PR #1, opened 2018-08-13
Remove svn tag

PaulWessel:
Get rid of $Id tags in files


CMakeLists.txt
```diff
@@ -1,6 +1,4 @@
 #
-#	$Id$
-#
```
leouieda: Yeah, that's great :tada:

PR #22, opened 2018-08-17
Templates for issue types and update PR template

leouieda:
Github allows using different templates for different types of issues. I
created one for bug reports and one for feature requests. We can add as
many as we want later ("questions" for example).

Update the PR template with a checklist of reminders to make sure the
code is following the general style, has documentation, tests, etc.


PaulWessel:
Good to go

PR #23, opened 2018-08-17
Update CMake README

PaulWessel:
The readme file for cmake discusses subversion.  I made the changes I think are needed for git.

The only question has to do with the GMT_SOURCE_CODE_CONTROL_VERSION_STRING since svn uses integers and git the hash?


leouieda:
I would suggestion using a git tag to mark the release commit. It's human readable at least and can be used to `git checkout` same way as a hash. 

That said, using the hash also works fine.

README.CMake
```diff
@@ -179,7 +179,7 @@ Set GMT_RELEASE_PREFIX in cmake/ConfigUser.cmake and run cmake. Then do
 You should then edit ${GMT_RELEASE_PREFIX}/cmake/ConfigDefault.cmake and
 set GMT_PACKAGE_VERSION_MAJOR, GMT_PACKAGE_VERSION_MINOR, and
 GMT_PACKAGE_VERSION_PATCH. Also uncomment and set
-GMT_SOURCE_CODE_CONTROL_VERSION_STRING to the current svn version. Then
+GMT_SOURCE_CODE_CONTROL_VERSION_STRING to the current git version. Then
```
joa-quim: But what would contain the GMT_SOURCE_CODE_CONTROL_VERSION_STRING? With SVN it was the revision number but in git we have only hash numbers.

PaulWessel: That is OK I think.  A release will say 6.0.0 but until then it says
pwessel@MacBeth:~-> gmt --version
6.0.0_r49baf49

joa-quim: Right, but when it's defined (gmt_version.h.in) it falls to ``@GMT_PACKAGE_VERSION_WITH_SVN_REVISION@``

```
#ifdef GMT_SOURCE_CODE_CONTROL_VERSION_STRING
#	define GMT_STRING "@GMT_PACKAGE_VERSION_WITH_SVN_REVISION@ (r@GMT_SOURCE_CODE_CONTROL_VERSION_STRING@)"
```

PaulWessel: Yes, we need to make some edits there as well.

leouieda:
It would be good to drop the leading `r` in the version string. The git hash has letters and numbers so that `r` might be mistaken for part of the hash.

seisman:
I tried to compile the master branch. It still gives a version string with a letter 'r'.
```
$ gmt --version
6.0.0_r63665d7
```
63665d7b6c1dbdca74cfc42c6549ed252838e336 is the hash of the latest commit.

PR #25, opened 2018-08-20
Backport fix to -I- option

joa-quim:
Do to not show the WKT info when it exists in a nc grid and -I- is set


PaulWessel:
Squashing and merging.

PR #30, opened 2018-08-21
Update GMT_Tutorial.rst

btozer:
missing "-" in W flag breaks GMT_tut_8. 
Low hanging fruit :-)


joa-quim:
Thanks for this catch.

PR #39, opened 2018-08-26
Update grdblend to work with memory files

PaulWessel:
grdblend was originally written to read row-by-row from any number of grids and blend overlapping rows to the output grid.  However, when used via an API some or all of the input grids may be in memory and grdblend had not been prepared for that yet.

It just needed some TLC to handle files that are actually memory grids and hence no need for row-by-row reading and freeing of grids and removal of temporary files.  Joaquim made most of the updates after my suggestions but left a few things hanging so it crashed.  I fixed these and tested the new build via gmtmex and it worked for Joaquim's example (his was in Julia but I can easier debug in Matlab):

G1 = gmt ('grdmath', '-R0/10/0/10 -I1 X Y');
G2 = gmt ('grdmath', '-R5/15/0/10 -I1 X Y');
G  = gmt ('grdblend', '-R0/20/0/10 -I1', G1, G2);

However, I did not actually check that the output is correct, just that it did no longer crash.

### DIFF
--- a/src/grdblend.c
+++ b/src/grdblend.c
@@ -267,15 +267,16 @@ GMT_LOCAL int init_blend_job (struct GMT_CTRL *GMT, char **files, unsigned int n
 	
 	for (n = 0; n < n_files; n++) {	/* Process each input grid */
 
-		if (L[n].download) {
+		if (L[n].download) {	/* Report that we will be downloading this SRTM tile */
 			char tile[8] = {""};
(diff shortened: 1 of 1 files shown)


joa-quim:
Confirm that ``G = gmt ('grdblend', '-R0/20/0/10 -I1', G1, G2);`` works but there seems to be some issue with the missing -R -I

PR #51, opened 2018-08-28
Fix grdimage crash when called from external interface

### LINKED ISSUE #50: grdimage crash when called from external interfaces
This Julia example
```
imshow("http://larryfire.files.wordpress.com/2009/07/untooned_jessicarabbit.jpg",
             frame="g", region="d", proj="I15c", image_in="r", show=true, Vd=1)
```
that translates to
```
grdimage  -Rd -JI15c -Bg -Dr -P http://larryfire.files.wordpress.com/2009/07/untooned_jessicarabbit.jpg -K > C:\tmp\GMTjl_tmp.ps
```
crashes at grdimage line 1346
```
	if (API->external && Ix && Iy) {	/* Restore old arrays since read-only image */
		gmt_M_free (GMT, I->x);	gmt_M_free (GMT, I->y);      // HERE
		I->x = Ix;	I->y = Iy;
	}
```
because the ``I`` object is NULL

joa-quim:
Fixes #50


leouieda:
Hi @joa-quim, I think the "Fixes #..." needs to be in the body of the PR.

joa-quim:
Thanks, I had an idea that it was something around this theme but not sure of the exact form.

PR #55, opened 2018-09-05
Add full in/out options to *ingrid* and *outgrid*

joa-quim:
But we have a problem. As it stands ``grdconvert`` says we need to state ``=ID`` in order to access to **+s**scale, etc. But that's not what we have in https://gmt.soest.hawaii.edu/doc/latest/GMT_Docs.html#grid-file-format-specifications where ``=ID``  is optional.

### DIFF
--- a/doc_classic/rst/source/grdconvert.rst
+++ b/doc_classic/rst/source/grdconvert.rst
@@ -32,7 +32,7 @@ be written and to specify scaling, translation, and NaN-value.
 Required Arguments
 ------------------
 
-*ingrdfile*
+*ingrdfile* [=id[+sscale][+ooffset][+ninvalid]]
     The grid file to be read. Append format =\ *id* code if not a
     standard COARDS-compliant netCDF grid file. If =\ *id* is set (see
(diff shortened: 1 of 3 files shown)


joa-quim:
But have you read the comments in PR body?

PaulWessel:
Sorry, no... What is the problem other than changing the docs to match reality?

joa-quim:
No problem, but not sure anymore what reality is. I remember that we visited this not long ago and I think we made ``=ID`` optional but it has to be provided when read/write native binary formats, so not so optional.

PaulWessel:
Think it says(?) optional if reading the default netCDF grid format.

joa-quim:
The point is. Should in ``grdconvert`` we remove the part that says ``If =id is set (see below),``?
Although true more most cases, that is a contradiction to saying that ``=ID`` is optional.

joa-quim:
This is clearly out of scope of this (closed) PR but since I started it... The CookBook section is badly incomplete as it misses the ``:diver`` option that is used when saving grids in GeoTiff floats or JPK2.

PR #60, opened 2018-09-08
Stop doing a "verify_sharedir_version"

joa-quim:
Since we hard-coded most of the postscript stuff we don't need this check anymore, which can even be harmfull on Win if no /DNO_SHAREDIR_VERIFY is used and only the gmt dll is updated.


PaulWessel:
The test is now failing; perhaps you can clean that part up first.

joa-quim:
Damn it. I did this PR directly from the git page and ofc didn't bother to check it builds. Seems, that I have to start over ... but latter.

PR #62, opened 2018-09-09
Improve a few doc pages

PaulWessel:
Joaquim was unhappy with how some options were documented in spectrum1d, gmt2kml, and grdimage.  I could not find the problem in grd2kml.

### DIFF
--- a/doc_classic/rst/source/gmt2kml.rst
+++ b/doc_classic/rst/source/gmt2kml.rst
@@ -15,7 +15,7 @@ Synopsis
 
 **gmt2kml** [ *table* ]
 [ |-A|\ **a**\ \|\ **g**\ \|\ **s**\ [*alt*\ \|\ **x**\ *scale*] ]
-[ |-C|\ *cpt* ] [ |-D|\ *descriptfile* ] [ |-E|\ [*altitude*] ]
+[ |-C|\ *cpt* ] [ |-D|\ *descriptfile* ] [ |-E| ]
 [ |-F|\ **e**\ \|\ **s**\ \|\ **t**\ \|\ **l**\ \|\ **p**\ \|\ **w** ]
 [ |-G|\ **f\|n**\ **-**\ \|\ *fill* ]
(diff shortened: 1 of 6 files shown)


doc_classic/rst/source/gmt2kml.rst
```diff
@@ -108,7 +108,7 @@ Optional Arguments
 
 .. _-E:
 
-**-E**\ [*altitude*]
+**-E**
     Extrude feature down to ground level [no extrusion].
 
 .. _-F:
```
joa-quim: Need to fix also the end of -N description that ends exactly like this

```
appearing where %d is placed, (4) give no arguments to turn symbol labeling off; line segments will still be named. Also note that
```

PaulWessel: Please be specific where this is since searching for this text in all *.c files returns no matches.  What file are we talking about here?  Your original message said grd2kml but now you say gmt2kml, but in either case I cannot find what you are listing.

joa-quim: It's man page of gmt2kml -N option (https://gmt.soest.hawaii.edu/doc/latest/gmt2kml.html#n)

```

-N[t|col |name_template|name]
    By default, if segment headers contain a -L“label string” then we use that for the name of the KML feature (polygon, line segment or set of symbols). Default names for these segments are
“Line %d” and “Point Set %d”, depending on the feature, where %d is a sequence number of
line segments within a file. Each point within a line segment will be named after the line
segment plus a sequence number. Default is simply “Point %d”. Alternatively, select one of
these options: (1) append col to supply individual symbol labels as the string formatted from
the col data column, (2) append t to let individual symbol labels be the trailing text of each
record (3) append a string that may include %d or a similar integer format to assign unique
name IDs for each feature, with the segment number (for lines and polygons) or point number
(symbols) appearing where %d is placed, (4) give no arguments to turn symbol labeling off; line
segments will still be named. Also note that 
```

Issue #45, opened 2018-08-27
Rename README.Cmake


seisman:
Currently, there are two README files in the repo: README.md and README.Cmake. GitHub display the content of README.Cmake at the homepage of the repo. However, README.Cmake is too technical, and README.md is more useful for potential new users. 

IMHO we should rename README.Cmake to make GitHub display README.md instead.

leouieda:
Hi @seisman, yeah I noticed that. I thought it would be good to morph the README.cmake into a CONTRIBUTING.md. We would need to add some things about the Github process and code quality, which we can copy from the GMT/Python CONTRIBUTING.md.

remkos:
Best approach would be to merge README.CMake into README.md. There is already some overlap between the two.

joa-quim:
I think we can this one now

PR #63, opened 2018-09-10
Merge README.CMake with README.md

remkos:
This addressed issue #45


remkos:
Hi @joa-quim. Here's another approach: leaving README.md more or less the same and replacing README.CMake with Building_GMT_with_CMake.md linked from the former.

remkos:
@joa-quim : You can see the rendered version here: https://github.com/GenericMappingTools/gmt/blob/master/Building_GMT_with_CMake.md

joa-quim:
Yes, now that it is merged I can but before it seemed to be in a branch of yours that I could not find in the main page branches list. Probably just ignorance.

leouieda:
> This welcome bot sometimes is a bit dumb. So now it's my first PR merge!

@joa-quim it's Remko's first PR. Note: the texts that the bot uses can be configured in the file `gmt/.github/config.yml` so we could have it say whatever we want (in case there is any GMT humor that you might want to introduce).

PR #64, opened 2018-09-12
Supress warnings on Windows

joa-quim:
When I build the Win 32 bits version I get tons of warnings ``conversion from 'uint64_t' to '::size_t', possible loss of data``. This PR shuts it down, but the question is: should we replace the 'uint64_t' by '::size_t', or are there instances where we don't want that?

### DIFF
--- a/src/gmt_dev.h
+++ b/src/gmt_dev.h
@@ -77,6 +77,11 @@ extern "C" {
 #endif
 #endif
 
+#ifdef _MSC_VER
+#	pragma warning( disable : 4091 )	/* 'static ': ignored on left of 'XXX' when no variable is declared */
+#	pragma warning( disable : 4244 )	/* conversion from 'uint64_t' to '::size_t', possible loss of data */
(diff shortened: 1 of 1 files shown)


PaulWessel:
This is a good question for Florian I think.  Of course, I don't really understand what a 32-bit version is.  Did the 1990s call and request a build? Joke aside, perhaps the better solution is to do an ifdef to determine if 32-bit and then typedef or define uint64_t to be size_t?

joa-quim:
Unfortunately, I'm stuck to use a GMT32 in the stand-alone Mirone (old compiler reasons). And since GDAL has moved to VS14+ only I am forced to rebuild all dependencies because for gmt6 I want to build with this compiler. So, given that we have 2 GMT versions (not counting GMT4), I have now to maintain:

- (32 & 64) x (GMT5 & master) x (VS12 & VS14) = 8 different builds.

Given this, this PR is more than good enough for me.

PR #69, opened 2018-09-16
Revert to old shorten path

PaulWessel:
The psl_shorten_path function has a bug so I am reverting to the old algorithm while we work on a solution.

**The problem**

given this file (t.txt) with content

0.1486E+03 0.1429E-04
0.1486E+03 0.3521E-04
0.1486E+03 0.3167E-04
0.1486E+03 0.2029E-04
0.1404E+03 0.5148E-05

the following command

gmt psxy -JX15/10.6l -R140/150/4.285e-06/0.0045  t.txt > gmt5.ps

simplifies the path to the point that the critical 2nd point is excluded.

Until we can fix the problem I have added "#define OLD_shorten_path" to postscriptlight.c so that we rely on the older (slightly slower but correct) line simplification algorithm.

### DIFF
--- a/src/postscriptlight.c
+++ b/src/postscriptlight.c
@@ -104,8 +104,6 @@
  *
  */
 
-#define OLD_shorten_path	/* Until bug in psl_shorten_path is fixed */
-
 /*--------------------------------------------------------------------
  *			SYSTEM HEADER FILES
  *--------------------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


remkos:
Instead of reverting, I have a fix using commit 5fdeb07
I can send that as a push request if that's more convenient.

PaulWessel:
Looks OK, tell me what to do to include it...

remkos:
Sorry, went to the wrong place. Please check out my revision above.

PR #74, opened 2018-09-19
Ensure geographic projection in pscoast

### LINKED ISSUE #73: pscoast -E gets confused by -R
Works
```
pscoast -JX12c/0 -Ba -EAR+p0.25p+gblue -P -K > lixo.ps
```
but not if we set -R manually
```
pscoast  -R-80/-50/-55/-20 -JX12c/0 -Ba -EAR+p0.25p+gblue -P -K > lixo.ps
pscoast [ERROR]: You must use a map projection or -Jx|X...d[/...d] for geographic data
```

PaulWessel:
Instead of making people remember to add "d" as in -JX15cd for degrees, just add it if the user forgot as it is easy to do.

Fixes #73

### DIFF
--- a/src/pscoast.c
+++ b/src/pscoast.c
@@ -768,23 +768,22 @@ int GMT_pscoast (void *V_API, int mode, void *args) {
 		GMT->common.R.wesn[XHI] -= 360.0;
 	}
 
-	if (Ctrl->Q.active && !GMT->common.J.active) {	/* Fake area and linear projection */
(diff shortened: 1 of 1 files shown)


joa-quim:
... due to a bug in a GMT.jl module. I have there a function that
```
cmd, opt_B, opt_J, opt_R = parse_BJR(d, "", "", O, " -JX12c/0")
```
where last argument is the default -J. Many copy-pastes and forgot to change the default to " -JX12cd/0" in pscoast (but pssolar was ok). Then, when it errored in GMT5.4 I just copy pasted the "should be working version".

PR #112, opened 2018-10-09
Groundwork for new insert module

PaulWessel:
To make figure inserts in modern mode

This will be used to add small figure inserts to the current figure.  Complies but does not yet work.


PaulWessel:
Not yet anyway. But will require R J of course to have been set since it is an overlay.

seisman:
@PaulWessel A quick question. As a non-native English speaker, I'm a little confused with the word "insert". It seems people usually call such a feature as "inset map". Is "figure inserts" correct?

PaulWessel:
Hi @seisman I think you are correct.  We'll use inset.

joa-quim:
I was referring to THIS_MODULE_KEYS (to external interfaces), not to the THIS_MODULE_NEEDS.

PaulWessel:
No data files used so KEYS are not needed.

PaulWessel:
FYI, gmt inset begin would write "V" to the PS file and gmt inset end writes "U"...  But yes, it is writing to the PS file.

leouieda:
Looks good to me. I agree with @seisman. Also `gmt insert` would easily be confused with other things like inserting values into grids or something like that. 

@joa-quim please refrain from merging things right away when there are still comments/changes to be made (like renaming the entire module or your comment about the THIS_MODULE_KEYS). That kind of defeats the purpose of going through a pull request and review. This way we can avoid having broken or incomplete commits in the history.

joa-quim:
Hmm, the 'talk' stated after the merge so I could not know that by the time I merged it.

leouieda:
Exactly. So unless it's a trivial fix it's best to wait for comments or confirm with the author before merging.

PR #122, opened 2018-10-14
Fix compiler warnings in grdppa.c

remkos:
Because I got a lot of warnings in compiling grdppa.c
Biggest problem was the missing `#include gmt_internals.h`

### DIFF
--- a/src/misc/grdppa.c
+++ b/src/misc/grdppa.c
@@ -26,7 +26,6 @@
  * ********************************************************************* */
 
 #include "gmt_dev.h"
-#include "gmt_internals.h"
 
 #define THIS_MODULE_NAME	"grdppa"
 #define THIS_MODULE_LIB		"misc"
@@ -37,6 +36,8 @@
 
 #define ij(h,i,j) ((i) + ((j)-1)*(h->mx) -1)
 #define ijk(h,i,j,k) ((i) + ((j)-1)*(h->mx) + ((k)-1)*(h->mx)*(h->my) -1)
+
+EXTERN_MSC void gmtlib_grd_flip_vertical (void *gridp, const unsigned n_cols, const unsigned n_rows, const unsigned n_stride, size_t cell_size);
 
 /* Control structure */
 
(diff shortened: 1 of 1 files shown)


PaulWessel:
So far, only gmt_*.c library files include gmt_internals.h and none of the modules do.  What does grdppa need that is not accessible via gmt_prototypes.h (which is included in gmt_dev.h for module use)?  TO keep this separation up we sometimes promote functions that are in gmt_internals.h to gmt_prototypes.h so that we don't need to add one-time EXTERM_MSC lines in the modules.  As I write this, gmt_prototypes.h is not a great self-intuitive name for what it means...

joa-quim:
Hmm, but I don't need to ``#include "gmt_internals.h"`` and neither needs CI because the PRs passed the build step.

remkos:
It does compile without `#include gmt_internals.h`, but then there is still a (reasonable) warning:
<pre>
[ 70%] Building C object src/CMakeFiles/supplib.dir/misc/grdppa.c.o
src/misc/grdppa.c:503:2: warning: implicit declaration of function 'gmtlib_grd_flip_vertical' is invalid
      in C99 [-Wimplicit-function-declaration]
        gmtlib_grd_flip_vertical (G->data, G->header->mx, G->header->my, 0, sizeof(G->data[0]));
        ^
1 warning generated.
</pre>
Obviously, `gmtlib_grd_flip_vertical` needs a prototype. I'll add it to `gmt_prototypes.h` then.

remkos:
In fact, no ... if you use a `gmtlib` routine, there is no way avoiding `#include gmt_internals.h` or `EXTERN_MSC`.

PaulWessel:
I guess this depends on whether there is any benefit to us to keep functions only used within the GMT library as a separate include file (gmt_internals.h).  If we were to include gmt_internals.h in gmt_dev.h, why even have separate gmt_prototypes.h and gmt_internals.h files?  Isn't it better to rename any gmtlib_* function to gmt_* and promote to gmt_prototypes.h instead?

joa-quim:
Yes, promoting it to gmt_* would be better. It's not rare that I need to do this operation to grids (mem layouts and stuff like that).

PR #136, opened 2018-10-21
Handle unit for vector shrinking in grdvector

PaulWessel:
The vector attribute processing keeps track of normalizing length and unit and applies the conversion later since some programs (like psxy) may read such instructions from the command line.  In grdcontour, however, there is no such reading and we must convert any length to inches before plotting.

The initial problem was detected by @joa-quim when comparing
-Q0.1i+e+n0.63c
-Q0.1i+e+n0.25c
which should be identical.  This PR makes them so.


joa-quim:
I'm afraid we have more of these. From ``endarrows.sh`` test
```
gmt psxy -R -J -W1p+o1.25/0.25i+v0.25i+gred -O t.txt -K >> $ps 
```
also shows a similar bug.

PR #140, opened 2018-10-22
Remove -dSAFER

joa-quim:
Back-port change from master where -dSAFER was removed because it now prevents transparency extension.


joa-quim:
Oh no, not again.

PR #141, opened 2018-10-22
Remove -dSAFER

joa-quim:
Back-port from master.  Remove -dSAFER because it now prevents transparency extension in ghostscript >= 9.25


PaulWessel:
Thanks for doing that.

PR #155, opened 2018-10-28
Set default -I in grdraster

### LINKED ISSUE #153: grdraster problem in 5.4.4, where is grdraster in GMT 6?
The MB-System program mbgrid has a longstanding capability to underlay a bathymetry or topography grid with  a global  or  regional topography model extracted using grdraster. With GMT 5.4.4 this capability is broken.

The problem is that grdraster is requiring the -Iincrement option when it used to use the increment defined in the source global model by default. For instance, the command

   % grdraster 28 -R-13.775815/-13.702018/53.549101/53.588000 -Gtmpgrdraster11839.grd

just yields the error message:

    grdraster: No increment given and no -I in effect.  Cannot initialize new grid

I tried to see if this is still an issue in the current code. However, when I cloned gmt-master, I no longer find grdraster defined anywhere in the codebase. Am I missing something, or is grdraster disappearing?

I'm working on MacOs 10.13.06.

PaulWessel:
Tnis had gotten lost in the porting.  Now selects the increment from the info file if -I not given.

Fixes #153 in 5.4.


seisman:
@PaulWessel You should not add gmt.history and tmpgrdraster11839.grd to the repo.

PaulWessel:
Sorry, should be gone now.

PR #160, opened 2018-10-30
Clarify some aspects about color usage in vectors.

joa-quim:


PaulWessel:
Please change "imply" to "implies" in all those places first.

Issue #53, opened 2018-08-29
Sort and separate the tests scripts


leouieda:
GMT has a lot of test scripts that create and compare plots. What we don't have is a measure of how many modules these tests call and how much of the code is covered. The tests also take a long time to run so they're not good for running on TravisCI (we don't want to wait 40min for a build to finish).

Quoting @PaulWessel s email:

> See plot of test durations (log-scale in seconds).  So most are around a few seconds (1-10) with a few slow ones (70 s); this is on my laptop.  I also attach the sorted list (time, test).  Before I do anything else we can talk about what might be useful? Make a plot of “coverage” somehow, meaning how many times any gmt module is called within the 812 tests?  I think we wanted a very basic set that at least ran all modules once on a minimal data set.

![test-time-distribution](https://user-images.githubusercontent.com/290082/44814161-8542b700-ab77-11e8-8f9f-de6cde740c47.png)
Distribution of test times.

Here is a file with the timing information for each test script: [test_sorted.txt](https://github.com/GenericMappingTools/gmt/files/2333715/test_sorted.txt)

leouieda:
To run these things on CI, it would be good to be able to specify different levels of tests that we want to run. Tests that are run on every build should be fast, ideally taking no more than 10min to run in total. We can run the full test suite with the slower tests on a nightly cron job. So it would help if we could "make test-fast" and  "make test-full", for example.

leouieda:
cc @joa-quim @seisman

seisman:
Inspired by https://github.com/obspy/obspy_github_api and https://github.com/obspy/obspy/blob/master/.travis.yml. 

Maybe we can separate it into several tests, and use some magic comment strings like `+DOCS`, `+TESTS:FULL`, `+TESTS:0` to trigger specific tests.

joa-quim:
Closing on a *too old and stalled* criteria.

PR #175, opened 2018-11-09
Build GMT docs

seisman:
This PR adds GMT documentations builds. Since the sphinx of travis's OS (Ubuntu 14.04) is too old, I install python and sphinx via miniconda.


joa-quim:
Does this mean that the docs are going to be (re)build for every commit?

seisman:
@joa-quim Yes.

joa-quim:
So, I'm afraid this is not such a good idea. Building the docs is a quite lengthy process and I do not see why a commit in the code would trigger it when the the two are not related.

seisman:
As mentioned in #53, it's possible to use some magic strings like `+DOCS` to dynamically trigger the documentations build.

PaulWessel:
@leouieda and I have discussed this. Using this feature depends on how quick it will be to build.  As it is now, all illustrations as stored as PS files and there is a conversion to PNG (and maybe PDF) taking place during the docs build.  We may find that we should also store the PNGs to avoid this step except when the PS has changed.  If it takes too long to build then we may wish to make such changes.

PaulWessel:
It is not possible to know for sure that a code change does not have an effect on the docs.  if we fit a bug in surface or a plot tool it may cause small changes in the plots.

seisman:
@PaulWessel @joa-quim Travis supports cron jobs. Currently, the master branch will be built every month. Maybe we can change the cron job settings and build the master branch everyday. 

If a build is triggered by a cron job, then we build the docs and also tests. Otherwise, we only build the sources, which is much faster.

seisman:
The [travis docs](https://docs.travis-ci.com/user/cron-jobs/#detecting-builds-triggered-by-cron) say that:

    To check whether a build was triggered by cron, 
    examine the TRAVIS_EVENT_TYPE environment variable to see if it has the value cron.

PaulWessel:
Perhaps a cron job is an acceptable solution.  I have gotten used to running the full tests before I submit a pull requests.  On my 24-core Mac the tests take ~3 minutes.  Building the entire docs from scratch is a bit faster since there are fewer PS to PNG processes; it takes 2.5 minutes.  If my patch breaks something in the tests I can then fix before finalizing the PR.  The cron job would instead be helpful to make sure we have a pretty recent version of the full docs available for users to browse. Maybe weekly cron?

leouieda:
> Bur aren't the tests exactly for that?

You're right. Building the docs always is probably not a great solution.

@seisman adding an `if` that checks for `TRAVIS_EVENT_TYPE` would be enough to make it trigger only in cron jobs. I think Travis can probably handle a nightly build.

.travis.yml
```diff
@@ -19,14 +19,32 @@ env:
         - COASTLINEDIR="$HOME/gmt-install-dir/coast"
         - PATH="$INSTALLDIR/bin:$PATH"
         - LD_LIBRARY_PATH="$INSTALLDIR/lib:$LD_LIBRARY_PATH"
+        - TRAVIS_EVENT_TYPE=cron
+        - if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
+            export PYTHON=3.6
+            export CONDA_REQUIREMENTS="ci/requirements.txt"
```
leouieda: No need for a requirements file. You can set `CONDA_INSTALL_EXTRA="sphinx"` (see [here](https://github.com/fatiando/continuous-integration/blob/master/.travis.yml#L52)).

.travis.yml
```diff
@@ -19,14 +19,32 @@ env:
...
+            export BUILD_DOCS=true
+          fi
 
 before_install:
     # Install GMT dependencies
     - bash ci/travis-setup.sh
+    - if [ "$BUILD_DOCS" == "true" ]; then
+        # Install GMT documentation dependencies
+        # Get the Fatiando CI scripts
```
leouieda: Travis doesn't like these comments inside blocks. You have to put them out of the `if` or they can't parse the yml file.

.travis.yml
```diff
@@ -19,14 +19,32 @@ env:
...
+        # Need to source the script to set the PATH variable globaly
+        source continuous-integration/travis/setup-miniconda.sh
+      fi
 
 install:
     # Build and install GMT
     - bash ci/travis-build.sh;
+    - if [ "$BUILD_DOCS" == "true" ]; then
+        # Build GMT documentations
+        bash ci/travis-build-docs.sh;
```
leouieda: Do we need a script for this?

leouieda: The build should probably be part of the `script` block as well since this is pretty much part of the tests.

.travis.yml
```diff
@@ -19,14 +19,32 @@ env:
         - COASTLINEDIR="$HOME/gmt-install-dir/coast"
         - PATH="$INSTALLDIR/bin:$PATH"
         - LD_LIBRARY_PATH="$INSTALLDIR/lib:$LD_LIBRARY_PATH"
+        - TRAVIS_EVENT_TYPE=cron
+        - if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
+            export PYTHON=3.6
+            export CONDA_REQUIREMENTS="ci/requirements.txt"
+            export BUILD_DOCS=true
```
leouieda:
```suggestion

```

Probably no need for this variable as well since we're only building on cron jobs.

.travis.yml
```diff
@@ -19,14 +19,32 @@ env:
...
+          fi
 
 before_install:
     # Install GMT dependencies
     - bash ci/travis-setup.sh
+    - if [ "$BUILD_DOCS" == "true" ]; then
```
leouieda:
```suggestion
    - if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
```

Could probably merge this block with the one further up. The `env` block is for setting the build matrix.

.travis.yml
```diff
@@ -19,14 +19,32 @@ env:
...
+        git clone https://github.com/fatiando/continuous-integration.git
+        # Download and install miniconda and setup dependencies
+        # Need to source the script to set the PATH variable globaly
+        source continuous-integration/travis/setup-miniconda.sh
+      fi
 
 install:
     # Build and install GMT
     - bash ci/travis-build.sh;
+    - if [ "$BUILD_DOCS" == "true" ]; then
```
leouieda:
```suggestion
    - if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
```

seisman:
@leouieda I've made changes as requested. Please review it again.

leouieda:
@seisman looks good to me. The docs seem to build without error and it takes 223s so not too bad. Should we set this to run nightly?

leouieda:
This is good to merge if @seisman has nothing more to add.

seisman:
> Should we set this to run nightly?

@leouieda I've set a nightly build. Feel free to set it weekly or monthly.

leouieda:
Perfect!

PR #183, opened 2018-11-16
kml2gmt, add KEYS for reading

joa-quim:
Missed the reading KEY. Avoid some failing tests for external but still does't work. It now crashes at line 240 when trying to read from ``fp``
```
while (fgets (line, GMT_BUFSIZ, fp)) {
```
because data is not it ``fp``


PaulWessel:
There were no KEY for input because we are not using the GMT i/o system and hence all the stuff about memory and passing info cannot work.  The input KML is not a GMT container.

PaulWessel:
I mean, the entire read section would have to use GMT_Get_Record and work with the trailing text part only.  It will be a major rewrite.  Do you have KML in memory that you need to pass into kml2gmt that way and cannot use a file?

joa-quim:
No, I'm just blindly converting all GMT modules to work from Julia.
But note that gmt2kml works (don't know what to do with the output, but it's there)
```
julia> D = gmt("pscoast -R-15/2/50/59:30 -Jm1i -M -W0.25p -Di");

julia> Dk = gmt2kml(D, F=:l, W=(1,:red))
1-element Array{GMT.GMTdataset,1}:
 GMT.GMTdataset(Array{Float64}(9222,0), Any["<?xml version=\"1.0\" encoding=\"UTF-8\"?>", "<kml xmlns=\"http://www.opengis.net/kml/2.2\">", "\t<Document>", "\t\t<Style id=\"st-39008-0\">", "\t\t\t<IconStyle>", "\t\t\t\t<Icon>", "\t\t\t\t\t<href>http://maps.google.com/mapfiles/kml/pal4/icon57.png</href>", "\t\t\t\t</Icon>", "\t\t\t\t<scale>1</scale>", "\t\t\t\t<color>ff80c0ff</color>"  …  "\t\t\t\t\t-3.14053559167,54.5825131609,0", "\t\t\t\t\t-3.16556038758,54.5603112841,0", "\t\t\t\t\t-3.156099794,54.5977721828,0", "\t\t\t\t\t-3.14442664225,54.5866330968,0", "\t\t\t\t\t-3.14053559167,54.5825131609,0", "\t\t\t\t</coordinates>", "\t\t\t</LineString>", "\t\t</Placemark>", "\t</Document>", "</kml>"], "", Any[], "", "")
```

PaulWessel:
Perhaps you should just delete this PR since there is much more to do than changing the KEYS.  Maybe make it an issue instead?

joa-quim:
Opened #976 to keep track of this

PR #188, opened 2018-11-26
Address topic http://gmt.soest.hawaii.edu/boards/1/topics/7807

PaulWessel:
While the longitudes where correctly inferred to be pixel registered, disaster struck for latitudes...
Should be a bit more foolproof now.

### DIFF
--- a/src/gmt_nc.c
+++ b/src/gmt_nc.c
@@ -627,8 +627,10 @@ GMT_LOCAL int gmtnc_grd_info (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *head
 		}
 		else if (has_vector) {	/* No attribute for range, use coordinates */
 			dummy[0] = xy[0], dummy[1] = xy[header->n_columns-1];
-			if (set_reg && fabs (fmod (dummy[0], dx)) > (0.4999 * dx))	/* Pixel registration */
+			if (set_reg && fabs (fmod (dummy[0], dx)) > (0.4999 * dx)) {	/* Most likely pixel registration since off by dx/2 */
 				registration = header->registration = GMT_GRID_PIXEL_REG;
+				dummy[0] -= 0.5 * dx;	dummy[1] += 0.5 * dx;
+			}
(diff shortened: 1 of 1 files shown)


joa-quim:
The problem is that we rely on the existence of "node_offset" or "node_offset" or "valid_range" or "valid_min"/"valid_max" to guess if data is grid or pixel but if none exists, we assume data is pixel reg.

PaulWessel:
Given the original grid actually says the x-values are 0.25, .75, ..., 359.75 it is admitting it is pixel registered, which matches the even number of nodes.  Even more so that min = -89.25 so I think only pixel registration makes sense.

joa-quim:
Yes, it has an even number of nodes but that doesn't necessarily make it more pixelated

```
length(0.25:0.5:359.75)
720
```

For this data, why a pixel reg -R[0 360] makes more sense then a grid reg of -R[0.25 359.75]?

```
C:\v>ncdump -h result_ice_repository.grd
netcdf result_ice_repository {
dimensions:
        lon = 720 ;
        lat = 360 ;
        time = 113 ;
```

PaulWessel:
Since we know it is a global grid it should be 360 by 180 degrees.  I would say 0/360/-90/90 as pixel reg make much more sense than 0.25/359.75/-89.75/89.75 as gridline reg.  Same file.

joa-quim:
OK, I by the global argument but the gmt_nc code is not basing its choice on a test for global regions.

PaulWessel:
No, it is basing it on xmin/dx giving a remainder of 0.5 *dx which is normally a good test to guess pixel registration.  It cannot know for sure of course, so a grid form 0.25 to 9.75 with dx 0.5 could be 0/10 pixel or 0.25/9.75 gridline. Hard to know what the user meant without more info in the header, but it is still a reasonable guess.  However, when the adjusted min/max gives a 360 range then I think we can be very sure.  I don't think we would treat these two cases differently but we certainly would be more confident in the latter case and could print a long-verbose message to that effect.

PaulWessel:
Note that the original grid fails to use COARDS convention for longitude attribute: It has "degrees east" instead of "degrees_east".  So it is also seen as a Cartesian grid.
Of course, it should not stop GMT from perhaps still figuring this out that it is geographic.

PR #189, opened 2018-11-26
Add test to check on http://gmt.soest.hawaii.edu/boards/1/topics/7776

PaulWessel:
Currently this test has axes labels that are rotated 180 with respect to axis annotations.  This PR just adds a new test (without any original PS) that fails.


PaulWessel:
We have always added tests from failures, then fix them so the test passes.  it gives assurance that if we screw up down the road then the test may once again fail and remind us of that.  Otherwise we may wait a long time for someone else to notice and then it is deja vu all over again.

leouieda:
A way to have the tests but not add extra failures is to include the test
in the PR that fixes the bug.

PR #192, opened 2018-11-27
Fix reading beyond band 9 of when only a single band is selected.

joa-quim:
Addresses an issue raised in the [forum ](https://gmt.soest.hawaii.edu/boards/1/topics/7807) where reading beyond band 9 was in fact reading band 0.
However, debugging this case revealed that band selections beyond 3'rth band for images that have more than 3 bands will also be wrong.


joa-quim:
What it does it to let you select the bands of an RGB image in any order and recombine them. But the logic breaks beyond 3 bands.

PR #196, opened 2018-12-02
Finalize all character adjustments for UTF-8

PaulWessel:
This PR includes the 7 missing characters not handled previously: ae, AE, Ydieresis, and the s,z,S,Z carons.  With this, all European letters present in the ISOLatin1+ encoding will be automatically converted from UTF-8 to their equivalent octal codes in our  character encoding tables.

### DIFF
--- a/doc_classic/rst/source/GMT_Docs.rst
+++ b/doc_classic/rst/source/GMT_Docs.rst
@@ -3150,6 +3150,10 @@ you must use the full octal code):
 | @s       | ß          | @i       | í          |
 +----------+------------+----------+------------+
 
+However, if your input text contains UTF-8 code characters (e.g., ü, Î)
+and you select the ISOLatin1+ character encoding then GMT will substitute
+the correct PostScript octal codes for you automatically.
+
 PostScript fonts used in GMT may be re-encoded to include several
 accented characters used in many European languages. To access these,
 you must specify the full octal code \\xxx allowed for
--- a/doc_modern/rst/source/GMT_Docs.rst
+++ b/doc_modern/rst/source/GMT_Docs.rst
@@ -3090,6 +3090,10 @@ you must use the full octal code):
 | @s       | ß          | @i       | í          |
 +----------+------------+----------+------------+
 
+However, if your input text contains UTF-8 code characters (e.g., ü, Î)
+and you select the ISOLatin1+ character encoding then GMT will substitute
+the correct PostScript octal codes for you automatically.
+
 PostScript fonts used in GMT may be re-encoded to include several
 accented characters used in many European languages. To access these,
 you must specify the full octal code \\xxx allowed for
--- a/src/postscriptlight.c
+++ b/src/postscriptlight.c
@@ -649,9 +649,11 @@ void psl_fix_utf8 (struct PSL_CTRL *PSL, char *string) {
 				strcat (out, tmp);
 			}
 		}
-		else if ((unsigned char)(string[k]) == 0305) {    /* Found Ydieresis, ae, AE, the S,Z,s,z carons */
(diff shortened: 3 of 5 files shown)


PaulWessel:
Yes, I do think that makes more sense.  But we should have some idea of what, if any, backwards issues it may cause?

joa-quim:
There may be. One kind of brute test is to change it and see if we get user complains.

PaulWessel:
Right. So it is ISO for SI units and Standard for US units.  I don't think it makes that much sense to associate measurement unit with font character set defaults.  I will set it to ISOLatin1+ and we can check.

PaulWessel:
Hundreds of minor failures because our originals use Standard+.  The differences are:

1. Negative numbers look different.  Per the encodings, we are actually using the hyphen and not the minus sign in plotted numbers when we use Standard instead of ISO, so we could declare a bug for that.
3. Some titles and label strings are slightly differently adjusted or at a different distance from the base line.  Not sure why that would be other than the PostScript operators that compute heights and widths do return slightly different values for the different but very similar looking character sets.
4. Actual hyphens.  On my keyboard, the numerical keyboard minus sign and my hyphen are the same character.  Per the encoding these are two separate things.  So right now, text with hyphens will in ISO come out with minus signs.  We may be able to control that under the hood since we usually format numbers (and hence can ensure we use minus) but not sure how we can "detect" the user meant a hyphen and replace with the actual ISO hyphen (octal 255).  Can you test if your keyboards prints different ASCII codes for your minus and hyphen keys?
5. Here and there the wrong octal code is used in pstext etc. We need a short-cut to place a degree symbol in other text since it is too fragile to use 312 or 260 and then when someone changes the charset we get the wrong symbol.  Maybe @* could mean degree and then we insert the correct degree symbol given the charset? Related to Degree symbols: In ISO we use the symbol called "degree" while in Standard we have used the symbol called "ring", which when combined with a or A spells my Norwegian name Pål.  It is the wrong symbol, hence we can declare a bug.

The diff PNG 
datagrep
plot shows the title adjustment and minus sign.

Actions:
Need to do some research and test regarding how to get hyphen symbol instead of minus in text when ISO is used.  I.e., how do you generate a hyphen symbol on European keyboards and is it different from minus? [See our appendix on ISOLatin1+ char plot and 035 versus 255.
Introduce @* as code for character-set-independent degree symbol (will select the correct "degree" symbol).
Use the proper "minus" symbol internally when formatting negative number and not just "-" which difference with character sets.

joa-quim:
My laptop's keyboard and I think, most or all, keyboards have only the minus character. I don't even know how to type an hyphen without resorting to some code. I guess that the utf8 hyphen can be detectable by your code and we should simply resort on that. that is, without further effort all '-' are minus.

PaulWessel:
Well, I was thinking of a different approach:

1. When GMT is plotting annotations etc. we know we are dealing with numbers and hence we select the minus code. That code differs across char sets.  This ensures uniformity and compliance to proper math type setting.
2. For free-standing text provided via pstext or -B etc, we are not concerned with formatting numerical data and any "-" character will be considered to be a hyphen and we insert the correct hyphen code, which also differs across char sets.  This also ensures consistency and avoid using very wide dashes (en-dash is very similar to math minus) in hyphenated words.
3. If you are very picky and want to write -200 in some text string and want the minus symbol then you can use the octal code for perfection.

Clearly, not super-easy to generate this from the keyboard (at least on a mac): https://apple.stackexchange.com/questions/39448/how-do-you-type-a-true-minus-sign-not-a-hyphen-on-a-mac

joa-quim:
OK, just learnt that we have in keyboards are hyphens, not minus signs.

PaulWessel:
So that is good in that users are typing and expecting hyphens, unless they use magic to get the minus sign. And for numbers in annotation we can control the use of minus.

PR #230, opened 2018-12-21
Explore adding long-format GMT options

PaulWessel:
This branch is work in progress on adding long-format option parsing to GMT.  E.g., --region=w/e/s/n instead of -Rw/e/s/n, --registration=pixel instead of -r, and so on.  At the moment only the common options and their most basic forms have been implemented, as well as local options for blockmean.
I will write more later about how this all works and what is required to be added to each module - blockmean.c static variable local_kw shows how.  Basically, for each short-format option we need to supply the corresponding long-form names.  Many GMT options support modifiers of the form +<letter>[<args> which in long form will be +<modifier-name>[=<arg>].  Finally, many GMT options have a directive character immediately following the option key and these will become <directive>[:<argument>] in long form.  Thus, what is parsed and converted to short-format options are optinos like

--longoption[=[<directive>:]<arg>][+<mod1[=<arg1>]][+<mod2[=<arg2>]]...


PaulWessel:
Sorry, got formatted.  The general thing I am parsing is of the form

`--longoption[=[<directive>:]<arg>][+<mod1[=<arg1>]][+<mod2[=<arg2>]]...
`

which will represent the vast number of module options.

PaulWessel:
Here I need help, @leouieda since the two conflicting files where deleted from master after the branch was created.  So they are in the branch but no longer in master.  We had this awhile earlier with another branch and you did the magic...

leouieda:
@PaulWessel should they be deleted from this branch as well?

PaulWessel:
Yes, I guess that would allow the merge later.

leouieda:
Done

PaulWessel:
Change KW_DIC to KEYWORD_DICTIONARY

PaulWessel:
I have made changes so that this branch can now be merged into master.  None of the work on long-options are exposed unless two ConfigUser.cmake add-definitions settings are added.  Hence, the branch passes the test and slowly adding the structure-arrays with keywords/values are simpler if in master.  Specifically, if you add

`add_definitions(-DUSE_COMMON_LONG_OPTIONS) 	# Turn on testing of upcoming long-option syntax for common GMT options
`

and build then this works:

`gmt basemap --region=0/40/0/50 --projection=M6i --frame --timestamp="My Map" -pdf map
`

and if you add

`add_definitions(-DUSE_MODULE_LONG_OPTIONS) 	# Turn on testing of upcoming long-option syntax for module options
`

then the block mean,mode,median modules can do things like

`gmt blockmode @ship_15.txt --increment=1 --region=-115/-105/20/30 --coltypes=g --extend --gridfile=field_%s.grd --fields=z,s,l,h`

Of course, this can all change as we experiment with it without having to maintain a separate branch that gets out of sync.  So while @joa-quim has already approved this a long time ago, I hope @seisman can improve it before I merge.

seisman:
OK, will try it in the next few days.

src/block_subs.h
```diff
@@ -87,6 +87,30 @@ struct BLOCK_CTRL {
...
 };
 
+GMT_LOCAL struct GMT_KEYWORD_DICTIONARY module_kw[] = { /* Local options for all the block* modules */
+	/* separator, short-option, long-option, short-directives, long-directives, short-modifiers, long-modifiers */
+	{ 0, 'A', "fields", "", "", "", "" },
+	{ 0, 'C', "center", "", "", "", "" },
+#if defined(BLOCKMODE)	/* Only blockmode has a -D option */
+	{ 0, 'D', "bin-width", "", "", "a,c,h,l", "average,center,high,low" },
+#endif
+#if defined(BLOCKMEDIAN)	/* Only blockmedian has a -T option */
```
seisman: The comment is wrong:
```
/* Only blockmedian has a -T option */
```

src/block_subs.h
```diff
@@ -87,6 +87,30 @@ struct BLOCK_CTRL {
...
+	{ 0, 'E', "extend", "b,r,s", "box-whisker,record,source", "l", "lower" },
```
seisman: The **-E** syntax is:
```
-Er|s[+l|h]
```
long modifier for +h is missing.

src/gmt_init.c
```diff
@@ -354,33 +354,46 @@ static struct GMT_FONTSPEC GMT_standard_fonts[GMT_N_STANDARD_FONTS] = {
...
-	{'o', "outcol"},
-	{'n', "interpol"},
-	{'p', "perspective"},
-	{'r', "registration"},
-	{'t', "transparency"},
-	{'x', "cores"},
-	{':', "order"},
-	{'\0', ""}	/* End of list marked with empty code and string */
+/* List of GMT common keyword/options pairs.  This list is used in gmtinit_kw_replace to convert
+ * the new long-format GMT options (e.g., --timestamp="My plot"+o5c/6c) to regular GMT short format
```
seisman:
```suggestion
 * the new long-format GMT options (e.g., --timestamp="My plot"+offset=5c/6c) to regular GMT short format
```


Is mixed long-format and short-format allowed? `--timestamp="My Map"+o1c/1c` gives an error:
```
gmt [ERROR]: Long-modifier form My Map+o1c/1c for option -U not recognized!
```

src/gmt_init.c
```diff
@@ -354,33 +354,46 @@ static struct GMT_FONTSPEC GMT_standard_fonts[GMT_N_STANDARD_FONTS] = {
...
+ * options (e.g., -U/5c/6c/"My plot") that the common and module parsers expect.
```
seisman:
```suggestion
 * options (e.g., -U"My plot"+o1c/1c) that the common and module parsers expect.
```

seisman:
Some failing examples:
```
gmt basemap --region=0/40/0/50 --projection=M6i --frame=afg --timestamp="Time stamp +offset" -pdf map
```

```
gmt basemap --region=0/40/0/50 --projection=M6i --frame=afg --timestamp="Time stamp"+command -pdf map
```

src/gmt_init.c
```diff
@@ -354,33 +354,46 @@ static struct GMT_FONTSPEC GMT_standard_fonts[GMT_N_STANDARD_FONTS] = {
...
+ * -DUSE_MODULE_LONG_OPTIONS will allow us to test any module_kw settings in the modules
+ *
+ * Without these we are blind to the keyword arrays.  Note that while you can test the
+ * common options without the module options, you cannot do the reverse.
+  */
+
+GMT_LOCAL struct GMT_KEYWORD_DICTIONARY gmt_common_kw[] = {
+	/* separator, short-option, long-option, short-directives, long-directives, short-modifiers, long-modifiers */
+	{   0, 'B', "frame",         "",        "",                                        "",         "" },
+	{ '/', 'I', "increment",     "",        "",                                        "e,n",      "exact,number" },
```
seisman: **-I** isn't a common option.

src/block_subs.h
PaulWessel: No, it is correct I think.  Of the three blockm*.c programs, only blockmedian has a -T option.

PaulWessel: Oh I see what you mean, this is -E... will fix

src/gmt_init.c
PaulWessel: Well, it is and it isn't.  Because it is almost a common option I keep it with the rest in the gmt_common_kw.

seisman: grdimage also has a **-I** option with a different syntax. Does it cause troubles for grdimage?

PaulWessel: I think in practice I will need to possibly move that 'I' entry in gmt_common_kw to the local_kw arrays as needed to avoid making it too complicated.  Will need to experiment with that.

seisman:
Another failing example:
```
gmt basemap --region=0/40/0/50 --projection=M6i --frame=afg --frame=+t"Title" --timestamp="Time stamp" -pdf map
```

PaulWessel:
I think --frame=+Title is wrong.  The equal means you are providing a directive, such as --frame=afg, whereas when just a modifier is given then there is no equal to assign.  So it should be --frame+tTitle I think.  However, that gives me a SEGV for another reasons (will fix).  I will also update the comments and code in the long-parser so it is easier to understand.

joa-quim:
I think ``--frame=[axes]+tTitle`` should be the way. That is, all *longoptions* should be *=something*

PaulWessel:
This will add more checks for -B since translating --frame=afg+tTitle would currently result in -Bafg+tTitle and 

`basemap [ERROR]: Syntax error -B option: Found frame setting modifiers (+b|g|n|o|p) mixed with axes settings!
`

while also points out t is missing from that message! Basically, the title is a frame setting and afg are axes settings and we don't mix them in GMT5+.  We dont want to produce GMT 4 long-B-strings so would have to teach the parser to create two separate -B option .  This increases complexity.  ANother solution is to use --frame for the frame and --axes for the axes.  Then it is cleaner.

PaulWessel:
Glad this issue came up.  I think using two keywords (frame and axes) makes perfect sense for this situation and for the GMT long-format command line.  Since we are still in the exploratory phase I think I will test that solution instead of adding lots of special checks for letting --frame split into multiple -B options.

joa-quim:
I've suffered a lot with this in the Julia wrapper too, and still have issues. Basically, I find the axes and frame definitions confusing and never remember it clearly without revisiting the **-B** docs. **-Bafg** is a frame setting (but where *afg* is generically presented as a *axes*), but selecting selecting one of them, *e.g.* **-Bx** becomes a axis setting.

PaulWessel:
No, -Bafg is an axes setting, not frame.  The two are clearly separated and described in our long -B page.

joa-quim:
Right, but look at this. The first option in a *frame* is called *axes*

image

PaulWessel:
Right, but that should actually say axis.  i.e., either x, y, or z.  The --frame option should specify which axes (--frame=west,south) while the --axes option should specify information about specific axes.

PaulWessel:
Sorry, I am wrong on that.  Yes, the frame setting specifies which axes you want, then the axis setting deals with its settings.  Perhaps we use --axis=x+ticks=afg or something.

joa-quim:
``+ticks=afg`` makes no sense. ``+ticks`` can only be *=yes* or absent. Same for *annot* & *grid*.
And maybe now you understand why I created also the *xaxis*, *yaxis*, *zaxis* (instead of *axis=x*)

PaulWessel:
Yep.  Well, I will experiment with something that works for the command line.  It does not have to be identical to your solution.  Since -B is broken into frame plus one or more axes calls I need to match that with the long-options as well.

PaulWessel:
It will continue to be challenging to let people place "+x" strings as part of a longer title or string argument, such as the `--timestamp="Time stamp +offset"` example above. The quotes are not preserved inside C and I believe our options are as before:

1. Escape such plus signs in free text so they are not a problem, e.g.,"\+x"
2. Place the entire string inside single quotes and then not split text based on modifier inside single quotes, i.e., `--timestamp="'Time stamp +offset'"`

I think using \+p for those strings you do now want parsed as modifiers is the simplest.  I have checked and the string inside C will have the \+ and hence we can simply (and temporarily) replace \+ with two non-text ascii codes, process things, and then if those two codes are present, adjust the text by removing the first and replacing the second with +.

PaulWessel:
Improved further. Added the --axes keyword to deal with axes and let --frame just deal with frame settings.  ANy label/text that contains +? sequences that should not be seen as modifiers must be escaped with a backslash and the entire string must be quoted.  WIth that, these work for me:

```
gmt basemap --region=0/40/0/50 --projection=M6i --frame=WSrt+title="Title"+fill=pink --axes=afg --timestamp="Time stamp" -pdf map
gmt basemap --region=0/40/0/50 --projection=M6i --frame=afg --timestamp="Time stamp \+offset" -pdf map
gmt basemap --region=0/40/0/50 --projection=M6i --frame=WSne --axes=x:afg --axes=y:af --timestamp="Time stamp"+command -pdf map
```

The --axes=x:agf follows the general longname=[directive[:arg]] general template since there are many GMT module options that requires this scheme, regardless if we decide to do --xaxis instead.

PaulWessel:
OK, will do that before committing.  Wait, this requires a compilation switch so I dont think we easily can unless you want to add 

```
add_definitions(-DUSE_COMMON_LONG_OPTIONS) 	# Turn on testing of upcoming long-option syntax for common GMT options
add_definitions(-DUSE_MODULE_LONG_OPTIONS) 	# Turn on testing of upcoming long-option syntax for module options

```
to the CI setup for master.  Otherwise the long-parsing is deactivated.

PaulWessel:
Let me know if you can add those statements to the CI cmake config file and then we could add a test or two to the master branch, @seisman.

seisman:
@PaulWessel Now you can add long-option tests to this branch.

seisman:
@PaulWessel The long-option test fails due to the timestamp of the **-U** option.

PaulWessel:
I guess we cannot really have the time-step in these tests as it changes.  I will remove it.

PR #234, opened 2018-12-23
Correctly format option -s

PaulWessel:
The GMT common option **-s** deviates from the standard use of directives and modifiers by ending with optional **a** or **r**.  While remaining backwards compatible, the correct syntax is to use these as modifiers **+a** or **+r**.  Thus, the blessed syntax is therefore **-s**[*cols*][**+a**|**+r**].  The new syntax is presented in the documentation, explanations, as well as various usages in scripts.


joa-quim:
But this is the case also of the -R....r variant

PR #237, opened 2018-12-23
Rename/modify gmtlib_count_slashes

PaulWessel:
Renamed to gmtlib_count_char and now takes a new argument (the character) we need to count.  This is motivated by need to also count commas in some places, so a general function that takes the character of interest is better.


seisman:
@PaulWessel `src/t.lis` shouldn't be included in this commit.

PaulWessel:
It should not.  Still struggling with git vs subversion work habits...will fix.

PR #241, opened 2018-12-27
pssac: don't draw line if -G is used and -W isn't

seisman:
Sometimes users may want to fill a seismogram but don't want to plot the line. Old pssac always draw the line no matter -W is used or not. The new pssac will not draw the line if -G is used but -W isn't.

The new pssac is slightly incompatible with the old one. Not sure if we should incorporate it into 5.4 branch.


seisman:
Merged into master and 5.4 branches.

PR #261, opened 2018-12-31
Modernize -G to use modifiers in pswiggle

PaulWessel:
The old version had a strange syntax with -,+,= as directives which would interfere with searching for modifiers.


PaulWessel:
Parser is backwards compatible.

PR #277, opened 2019-01-02
Add coverage reports

seisman:
This PR adds coverage reports for gmt C codes.

We need to run the whole tests to generate the coverage reports. Because these tests take too long (~30 min on travis), we cannot build the coverage reports for every commit. Currently, the tests and coverage are only activated by cron jobs, so we don't need to wait the reports for every commit, but still get latest coverage reports every day.

Although there're some failing tests, we still get coverage reports. You can check the coverage reports generated by this PR here (https://codecov.io/gh/GenericMappingTools/gmt/tree/561266b5cdce0699f8fff53068691e7e38942e26). Currently, the coverage is 57%.

@GenericMappingTools/core Any comments or suggestions?

**Please don't merge this PR even after approval, I still need to do some cleanup before merging.**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,8 @@ script:
         cd build; make -j docs_html; cd ..;
         cd build; make -j check; cd ..;
       fi
+    - export COVERAGE=true
+    - cd build; make -j check; cd ..;
 
 after_success:
     - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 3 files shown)


PaulWessel:
That is great, @seisman!  I will let @leouieda be the one that approves this PR.

joa-quim:
``Coverage``, Uau. Lot's of new work in perspective to increase that percentage.

PR #282, opened 2019-01-04
Add FindGMT.cmake file (from GMTSAR)

joa-quim:


PaulWessel:
Why would GMT need to find GMT? It is installing GMT.

joa-quim:
Others may need it. For example vcpkg wants one to when it runs the vcpkg_test_cmake() function, though I could not make the damn thing find it and hence I can build GMT with vcpkg but not install it.

PR #283, opened 2019-01-04
Fix "Shebang" problems in bash script headers

holishing:

Fixes some issues in https://gmt.soest.hawaii.edu/boards/2/topics/7449

continue the fixes in 7febdf7, for let these executable script or sample
scripts more compatible with other Unix-Like system (such as FreeBSD,
other BSD versions...)

**Something need help to be checked**
if these changes will affects related behaviors of Travis CI or other test services.


PaulWessel:
Hi @leouieda, @seisman can you seen any reasons why we should not accept this PR?

seisman:
Seems good to me.

PR #285, opened 2019-01-04
Deploy built documentations to gh-pages from Travis

seisman:
Uses the Fatiando a Terra CI scripts (https://github.com/fatiando/continuous-integration)
to deploy the built docs to the `gh-pages` branch.
This branch is served at https://genericmappingtools.github.io/gmt/dev/
The built HTML from the master branch is placed on the `/dev` folder of gh-pages.
Whenever a new tag is created, the scripts will create a new folder with that tag
name (eg `/6.0.0`) and link `/latest` to the new folder. 
This way https://genericmappingtools.github.io/gmt/latest always points to last
release made and https://genericmappingtools.github.io/gmt/dev/ always reflects
the master branch.


leouieda:
@seisman this should be pretty straight forward now. One of the only things missing is an encrypted Github token. I can finish this if you'd like, as I've done this sort of thing many times before.

seisman:
@leouieda Go ahead please.

leouieda:
Merged the master branch and testing the deployment from this branch. If all is work, will limit docs deployment to the daily cron job.

.travis.yml
```diff
@@ -55,7 +57,10 @@ env:
         - PATH="$INSTALLDIR/bin:$PATH"
         - LD_LIBRARY_PATH="$INSTALLDIR/lib:$LD_LIBRARY_PATH"
         - BUILD_DOCS=false
+        - DEPLOY_DOCS=false
         - TEST=false
+        # Encrypted Github token (GH_TOKEN)
+        - secure: "cwHlTNa+b2lPiCbN5Ja9alj6EkKwJsuR5QcNPORrp3uCOQF9+cJUFEyMIGVGzrzZQnpjn05XYDr6eGFUYkbaR1QVUQdNH+2rCiScFHe94fRPJlJabxXexCfvZCNV8VRi6NSyFPuZKkCVs10bQ7Lk5dPS/V4Dalz1xyURcvI8H8XQFTmDYb/7QFw73l0WzNf+Y1rzNpRZEQ4ERhmnir489njCv1w7oJ0C+BrRvPqTmNmfXbrY6Bcana6BJvvumSY1xRaw60f/UNKYTlTVlYTVSWK1kq9mTfkhiJhe5oNoWchKloiOg3fA2XtAuucuXSdCjqo9aWHMxKJy+jQ7/BeId9hacvB9pVclYoOLARBcPoWDb21eKJZ8A70yKlUM3ryhVXM4drHzM4ITitPA1sEJWNJ2BQO8Cils0FfcOLvBR3NdbLi8RdHZItmU/Oe8o1srUTnnkdPRiEWbmAEs4pVrsHY9sD0660PXTFBSQ7sDXeyTl2etG/YSQKvE3ysOnZwhEuTXqf8cpo+tSbjindvWSyyN0HSuJ0bd7Ei+qWUQ3nJNEtHpiq8Ms6KUPHjPpWO2BEl+FQHWFpmRtwBg/REk36VWIl1UHSKuKkqxUT4qiZeL5zlrHGso+7QGhVjkYj5un3XoVObopFNVUB8y56xeW21gI51HK48cPd5rMnX95KE="
```
seisman: @leouieda Can you access this link (https://codecov.io/gh/GenericMappingTools/gmt/settings) for codecov token? If you can,  I think it would be better to add the codecov token here.

leouieda: I can access it but I can't copy it. Is there any reason why you prefer having it here?

seisman: I prefer adding CODECOV_TOKEN here, so that we can manage these two tokens at the same place. Currently, GH_TOKEN is in .travis.yml, while CODECOV_TOKEN in travis's web interface.

leouieda:
OK, seems to be working. The docs are built and served under https://genericmappingtools.github.io/gmt/dev/

For now, https://genericmappingtools.github.io/gmt/latest also points to the same docs but it will be replaced with the docs for the latest release. This happens automatically whenever a new tag is created. `/dev` will always point to our daily builds of master.

cc @PaulWessel @joa-quim

.travis.yml
leouieda: OK. We could also move the GH_TOKEN to the web interface then. As I said, I can't add the codecov token because I don't have access to it. But if you have it, then could you add it here instead?

seisman: @leouieda I've added encrypted CODECOV_TOKEN to travis.yml and removed it from the web interface.

leouieda: 👍 thanks, Dongdong!

leouieda:
This is pretty much done on my part. After approval, we should be good to merge. The PR description is updated and can be used for the commit message.

PR #289, opened 2019-01-06
Avoid names reserved for the compiler

PaulWessel:
We use defines like **_GMT_H** but anything that starts with underscore is reserved for compilers.

### DIFF
--- a/src/gmt_common.h
+++ b/src/gmt_common.h
@@ -28,8 +28,8 @@
  * \brief Holds current selections for the family of common GMT options
  */
 
-#ifndef _GMT_COMMON_H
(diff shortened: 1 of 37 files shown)


elfring:
Thanks for [your adjustment of some affected include guards](https://github.com/GenericMappingTools/gmt/pull/289/commits/084dcf14f3c09891ec24ff1660fda75d13300237 "Avoid names reserved for the compiler").
[A few update candidates were left over](https://github.com/GenericMappingTools/gmt/blob/3344c5786a94e0edb1902e3c44f6f29967a038a0/src/common_string.h#L31 "Update candidate").

PR #301, opened 2019-01-10
Fix pen style

seisman:

This PR fixes a issue introduced by #243. When pen color is omitted, pen style names `dot` and `dash` are not recognized. 

Code to reproduce the bug:
```
gmt psxy -R0/10/0/10 -JX10c/10c -Baf -W1p,dash > test.ps << EOF
2 2
6 6
EOF
psxy [ERROR]: Colorname dash not recognized!
```

### DIFF
--- a/doc_classic/rst/source/explain_pen.rst_
+++ b/doc_classic/rst/source/explain_pen.rst_
@@ -12,7 +12,7 @@ Specifying Pens
     **thick**\ [**er**\ \|\ **est**], **fat**\ [**ter**\ \|\ **test**],
     or **obese**. *color* specifies a gray shade or color (see
     `Specifying Color`_ above). *style* can be any of 'solid', 'dashed'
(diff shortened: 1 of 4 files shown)


PaulWessel:
The change in gmt_support.c has some issues:

- the comparison for dashdot will never be reached since the check for dash with 4 chars will find it first
- there is a stray length of 57 in the last strncmp

I think we should check for dot (leading 3 chars) and dash (leading 4 char) so that dotted and dashed are still OK.

seisman:
> the comparison for dashdot will never be reached since the check for dash with 4 chars will find it first


Yes, you're right. Does changing the order of the comparisons work?
```
if (!strncmp (word, "dotdash", 7U)  || !strncmp (word, "dashdot", 7U) || \
    !strncmp (word, "dash", 4U) || !strncmp (word, "dot", 3U) || \
    !strncmp (word, "solid", 5U))
```

Issue #298, opened 2019-01-08
Error message when using remote file


seisman:
```
$ gmt grdinfo @earth_relief_02m.grd
grdinfo: Libcurl Error: Remote file not found
grdinfo: Failed to remove /Users/seisman/.gmt/cache/earth_relief_02m.grd! [remove error: No such file or directory]
grdinfo: Could not even remove file /Users/seisman/.gmt/cache/earth_relief_02m.grd
/Users/seisman/.gmt/earth_relief_02m.grd: Title: Earth Relief at 2 arc minute
/Users/seisman/.gmt/earth_relief_02m.grd: Command: grdconvert ETOPO2v2g_f4.nc earth_relief_02m.grd=ns -fg
	(old cmd) grdreformat ETOPO2v2g_GMT_int.grd ETOPO2v2g_GMT_flt2.grd
/Users/seisman/.gmt/earth_relief_02m.grd: Remark: Converted to chunked int NC4 grid from ETOPO2v2g_f4.nc
/Users/seisman/.gmt/earth_relief_02m.grd: Gridline node registration used [Geographic grid]
/Users/seisman/.gmt/earth_relief_02m.grd: Grid file format: ns = GMT netCDF format (16-bit integer), COARDS, CF-1.5
/Users/seisman/.gmt/earth_relief_02m.grd: x_min: -180 x_max: 180 x_inc: 0.0333333333333 name: longitude [degrees_east] n_columns: 10801
/Users/seisman/.gmt/earth_relief_02m.grd: y_min: -90 y_max: 90 y_inc: 0.0333333333333 name: latitude [degrees_north] n_rows: 5401
/Users/seisman/.gmt/earth_relief_02m.grd: z_min: -10722 z_max: 8046 name: z
/Users/seisman/.gmt/earth_relief_02m.grd: scale_factor: 1 add_offset: 0
/Users/seisman/.gmt/earth_relief_02m.grd: format: netCDF-4 chunk_size: 129,129 shuffle: on deflation_level: 3
```
* Operating system: macOS
* Version of GMT: 5.4.5

PaulWessel:
Cannot reproduce with 6.0.  Will try tomorrow with 5.4.5 at work.

PR #303, opened 2019-01-10
Fix access to remote files in 5.4.x

PaulWessel:
I think we updated syntax on remote files in 6 and decided to always use the leading @ symbol.  In 5.4 just giving earth_relief_xxm.grd was OK but now it tries to read from the cachhe dir rather than the top dir.  This change fixes the #298.


btozer:
For me this seems to be a problem using both 5.4.5 and 6.0

on gmt 5.4.5 @earth_relief_XX fails e.g.

grdimage: Failed to remove /Users/btozer/.gmt/cache/earth_relief_05m! [remove error: No such file or directory]

and -Vd reports:

grdimage: Downloading file ftp://ftp.soest.hawaii.edu/gmt/data/cache/earth_relief_05m
grdimage: Libcurl Error: Remote file not found

on inspection of the ftp site it seems earth_relief_XX are stored in 

ftp://ftp.soest.hawaii.edu/gmt/data/

not

ftp://ftp.soest.hawaii.edu/gmt/data/cache/

so it seems the problem is simply a dir path issue. i.e. remove cache/ from the file path to fix?

PR #146, opened 2018-10-24
WIP: Setup travis build on OSX

seisman:

PR #317, opened 2019-01-16
Add OSX builds and refactor TravisCI build config

leouieda:
Refactor the TravisCI build configuration to make it easier to add OSX support
and to deploy the docs to Github pages later on. Set different builds in the 
matrix for cron jobs and separate testing and building the docs. This is 
necessary since the tests are known to always fail and we want to deploy the 
docs anyway. The build was always passing even if the docs or tests ended in 
failure because the `cd build; ...; cd..;` command always passes. The return 
value is only taken from the last `cd`.


leouieda:
I think I managed to get a simpler setup than #146 and the build time for OSX is 7.5 min. Even then, we always have the option of ignoring the CI (though we shouldn't).

Running a last test to see if the build works on the cron jobs as well.

.travis.yml
```diff
@@ -1,50 +1,120 @@
...
+          if: type != cron
+        - name: "Linux"
+          os: linux
+          #if: type = cron
+          env:
+            - TEST="all"
+        - name: "Mac"
+          os: osx
+          if: type != cron
+        - name: "Max"
```
seisman: Max -> Mac

ci/build-gmt.sh
```diff
@@ -0,0 +1,52 @@
...
+# To return a failure if any commands inside fail
+set -e
+
+cat > cmake/ConfigUser.cmake << 'EOF'
+set (CMAKE_INSTALL_PREFIX "$ENV{INSTALLDIR}")
+set (GMT_LIBDIR "$ENV{INSTALLDIR}/lib")
+set (DCW_ROOT "$ENV{COASTLINEDIR}")
+set (GSHHG_ROOT "$ENV{COASTLINEDIR}")
+
+set (CMAKE_BUILD_TYPE Debug)
```
seisman: Duplicate CMAKE_BUILD_TYPE if "COVERAGE" == "true".

.travis.yml
```diff
@@ -1,50 +1,120 @@
...
+    - if [ "$TEST" == "all" ]; then
+        make -j check;
+      fi
+    - if [ "$BUILD_DOCS" == "true" ]; then
+        make -j docs_html;
+      fi
+    - cd ..
+
+# Things to do if the build is successful
+after_success:
```
seisman: Although some tests fail, we can still upload coverage reports.

.travis.yml
leouieda: :+1:

.travis.yml
leouieda: Right, I forgot about that. At least now the tests really show up as failures instead of passing. Hopefully that red badge will be good motivation to fix them or mark as won't fix :wink:

ci/build-gmt.sh
leouieda: :+1: removed the one inside the if block

leouieda:
@seisman should be good to go. Could you do one last review?

.travis.yml
```diff
@@ -1,52 +1,124 @@
...
         - INSTALLDIR="$HOME/gmt-install-dir"
         - COASTLINEDIR="$HOME/gmt-install-dir/coast"
         - PATH="$INSTALLDIR/bin:$PATH"
         - LD_LIBRARY_PATH="$INSTALLDIR/lib:$LD_LIBRARY_PATH"
-        - if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
-            export PYTHON=3.6
-            export CONDA_INSTALL_EXTRA="sphinx"
-            export COVERAGE=true
-          fi
+        - COVERAGE=true
```
seisman: The environment variable COVERAGE seems useless, since we always need to run the tests to generate coverage reports.

ci/build-gmt.sh
```diff
@@ -0,0 +1,51 @@
...
+set (DO_TESTS TRUE)
+set (DO_ANIMATIONS TRUE)
+set (N_TEST_JOBS 2)
+EOF
+
+if [ "$COVERAGE" == "true" ]; then
+    cat >> cmake/ConfigUser.cmake << 'EOF'
+set (CMAKE_C_FLAGS "-Wall -Wdeclaration-after-statement -coverage -O0")
+EOF
+fi
```
seisman: We don't need to enable the tests in compile-only jobs, so we can change it to:
```
if [ "$TEST" == "all"]; then
    cat >> cmake/ConfigUser.cmake << 'EOF'
set (CMAKE_BUILD_TYPE Debug)
enable_testing()
set (DO_EXAMPLES TRUE)
set (DO_TESTS TRUE)
set (DO_ANIMATIONS TRUE)
set (N_TEST_JOBS 2)
set (CMAKE_C_FLAGS "-Wall -Wdeclaration-after-statement -coverage -O0")
EOF
fi   
```

seisman:
@leouieda The PR looks good. I just made two comments.

seisman:
@leouieda codecov needs a token to upload the coverage reports. The token was added via the travis settings panel. You can encrypt this token and add it to .travis.yml.

.travis.yml
leouieda: Yeah, I agree. I'll remove it and use the `TEST` variable instead.

ci/build-gmt.sh
leouieda: Are any of those required for building the docs?

leouieda:
> @leouieda codecov needs a token to upload the coverage reports. The token was added via the travis settings panel. You can encrypt this token and add it to .travis.yml.

No need. Plus, I can't access the unencrypted token from the web interface.

ci/build-gmt.sh
leouieda: Never mind, already found it.

PR #320, opened 2019-01-17
Let psmeca support perspective view

seisman:


GeohunterX:
Hi! I just stumbled over this request since I would need and really apprechiate this option, too. Any changes so far? Guess it's not implemented yet since I am getting "psmeca Error: Unrecognized option -p"?

It would also be really cool to have perspective view for pscoupe as well for nice 3D block diagrams!

seisman:
@GeohunterX It's already implemented and merged into the master branch. But it's not officially released yet.

It seems you're using GMT 5.4.5 (the latest release) or older. You need install the latest GMT developing codes from the master branch, then `psmeca -p` would be available.

GeohunterX:
Yes you're completely right I'm on 5.4.5

Thanks for your fast reply and for the solution! I will check out how to get the master branch (I'm still new to linux).

Thank you!!!

seisman:
@GeohunterX #538 has been merged into master branch and pscoupe also supports -p option.

GeohunterX:
@seisman @PaulWessel - you're amazing! Thank you so much for this incredibly fast implementation!!!

It's working perfectly :-)

Issue #316, opened 2019-01-16
Enhance custom symbol conditional statements to also work on characters/strings 


KristofKoch:
Conditional tests in a custom symbol are a great way to change to symbol depending on data passed to the symbol. This works already great for numerical values. 

I'd like to request to **make conditional testing available for characters/strings as well**. The _is equal_ and _is not equal_ operators would of interest. This will be useful when the symbol contains text or single characters. 

# Current situation:

A single character is passed to the symbol and the symbol has to display the character in a special way. Currently I have to pass the character plus a "control variable" where a conditional statement can act on to the symbol. 

## Placement file:
````
175.9   -38.8   1  A
173.75  -39.25  2  B
````

## Symbol code snippet:
````
if $1 == 1 then {
    0   0.1  0.8 $t l -W- -Gblack
} elseif $1 == 2 then {
    0  -0.1  0.8 $t l -W- -Gblack
}
````

# Improved Situation:

No need for the "control variable" when it is possible to test against characters/strings:

## Placement file:
````
175.9   -38.8   A
173.75  -39.25  B
````

## Symbol code snippet:
````
if $t == A then {
    0   0.1  0.8 $t l -W- -Gblack
} elseif $t == B then {
    0  -0.1  0.8 $t l -W- -Gblack
} ...
````

The placement file doesn't need redundant information and the symbol can act on the variable directly.

This feature request is inspired by [takaya nambas question in the forum about circled characters](http://gmt.soest.hawaii.edu/boards/1/topics/7982). The justification CM doesn't help much if you want to center a character in a circle. The optical middle of a character isn't always the geometric center. The letter "A" needs to sit a bit higher while the letter "W" needs to sit lower to appear visually pleasing centered. Another use case are warning levels like "dormant", "elevated", "active", "erupting" for volcanos …

**Comment by KristofKoch** [comment]
As stale bot got this issue in its claws – any news on this one?

**Comment by PaulWessel** [comment]
Hi @KristofKoch, I just tried it out and perhaps it already works.  I did this:
```
cat << EOF > t.txt
0	0	A
1	1	B
EOF

```
Then I made this custom symbol (text.def):
```
#
N: 1 s
if $t == A then {
    0   0.1  0.8 $t l -W- -Gred
} elseif $t == B then {
    0  -0.1  0.8 $t l -W- -Gblue
}

```

and ran

[...]

PR #323, opened 2019-01-17
Lay groundwork for doing text comparisons in custom symbols

PaulWessel:
This follows up on #316.  Does not break anything but also not tested for the new feature yet.

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -2554,11 +2554,59 @@ enum plot_operand {
 	RIGHT_OPERAND2 = 2
 };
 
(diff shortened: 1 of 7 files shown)


KristofKoch:
Please pardon my lack of understanding - is it correct that text comparisons should now work in 6.0.0_31f9585 ?

PaulWessel:
No not yet, I have not even tried to see if what I implemented is correct yet.  I just put in the pieces.  On travel the next 2 says so be patient.

KristofKoch:
Thank you for clarification - safe travels!

PR #331, opened 2019-01-21
Add a Code of Conduct from Contributor Covenant

leouieda:
Adds a template code of conduct for the project saying the basic things:
be nice, don't harass, be welcoming. Also includes a way of reporting 
transgressions (right now, emailing me or @PaulWessel ). Is that OK? 

This is standard in many open-source projects and is a good thing to have
just in case.

**Reminders**


leouieda:
@GenericMappingTools/core please take a look at the CoC. It's pretty straight forward but we should be aware of it and call people out if they misbehave.

PR #353, opened 2019-01-24
Speedup Travis by caching, build animations, and fix osx path issues

ldldr:
These changes should speed up the Travis build.

- Let travis cache apt packages, pip cache dir as well as datasets tarballs.
- Use Ninja as CMake generator.
- Use ccache to avoid building objects that haven't changed
- Retry downloads of datasets if they fail
- Use @seisman's dataset mirrors instead of the Hawaii ones
- Build animations for the HTML docs
- Fixes the OSX library path issues

Animations are known to fail because they take too long and Travis times out.


ldldr:
This cuts down the average build time from 5 to 2 minutes on Linux. It also fixes the issue with the runtime paths for dynamic library files on OSX

ldldr:
It basically works but the doc build times out after 50 minutes because of the animations taking to long (https://travis-ci.org/GenericMappingTools/gmt/jobs/484607685 and https://travis-ci.org/GenericMappingTools/gmt/jobs/484585377). Anything particular that we can do to cut down the time? Optimize gmtmovie, do less frames or leave out webm format?

PR #362, opened 2019-01-24
Clarified -jx option in Building GMT (quick start)

KristofKoch:
Clarified the usage of *-jx* option when running *make -jx*. Previously the example command and the explanation didn't match. 

At least I struggled quite a bit with the *-j* option and the description when installing GMT6 the first time.

While I understand that on most machines *make -j* isn't a problem, it might be on shared machines. Therefore I expanded the explantation a bit to be more helpful for novice users.


KristofKoch:
Hi Paul,

I incorporated your suggestion.

PR #365, opened 2019-01-24
bug fixes in earthtide time interval

mrajner:
This 3 commits fixes problem of incorrect time delta in sun/moon position computation. Moreover the output was confusing as the gregorian date was as expected, while computed values were for different epochs.

The other fix is with day interval unit (`-T.../.../...d`), which was not specified in docs for earthtide, however it was parsed correctly giving proper gregoriand dates but with wrong earth/sun position, hence tidal components.

The simple sh test file is attached in separate commit. You can run this script with current version to see the problems described.

Please see extended commit message for additional information and description.


joa-quim:
Thanks for catching this and providing the extra test.

PR #373, opened 2019-01-26
Generate smaller animation examples

PaulWessel:
I have changed several parameters so that the number of frames have been reduced for the long animation 06 and using a cruder data grid for 07.  What took ~15 minutes to run for me now takes ~3 minutes.  Hopefully this is simple enough for travis to complete


leouieda:
I merged in the latest changes from #353 and triggered a build of the docs in this PR. If the build doesn't timeout we can revert the docs build back to the cron job.

PaulWessel:
How did it go @leouieda, can we merge this one?

leouieda:
Not yet. The doc build failed but I'm not sure why. It seems that it tried to build the latex PDF as well. That's not really necessary, is it?

PaulWessel:
I don't think so, if there are stuff that truly fails it will fail in the rst to html step as well.  The PDF should be triggered separately and perhaps only when we do a release.

seisman:
https://github.com/GenericMappingTools/gmt/blob/ebbe76c37aa8acd67704a29f111d57905b920a18/.travis.yml#L153-L160

Do we really need to compile the PDF here?

PaulWessel:
No, I think the PDF build should only be triggered during a release, not the regular builds.  Anything that crashes the PDF build would also crash the HTML build I think so no point in doing both.

seisman:
@PaulWessel @leouieda The builds now pass. I think it's OK to merge now.

Issue #388, opened 2019-02-01
Inconsistent rendering of hyphens/minus


seisman:
```
echo 5 2 -A -B -C -123 | gmt text -Jx1c -R0/10/0/3 -Baf -B+t'-A -B -C -123' -png hyphens
```

![image](https://user-images.githubusercontent.com/3974108/52096546-40db9d00-2595-11e9-9380-6cb79d1cedb4.png)

The hyphen in `-A` is different from others.

**System information**

* Operating system: macOS
* Version of GMT: gmt latest

PaulWessel:
Hi @seisman, see if you agree that we can close this one.

PR #391, opened 2019-02-01
Try to be more consistent with minus and hyphens

PaulWessel:
Because the different character sets have different octal codes for minus and hyphen we may sometimes have to replace the typed hyphen character with a minus and vice versa.  Now only adding in a minus if what follows is a number and it is the only number.  Partly addresses #388.

### DIFF
--- a/src/postscriptlight.c
+++ b/src/postscriptlight.c
@@ -632,11 +632,13 @@ static void psl_fix_utf8 (struct PSL_CTRL *PSL, char *in_string) {
 	char *out_string = NULL;
 
 	if (!strncmp (PSL->init.encoding, "Standard+", 9U)) {	/* For Standard+ encoding we need to swap leading minus values encoded as hyphen with the actual minus symbol */
-		if (in_string[0] == 0055)	/* Found leading hyphen which we interpret to be a minus sign */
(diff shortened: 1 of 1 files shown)


seisman:
I don't think we can find a perfect solution. For example, if someone want to write `Y = aX - b` in a plot. It's impossible to know that we should replace `-` with minus symbol in this case.

IMO, we only need to do the hyphen-minus replacements for annotations, since it's out of control by users. For labels, titles and other texts, users have full control, they can determine what they want to use.

PaulWessel:
OK, then I think we are OK.  The fix I added caught the leading -A and no longer replaces that with a minus symbol.  I think that is the correct guess and if the user means "minus variable A" then they can use octal codes as needed.

PaulWessel:
Yes, that is possible but requires a few changes since I implemented the current scheme inside postscript light so it does not know what it is being passed.  Doing it at a higher level (in gmt_plot.c) will be required.

seisman:
IMO, postscript library shouldn't make the changes.

PaulWessel:
OK, we will keep this open and await my return to Hawaii early next week.  We will move the hyphen/minus code portion into the annotation machinery then.

PR #393, opened 2019-02-01
Update some examples to modern mode

seisman:


joa-quim:
Actually we need the classic versions too. The idea is to have the examples in separate files (one classic and one modern) and include them in the rst versions.

seisman:
@joa-quim The changed files are only for modern mode.

joa-quim:
Nope. We only have one ``doc`` now and it will have to do both the classic and modern roles.

PR #395, opened 2019-02-01
Default is gridline, not pixel

joa-quim:


PaulWessel:
Well, not sure.  If you read the whole paragraph you will see that with no -r then gridline is the default.  However, if you give -r with no argument then it defaults to pixel. Adding g or p overrides this.  Perhaps this can be said better as I am using "default" in two different ways.  Perhaps I can work on this on the planes....

PR #403, opened 2019-02-06
Add missing KEYs to x2sys_cross

joa-quim:
Missed KEYs for input but things still don't work and external calls ends up crashing. The problem is that ``x2sys_read_file()`` calls ``x2sys_get_data_path (GMT, path, fname, s->suffix)`` but ofc there are no suffixes for in-memory data files. But other than this, ``x2sys_read_file()`` doesn't seem fit to read in-mem files.

### DIFF
--- a/src/gmt_supplements_module.c
+++ b/src/gmt_supplements_module.c
@@ -84,7 +84,7 @@ static struct Gmt_moduleinfo g_supplements_module[] = {
 	{"rotconverter", "spotter", "Manipulate total reconstruction and stage rotations", ">D}"},
 	{"rotsmoother", "spotter", "Get mean rotations and covariance matrices from set of finite rotations", "<D{,>D}"},
 	{"x2sys_binlist", "x2sys", "Create bin index listing from track data files", ">D}"},
-	{"x2sys_cross", "x2sys", "Calculate crossovers between track data files", ">D}"},
+	{"x2sys_cross", "x2sys", "Calculate crossovers between track data files", "<D{,>D}"},
 	{"x2sys_datalist", "x2sys", "Extract content of track data files", ">D}"},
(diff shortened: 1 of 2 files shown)


joa-quim:
And what does it need to make it really work?

PaulWessel:
Not sure without looking.  This tool was meant to read thousands of files from disk so having them in some sort of array of datasets in memory is quite different and maybe not practical.  It is not clear if whatever work that would take to implement is warranted.  If you just have a few things and care about the crossings only then gmt spatial can do that now.  if you need to data crossings etc like x2sys was originally used for then sticking with files is not so bad.

joa-quim:
Hadn't thought on gmtspatial (that I never used for this). My use case is to compute just the internal crossover errors so a simpler case for x2sys_cross. However trying in command line I can't make gmtspatial work

This works fine
```
x2sys_cross -TLINE l9-25.dat -Qi > cros_9-25.dat
```
but this gives incomprehensible results 
```
gmtspatial l9-25.dat -Ii > cros_9-25.dat
```

l9-25.dat
```
-43.887306	24.280865	17.824999
-43.841675	24.291063	15.884999
-43.796387	24.301170	15.775000
-43.751442	24.311186	16.859999
-43.706829	24.321112	17.590000
...
```

cros_9-25.dat
```
-43.857342	24.2326075	2708.5	2708.5	l9-25.dat	l9-25.dat
-32.4217205	25.901882	2707.5	2707.5	l9-25.dat	l9-25.dat
-39.265007224	24.9017640474	2707.79861523	4061.79841888	l9-25.dat	l9-25.dat
-34.6451498353	25.5769370307	2707.59702207	5415.59669112	l9-25.dat	l9-25.dat
...
```

Issue #389, opened 2019-02-01
subplot not working with -Jp polar projection


seisman:
**Description of the problem**
In subplot mode, the width or scale of each subplot should be omitted, since subplot will automatically calculate the subplot width and append the width to -J option. However, this doesn't always work for -Jp, since figure width is not always the last arguments.

**Full script that generated the error**

```
gmt begin Jp pdf
  gmt subplot begin 2x2 -Fs5c 
    gmt basemap -JPa5c/45 -R0/90/0/1 -Bxa45f -c1,1 
    gmt basemap -JPa5c/45 -R0/90/0/1 -Bxa45f -c1,2
    gmt basemap -JPa5c/45 -R0/90/0/1 -Bxa45f -c2,1
    gmt basemap -JPa5c/45 -R0/90/0/1 -Bxa45f -c2,2
  gmt subplot end
gmt end
```

**Full error message**

```
psbasemap [ERROR]: Syntax error -J option. Correct syntax:
	-Jp[a]<scale>[/<origin>][r] OR -JP[a]<width>[/<origin>][r]
	  <scale> is cm/units, or use <width> in cm
	  Optionally, prepend a for azimuths, append theta as origin [0],
	  or append r to reverse radial coordinates.
psbasemap [ERROR]: Offending option -JPa5c/45/1.9685i
```

**System information**

* Operating system: macOS
* Version of GMT: 6.0.0_32e95ef

PaulWessel:
Hm, I wonder if a more robust implementation would be to specify a special symbol for the width or scale so we know where to replace it with the computed value.  Perhaps -JPa=/45 and the = is replaced by the computed width (-Jpa1.9685i/45)?

seisman:
The syntax of -Jp projection `-JP[a]<width>/<origin>[r][z]` is a little weird to me. Most other projections put `<width>` as the last argument, while `-JP` as the first one. Prepending `a` and appending `r` or `z` make it more complicated.

PaulWessel:
True, that is a weird syntax.  What idiot came up with that?  Oh, me...
So if we switched it in a backwards compatible way to
**-JP**[**a**]_width_[**+o**_origin_][**+r**][**+z**]
or something then it would be the last argument (once modifiers are recognized).  We cannot just switch width and origin in the old syntax since a/b looks the same as b/a.

seisman:
[...]

PR #421, opened 2019-02-11
Use ? to indicate scale or width in projections in subplots under modern mode

PaulWessel:
We cannot specify the actual width or scale in subplots, and because some projections have many arguments and the width may not be the last one, we will use ? to indicate where the computed scale or width should be placed.  Addresses #389.


seisman:
@PaulWessel Don't forget to update the documentation of subplot.

Issue #402, opened 2019-02-06
Nicer clipping of lines/polygons at map region boundaries


KristofKoch:
(Issue is based on this forum thread: http://gmt.soest.hawaii.edu/boards/1/topics/8037)

**Description of the problem**

When a line or polygon intersects the map region boundary its clipping point is exactly at the boundary. While logically correct, it is not visually pleasing for _line pen width_ > _map boundary pen width_ : 

![image](https://user-images.githubusercontent.com/45796239/52325656-1e5ae100-29e6-11e9-913d-e44b83971fd6.png)
The issues are the "bleeding" over the map region boundary and the "missing" wedge inside the plot. Basically the current implementation is only correct for zero _line pen width_ or for an intersect angle between the line and the boundary of exactly 90°. 

The desired result is more like this mockup:

![image](https://user-images.githubusercontent.com/45796239/52325882-fd46c000-29e6-11e9-8edf-7ccbd08d91ed.png)
**Possible solution**

The solution @PaulWessel suggested in the forum thread:
> Whenever a line intersects the map boundary and we clip the line to the starting point, we must instead extend the start of the line geometrically backwards/outwards so its end point is 0.5 pen thickness from the boundary. […] GMT will set up a clip path so that we will achieve the desired result.

![ugly_explain-o-graph](https://user-images.githubusercontent.com/45796239/52326020-84943380-29e7-11e9-8823-44fb2d1d531c.png)

This is basically a visualization of Pauls idea.

**Script that generated the error**

This is the simple script that generates the _red line example_ above. 

```
#!/bin/bash

FILE=test_bleeding_pen.ps

gmt psxy \
    -R7.5/49.5/9.5/50.5r \
    -JL8.5/50/49.75/50.25/15c \
    -Bwesn \
    -W35p,red \
    > $FILE << END
 7.6 49.4
10   49.9
END
```

**System information**

* Operating system: Ubuntu 18.04.1 LTS (GNU/Linux 4.15.0 x86_64)
* Version of GMT: 6.0.0_f5755c1 [64-bit] [MP] [4 cores]

PaulWessel:
[...]

PR #426, opened 2019-02-14
Consider line thickness when clipping lines against rectangular boundaries

PaulWessel:
This PR is in support of the problem discussed in #402 and follows the solution I outlined earlier today.  Currently, the algorithm is limited to maps with rectangular boundaries as well as lines that do not wrap across 360 degrees.  Once the current PR is tested and found acceptable then I will extend that to the other two cases. This PR should not be accepted without trying and running the examples.  It currently solves @kristofkoch example and passes the trimline.sh test.  Please switch to this branch and run tests and comment first.  It currently breaks a dozen or so other scripts, but in a good way: Part of my solution requires map clipping to be on (in psxy, we only turned on clipping for symbols, not lines, since lines where clipped numerically, but this is part of the problem).  With clipping, lines close the the boundary, especially those subparallel or parallel to the border, will be clipped and look thinner than before where they exit.  This is correct behavior but "breaks" the tests.
Ideally, @KristofKoch, @joa-quim and @seisman should give this a try and comment on any issues they discover.


KristofKoch:
I came up with some tests as @PaulWessel requested and got some mixed results.

It fails in **test case 1** with a multi-segment line where a part of the line is outside the map region:

mulit-segment

The clipping works, but the "infill" is missing. If only the start and end point are outside the map region everything is fine.

**Test case 2 and 2.1** are interesting - here I have a line bowing into the map region and something goes wrong. The thin black line is there to show the actual great circle path. I'm aware that the Mercator Projection is far from ideal for high latitudes but it nicely illustrates the problem.

great_circle

While the clipping and infill at the border work as expected the line has an "interesting" line end at the middle line. I suspect this to be unrelated to the clipping of lines.

Things become interesting when I supply a `lat0` in test case 2.1:

great-circle-lat0

`-JMlon0/lat0/size` makes for some interesting lines. I suspect this to be another problem and not related with the line clipping we are talking about here.

The test cases 3 & 4 where I tested for crossing 0 and 180 degrees longitude worked as expected.

I added a `pscoast` call with a thick pen for the coastline as **test case 5**. My assumption that under the hood `pscoast`uses the same code as `psxy` is clearly wrong:

pscoast

However I would expect the same behavior for `pscoast` as for `psxy`.

**Test Script**

```
#!/bin/bash

# CASE 1
#
# 3 lines; mid latitudes, one single segment
# line and two multisegment lines. Of the multi-
# segment lines one has all points outside the 
# map region and one has only start and end
# points outside the map region
CASE1=test_clipping_line.ps


# CASE 2
#
# 3 lines; two great circle starting and endling 
# outside the map region in high latitudes with 
# a suboptimal projection and the line bowing 
# into the map and one line starting and ending
# inside the map region
CASE2=test_clipping_line_gc.ps

# CASE 2.1
#
# same as case 2 but additionally with lat0:
# -JMlon0/lat0/size
CASE21=test_clipping_line_gc_lat0.ps


# CASE 3
#
# 3 lines; one crossing 0deg longitude, one
# ending at 0deg longitude and one ending just
# before 0deg longitude
CASE3=test_clipping_line_0lon.ps


# CASE 4
#
# a line crossing +/- 180deg longitude
CASE4=test_clipping_line_180lon.ps


# CASE 5
#
# a pscoast call with a thick pen for the
# coastline
CASE5=test_clipping_line_pscoast.ps


# CASE 1

gmt psxy \
    -R7.5/49.5/9.5/50.5r \
    -JL8.5/50/49.75/50.25/15c \
	-Bwesn \
	-W35p,red \
	> $CASE1 << END
>
 7.6 49.4
10   49.9
>
 7   50
 8.5 50.55
10   50
>
 7   49.5
 8.5 50.1
10   50
END


# CASE 2 & 2.1

cat > gc.d << END
>
-74.5 81.9
-15.5 81.9
>
-74.5 84
-15.5 84
>
-85.5 85
-4.5  85
END

# case 2
gmt psxy \
    -R-85/82/-5/87r \
    -JM-45/15c \
	-Bwesn \
	-W35p,red \
		-K \
	> $CASE2 < gc.d

gmt psxy \
    -R -J \
	-Bwesn \
	-Wthinnest,black \
		-O \
	>> $CASE2 < gc.d

# case 2.1
gmt psxy \
    -R-85/82/-5/87r \
    -JM-45/84.5/15c \
	-Bwesn \
	-W35p,red \
		-K \
	> $CASE21 < gc.d

gmt psxy \
    -R -J \
	-Bwesn \
	-Wthinnest,black \
		-O \
	>> $CASE21 < gc.d

rm gc.d


# CASE 3

gmt psxy \
    -R-1/49.5/1/50.5r \
    -JL8.5/0/49.75/50.25/15c \
	-Bwesn \
	-W35p,red \
	> $CASE3 << END
>
-0.8  49.49
 2    49.9
>
-1.5  50.25
 0    50.51
>
 1.5  50.25
 0.01 50.51
END


# CASE 4

gmt psxy \
    -R179/49.5/-179/50.5r \
    -JL180/50/49.75/50.25/15c \
	-Bwesn \
	-W35p,red \
	> $CASE4 << END
-178.8 49.4
 178.8 49.9
END


# CASE 5

gmt pscoast \
	-R173/-40/178/-37.5r \
	-JL176/-38.5/-38/-39.5/15c \
	-Df \
	-Bwesn \
	-W35p,red \
	--PS_LINE_CAP=round \
	--PS_LINE_JOIN=round \
	> $CASE5
```

PaulWessel:
Thanks for the feedback.  I will have a look.

KristofKoch:
Hi Paul, thank you for looking into this. I played around with f0ec063 and wasn't able to break it with examples similar to case 1.

PaulWessel:
Apart from the fat-pen weirdness (#431) all of the examples in @KristofKoch's list look good to me now.  There are a few places where gmt_plot_line is used that is not yet set up to clip (pssac.c) and there are places where gmt_plot_line is not used to draw the lines (psxyz, pswlggle, etc).

KristofKoch:
I tried hard to break 40c52e7 with my use cases - no luck. Thank you very much @PaulWessel! Even `pscoast` looks nice now (that's what sparked the whole thing over [in the Forum](http://gmt.soest.hawaii.edu/boards/1/topics/8031))



bleeding pen
**OLD**




no bleeding
**NEW with 40c52e7**

Thanks again, this is awesome!

PaulWessel:
Great.  Does anyone have any objections to merging this PR into master (after I update the many PostScript plots that will change a bit after this new code is added)? E.g., @joa-quim or @seisman?

PaulWessel:
Need to fix a few (new) bugs.  There are a handful of tests that now fail due to "streaking" across the map from periodic boundaries.  oblique.sh I think is one of them?

PaulWessel:
Hi @GenericMappingTools/core and @KristofKoch, I have continued to work on this PR and I have finally fixed a few things that caused me trouble.  It now passes all the tests that master passes.  It would be great to have you try this branch and see if you can break it w.r.t. master.  Otherwise, I think we should merge it before the GMT 6b release.  Please try it this week and by the end of the week, if nothing bad is found (that is not also in master), we merge.

PaulWessel:
Reminder to @GenericMappingTools/core and @KristofKoch to give this branch a final go-through.

KristofKoch:
@PaulWessel I can test-drive this branch on Wednesday (05JUN19) earliest. My day-job keeps me currently busy.

PaulWessel:
Hi @joa-quim and @seisman : Last call for testing this PR.  After adding the RMS checks etc I only have two failures on macOS: test_cm4.sh (because the source files changed) and pseudo.sh (to be fixed). PLEASE build this PR and run the tests or whatever and let me know so I can merge with better confidence.

PaulWessel:
If @abanbj has time to test this PR too that would be great.

seisman:
Just play with this PR. It looks good to me.

abanbj:
I ran @KristofKoch 's 'test suite' and went fine. One small thing though; there is a small narrow 'plot' area that extends beyond the frame (zoom in). Is this a potential problem?

PaulWessel:
Hi @abanbj, could you be more specific where this is? I just ran them myself and zoomed in around the boundaries and no such thing.  It would be odd given the map clip has been activated.

abanbj:
Its the same for all figures generated with @KristofKoch 's scripts. E.g. *test_clipping_line*. Also I converted all PS to PDF if that makes a difference. Attached you see a zoomed in part of the frame. White is extending beyond the frame. Maybe all figures have this?! Might be nothing.

frame_white

PaulWessel:
I see no such thing, must be your gs version of viewer?  I am pretty sure since we set a clip path for that rectangle and nothing can show outside unless there is a gs bug.

abanbj:
Checked now with two different viewers, and it's all the same. So must be gs then I guess.

```
$ gs --version
9.26
```

PaulWessel:
How about @joa-quim and @seisman; do you see anything when zooming in.  I get this - clean as a whistle.  This is gs 9.26 as well, on macOS.  Below is conversion to PNG so no viewer involved.

test_clipping_line

joa-quim:
I see nothing strange. gs 2.22

abanbj:
NB! Ignore the gray stuff in my figure (thats just the pdf readers background). It's the white area outside the frame I'm thinking about.

seisman:
Not sure if we're talking about the same thing, as @abanbj's figure is a little misleading.

This is what I got after converting the plot to PNG or PDF.

image

I don't see any "colorful" pixels beyond the black frame. But like @abanbj said, I see some "white" regions after converting the plot to PDF or PNG formats. It may mean that the HiResBoundingBox is not exactly calculated. 

I also see something weird with the PNG figure. I can see some gray lines in the north and east sides of the frame. 

Upper left corner:
image

Lower right corner:
image

joa-quim:
I also see the grays but a bit differently (UL corner) but nothing in the PDF. It must be some anti-aliasing effect of the rasterization process.
Capture

PaulWessel:
Your gray lines are aliasing for fractional pixels.  If I run `gmt psconvert -Tg test_clipping_line.ps -A -E4800 ` and open the 28490 x 22130 PNG in Photoshop and zoom in real hard I see a tiny row of gray pixel beneath the large wall of black.  This is expected: The line thickness is not set up to be a multiple of pixels sizes.
I can confirm the thin white border.  This may be related to the Highres BB, not sure, but we can test this.  All of this should be independent of the particular PR I think.

PaulWessel:
Did not yet see your comment, @joa-quim but that is exactly it.

abanbj:
Thanks @seisman - that's exactly what I mean.

PaulWessel:
I just changed the printing of highres BB to using %.8f instead of %.4f and makes not difference to the width of the white pad.

PaulWessel:
Note that a plain `gmt psbasemap -R7.5/49.5/9.5/50.5r-JL8.5/50/49.75/50.25/15c -Bwesn > t.ps` yields exactly the same white gap, in both master and this PR.  So not related to the new fat line clipping.

abanbj:
You're totally right, Paul. But then, how come, if you include ``--MAP_FRAME_TYPE=INSIDE`` there is no such gap? Why would this make a difference?

```
gmt psbasemap -R7.5/49.5/9.5/50.5r -JL8.5/50/49.75/50.25/15c -Bwesn --MAP_FRAME_TYPE=INSIDE > t_i.ps
```

PaulWessel:
Not sure, but if you diff the PS produced with or without INSIDE you will see the difference is that INSIDE sets the clip path.  That path goes right down the middle of that black frame and hence it is black all the way.  The other case doe not set clippath and relies on boundingbox.  It is possible the highres BB is not as high res as we would like.

PR #445, opened 2019-02-27
Improve/fix the gmt_ogrread function and include in the lib code.

joa-quim:
Start the work of reading the OGR vector formats directly via GDAL. The ``gmt_ogrreaf()`` uses a design inherited from it's Matlab ancestor in Mirone and hence it's structure must be improved to fit better with the GMT needs. But as it is we can already load data directly. We still need to write code to parse its contents and feed the corresponding GMT data structures, so as it is we can't use it yet but as a proof o concept I added one such type of function parser in Julia that converts the gmt_ogrreaf() output into a ``GMTdataset``. It works like this.

```
# Create an API session
API = GMT.GMT_Create_Session("GMT", 2, GMT.GMT_SESSION_NOEXIT + GMT.GMT_SESSION_EXTERNAL + GMT.GMT_SESSION_COLMAJOR);

# Read a shape file directly from ... wherever it is
ex = GMT.gmt_ogrread(API, "/vsicurl/https://raw.githubusercontent.com/OSGeo/gdal/master/autotest/ogr/data/poly.shp");

# Plot it
plot(GMT.ogr2GMTdataset(ex), show=1, fmt=:png)
```
gmtjl_tmp

### DIFF
--- a/src/gmt_customio.c
+++ b/src/gmt_customio.c
@@ -1723,7 +1723,7 @@ int gmt_srf_write_grd (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *header, gmt
 #include "gmt_gdalread.c"
 #include "gmt_gdalwrite.c"
 #include "gmt_ogrproj.c"		/* For coordinate conversions but can "enter" here too */
-#if (GDAL_VERSION_NUM >= 2000)
+#if GDAL_VERSION_MAJOR >= 2
 #include "gmt_ogrread.c"
 #endif
 /* GDAL support */
--- a/src/gmt_customio.c
+++ b/src/gmt_customio.c
@@ -1723,7 +1723,9 @@ int gmt_srf_write_grd (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *header, gmt
 #include "gmt_gdalread.c"
 #include "gmt_gdalwrite.c"
 #include "gmt_ogrproj.c"		/* For coordinate conversions but can "enter" here too */
+#if (GDAL_VERSION_NUM >= 2000)
 #include "gmt_ogrread.c"
+#endif
 /* GDAL support */
 /*-----------------------------------------------------------
  * Format :	gd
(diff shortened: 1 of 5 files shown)


PaulWessel:
Looks like there are compiling errors in Travis for Linux - could you check and see what is missing (if anything)?

joa-quim:
Don't get it. It looks like it's not finding some GDAL pieces (can't be the whole GDAL because it doesn't complain with gdalread). It builds fine on OSX though.

PaulWessel:
Maybe @leouieda can sniff out what is happening?

joa-quim:
I bet it's our friend Ubuntu being _stable_. We need at least GDAL2.0 but it likely is still at 1.11 (a couple of years old)

PaulWessel:
Wonder if possible to isolate under a "#if GDAL_VERSION > 2" test?

joa-quim:
Yes. It would be ``#if (GDAL_VERSION_NUM >= 2000)`` but I would like to confirm that this is the source of the errors. My Winbuntu, where the same error occurs, refused to update GDAL beyond 1.11 (from 2015). What a beautiful world. I'm now upgrading the entire WSL OS but it promised to take the whole night to do that.

leouieda:
> I bet it's our friend Ubuntu being _stable_

Travis uses Ubuntu Xenial and the [latest gdal there is 1.11.3](https://packages.ubuntu.com/source/xenial/gdal). But I don't think there is a good way around this other than getting gdal from another source. This is why I use the conda packages for pygmt but that comes with its own set of headaches. 

If it breaks on Travis, you can bet it will break for all our Ubuntu users who follow our install instructions and don't build GDAL from source.

joa-quim:
OK, it turned out that upgrading Winbuntu was quicker then "the whole night" and I now have a 
```
j@MacMarilu:~/programs/build$ gdalinfo --version
GDAL 2.2.3, released 2017/11/20
```
With this it builds fine. So I don't understand why Travis is still failing since I added the #if (GDAL_VERSION_NUM >= 2000) in https://github.com/GenericMappingTools/gmt/pull/445/commits/160cb7662e2936e3c4e471e4ceef05904e9746d1

joa-quim:
Replaced the versioning condition by ``#if GDAL_VERSION_MAJOR >= 2`` it now passes also in Linux.

PR #447, opened 2019-02-27
Find git and unix commands before including ConfigCMake

### LINKED ISSUE #352: Short SHA missing from version information
This may be a deliberate thing, but if not;

The short hash that is appended to the version number (``6.0.0_[short SHA]``) is gone. I'm using d0302c4.
```
$ gmt --version
6.0.0
$ gmt

          GMT - The Generic Mapping Tools, Version 6.0.0 [64-bit]
(...)
(...)
```

seisman:
This PR fixes #352.

ConfigCMake needs GIT to determine the short SHA. However, the GIT variable is set after including ConfigCMake.cmake. This explains why the short SHA is missing. 

This PR changes the order of including ConfigCMake and finding git.


abanbj:
Looks good!

PR #449, opened 2019-02-27
Initialize to NULL

joa-quim:
This is in response to this compiler warning. But I'm not certain that's the only fix.
``c:\progs_cygw\gmtdev\gmt5\master\src\grdvector.c(650) : warning C4700: uninitialized local variable 'last_headpen' used``


PaulWessel:
Did you compile this before submitting? You cannot set a structure to NULL, only a structure pointer.  Need to use gmt_M_memset on this.

joa-quim:
Sorry, no, no compiling. Just did it on the web page because making branches for small things is becoming a bit boring.

PR #452, opened 2019-02-28
Add GMT QR EPS symbol to custom symbols

PaulWessel:
By adding the symbol QR.eps to the regular custom GMT symbols one can easily place the GMT QR Code on plots using plot **-Sk**QR/_size_.  For instance, this command makes the following plot:

`echo 0 0 | gmt psxy -R-2/2/-2/2 -JX4i -Baf -SkQR/3i -P > gmt-qr.ps
`

gmt-qr


KristofKoch:
Gentlemen,

I noted that the encoded URL is not a GMT related URL but some sort of shortening service giving http://l.ead.me/bb4l0i as target. 

As GMT plots tend to be around much longer than the average web service I would suggest to encode the URL to the GMT website directly.

While this makes the QR code four rows & columns larger it is way more future proof than relying on an external service.

I made one for your convenience pointing directly at http://www.generic-mapping-tools.org/

gmt2

Download the eps file: [gmt-qr.eps.zip](https://github.com/GenericMappingTools/gmt/files/2918375/gmt-qr.eps.zip)

you can easily check its encoded content at https://zxing.org/w/decode.jspx

leouieda:
@KristofKoch I agree with you on that. Thanks!

PR #459, opened 2019-03-01
Rename br.d to gmt_br.locale

seisman:


PaulWessel:
I will let our Portuguese and Brazilian developers slug this out... I seem to remember strong statements like "there is no Brazilian language - it is Portuguese!".

joa-quim:
Sure. I remember clearly to have said that some > 10 years ago. As well as that we didn't have that duplication for French or Spanish. The Fact that we have it for .uk and .us (forgeting the .au, .nz, .sa, .etc) instead of .en is also a sad historical feature.

leouieda:
>  "there is no Brazilian language - it is Portuguese!"

:rofl: there are actually several Brazilian languages but like the US languages, there is barely anyone speaking them anymore.

If GMT was a word processor, sure Brazilian and Portuguse portuguese are very different. But for mapping there is no significant difference.

seisman:
What should we do here? Delete `br.d` or leave it as it is?

PaulWessel:
I guess if there could be entries that would be different in the two versions of Portuguese.  It is the same for UK and US: Those files are identical, but if we were to add other things (e.g., spelling of meter vs metre) they could diverge.  Perhaps @leouieda and @joa-quim could advise/agree on if it is best to keep br for now in light of that or if spelling (unlike pronunciation) is the same always.

joa-quim:
Given that we talking about names of months, days of week and wind rose directions, where there is absolutely no difference between the way they are used in both countries, my opinion is that if we want to keep different files than we should add entries also (at least) to those listed in

https://en.wikipedia.org/wiki/List_of_territorial_entities_where_Portuguese_is_an_official_language
https://en.wikipedia.org/wiki/List_of_countries_where_Spanish_is_an_official_language
https://en.wikipedia.org/wiki/List_of_territorial_entities_where_French_is_an_official_language
https://en.wikipedia.org/wiki/List_of_territorial_entities_where_English_is_an_official_language

remkos:
This discussion would extend to nl_nl and be_nl as well.
We don't use those kind of <country>_<language> notations, so I would argue that our `pt.d` revers to the *language*, not the *country*. Same as `nl.d` connotes the Dutch language, not the country.

Yes, we have `us.d` and `uk.d`, as I can see there are in fact significant differences spelling, even in science, like "meter" and "metre".

Hence, unless we go as far as creating `pt_pt.d` and `br_pt.d`, I would say there is no `br.d`.

leouieda:
I agree that there is probably no need for a `br` version.

seisman:
I've removed `br.d`.

PR #460, opened 2019-03-01
Disallow negative contour intervals

### LINKED ISSUE #456: Issue warning if negative contour interval is given and suggestion to improve grdcontour
1. Issue a warning/error if the user specifies a negative contour *interval*. It has happened to me that I forgot to add the ``+`` to draw a single negative contour. E.g.

```
$ gmt grdimage @earth_relief_15m -RNO,IS -JS0/90/20c -K > temp.ps
$ gmt grdcontour @earth_relief_60m -R -J -C-500 -O >> temp.ps
```
This will give a  plot with no contours. 

2. If you want to draw a very limited amount of contours (e.g. one or two) contours, having to make a contour level file is cumbersome. What about having the ability to specify "several single" contours? E.g. if I want to draw the -2000, -1000 and -500 contour, I could specify ``-C+-2000,-1000,-500``.

PaulWessel:
Stuff like **-C**-100 should result in  an error since intervals cannot be negative.
Fixes #456

### DIFF
--- a/src/grdcontour.c
+++ b/src/grdcontour.c
@@ -401,8 +401,12 @@ GMT_LOCAL int parse (struct GMT_CTRL *GMT, struct GRDCONTOUR_CTRL *Ctrl, struct 
 				}
 				else if (opt->arg[0] == '+' && (isdigit(opt->arg[1]) || strchr ("-+.", opt->arg[1])))
 					Ctrl->C.single_cont = atof (&opt->arg[1]);
-				else
+				else if (opt->arg[0] != '-')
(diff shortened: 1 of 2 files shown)


leouieda:
@PaulWessel if this PR fixes the issue mentioned, writing `Fixes #456` will automatically close the issue. But it has to be those words and only one issue number.

PaulWessel:
Thanks @leouieda, I confirm that this Fixes #456.

leouieda:
:+1: updated the PR description

PR #466, opened 2019-03-04
Let the QR symbol honor -G and -W if set

PaulWessel:
This way the QR code can be set in a different color.  If **-W** is used then we also draw the outline of the quiet box around the code.  The quiet area is still painted white always.  We could decide to let users paint the quite area separately.  In most cases the background is probably white already but if it is a map or something it is probably OK to ask the user to paint that square.  Thoughts?


seisman:
@PaulWessel Please ignore the appveyor failures.

PR #473, opened 2019-03-05
Separate the making ellipse coordinates from plotting them

PaulWessel:
The old _gmt_geo_ellipse_ would create coordinates for an ellipse and plot them on a geographic map in psxy and psxyz.  However, there are times when we may want to output such coordinates.  I have split that function into two functions _gmt_get_geo_ellipse_ and _gmt_plot_geo_ellipse_ where the latter calls the former.  We can now consider other uses of gmt_get_geo_ellipse.  Test run the same after as before my changes.

### DIFF
--- a/src/t.lis
+++ b/src/t.lis
@@ -1,5 +1,0 @@
-./gmt_plot.c
-./gmt_prototypes.h
-./psxy.c
-./psxy_new.c
-./psxyz.c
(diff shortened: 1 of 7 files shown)


src/t.lis
```diff
@@ -0,0 +1,5 @@
+./gmt_plot.c
```
seisman: This file shouldn't be added.

Issue #474, opened 2019-03-05
Ability to create ellipse coordinates and sample along ellipse


PaulWessel:
Given the PR #473 it is possible to add the capability of creating ellipse coordinates from various modules in GMT.  I have two places in mind:

1. In grdtrack we have a modifier under **-E** that allows sampling of grids along a circle of given radius (**+o**_radius_).  We could allow a new modifier **+e**_major_/_minor_/_azimuth_ to do the same along an ellipse instead.
2. project allows users to create coordinates for a great or small circle.  I think this is the module best suited to take a new option to create an ellipse instead, e.g, **-M**_major_/_minor_/_azimuth_ in conjunction with **-C** and _-G_.

PaulWessel:
Decided at this point there is not a great need to add the +e modifier in grdtrack.  Users can pipe output from project if they need an ellipse since project can now make both a geographic and Cartesian ellipse path.  Thus closing this issue.

PR #479, opened 2019-03-05
Add option -Z to project for generating coordinates of ellipse

PaulWessel:
In conjunction with setting a center (**-C**) and an increment in km (**-G**), the new **-Z**_major_/_minor_/_azimuth_ option can specify the parameters of a geographic ellipse and return the coordinates that define its perimeter.  A modifier **+e** can be used to adjust the initial increment so that the adjusted increment fits exactly into the perimeter.

See discussion in issue #474 for some background.


PaulWessel:
I also updated the documentation and added an example to the list of examples in the man page.

PR #483, opened 2019-03-06
Remove misc since it dosen't exist

seisman:


src/gmt_supplements_module.c
```diff
@@ -57,7 +57,6 @@ static struct Gmt_moduleinfo g_supplements_module[] = {
 	{"mgd77path", "mgd77", "Return paths to MGD77 cruises and directories", ">D}"},
 	{"mgd77sniffer", "mgd77", "Along-track quality control of MGD77 cruises", ""},
 	{"mgd77track", "mgd77", "Plot track-line map of MGD77 cruises", ">X}"},
-	{"dimfilter", "misc", "Directional filtering of grids in the space domain", "<G{,GG},>DQ"},
 	{"grdppa", "misc", "Automatic extraction of ridge or valley axes", "<G{,>D}"},
```
seisman: Why was `grdppa.c` deleted in 9f24c96d09e42d66ddbca4b1505b5c18498e7847?

joa-quim: Well, ``grdppa`` is from an old algorithm and probably not at the GMT quality standards. The idea is interesting and useful but should probably resurrect in a form of more modern code. 
And yes, please delete the ``grdppa`` on ..._module.c though I think that one is generated automatically from a script.

PR #503, opened 2019-03-14
Enhance mapproject -W to return reference point coordinates

PaulWessel:
The **-W** option is used to obtain a map's width or height.  I have expanded it to take typical reference points (as used when placing embellishments, such as colorbar or roses) so that mapproject can return the geographic coordinates of the reference point.  For instance, if you need to map coordinates of the center of an oblique map, you use **-Wj**CM. Users can specify the reference point via **j**_code_, **n**_relx/rely_, or **x**_plotx/ploty_.  The inverse is also possible, i.e., using **g**_lon/lat_ and return the plot coordinates of the map point.
This format simplifies scripting by avoiding sending one-line coordinates into mapproject and using **-I** for such common calculations.

### DIFF
--- a/doc/rst/source/mapproject.rst
+++ b/doc/rst/source/mapproject.rst
@@ -27,7 +27,7 @@ Synopsis
 [ |-S| ]
 [ |-T|\ [**h**\ ]\ *from*\ [/*to*] ]
 [ |SYN_OPT-V| ]
-[ |-W|\ [**w**\ \|\ **h**] ]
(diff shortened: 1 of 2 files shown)


KristofKoch:
That's pretty cool – I could make great use of that. Saves me from figuring out from scratch how to charm `mapproject` into giving me the coordinates for the upper right corner every time again.

PaulWessel:
It should be especially handy for oblique projections and map scales, since in those cases we don't easily know map dimensions which would be needed to then invert for coordinates; this option does that all in one.

Issue #498, opened 2019-03-12
gmt docs behavior on server installations


KristofKoch:
**Description of the desired feature**

Have `gmt docs` behave differently on desktop and server installations.

* on desktop installations everything is fine.
* on server sessions there is no browser to open - therefore I'd like to suggest to display the URL to the online documentation and maybe a hint for the man page and a note that there is desktop environment detected. 

**Detailed …**

Currently `gmt docs` tries to open the default browser. While this is nice on desktop installations it understandably fails on server installations. 

```
$ gmt docs coast -J
sh: 1: xdg-open: not found
docs [ERROR]: Opening http://gmt.soest.hawaii.edu/doc/latest/coast.html#j in the default browser via xdg-open failed with error 32512
docs: No such file or directory
```

According to `xdg-open` man page: "xdg-open is for use inside a desktop session only."

As I normally SSH into a remote machine to work with GMT the current incarnation of `gmt docs` is almost useless to me. It fails in a way that is still usable but I guess this is accidental. 

While there are man pages I prefer working with the online documentation (better layout, links, easier to read). Therefore I would still appreciate the URL to the online documentation. This allows to easily call the URL on the local machine (Cmd-click in _iTerm_ on MacOS for example) while working on the remote machine.

**My suggestion**
```
$ gmt docs coast -J
You are working in a server environment. Try opening
http://gmt.soest.hawaii.edu/doc/latest/coast.html#j
in a web browser on your local machine. 
```

Your thoughts?

joa-quim:
Hmmm, this bit of code at the end of ``docs.c`` should have done almost what you propose. I wonder why the ``system`` command have not triggered that error case.
```
	sprintf (cmd, "%s %s", can_opener[k], URL);
	if ((error = system (cmd))) {
		GMT_Report (GMT->parent, GMT_MSG_NORMAL, "Opening %s in the default browser via %s failed with error %d\n", URL, can_opener[k], error);
		perror ("docs");
		Return (GMT_RUNTIME_ERROR);
	}
```

KristofKoch:
@joa-quim as far as I can see it did trigger (third line in my output). But it seems GMT assumes `xdg-open` to be present and doesn't check for it. This throws the first error message (`sh: 1: xdg-open: not found`) in my understanding.

joa-quim:
[...]

PR #510, opened 2019-03-15
Soften error message when docs cannot be opened automatically.

joa-quim:
Address issue #498

### DIFF
--- a/src/docs.c
+++ b/src/docs.c
@@ -169,8 +169,9 @@ int GMT_docs (void *V_API, int mode, void *args) {
 
 	sprintf (cmd, "%s %s", can_opener[k], URL);
 	if ((error = system (cmd))) {
-		GMT_Report (GMT->parent, GMT_MSG_NORMAL, "Opening %s in the default browser via %s failed with error %d\n", URL, can_opener[k], error);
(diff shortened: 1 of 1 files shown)


joa-quim:
I was just trying to avoid that ugly ``docs [ERROR]: `` since this not really a user error. Looked into the code but without stepping though it's really hard to tell if Report is writing to stderr or stdout. Is there a way (without fprintf) to write the text to stderror and not printing the ``docs [ERROR]: `` bit?

PaulWessel:
I think we should check the system return and if xdg-open is not found then it _is_ an error and we do as before.  However, if it is not that type of error then we can do a soft GMT_Message and then just do a printf of the URL.

joa-quim:
But this only about ``xdg-open``? And what other type of error would merit a soft message whilst this get the harsh one?
This issue is getting more complicated than it worth.

PaulWessel:
yes, I think open and cmd start will always be available.  I think only "not found" is an error.  Reasons for other failures would be considered soft and have to do with X11 tunneling etc.  Given that we are addressing an issue affecting 1% of users I think handling this 1st order error is more than we need to worry about.

PR #519, opened 2019-03-15
Fix bad allocation size.

joa-quim:


joa-quim:
``gmtread("Isle_wight.tif=gd", img=1, band=0);`` was not allocating ``whichBands`` because ``nn = atoi(prhs->B.bands)`` = 0

PR #528, opened 2019-03-18
Let docs take option to just print the URL

### LINKED ISSUE #498: gmt docs behavior on server installations

PaulWessel:
Giving the **-Q** option to docs means we simply prints the URL instead of opening it.  The **-Q** option must obviously follow the docs module name since there may be other module options following the module of interest.  Closes #498.

### DIFF
--- a/doc/rst/source/docs.rst_
+++ b/doc/rst/source/docs.rst_
@@ -9,7 +9,7 @@ Synopsis
 
 .. include:: common_SYN_OPTs.rst_
 
-**gmt docs** *<module-name>* [*-option*]
+**gmt docs** [ **-Q** ] *<module-name>* [*-option*]
 
 |No-spaces|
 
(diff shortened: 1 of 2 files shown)


PaulWessel:
quick implementation so and time to do before coffee.

PR #546, opened 2019-03-20
Could not set zero offset in grdgradient

PaulWessel:
Because of a bad check used in determining if **-N** set the offset, an explicit zero offset given as **-N+o**0 would always be ignored and the mean of the intensities would be used instead.  We now keep track of which modifiers were set during parsing instead of checking what the value is.


PaulWessel:
All tests pass as before, I am self-approving since nobody else up.

PR #558, opened 2019-03-22
Standardize -g to use modfiers and not -|+ as flags

PaulWessel:
The **-g** option for detecting data gaps had rarely used feature of giving a signed gap.  Then, the sign of the gap going from previous to current record mattered [normally we use the absolute value of any difference].  However, this use of **+** can interfere with modifier parsing.  Instead ,I have introduced modifiers **+n** (negative) and **+p** (positive) to handle gap signs.  Furthermore, the feature was not documented consistently and not used in any examples.  Finally, the option to require all gap specifications be satisfied (**-ga**) was incorrectly referred to as **-g:** in the main documentation.  The revision lead to no new failures in the 825 tests.


doc/rst/source/std_opts_classic.rst_
```diff
@@ -34,7 +34,7 @@
 | **-f**\ [**i**\ \|\ **o**]\ *colinfo*                               | Set formatting of ASCII input or output :ref:`(...) <-f_full>`           |
 +---------------------------------------------------------------------+--------------------------------------------------------------------------+
 | **-g**\ [**+**]\ **x**\ \|\ **X**\ \|\ **y**\ \|\                   |                                                                          |
-| **Y**\ \|\ **d**\ \|\ **D**\ *gap*\ [**u**]                         | Segment data by detecting gaps :ref:`(...) <-g_full>`                    |
+| **Y**\ \|\ **d**\ \|\ **D**\ *gap*\ [**u**][**+n**\ \|\ **p**\ ]    | Segment data by detecting gaps :ref:`(...) <-g_full>`                    |
```
joa-quim: Don't need the last ``\ ]``.  ``[**+n**\ \|\ **p**] `` is good eenough.

PR #560, opened 2019-03-22
Introduce common option -j and eliminate sign-based increments

PaulWessel:
GMT still has some awkward syntax here and there.  One of them is the idea of encoding how a spherical distance is computed via an optional sign to items that are positive definite.  E.g., you have a radius of 10 km but want the calculation to be done using ellipsoidal geodesics. That means specifying the radius as +10k, where the + symbol means ellipsoidal.  A quick-and-dirty flat Earth calculation requires -10k, where the - symbol does _not_ mean minus but 'flat Earth'.  To make matters worse, there are cases where we only want to pass the unit we want used for calculations, and then odd constructs like **+u+k** appears (modifier **+u** selects the unit, which is **+k**, i.e., ellipsoidal km).  But the **+k** also looks like a modifier in its own right, complicating unique parsing.  Clearly this had to go, but in a backwards compatible way.

Given the wide use of this scheme in numerous options throughout GMT, options that had a rich range of other modifiers, the best solution was to delegate the decision on how to compute distances to a dedicated (and new) GMT common option, **-j**_mode_.
The modes recognized are the same as before: **-jg** for great circle [Default], **-jf** for Flat Earth, and **-je** for ellipsoidal (geodesic).  I have updated source code an documentation to use the new syntax but in a backwards compatible way: The old syntax is no longer documented but is honored for a compatibility level of 6.  That way users will be warned of their deprecated syntax but it still works for them.  There are currently ~10 modules that uses **-j** but this may increase in the future.

I have run the 825 examples and there are no changes before and after using this branch.  I therefore think it is good to go but it would be helpful if @joa-quim and @seisman gave the branch a spin first toes if they notice something that should be fixed first.


joa-quim:
Run the tests and nothing suspicious.

seisman:
Looks good to me. 

@PaulWessel You committed a junk file `src/t.nc` in this PR. Remove it before merging.

PR #565, opened 2019-03-22
Check for =format in gmt_gdalwrite.c

### LINKED ISSUE #562: grdimage -A, mess in the file extension
The test ``readwrite_withgdal.sh`` now fails because the line that creates an image from a grid errors with:

```
gmt grdimage lixo.grd -Alixo.png=PNG -JX4c -Clixo.cpt
grdimage [ERROR]: Unupported image format. Supported formats are:
BMP,GIF,JPG,PNG & TIF
grdimage [ERROR]: Alternatively, append :<driver> for a valid GDAL driver
grdimage (api_export_image): Unknown Error [lixo.png=PNG]
```
The problem is in the file name ``lixo.png=PNG`` where the ``=PNG`` was not stripped out. In fact we don't need to specify the driver, but if we do, somewhere in the code we used to convert ``=PNG`` to ``=gd:PNG`` (which the form expected by ``gmt_gdalwrite``). That conversion was broken and likely recently because this test used to pass. This is also an annoying issue that keeps resurfacing again and again.

PaulWessel:
When a user gave output file as plot.png=PNG it failed to properly parse the driver.  Closes #562.

### DIFF
--- a/src/gmt_gdalwrite.c
+++ b/src/gmt_gdalwrite.c
@@ -115,6 +115,10 @@ int gmt_export_image (struct GMT_CTRL *GMT, char *fname, struct GMT_IMAGE *I) {
 			if ((c = strstr(fname, "=gd")))		/* Check if we have not a stray =gd in name*/
 				c[0] = '\0';					/* Yes, remove it too */
 		}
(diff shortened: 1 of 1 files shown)


PaulWessel:
Yes, not sure, but it seems a good place to make the check.  Will commit when checks complete.

PR #569, opened 2019-03-24
Let subplot set and common option -c advance to next panel in the sequence

PaulWessel:
Instead of always require _row,col_ of the next panel, let a lack of argument simply advance to the next panel in the sequence determined by **-A.**


PaulWessel:
Give it a spin.  I tested in a little bit, e.g., test/modern/panels.sh.

seisman:
Following script works for 2x2 subplots, but fails for 1x4 or 4x1.
```
gmt begin subplot png
gmt subplot begin 2x2 -Fs5c/3c -A'(a)'+JTL+o0.2c/0.4c -M0.5c/0.2c
for color in blue red lightblue lightred
do
    gmt basemap -R0/5/0/5 -Ba -BWSen+g$color -c
    echo 3 3 | gmt plot -Sc1c -W1p
done
gmt subplot end
gmt end
```

PR #606, opened 2019-04-05
Do not set tesselate when plotting points

PaulWessel:
The KML tessselate setting applies only to lines and polygons but a recent change addressing another issue inadvertently wrote out the tesselation setting even for points.  Now fixed. I also replaced some spaces with tabs.


PaulWessel:
Even more reasons to not do line breaks for moderately long lines...

PR #608, opened 2019-04-07
Allow pscontour and grdcontour to take comma-separated lists

### LINKED ISSUE #461: Allow specifying contours as comma-separated values
Contour plotting/dumping (``grdcontour``/``pscontour``) may be specified by 1) a cpt-file 2) a two column text file or 3) a fixed contour interval/single contour if prepended with a +. An additional useful option would be to be able to specify contours as a comma-separated list with ``-C``. If one wants to plot e.g. contours 200, 500, 1000 and 2000 this could be specified as ``grdcontour [..] -C200,500,1000,2000``.

PaulWessel:
This PR lets both contour and grdcontour accept a comma-separated list of specific contours.  The special case of a single contour needs to be given a trailing comma to not be confused with a contour interval.  This buries the old scheme of giving a single contour by using a leading +, which always looked odd for negative contours (e.g., **-C**+-10).  It is now **-C**-10, with its trailing comma, or **-C**10, for the positive 10 contour.  The old scheme is still supported of course but no longer documented.  The list mechanism allows both **-A** and **-C** to be used at the same time, provided they are both lists.  Closes #461.

### DIFF
--- a/doc/rst/source/contour.rst
+++ b/doc/rst/source/contour.rst
@@ -16,9 +16,9 @@ Synopsis
 
 **gmt contour** [ *table* ] |-J|\ *parameters*
 |SYN_OPT-Rz|
-[ |-A|\ [**-**\ \|\ [+]\ *annot\_int*][*labelinfo*] ]
+[ |-A|\ [**-**\ \|\ *contours*][*labelinfo*] ]
 [ |SYN_OPT-B| ]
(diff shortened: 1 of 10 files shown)


abanbj:
Great! Works!

PaulWessel:
Good, awaiting approval by @GenericMappingTools/core.

PR #622, opened 2019-04-10
Update ConfigDefault.cmake

PaulWessel:
Switch data URL to new domain oceania.generic-mapping-tools.org.

### DIFF
--- a/cmake/ConfigDefault.cmake
+++ b/cmake/ConfigDefault.cmake
@@ -116,7 +116,7 @@ endif (NOT DEFINED GMT_RELEASE_PREFIX)
 
 # Default location of remote data server
 if (NOT DEFINED GMT_DATA_URL)
-	set (GMT_DATA_URL "http://www.soest.hawaii.edu/gmt/data")
+	set (GMT_DATA_URL "http://oceania.generic-mapping-tools.org")
 endif (NOT DEFINED GMT_DATA_URL)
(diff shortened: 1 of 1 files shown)


leouieda:
@joa-quim yep, it should. But look at the address. It should be http://www.soest.hawaii.edu/gmt/data/ which is the current GMT_DATA_URL. So now we can do this: http://oceania.generic-mapping-tools.org/gmt_md5_server.txt

PaulWessel:
One day there will be a europe.generic-mapping-tools.org and your downloads will be sweet and fast.

leouieda:
@joa-quim we should try to setup a european mirror to test this out. All we need is an http server with the data files. Paul can register the redirect the europe.generic-... domain and it should be good to go.

joa-quim:
I thought  http://oceania.generic-mapping-tools.org was supposed to link to  www.generic-mapping-tools.org

PaulWessel:
No, those addresses are only for data.  We wont be doing mirrors of web pages.  www.generic-mapping-tools.org will eventually show a landing page with options to see docs, Julia, Discourse, Python, etc. We are not setting up a web browsing data directlry,.

leouieda:
We discussed this but there is no need for documentation mirrors. They are just plain HTML pages. And who knows from where Github hosts these files.

PR #623, opened 2019-04-10
Fix scripts in tools directory

seisman:


seisman:
> I think there is no longer a postscriptlight subdirectory.

Maybe we can keep it for compatibility.

PaulWessel:
That is fine by me.

Issue #638, opened 2019-04-15
Let the API return an empty set instead of NULL?


PaulWessel:
Modules such as gmtselect, blockmean, and many others perform record-by-record i/o.  When these are called by the API in, say, MATLAB, and the particulary combination of input data and options produce no output, the API simply returns NULL instead of an empty data set.  I think this leads to special checks in the external interface to handle these cases whereas it would be better that these modules simply returned an empty dataset.  Perhaps @joa-quim could test the equivalent of this in Julia:

`echo 0 0 0 | gmt blockmean -R1/2/1/2 -I1
`

and see if it returns an empty set of crashes.

joa-quim:
Haven't we addressed this before?

```
julia> blockmean([0.0 0 0], R="1/2/1/2", I=1)
0-element Array{GMT.GMTdataset,1}
```

PaulWessel:
Hm, interesting.  Would you be able to run that in GMTMEX?  I brought this issue up because a specific plot using SRTM collects the tiles and then uses the region perimeter as points fed to gmtselect -N to determine if any of the points are not on land.  If not then we add in the 30s grid to fill in the oceans.  In my case none of the points were on land so the gmtselect call returned a NULL data set and crashed grdimage...

joa-quim:
``gmtmex('blockmean -R1/2/1/2 -I1', [0 0 0])`` crashes ML. Maybe I have some test to catch this case in Julia because otherwise J & ML behave exactly the same.

PaulWessel:
OK "good", that makes sense to me; Julia (or you) must do some other check somewhere.

PaulWessel:
Fixed crash in gmtmex which was caused by foolish behavior: accessing table[0]->n_headers in the case when there are no tables in the structure.  This is why it crashed in mex but not in Julia since you probably do not do the same thing there.

PR #645, opened 2019-04-17
Return an empty dataset instead of NULL

PaulWessel:
If a record-by-record writing module ends up not writing any records, we make sure the API passes an empty dataset back instead of a NULL pointer.  This addresses issue #638.


PaulWessel:
Ideally please rebuild and try the mex again to see if no longer crashing.

joa-quim:
Still crashes ML

PaulWessel:
OK, well, will need to debug then some time after this week ends...

joa-quim:
No crash now.

PR #650, opened 2019-04-22
Let the implicit -R be saved in gmt.history too (classic mode)

joa-quim:
In classic mode this, for example, ``gmt grdimage lixo.grd -JX10 > ...`` would not save the data's -R in ``gmt.history`` (it does for Modern mode). This feature is needed in the Julia wrapper to transmit -R to next command saving the user to do it explicitly. This PR solves that (and could probably be improved to avoid repeat code).


joa-quim:
I have no idea on the work that it would take to rewrite the wrapper to use modern mode (quite a bit for sure). And as I told you, that would mean loosing the ability to run the Julia wrapper from GMT5. Don't think that's a such a good trade off. Now that I think of it, this looks more a bug of classic mode that affects also the mex wrapper.

PaulWessel:
Maybe if you promise to give me a free beer at AGU I could be convinced that perhaps our implicit use of the grid domain to set -R should under classic mode result in a -R in gmt.history.  But as you implemented it there is the extension to examining data sets etc.

joa-quim:
It was damn hard to find how to make the implicit -R land into gmt.history, so I copied the chunk of code that puts it into ``GMT->init.history`` (that's why I made the comment that the PR could be improved to avoid code repetition). 
And if you want to be beer-bribed I'm fine with it.

PaulWessel:
Testing but breaking many tests now - will try to figure out why.

PaulWessel:
I am ready to give up on this - it is taking me down too many rabbit holds to redesign something to mimic modern mode.  OK to give up?

joa-quim:
OK, it will just make the wrapper a little bit slower  since I’ll have to do the region inquire from more gmt calls. 

Sent from my iDedo

No dia 12/05/2019, às 20:45, Paul Wessel <notifications@github.com> escreveu:

PaulWessel:
Sorry I was not able to do this one.  Closing this PR.

PR #651, opened 2019-04-23
Azure Pipelines CI for Windows via conda

seisman:
This PR adds the azure-pipelines CI for Windows. The dependencies (netcdf, gdal, et al.) are installed via conda.

The compilation takes about 12 minutes. And the simple tests all passed except the last one.
```
+ gmt begin
begin [ERROR]: Unable to create a workflow directory : C:/Users/VssAdministrator/.gmt/sessions/gmt6.3384
```


ci/azure-pipelines-windows.yml
```diff
@@ -0,0 +1,38 @@
...
+  displayName: Add conda to PATH
+
+- bash: |
+    set -x -e
+    conda config --set always_yes yes --set changeps1 no --set show_channel_urls true
+    conda install libnetcdf libcurl libgdal pcre
+    conda list
+  displayName: Install dependencies via conda
+
+- bash: echo "##vso[task.setvariable variable=INSTALLDIR]D:/a/1/s/gmt-install-dir"
```
seisman: Here, I use `D:/a/1/s` rather than `$BUILD_SOURCEDIRECTORY` (D:\a\1\s).

If `$BUILD_SOURCEDIRECTORY` is used, then 
```
GMT->session.SHAREDIR = D:s/gmt-install-dir/share
```
and gmt will fail to find some files in share directory.

ci/build-gmt-windows.sh
```diff
@@ -0,0 +1,53 @@
...
+
+echo ""
+echo "Using the following cmake configuration:"
+cat cmake/ConfigUser.cmake
+echo ""
+
+mkdir -p build && cd build
+
+# Configure
+cmake .. -G "Visual Studio 15 2017 Win64" -DCMAKE_PREFIX_PATH=${CONDA}/Library
```
seisman: `-DCMAKE_PREFIX_PATH=${CONDA}/Library` is necessary to let cmake know where to find the DLL files.

leouieda:
>  And the simple tests all passed except the last one.

Maybe try setting the GMT home directory to `D:/a/1/s/`. This seems to be a permission problem. Azure is kind of weird.

joa-quim:
I'm not sure on how I can help here. I see that the build fails because
```
begin [ERROR]: Unable to create a workflow directory : C:/Users/VssAdministrator/.gmt/sessions/gmt6.3384
```
though this is not completely new for me (I think I remember to have had people asking me about some similar problem), I never had it, so hard to debug.

Regarding the ``D:\a\1\s``, maybe you need to escape the slashes. I.e. ``D:\\a\\1\\s``

seisman:
The windows build passes now. It takes ~16 minutes.

ci/azure-pipelines-windows.yml
```diff
@@ -0,0 +1,43 @@
+# Template for the basic Windows steps in Azure Pipelines
+
+steps:
+
+- bash: echo "##vso[task.prependpath]$CONDA\Scripts"
+  displayName: Add conda to PATH
+
+- bash: |
+    set -x -e
+    conda config --set always_yes yes --set changeps1 no --set show_channel_urls true
```
leouieda:
```suggestion
    conda config --set always_yes yes --set changeps1 no --set show_channel_urls true
    conda install pycryptosat --yes
    conda config --set sat_solver pycryptosat
    conda config --set safety_checks disabled
```

We're spending 8.5 min just installing dependencies. Try adding these lines to see if we can get conda to resolve the dependencies faster. Maybe it will cut down on the time a bit.

ci/build-gmt-windows.sh
leouieda: Is there any way to do parallel builds?

joa-quim: Ninja does parallel builds. That's what I use.

seisman: I tried ninja once but failed. Have you tried Ninja on your machine? @joa-quim

seisman:
@leouieda Now the Windows build takes ~9 minutes.

Issue #656, opened 2019-04-24
When the coordinate is a character string, not a number, the internal storage memory used by gmt6 increases without end ON MAC


wangliang1989:
When the coordinate is a character string, not a number, the internal storage memory used by gmt6 increases without end ON MAC.

**Full script that generated the error**

```
#!/bin/bash

gmt begin
gmt figure draw_example_b png A1c
gmt coast -JH30c -Rg -BSWen -Bxaf -Byaf -Bag -Dl -A250 -Ggrey -Scornsilk
gmt plot -W2p,blue,solid -Am << EOF
>
100 -85
100 85
>
30 60
150 abc
EOF
gmt end
```

**Full error message**

```
No message error, but program won't finish
```

**System information**

* Operating system: MacOS
* Version of GMT: gmt6

PaulWessel:
Thanks for pointing that out - it is being fixed via #659.

PR #659, opened 2019-04-24
Make sure there are no NaNs in fixup_path function

PaulWessel:
When input contains junk you may end up with NaN in the coordinate array.  As per issue #656 some modules will hang when trying to interpolate unless **-A** is used.  Now it will give a warning and skip the interpolation.

### DIFF
--- a/src/gmt_vector.c
+++ b/src/gmt_vector.c
@@ -1396,7 +1396,7 @@ uint64_t gmt_fix_up_path (struct GMT_CTRL *GMT, double **a_lon, double **a_lat, 
 	lon = *a_lon;	lat = *a_lat;	/* Input arrays */
 
 	if (gmt_M_is_dnan (lon[0]) || gmt_M_is_dnan (lat[0])) {	/* If user manages to pass NaN NaN records then we check on the first record and bail */
-		GMT_Report (GMT->parent, GMT_MSG_VERBOSE, "Your data contains NaNs - no resampling taken place!\n");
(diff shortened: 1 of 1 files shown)


seisman:
Users cannot see the warning unless -V is used. Is it desired?

PaulWessel:
It is complicated.  For many purposes it is not that harmful to have junk in the data causing NaNs.  We have in fact a GMT defaults that controls this:

`--IO_NAN_RECORDS=skip|pass
`

The default is _pass_, meaning reading junk does not necessarily cause an error.  So in the users case the data was read and the second latitude was NaN.  Then, plot wants to resample along a great circle and it did not check for NaNs, thus hanging.  For those reasons, we have not flagged these as errors.  One can set **--IO_NAN_RECORDS**=_skip_ and then we get

psxy [ERROR]: Encountered first invalid ASCII data record near/at line # 3

> psxy [ERROR]: Likely causes:
> psxy [ERROR]: (1) Invalid x and/or y values, i.e. NaNs or garbage in text strings.
> psxy [ERROR]: (2) Incorrect data type assumed if -J, -f are not set or set incorrectly.
> psxy [ERROR]: (3) The -: switch is implied but not set.
> psxy [ERROR]: This file had 1 data records with invalid x and/or y values
> 

That is why I am only issuing the new message via **-V**.

seisman:
Sounds reasonable. Good to merge.

PR #662, opened 2019-04-25
Add USTC mirror address in README

wangliang1989:
Add USTC mirror address in README


seisman:
we need check the mirror list and fix these ftp addresses.

PR #671, opened 2019-04-29
Improve test for geometry inside the API

PaulWessel:
Since gmtmex may not know what a dataset will be used for it assigns it the geometry PLP (point, line, polygon).  Later, we encountered a check specifically for point, which is part of PLP, but we used equality instead of logical and.  @joa-quim, this should allow that greenspline 1-D test to work in mex.


joa-quim:
Yes, it works now. Thanks

PR #694, opened 2019-05-07
pspolar: Let beachball size scaling with magnitude, like psmeca

### LINKED ISSUE #387: pspolar bubble scaling
Hello,
We know that psmeca beach ball size is scaled to a magnitude 5, whereas pspolar bubble is not. If one is using psmeca before using pspolar in a script, then one has to scale the bubble of the pspolar manually. e.g.

example shell code
mag=2.5
ball_size='4i'
polar_ballsize=`echo "scale=3; ${ball_size::${#ball_size}-1}*$mag/5" | bc`

Can't you use the same function for scaling in psmeca and pspolar? This sounds like a reasonble modification to pspolar function.

seisman:

This PR add the feature discussed in #387 to pspolar, and fixes some typos in documentations. Closes #387.

### DIFF
--- a/doc/rst/source/supplements/meca/polar.rst
+++ b/doc/rst/source/supplements/meca/polar.rst
@@ -15,10 +15,10 @@ Synopsis
 
 **gmt polar** [ *table* ] |-D|\ *lon/lat* |-J|\ *parameters*
 |SYN_OPT-R|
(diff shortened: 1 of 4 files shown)


79seismo:
Thanks very much!

PR #711, opened 2019-05-12
Add -J to grdedit without changing projection environment

PaulWessel:
Here, **-J** is not used to project data, just supplying meta data.

### DIFF
--- a/src/grdedit.c
+++ b/src/grdedit.c
@@ -243,7 +243,7 @@ GMT_LOCAL int parse (struct GMT_CTRL *GMT, struct GRDEDIT_CTRL *Ctrl, struct GMT
 
 int GMT_grdedit (void *V_API, int mode, void *args) {
 	/* High-level function that implements the grdedit task */
-	bool grid_was_read = false;
+	bool grid_was_read = false, do_J = false;
 	
 	unsigned int row, col;
 	int error;
@@ -252,7 +252,7 @@ int GMT_grdedit (void *V_API, int mode, void *args) {
 
 	double shift_amount = 0.0;
 
-	char *registration[2] = {"gridline", "pixel"}, *out_file = NULL;
+	char *registration[2] = {"gridline", "pixel"}, *out_file = NULL, *projstring = NULL;
 
 	struct GRDEDIT_CTRL *Ctrl = NULL;
 	struct GMT_GRID *G = NULL;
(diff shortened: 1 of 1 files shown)


src/grdedit.c
```diff
@@ -278,6 +278,13 @@ int GMT_grdedit (void *V_API, int mode, void *args) {
 
 	/*---------------------------- This is the grdedit main code ----------------------------*/
 
+	if ((do_J == GMT->common.J.active)) {	/* Only gave -J to set the proj4 flag, so grid is already projected (i.e., Cartesian) */
```
joa-quim: Maybe you meant 
    if ((do_J = GMT->common.J.active)) {

src/grdedit.c
```diff
@@ -278,6 +278,13 @@ int GMT_grdedit (void *V_API, int mode, void *args) {
...
+		projstring = gmt_export2proj4 (GMT);	/* Convert the GMT -J<...> into a proj4 string */
+		gmt_set_cartesian (GMT, GMT_IN);	/* Force Cartesian since processing -J changed that */
```
joa-quim: Nope. We also may want to explicitly set -J"+proj=longlat ..."

PaulWessel: Yes, meant assignment.  I can exclude the longlat case from setting Cartesian; that is the only case no?

joa-quim: From what remember now, yes. But notice that we can also say +proj=latlong

PaulWessel:
Same result as before since unaffected by the longlat/latlong check.  So good to go I think.

joa-quim:
But you also changed the assignment issue.
In your test the -R is still getting modified.

x_min: -556237.453966 x_max: 556957.453966

PaulWessel:
Isn't the whole point that R gets changed to meter so it matches the specific projection?  Otherwise -r has random inches of some sort.

joa-quim:
You gave it -R-556597.453966/556597.453966/1109503.57278/2260011.53508 
Why would this be modified by the program?

PaulWessel:
I see; -556597.453966/556597.453966 becomes -556237.453966/556957.45396608 which is way larger than numerical noise.  It has to do with setting -J and we get stuff like
grdedit [WARNING]: Warning -R: Given west and east values [-556597 556597] were adjusted so not exceed multiples of 360 [-556237 556957].  I will try to parse it a second time after undoing the -J.

PaulWessel:
OK, should work now.  Give the original region I set via -R.

joa-quim:
Yep, that's the idea.

PR #728, opened 2019-05-17
Drop find_programs gs

seisman:
`gswin32` is another possible name of ghostscript.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,7 @@ set (CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules/"
 # Find UNIX commands
 include (FindUnixCommands)
 find_package (Git)
-find_program (GS gs gswin64)
+find_program (GS NAMES gs gswin64 gswin32)
 find_program (XZ NAMES xz)
 
 # Include configuration options (default options and options overridden by user).
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,6 @@ set (CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules/"
 # Find UNIX commands
 include (FindUnixCommands)
 find_package (Git)
-find_program (GS NAMES gs gswin64 gswin32)
 find_program (XZ NAMES xz)
 
 # Include configuration options (default options and options overridden by user).


CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ set (CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules/"
 # Find UNIX commands
 include (FindUnixCommands)
 find_package (Git)
-find_program (GS gs gswin64)
+find_program (GS NAMES gs gswin64 gswin32)
```
joa-quim: Actually I think what is used on Windows are the gswin64c/gswin32c but I don't know why do we search for the ghost on Windows. The only module that uses it is _psconvert_, and it finds the executable via a registry inquire, which was a complicated thing to achieve because Windows64 have two registries one for 32 and another 64 and the Ghost will be in either one or the other depending on its bitage.

seisman: You're right. Seems we even don't need to search gs on Linux. The variable `GS` is never used in the cmake configuration or the psconvert source code. Maybe we can remove the `find_program (GS ...)` line?

joa-quim: Indeed, it looks we can drop it.

PaulWessel:
Question: Since GMT requires gs to be present, especially modern mode, I think finding gs was our proxy for making sure the user has gs installed.  Will this go away now?  I.e., users can build GMT if they do not have gs installed?

seisman:
`find_program (GS gs)` tries to find gs and save its full path to the variable GS. However, if gs isn't installed or isn't found, `find_program` doesn't give any warnings or errors.

joa-quim:
But I doubt it ever detected the presence the Ghost on Windows that does not install itself in PATH. This is what we do in _psconvert_
```
#ifdef WIN32
	if (ghostbuster(GMT->parent, C) != GMT_NOERROR)  /* Try first to find the gspath from registry */
		C->G.file = strdup ("gswin64c");     /* Fall back to this default and expect a miracle */
#else
```

seisman:
If you want to make sure gs is installed, we need something like
```
find_program (GS NAMES gs gswin64c gswin32c)
if (NOT GS)
message (..)
endif (NOT GS)
```

PaulWessel:
Well, I think we need to; it is not an optional install in my mind since psconvert calls it, and movie calls psconvert indirectly per frame via gmt end.

seisman:
Strictly speaking, gs is a runtime dependency but not necessary when building GMT. I agree we need to check if gs is available.

PaulWessel:
Yes, and I dont know what the standard protocol is for tools that depends on executables.  There must be other packages that depends on executables instead of includes and libraries.  What do they do?

seisman:
I don't know if there are any standard way to do that. But we have a FindSphinx.cmake in cmake/modules. And this file (https://github.com/Kitware/CMake/blob/master/Modules/FindCygwin.cmake) shows that we can search gs from registry.

seisman:
This PR is already merged and there is no way to reopen it. I've opened an issue #733 in order not to forget it.

PaulWessel:
What you think, @joa-quim ? Is this something to pursue or is it just more work for no gain?

joa-quim:
Let's drop it. I see no real gain in detecting gs presence while building on Win

PR #731, opened 2019-05-18
Azure Pipelines CI for Windows via vcpkg

seisman:

This PR adds the azure-pipelines CI for Windows. It's similar to #651, except that #651 uses conda to install dependency libraries, while this PR uses vcpkg to build all libraries from source.

Currently the whole build takes ~70 minutes, and building dependencies via vcpkg takes more than 60 minutes. The build time can be reduced to <5 minutes after Azure Pipelines supports caching, which should be available in next few weeks (https://github.com/Microsoft/azure-pipelines-yaml/pull/113).

Although now the Windows build is extremely time-consuming, it's still helpful to run the Windows builds in cron jobs.


joa-quim:
Building GDAL in vcpkg used to take an eternity due to a dependency on BOOST but I think I dropped that with a PR to the GDAL portfile some months ago.

seisman:
@GenericMappingTools/core What do you think about this PR? Can we merge it and run the Windows builds nightly?

Issue #736, opened 2019-05-19
GMT system calls for executables


PaulWessel:
The program **docs** makes system calls to a program that can open files.  The program depends on the OS.  Currently, we have

- Windows: Call **cmd /c start**
- OS X: Call **open**
- Else: Call **xdg-open**

I am pretty sure that the cases for Windows and OS X are find as **cmd** and **open** are part of the standard install, but I do not think **xdg-open** is a guaranteed application on all other Unix or Linux platforms.  Thus, perhaps we need CMake to test for this program and if not found we must complain.  Furthermore, the requirement of **xdg-open** should be discussed as one of the prerequisite applications together with **gs**, **gm**, **ffmpeg**, and **ogr2ogr**.

PaulWessel:
Closed as implemented.

PR #742, opened 2019-05-19
Look for required executables and warn if not found

PaulWessel:
Sine GMT uses the executables gs, gm, ffmpeg, and xdg-open we now look to see if they are present in the path and warn if they are not.  Compilation will continue regardless of result. See #736 for context.


seisman:
Don't forget `ogr2ogr`.

src/CMakeLists.txt
```diff
@@ -39,6 +39,56 @@ endif (WIN32)
...
+# Find executable needed by gmt
+find_program (OGR2OGR ogr2ogr)
+if (NOT OGR2OGR)
+	message (WARNING "Cannot find program ogr2ogr - gmt will be unable convert Shapefiles to GMT/OGR format")
+	set (GMT_CONFIG_OGR2OGR_MESSAGE "no" CACHE INTERNAL "OGR2OGR config message")
+else (NOT OGR2OGR)
+	set (GMT_CONFIG_OGR2OGR_MESSAGE "yes" CACHE INTERNAL "OGR2OGR config message")
+endif (NOT OGR2OGR)
+# Find executables needed by psconvert
+find_program (GS NAMES gs gswin64c gswin32c)
```
seisman: `find_program (GS NAMES gs gswin64c gswin32c)` doesn't work for Windows, as psconvert searches gswin32c/gswin64c by reading registry.

PaulWessel: So perhaps I just place that under if (UNIX)?

joa-quim: The true reason is that GS doesn't add to the PATH so any program that wants to find it must look into the registry.

PaulWessel: So is this also true for ogr2ogr, ffmpeg, gm, etc., etc? I.e., place those under a UNIX test?

seisman: find_program can search the paths via registry.
See https://github.com/Kitware/CMake/blob/master/Modules/FindCygwin.cmake.

seisman: > So is this also true for ogr2ogr, ffmpeg, gm, etc., etc? I.e., place those under a UNIX test?

I don't think we should place all those under a UNIX test. For Windows, if cmake cannot find `ogr2ogr` in PATH, then we also won't be able to run it by a system call.

As I understand it, `gs` is the only special case, because we search it both in PATH and in registry.

PaulWessel: OK, I will only isolate the gs check.

PaulWessel: @seisman and @joa-quim, let me know if this is OK now.

joa-quim: I agree it's good enough.

PaulWessel: OK, need someone to review and OK

PR #750, opened 2019-05-21
Prevent movie label overflow and increase to 32

PaulWessel:
The repeatable movie **-L** option can now handle up to 32 labels, and this is now documented better.  Other changes are spelling and typos.


seisman:
Is this PR the same as #748?

PaulWessel:
Yes, how did that happen... I am closing it.

PR #758, opened 2019-05-23
Update instructions about how to install dependencies for Archlinux and FreeBSD

holishing:
Briefly tested on those platforms and found related packages.
So adding instructions for those two platforms to let people who use those platform easier to set up developing environment.


seisman:
@holishing Can you also help add some instructions about how to install the pre-built GMT binary package on these platforms? The instructions should be added to INSTALL.md.

PR #759, opened 2019-05-23
Missing 'k' annot unit description

joa-quim:


doc/rst/source/explain_-B_full.rst_
```diff
@@ -80,6 +80,7 @@
     - **u** (ISO week, plot using 2 digits)
     - **r** (Gregorian week, 7-day stride from start of week :ref:`TIME_WEEK_START <TIME_WEEK_START>`)
     - **K** (ISO weekday, plot name of day)
+    - **k** (weekday, plot name of weekdays in selected language)
```
seisman: Below is the description from GMT_Docs.rst. 
image

`k` should be `Plot number of day in the week`

joa-quim: But it's missing in https://www.generic-mapping-tools.org/gmt/latest/gmt.html#b-full (that uses ``explain_-B_full.rst_``
And when we see the online print of ``psbasemap -``, it says (what I have copied)
```
               K: ISO weekday - format annotation according to FORMAT_DATE_MAP.
               k: weekday - plot name of weekdays in selected language [us].
```
Checking what it really does, ``Plot number of day in the week`` is the right description

joa-quim: There are more discrepancies between the two docs, but I have a further problem with this PR. I made it directly from the Github  page and now I don't know how to modify it.

seisman: First click "Files changed" tab, then click the three dots at the upper right corner (just below the "Review changes" button), then you can edit this file again. But remember to directly push the changes to the branch `joa-quim-patch-1` instead of creating a new one.

seisman:
@PaulWessel This PR shouldn't be merged. As @joa-quim said, there are some discrepancies about the meaning of K and k, between the documentation and `gmt psbasemap -` print.

PaulWessel:
Sorry, I should be able to fix this to match reality.

PR #766, opened 2019-05-24
Use cmake in building instructions

### LINKED ISSUE #764: Use cmake instead of make in building instructions

In [BUILDING.md](https://github.com/GenericMappingTools/gmt/blob/master/BUILDING.md), we use make commands 
```
make -j
make -j docs_man 
...
make -j install
```
to build the sources, documentations and running tests.

`make` is not cross-platform compatible. It works for cmake generators like "Unix Makefiles", but not for Ninja and Visual Studio.

I think we can change make to cmake, like
```
cmake --build .
cmake --build . --target docs_man
...
cmake --build . --target install
```

making the "make commands" platform independent.

seisman:
Fixes #764


PaulWessel:
Should we announce somewhere that once can select the build tool to be make, ninja, or ???, which may add one more requirement (e.g., ninja)?  I assume that the average person building GMT has not yet heard of ninja and why one may wish to use it.

Issue #763, opened 2019-05-24
grdlandmask with -A+aiS option


macpurity:
**Description of the problem**

When issuing a grdlandmask command to create a southern hemisphere mask for Antarctica, and then making a world plot of test.grd, it seems that zones above -60° also have been masked.

Documentation indicates that all features north of 60°S should be skipped, but they're not.

**Full script that generated the error**

```
gmt grdlandmask -Gtest.grd -I0.01/0.01 -R-180/180/-90/0 -A+aiS -Df -r
```

**Full error message**

```
No error message - just erroneous results
```

**System information**

* Operating system: OSX 10.13.6
* Version of GMT: Version 6.0.0_ac8b8ab_2019.05.20 [64-bit] [8 cores]

macpurity:
Example of map of test.grd

macpurity:
It seems the option order matters! Does -A+aSi operates differently than -A+aiS? I say yes!

This command worked, as expected, doing a 10° wide (lon) slice in low resolution GHSSG:

gmt grdlandmask -Gtest.grd -I0.01/0.01 -R30/40/-90/0 -A+aSi -Dl -r

But, switching to a higher resolution gave segmentation fault errors:

gmt grdlandmask -Gtest.grd -I0.01/0.01 -R30/40/-90/0 -A+aSi -Dh -r
ERROR: Caught signal number 11 (Segmentation fault) at
0   libgmt.6.dylib                      0x0000000108e062e5 gmt_assemble_shore + 3557
1   ???                                 0x0000010000000001 0x0 + 1099511627777
Stack backtrace:
0   libgmt.6.dylib                      0x0000000108d7a2e4 sig_handler + 580
1   libsystem_platform.dylib            0x00007fff6f36ef5a _sigtramp + 26
2   ???                                 0x00007fffebe864a2 0x0 + 140737151263906
3   libgmt.6.dylib                      0x0000000108f92b9c GMT_grdlandmask + 5692
4   libgmt.6.dylib                      0x0000000108d94464 GMT_Call_Module + 628
5   gmt                                 0x0000000108d6bede main + 1070
6   libdyld.dylib                       0x00007fff6f060015 start + 1

Something is definitely amiss. The command, when switching +aSi to +aiS does NOT skip zones above -60°.

gmt grdlandmask -Gtest.grd -I0.01/0.01 -R30/40/-90/0 -A+aiS -Dl -r

PaulWessel:
[...]

PR #768, opened 2019-05-25
Correct parsing of -A for modules using GSHHG data

PaulWessel:
For modules that use the GSHHG dataset, the **-A** option offers a special modifier **+a** affecting Antarctica: One can select either the **i**ce front or **g**rounding line as the coastline, and there are expert settings to either **s**kip _all_ of Antarctica or **S**kip everything _but_ Antarctica.  There were several problems:

1. It was not possible to select two **+a** modifier codes simultaneously, such as _both_ **g** and **S**.
2. The flags used internally to know what to do where not powers of 2 and combined poorly.
3. Logical tests involving these flags therefore were subject to unintended consequences.

I have updated the internal flags to be compatible with multiple selections and I have fixed various logical tests to use bit-comparisons instead of plain equality.  The parsing of **-A** now allows one or two codes to be appended and the documentation has been updated accordingly.  I will also port this to the 5.4 branch.  This PR closes issue #763 .

### DIFF
--- a/doc/rst/source/coast.rst
+++ b/doc/rst/source/coast.rst
@@ -15,7 +15,7 @@ Synopsis
 
 **gmt coast** |-J|\ *parameters*
 |SYN_OPT-R|
-[ |-A|\ *area* ]
+[ |SYN_OPT-A| ]
 [ |SYN_OPT-B| ]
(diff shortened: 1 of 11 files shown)


PaulWessel:
Thanks, done for both PRs.

Issue #784, opened 2019-05-27
-c option is missing in usages and documentations of all modern plotting modules


seisman:
All modern plotting modules don't print -c option in their usage and documentations.

PaulWessel:
OK, I will fix this to avoid changing too much.

PaulWessel:
This is also a time to reflect if **-c**_row_,_col_ should really be **-c**_row_/_col_ which is more typical GMT usage, no?

seisman:
`gmt subplot set` use the `row,col` style:
```
gmt subplot set [ row,col ] [ -Afixedlabel] [ -Cside/clearance[u] ] [ -V[level] ]
```
It seems a little weird to have `gmt subplot set 1/3`.

seisman:
After merging #785, there're still some issues.

Please check output of following commands:
- `gmt basemap -`:  unwanted ` ]` in -c option
- `gmt psbasemap -`: unwanted `[]` 
- `gmt sac -` and `gmt pssac -`: both print -O, -P, -K, -c options.

PaulWessel:
A bit mystified by the gmt sac - case.  Running it in Xcode I come to the printing of the usage message and all the GMT_P_OPT variables show as being blank "" but yet the print statement produces -P, etc.  This seems to be specific to all PS producing modules in the supplements.  The core works fine, so something to do with extern char * settings.  I think I need to allocate these on the stack.  Will try that instead.

PaulWessel:
OK, figured it out (had to re-learn how extern and declarations work...).  PR coming up soon.

Issue #777, opened 2019-05-26
gmt coast should print usage in modern mode


seisman:
In classic mode, typing `gmt pscoast` will print the module usage.

In modern mode, `gmt coast` will report some errors:
```
$ gmt coast                                                         
pscoast [ERROR]: Syntax error: Must specify -R option
pscoast [ERROR]: Syntax error: Must specify at least one of -C, -G, -S, -I, -N, -Q and -W
```
The same goes for other modules for modern mode, and for `gmt meca`, it crashes.

Is it possible that `gmt coast` can print the full usage?

PaulWessel:
There are some issues with this. For instance, gmt plot by itself may be a valid modern mode command if **-R -J** has already been set and the data are piped in.  We would only be able to check that by detecting a session has started and **-R -J** has been set.  Even then, we do not know the intention of the user.  However, I think "gmt plot" may be the only "valid" command and I would be OK insisting the user add **-W**_pen_ to get what they want here, otherwise it should be "show me usage".
I will have to see how this may be implemented.

PaulWessel:
I propose that all plot modules in _modern_ mode must supply an explicit option. Currently, I think only two modules could be used in a script where **-R -J** is set explicitly by earlier modules and you could do things like

cat my data | gmt clip
cat my_points | gmt plot

Since the modules are unaware of the stdin stream we would have to allow

gmt clip
gmt plot

to wait for input rather than print the usage message.  This seems too far out for me, hence the suggestion of requiring an option (or file) to run the module.  While for plot there is no problem (give **-W**_pen_ instead of relying on a default), with clip there is no option to activate clipping.  There is **-C**[**a**|_n_] to deactivate but the default is to activate.  Hence, to get cat my_data | gmt clip to take an option we need to invent one.  My suggestion would be **-Cb** for begin clipping, and let that be the default.  Let me know what @GenericMappingTools/core think.

PaulWessel:
Note that this would also apply to gmt info under modern mode.  cat file | gmt info currently works in classic but if run in modern mode an option would be required, such as **-Id** for default behavior.

joa-quim:
Confusing. I see no meaning in asking for the module's usage in the middle of modern session. Since a ``gmt begin`` has already been issued, user has only to know what he wants to do. Also, obliging to give 
needless options seems to go contrary to the simplification route.

[...]

PR #785, opened 2019-05-27
Add call to GMT_Option for -c in all plotting modules

PaulWessel:
In response to #784.  Well, except inset which cannot start a new panel.  Still subject to #777.  Meant to only be listed under modern mode.  Still need to do synopsis.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -15644,7 +15644,7 @@ int gmtlib_get_num_processors() {
 }
 
 int gmt_report_usage (struct GMTAPI_CTRL *API, struct GMT_OPTION *options, unsigned int special, int (*usage)(struct GMTAPI_CTRL *, int)) {
-	/* Handle the way classic and modern mode modules report their usage msssages.
+	/* Handle the way classic and modern mode modules report their usage messages.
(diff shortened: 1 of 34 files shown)


seisman:
The rST documentations should also be updated.

PaulWessel:
Yes.  I have to head out for a few hours - if you have time please do it and let me know you are working on it so we don't both do it.

seisman:
OK. I'll open another PR to address the issue of documentations.

seisman:
I tried 
```
gmt coast -
```
and
```
gmt begin map
gmt coast -
gmt end
```
both don't show -c option.

PaulWessel:
Yes, will have to debug - now that I am back.

PaulWessel:
Funny, pscoast was left of my list as I created the list by searching for KEYS with >X} but pscoast.c is a chameleon that may write data...  Now works for me.  So gmt coast - or gmt coast -? works and shows the **-c** and explanation, while gmt pscoast - or -? gives classic usage with **-O -K** but no **-c**. Typing gmt coast with no args give errors about **-R** etc as before.  Please see gmt_report_usage in gmt_init.c for the explanation.

PR #801, opened 2019-05-29
Use ctest instead of make check to avoid the stalled build issue on Travis CI

seisman:
Although `cmake --build . --target check` is equivalent to `ctest --force-new-ctest-process -j4`, the former one causes troubles on Travis, as it wont't output anything unless all tests are completed.


seisman:
> -j4? Aren't Travis running machine single core?

Travis are running on dual-core machine. From the build output, it looks like travis can run four tests parallelly.

joa-quim:
So, maybe just **-j** which runs with all available cores, no?

seisman:
No. Unlike `cmake -j`,  `ctest -j` uses a single core.

joa-quim:
OK

PR #804, opened 2019-05-30
Fix some deprecated syntax in examples

seisman:
Please see the commit messages for deprecation warnings of old scripts.

The PS files in these examples need to be re-generated. Someone using US defaults please help with it.


seisman:
> What is the basis for changing the 3.5i to 3.8i in ex38? I could not figure it out

Check the figure of ex38 at http://docs.generic-mapping-tools.org/latest/gallery/ex38.html. Note the slight overlapping between left and right panels.

image

seisman:
@PaulWessel Don't forget to re-generate the PS files.

PR #806, opened 2019-05-30
Let -Tmin/max/inc behave like array creation in MATLAB, etc.

PaulWessel:
The new **-T** parser in places like makecpt or gmtmath is pedantic and will complain if (_max_-_min_) is not an integer times _inc_.  However, in MATLAB and other langulages, creating an array like x = 0:2:11 will result in 0, 2, 3, 4, 10 with no fussing whatsoever.  I have changed the parser to only warn about this (requires **-V**).  But perhaps it should not even be a warning (i.e., require **-Vl**)?


PaulWessel:
Are you OK with a warning if you do -T0/9.5/1? You don't get one in MATLAB if you do 0:1:9.5.

joa-quim:
No need for warning. Just docu it and it's done.

PR #809, opened 2019-05-30
Update site address of documentations

seisman:


seisman:
Currently, latest is just a symbol link to dev. After 6.0.0 is released, latest will point to the 6.0.0 documentations, while dev will point to the documentations of the master branch.

I'm OK with both. I always would like to see the real "latest" documentation from master.

PR #815, opened 2019-06-01
Allow different CPT defaults if using earth|srtm_relief files

PaulWessel:
The GMT CPT default is rainbow but it is not suitable for DEMs since there is no clean separation at sealevel.  This PR introduces the notion of several GMT CPT defaults.  For now there are the default rainbow, the geo for earth_relief, and the srtm for srtm_relief (no ocean).  All tests that passed before still pass.


PaulWessel:
Yes, I was not sure where.  Right now it would only affect grdimage and grdview, so could go in their man pages.  Alternatively, I could add something to the GMT_Docs section on CPT files. Or can do both. Other suggestions?

PR #818, opened 2019-06-02
Allow setting GDAL_DATA and PROJ_LIB via environment variables.

joa-quim:
With the advent of GDAL3+PROJ4 V4 there might be conflicts between the different versions of files that PROJ4 needs to find out, for example, what a EPSG number is (@seisman  felt it already). On the other hand, in Windows I want to ship a full working GDAL+PROJ4 and for that we need a collection of auxiliary files pointed by GDAL's GDAL_DATA and PROJ's PROJ_LIB environmental variables.

This PR lets GMT users set them via the ``LOCAL_GDAL_DATA`` and ``LOCAL_PROJ_LIB`` (better names may be proposed) env variables. Furthermore, if all those files are copied into GMT's ``share/GDAL_DATA`` directory (like the Windows installer will do) then that dir will take preference to the GDAL's & PROJ's pointed ones.

Well, this is the intention. Only tested on Windows (Winbuntu is several years away of having a GDAL3) and it worked.

Reminder, this is a GDAL3 only thing (#ifdefed with GDAL_VERSION_MAJOR >= 3)

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -14663,6 +14663,10 @@ struct GMT_CTRL *gmt_begin (struct GMTAPI_CTRL *API, const char *session, unsign
 
 	struct GMT_CTRL *GMT = NULL;
 	char version[GMT_LEN8] = {""};
+#if defined(HAVE_GDAL) && (GDAL_VERSION_MAJOR >= 3)
+	char *path1 = NULL, *path2 = NULL, *paths[2] = {NULL, NULL};
+	int  local_count = 0;
+#endif
 
 #ifdef __FreeBSD__
 #ifdef _i386_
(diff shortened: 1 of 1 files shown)


joa-quim:
I have one doubt about line 14717

paths[0] = strdup(dir);

I’m not freeing that paths[0], so it leaks. But the other in cases paths[0|1] contains the string return bu getenv and those we should not free.

PaulWessel:
Are you done with paths after the call to the GDAL function? if so, can't you just use a bool variable that is set to true when you strdup and then you free paths[0] if that variable is true?

joa-quim:
Quite likely, freeing it the call to ``OSRSetPROJSearchPathsfunction`` would screw (crash?) everything because the paths would disappear. Unless GDAL makes a copy of the input arg.

PaulWessel:
It would be bad design if it did not make its own copy based on the input string pointer.  Is there any documentation on this?

joa-quim:
https://gdal.org/api/ogr_srs_api.html?highlight=osrsetprojsearchpaths#_CPPv421OSRSetPROJSearchPathsPPCKc

and an obscure ``auto`` C++ type (well, everything in C++ is obscure) in this (quite helpful) test

https://github.com/OSGeo/gdal/blob/master/autotest/cpp/test_osr_set_proj_search_paths.cpp

joa-quim:
Tried this variation that frees the strdup allocated mem and it seems to keep working.

PaulWessel:
Good, I would be surprised if a function call result would depend on arguments to it that change later.

Issue #499, opened 2019-03-12
MAP_ANNOT_OBLIQUE 2 fails with MAP_FRAME_TYPE inside


KristofKoch:
**Description of the problem**

When using

```
gmt set MAP_FRAME_TYPE inside
gmt set MAP_ANNOT_OBLIQUE 2
```

no annotation is plotted along the upper and lower plot borders. I expected longitude annotations parallel to the plot border placed on the inside of the border.

Using 

```
gmt set MAP_FRAME_TYPE plain
gmt set MAP_ANNOT_OBLIQUE 2
```

everything works as expected with parallel annotations on the upper and lower border of the plot.

I expect the parallel annotations parallel to the plot border regardless of `MAP_FRAME_TYPE`. Combinations of bit 2 in `MAP_ANNOT_OBLIQUE` don't work either. 

**Full script that generated the error**

This test script generates two identical plots except for `MAP_FRAME_TYPE` to highlight the problem.

```
#!/bin/bash

FILE1=test_map_annot_plain.ps
FILE2=test_map_annot_inside.ps

GMT_R="-R-125.1/31.6/-108.18/41.14r"
GMT_J="-JL-117.3/36.7/32.6/40/16c"

#
# ANNOTATIONS OUTSIDE (WORKS)
#

# get rid of anything that might influence the plot
rm gmt.conf
rm gmt.history

gmt set MAP_FRAME_TYPE plain
gmt set MAP_ANNOT_OBLIQUE 2

gmt psbasemap \
	${GMT_R} ${GMT_J} \
	-BWESN \
	-Bpa5f1g5 \
	> $FILE1

#
# ANNOTATIONS INSIDE (FAILS)
#

# get rid of anything that might influence the plot
rm gmt.conf
rm gmt.history

gmt set MAP_FRAME_TYPE inside
gmt set MAP_ANNOT_OBLIQUE 2

gmt psbasemap \
	${GMT_R} ${GMT_J} \
	-BWESN \
	-Bpa5f1g5 \
	> $FILE2
```

**System information**

* Operating system: 18.04.1 LTS (GNU/Linux 4.15.0 x86_64)
* Version of GMT: Version 6.0.0_68cf5ae_2019.03.08 [64-bit] [MP] [4 cores]

joa-quim:
This (no ``MAP_ANNOT_OBLIQUE``) gives me what you expect
```
[...]

PR #846, opened 2019-06-06
Fix missing longitude annotation for frame-type inside

PaulWessel:
The labels were there in the PostScript but due to a bug they got placed outside the frame and clipped.  See #499 for context. I hope @KristofKoch can give this a test before we merge.


seisman:
I tested it with the example script in #499. It works better now, but still imperfect. 

The annotation at the lower left corner is still clipped.

image

PaulWessel:
Well, there are more than one way to look at this.  The string is simply clipped, which is OK.  However, in an ideal world we would not place that number since it will be clipped.  To detect that is not trivial, and could only be done correctly via PostScript coding, i.e., detect that the LB point of string is outside and thus not plot.  That means knowing string width and subtract half width from plot (centered) point and determine it is negative.   I am not prepared to put in that type of work for this.  The next simplest thing to do is approximate: Guesstimate the half-width of that string in GMT (i.e., in guessed inches) and do the same assessment there.  I recommend we accept this PR and perhaps leave some comment in an issue about this not being perfect yet.

joa-quim:
Yes, agree (and forget my dumb comment, confused the function names)

PR #855, opened 2019-06-07
Fix CMake CMP0026

### LINKED ISSUE #682: Update CMakeLists.txt to avoid deprecation warnings
**CMake complaints**

When we run cmake we get a bunch of messages that I would like to go away:

> CMake Deprecation Warning at CMakeLists.txt:61 (cmake_policy):
>   The OLD behavior for policy CMP0026 will be removed from a future version
>   of CMake.
> 
>   The cmake-policies(7) manual explains that the OLD behaviors of all
>   policies are deprecated and that a policy should be set to OLD only under
>   specific short-term circumstances.  Projects should be ported to the NEW
>   behavior and not rely on setting a policy to OLD.

and this one

> CMake Warning (dev) at src/CMakeLists.txt:255 (find_package):
>   Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
>   Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
>   command to set the policy and suppress this warning.
> 

and

> CMake Warning (dev):
>   Policy CMP0058 is not set: Ninja requires custom command byproducts to be
>   explicit.  Run "cmake --help-policy CMP0058" for policy details.  Use the
>   cmake_policy command to set the policy and suppress this warning.
> 

Does anyone out there have suggestions on how to get rid of these by updating our scripts to not be causing these warnings?

**Are you willing to help implement and maintain this feature?** Yes, but confused.

seisman:
[CMP0026](https://cmake.org/cmake/help/latest/policy/CMP0026.html) (introduced in CMake 3.0) disallows use of the LOCATION property for build targets, because the LOCATION property is not fully determined until generate-time.

This means we cannot get LOCATION property of the libraries.  

THis PR sets CMP0026 policy to new (i.e. disallows use of the LOCATION property). The filenames of the libraries are now determined by OUTPUT_NAME and PREFIX in the target property, and also CMAKE_SHARED_LIBRARY_SUFFIX, CMAKE_STATIC_LIBRARY_SUFFIX, CMAKE_SHARED_MODULE_SUFFIX settings.
 
I tried it with cmake 2.8.12 on an old Linux. It works well.

Closes #682.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,20 +46,12 @@ cmake_minimum_required (VERSION 2.8.7)
 # Use NEW behavior with newer CMake releases
 foreach(p
 		CMP0025 # CMake 3.0: Compiler id for Apple Clang is now AppleClang
+		CMP0026 # CMake 3.0: Disallow use of the LOCATION target property
 		CMP0058 # CMake 3.3: Ninja requires custom command byproducts to be explicit
 		CMP0074	# CMake 3.12: find_package uses PackageName_ROOT variables
 		)
 	if(POLICY ${p})
 		cmake_policy(SET ${p} NEW)
-	endif()
-endforeach()
-
-# Use OLD behavior with newer CMake releases
-foreach(p
-		CMP0026 # CMake 3.0: Disallow use of the LOCATION target property
-		)
-	if(POLICY ${p})
-		cmake_policy(SET ${p} OLD)
 	endif()
 endforeach()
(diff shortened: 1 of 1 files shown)


seisman:
GMT_CORE_LIB_NAME, GMT_CORE_LIB_NAME and GMT_SUPPL_LIB_NAME are also used in other source codes. This PR will make troubles.

PR #858, opened 2019-06-07
Simplify test scripts gmtest.in and their CMakeLists.txt files

PaulWessel:
I have cleaned up the testing schemes (CMakeLists.txt (3), gmtest.sh (3), and animate.in):

- Eliminate special provisions for classic versus modernize checks in the three gmtest.in scripts. We no longer plan (or have been doing) a conversion from classic to modern on the fly.
- Use a simple grep-command to detect if a script is modern mode (look for "gmt begin") and use that to handle the setting of the session PPID.
-  Flag all known failure scripts with the keyword **GMT_KNOWN_FAILURE**.  Any script with that text in it will not be compared with an original PostScript file (which often is missing) and instead will FAIL (known failure) and not be added to the failure count.
- Eliminate **MODERNIZE_TESTS** from the make/Config*.cmake files.

The only desired modification would be a report of the count of known failures at the end, so we are at least reminded when running the tests.  However, the log file lists all the information

### DIFF
--- a/cmake/ConfigUserTemplate.cmake
+++ b/cmake/ConfigUserTemplate.cmake
@@ -216,8 +216,6 @@
 #set (DO_EXAMPLES TRUE)
 #set (DO_TESTS TRUE)
 #set (DO_ANIMATIONS TRUE)
-# Auto-convert classic scripts to modern during testing:
-#set (MODERNIZE_TESTS TRUE)
 # Number of parallel test jobs with "make check":
 #set (N_TEST_JOBS 4)
 
--- a/doc/examples/CMakeLists.txt
+++ b/doc/examples/CMakeLists.txt
@@ -153,12 +153,6 @@ if (UNIX AND DO_ANIMATIONS)
 endif (UNIX AND DO_ANIMATIONS)
 
 # run examples (test)
-if (MODERNIZE_TESTS)
-	set (SCRIPT_MODE "M")
-else (MODERNIZE_TESTS)
-#set (SCRIPT_MODE "C")
-	set (SCRIPT_MODE "")
-endif (MODERNIZE_TESTS)
 
 if (DO_EXAMPLES) # AND UNIX)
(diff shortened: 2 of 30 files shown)


joa-quim:
Would be good to have a way to exclude the movie and API tests on Windows. As they are they cannot work (movie runs bash scripts and I was never able to make that run on Win. API uses gcc).

PaulWessel:
Well, the cmake Template has this
#set (DO_ANIMATIONS TRUE)
...
#set (DO_API_TESTS ON)

so perhaps you have turned them on in your ConfigUser.cmake file?

seisman:
I still have some failures:
```
	331 - grdfilter/openmp.sh (Failed)
	493 - pscoast/pseudo.sh (Failed)
	517 - pslegend/stacklegendmod.sh (Failed)
	579 - psxy/antwrap.sh (Failed)
	613 - psxy/decoratedlines.sh (Failed)
	669 - psxy/trimline.sh (Failed)
	750 - mgd77/test_cm4.sh (Failed)
```

And this is what I got in the test log `build/test/fail_count.d`. I expect to see a full test log information.
```
2019-06-07 09:23:29 pscoast/pseudo.ps: RMS Error = 0.128
2019-06-07 09:23:33 pslegend/stacklegendmod.ps: RMS Error = NA
2019-06-07 09:24:10 psxy/trimline.ps: RMS Error = 0.018
```

PaulWessel:
I have not flagged those as known failures since I hope to fix some of them.  Also, openmp.sh is not a failure since it was not run but currently reports FAIL.  I don't know anything about fail_count.d; I have only looked at Testing/Temporary/LastTest.log.

joa-quim:
>Well, the cmake Template has this
>#set (DO_ANIMATIONS TRUE)
>...
>#set (DO_API_TESTS ON)

I have them set to FALSE but still show up as failures

PR #859, opened 2019-06-07
Ignore zmin/zmax in -R if -Jz/Z is not given

PaulWessel:
Deescalate from fatal error to warning.


PaulWessel:
OK, perhaps you can look at that since you are on the hunt for flatters.

PR #881, opened 2019-06-11
Give more control regarding MAP_FRAME_TYPE=graph

PaulWessel:
This PR allows users to change the default parameters for the length of the extension needed to add a vector at the end of the axis under **MAP_FRAME_TYPE**=_graph_ mode [7.5%].  You can append ,_length_[_unit_], where _unit_ may be **%** (extension to tip of vector is this fraction of the axis length) or give an absolute length (with units **c**, **i**, **p**).  Then we extend the axis by that amount before adding the vector head.


joa-quim:
Nice. But we should still address the case like my Julia example where I was forced to call basemap twice in order to have the tick marks on the inside side. What about a new **MAP_FRAME_TYPE=graph+inside[,length[unit]]**

PR #893, opened 2019-06-13
Add a template for GMT release checklist

seisman:
Created an issue template for GMT release checklist. For each upcoming release, we can open an issue using this template, and tick them off one by one.


seisman:
Feel free to add more to the checklist

PaulWessel:
Do we tag after we build the tarballs? With svn we always made the 5.4.x tag and committed, then made the tarball etc so it would have the right settings.

seisman:
I don't think tagging or not makes any differences in git. A tag is just a pointer to a specific commit hash. 

In the checklist, I put tagging after making tarballs, because sometimes our tarballs/installers may not work as expected (e.g. #909). If we tag and push it to GitHub, then realize that our tarballs don't work well, we have to make a new tag (i.e. increasing GMT version number).

seisman:
Need updates after we finish #962.

leouieda:
The only way the order would make a difference is if GMT used the git tag for something. But since the version number is set by cmake then it shouldn't matter.

seisman:
I updated the release checklist based on our release experience with 6.0.0rc1. This PR is ready for review.

PR #894, opened 2019-06-13
Fix segyprogs to segy

seisman:

The official name of segy supplement is `segy`. Sometimes we use segyprogs. This PR fixes segyprogs to segy.


joa-quim:
Should we merge ``segy`` into the new supp ``seis``?

seisman:
I prefer to keeping them separated.

PR #909, opened 2019-06-14
List file to add extra exe's to the Win installer

joa-quim:
On Windows the installer ships with a complete GDAL + PROJ4 (V6). But for that I need to provide the list of the exe's to include in the installer. This file has that, but for my own paths.

### DIFF
--- a/cmake/modules/FindFFTW3.cmake
+++ b/cmake/modules/FindFFTW3.cmake
@@ -70,6 +70,11 @@ if (FFTW3F_LIBRARY)
 	check_symbol_exists (fftwf_import_wisdom_from_filename fftw3.h FFTW3F_VERSION_RECENT)
 	cmake_pop_check_state() # restore state of CMAKE_REQUIRED_*
 
+	# For some reason in my machine check_symbol_exists() call above is not able to its job
+	if (WIN32 AND NOT FFTW3F_VERSION_RECENT)
+		set (FFTW3F_VERSION_RECENT TRUE)
+	endif ()
+
 	# old version warning
 	if (NOT FFTW3F_VERSION_RECENT)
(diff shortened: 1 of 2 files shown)


cmake/modules/FindFFTW3.cmake
```diff
@@ -70,6 +70,11 @@ if (FFTW3F_LIBRARY)
 	check_symbol_exists (fftwf_import_wisdom_from_filename fftw3.h FFTW3F_VERSION_RECENT)
 	cmake_pop_check_state() # restore state of CMAKE_REQUIRED_*
 
+	# For some reason in my machine check_symbol_exists() call above is not able to its job
```
seisman: I don't think it's a good idea to commit your WIN32 trick here, since the Windows CI doesn't have such problem.

joa-quim: Hmm, that change slipped in but I actually wanted to submit it as an separate PR. Yes, you are right that Windows CI does not seem to have that problem. But I do, and since it's me who makes the installers, I need to have it working preferably to always have an uncommitted change in my local repo (which the tyrant Git refuses to update)

PaulWessel: Any way to work around having to commit this into the FindFFTW.cmake file, i.e., via some conditional include statement or similar?

joa-quim: I already wrapped it under a ``if (WIN32 AND NOT FFTW3F_VERSION_RECENT)``. According to the comment in line 69, this would only break for the case of someone building on windows with a static FFTW3 build. A case that as far as we know, never happened before, nor is likely to happen in a foreseeable future.

seisman:
These changes were merged in #922.

PR #911, opened 2019-06-14
Make sure char strings for paths are PATH_MAX long

### LINKED ISSUE #910: Long TMPDIR causes troubles
I'm using macOS 10.14.5. On my mac, TMPDIR is defined as:
```
$ echo $TMPDIR
/var/folders/mz/d3wtwjtx4fg8qlds5lgt1c240000gn/T/
```
I never define TMPDIR myself, so I believe this is the default value set by macOS. The long TMPDIR will cause some segmental faults.

Below is an example:

In `gmtlib_get_region_from_data`, the variable `tmpfile` is only 64-bytes.

https://github.com/GenericMappingTools/gmt/blob/127bb214ce9012334c17ec915fe84804a61ade12/src/gmt_init.c#L11777

Then the `strcat` function aborts with a segment fault, because the temporary file path needs 70 bytes.
https://github.com/GenericMappingTools/gmt/blob/127bb214ce9012334c17ec915fe84804a61ade12/src/gmt_init.c#L11811

To reproduce the issue:
```
export TMPDIR /a/very/long/path/to/somewhere
gmt begin plot-test
gmt plot -JX10c -Baf << EOF
0 0
5 5
EOF
gmt end
```

Changing GMT_LEN64 to GMT_LEN128 can fix the issue in gmtlib_get_region_from_data. Other functions may have the same issue.

PaulWessel:
We had som that were only 64 and got buffer overruns.  Closes #910.

### DIFF
--- a/src/blockmean.c
+++ b/src/blockmean.c
@@ -558,7 +558,7 @@ int GMT_blockmean (void *V_API, int mode, void *args) {
 	}
 	if (Ctrl->G.active) {	/* Writes the grid(s) */
 		unsigned int kk;
-		char file[GMT_LEN128] = {""};
+		char file[PATH_MAX] = {""};
 		for (k = kk = 0; k < NF; k++) {
 			if (strstr (Ctrl->G.file[kk], "%s"))
 				sprintf (file, Ctrl->G.file[kk], code[k]);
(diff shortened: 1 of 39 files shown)


joa-quim:
Hope we won~t have consequences of this. Quite often there are strncmp(...) that pass the size of the stack allocated sizes and hence need to be updated as well.

PR #936, opened 2019-06-15
WIP: Add nightly packaging to CI

seisman:

This PR makes source packages and installer nightly, to test if changing CMakeLists.txt breaks packaging.

- Linux: Make source packages only
- macOS: Make both source packages and Bundles
- Windows: Make both source packages and binary installers 

This also make it possible to distribute nightly built installers for macOS and Windows.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -161,7 +161,7 @@ jobs:
 
   variables:
     BUILD_DOCS: false
-    TEST: false
+    TEST: true
 
   steps:
 
@@ -192,7 +192,7 @@ jobs:
   - bash: |
       set -x -e
       cd build
-      cmake --build . --target gmt_release
-      cmake --build . --target gmt_release_tar
+      #cmake --build . --target gmt_release
+      #cmake --build . --target gmt_release_tar
       cmake --build . --target package
     displayName: Package GMT
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -115,6 +115,40 @@ jobs:
     condition: succeededOrFailed()
     displayName: Upload test coverage
 
+# Mac - Pacakge
+########################################################################################
+- job:
+  displayName: 'Mac | Package'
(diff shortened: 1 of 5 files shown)


joa-quim:
Cool. But I'm sure we'll still have to tune the Win installer section. For example, in my ``ConfigUser.cmake`` I have
```
	# CPACK doesn't copy the omp dll probably becuase it uses the loadlibrary dynamic linking tech
	if (BITAGE EQUAL 64)
		if (VC STREQUAL "VC12")
			set (OMP_DLL_LOC "C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/redist/x64/Microsoft.VC120.OPENMP/vcomp120.dll")
		else ()
			set (OMP_DLL_LOC "C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Redist/MSVC/14.16.27012/x64/Microsoft.VC141.OPENMP/vcomp140.dll")
		endif ()
	else (BITAGE EQUAL 64)
		if (VC STREQUAL "VC12")
			set (OMP_DLL_LOC "C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/redist/x86/Microsoft.VC120.OPENMP/vcomp120.dll")
		else ()
			set (OMP_DLL_LOC "C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Redist/MSVC/14.16.27012/x86/Microsoft.VC141.OPENMP/vcomp140.dll")
		endif ()
	endif (BITAGE EQUAL 64)
```
as well as other sections to copy the GDAL_DATA and PROJ_LIB directories so that GDAL will work without users even knowing that we set up the ENV vars under the hood. But this is for GDAL3 and Vcpkg is still on ``GDAL2.4.1``.

Ah, and there is also the issue of the ``gmtmex`` that can't be built by the CI script. If there is a way of it fetching the mex file (a dll) from somewhere we could think of a solution for this.

seisman:
Yes, we definitely need tune the cmake files if we want a nightly release. We can do it after the beta release.

PR #948, opened 2019-06-16
WIP: Enable tests on Windows CI

seisman:


seisman:
@joa-quim How long does it take to run all the tests on Windows? It seems it will take several hours here.

seisman:
Attached is the test log file on Windows CI: [33.txt](https://github.com/GenericMappingTools/gmt/files/3295324/33.txt).

Bad news: it takes 5 hours to run 200 tests.

Good news: most tests pass, while some tests fail due to the missing `bc` command and something else.

PaulWessel:
Wow, I guess this is serious time-slicing?

joa-quim:
It takes 273 sec with -j8 and 949 sec with ``ninja check``. But it only runs 207 tests. What about the others?

And worst. The testing system is broken. It doesn't write the ``test`` directory with the failures. Now we only know it failed but cannot go see why.

joa-quim:
Hmm, deleted the ``build`` dir and restarted and now things are back to normal. With -j8 it takes 980 sec

PR #954, opened 2019-06-16
Make macOS bundle on macOS CI

seisman:
Let macOS CI build source tarballs and macOS bundle (dmg file) nightly, to avoid any changes breaking the packaging process.

Now we have a nightly macOS bundle. After confirming that the bundle works for everyone, we can upload it somewhere.


seisman:
@PaulWessel This PR is ready for review.

PR #950, opened 2019-06-16
Add debug printout for -Aa in gmtspatial

PaulWessel:
Under #ifdef DEBUG/endif, added **-Vd** print statements to see what the sorted pairs look like.  No effect on non-debug enabled builds. Needed to try to figure out why macOS and Linux give different answers for test gmtspatial/poldecimate.sh.

PR #956, opened 2019-06-16
Use mergesort in gmtspatial NN so results are stable

PaulWessel:
We found that the nearest neighbor analysis gave slightly different results on macOS and Linux, and traced it to the fact that _qsort_ is an unstable algorithm (if given identical items then the order is undefined).  Thus, we switch to _mergesort_ (which is stable) in this application.  Exception is (for now) Windows which appears not to have a built-in mergesort function.  See #950 for context.

### DIFF
--- a/src/gmt_config.h.in
+++ b/src/gmt_config.h.in
@@ -128,6 +128,7 @@
 #cmakedefine HAVE_POPEN
 #cmakedefine HAVE__POPEN
 #cmakedefine HAVE_POSIX_MEMALIGN 1
+#cmakedefine HAVE_MERGESORT
 #cmakedefine HAVE_QSORT_R_GLIBC
 #cmakedefine HAVE_SETLOCALE
 #cmakedefine HAVE__SETMODE
--- a/src/gmtspatial.c
+++ b/src/gmtspatial.c
@@ -146,6 +146,18 @@ struct PAIR {
 	uint64_t pos;
 };
 
+#ifdef __APPLE__
+/* macOX has it built in, so ensure we define this flag */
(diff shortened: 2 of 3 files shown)


seisman:
The linux build fails.

PaulWessel:
Yep, working on including a mergesort.c.  Could you check if it is hiding in some basic place? It seems incredible it is not available as I google mergesort Linux and find man pages for it...

joa-quim:
[Here ](https://www.bogotobogo.com/Algorithms/mergesort.php)they have a multi-threaded  (pthread) mergesort.

PaulWessel:
But only sorting integers and nowhere near the prototype with generic pointer args we need.

seisman:
mergesort is provided by bsd system. For ubuntu, it's available from libbsd http://manpages.ubuntu.com/manpages/trusty/man3/heapsort.3.html

PaulWessel:
Would it be too much to add dependency on libbed for the non-BSD linux flavors that dont have mergesort natively?

PaulWessel:
@seisman, since it now passed, could you try build and run that poldecimate test and send the plot?

seisman:
@PaulWessel Please see the attached PS file [poldecimate.ps.txt](https://github.com/GenericMappingTools/gmt/files/3294869/poldecimate.ps.txt).

You need remove the .txt suffix.

PaulWessel:
Thanks, same number as for macOS so I will merge this PR.  @joa-quim can comment out lines 154 and 159 in gmtspatial.c and see if Windows can compile and run the same way.

PR #970, opened 2019-06-17
Enable GMT_USE_THREADS and GMT_ENABLE_OPENMP

seisman:


PaulWessel:
One issue with enabling threads/open-mp when running tests that I have found is that if I use my usual 24 cores and then some of the jobs use openmp that also takes over logs of cores, things may bog down.  Not sure how this will work on CI.

seisman:
OK. I can try openmp with the full tests.

seisman:
@PaulWessel I added these two lines to ConfigUser.cmake
```
set (GMT_USE_THREADS TRUE)
set (GMT_ENABLE_OPENMP TRUE)
```
but the openMP support is still disabled on macOS. Any idea how to enable openmp on macOS?

PaulWessel:
Yes, I think so.  Cannot use Apple clang sine no OpenMP support yet (really, Apple!). but install gcc-mp-x.  I then have a file in my home dir called cache.cmake which has stuff like this:

> SET ( CMAKE_C_COMPILER "/opt/local/bin/gcc-mp-9" CACHE STRING "GNU MP C compiler" )
> SET ( CMAKE_CXX_COMPILER "/opt/local/bin/g++-mp-9" CACHE STRING "GNU MP C++ compiler" )
> SET (CMAKE_C_FLAGS -flax-vector-conversions)
> SET (CMAKE_C_FLAGS_DEBUG -flax-vector-conversions)
> SET (CMAKE_C_FLAGS_RELEASE -flax-vector-conversions)
> SET (OpenMP_C_FLAGS -flax-vector-conversions)

and I pass -C ~/cache.cmake to cmake.

seisman:
@PaulWessel This PR enables threads and openmp support on Linux, and threads support on macOS. 
Currently, we don't have a simple solution to enable openmp support on macOS.

All the tests pass. I think it's ready to merge.

PR #983, opened 2019-06-19
Fix sharedir path

PaulWessel:
Work in progrress on this.


seisman:
@leouieda Can you check if this commit works in conda builds.

leouieda:
Running at https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=43123 I'll post here when it finishes building

leouieda:
OK, all checks pass on all 3 OSs. This is good to go on the conda side.

seisman:
This PR is OK to me. The only tiny issue is that the bundle will still try to search Paul's path first. Of course, Paul's path doesn't exist in users' computer, then the bundle skip it to the correct path.

PaulWessel:
Maybe we can make this better after rc1 is done.

leouieda:
:+1: I'm in favor of that

seisman:
Yes, I agree. I also don't think my PR is good enough.

PR #984, opened 2019-06-19
WIP: Fix GSHHGDIR and DCWDIR

seisman:
@PaulWessel Can you check if this works for you.

### DIFF
--- a/share/CMakeLists.txt
+++ b/share/CMakeLists.txt
@@ -34,12 +34,12 @@ if (GSHHG_PATH AND COPY_GSHHG)
 		DESTINATION ${GMT_DATADIR}/coast
 		COMPONENT GSHHG)
 	# location must be blank in gmt.conf
-	set (GSHHG_INSTALL_PATH)
+	set (GSHHG_INSTALL_PATH CACHE INTERNAL "Location of GSHHG Data")
 	# create coastline.conf which is needed for in-build-dir ctest
 	configure_file (coastline.conf.in coastline.conf @ONLY)
 else (GSHHG_PATH AND COPY_GSHHG)
 	# set installed location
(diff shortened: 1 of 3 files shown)


PaulWessel:
OK trying now.

seisman:
In this function, GMT tries to find DCW in GMT->session.DCWDIR (if set) first, then search via gmt_getsharepath.
https://github.com/GenericMappingTools/gmt/blob/f75038ab617363e358a1555b8d54b61ff6e37660/src/gmt_dcw.c#L71-L77

When COPY_DCW is set to TRUE, we know that DCW data must be in SHAREDIR/dcw, so we should skip searching in GMT->session.DCWDIR.

Now the problem is,  GMT->session.DCWDIR is always set to `DCW_INSTALL_PATH`:
https://github.com/GenericMappingTools/gmt/blob/f75038ab617363e358a1555b8d54b61ff6e37660/src/gmt_init.c#L5473-L5475

And `DCW_INSTALL_PATH` is always set to `DCW_PATH` in https://github.com/GenericMappingTools/gmt/blob/f75038ab617363e358a1555b8d54b61ff6e37660/src/config.h.in#L68-L69.

PaulWessel:
Your first change worked for me. Do I need to try ff7f807 as well?

seisman:
The first commit breaks CI, because we don't set "COPY_GSHHG" and "COPY_DCW" in our CI.

PaulWessel:
We still need this merged too, no?

seisman:
No. #983 already works. This PR tries to make it better.

I think we can leave it after rc1.

joa-quim:
Ping on this. Implement it (after conflict solving)? Looks OK to me.

PR #991, opened 2019-06-20
Set COPY_GSHHG and COPY_DCW to TRUE for packaging standalone installers

seisman:
Need copy gshhg and dcw data to make macOS bundle and Windows installers.


leouieda:
@seisman I'm not sure what this PR does. Is this setting up for deploying packages from CI in the future?

seisman:
I'm exploring the possibility of making nightly releases.

To make standalone installers, we have to set COPY_GSHHG and COPY_DCW to TRUE so that these coastline data are included in the installers.

Issue #995, opened 2019-06-24
Error with begin -Vd


seisman:
**Description of the problem**

Script to reproduce the bug:
```
gmt begin map -Vd
gmt coast -Rg -JH10c -Baf -W1p
gmt end
```

Debug information:
```
$ sh test.sh
gmt [DEBUG]: Obtained the ppid from parent: 93093
gmt [DEBUG]: Enter: gmtinit_new_GMT_ctrl
gmt [DEBUG]: GMT->session.SHAREDIR = /Users/seisman/local/GMT-6.0.0/share
gmt [DEBUG]: GMT->session.HOMEDIR = /Users/seisman
gmt [DEBUG]: GMT->session.USERDIR = /Users/seisman/.gmt [created]
gmt [DEBUG]: GMT->session.CACHEDIR = /Users/seisman/.gmt/cache [created]
gmt [DEBUG]: GMT: 0. Will try to find subdir=postscriptlight stem = PSL_custom_fonts suffix=.txt
gmt [DEBUG]: GMT: 1. gmt_getsharepath trying current dir
gmt [DEBUG]: GMT: 2. gmt_getsharepath trying USERDIR /Users/seisman/.gmt
gmt [DEBUG]: GMT: 3. gmt_getsharepath trying USERDIR subdir /Users/seisman/.gmt/postscriptlight
gmt [DEBUG]: GMT: 4. gmt_getsharepath trying SHAREDIR subdir /Users/seisman/local/GMT-6.0.0/share/postscriptlight
gmt [DEBUG]: GMT: 5. gmt_getsharepath failed
gmt [DEBUG]: Map distance calculation will be Cartesian
gmt [DEBUG]: Exit:  gmtinit_new_GMT_ctrl
gmt [DEBUG]: Enter: New_PSL_Ctrl
gmt [DEBUG]: Exit:  New_PSL_Ctrl
gmt [DEBUG]: Enter: gmt_manage_workflow
gmt [DEBUG]: Exit : gmt_manage_workflow
gmt [DEBUG]: Enter: PSL_beginsession
gmt [DEBUG]: Exit : PSL_beginsession
gmt [DEBUG]: Enter: PSL_setdefaults
gmt [DEBUG]: Exit : PSL_setdefaults
gmt [DEBUG]: Enter: gmtlib_io_init
gmt [DEBUG]: Exit : gmtlib_io_init
gmt [DEBUG]: Enter: gmt_hash_init
[...]

PR #996, opened 2019-06-24
Properly extract session and graphics format for begin

PaulWessel:
The session name and graphics format could get confused if **-Vd** was given.  Closes issue #995.

### DIFF
--- a/src/begin.c
+++ b/src/begin.c
@@ -19,7 +19,7 @@
  * Version:	6 API
  *
  * Brief synopsis: gmt begin starts a modern mode session.
- *	gmt begin [<prefix>] [<format>]
+ *	gmt begin [<prefix>] [<format>] [-V<level>]
  */
 
 #include "gmt_dev.h"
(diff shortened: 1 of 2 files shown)


joa-quim:
I think we must create the 6.0 branch first before proceed with the normal bug fixes flow.
I was expecting to come home to bring this issue up again.

PaulWessel:
Ok, perhaps you guys can sort that out; I won’t be able to do more for a
few days.

SOEST, U of Hawaii at Manoa

seisman:
I've created the 6.0 branch. Now it's even with the master. 

As this PR fixes a bug in 6.0.0rc1, we should merge it into 6.0 branch and then merge the 6.0 branch into master. Is this what you meant @joa-quim

joa-quim:
Yes. Except that the order is unclear to me (though at the end it should give the same). If we merge to master then we should rebase the 6.0 branch.

seisman:
> If we merge to master then we should rebase the 6.0 branch.

The order matters if we have added some new features to the master branch. Rebasing is not a good idea for long-term branches, as it always rewrites some git history.

seisman:
I just changed the base branch of this PR to 6.0.

joa-quim:
Yes, you are right. I guess I was thinking on this case only.

PaulWessel:
So do we also from time to time merge 6.0 into master then, since it needs the same fix.

joa-quim:
It's like  we used to do in SVN  Everything in 6.0 should be also in master, but not the converse.

PaulWessel:
Thanks, just merged 6.0 into master (one commit).

PR #998, opened 2019-06-25
Fix error in gmt2kml -Nt

### LINKED ISSUE #997: gmt2kml appears to fail when using trailing record text as label
When using gmt2kml in gmt6, I couldn’t use trailing record text as label.

Using gmt 5.4.5, I get what I expected, showed below,
but with gmt 6, I didn’t get the expected label, but a generic “Point 0”.

I tried using -Ncol, but appears it didn’t work too.

I hope I understood properly the usage for option -N:
1. Specify -Ncol to use a column from the data record a single-word label (-Fe|s|t only).
2. Specify -Nt if the trailing record text should be used as label (-Fe|s|t only).

```
gmtswitch 5.4.5
echo "-45 -30 1000 sjsp" | gmt 2kml -Ag -Fs -N+ | grep name
                        <name>stdin</name>
                                <name>Point Set 0</name>
                                        <name>sjsp</name>
gmtswitch 6.0.0
echo "-45 -30 1000 sjsp" | gmt 2kml -Ag -Fs -Nt | grep name
                        <name>stdin</name>
                                <name>Point Set 0</name>
                                        <name>Point 0</name>

```

**Full error message**

```
PASTE ERROR MESSAGE HERE
```

**System information**

* Operating system:Ubuntu 18.04.2 LTS
* Version of GMT: 6.0.0_279da86_2019.06.23

PaulWessel:
The if test checking for this particular option had a bug (checked the wrong array item) and was never accepted.  This PR closes #997.

### DIFF
--- a/src/gmt2kml.c
+++ b/src/gmt2kml.c
@@ -443,9 +443,8 @@ GMT_LOCAL int parse (struct GMT_CTRL *GMT, struct GMT2KML_CTRL *Ctrl, struct GMT
 					Ctrl->N.mode = GET_COL_LABEL;
 					Ctrl->N.col = GMT_Z;
 				}
-				else if ((opt->arg[0] == 't' || opt->arg[0] == '+') && opt->arg[0] == '\0') {	/* Trailing text (+ is backwards compatible)  */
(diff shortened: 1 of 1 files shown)


seisman:
Merge this PR?

PaulWessel:
Sorry, travel made me forget.

PR #1001, opened 2019-06-25
Add GMT_KNOWN_FAILURE flag to "pass" the known failures

### LINKED ISSUE #852: Allow per-test RMS threshold for plot comparisons
**Ability to override default RMS setting given to GraphicsMagic**

The testing of our ~800 GMT scripts that make plots involve comparison between a saved original PostScript file and the one produced by the script.  We call _GraphicsMagick_ to make rasters of both and then compare them and compute a RMS difference.  If this difference exceeds the given threshold (set in CMake config as **GRAPHICSMAGICK_RMS**) the test fails, otherwise it passes.  Unfortunately, subtle differences between compilers and different OS yield slightly different plots even though there are no bugs.  For some plots these differences add up more than for others, resulting in false positives (i.e., failures that are not real failures).  These false positives makes it harder to detect new failures introduced by code changes.  We would like the ability to give an optional **GRAPHICSMAGICK_RMS** in some scripts and have that value used in the comparison.

Here is a potential solution to improve this situation:

1. Add a script-specific **GRAPHICSMAGICK_RMS** setting for false positive scripts to avoid seeing the flagged as failures
2. Add a flag **KNOWN_FAILURE** to scripts that we know fail so that they do not count towards the total failure when CI runs, since this masks any new failures
3. Modify gmtest.in to act on these optional variables accordingly.
4. Make an issue per known failure and write what we think the problem and potential solution may be so that contributors can have a look and possible fix the problem.

seisman:
Fixes #852.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -50,7 +50,7 @@ jobs:
 ########################################################################################
 - job:
   displayName: 'Mac | Test'
-  condition: eq(variables['Build.Reason'], 'Schedule')
+  #condition: eq(variables['Build.Reason'], 'Schedule')
 
   pool:
     vmImage: 'macOS-10.14'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -50,7 +50,7 @@ jobs:
 ########################################################################################
 - job:
   displayName: 'Mac | Test'
-  #condition: eq(variables['Build.Reason'], 'Schedule')
+  condition: eq(variables['Build.Reason'], 'Schedule')
 
   pool:
     vmImage: 'macOS-10.14'
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ matrix:
           if: type != cron
         - name: "Linux (cron - test)"
           os: linux
-          if: type != cron
+          if: type = cron
           env:
             - TEST="true"
         # Only build the docs on Linux cron jobs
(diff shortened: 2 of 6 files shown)


seisman:
`mgd77/test_cm4.sh` and `psxy/antwrap.sh` still fail.

joa-quim:
I still don't understand well this idea of hiding the failing tests. What then are they good for?

seisman:
> I still don't understand well this idea of hiding the failing tests. What then are they good for?

We run all the tests nightly to make sure recent changes don't break the codes. Usually, we have two ways to know if all the tests pass:

1. Check the badges in README.md [TravisCI](https://travis-ci.org/GenericMappingTools/gmt) [Azure](https://dev.azure.com/GenericMappingTools/GMT/_build/latest?definitionId=2&branchName=master)

2. Check the build logs.

It's obvious that the first one is much simpler. However, it doesn't work well for our case. Since we always have some "known to fail" tests, we  have to read build logs to check if there're any new "failing" tests. After hiding these "known to fail" tests, we can know if all the tests pass at a glance.

seisman:
There is still one failing test (mapproject/proj4.sh) on Linux. Probably it's caused by the old versions of gdal (1.11.3) or proj (4.9.2).

joa-quim:
I see. But the price of this is to almost throw away the failing tests. We should move them into a separate category (directory?) to not hide them under the carpet.

joa-quim:
The ``mapproject/proj4.sh`` failures are boring. They mostly happen due to changes in PROJ4.

PaulWessel:
We need a way to run the filling test so they fail without having to find and edit individual scripts. but we can discuss this at Scripps.

seisman:
Can we merge this? Or leave it to the summit?

PaulWessel:
Either way, we will discuss this later this summer so perhaps nice to have things pass for now?  I won't forget about the failures!

seisman:
OK. merged.

PR #1008, opened 2019-06-27
Let gmt docs open plot files without prepending file://

### LINKED ISSUE #1007: gmt docs cannot open local plots
If I make a local png, such as

`gmt coast -R0/30/0/30 -JM5i -B -Gred -png map
`

then decide to view it via gmt docs, it fails:

> gmt docs map.png
> The file /map.png does not exist.
> docs [ERROR]: Opening file:///map.png in the default browser via open failed with error 256
> docs: No such file or directory

Perhaps the file:///map.png is required with Windows "cmd /c start" or Linux "xdg-open, but it sfor sure does not work with macOS "open".  One solution is to do the #ifdef __APPLE__ to handle this by just passing the file name, but first please tell me if it works for @joa-quim on Windows and @seisman on Linux?  Also, the error message talks about browser but when it is an image of PDF it is opened by Preview, so we should improve our messaging as well.

PaulWessel:
Creating URLs starting with file:// should only be done for local documentation (HTML) pages.  This PR closes #1007.

### DIFF
--- a/src/docs.c
+++ b/src/docs.c
@@ -132,7 +132,7 @@ int GMT_docs (void *V_API, int mode, void *args) {
 			snprintf (URL, PATH_MAX, "%s", docname);
 	}
 	else {	/* One of the fixed doc files */
-		sbprintf (URL, PATH_MAX, "file:///%s/doc/html/%s", API->GMT->session.SHAREDIR, module);
+		snprintf (URL, PATH_MAX, "file:///%s/doc/html/%s", API->GMT->session.SHAREDIR, module);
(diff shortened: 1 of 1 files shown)


seisman:
Don't forget to improve the message:
```
Opening file:///map.png in the default browser via open failed with error 256
```

PaulWessel:
I removed the reference to the browser.

PaulWessel:
If you are still up, please approve so I can merge this in with 6.0 and master, and then add my master PR for gmt end show.

PaulWessel:
Meant to ping @seisman on this.  Going to store...

PR #1010, opened 2019-06-27
Let gmt end take show verb to display all illustrations via gmt docs

PaulWessel:
**Add optional verb show to gmt end**

When writing a GMT modern script that makes one or more figures, it is nice to be able to automatically open the finished PDFs or PNGs in viewers.  Instead of manually adding commands to do that, I have added an optional keyword show to the gmt end command.  As an example,  this script

```
gmt begin
  gmt figure map1 png
  gmt coast -R0/30/0/30 -JM5i -Gred -B
  gmt figure map2 pdf,PNG
  gmt coast -R0/30/0/30 -JM5i -Ggreen -B
gmt end show
```

will open up the previewer for three figures.

I am open to suggestions here, e.g., perhaps show should be an option instead of a verb, but I was guided by others like gmt subplot begin|end.


seisman:
It looks pretty good.

PaulWessel:
I should have had this idea some weeks ago.  It seems like a really handy thing to have in GMT 6 from the get-go.  We will soon be doing workshops and will wish we could end scripts with "gmt end show".  What do @seisman, @joa-quim, and @leouieda think about merging this into master and release rc2 next week before we send homework to the UNAVCO students about GMT installation?  Remember, gmt docs file is broken in rc1.

seisman:
I also would like to see it in GMT 6.0.0, rather than waiting for the 6.1.0 release.

If we're going to add the feature before 6.0.0rc2, then we should merge it into 6.0 branch, then merge 6.0 into master.

joa-quim:
> I should have had this idea some weeks ago. 

Actually I had this idea sometime ago and you were a bit resistant to it at the time.
Yes, I agree to include this on future 6.0.0

leouieda:
:+1: I'm all for it

seisman:
As we all agree, I just changed the base branch to 6.0, in case we forget it when merging.

PR #1013, opened 2019-06-27
Use examples in centimeters instead of inch

joa-quim:


seisman:
@joa-quim I think you shouldn't push commits directly to the 6.0 branch. Instead, you should create a new branch based on the 6.0 branch, then open a PR and select 6.0 as the base branch.

PaulWessel:
Agreed. Took me a few tries to get it.

SOEST, U of Hawaii at Manoa

joa-quim:
OK, I think I see a way to do it from within V. Studio Code.

PR #1016, opened 2019-06-27
Place gridlines behind symbols like in psxyz

### LINKED ISSUE #1015: Grid mesh on top for 2D and under for 3D
Paul, we discussed this a while ago and ended up forgetting it. The way the grid mesh is set in not consistent. We plot it on top of symbols for ``psxy`` and under them with ``psxyz``. This is well depicted by looking at the ``headercheck.sh`` test (in both psxy & psxyz test directories). The 3D case looks nicer. 

Changing this behavior is a bit of a break, but as is it's not consistent and the time to do it is (still) really now.

PaulWessel:
This old "feature" is deemed a bug and we now will draw the gridlines before laying down symbols, lines, and polygons.  This will trigger test failures and require updating a bunch of PostScript originals next.  Closes #1015.

### DIFF
--- a/src/psxy.c
+++ b/src/psxy.c
@@ -993,11 +993,12 @@ int GMT_psxy (void *V_API, int mode, void *args) {
 		gmt_plotend (GMT);
 		Return (GMT_NOERROR);
 	}
+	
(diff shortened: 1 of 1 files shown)


PaulWessel:
OK, wil merge in and then merge into master as well.

PR #1036, opened 2019-06-29
Replace apt-get with apt

seisman:
Apt is recommended since Ubuntu 16.04.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,8 +81,8 @@ before_install:
         echo "Copy cached deb packages"
         sudo rsync -av ${DEB_CACHE}/ /var/cache/apt/archives/
         echo "Start to install software"
-        sudo apt-get update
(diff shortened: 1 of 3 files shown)


seisman:
I also cherry-picked it into master branch.

holishing:
I wonder that if it will show in build log of CI:

```
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

for e.g.
Line 313 in build log in [Travis CI](https://travis-ci.org/GenericMappingTools/gmt/jobs/552220181#L313)

Although it doesn't matter in my experience with testing in Docker automated build before.

seisman:
You're right. apt is designed as a end-user tool, not suitable for scripts (https://askubuntu.com/questions/990823/). 

Maybe we should revert this commit.

PR #1042, opened 2019-06-29
Let first plot under subplot default to first panel if not set via -c

PaulWessel:
When a subplot is started, it is convenient to have a default for which panel to first plot in.  That default will be the first panel, of course, and now we select that if the first plot command in a subplot does not set **-c** explicitly.

### DIFF
--- a/doc/rst/source/explain_-c_full.rst_
+++ b/doc/rst/source/explain_-c_full.rst_
@@ -3,4 +3,6 @@
 **-c**\ [*row*\ ,\ *col*]
     Used to advance to the selected subplot panel.  Only allowed when in
     subplot mode.  Available to all plot modules.  If no arguments are
-    given then we advance to the next panel in the selected order.
+    given then we advance to the next panel in the selected order. If
+    no **-c** is given and we just entered subplot mode then the first
+    panel is selected.
--- a/doc/rst/source/explain_-c_full.rst_
+++ b/doc/rst/source/explain_-c_full.rst_
@@ -5,4 +5,4 @@
     subplot mode.  Available to all plot modules.  If no arguments are
     given then we advance to the next panel in the selected order. If
     no **-c** is given and we just entered subplot mode then the first
-    panel is selected.
+    panel (top, left) is selected.
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -11528,6 +11528,7 @@ GMT_LOCAL int get_current_panel (struct GMTAPI_CTRL *API, int fig, unsigned int 
 		API->error = GMT_RUNTIME_ERROR;
 		return GMT_RUNTIME_ERROR;
 	}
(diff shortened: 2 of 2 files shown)


joa-quim:
So I guess now this test in subplot.c is no longer needed
```
			case 'C':
				if (Ctrl->In.mode != SUBPLOT_SET) {
					GMT_Report (GMT->parent, GMT_MSG_NORMAL, "Option -C is only available when setting the active panel\n");
					n_errors++;
				}
```

PaulWessel:
No, I dont think so. Why? You need gmt subplot set to adjust those issues on a per-panel case, if I remember this correctly.

joa-quim:
I guess that I'm confused by the error message
```
"Option -C is only available when setting the active panel"
```
and the active panel is not set by default to **-c1,1**, but that occurs later to the call to ``subplot``, so I'm not getting what this test does.

PaulWessel:
It means you cannot use -C when you are doing
gmt subplot begin
gmt subplot end

it can only be used with
gmt subplot set

SOEST, U of Hawaii at Manoa


On June 30, 2019 at 9:02:23 AM, Joaquim (notifications@github.com) wrote:

I guess that I'm confused by the error message

"Option -C is only available when setting the active panel"

and the active panel is not set by default to *-c1,1*, but that occurs
later to the call to subplot, so I'm not getting what this test does.

—
You are receiving this because you modified the open/close state.
<https://github.com/GenericMappingTools/gmt/pull/1042?email_source=notifications&email_token=AGJ7IX5OJVLFXND25L4UQKDP5D7L7A5CNFSM4H4LFVXKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODY4RXCI#issuecomment-507059081>,
or mute the thread

joa-quim:
Ah, I saw the ``set`` case in code, but found nothing about it in docs

PaulWessel:
Look again, there are three sections in the subplot rst: begin, set and end.

joa-quim:
It's getting more confuse. So we can't do ``gmt subplot begin -C`` (begin lists -C) unless we have done ``gmt subplot set`` which also has **-C**?

PaulWessel:
Yes, I agree this is odd.  The way it works is that gmt subplot begin **-C** sets clearances for all panels equally.  If you then wish to mess with the clearance for a particular panel then you must use gmt subplot set **-C**.  The only place I see we are doing **-C** is in ex02.sh in the exmod tests so I dont have a case where we use it with begin.  We can remove that restriction and see what happens...

joa-quim:
Hmm, that example doesn't use ``set``, just ``gmt subplot 2,1 -Ce1.1i`` Does it mean that in lack of ``begin|end`` it defaults to ``set``?

PaulWessel:
Yes.

PR #1049, opened 2019-06-30
VS isdigit in debug mode has an assert C>=0 && c<= 255. 

joa-quim:
Do that here otherwise it crashes in debug mode.
I still another one like this in ``struct_geo.sh`` but these crashes are damn hard and time consuming to find.

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -8233,7 +8233,8 @@ bool gmt_not_numeric (struct GMT_CTRL *GMT, char *text) {
 	if (!text) return (true);		/* NULL pointer */
 	if (!strlen (text)) return (true);	/* Blank string */
 	if (isalpha ((int)text[0])) return (true);	/* Numbers cannot start with letters */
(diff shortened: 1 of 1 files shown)


joa-quim:
Possibly that char is not ascii so we are pushing the situation. Why (int)t[0] is giving a negative number?

PaulWessel:
What is the text that is being checked? Has Euro accented characters?

joa-quim:
From utf8.sh. This line
```
echo 0 1 ø ù ú û ü ý þ ÿ | gmt pstext -R-1/6/0.6/9.6 -Jx0.9i -P -B0  -F+f24p+jBL > lixo.ps
```

seisman:
This PR should be merged into 6.0, right?

joa-quim:
right. Doing to many things at same time.

seisman:
OK. I'll cherry-pick the commit to 6.0 branch.

PR #1051, opened 2019-06-30
Enhance trailing text selection by specifying specific word

PaulWessel:
This PR adds an optional word selection to the trailing text selection **t** used by **-i** and **-o**.  By specifying **t**_word_ we select a single word from the trailing text [Default is the entire trailing text].  Here, _word_ is the number of the word, starting at 0 for the first word.

Some examples of behavior:

```
echo 1 2 3 A B C | gmt convert -i2,1,t1
3	2	B
echo 1 2 3 A B C | gmt convert -o0,t2 
1	C
echo 1 2 3 A B C | gmt convert -ot0   
A
```

There may be more work to do before this works in pstext which as its own peculiar input checking, hence the WIP.  However, in doing this I ran into an issue that requires discussion:

In GMT we long ago decided that the first column is 0.  The **-i** and **-o** options and many other places follow that scheme.  However, in updating the GMT documentation for the word enhancement I found this passage pertaining to the extraction of a single word from the trailing text for use as a label in a custom symbol:

> Normally, the **l** macro code will place a hard-wired text string.  However, you can also obtain the entire string from your input file via a single symbol variable **$t** that must be declared with type **s** (string).  The string will be taken as all trialing text in your data record.  To select a single word from the trailing text you just use $k, where k starts at 1 for the first word, regardless of how many numerical columns that precede it.

So in this particular setting we use word column starting at 1.  In contrast, gmt movie also handles individual words in trailing text:

> ..., while **t**_col_ uses word number *col* from the trailing text in *timefile* (requires **-T** ... **+w**; first word is 0).

It would seem the custom symbol word counting system violates the "start at 0" GMT rule in operation everywhere else and I propose we flag this as a bug that needs to be fixed.  I ping @KristofKoch here since he was the one arguing for the word feature, but the implementation error is mine.

Finally (and again), this enhancement seems super-helpful (avoids reliance on awk to deal with parts of trailing text) and should have been included in 6.0.0, hence I would suggest we merge this into 6.0.0 and release as rc2.


seisman:
I'm OK with merging this feature to 6.0 branch.

PaulWessel:
Given we now have merged #1064, most of my discussion above is moot. What remains is simply the addition of trailing text word control in **-i** and **-o**, plus a different solution for pstext where the **+w**_word_ modifier to **-F** acts the same way.

PaulWessel:
How about you, @joa-quim? OK to merge this guy? FYI, since pstext doesn't use **-i** I added -**F+w**, i.e.,

`echo 0 0 CM This is Bullshit people | gmt pstext -R-2/2/-2/2 -JX4i -P -B0 -F+f24p+w2+j > t.ps
`

But as I write this, with pstext not taking **-i** I could add **-i** as a **module** option to pstext and get the word that way.  This would let me remove the **+w**_word_ scheme and implement it via a local _-i_ option.  Perhaps better?

seisman:
I like `pstext -i`. It's more intuitive.

joa-quim:
Sure (confess I've not been following this issue closely. Just too many things)

PaulWessel:
OK, I will work on a local -i.

PaulWessel:
Thanks for the feedback, it is now **-it**_word_:

`echo 0 0 CM This is Bullshit people | gmt pstext -R-2/2/-2/2 -JX4i -B0 -F+f24p+j -it2 -pdf map
`

joa-quim:
You mean, we have now a *-i* which is not the global **-i**?

PaulWessel:
We have a **-i** in pstext that is not the global **-i** since that one deals with the numerical and text portion of a record.  pstext, going back to the Reagan administration, has a weird i/o where you can list the order of things like font, justification, angle, so if you want

> x y justification angle text

then you are violating the GMT record format by putting text (justification) ahead of another numerical column (angle).  Unless you are willing to say that all numerical items, including angle, must come before all text items, then we cannot use the global **-i**.  This is why I first implemented **-F+w**_word_ but it is also possible to have a local **-i** act in the same way, except it is only limited to the word selection, hence **-it**_word_.

joa-quim:
Ok, it's fine for me  (I'm not the one who puts consistency above all)

PaulWessel:
There is also the possibility down the road that I figure out a way to let pstext use the actual global **-i** instead, and then we don't have to change the syntax again.  OK, need your review.

cmake/ConfigDefault.cmake
```diff
@@ -39,7 +39,7 @@ set (GMT_VERSION_YEAR "2019")
 
 # The GMT package version.
 set (GMT_PACKAGE_VERSION_MAJOR 6)
-set (GMT_PACKAGE_VERSION_MINOR 0)
+set (GMT_PACKAGE_VERSION_MINOR 1)
```
joa-quim: This should not be here. The other time was to merge in master. This one goes to 6.0

PR #1056, opened 2019-06-30
Failed to get the remote file tag when file extension was given

### LINKED ISSUE #1050: Print information when downloading earth_relief data

https://github.com/GenericMappingTools/gmt/blob/9924521a0381e7308aaada6f611ee1f2f9489496/src/gmt_remote.c#L71-L75

This function says it will print some information if the @earth_relief_xxx.grd is downloaded for the first time.

I removed all earth_relief files in my `.gmt` directories, and ran: 
```
$ gmt which -Gu @earth_relief_20m
/Users/seisman/.gmt/server/earth_relief_20m.grd
```
It didn't show any information.

Again, removed all earth_relief files and ran:
```
$ gmt which -Gu @earth_relief_10m @earth_relief_20m
/Users/seisman/.gmt/server/earth_relief_10m.grd
earth_relief_20m: Download file from the GMT ftp data server [data set size is 783K].
earth_relief_20m: Earth Relief at 20x20 arc minutes obtained by Gaussian spherical filtering (37 km fullwidth) of ETOPO1 (Ice g) [NOAA].

/Users/seisman/.gmt/server/earth_relief_20m.grd
```
It showed the information for earth_relief_20m, but not earth_relief_10m.

I expect to see the information each time the earth_relief files are downloaded.

PaulWessel:
If a user specified a remove file including the extension (@earth_relief_10m.grd) then the attribution code did not determine the 10m tag but instead got grd.  Now check that if .grd is given then we adjust accordingly.  This PR closes #1050.  In general, the extension is not required.

### DIFF
--- a/src/gmt_remote.c
+++ b/src/gmt_remote.c
@@ -72,7 +72,8 @@ GMT_LOCAL int give_data_attribution (struct GMT_CTRL *GMT, const char *file) {
 	/* Print attribution when the @earth_relief_xxx.grd file is downloaded for the first time */
 	char tag[4] = {""};
 	int k, match = -1, len = (int)strlen(file);
-	if (strstr (file, ".grd")) len -= 4;    /* If extension was provided we must skip that as well */
+	if (strstr (file, ".grd")) len -= 4;		/* If .grd extension was provided we must skip that as well */
+	else if (strstr (file, ".nc")) len -= 3;	/* If .nc extension was provided we must skip that as well */
 	strncpy (tag, &file[len-3], 3U);	/* Get the xxy part of the file */
 	for (k = 0; k < GMT_N_DATA_INFO_ITEMS; k++) {
 		if (!strncmp (tag, gmt_data_info[k].tag, 3U)) {	/* Found the matching information */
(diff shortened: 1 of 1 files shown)


src/gmt_remote.c
```diff
@@ -72,9 +72,11 @@ GMT_LOCAL int give_data_attribution (struct GMT_CTRL *GMT, const char *file) {
 	/* Print attribution when the @earth_relief_xxx.grd file is downloaded for the first time */
 	char tag[4] = {""};
 	int k, match = -1, len = (int)strlen(file);
-	strncpy (tag, &file[len-3], 3U);
+	if (strstr (file, ".grd")) len -= 4;		/* If .grd extension was provided we must skip that as well */
+	else if (strstr (file, ".nc")) len -= 3;	/* If .nc extension was provided we must skip that as well */
```
seisman: Does this mean I can use `gmt which -Gu @earth_relief_20m.nc`?

It gives some errors:
```
$ gmt which -Gu @earth_relief_20m.nc
gmtwhich [ERROR]: Libcurl Error: HTTP response code said error
gmtwhich [ERROR]: File earth_relief_20m.nc.grd not found!
```

PaulWessel: No, I was being tired and stupid.  I just removed that.  The files are *.grd but users should only specify @earth_relief_xxy with no extension.

PR #1057, opened 2019-06-30
Handle very long KML coordinate records

### LINKED ISSUE #1012: kml2gmt failing

Simply want to make a quick conversion from kml to GMT. Using the command pasted below.

Yields only the initial header record, then nothing else (no coordinates). File attached. Is it a file issue or a read-in issue?

```
kml2gmt IS2_RGT_0002_cycle4_28-Jun-2019.kml -Fl
```

**Full error message**

```
None - even with -Vd. Just outputs the header, then nothing else.
```

**System information**

* Operating system: OSX 10.13.6
* Version of GMT: Version 6.1.0_908ab21_2019.06.27 [64-bit] [8 cores]

PaulWessel:
Deals with the issue pointed to by, and closes #1012.

Some applications write the coordinates in a KML line or polygon on a single record, regardless of how many.  This meant kml2gmt failed since it was limited by GMT_BUFSIZ.  I extended that buffer to allocate GMT_INITIAL_MEM_ROW_ALLOC characters instead ( ~2 milion characters).

### DIFF
--- a/src/kml2gmt.c
+++ b/src/kml2gmt.c
@@ -166,9 +166,9 @@ int GMT_kml2gmt (void *V_API, int mode, void *args) {
 	size_t length;
 	bool scan = true, first = true, skip, single = false, extended = false;
 	
-	char line[GMT_BUFSIZ] = {""}, buffer[GMT_BUFSIZ] = {""}, name[GMT_BUFSIZ] = {""};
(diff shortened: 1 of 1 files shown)


seisman:
@PaulWessel  Please use `closes #1012` instead of `closes issue #1012`.

PaulWessel:
A few more times and I will get it right (maybe).

Issue #1062, opened 2019-06-30
Custom symbol selection of word for label uses wrong count


PaulWessel:
**Description of the problem**

We recently added the capability of selecting a single word from a record's trailing text to be used in a custom symbol via the psxy **-Sl**_size_**+t**_word_ mechanism.  The syntax that I foolishly selected was this: **$t**:_word_, where **$t** represents trailing text and the optional :_word_ selects a single word from **$t**, using the integer _word_ to pick the item.  There are two flaws with this scheme:

1. There should not be any colon needed, as **$t**3 is easy to parse, and that matches the syntax in **-i** and **-o**.
2. The first word should be 0, not 1 (as implemented) and this breaks compatibility with how all other GMT options count – they all start at 0.

The fix is to declare this a bug and use **$t**[_word_], with _word_ starting at 0 for first word.  I believe the only real user of this feature is @KristofKoch and he can handle this correction.  let me know if there are comments, otherwise I will build a PR.

KristofKoch:
@PaulWessel, no problem - I will adjust my scripts accordingly. Thank you for the heads up!

PaulWessel:
Great, we have merged this PR so just replace **$t**:_n_ with **$t**n-1 after you build the latest.

PR #1064, opened 2019-06-30
Correct how trailing text words are accessed in custom symbols

PaulWessel:

See #1062 for context.  This PR reinstates the rule that 0 is the first item in a list (here, the words in the trailing text).  It also eliminates the colon between **$t** and the _word_ number since this syntax clashes with how **-i** and **-o** specify single words.  The only place we use colon is in ranges, e.g., **-i**1:5 (or **-i**1-5).


PaulWessel:
FYI, there is no backwards compatibility issues here since this feature was added in 6 only.

PR #1067, opened 2019-06-30
Let cmake report the gs version

PaulWessel:

It would be helpful if the cmake configuration could not just say "yes" when it finds a required item but also report the version number.  I added this for gs since gs --version prints that number. Now, cmake says

> *  Found GhostScript (gs)     : yes (9.27)

I would like the same for gm, ffmpeg but those guys think it is normal to print out everything they know about their builds, so it will require things like 

`ffmpeg -version | awk '{if (NR == 1) print $3}'
`

which I don't quite know how to do in cmake.


seisman:
We already know that gs 9.27 has serious bugs. Do you think we should warn users not to use this version?

PaulWessel:
Yes, we probably should.  This is very annoying.  I wonder how other developers test their code.  In just a few weeks we have seen deadly bugs introduced in netcdf, gs, and gm...
Perhaps these steps?

1. Make sure we build the macOS bundle with gs 9.2x < 27
2. Make sure the Windows installer that @joa-quim builds have 9.2x < 27
3. Add information to our prerequisites page that mentions these problems.

Come to think of it, wasn't 9.27 meant to fix another bug we pointed out in 9.26?  Do we need to go back further?

joa-quim:
Yes, 9.27 was supposed to fix the issue that prevented setting the 15 cm padding of our wall sized page.

We don't ship gs in the Win installer, but should we? I do it in Mirone and surprisingly only the exe and dll do the conversion job nicelly. No need of any other file (fonts, etc).

PaulWessel:
Not sure, up to you I guess.  Actually, we don't distribute - we had that discussion with Artifex some years ago and cannot do it.

joa-quim:
I remember that discussion but not the arguments anymore. GS is GPL, why the hell can't we redistribute.

joa-quim:
This PR broke the Win build (there is no ``gs`` exe on Windows). To fix it I ended up committing directly to 6.0 & master (endless git shits). Now that I've looked more on this one, I don't think is so interesting to give that info compared to the extra complexity it introduces in the CMakeList file.

PaulWessel:
Maybe just do this under *nix.

Issue #1035, opened 2019-06-29
Can't find custom symbols on Windows


seisman:
**Description of the problem**

When running the tests on Windows CI (#948), I see some scripts failing with errors like:
```
686/839 Test #689: psxy/custom_symbol.sh ...................***Failed    3.69 sec
psxy [ERROR]: Could not find either custom symbol or EPS macro D:A:/1/s/test/psxy/comet
psxy [ERROR]: Could not find custom symbol D:A:/1/s/test/psxy/comet
ERROR: gmtest:58
memtrack errors: 0
exit status: 1
ERROR: custom_symbol.sh:7
memtrack errors: 0
exit status: 1
```

The output of `gmt defaults -Vd` is:
```
gmt [DEBUG]: Obtained the ppid from parent: 5344
gmt [DEBUG]: Enter: gmtinit_new_GMT_ctrl
gmt [DEBUG]: GMT->session.SHAREDIR = D:/a/1/s/gmt-install-dir/share
gmt [DEBUG]: GMT->session.HOMEDIR = D:/a/1/s
gmt [DEBUG]: GMT->session.USERDIR = D:/a/1/s/.gmt [created]
gmt [DEBUG]: GMT->session.CACHEDIR = D:/a/1/s/.gmt/cache [created]
```

These directories all look good. But the path to search custom symbols is obviously wrong. Any idea how it happens? 

Refer to the [build log](https://dev.azure.com/GenericMappingTools/GMT/_build/results?buildId=1105) if you want to see more output.

PaulWessel:
I suspect gmtinit_handle_dosfile in gmt_init.c. It temporarily replaces : with ; and then later back again in order to parse -B properly.  But don't see how it is called in the custom symbol context.  Not sure how D:/a/1 can become D:A:/1 otherwise.  There has to be some wrong logic somewhere in the code (gmt_init.c, gmt_io.c) where we have an unstated assumption of the dir name after X: being longer than 1.

PaulWessel:
Is this another manifestation of minGW replacing /a with A: inside a path so that D:/a/ becomes D:A:/?

PaulWessel:
From what you and @joa-quim described with minGW replacing stuff, here we are basically issuing this GMT call:

`gmt psxy -R0/20/10/30 -JM6i -P -Bag -W0.5p,red -SkD:/a/1/s/test/psxy/comet/2.5i -K -Xc -Yc > $ps
`

[...]

PR #1069, opened 2019-07-01
Try to defeat minGW insanity on path replacements

PaulWessel:
With minGW replacing /a with A: we must undo this craziness before trying to open files.  See #1035 for context.


seisman:
Closed as it won't be merged.

PR #1072, opened 2019-07-01
Fix CID 199098

joa-quim:


joa-quim:
Shit. I forgot to change it again to merge in 6.0. This is really to boring.

PaulWessel:
You are actually slower on the uptake that me! That makes me happy.
Just do the same to 6.0 then.

joa-quim:
Fck. Again, I create a branch on 6.0 which is useless because it insists in merging on master.
I'll do it on 6.0 from the Github page.

PR #1074, opened 2019-07-02
Add admin dir with files needed for building a rc or final release

PaulWessel:
We place convenience scripts and CMake settings in the new admin dir.  They are not part of the tar distribution but used by the developers when building the tar balls and macOS bundle for the releases.

### DIFF
--- a/admin/build-release.sh
+++ b/admin/build-release.sh
@@ -25,7 +25,7 @@ cp -f admin/ConfigReleaseBuild.cmake cmake/ConfigUser.cmake
 rm -rf build
 mkdir build
 cd build
-cmake -G Ninja -DGMT_PACKAGE_VERSION_SUFFIX=${tag} -DGMT_PUBLIC_RELEASE=TRUE ..
+cmake -G Ninja ..
 # 3. Build the release and the tar balls
 cmake --build . --target gmt_release
 cmake --build . --target gmt_release_tar
--- a/admin/build-release.sh
+++ b/admin/build-release.sh
@@ -35,11 +35,9 @@ rm -f gmt-${Version}${tag}-src.tar
 cmake --build . --target install
 # 6. Build the macOS Bundle
 cpack -G Bundle
-# 7. Rename bundle to match standard package prefix
-mv GMT-${Version}${tag}-Darwin.dmg gmt-${Version}${tag}-darwin-x86_64.dmg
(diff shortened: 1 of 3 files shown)


admin/build-release.sh
```diff
@@ -0,0 +1,48 @@
...
+# 1. Set basic ConfigUser.cmake file for a release build
+if [ -f cmake/ConfigUser.cmake ]; then
+	cp cmake/ConfigUser.cmake cmake/ConfigUser.cmake.orig
+fi
+cp -f admin/ConfigReleaseBuild.cmake cmake/ConfigUser.cmake
+# 2. Make build dir and configure it
+rm -rf build
+mkdir build
+cd build
+cmake -G Ninja -DGMT_PACKAGE_VERSION_SUFFIX=${tag} -DGMT_PUBLIC_RELEASE=TRUE ..
```
seisman: We need set GMT_PACKAGE_VERSION_SUFFIX and GMT_PUBLIC_RELEASE in the `cmake/ConfigDefault.cmake` file and commit the changes before making a release. So we don't need set them in the command line.

seisman:
The `admin` directory should be removed before making a source tarball. You need change this file:

https://github.com/GenericMappingTools/gmt/blob/0e8a8cd8533dd24f15f14393ccd433e523fa7560/CMakeLists.txt#L107-L112

admin/build-release.sh
```diff
@@ -0,0 +1,48 @@
...
+# 3. Build the release and the tar balls
+cmake --build . --target gmt_release
+cmake --build . --target gmt_release_tar
+# 4. Remove the uncompressed tar ball
+rm -f gmt-${Version}${tag}-src.tar
+# 5. Install executables before building macOS Bundle
+cmake --build . --target install
+# 6. Build the macOS Bundle
+cpack -G Bundle
+# 7. Rename bundle to match standard package prefix
```
seisman: @PaulWessel Please review #1078. If #1078 is merged, you need remove the step 7.

PaulWessel:
OK, step 7 removed.

seisman:
@PaulWessel Please also see other comments.

admin/build-release.sh
PaulWessel: OK, will remove those too.

PR #1077, opened 2019-07-02
WIP: Remove unneeded cpack generators

seisman:

Running `cmake --build . --target package` can package GMT in different formats. We only need NSIS on Windows and Bundle on macOS. Other formats like ZIP, TGZ, TBZ2 are not needed.


joa-quim:
Actually the ZIP format can be interesting since what it produces is just zip files instead of an installer. This has it's own value. Either for people who don't like installers or may face permissions problems in machines they do not administrate.

However, building those takes an awful long time (and writes hundreds of Mb in disk) so what I really would like to see is to have the possibility to state which ones we want to build (defaulting to NSIS).

seisman:
> Actually the ZIP format can be interesting since what it produces is just zip files instead of an installer. This has it's own value. Either for people who don't like installers or may face permissions problems in machines they do not administrate.

Agree. I also like a portable installer in ZIP format. But currently, it generates several ZIP files (i.e. GMT-6.0.0-Runtime.zip, GMT-6.0.0-Documentation.zip, GMT-6.0.0-GSHHG.zip, GMT-6.0.0-DCW.zip, ...). Is it possible to have all these in only one ZIP file?

> However, building those takes an awful long time (and writes hundreds of Mb in disk) so what I really would like to see is to have the possibility to state which ones we want to build (defaulting to NSIS).

You can use `cpack -G NSIS` to generate the NSIS format only.

joa-quim:
Thanks for the ``cpack -G NSIS``. Will use it all the times.
Regarding the other files that the ``ZIP`` target writes (TWICE), sory don't know.

seisman:
Let's keep this PR WIP and I'll see if there are some configurations to generate an all-in-one ZIP format.

joa-quim:
Yep.

seisman:
See #1080 instead.

PR #1093, opened 2019-07-03
Fix a crashing point if bad Fs

joa-quim:
This would crash because it would try to access to the NULL pointer

gmt subplot begin 3x2 -BWSne -M0 -A -Fs2i,


PaulWessel:
So, does this go into 6.0 after we set the tag?  This isn't a critical bug fix I think.  We should set that tag first etc and then merge this, no?

joa-quim:
Yes. No need to restart because of this one.

PaulWessel:
OK, heading in to work now and will deal with placing the produces in 1+ hour.

PR #1094, opened 2019-07-03
Fix CPack output filename on Windows

seisman:

`CPACK_SYSTEM_NAME` is empty here, resulting in `gmt-6.0.0rc2-.exe` on Windows. This PR fixes the issue by checking the 64 bits or 32 bits of the target architecture.


seisman:
@joa-quim Can you check if this PR can fix the installer name problem? 

You don't need to generate the installers, just run the cmake configuration, and you should see something like:
```
-- Performing Test HAVE_TRADITIONAL_CPP
-- Performing Test HAVE_TRADITIONAL_CPP - Failed
-- Using CFLAGS = '-Wextra -Wall -Wdeclaration-after-statement -O3 -DNDEBUG'
###gmt-6.1.0rc2-darwin-x86_64###
*
*  GMT Version:               : 6.1.0rc2
*
*  Options:
*  Found GSHHG database       : /Users/seisman/Gits/gmt/gmt/share/coast (2.3.7)
*  Found DCW-GMT database     : /Users/seisman/Gits/gmt/gmt/share/dcw
```

joa-quim:
Yes, it prints the right name ``###gmt-6.1.0rc2-win64###``
Funny, I have a variable (``BITAGE``) in my ``ConfigUser.cmake`` that does exactly that bits detection test.

PR #1100, opened 2019-07-04
Fix spelling errors.

sebastic:
The lintian QA tool reported a spelling error for the Debian package build of GMT 6.0.0rc1:

 * reencode -> re-encode


PaulWessel:
Thanks, I fixed these but via the original files.

PR #1107, opened 2019-07-04
Fixing various typos in the docs, plus allow -Zfirst/last in gmt convert

PaulWessel:
The gmt convert **-Z** option is odd compared to others in that it allows a colon separating a _start_ and _stop_ setting.  I switched the docs and parsing to look for a slash but also handle the colon as backards compatibility.  The rest are typos and formatting issues in the rst docs.
@seisman, please look at the synopsis for coupe which has a trailing **/d** in the **-S** option.  Probably should be something else?  I did not change it.


doc/rst/source/gmt.rst
```diff
@@ -55,7 +55,16 @@ or the session ends.  The *prefix* is used to build final figure names when exte
 are automatically appended. The *format* setting is a comma-separated list of desired
 extensions (e.g., pdf,png). 
 
-**gmt end**
+**gmt insert** [*arguments*]
```
seisman: Typo here. Should be inset

PR #322, opened 2019-01-17
Add grdcontour option for grd2kml

btozer:
**Description of proposed changes**

This branch is a work in progress to add a grdcontour option for grd2kml. This is required for producing final Google Earth products for our global grids (e.g. SRTM15+ bathymetry and gravity). 

Creating the contours as a separate vector file and then overlying these on the raster grid isn't an option as Google Earth crashes when importing vectors that exceed a certain size.  

So far, the code has only been edited with some basic groundwork (this may need reviewing and fixing).  But the idea is to add an option flag (e.g. -W) for overlaying contours using either:

a constant sample spacing and pen attributes (e.g. spacing=500m pen=1p,black)
gmt grd2kml input.grd -I+a340 -Ctable.cpt -W500+w1p,black -Noutput.kml

or

an input text file that contains the contour levels and corresponding pen attributes
gmt grd2kml input.grd -I+a340 -Ctable.cpt -Wcontours.txt -Noutput.kml

contours.txt =
-4000 1p,grey
-3000 1p,grey
-2000 3p,black
-1000 1p,grey

"Under the hood" - grdcontour will need to be called after grdimage and before psconvert.

Fixes #142 

**Reminders**

- [ ] Make sure that your code follows our style. Use the other functions/files as a basis.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Describe changes to function behavior and arguments in a comment below the function declaration.
- [ ] If adding new functionality, add a detailed description to the documentation and/or an example.

PR #1115, opened 2019-07-06
Let grdcontour/pscontour be able to have unique pen per contour

PaulWessel:
The **-C**_file_ option allows these modules to read specific contours, flag them as **C** or **A**, and optionally get a fixed annotation angle.  This PR adds an optional 4th item, a _pen_, that may be read from the file.  This allows each contour to be drawn with a unique pen.  If the pen is not present then we default the the one set via **-W**.  This enhancement is in support of #322.


PaulWessel:
Hi @seisman, why is Travis CI saying it is in the progress while the Details says it passed 20 hours ago?

seisman:
I don't see any travis ci builds running now. I think the travis ci build of this PR passes.

PaulWessel:
Great, please review this and #1116 if you can.

Issue #1112, opened 2019-07-06
Let *contour accept different pens for different contours


PaulWessel:
**Description of the desired feature**

This request is tied to the #322 PR. Implementation of that feature would be very much simplified if grdcontour had a way to accept a list of contours with specific pens.  As it is, it would be much more involved to do so.  Right now, there are four ways to specify contours:

1. A fixed contour interval
2. A comma-separated list of contours
3. A file with contor levels in col 1 and **C**|**A** in col 2 [and optional angle in col 3]
4. A CPT file

The closest thing to what is needed is way 3 if we added an optional col 4 with pens.  This would just be an enhancement to grdcontour and pscontour and would let grd2kml prepare the required input file by simply adding the two missing columns with a **C** in col2 and the unused 0 in col 3.

Any other ideas on how to handle this? It seems like a useful addition to the contour tools and has the side-effect of making implementation of #322 simpler.

PaulWessel:
I will answer this myself: I think this is the best solution since it does not add any new option and it is optional within an existing option.  This is intended for master and not for 6.0.0 release.

seisman:
Closed by #1115.

PR #1117, opened 2019-07-06
Add -Wfile option in grd2kml for contour overlays

PaulWessel:
Enable the overlay of select contours, as suggested by #322.  Need some testing first and merging of #1112.


PaulWessel:
I have finished implementing this feature.  It took a few interesting turns. First, we needed the ability to pass a virtual file with contour levels and pens to grdcontour from grd2kml.  However, we previously allowed passing virtual CPT files so now grdcontour did not know what type of virtual file was passed.  This was solved by adding a new API function GMT_Inquire_VirtualFile, which returns the family of the data object in question (e.g., GMT_IS_DATASET, GMT_IS_PALETTE, etc.). Then, this exposed a bug in the initial machinery for reading contours and pens which was fixed.  Finally, I felt the new option to specify contours should have a simple way to select all the contours implied by the cpt and instead just specify a common pen, hence I settled on the syntax **-W**_contfile_|_cpt_.  If the argument cannot be accessed we assume it is a pen and not a file. FYI, @btozer, this new option will be available in master (i.e., probably released as 6.1)  and not the upcoming 6.0.  Please build from master (after this PR is merged) and let me know if there are any issues.

Attached is a screen dump from Google Earth which shows the result of my test command

`gmt grd2kml ellice.nc -I+d -Nellice -L512 -T"Ellice Basin Bathymetry" -Q -W1p -Ct.cpt
`

conts

PR #1134, opened 2019-07-09
GMT_Create_Options must always allocate memory

### LINKED ISSUE #1132: gmt_init_module is taking liberties it shouldn't

All modules start with an initialization section that is very similar.  They will do this:

`options = GMT_Create_Options (API, mode, args);	/* Set or get option list */
`

and

`if ((GMT = gmt_init_module (API, THIS_MODULE_LIB, THIS_MODULE_NAME, THIS_MODULE_KEYS, THIS_MODULE_NEEDS, &options, &GMT_cpy)) == NULL) bailout (API->error); /* Save current state */
`

before carrying out parsing of options.  However, we have a bug here:

GMT_Create_Option either creates and returns a linked list of GMT_OPTION structures or it just passes back its input argument (since it already is a linked list of structures).  In all modules called from the command line we know the former is the case, hence options is created from the _argc, argv_ arguments passed in, and that linked list is freed when the module returns by a call to gmt_M_free_options.

However, API users who may create their own linked list of options and pass those plus mode = GMT_MODULE_OPT expect then to free their linked list after the module (GMT_Call_module) has returned.  Unfortunately, gmt_init_module will in some cases make changes to the option list, and this may in fact change the order of things in the list.  I think this is likely to be the cause of @joa-quim crashes related to memory corruption.

PaulWessel:
When users of the API passes a linked list of GMT_OPTION struct pointers to _GMT_Create_Option_ it cannot just return that pointer but must duplicate the list first and return the head of that duplicate.  That way, _gmt_M_free_options_ can always free the options link so that we honor the "const" modifier for the input argument to _GMT_Create_Options_. See #1132 for context.  Closes #1132.

### DIFF
--- a/src/gmt_parse.c
+++ b/src/gmt_parse.c
@@ -449,16 +449,18 @@ GMT_LOCAL struct GMT_OPTION *fix_gdal_files (struct GMT_OPTION *opt) {
 /*! . */
 struct GMT_OPTION *GMT_Create_Options (void *V_API, int n_args_in, const void *in) {
 	/* This function will loop over the n_args_in supplied command line arguments (in) and
-	 * returns a linked list of GMT_OPTION structures for each program option.
-	 * These will in turn will be processed by the program-specific option parsers.
+	 * then creates a linked list of GMT_OPTION structures for each program option.
+	 * These will in turn will be processed by the module-specific option parsers.
 	 * What actually happens is controlled by n_args_in.  There are three cases:
-	 * n_args_in < 0 : This means that in is already a linked list and we just return it.
+	 * n_args_in < 0 : This means that in is already a linked list and we just duplicate and return a copy.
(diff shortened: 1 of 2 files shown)


PaulWessel:
This may or may not help with your Julia memory issues, but is the right thing to do anyway.

joa-quim:
Nice. Julia crashes disappeared. But if it's a duplicate that is being returned who frees the original?

And only now I noticed this deadly weapon
```
#define gmt_M_free_options(mode) {if (GMT_Destroy_Options (API, &options) != GMT_OK) exit (GMT_MEMORY_ERROR);}
```
We CANNOT have *exits* in external interfaces as the blow out the caller as well.

PaulWessel:
Since you are creating they original, you need to delete it.  The module takes care of the duplicate.
That if test will never return not GMT_OK but we could simply remove the if-test part and just call destroy and not worry about it.

PR #1150, opened 2019-07-10
Add ffmpeg exe and dlls to wininstaller

joa-quim:


joa-quim:
I built ffmpeg with vcpkg. Less than 1/3 of the size of those at the official site (built probably on linux). The instructions to install ffmpeg on our hw1 page are simply terrifying.

PR #1157, opened 2019-07-12
Let pstext handle lack of input when -F+c+t is used.

PaulWessel:
Must turn of implicit input.


seisman:
This PR may introduce a bug.

Try running this command:
```
echo 0 0 | gmt pstext -R0/40/20/60 -JM12 -Bafg -F+f30+cMC+tMAP -P > lixo.ps
```
There should be a fancy frame for this basemap.

joa-quim:
You are right about the fancy frame, but it's not because of the ``echo 00 | ...`` part. Without it the result is the same.
```
gmt pstext -R0/40/20/60 -JM12 -Bafg -F+f30+cMC+tMAP -P > lixo.ps
```

joa-quim:
At least it didn't break previous behavior. Still works
```
echo MAP | gmt pstext -R0/40/20/60 -JM12 -Bafg -F+f30+cMC -P > lixo.ps
```

joa-quim:
It's not only the fancy. No annots whatsoever.

PaulWessel:
Yes, the clip path is on during annotation - will see why.  With -N it looks fine...

PR #1159, opened 2019-07-12
Add example on how to plot text without an input file.

joa-quim:


joa-quim:
FYI. A ``lifigurexo`` is a mix of a ``lixo`` and a ``figure``

Issue #1171, opened 2019-07-14
gmt end show gives error for figure names starting with a module name


PaulWessel:
**Description of the problem**

If a figure has a name that starts with a valid classic module name then the call to docs fails. This script shows the problem:

```
gmt begin pscoast_fig png
	gmt pscoast -RFR -Gred -B
gmt end show
```

which fails to open the PNG file and says

> docs [ERROR]: For now, HTML documentation only uses GMT modern mode names, hence pscoast_fig.png will display as coast
> docs [ERROR]: Shared GMT module not found: pscoast_fig.png 
> end [ERROR]: Failed to call docs
> end [ERROR]: process_figures returned error 78

This only happens for a classic module name.  Renaming the figure name to coast_fig works.

PaulWessel:
Some future improvements to docs:

1. Loop over its arguments so more than one file or document reference can be given.
2. Call GMT_Parse_Common so that **-V** can be handled in a normal way.

PR #1174, opened 2019-07-15
Improved option parsing for docs

PaulWessel:
See #1171 for context.  This PR improves argument parsing for docs.  Apart from the restriction on -Q (must be first argument to docs if used), users can place arguments in any order.  Under macOS, multiple figure references are collected and opened with a single call to open.  Usage message and man page updated to reflect reality. Closes issue #1171.

### DIFF
--- a/doc/rst/source/docs.rst_
+++ b/doc/rst/source/docs.rst_
@@ -2,14 +2,14 @@
 
 .. only:: not man
 
-    Show the HTML documentation for selected module
+    Show the HTML documentation for selected module or display plots
 
 Synopsis
 --------
 
 .. include:: common_SYN_OPTs.rst_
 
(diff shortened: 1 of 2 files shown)


PaulWessel:
Because there are ifdefs for the three major OS and I only tested on macOS, it would be helpful to test this PR under Windows (@joa-quim) and Linux (@seisman) before approving.

doc/rst/source/docs.rst_
```diff
@@ -81,6 +86,13 @@ To display the URL to the surface man page,.run
 
     gmt docs -Q surface
 
+To open all png files in current directory and the documentation for :doc:`begin`, run
```
seisman:
```suggestion
To open all png files in current directory and the documentation for :doc:`movie`, run
```

PR #1182, opened 2019-07-16
Let pscoast not insist on -R -J when not needed

### LINKED ISSUE #1181: Some guys should be allowed to not set -R if first in a modern script
```
gmt begin
     gmt coast -EIT -M

pscoast [ERROR]: Syntax error: Must specify -R option
```

So far the workaround is to set a _e.g._ bogus **-Rg** option. Otherwise, we can move that line to before the ``gmt begin`` but then it will have to be ``gmt pscoat ...``, which probably is even more confusing

PaulWessel:
The case in #1181 did not require **-R -J** but coast was insisting.  It had to do with the THIS_MODULE_NEEDS setting being "JR" but this should be relaxed with just -E -M are given.  I now modify that inside gmt_init_module.  Closes #1181.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -12259,7 +12259,7 @@ GMT_LOCAL int gmtinit_set_last_dimensions (struct GMTAPI_CTRL *API) {
 }
 
 /*! Prepare options if missing and initialize module */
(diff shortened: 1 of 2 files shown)


src/gmt_init.c
```diff
@@ -12285,6 +12285,7 @@ struct GMT_CTRL *gmt_init_module (struct GMTAPI_CTRL *API, const char *lib_name,
 
 	bool is_PS;
 	unsigned int srtm_res = 0;
+	char *required = (char *)in_required;
```
joa-quim: Let's hope this does not screw on VS (it normally only accepts declaration = setting in last declaration line)

PR #1195, opened 2019-07-22
Ensure that if -Lc or -Lt are used there is a data table in -T

PaulWessel:
Otherwise movie will crash.


PaulWessel:
Maybe @seisman can approve this one?

PR #1208, opened 2019-07-23
Add explanation of why it is called coupe

PaulWessel:
Per Feigl it means to cut, related to the crossection.

### DIFF
--- a/doc/rst/source/supplements/seis/meca_notes.rst_
+++ b/doc/rst/source/supplements/seis/meca_notes.rst_
@@ -5,7 +5,7 @@ Bomford, G., Geodesy, 4th ed., Oxford University Press, 1980.
 
 Aki, K. and P. Richards, Quantitative Seismology, Freeman, 1980.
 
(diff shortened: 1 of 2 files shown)


doc/rst/source/supplements/seis/meca_notes.rst_
```diff
@@ -5,7 +5,7 @@ Bomford, G., Geodesy, 4th ed., Oxford University Press, 1980.
 
 Aki, K. and P. Richards, Quantitative Seismology, Freeman, 1980.
 
-\F. A. Dahlen and Jeroen Tromp, Theoretical Global Seismology, Princeton, 1998, p.167. Definition of scalar moment.
```
seisman: We need the leading `\` to make rst happy. Otherwise we may get something like 

- A. Dahlen and Jeroen Tromp, Theoretical Global Seismology, Princeton, 1998, p.167. Definition of scalar moment.

PR #1222, opened 2019-07-24
Clarify that modern and classic have different defaults for colorbars

PaulWessel:
In classic mode, psscale expects **-C**_cpt_ or it will read stdin.  In modern mode, colorbar will use the current CPT unless a **-C**_cpt_ is given to override that choice.  I also added a note about macOS Preview lameness regarding course PDF images.

### DIFF
--- a/doc/rst/source/colorbar_common.rst_
+++ b/doc/rst/source/colorbar_common.rst_
@@ -41,8 +41,9 @@ Optional Arguments
 .. _-C:
 
 **-C**\ [*cpt*]
-    *cpt* is the CPT to be used. If no *cpt* is appended then we use the
-    current CPT.  If no **-C** is given the we read stdin.  By default all
+    *cpt* is the CPT to be used. If no *cpt* is appended or no **-C** is given
+    then we use the current CPT (modern mode only).  In classic mode, if no **-C**
+    is given the we read stdin.  By default all
     color changes are annotated. To use a subset, add an extra column to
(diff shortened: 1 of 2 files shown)


doc/rst/source/colorbar_common.rst_
```diff
@@ -41,8 +41,9 @@ Optional Arguments
 .. _-C:
 
 **-C**\ [*cpt*]
-    *cpt* is the CPT to be used. If no *cpt* is appended then we use the
-    current CPT.  If no **-C** is given the we read stdin.  By default all
+    *cpt* is the CPT to be used. If no *cpt* is appended or no **-C** is given
+    then we use the current CPT (modern mode only).  In classic mode, if no **-C**
+    is given the we read stdin.  By default all
```
seisman:
```suggestion
    is given then we read stdin.  By default all
```

doc/rst/source/colorbar_notes.rst_
```diff
@@ -4,7 +4,10 @@ Notes
 #. When the CPT is discrete and no illumination is specified, the
    color bar will be painted using polygons. For all other cases we must
    paint with an image. Some color printers may give slightly different
-   colors for the two methods given identical RGB values.
+   colors for the two methods given identical RGB values.  See option **-N**
+   for affecting these decisions.  Also not that for years now, Apple's
```
seisman:
```suggestion
   for affecting these decisions.  Also note that for years now, Apple's
```

PR #1267, opened 2019-08-01
Remove --PAR=value from session commands that don't use it

PaulWessel:
E.g., begin, end have no use for **--PAR**=_val_ but it was listed on usage message.  The docs had it correctly.


seisman:
Don't forget to update the manpages.

PaulWessel:
Hm, I looked in begin.rst_ etc., and they did not list --PAR-=value there.  What did I miss?

seisman:
image

--PAR=value is listed in `explain_help.rst_`.

PaulWessel:
Yes, I need to do that separately.  Right now, idoing
.. include:: explain_help.rst_

prints many things, including the --PAR=value.  We will need to make a separate version of that file that does not print the PAR-value since only these modules should not have it.  It is the same problem in the C code with usage.  they still list PAR-value and will need a similar fix.  So this PR only deals with the first part of the update and can be approved I think.

PR #1291, opened 2019-08-02
Make it such that -J is not needed when creating a Geotif from a georeferenced nc grid.

joa-quim:
This will also easy life for MB-System users to replace their mbgrdtiff for grdimage.


seisman:
Should this commit also go to 6.0 branch?

Issue #1207, opened 2019-07-23
Improve meca syntax


PaulWessel:
**Description of the problem**

Some of the seis options are very old-style, e.g. the **-S** option in meca has several optional items (font size, offset) given via slash-separated items, and there is a **+u** to change the justification of the text but only above and below is available.  Instead, we should introduce backwards-compatible optional modifiers **+f**_font_, **+o**_offset_ and **+j**_justification_ as is done elsewhere in GMT.

seisman:
Also, -S option should be a required option.

seisman:
There is also an incompatibility about the required columns for input data. For example, -Sa needs 10 columns. 
```
lon lat depth strike dip rake newX newY event_tile
```

In GMT4, the last three columns are optional. Since GMT5, only the last column is optional.

seisman:
-S option was improved in #1298 and 1304, and now it's a required option.

For the incompatibility of required input columns, it seems difficult to have a flexible number of columns with current GMT I/O mechanism. 

Closed it as won't be fixed.

PR #1298, opened 2019-08-04
meca: Improve -S option while keeping backwards-compatibility

seisman:
As pointed in #1207, the -S option in meca and coupe are old-style. 

This PR introduces  backwards-compatible optional modifiers **+f**, **+j** and **+o** to meca -S option. All seis_* tests pass.

Compared to the old-style syntax, the new syntax is easier to use and is more flexible:

- `+f<font>` now can control font, size and color of event label
- `+j<justify>` can put event label at any position
- `+o<dx>/<dy>` can add more offsets to event label

The behavior of `+j<justify>` and `+o<dx>/<dy>` is very similar to placement of embellishments with **-DJ***justify*+o*dx*/*dy*. See the figure below with different justification code.

image

**Questions:**

1. -S is a required option. Do I need to update **THIS_MODULE_NEEDS**?
2. **THIS_MODULE_OPTIONS** is `-:>BHJKOPRUVXYdehipt`. What's the meaning of `H`?

**Known issues:**

- `-Fr` (adding a text box to event label) doesn't work, even in GMT5. Currently I disable the feature and will see how to fix it.

### DIFF
--- a/doc/rst/source/supplements/seis/explain_meca_-S.rst_
+++ b/doc/rst/source/supplements/seis/explain_meca_-S.rst_
@@ -1,11 +1,11 @@
-**-S**\ *<format><scale>*\ [/*fontsize*\ [/*offset*]][**+u**\ ]
+**-S**\ *<format><scale>*\ [**+f**\ *font*][**+j**\ *justify*][**+o**\ *dx*\ [/*dy*]]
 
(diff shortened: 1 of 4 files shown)


seisman:
@PaulWessel I've updated THIS_MODULE_OPTIONS.

Issue #871, opened 2019-06-08
GMT_DOC_URL: link to latest or specific version


seisman:
**Description of the desired feature**

GMT_DOC_URL is used by `gmt docs` to open HTML documentations. Currently it's set to http://docs.generic-mapping-tools.org/latest.

Before making public releases, we need to determine where GMT_DOC_URL links to.

1. always links to `latest`.
2. links to the version of the current release (i.e. http://docs.generic-mapping-tools.org/6.0.0 when releasing 6.0.0)

I prefer the second option, so that users always read the documentations matching the GMT version they're using.

PaulWessel:
Hard to argue with that, I agree on option 2 as well.

joa-quim:
Many projects in Github have two badges pointing to documentation. One with the latest official version and the other one with 'latest'. That's what we have also at the Hawaii GMT site (less the badges).

PaulWessel:
But we cannot have two defaults in GMT.  This is where

gmt docs gravies

will take you.

joa-quim:
If you look a https://github.com/GenericMappingTools/GMT.jl/tree/gh-pages you'll find the docs of previous versions. But that was generated by the registration script so I don't know how to do that. What we have in the ``gh-pages`` branch it's only ``dev``. If we had a ``stable`` or a ``6.0.0``, than we could have the two badges.

joa-quim:
Ops, it looks I missed the ``gmt docs`` part. Thought we were talking about the links on the Github page.

seisman:
Closed by #1551.

Issue #870, opened 2019-06-08
Move Matlab wrapper documentations to https://github.com/GenericMappingTools/gmtmex


seisman:

seisman:
@PaulWessel @joa-quim Can any of you take some time migrating the file matlab_wrapper.rst to the gmtmex wiki page?

PaulWessel:
Hi @joa-quim, could you help? I have to do Chair work for a while now after the paper and proposal both have been submitted.

joa-quim:
I just changed the link at the top of the readme page. We can leave the ``matlab_wrapper.rst`` file trailing around but no included in case we later want to recover some bits of it.

meghanrjones:
Could we just move this ReST file to the [GMT sandbox repo](https://github.com/GenericMappingTools/sandbox) so that we can close this issue?

PaulWessel:
Yes, that would be OK with me so we have a saved copy of it.

Issue #851, opened 2019-06-06
Relpace classic gallery scripts with modern mode


PaulWessel:
**Let the gallery use modern mode**

With the push to modern mode it makes sense to simplify our gallery by using modern mode as well.  Right now, only 13 of the 50 scripts have a modern mode equivalent in test/exmod.  Perhaps we should make a push to convert the rest in test/exmod first and then we can switch classic and modern when ready?

seisman:
I'll help with the conversions.

PaulWessel:
Thanks, just add in test/exmod for now. To avoid us both doing the same, how about you start on the lower half (1-25) and I start 26-50?

seisman:
OK.

seisman:
@PaulWessel I think we should **NOT** add the modern examples to test/exmod.

The examples are pretty large (~40 Mb). Converting these examples to modern mode and adding them to test/exmod means adding another 40 Mb to the repository. 

Although the classic examples will be deleted someday, the 40 Mb files always stay somewhere in the git history. Currently, the repository size is as large as 700 Mb, and the `.git` directory is almost 500 Mb. 

We should try our best to avoid adding more large files to the git history.

Issue #816, opened 2019-06-01
Redesign GMT documentations


seisman:
**Description of the problem**

GMT documentation is using a pretty old RTD theme. Any special reason for this old theme? What about updating to the latest [sphinx_rtd_theme](https://github.com/rtfd/sphinx_rtd_theme), which is possibly the most popular sphinx theme.

PaulWessel:
With GMT6 coming out real soon, seems very timely to change the look and feel of the docs.  Leo is working on a web landing page. What you think, @leouieda and @joa-quim?

joa-quim:
Yes, let's do like women and relocate the furniture around the room.

PaulWessel:
Spoken like a man who has endured an involuntary room rearrangement.

joa-quim:
Well, haven't we all?
(mothers count too)

leouieda:
This is more like a bachelor cleaning up the mess in the apartment so he can get a date because no respectable woman would want to live there.

leouieda:
I agree with @seisman that we should update the theme while we're at it. At some point things are gonna stop working because the dependencies are too old.

PaulWessel:
If this is just changing a setting or two in the conf.py.in then by all means give it a try now.

seisman:
Closed by #1308.

PR #1308, opened 2019-08-06
Redesign documentations

seisman:
@leouieda I create the PR so that we can easily track the progress and contribute PRs to the docs-refactor branch.

This PR is related to #805, #816, #851, #870, #871.


leouieda:
Thanks, Dongdong!

leouieda:
Hi @seisman sorry for the delay. I'm pushing in some changes I made recently. Please don't merge master or 6.0 branches here since it creates a lot of conflicts that need to be resolved. Or at least give me a heads up :wink:

seisman:
It seems @leouieda is busy and have limited time working on the documentations. Maybe I can take over this branch and commit changes directly, so that we can have the new documentation ready when releasing 6.0.0.

PaulWessel:
I know, but I think we need to give @leouieda a chance to resurface on Monday at the latest.  If nothing by Monday night GMT then you can make changes.

seisman:
👌

leouieda:
Hi guys, sorry for the disappearance. It's ok with me if @seisman takes over this PR. A few things that were left unfinished:

- [ ] Break the tutorial document into several documents. I'm thinking one for each chapter and include them separately in the `index.rst` toc table.
- [ ] Add some text to the front page about GMT, how the docs are laid out and where to start if you don't know anything about GMT.
- [ ] A page explaining modern mode (*Modern mode primer*) for people who know GMT. Could be taken directly from the paper.

seisman:
Although the new documentations still needs more improvements, I think it's OK to replace the old documentations. As we have many files renamed or removed in the new documentations, modifying the old documentations may lead to a lot of conflicts which need to be solved. So I think we can merge it into 6.0 branch ASAP, so that we can improve the documentations efficiently.

You can preview the new documentations by [visit this temporary site](https://seisman.github.io/gmt-docs-preview/).

leouieda:
@seisman great job on this! It looks really good. I agree that this can be merged and worked on in later PRs. The only question I have is about the classic mode tutorial and cookbook links. It's a bit strange to me that they go to the GMT 5 docs. I'd rather not have those toctree items and put a note on the tutorial and cookbook pages points to GMT 5 with some explanation why. Like we did in the Core Modules page. Something like:

> Looking for the classic mode tutorial? Since classic mode commands haven't changed since GMT 5, please visit the [GMT 5 tutorial page](https://docs.generic-mapping-tools.org/5.4/GMT_Tutorial.html) instead.

joa-quim:
Sorry, I've not paid much attention to this. I looks very nice.
I note however that we lost the [GMT -J codes](https://docs.generic-mapping-tools.org/dev/proj_codes_GMT.html).

Also, history is round. we came back to GMT3-4 were modules were resented in thematic categories. I'm sorry to disagree here but I always searched in the alphabetic order.

We should have a GMT ``Core Modules - thematic`` and a ``Core Modules - alphabetic`` entries that users can change as bookmarks depending on their preferences.

PaulWessel:
Hi @seisman and @leouieda - this looks very nice; thank you!  I think this looks good enough to publish for GMT 6.0.0.  I do agree with @joa-quim that having alphabetical access is what you want when you know which module you need to look up instead of hunting for it a bit under the categories, but we can add that as he suggests if we find it really is a problem.  The -J codes page: Seems like that could easily be added under the Reference Documentation heading.  Maybe call it "Projection Specifications" and add it just after common options?

seisman:
I've addressed some of the comments above, and the new documentation is available from https://seisman.github.io/gmt-docs-preview/.

seisman:
The core module list (https://seisman.github.io/gmt-docs-preview/modules.html) has the same problem, because we're using multiple tables for thematic categories. Will see how to fix it later.

KristofKoch:
May I show you a problem I just noticed with the theme @seisman chose:

Screenshot

When using the search function the term searched for becomes highlighted and the word containing the search term becomes separated into two words. This surely will lead to errors when user don't use copy&paste but read the docs and type their own code. `pssolar` ≠ `ps solar`

joa-quim:
Hmm, is it a Mac'ish thing? It doesn't happen to me on Win with both Firefox and Epic. But I'm looking at Dongdong's link page where, BTW, there is no ``pssolar`` reference.

seisman:
The new site is online https://docs.generic-mapping-tools.org/latest/ now. @joa-quim

seisman:
I can reproduce it with Chrome and Safari, and just found a solution. If you think it's a problem, then I can fix it in minutes.

PaulWessel:
Looks like pssolar is missing from classic module list?

Issue #195, opened 2018-12-01
Incorrect cropping if the origin of second panel is >12 cm below the first panel


seisman:
**Description of the problem**

**Full script that generated the error**

```
gmt begin map jpg
gmt basemap -JX10c/10c -R0/10/0/5 -Baf
gmt basemap -JX10c/10c -R0/10/0/5 -Baf -Y-13c
gmt end
```

**Output Figure**
![map](https://user-images.githubusercontent.com/3974108/49330616-fcb12980-f55e-11e8-9995-715d814e1dce.jpg)

**System information**

* Operating system: macOS
* Version of GMT: 6.0.0_15a9160

PaulWessel:
In modern mode we basically have an enormous canvas that we clip to the final size.  However, if you purposefully move the origin way to the left or down then this will happen.  Perhaps we should position the origin at (1 meter, 1 meter) to handle relative innocent shifts like this from still working.

seisman:
Yeah. I think most people like to plot the top left panel first. 1 meter seems enough for most plots.

PaulWessel:
We were using GMT_PAPER_MARGIN ( = 5 inches) for this or ~12.7 cm as you found.  I am setting it to 40 inches (a bit over 1 meter) and will submit PR once make check returns without any side-effects.

joa-quim:
Well, we have a 100 m wall don't we? So 1 m may not be that much for people that like to start from top down (but I surely prefer an Y axis positive up).

PaulWessel:
It is 11.55 meters. (32767 points)

PaulWessel:
Bug has now been fixed in ghostscript repository so we are awaiting the 9.27 release before we can close this issue.

meghanrjones:
Just an update that I still get the same incorrect cropping with gs 9.55.0.

PaulWessel:
And I see GMT_PAPER_MARGIN is still 5 in 6.3.0.  Not sure what happen here - the above says we changed to 1 m.  I will again set to 1 m (~40 inches) and we can run tests etc.  Unclear to me why this did not stick.  I agree setting aside a full meter is fine given the canvas is 11 m.

meghanrjones:
[...]

PR #1315, opened 2019-08-06
Reset subplot row,col|index to start at zero

PaulWessel:
In order to be consistent with all other places in GMT where rows or columns are given, we will set subplot row,col also follow the same convention.  Thus 0,0 is the top left panel.
Implementation changed some unsigned ints to int and several doc and test scripts needed to be updated.  Only failure (for me right now) was anim_08.sh which seems to have to do with NEIC data changes.

### DIFF
--- a/doc/rst/source/explain_-c_full.rst_
+++ b/doc/rst/source/explain_-c_full.rst_
@@ -7,4 +7,4 @@
     no **-c** is given and we just entered subplot mode then the first
     panel (top, left) is selected.  Instead of *row, col* you may give
     the one-dimensional *index* which depends on the order you set via **-A**
-    when the subplot was defined.
+    when the subplot was defined. Note: *row*, *col*, and *index* all start at 0.
--- a/doc/rst/source/subplot.rst_
+++ b/doc/rst/source/subplot.rst_
@@ -172,11 +172,11 @@ Optional Arguments
 ------------------
 
 *row,col*
-    Sets the current subplot until further notice.  If not given we go to the next panel by order
+    Sets the current subplot until further notice.  Note: First *row* or *col is 0, not 1. If not given we go to the next panel by order
     specified via **-A**.  As an alternative, you may bypass the **set** mode and
     instead supply the common option **-c**\ [*row,col*] to the first plot command you issue in that subplot.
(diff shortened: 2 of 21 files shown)


seisman:
The script below works but generate some "error" messages:
```
gmt begin map pdf
gmt subplot begin 3x2 -Fs5c/5c
    gmt basemap -R0/10/0/10 -Baf -c0
    gmt basemap -R0/10/0/10 -Baf -c3
gmt subplot end
gmt end show
```

Error message:
```
gmt [ERROR]: Index 0 goes to (0, 0)
gmt [ERROR]: Index 3 goes to (1, 1)
```

seisman:
`-c-1` also works. Is it a feature?
```
gmt begin map pdf
gmt subplot begin 3x2 -Fs5c/5c -A
    gmt basemap -R0/10/0/10 -Baf -c-1
    gmt basemap -R0/10/0/10 -Baf -c3
gmt subplot end
gmt end show
```

PaulWessel:
oops.  What do you mean it works? It goes to the first panel?
I will add a check that the index is not negative (I think I check that row,col are not negative).

seisman:
> oops. What do you mean it works? It goes to the first panel?

Yes, `-c-1` goes to the first panel.

seisman:
Another issue is the figure is incorrectly cropped at the bottom. To reproduce the issue, run:
```
gmt begin map pdf
gmt subplot begin 3x2 -Fs4c/4c -A
for i in $(seq 0 5); do
    gmt basemap -R0/10/0/10 -Baf -c
done
gmt subplot end
gmt end show
```

image

The cropping is correct if I change `3x2` to `5x2`. Thus the issue may be not related to #195.

PaulWessel:
Gets worse if I extend to 4x2. ghostscript again? I mean, gs is doing the cropping. Doing ps I see we are running off the page, so perhaps not gs.  THis is not related to the new row,col stuff but was there before, right?

seisman:
What's the plot origin for the first panel?

PaulWessel:
I will look into it, but since this problem exists in 6 with old 1-n, 1-m specifications I think we can merge this PR in and then work on this problem.

PR #1316, opened 2019-08-06
Add modifier +s to -Sr to specify a rectangle via opposite diagonal corners

PaulWessel:
It is often desirable to draw a rectangle based on the coordinates of one of the diagonals. Thus, **-Sr+s** will do this and expects

`lon1 lat1 lon2 lat2
`

for points on any diagonal.  It creates a polygon and is thus subject to **-A** as well as **-W -G**.


src/gmt_plot.h
```diff
@@ -139,6 +139,7 @@ struct GMT_SYMBOL {
 	bool shade3D;		/* true when we should simulate shading of 3D symbols cube and column */
 	bool fq_parse;		/* true -Sf or -Sq were given with no args on command line and must be parsed via segment headers */
 	bool accumulate;	/* true if -So takes many band z and they are increments, not total z values */
+	bool diagonal;		/* true if -SJ+s is given */
```
seisman:
```suggestion
	bool diagonal;		/* true if -Sr+s is given */
```

Issue #1342, opened 2019-08-09
Should undocumented figure fill/outline in subset be removed?


PaulWessel:
**Description of the problem**

During implementation I added various debug-related modifiers to **-F** in subplot:

- **+d** draws the rectangles around what is considered the total dimensions of each subplot (this includes area computed to hold ticks, annotations)
- **+g** sets fill for the total composite figure domain
- **+p** sets pen to draw outline of total composite figure domain.

For instance, this code:

```
gmt begin panels
  gmt subplot begin 2x2 -Fs3i+glightgray+pfaint+d -M5p -A -SCt -SR -Bwsne
    gmt basemap -R0/80/0/10
    gmt basemap -c
    gmt basemap -c
    gmt basemap -c
  gmt subplot end
gmt end show
```

yields the plot below  My question: is either **+g** or **+p** useful to users and should be documented? I cannot see the point of **+p**, but perhaps **+g**?

![panels](https://user-images.githubusercontent.com/26473567/62748682-36430a00-b9f5-11e9-9fd8-695308b8ab19.png)

seisman:
I don't think any of these three modifiers are useful to users. I suggest keeping it as is.

PaulWessel:
I agree, will keep them hidden for developers only.

joa-quim:
Guys, you were too quick on this one. I think the **+d** could be interesting in some cases. Imagine for example a portfolio of global map projections, each one seating in its own box. But it should be expanded to **+d<pen>** (didn't get what **+p** does as it seem to do what I understood **+d** would do, except the red).

seisman:
**+d** draws the box for each subplot, and **+p** draws the box of the whole figure.

PaulWessel:
The +d is debug and will not be revealed to users.  For +p and +g, I can see uses to paint and outline the background when no annotations and frames are used internally.  Consider this plot:
![four](https://user-images.githubusercontent.com/26473567/62803600-9930b180-ba86-11e9-9eff-179b98823468.png)

[Strange, for me it truncates the bottom but the plot is correct].  The script is

```
gmt begin four png
  gmt subplot begin 2x2 -Fs8c+p1p+gyellow -B+n -Vd -M0
    gmt pscoast -Rg -JG0/0/? -Gred -Sblue -B0g
    gmt pscoast -JG90/0/? -Gred -Sblue -B0g -c
    gmt pscoast -JG180/0/? -Gred -Sblue -B0g -c
[...]

PR #1352, opened 2019-08-09
Document subplot -F modifiers +g, +p, and new +c

PaulWessel:
These were not documented but where there all along, except **+c**.  This lets you draw the background figure canvas, possibly expanded by **+c**_clearance_, and specify fill and outline.  Makes most sense if **-B+n** was used to set up the subplot.  See #1342 for discussion.

### DIFF
--- a/doc/rst/source/subplot.rst_
+++ b/doc/rst/source/subplot.rst_
@@ -74,7 +74,7 @@ Required Arguments
     a zero *height*, or (2) you can select *height* based on trial and error to suit your plot layout.
 
     Optionally, you may draw or paint the figure rectangle behind the subplots, and even expand it via **+c**.  This
-    is most useful if you supply **-B=n** to subplot begin, meaning no ticks or annotations will take place in the
+    is most useful if you supply **-B+n** to subplot begin, meaning no ticks or annotations will take place in the
     subplots.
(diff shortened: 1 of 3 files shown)


doc/rst/source/subplot.rst_
```diff
@@ -73,6 +73,10 @@ Required Arguments
     to compute the height of each subplot.  All subplots must share the same region and projection and you specify
     a zero *height*, or (2) you can select *height* based on trial and error to suit your plot layout.
 
+    Optionally, you may draw or paint the figure rectangle behind the subplots. and even expand it via **+c**.  This
```
seisman:
```suggestion
    Optionally, you may draw or paint the figure rectangle behind the subplots, and even expand it via **+c**.  This
```

doc/rst/source/subplot.rst_
```diff
@@ -73,6 +73,10 @@ Required Arguments
...
+    is most useful if you supply **-B=n** to subplot begin, meaning no ticks or annotations will take place in the
```
seisman:
```suggestion
    is most useful if you supply **-B+n** to subplot begin, meaning no ticks or annotations will take place in the
```

PR #1353, opened 2019-08-09
Add GMT_GRAPHICS_FORMAT [pdf] as new GMT default setting

### LINKED ISSUE #803: Add a GMT default to configure default figure format in modern mode

In modern mode, if formats are not specified in `gmt begin` or `gmt figure`, the default figure format is PDF. It would be good to allow users to select the default figure format(s).

PaulWessel:

This only affects the default graphics format in modern mode.  Rather than always being pdf, it can now be changed to another valid format via gmt set. Closes #803.

### DIFF
--- a/doc/rst/source/begin.rst_
+++ b/doc/rst/source/begin.rst_
@@ -47,7 +47,7 @@ Optional Arguments
 
 *formats*
     Give one or more comma-separated graphics extensions from the list of allowable graphics formats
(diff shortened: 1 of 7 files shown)


seisman:
Is it possible to support multiple formats, e.g. `png,pdf`?

PaulWessel:
Not as implemented. I could expand it.

PaulWessel:
Let's leave that for 6.1 if we decide to do it.

PR #1354, opened 2019-08-09
Improve parsing of -J and autoscaling for Cartesian projections in subplots

seisman:


PaulWessel:
Yes, that is the part I added. Wait for CI to clear it and merge?

seisman:
Looks good to me. I can't approve it because I opened this PR.

You may change the PR title and add some description to make the PR self-explained.

PaulWessel:
Basically, Cartesian projections will fill the subplot dimensions, so you use -Fs to set the subplot sizes and then Cartesian plots should auto-scale there if you dont give -J.  If you do give a -J then you are overriding that and your plot may no longer be centered in the subplot; see -C for fudging that.

PR #1368, opened 2019-08-12
Discuss run-time installs for the bundle

### LINKED ISSUE #1099: Run-time dependency for macOS installer package

Mac users who just install the GMT bundle will get everything needed to run the basic GMT tools (executable, all required shared libraries like netcdf, GDAL, etc).  However, GMT also expects several executables to be available in the PATH for things to work at run-time:

- ghostscript (gs): psconvert issues system calls to gs to convert PostScript to other formats.
- gdal_translate: psconvert calls gdal_translate to make Geotiffs.
- ogr2ogr: GMT calls ogr2ogr to convert ESRI shape files to GMT/OGR format.

While the GDAL library is present in the bundle, the run-time executables gdal_translate and ogr2ogr are not.  Likewise, while macOS supplies Preview to view files, including PostScript, there is no gs executable installed (need to verify this).  Our discussions with Artifex about suppling a gs executable in our installers have met with resistance.  So, we could presumably add the two GDAL executables to the bundle (@seisman, is there a CMake setting for adding extra executables to the bundle?) but we still need a solution for installing gs.  It seems perhaps the easiest way is to use homebrew (e.g., http://macappstore.org/ghostscript/).  One possibility would then be this:

If macOS and running the bundle gmt executable and there is no gs, set up a system call to launch and install gs via brew.

PaulWessel:
The macOS installer does not have run-time dependencies. Closes #1099.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -93,12 +93,13 @@ run-time dependencies on executables that you will need to install separately:
 - [GraphicsMagick](http://www.graphicsmagick.org/) (Convert images to animated GIFs)
 - [FFmpeg](http://www.ffmpeg.org/) (Convert images to MP4 or WebM videos)
 
-If you are already using any of fink, macport, or homebrew then please see
-[build from source](BUILDING.md) for how to install these packages.  If you do not then
-we recommend installing [homebrew](https://brew.sh/), then run
+Install these via fink, macport, or homebrew.  If you do not use any of these, then
+we recommend installing [homebrew](https://brew.sh/). Now, run either the fink, port,
+or brew command:
 
+    fink install gdal ghostscript graphicsmagick ffmpeg 
+    port install gdal ghostscript graphicsmagick ffmpeg 
     brew install gdal ghostscript graphicsmagick ffmpeg 
-
 
 ### Install via macports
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -85,6 +85,21 @@ or:
...
+Note: While the installer comes with all GMT executables and needed libraries, there are some
+run-time dependencies on executables that you will need to install separately:
+
+- [GDAL](https://www.gdal.org/) (Convert shapefiles and produce geotiffs)
+- [Ghostscript](https://www.ghostscript.com/) (Convert PostScript to PDF or raster images)
+- [GraphicsMagick](http://www.graphicsmagick.org/) (Convert images to animated GIFs)
+- [FFmpeg](http://www.ffmpeg.org/) (Convert images to MP4 or WebM videos)
+
+If you are already using any of fink, macport, or homebrew then please see
+[build from source](BUILDING.md) for how to install these packages.  If you do not then
```
seisman: I think we should list how to install these packages via brew, fink and macport, so that users don't need to read the complicated building instruction.

PaulWessel:
Not sure about fink naming of packages.

PR #1371, opened 2019-08-13
Save and restore original -B settings during inset building

PaulWessel:

We need to keep the last **-B** option used before gmt inset starts an inset, so that we can restore that setting as the current last **-B** setting when the inset ends, regardless of what -B options might have been used during the inset.  This is needed so that any colormap command after the inset ends can determine if the chosen placement is on an axis that was annotated or not and adjust accordingly.


PaulWessel:
Can any of @GenericMappingTools/core have a look and approve? Tests pass.

PaulWessel:
Good.  This means the colorbar command in Fig_4.sh can again be placed naturally after the inset end command and still work as expected.

PR #1374, opened 2019-08-14
Fix documentations of begin, end, docs

seisman:


doc/rst/source/docs.rst_
```diff
@@ -9,7 +9,7 @@ Synopsis
 
 .. include:: common_SYN_OPTs.rst_
 
-**gmt docs** [ **-Q** ] [ |SYN_OPT-V| ] [ *module-name* [*-option*] ]
+**gmt docs** [ **-Q** ] [ |SYN_OPT-V| ] *module-name* [*-option*]
```
seisman: *module-name* is not optional.

doc/rst/source/docs.rst_
```diff
@@ -62,8 +64,6 @@ To see the documentation of the **-B** option in *pscoast*
 
     gmt docs pscoast -B
 
-And do not be surprised to see that it says *coast* instead of *pscoast*.
```
seisman: docs now supports both coast and pscoast

Issue #1397, opened 2019-08-17
inset: inappropriate default value for margins


seisman:
The default value for margins is 0.5 cm in inset. This value is too large for normal-size plots.

https://github.com/GenericMappingTools/gmt/blob/ab17407289f70193e96404cb85a2df3728f203d0/test/exmod/ex44.sh#L8-L18

Take the ex44 as an example, the figure width is 6 inch, which is a suitable size to fit a A4 paper. In such plots, an inset box size of 2 inch seems reasonable. Below is the figure generated without the **-M0** in **inset begin**. It's obvious that the default value is not a good choice for such plots.

![image](https://user-images.githubusercontent.com/3974108/63206579-29718680-c085-11e9-989b-8999eaaea6ce.png).

PaulWessel:
Agree, kind of a random value I think.  0.5 cm is ~14 points and we use points for lots of these defaults.  Want to try 6p an see if that looks OK? I agree it should be a good default for a page sized plot with a smaller inset.  Or maybe 5p? We have several MAP_* defaults that are 5 points.

seisman:
For a circular inset map (i.e. the lower panel), any value between 0-6p seems good. But for a rectangle inset map (i.e. the upper panel), I think the best default value is 0p.

PaulWessel:
Given that there is clearly no single value that can work as a reasonable default in all cases, I agree that the default should be zero and -M needs to be used to change that.

PaulWessel:
Want me to start on that?

seisman:
OK. Go ahead.

PR #1398, opened 2019-08-17
Let inset margins default to none

PaulWessel:
Since there is no sensible default we can choose, it must default to zero.


PaulWessel:
Closes #1397.

PR #1404, opened 2019-08-17
Let subplot -Wpen draw optional dividing lines

PaulWessel:
There are times, especially with **-B+n**, where drawing lines between subplots is required.  Closes #1403.

PR #1407, opened 2019-08-18
Improved -C option in subplot

PaulWessel:
When a constant clearance is common for both w and e we can use **-Cx**_gap_, and if common to both s and n we can use **-Cy**_gap_, and if constant on all sides we simply do not indicate a side, i.e., **-C**_gap_.

### DIFF
--- a/doc/rst/source/subplot.rst_
+++ b/doc/rst/source/subplot.rst_
@@ -17,7 +17,7 @@ Synopsis (begin mode)
 **gmt subplot begin** *nrows*\ **x**\ *ncols*
 **-F**\ [**f**\ \|\ **s**\ ]\ *width*\ /*height*\ [**+f**\ *wfracs*\ /*hfracs*\ ][**+c**\ *dx/dy*\ ][**+g**\ *fill*\ ][**+p**\ *pen*\ ]
 [ **-A**\ *autolabel* ]
-[ **-C**\ *side*\ /*clearance*\ [**u**]]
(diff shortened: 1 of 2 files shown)


doc/rst/source/subplot.rst_
```diff
@@ -103,10 +103,12 @@ Optional Arguments
 
 .. _subplot_set-C1:
 
-**-C**\ *side*\ /*clearance*\ [**u**\ ]
+**-C**\ [*side*]\ /*clearance*\ [**u**\ ]
     Reserve a space of dimension *clearance* between the margin and the subplot on the specified
-    side, using *side* values from **w**, **e**, **s**, or **n**.  The option is repeatable to set aside space
-    on more than one side.  Such space will be left untouched by the main map plotting but can
+    side, using *side* values from **w**, **e**, **s**, **n**, or **xx** for both **w** and **e**
```
seisman:
```suggestion
    side, using *side* values from **w**, **e**, **s**, **n**, or **x** for both **w** and **e**
```

doc/rst/source/subplot.rst_
```diff
@@ -115,7 +117,7 @@ Optional Arguments
 
 .. |Add_-J| unicode:: 0x20 .. just an invisible code
 .. include:: explain_-J.rst_
-
+s
```
seisman:
```suggestion

```

PaulWessel:
Sorry, included the typo fixes in the next branch #1404.

Issue #1416, opened 2019-08-19
Transparency doesn't work with ghostscript 9.28rc1


seisman:
Test script:
```
gmt begin transparency
gmt basemap -R0/10/0/4 -Jx1c -Baf
echo 2 2 | gmt plot -Sc2c -Gred
echo 5 2 | gmt plot -Sc2c -Gred@50
echo 8 2 | gmt plot -Sc2c -Gred -t50
gmt end show
```

PaulWessel:
Make the PS, then run psconvert -S to see what the options to gs are and see if that is among the new problems they are trying to fix with the SAFER stuff.

PaulWessel:
Try to add -dNOSAFER and see if that changes anything,

seisman:
Output of **psconvert -S**:
```
$ gmt psconvert -A -Tf -S transparency.ps                                                                                                  
gs -q -dNOPAUSE -dBATCH -dPDFSETTINGS=/prepress -dDownsampleColorImages=false -dDownsampleGrayImages=false -dDownsampleMonoImages=false -dUseFlateCompression=true -dEmbedAllFonts=true -dSubsetFonts=true -dMonoImageFilter=/FlateEncode -dAutoFilterGrayImages=false -dGrayImageFilter=/FlateEncode -dAutoFilterColorImages=false -dColorImageFilter=/FlateEncode -dSCANCONVERTERTYPE=2 -dMaxBitmap=2147483647 -dUseFastColor=true -sDEVICE=pdfwrite  -g3160x1510 -r720 -sOutputFile='transparency.pdf' './psconvert_41626d.eps'
```

I then run the command above with the *-dNOSAFER* flag, the transparency works.

PaulWessel:
OK, I have asked the question of the gs developers if this is intentional or not.

seisman:
Another small issue is, when I run `gmt psconvert -A -Tj -S transparency.ps`, it prints two gs commands, with the second one starting with "psconvert [ERROR]:":
```
$ gmt psconvert -A -P -Tj -S transparency.ps
[...]

PR #1429, opened 2019-08-21
Make sure psconvert -A is used in modern mode if not PostScript

PaulWessel:
Users may select special settings and forget to add A.  See #1416.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -15554,6 +15554,8 @@ GMT_LOCAL int process_figures (struct GMTAPI_CTRL *API, char *show) {
 						strcat (cmd, option);
 					}
 				}
+				if (not_PS && strchr (fig[k].options, 'A') == NULL)	/* Must always add -A if not PostScript */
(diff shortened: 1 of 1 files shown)


seisman:
The script below doesn't work any more.
```
gmt begin bug1m png -A2c
gmt basemap -R0/10/0/5 -Jx1c -B0
gmt end show
```

PaulWessel:
Hm, works fine here (9.27).  What goes wrong?

seisman:
With **-A2c**, I expect to see extra margins around the four corners, but actually, I don't see any margins.

PaulWessel:
Ah yes, you are right about that...

seisman:
Maybe it's not related to this PR. There are also no extra margins with the latest 6.0 branch.

PaulWessel:
There is no - in A, just A2c

seisman:
Oops, my mistake.

PaulWessel:
But no error given, so still a problem.  I will check.

PR #1473, opened 2019-08-30
Update INSTALL and BUILDING instructions

seisman:


seisman:
@PaulWessel I just made some changes and mentioned how to install ghostscript, graphicsmagick and ffmpeg for homebrew, fink and macports users, and also macOS bundle users.

PR #1477, opened 2019-09-01
On Windows, check if a ghostscript is present in GMT/bin.

joa-quim:
In anticipation to the possibility of us shipping gs in the Win installer.
On Windows, check if a ghostscript is present in GMT/bin. If yes use it instead of any installed one.


joa-quim:
The entire ``ghostbuster()`` function is under a ``#ifdef WIN32``

Issue #1421, opened 2019-08-20
Revisit the handling of hyphen-minus inconsistence


seisman:
Related to #196, #198, #391, #644.

Sorry to be too picky. I'm not quite satisfied with how GMT handles hyphen and minus. 

See the script and the generated figure below as a possible use case. If I don't have much knowledge about the difference between hyphen and minus, I think I would be confused why spaces in a string change the appearance of **-**.

```
gmt begin isolatin png
gmt basemap -R-10/10/-10/10 -JX10c/10c -Baf
gmt text -F+f18p << EOF
0 0 123 - 456
0 2 123- 456
0 4 123 -456
0 6 123-456
0 -2 ABC - DEF
0 -4 ABC- DEF
0 -6 ABC -DEF
0 -8 ABC-DEF
EOF
gmt end show
```

![isolatin](https://user-images.githubusercontent.com/3974108/63317713-e7914c00-c2e1-11e9-891b-989e2dee7049.png)

I know that GMT internally converts some hyphens to minus, following some rules. I don't know the exact rules, but I think it's based on the definition of hyphen, as mentioned in comments of previous issues/PRs: 
> a short line (-) used to connect the parts of a compound word or the parts of a word divided for any purpose

However, I don't think the definition is always true. I can imagine many use cases where I need a hyphen, not a minus. For example, command line option **-B** should use a hyphen, instead of a minus. Sometimes, we even need a en-dash or a em-dash. I don't think there are perfect rules to determine when a hyphen should be used or not.

Actually, I quite like the idea proposed by @PaulWessel in https://github.com/GenericMappingTools/gmt/pull/196#issuecomment-443550172. 

> 1. When GMT is plotting annotations etc. we know we are dealing with numbers and hence we select the minus code. That code differs across char sets.  This ensures uniformity and compliance to proper math type setting.
> 2. For free-standing text provided via pstext or -B etc, we are not concerned with formatting numerical data and any "-" character will be considered to be a hyphen and we insert the correct hyphen code, which also differs across char sets.  This also ensures consistency and avoid using very wide dashes (en-dash is very similar to math minus) in hyphenated words.
> 3. If you are very picky and want to write -200 in some text string and want the minus symbol then you can use the octal code for perfection.

I think we should follow the idea, and let users decide which one to use. The only exception is annotations, as pointed by the first point of Paul's comment, in which we should always use a minus.

PaulWessel:
[...]

PR #1479, opened 2019-09-01
Improved handling of minus and hyphen in text

PaulWessel:

See #1421 for background.  This PR addresses this by replacing hyphens with minus sign _only_ during axes annotations.  For the example in #1421 , we now get this:
isolatin

The implementation is a new PSL function _PSL_settextmode_ that can be used to switch the text machine from replacing hyphens with minus sings (for annotations) or the other way around (for text).  The changes brings 52 test "failures" and I wanted to confirm with you what you thing (please build this PR and run tests).  What I see are these types of failures:

1. Powers of ten (10^-3) now uses the minus symbols whereas in the past those remained hyphens.  This is an improvement; e.g. ex03.sh
2. Lots of basic text and labels that contained hyphens were typeset as minus before (e.g., new_gaps.sh printing GMT option settings and the leading hyphen would become a minus.  This is good. Most of the failures seems to be text with hyphens incorrectly changed to minus.
3. Plotting negative numbers with pstext (e.g., center_bf_g.sh, reduce.sh) now have hyphens. Do we rely on octal codes here (awkward when plotting data tables) or do we check for a single word that is a number and then switch automatically?  Other than -F+z we cannot really know.
4. Sometimes we set an equation and it may contain minus sign, with pstext. We could scan for @~ (switching to symbol font) as an indication of math and use minus.  Thoughts on that one?  Other places we cannot know, e.g., -6x4 labels in geodesy_01.sh.  In some doc scripts we actually use the octal code for a long-dash or minus to force the issue (GMT_-OK.sh).
5. Embedded negative numbers in magnetic rose direction now has minus sign (e.g. rose_map1.sh).
6. Lastly, time axis with primary and secondary annotations (e.g., time_testing_6.sh).  Right now, the labels January-04 uses a minus instead of a hyphen.  This may be a good effect, but it is also easy to turn off minus here since all those numbers are positive, hence we can be reasonably assured that a hyphen is a hyphen.  Shall I force these to be hyphens?

Let me know what you are willing to consider good failures and what you think we need to do further with this PR.

### DIFF
--- a/doc/rst/source/postscriptlight.rst
+++ b/doc/rst/source/postscriptlight.rst
@@ -439,6 +439,14 @@ affect the current state of parameters such as line and fill attributes.
     are interpreted as the cutoff acute angle (in degrees) when mitering
     becomes active.
 
+**long PSL_settextmode** (**struct PSL_CTRL** *\*P*, **long** *mode*)
+
+    Changes between the two modes PSL_TXTMODE_MINUS and PSL_TXTMODE_HYPHEN.
+    When the minus mode is active we assume we are plotting annotation
+    strings with numbers and all hyphens are translated to minus codes
+    which differs based on char sets.  Likewise, in hyphen mode any
+    minus character is typeset as a hyphen in the current char set.
+
 **long PSL_settransparency** (**struct PSL_CTRL** *\*P*, **double**
 *\*transparency*)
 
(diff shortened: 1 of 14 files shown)


seisman:
According to the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601): the separator used between date values (year, month, week, and day) is the hyphen. But now it's minus.

PaulWessel:
OK, so we can change annotation status to hyphen when we have a time-axis then, since there won't be any negative dates or hours (and unlikely negative years).

PaulWessel:
After letting pstext -F+z turn on minus mode we are down to 44 failures.  All fall into two categories: Powers of ten where the originals have hyphens for negative powers, and text or titles or labels including GMT option strings, e.g., -JM6i, where the hyphen in correctly is convert to a minus symbol.  I think this PR is ready for prime time and have removed the WIP

PaulWessel:
This PR closes #1421.

PR #1487, opened 2019-09-03
Settings to allow including an optional ghostscript install

joa-quim:
Note: Artifex also adds a _bitage_ (32 or 64) to their binaries, so it fits perfectly to my ``ConfigUser.cmake`` definitions, but to make more general the BITAGE setting should be put somewhere at a basic level (``ConfigDefault.cmake``?)  so that the install bit at ``share/CMakeLists.txt`` is available to all.
```
	set (BITAGE 32)
	# Detect if we are building a 32 or 64 bits version
	if (CMAKE_SIZEOF_VOID_P EQUAL 8)
		set(BITAGE 64)
	endif ()
```


joa-quim:
I had the same thought but looking at what is installed by the official installer only some files under ''lib'' may (or not) be needed.

seisman:
For users who want to typeset Chinese characters, files in the Resource directory are required.

joa-quim:
The source tree has a dir called ``Resources`` but there is none in the dir created by the windows installer.

seisman:
This is what I have with the gs 9.26 Windows installer.
image

joa-quim:
Ah, I'm still on 9.22 (but that in the installer is built from git source)

PR #1488, opened 2019-09-03
CPT to plot magnetic anomaly maps.

joa-quim:
It's the one used in IPGP for decades


joa-quim:
The ``GMT_App_M_1?`` are complicated and the three pages are full. What you mean _do it in the cache_

joa-quim:
And speaking of new CPT's. Shouldn't we add the [new rainbow](https://ai.googleblog.com/2019/08/turbo-improved-rainbow-colormap-for.html?m=1) from Google that I mentioned you?

PaulWessel:
There is room for adding a pair of CPTs so if you get the google one too.  However, that one has no hinge and should be normalized to 0-1 in z (see others).  I think the mag has a hinge (zero mag) but no zero entry in the CPT.  It would probably be best to split that -50 to 50 up in to -50 to 0 and 0 to 50 for that reason, and add the HINGE flag.
I can help with App M.

PaulWessel:
OK, what do we call the google cut? google? spectrum?  Once you add it I can go to work on App M.

joa-quim:
Well, they called it _turbo_. It's a stupid name but that's how it's going to be known.
Still trying to see how to reformat it.

PaulWessel:
OK, turbo is fine since it is not already used.

PaulWessel:
Need entry in gmt_cpt_masters.h for turbo.

joa-quim:
Done. I think this turbo is a candidate to a new default CPT (if that is still on the table)

PaulWessel:
So some difficulty with App. We have 71 CPTs in gmt_cpt_masters.h, but one we skip (srtm which is same as dem4 on land only).  So 70 minus the 24 from Fabio = 46.  To split into two pages with two columns that number needs to be divisible by 4.  So we actually need 48.  Are we able to come up with 2 more CPTs.

PaulWessel:
How about cequal [http://soliton.vm.bytemark.co.uk/pub/cpt-city/imagej/tn/cequal.png.index.html] and columbia [http://soliton.vm.bytemark.co.uk/pub/cpt-city/wkp/shadowxfox/index.html]?

joa-quim:
And what about dropping 2? We already have more than enough generic ones.

PaulWessel:
Would give backwards compatibility headaches.  Better to add two.  OK, could remove two from the table but leave in share - that is backwards compatible since the gmt_cpt_masters.h is only used for listing (and by App M).  What to remove?

PaulWessel:
Maybe seafloor and wysiwyg?

joa-quim:
wysiwyg is a good candidate. And the _acton_? We only break some compat if remove old ones. New ones were technically never really released.

PaulWessel:
acton is part of Fabio so does not help with the 24 problem.

PaulWessel:
I think seafloor is pretty useless and there is ibcao, bathy, abyss already.

joa-quim:
Yes, we have several seaflors. Panoply seems pretty useless too.

PaulWessel:
I'll remove seafloor and wysiwyg.  Remko loves panoply...

seisman:
Why do we have to remove two CPT? Can we list all 70 cpts in one long paper? Or can we have 24, 22, and 24 cpts in three pages?

PaulWessel:
OK, done and updated M 1b PS.  Now for discussion of turbo vs rainbow.  I would think we agree that turbo is a much improved rainbow without being washed out as in Fabio, no?  Maybe @seisman might agree as well?

PaulWessel:
Words of wisdom from @seisman. We don't have to make 3 separate M plots.  We could keep the Fabio stuff separate and then do everything else as a longer page.  We are no longer dealing with page plotting.  I will see what I can do - the App M script is messy...

seisman:
I once tried to rewrite the three App M scripts with subplot. It's more readable but much slower.

PaulWessel:
I got things to fit as @seisman suggested with 24, 22, 24 on three pages.
Opinions on making turbo the new rainbow default CPT in GMT 6?

PaulWessel:
Changing to turbo will probably affect some of our PS origs.

seisman:
I have no objection.

joa-quim:
I'm very favorable. It's not only the perceptual thing but also that it drops the violet, which always felt wrong on topography maps.

PaulWessel:
Running all tests but having trouble:

psxy [ERROR]: Unable to down file http://oceania.generic-mapping-tools.org/gmt_md5_server.txt

Not sure why this is happening but I have seen it before.  So lots of failures obviously if we cannot get the cache files.

PaulWessel:
We are finding out that possibly due to a UH firewall change, nodes INSIDE the UH network cannot access oceania but nodes on the outside can. The SOEST IT staff says it has nothing to do with the SOEST firewall, but there seems to be some infinite loop being set up when we try within UH.

seisman:
I'm back now. Should I push the updated PS files to this branch?

PaulWessel:
Yes please, I am unable to access the cache since I am inside the UH network...

PaulWessel:
So if tests pass for you then I guess we can merge this PR?

PaulWessel:
Implemented.

PR #1490, opened 2019-09-03
Synchronize cmake settings with admin/ConfigReleaseBuild.cmake

seisman:
In anticipation to the possible nightly dev releases (macOS and Windows), I synchronized the settings in admin/ConfigReleaseBuild.cmake to the CI cmake settings. I also changed the CMAKE_BUILD_TYPE to _RelWithDebInfo_, as it's said to be better than _Release_.

### DIFF
--- a/admin/ConfigReleaseBuild.cmake
+++ b/admin/ConfigReleaseBuild.cmake
@@ -4,7 +4,7 @@
 # your environment and pointing to the latest releases.
 #
 #-------------------------------------------------------------
-set (CMAKE_BUILD_TYPE relwithdebinfo)
+set (CMAKE_BUILD_TYPE Release)
 set (CMAKE_INSTALL_PREFIX "gmt-${GMT_PACKAGE_VERSION}")
 set (GSHHG_ROOT "$ENV{GMT_GSHHG_SOURCE}")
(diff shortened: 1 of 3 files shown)


joa-quim:
Why is _RelWithDebInfo_ better? It will ship with the .pdb (on Win) that are only useful to people that know how to debug.

seisman:
As I understand it, *RelWithDebInfo* mode produces fully optimized code, the same as _Release_ mode, but also allows debugging. So it provides more possibilities. I think that's why @PaulWessel uses *RelWithDebInfo* when building the macOS bundle.

PaulWessel:
The real reason is that Florian @ldldr told me to...

joa-quim:
That debug info is very limited. Because the code is optimized like in _Release_ most of variables are not available. From my experience it only serves to know where shit happened but next I had to use a true Debug build.

PaulWessel:
True.  When any bugs or crashes are reported, I will need to debug using the full debug, and if it is only a problem that happens with the optimized executables then we deal with valgrind and print statements to get closer to the issue.  I never try to debug with those compiler settings...

seisman:
OK. I've changed the CMAKE_BUILD_TYPE to Release on both macOS and Windows.

Issue #1496, opened 2019-09-04
movie generates wrong scripts if the main script ends without a newline


seisman:
An error originally reported by @liuzhumei

![image](https://user-images.githubusercontent.com/3974108/64222292-9cf4ff80-ce9d-11e9-8ff3-8a2a9faf2097.png)

Note the `endREM` at line 16. Maybe its because the globe.bat script doesn't end with a newline?

seisman:
Same error for bash script without an EOF newline:
```
#!/usr/bin/env bash
# Main frame loop script
export GMT_SESSION_NAME=$$
# Include static and frame-specific parameters
source movie_init.sh
source movie_params_$1.sh
mkdir ${MOVIE_NAME}
cd ${MOVIE_NAME}
gmt begin
# 	Set output PNG name and plot conversion parameters
	gmt figure ${MOVIE_NAME} png E${MOVIE_DPU},A+n+r
	gmt set PS_MEDIA 6ix6i DIR_DATA /Users/seisman/Desktop/workspace,/Users/seisman/Desktop/workspace/globe
   gmt coast -Rg -JG${MOVIE_FRAME}/20/${MOVIE_WIDTH} -Gmaroon -Sturquoise -Bg -X0 -Y0
gmt end# Move PNG file up to parent directory and cd up one level
mv -f ${MOVIE_NAME}.png ..
cd ..
# Remove frame directory and frame parameter file
rm -rf ${MOVIE_NAME}
rm -f movie_params_$1.sh

```

PaulWessel:
Just so I understand, this is the final last line in the input script?

seisman:
Yes, the content of the main script is:
```
gmt begin
   gmt coast -Rg -JG${MOVIE_FRAME}/20/${MOVIE_WIDTH} -Gmaroon -Sturquoise -Bg -X0 -Y0
gmt end
```
The newline character after **gmt end** make a big difference.

seisman:
Closed by #1497.

PR #1497, opened 2019-09-04
Watch is input scripts are missing a newline in the last line

PaulWessel:
If the user is sloppy and doesn't hit return on the last script line then we need to compensate.  Closes issue #1496.


seisman:
It works but is a little fragile if there are more than one space between **end** and **show**, e.g.
```
gmt end       show
```

PaulWessel:
Latest PR deals with the words separately.

PR #1512, opened 2019-09-05
Update install instructions

seisman:
- Windows installers also provide ghostscript
- Windows users need to install graphicsmagick separately
- Update install instruction for macOS bundle.


joa-quim:
So we are set that the Windows installer will ship ghostscript? What I did was experimental work.
IMO we should do it. I only have to add to the description in the installer that the binaries are compiled from Artifex unmodified source code.
Also found that QGis (silently) ships with ghostscript.

PR #1521, opened 2019-09-05
Add option to force docs to read from the GMT server

### LINKED ISSUE #1518: Allow gmt docs to open the remote documentation
`gmt docs coast` opens the local documentation if it exists (e.g. `file:///Users/seisman/local/GMT/share/doc/html/coast.html`). With **-Q**, `gmt docs -Q coast` prints the URL of the local documentation.

Since we're continuously fixing and improving the docs, users may want to read the remote documentation (i.e. docs.generic-mapping-tools.org) even though they have one in their local computer. 

Thus, we may add an option or a modifier to allow opening the remote documentation and printing the remote URL.

PaulWessel:
The new **-S** option overrides the default to use the local doc files.  Closes #1518.

### DIFF
--- a/doc/rst/source/docs.rst_
+++ b/doc/rst/source/docs.rst_
@@ -9,7 +9,7 @@ Synopsis
 
 .. include:: common_SYN_OPTs.rst_
 
-**gmt docs** [ **-Q** ] [ |SYN_OPT-V| ] *module-name* [*-option*]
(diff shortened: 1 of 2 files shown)


seisman:
Nice work! Change it to WIP?

PaulWessel:
To be merged after rc4 is released.

PR #1522, opened 2019-09-05
Add share/proj to the list of custom dirs that GDAL may consult

joa-quim:
I used this to make the Win installers but we can merge this after RC4.


PaulWessel:
Sorry for holding things up - I was/am teaching...
WHo will squash it?

seisman:
IMO you can squash and merge it into 6.0 branch, and rebuild the tarballs and macOS bundle again, so that the five packages (2 source packages, 2 Win installers and 1 macOS bundle) are all built from the same codes.

PaulWessel:
OK, sqauashing away in a second.

PR #1530, opened 2019-09-06
Use https for URL to GMT data server

PaulWessel:
Run the encrypted service.  Please test from outside hawaii.edu to make sure there are no side-effects since I cannot do it.

### DIFF
--- a/cmake/ConfigDefault.cmake
+++ b/cmake/ConfigDefault.cmake
@@ -121,7 +121,7 @@ endif (NOT DEFINED GMT_RELEASE_PREFIX)
 
 # Default location of remote data server
 if (NOT DEFINED GMT_DATA_SERVER)
-	set (GMT_DATA_SERVER "http://oceania.generic-mapping-tools.org")
+	set (GMT_DATA_SERVER "https://oceania.generic-mapping-tools.org")
 endif (NOT DEFINED GMT_DATA_SERVER)
 
 # You can set the build configuration type as a command-line argument to 'cmake' using -DCMAKE_BUILD_TYPE:STRING=Debug for example.
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -380,7 +380,7 @@ fonts can be found in the :doc:`gmt` man page.
 .. _GMT_DATA_SERVER:
 
 **GMT_DATA_SERVER**
-    Address of the data directory on the remote server [The SOEST server].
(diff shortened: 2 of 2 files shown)


seisman:
Does it work for you? This is what I got:
```
$ gmt which -Gl @AFR.nc
gmtwhich [ERROR]: Unable to down file https://oceania.generic-mapping-tools.org/gmt_md5_server.txt
gmtwhich [ERROR]: Libcurl Error: Couldn't connect to server
gmtwhich [ERROR]: Libcurl Error: Couldn't connect to server
gmtwhich [ERROR]: File AFR.nc not found!
```

PaulWessel:
I cannot test until I go home.  The libcurl function may need an extra setting to do this, perhaps?

seisman:
I don't think it's libcurl settings. You can try visit the URL via you browser. http://oceania.generic-mapping-tools.org is redirected to http://www.soest.hawaii.edu/gmt/data/, with a 403 forbidden code. But https://oceania.generic-mapping-tools.org cannot be reached. Maybe https isn't enabled by the DNS provider?

PaulWessel:
Try in a few minutes.  The DNS services forwards

oceania.generic-mapping-tools.org to https://www.soest.hawaii.edu/gmt/data

now (it was http).  So not sure what happens if you try http.  I guess rc4 will have http so it may break?  But from the entry above there is no left-side http or https I can set.

seisman:
Not working for me now. I'll test it later. I believe it's possible to have both http and https. Just like https://www.generic-mapping-tools.org and http://www.generic-mapping-tools.org.

seisman:
I still cannot visit the https URL now.

PaulWessel:
I just now changed it back to http:

PaulWessel:
So the GMT server supports either one of course (I remove the local rays_11.txt between each try):

```
gmt set GMT_DATA_SERVER http://www.soest.hawaii.edu/gmt/data
gmt which -Gc @rays_11.txt -Vd
gmt set GMT_DATA_SERVER https://www.soest.hawaii.edu/gmt/data
gmt which -Gc @rays_11.txt -Vd
```
But doing the same with oceania fails, only http works.

PaulWessel:
Any thoughts on this @leouieda?

PaulWessel:
Closing this since does not work for https.

leouieda:
@PaulWessel I think it's a good idea to do https whenever possible. I'm sure SOEST IT can setup SSL for the data server.

PR #1539, opened 2019-09-08
Use a functinon to check if an executable exists via popen

PaulWessel:
We had several places where we did a special popen check, but I have consolidated those actions in a new gmt_check_executable () function that handles the common tasks.  It is being called by psconvert.c and movie.c

### DIFF
--- a/src/gmt_prototypes.h
+++ b/src/gmt_prototypes.h
@@ -333,7 +333,7 @@ EXTERN_MSC bool gmt_this_alloc_level (struct GMT_CTRL *GMT, unsigned int alloc_l
 
 /* gmt_support.c: */
 
-EXTERN_MSC bool gmt_check_executable (struct GMT_CTRL *GMT, char *in_cmd, char *pattern, char *text);
+EXTERN_MSC bool gmt_check_executable (struct GMT_CTRL *GMT, char *program, char *arg, char *pattern, char *text);
 EXTERN_MSC void gmt_filename_set (char *name);
 EXTERN_MSC void gmt_filename_get (char *name);
 EXTERN_MSC bool gmt_no_pstext_input (struct GMTAPI_CTRL *API, char *arg);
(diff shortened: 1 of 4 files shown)


seisman:
The Windows build fails.

PaulWessel:
So it is because of directory with spaces in the name, Program Files.  I guess

	sprintf (cmd, "%s --version", Ctrl->G.file);
	if (gmt_check_executable (GMT, cmd, NULL, cmd)) {	/* Found GhostScript */

needs to be

	sprintf (cmd, "'%s' --version", Ctrl->G.file);

to ensure it is seen as a single name?

PaulWessel:
@seisman I added single quotes around a program with spaces, but I don't see that reflected in the error message? Any suggestions?

PaulWessel:
I don't get it.  It now says (I forced the first message to see what is passed to popen)

> psconvert [ERROR]: gmt_check_executable: Pass to popen (in []):
['C:\Program Files\gs\gs9.26\bin\gswin64c.exe' --version > NUL]
> The filename, directory name, or volume label syntax is incorrect.
> psconvert [ERROR]: Cannot execute GhostScript ("C:\Program Files\gs\gs9.26\bin\gswin64c.exe").

I don't see the problem.

seisman:
Always use double quotes?
image

PaulWessel:
I think I have tried every combination, but perhaps now.  I will modify to do double quotes.

PaulWessel:
Seems like nothing works, single double nothing.

PaulWessel:
If you have Windows anywhere, can you paste in that command, with our without quotes etc and see if it should run?

seisman:
image

It seems the command `"C:\Program Files\gs\gs9.26\bin\gswin64c.exe" --version > NUL` outputs stdout (gs version) to NUL. If that's true, how do you parse the version string? Maybe you should use `2> NUL`?

PaulWessel:
Please try with a command that fails (e.g. gs7897987) vs one that works.  So it writes the version to > NUL but error needs 2> NUL?

seisman:
image

PaulWessel:
OK, pushed 2> NUL

PaulWessel:
Worked apparently, so OK to merge?

joa-quim:
The function ``ghostbuster()`` already wrapped the G->file under double quotes. Is it now double wrapped?

PR #1559, opened 2019-09-13
Add the DOI for the source code release on zenodo

PaulWessel:
We have reserved this DOI and will publish it when we release 6.0.0.  It is displayed when the user runs

`gmt --show-doi
`


seisman:
Use https://doi.org/xxxxx ?

PaulWessel:
Yes of course

PR #1569, opened 2019-09-16
Build and deploy documentations for 6.0 and master branch

seisman:


seisman:
As we're continually improving the new documentations, we don't want to wait for one-day to see the changes.

Issue #1583, opened 2019-09-18
Use begin and end in the tutorial


seisman:
In the tutorial, we should always use 
```
gmt begin map
gmt ...
gmt end show
```

instead of the one-liner commands.

PaulWessel:
I agree that it is just plain confusing to switch between one-lines and the regular modern mode syntax.  One you know your way around GMT you can cheat and use those one-liners for simple stuff.

The other problem with the tutorial (I just got a new graduate student and he is learning GMT by working through the tutorial) is that we are mixing some classic mode calls, e.g., how to make a CPT, such as

`gmt makecpt -Crainbow -T1000/5000/500 -Z > topo.cpt
`

but then we may have a modern mode script with

`gmt grdimage @tut_relief.nc -JM6i -B -Ctopo.cpt
`

and my student simply pasted the makecpt command into the gmt begin/end script but of course in modern mode you need -H for this, or even better, you just do

```
gmt makecpt -Crainbow -T1000/5000/500 -Z
gmt grdimage @tut_relief.nc -JM6i -B
```

So separating them out can also be confusing.

PaulWessel:
I will work on this now.

seisman:
Closed by #1586

PaulWessel:
What about man pages? We have lots of one-liners in the man pages and those examples are not particularly helpful for new users trying to build a modern mode script by hunting for examples.  Copying

`gmt colorbar -Ct.cpt -Dx8c/1c+w12c/0.5c+jTC+h -Bxaf+l"topography" -By+lkm -pdf map
`

into a gmt begin myplot pdf modern mode script is disastrous.  I know we wanted 1-liners because they are short, but are we setting people up for trouble?

seisman:
Yes, as the one-liner syntax is introduced in an appendix of the long cookbook, most users may never realize the existence of the one-liner syntax.

PaulWessel:
There are also lots of man pages examples that has one-liners that cannot work by themselves, e.g., clip sets a clip path but we have no example on the man page that users can copy and paste to see the effect.  There must be lots of those cases across most man pages.

joa-quim:
When one liners are clearly incorrect (clip, makecpt, etc), ofc I agree that they should be replaced by a
```
gmt begin
   makecpt ...
   ...
gmt end
```
[...]

PR #1586, opened 2019-09-18
Let tutorial only used gmt begin/end examples

PaulWessel:
Introducing the one-liner shorthand is not very smart in a tutorial trying to teach new users how to write GMT modern mode scripts.  Also, many examples had older modern syntax that predated the current CPT scheme.  A few original PS files had to be updated.

Still need to teach grdview to use curent CPT without needing **-C**.


PaulWessel:
Closes #1583.

PR #1589, opened 2019-09-19
Ensure proper interpolation in grdinterpolate

PaulWessel:
When the output time fell exactly on a layer time our search to limit the number of layers resulting in one layer - yet the spline interpolation requires at least 2 layers.
Also, if the time coordinate is actually time, we use -f2T but need to reset to float before writing grids (since the data written are not time)


seisman:
@PaulWessel You can merge this PR if you want. Just ignore the travis-ci check, as we no longer use travis. Don't understand why it's here.

PR #1602, opened 2019-09-20
Update azimuth test with knowledge of back-azimuth

PaulWessel:
This failure was a user error (using forward instead of back azimuth at the wrong time).  Updated the script and added PS original.


test/mapproject/azimuth.sh
```diff
@@ -13,9 +13,9 @@ gmt begin azimuth ps
   lon=156.293957716
```
seisman: `omega=1.08` isn't used. Please also remove it.

Issue #1388, opened 2019-08-15
an extra line in the national boundaries


bridgeW:
**Description of the problem**

an extra line plotted in the solar results. There was a grey line tangential to the west edge of the Eger VI projection map  when I used the SOLAR module to plot the day-night terminator.

**Full script that generated the error**

```
export GMT_SESSION_NAME=$$
#!/usr/bin/env bash
gmt begin solar pdf
DATEstr=2019-04-24T12:15:30

gmt coast -Rd -JKs100/10i -Dl -A5000 -W0.5p -N1/0.5p,gray -S175/210/255 -Bafg --MAP_FRAME_TYPE=plain
gmt solar -Td+d${DATEstr} -Gnavy@95
gmt solar -Tc+d${DATEstr} -Gnavy@85
gmt solar -Tn+d${DATEstr} -Gnavy@80
gmt solar -Ta+d${DATEstr} -Gnavy@80

gmt solar -I+d${DATEstr} -C | gmt plot -Sc0.5c -Gyellow -t40
echo 90 28 | gmt plot -SE-10c -Wred

gmt end
```

**Full error message**

No error message comes out.

**System information**

* Operating system: MacOS 10.14.4
* Version of GMT: v6.0.0rc3

seisman:
A minimum script to reproduce the bug:
```
gmt begin solar pdf
gmt coast -Rd -JKs100/10i -Dl -A5000 -W0.5p -N1/0.5p,gray -S175/210/255 -Bafg
gmt end show
```
It looks like a bug of GSHHG data or **coast**.

PaulWessel:
It comes from the political boundaries (-N1) and affects all resolutions.  Will try to dig in later today or during workation.

joa-quim:
Is the ``export GMT_SESSION_NAME=$$`` line now mandatory? Without it I get

```
C:\v>C:\programs\WinGit\bin\bash.exe t.sh
gmt.exe [ERROR]: Shared GMT module not found: coast

ERROR: No module named coast was found.  This could mean one of four cases:
  1. There actually is no such module; please check your spelling.
  2. You used a modern mode module name while running in GMT classic mode.
  3. Module exists in the GMT supplemental library, but the library could not be found.
  4. Module exists in a GMT custom library, but the library could not be found.
Shared libraries must be in standard system paths or set via environmental parameter %%PATH%%.

gmt.exe [ERROR]: Not available in classic mode
```

PaulWessel:
[...]

PR #1611, opened 2019-09-22
Make sure clipping is on if line extension is applied to pscoast -N

PaulWessel:
See issue #1388 for context.  The problem was that we implemented the feature that allows fat lines to properly be clipped at map boundaries (by extending the line outside the map and turn clipping on) but for this case (-N) the clipping was not activated, allow the extension line to be seen.  Now the decision to clip matches the code that allows line extension plus it applies to **-W -N** and **-I** as long as **-M** is not issued.


PaulWessel:
Closes #1388.

PR #1612, opened 2019-09-22
Add -W to psclip and give working example in man page

PaulWessel:
I have added a working copy/paste example to the clip/psclip man pages and in the process I added a basic **-W**_pen_ option to psclip.c so we can visualize the clip path as well.  Also added the example as a test for psclip.

### DIFF
--- a/doc/rst/source/clip.rst
+++ b/doc/rst/source/clip.rst
@@ -49,10 +49,10 @@ being partly visible or missing altogether, try
 
     gmt begin clip
       gmt clip -R0/6/0/6 -Jx2.5c -W1p,blue << EOF
-0 0
-5 1
-5 5
-EOF
+    0 0
+    5 1
+    5 5
+    EOF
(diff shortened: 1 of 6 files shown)


doc/rst/source/clip.rst
```diff
@@ -41,18 +42,23 @@ Synopsis
...
-    gmt clip my_region.xy -R0/40/0/40 -Jm0.3i
-
-To deactivate the clipping in an existing plotfile, run:
-
-   ::
-
-    gmt clip -C
+    gmt begin clip
+      gmt clip -R0/6/0/6 -Jx2.5c -W1p,blue << EOF
+0 0
```
seisman: I think the indentations here are incorrect.

PaulWessel: You mean you want space in front of those coordinates?

seisman: Yes, the coordinates should have the same indentations as `gmt begin`. 

This is now what it looks like:
image

PR #1621, opened 2019-09-22
Prevent gmt show end from displaying in tests

PaulWessel:
By setting export **GMT_END_SHOW**=off in doc/scripts/gmtest.in we turn off the automatic display of plots.  There are no such examples, animations, or tests.


PaulWessel:
Perhaps this should be documented in the one-liner.rst file?

PaulWessel:
Closes #1583.

seisman:
@PaulWessel Can you allow **GMT_END_SHOW** to disable auto-show in "gmt end show"?

PR #1625, opened 2019-09-23
Doc improvements

PaulWessel:
Some improvements to a few rest files (working examples in grcontour, starting on the same in grdimage), plus placing c2308.txt in cache so eventually doc examples for x2sys can use @c2308.txt as now used in a test.


doc/rst/source/grdcontour.rst
```diff
@@ -50,48 +50,54 @@ Examples
...
+To contour the remote file AK_gulf_grav.nc every 25 mGal on a Mercator map at
+0.5 inch/degree, annotate every 50 mGal (using font size = 10p), using 1
+degree tick marks, and draw 30 minute gridlines:
 
    ::
 
-    gmt grdcontour hawaii_grav.nc -Jm0.5i -C25 -A50+f10p -B1g30m -pdf hawaii_grav
+    gmt grdcontour @AK_gulf_grav.nc -JM16c -C25 -A50+f10p -B -pdf alaska_grav1
 
 To do the same map but only draw the 50 and 150 and annotate the 100 contour:
```
seisman: FYI, you can have a code block by ending a paragraph with `::`. It's a little simpler. 

For example, you can use 
```
To do the same map but only draw the 50 and 150 and annotate the 100 contour::

    gmt grdcontour @AK_gulf_grav.nc -JM16c -C50,150 -A100,+f10p -B -pdf alaska_grav2
```
instead of 
```
To do the same map but only draw the 50 and 150 and annotate the 100 contour:

   ::

    gmt grdcontour @AK_gulf_grav.nc -JM16c -C50,150 -A100,+f10p -B -pdf alaska_grav2
```

PR #1632, opened 2019-09-23
Increase hash character array to 128.

PaulWessel:
The new SHA256 takes up 64 characters so neds to be > 64 anyway. aking more room for the future.


seisman:
master or 6.0 branch?

PR #1634, opened 2019-09-23
Enable x2sys track files to be remote

PaulWessel:

Commands like x2sys_cross @2308 will now work by downloading the remote file to the current directory first.  Thus, x2sys_06.sh can use a remote file and work.


seisman:
Is this the same as #1630.

PaulWessel:
Yes it is , oops busy morning.  I guess delete?

PR #1646, opened 2019-09-25
Handle general URL files

PaulWessel:
URLs need to be checked if they exist locally.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -15219,39 +15219,11 @@ GMT_LOCAL bool check_if_we_must_download (struct GMT_CTRL *GMT, const char *file
 	unsigned int pos = gmtlib_get_pos_of_filename (file);	/* Find start of filename */
 	if (pos == 1) return true;  /* Must always check since file on server might have changed and should be refreshed first */
 	else if (kind == GMT_URL_QUERY)
-		return true;	/* A query can never exist locally */
+		return true;	/* A query can never exist locally so must follow the URL */
 	else if (!gmt_access (GMT, &file[pos], F_OK))
-		return false;	/* File exists already so no need to download */
+		return false;	/* Regular file exists already so no need to download */
 	return true;	/* File not found */
 }
-
-#if 0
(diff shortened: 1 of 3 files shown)


PaulWessel:
This PR lets URL files be downloaded to current directory, but the failure in mfilter.sh has to do with header records that GMT cannot recognized.  I dont understand how this worked earlier.

PaulWessel:
A question would be if GMT should know that records starting with quotes must be headers.  Currently only records starting with # are headers unless **-hi**_n_ is used.

PaulWessel:
What do you think @joa-quim ?

seisman:
The command below only plots one circle. Possibly all invalid records are ignored.
```
gmt plot -R0/10/0/10 -JX10c -Baf -Sc0.5c -Gred -Vd -pdf map << EOF
" 1 1
" 2 2
' 3 3
$ 4 4
5 5
( 6 6
EOF
```

PaulWessel:
Problem is there is no **-f0T** in filter1d call so it does not know that those 1958-* text strings are time stamps when it finally gets there.  It could learn that by examining one of those records but it never gets there since the header record is scanned as the first data record and has no time data to learn from.

PaulWessel:
I am hesitant to skip records starting with " since there are gmt modules that can read just trailing text, so this would skip such record entirely.  If we know there must be numerical portions then we could do that.  Clearly, filter1d is such a case but the decision needs to be known deep in the io function.

joa-quim:
Hmm, dangerous ground. I agree that we should not consider comments lines starting with ". Specially not at this point of time-line. If CSV was a reliable extension/format/whatever then we could dare more but unfortunately global ignorance reign and it so often is used just to mean a text file. Current case could perhaps  be solved if we could analyze a couple of lines at the END of file to know the number of columns (using ',' as separator) and then restart from beginning and skip all lines that have a different number of columns. But, as I said above, this is guess-work on dangerous  land.

PaulWessel:
Yes.  And reading the end is tricky with pipes.  Would have to save to temp file, etc.

PR #1648, opened 2019-09-25
Check if feature/pen syntax is wrong

PaulWessel:
pscoast has several options like **-N**_feature_[/_pen_] but it would not detect and process wrong syntax like -N1,1p (the pen is simplly ignored).  This PR adds checks for such incorrect syntax which yields an error message.  One of the examples in the G-cubed paper made that mistake. I have asked Wiley to fix it before publication.

### DIFF
--- a/src/pscoast.c
+++ b/src/pscoast.c
@@ -363,6 +363,11 @@ GMT_LOCAL int parse (struct GMT_CTRL *GMT, struct PSCOAST_CTRL *Ctrl, struct GMT
 					n_errors++;
 					continue;
 				}
+				if (strchr (opt->arg, '/') == NULL && strchr (opt->arg, ',')) {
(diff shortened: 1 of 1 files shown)


src/pscoast.c
```diff
@@ -463,6 +473,11 @@ GMT_LOCAL int parse (struct GMT_CTRL *GMT, struct PSCOAST_CTRL *Ctrl, struct GMT
 				break;
 			case 'W':
 				Ctrl->W.active = true;	/* Want to draw shorelines */
+				if (strchr (opt->arg, '/') == NULL && strchr (opt->arg, ',')) {
```
seisman: Some tests fail due to a bug introduced here. The syntax of -W option is **-W**[_level_/]_pen_.

PaulWessel: Correct, too much copy/paste on this one.

PR #1658, opened 2019-09-27
Update building instructions

seisman:
1. Windows users also need to add GDAL's tools path to PATH
2. Describe how to set PATH for Linux/macOS users

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -246,15 +246,18 @@ After installing vcpkg, you can install the GMT dependency libraries with (it ma
     # hook up user-wide integration (note: requires admin on first use)
     vcpkg integrate install
 
-After installing these dependency libraries, you need to add the bin path
-(i.e. `C:\vcpkg\installed\x64-windows\bin`) to the system PATH,
-to allow executables find the DLL shared libraries.
+After installing these dependency libraries, you also need to add
+vcpkg's bin path (i.e. `C:\vcpkg\installed\x64-windows\bin`) and
+GDAL's bin path (i.e. `C:\vcpkg\installed\x64-windows\tools\gdal`),
+to the system environmental variable `PATH`,
+so that GMT executables can find the DLL shared libraries and
+tools (`gdal_translate` and `ogr2ogr`) provided by GDAL.
 
 ## Getting GMT source codes
 
 The latest stable release of the GMT source codes (filename: gmt-x.x.x-src.tar.gz)
 are available from the [GMT website](https://www.generic-mapping-tools.org) and
-[GitHub Release Page](https://github.com/GenericMappingTools/gmt/releases)
+[GitHub Release Page](https://github.com/GenericMappingTools/gmt/releases).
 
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -246,15 +246,18 @@ After installing vcpkg, you can install the GMT dependency libraries with (it ma
...
 
-After installing these dependency libraries, you need to add the bin path
-(i.e. `C:\vcpkg\installed\x64-windows\bin`) to the system PATH,
-to allow executables find the DLL shared libraries.
+After installing these dependency libraries, you also need to add
+vcpkg's bin path (i.e. `C:\vcpkg\installed\x64-windows\bin`) and
+GDAL's bin path (i.e. `C:\vcpkg\installed\x64-windows\tools\gdal`),
+to the system environmental variable `PATH`,
+so that GMT executables can find the DLL shared libraries and
+tools (`gdal_translate` and `ogr2ogr`) provided by GDAL.
```
joa-quim: Are you sure about this? I only build GMT with vcpkg once and don't remember the details anymore but I think that at the end one can make a package with everything. Otherwise is not only GDAL but all of it's dependencies.

joa-quim:
Well, I made a comment.

seisman:
@joa-quim I can't see any comments.

joa-quim:
I'm in a train but I can see it
Capture

joa-quim:
And it's my fault that I never finished that work, but what we really need is a formula to build GMT directly with vcpkg.

BUILDING.md
seisman: vcpkg installs all the DLL libraries in its bin path (i.e. `C:\vcpkg\installed\x64-windows\bin`). I think the `vcpkg integrate install` command does the tricks so that we can build any packages without adding the bin path to PATH. 

But the problem is, when we run the gm executable, it won't know where to find the DLL libraries. When building GMT, we also build gshhg_version and run it to extract the version information from GSHHG files, so we have to add the vcpkg's bin path to PATH before building GMT.

It's the same for GDAL's bin path (C:\vcpkg\installed\x64-windows\tools\gdal), gmt can't call gdal_translate and ogr2ogr if we don't add it to PATH.

seisman: My changes here may be a little misleading. 

What I meant is: the vcpkg's bin path is needed to allow GMT executables to find the DLL libraries, and the GDAL's bin path is needed to allow GMT find gdal_translate and ogr2ogr. 

Does it make more sense to you?

joa-quim: Can't continue this discussion right now. Train is arriving and I'll go dinner somewhere.
I'm not sure of nothing but I got the impression that one would build with vcpkg and then create a package to install in a tidier place instead of using directly the vcpkg directory source/build tree.

seisman: I'm pretty sure when I run `cmake --build . --target install`, not all the DLL libraries are copied to the GMT's bin path.

Issue #1644, opened 2019-09-25
Three failing tests


seisman:
```
1/3 Test #238: filter1d/mfilter.sh ..............***Failed    0.22 sec
filter1d [ERROR]: Error for input file: No such file (weekly_in_situ_co2_mlo.csv)
gmt [ERROR]: Error parsing filter1d options
ERROR: gmtest:58
memtrack errors: 0
exit status: 1

2/3 Test #395: grdimage/url_map.sh ..............***Failed    0.24 sec
grdimage [ERROR]: Error for input file: No such file (untooned_jessicarabbit.jpg)
grdimage [ERROR]: Syntax error: Must specify one (or three) input file(s)
ERROR: gmtest:58
memtrack errors: 0
exit status: 1

3/3 Test #830: x2sys/x2sys_06.sh ................***Failed    0.25 sec
x2sys_cross [ERROR]: x2sys_read_file : Cannot find track c2308.txt
(null): Error from fclose [@c2308]
ERROR: gmtest:58
memtrack errors: 0
exit status: 1

```

PaulWessel:
Yep, working on it.

PaulWessel:
Hm, I had not committed a change to deal with this.  Testing my branch works in getting the file, but now filter1d complains about no data columns.  Looking at the file it has lots of header records that I dont think we can recognize:
"-------------------------------------------------------------------------------------------"
" Atmospheric CO2 concentrations (ppm) derived from in situ air measurements                "
" at Mauna Loa, Observatory, Hawaii: Latitude 19.5°N Longitude 155.6°W Elevation 3397m      "
...

I don't know if they changed their format recently?  GMT expects headers to start with #.

seisman:
I can't get the files with the latest 6.0 branch. Checkout the codes before commit https://github.com/GenericMappingTools/gmt/commit/8cf32eeeaa4175f69834b67231b8427bebab54f5, then I can get the files.

PaulWessel:
Now, oddly, in my cache I got that file this morning, and it has these headers:
#-------------------------------------------------------------------------------------------#
# Atmospheric CO2 concentrations (ppm) derived from in situ air measurements                #
# at Mauna Loa, Observatory, Hawaii: Latitude 19.5<C2><B0>N Longitude 155.6<C2><B0>W Elevation 3397m      #
#                                                                                           #
# Source: R. F. Keeling, S. J. Walker, S. C. Piper and A. F. Bollenbacher                   #

[...]

PR #1663, opened 2019-09-28
Let x2sys use cache as well

PaulWessel:
Any cached @files should behave like cached files everywhere else in GMT.  Now downloaded to the user's cache.  The only thing special here is that we may have to add a file extension to the cache name before checking, e.g., a file given as @c2308 will need to become @c2308.txt before checking.


seisman:
Closing #1644.

PR #1664, opened 2019-09-28
docs: Add rst files to hidden toctrees

seisman:
To suppress the annoying "document isn't included in any toctree" warnings.


seisman:
There are still some warnings left. I'll try how to fix them later.

PR #1674, opened 2019-09-29
docs: Remove isogmt

seisman:
isogmt is a Bash script for GMT isolation mode, which is no longer
needed in GMT 6 modern mode.


seisman:
OK. Closing.

PR #1678, opened 2019-09-30
docs: Cleanup the messy explain_-bi_full.rst_

seisman:
Remove the -bi example written in raw HTML codes.
image


seisman:
Pop-up is a good idea, but it seems not working anymore in the new theme. OK, let's keep them and see if there is a fix.

joa-quim:
psxy -L has the same issue. The pop-up was very nice because it could focus one the particular option only.

PR #1680, opened 2019-09-30
Update install instruction via conda

seisman:
The updated instruction doesn't work now, but it will work after the official 6.0.0 is released. 

**DO NOT MERGE IT NOW.**


seisman:
ghostscript is automatically installed when running `conda install gmt -c conda-forge`.

PaulWessel:
That begs the question if the same should not happen automatically for the other run-time requirements?

seisman:
Currently not, only ghostscript is installed automatically. I'll see if I can add ffmpeg and graphicsmagick as the run-time dependencies.

PR #1684, opened 2019-09-30
CI: Use new GMT data server URL and new hash table

seisman:


seisman:
curl raises an error on Linux and Windows:
```
$ curl https://oceania.generic-mapping-tools.org/gmt_hash_server.txt
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

PaulWessel:
Hm, when we do this via libcurl we specifically set some flags.  Since you are calling curl as a command there are probably an option that bypasses this issue.

seisman:
`curl -k` works.

PR #1685, opened 2019-09-30
CI: Replace md5sum with sha256sum

seisman:


seisman:
On Windows, we can use `CertUtil -hashfile gmt-xxx-win64.exe sha256`.

Issue #1702, opened 2019-10-01
Bands (+b) are broken and no 'image' module


joa-quim:
This
```
gmt image @needle.jpg=gd+b2,1,0 -Dx0/0+w7c -P -Y15c -K > lixo.ps
```
and this
```
gmt image @needle.jpg=gd+b0,1,2 -Dx0/0+w7c -P -Y15c -K > lixo.ps
```
produce the same output (**+b** seems to be ignored). 

and ``gmt image`` not found
```
gmt image @needle.jpg=gd+b2,1,0 -Dx0/0+w7c -P -Y15c -K > lixo.ps
gmt [ERROR]: Shared GMT module not found: image
```
Trying to see what's going on with -b but will have to leave in a while.

PaulWessel:
But otherwise you are right:

```
gmt image @needle.jpg=gd+b2,1,0 -Dx0/0+w7c -pdf one
gmt image @needle.jpg=gd+b0,1,2 -Dx0/0+w7c -pdf two
```

produces the same plot.

PaulWessel:
That must be in the GDAL read part, no?

joa-quim:
No, it's psimage L445

PaulWessel:
You are right.  So need to add a c[0] = '=' just before L471 I think?

joa-quim:
Not enough. Need to follow track but have only ~20 min before going to airport pick up Alexandra

```
ERROR 5: C:/Users/joaqu/.gmt/cache/needle.jpg: GDALDataset::GetRasterBand(0) - Illegal band #

ERROR 10: Pointer 'hBand' is NULL in 'GDALGetRasterDataType'.

ERROR 10: Pointer 'hBand' is NULL in 'GDALRasterIO'.

psimage [ERROR]: GDALRasterIO failed to open band 2 [err = 3]
```

joa-quim:
> There is no image command in classic mode.

This is too confusing. Classic mode should accept the modern mode names as well.

PaulWessel:
Classic mode has no idea there are modern mode names and neither do classic mode users.  Mixing that up is what would be truly confusin.

joa-quim:
There will be not clear cut division between classic and modern mode users. Classic will be modern too (the inverse will be much more rare) and that's where the forbidding rules will produce confusion.

PaulWessel:
So on further review ... we accept pscoast and psbasemap etc in modern mode so not sure why image is not accepted...

[...]

PR #1711, opened 2019-10-01
Truly test selecting a different band ordering (would have catch #1702)

joa-quim:
This will produce a failing test, The orig PS will need to be updated


seisman:
This PR should be merged into 6.0 branch, right?

PaulWessel:
I believe so

PR #1716, opened 2019-10-02
docs: Cleanup Julia and matlab scripts

seisman:
These Julia and Matlab scripts are not used in the GMT documentation.
Even if these scripts are useful, they should be put in their own
repositories.


PaulWessel:
Hi @joa-quim, could you copy/paste/add these to the suitable repos?

joa-quim:
The point is that those tests (and there are couple more) are more or less useless outside the GMT tests tree. Not these in particular that require no input, but others. And I don't see the logic of the purge. We have bash and bat tests, having them in other languages is just an extension. Aside, regarding the 2 julia tests, they are very outdated and can be removed. The other 2, well yes, there not particularly useful.

seisman:
The reason of the purge is that these scripts are never tested or maintained.

joa-quim:
The fact that there is no automatic service running those scripts doesn't mean they are never used (I wrote them for a purpose). Again, regarding these 4 scripts the julia ones can be removed but for the Matlab I have hopes that we will be able to run CI Matlab scripts. Talked yesterday with a ML engineer and she told me they are working on a solution for that.

PaulWessel:
But isn't the point that they cannot run without also installing the gmt mex and Julia stuff? Installing just GMT is not enough, hence they below in the gmt mex and gmt julia repos since both of those depend on GMT.  They dont have to live in the gmt repo to be run from some specific directory.

joa-quim:
I've been telling that there are more of those .m scripts that sit next to their bash cousins, but I'm OK to delete them all.

Capture

seisman:
I didn't notice that there are so many matlab scripts hidden in the test directory. They definitely should be moved to the gmtmex repository.

PR #1760, opened 2019-10-08
Set clip path for map insets by default

PaulWessel:

As for gmt text, we now let gmt inset begin set a clip path so that anything extruding outside the inset panel will be clipped. This behavior can be turned off with **-N**.  Implemented with hidden parameter in the PostScript since gmt inset end needs to restore the previous clip path but does not take any options.

### DIFF
--- a/doc/rst/source/inset.rst
+++ b/doc/rst/source/inset.rst
@@ -23,6 +23,7 @@ Synopsis (begin mode)
 |-D|\ *inset-box*
 [ |-F|\ *box* ]
 [ |-M|\ *margins* ]
+[ |-N| ]
 [ |SYN_OPT-V| ]
 [ |SYN_OPT--| ]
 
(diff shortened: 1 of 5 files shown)


seisman:
This script fails with segmentation fault:
```
gmt begin inset pdf
  gmt psbasemap -R0/7/3/7 -Jx1i -B0
  gmt inset begin -DjTR+w3i/2i
    gmt plot -R0/7/3/7 -B @Table_5_11.txt -Sc0.15i -Glightgreen -Wfaint
    gmt plot @Table_5_11.txt -W1.5p,gray
    gmt plot @Table_5_11.txt -St0.15i -Gorange
  gmt inset end
gmt end show
```

seisman:
Should the ticks at north and south sides also trimmed?
image

PaulWessel:
Don't think we can.  This happens because the inset dimension and actual plot aspect ratio differs and only the E/W exceed the max panel dimensions.  Normally, users will need to use **-M** to set margins, and/or add clearance via **-F+c**.  If we wanted to clip to the map boundary inside the map inset then we would have to pass some information to the module inside the inset that sets up **-R -J** etc.   But we never clip annotations in other GMT plots (we clip other things but not annotations).  I think here we just don't want stuff to plot outside the panel dimensions.  Think you need **-M**0.75c here to make this work well.

PR #1763, opened 2019-10-09
Drop UNIX manpages

### LINKED ISSUE #1728: Drop UNIX man pages?
The UNIX man pages seem useless for most users, as now they can access the HTML version local/remote documentation via `gmt docs`. The UNIX man page format is also limited to plain text, preventing us from adding non-text elements (e.g. figures) to the manual.

Do you think we should drop the UNIX man pages?

seisman:
This PR remove all UNIX manpage stuff. Closes #1728.

The following code snippet use the `GMT_MANPAGE` macro, which seems never used and should be removed. But I'm not sure what to do here.
https://github.com/GenericMappingTools/gmt/blob/7ead9ce7af4643ef535d96cc07701735c8427f65/src/gmt_version.h.in#L51-L62


PaulWessel:
Regarding your question above, if GMT_MANPAGE is never set anywhere then it seems we just remove the ifndef GMT_MANPAGE and the outer endif, no?

PR #1771, opened 2019-10-09
Failed to separate gmt begin from gmt inset begin, etc

PaulWessel:
Both **gmt inset begin** and **gmt subplot begin** was incorrectly detected as **gmt begin** by my naive test.  This caused bad failures of course, with final scripts that lacked gmt inset begin but had gmt inset end.  We now check that only lines with both gmt and begin but not having either inset or subplot are genuine starts of modern mode scripts (same deal with **gmt end**).

### DIFF
--- a/src/movie.c
+++ b/src/movie.c
@@ -876,10 +876,15 @@ GMT_LOCAL bool is_comment (char *line) {
 	return (line[k] == '#') ? true : false;		/* Will return true for lines starting with some tabs, then # */
 }
 
+GMT_LOCAL bool is_not_insetsub (char *line) {
+	/* Return true if neither inset nor subplot strings are found */
+	return (strstr (line, "inset") || strstr (line, "subplot")) ? false : true;
+}
+
 GMT_LOCAL bool is_gmtbegin (char *line) {
 	/* To handle the cases where there may be more than one space between gmt and begin... */
-	if (is_comment (line)) return false;	/* Must avoid finding # gmt begin */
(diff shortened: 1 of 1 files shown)


PaulWessel:
Hi @seisman, the two recent fixes to movie.c points out that since it has not seen that much use yet, there may be some dumb first-order bugs still in it.  I found the last two (not considering commented out "# gmt begin" lines or thinking "gmt inset begin" was "gmt begin", if you have thoughts on other dumb things that could sneak through the tests when rewriting a script please let me know.

seisman:
I think there is a more robust way to detect if a line is "gmt begin" or "gmt end". 

1. Split the current line by spaces and tabs (using `strtok`).
2. Check if the first substring is `#` or `REM`.
3. Check if the 2nd is "gmt" and the 3rd is "begin" or "end".

PaulWessel:
Hi @seisman, I followed your suggestion.  All movies run for me except anim_08.sh.  Now, I have trouble with the curl command from time to time, but when I run this it just dies here:

> ...
> movie [INFORMATION]: Create parameter initiation script movie_init.sh
> movie [INFORMATION]: Create preflight script movie_preflight.sh and execute it
> Abort trap: 6

Running the same movie command in Xcode works...  I suspect it is something odd with my machine - could you please try?  if it works for you then I think you can approve the PR>

PaulWessel:
Actually had a bug in that the word array could overflow...Fixed now and runs again.

PR #1773, opened 2019-10-10
Address bad wrapping in pscoast

### LINKED ISSUE #1759: pscoast -JX or -JQ wrapped across dateline generates garbage
I am using GMT5 v5.4.1 (on CentOS 7 linux). I am trying to use pscoast to generate a map that wraps across the international date line. The command I am using is as follows:

```
gmt pscoast -JX5.16558972d/7d -R-42.2839/223.375/-90/90 -BWeSn -Bxf10ag40 -Byf5ag20 -Dl -S180/220/255 -W1/.2 -W2/.05 -G200 -X.75 -Y.75 -K
```

I have also tried replacing the "-JX..." above with "-JQ265.6589/6i".

In both cases, the map I get is garbagified, with large rectangles of undrawn space covering the continents.

jx

**System information**

* Operating system: CentOS Linux release 7.6.1810
* GMT version (`gmt --version`): 5.4.1

PaulWessel:
A 2016 longitude wrap fix should only apply for global maps (360-degree range) only.  Adding a test that now passes. Closes #1759.

### DIFF
--- a/src/gmt_map.c
+++ b/src/gmt_map.c
@@ -1709,7 +1709,7 @@ uint64_t map_wesn_clip (struct GMT_CTRL *GMT, double *lon, double *lat, uint64_t
 /* This is new approach to get rid of those crossing lines for filled polygons,
  * i.e., issue # 949.  Also see comments further down.
  * P. Wessel, Dec 1 2016 */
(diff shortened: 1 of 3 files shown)


seisman:
I still have the bug in #1759 and the test pscoast_JQ.sh still doesn't work.

PR #1785, opened 2019-10-11
Use the SHELL/shell env variable in script and use if known

PaulWessel:
If we can determine the shell then we use that as the shell in the start line, since we then can pass **-e** which stops the shell upon the first error in the script.  This is better than having a GMT script continue despite an error.  For most users of bash or csh it would then start with

#!/bin/bash -e

but with zsh and tcsh being other shells that are often used we should write

#!/bin/zsh -e

etc, and not bash.


seisman:
I like the idea of `-e` to stop at the first error.

For the shebang, I have a different opinion. I use zsh as my default shell, because it provides better completion and other useful features in the terminal, but I still write and run bash script.

PaulWessel:
OK, I can see that zsh has useful things for interactive sessions but those might not be used in scripts.  So you would issue bash is zsh and csh if tcsh then probably?

seisman:
Yes, actually most of the time I run a script with `bash xxx.sh` rather than `./xxx.sh`.

PaulWessel:
So we use the env command to find bash but that way we cannot pass -e.  In bash we can do set -e but bot in csh. So if SHELL returns /bin/zsh, do we date to assume that path is true for all shells? I think not.  So maybe if zsh we do env bash plus set -e, else we use path -e?

seisman:
I'm not sure if I understand you correctly. The following script stops at the basemap command (note the wrong `-Rg1` option):
```
#!/usr/bin/env bash -e
gmt begin
gmt basemap -Rg1 -JH10c -Baf
gmt coast -W1p
gmt end show
```

I also replace bash with csh, zsh and tcsh, they all work.

PaulWessel:
Great, stack overflow said it did not work but I did not check. OK so we do this then.  There is no similar solution for DOS.

seisman:
I'm afraid it works for macOS only. Linux gives erros:
```
$ ./test.sh
/usr/bin/env: zsh -e: No such file or directory
```

PaulWessel:
We need to wrap this up now. Should we do this:

- If SHELL returns bash, use shebang path with -e, e.g., #!/usr/bin/bash -e
- if SHELL returns zsh, use shebang path with env "#!/usr/bin /env bash" and then call set -e explicitly
- if SHELL returns a csh flavor we use that path and add -e as well
- Otherwise we leave the -e stuff off since we don't know.

seisman:
I think we can use `#!/usr/bin/env -S bash -e`. It works on both Linux and Mac.

PaulWessel:
OK great, so do that if bash, sh, zsh, ksh (korn shell) then do bash with -Se via env, and if csh, tcsh then I do a csh via env -S the same way. For DOS we do nothing different than now.

PaulWessel:
Output for me is now

```
#!/usr/bin/env -S bash -e
# GMT modern mode bash template
# Date:    2019-10-15T11:37:42
# User:    pwessel
# Purpose: Purpose of this script
export GMT_SESSION_NAME=$$	# Set a unique session name
gmt begin figurename
	# Place modern session commands here
gmt end show
```

I have two questions:

1. Should the gmt begin be more complete, i.e., `gmt begin filename [formats] [convertoptions]`?
2. Should there be a way to specify which shell I want? I may want to write a csh script for a demo but I cannot unless I change my shell.   Maybe add **--new-scrip**t[_=dos|bash|csh_]?

PaulWessel:
I implemented the above, but say batch instead of dos.

seisman:
> 1. Should the gmt begin be more complete, i.e., `gmt begin filename [formats] [convertoptions]`?

I don't like it. PDF is a good default format and *convertoptions* also has a reasonable default. The current template script can be run without changing any single line. With a "complete" begin command, users may have to understand the meaning of _formats_ and 
_convertoptions_ and then remove/change them.

> 2. Should there be a way to specify which shell I want? I may want to write a csh script for a demo but I cannot unless I change my shell.   Maybe add **--new-scrip**t[_=dos|bash|csh_]?

Good idea. I think we can make it more general: **--new-script**=*shell*, and *shell* can be any shell, including bash, csh, et al, and also "dos" or "batch" for Windows users.

PaulWessel:
OK, agree with point one.  Keep gmt begin as is.

 I can add the word "dos" to mean batch as well.

RIght now I check if the arg matches one of the three types we support.  What should --new-script=zsh do?  or --new-script=/var/weird/voodoshell do? We don't want nonfunctional scripts because a user shell dont actually exist.

PR #1797, opened 2019-10-12
Delay hash refresh by 24 hours if server is down

PaulWessel:
If getting the gmt dataserver hash file takes to long (> 10seconds) then the server is down, probably.  Prevent this from begin repeated by all GMT remote data access commands by updating the modification time on the local hash file so the 24 hour cycle is reset.
Not tested; my hash file will be 24 hours old in 2 hours and can test then...

### DIFF
--- a/src/gmt_io.h
+++ b/src/gmt_io.h
@@ -245,6 +245,7 @@ struct GMT_IO {				/* Used to process input data records */
 	bool first_rec;			/* true when reading very first data record in a dataset */
 	bool trailing_text[2];	/* Default is to process training text unless turned off via -i, -o */
 	bool hash_refreshed;		/* true after calling the hash_refresh function the first time */
+	bool internet_error;		/* true after failing to get hash table due to time-out */
 	uint64_t seg_no;		/* Number of current multi-segment in entire data set */
 	uint64_t seg_in_tbl_no;		/* Number of current multi-segment in current table */
 	uint64_t n_clean_rec;		/* Number of clean records read (not including skipped records or comments or blanks) */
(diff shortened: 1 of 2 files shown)


PaulWessel:
Did you try it and it worked?

joa-quim:
No, doesn't even build
```
C:\progs_cygw\GMTdev\gmt5\master\src\gmt_remote.c(251): warning C4013: 'utimes' undefined; assuming extern returning int
```

joa-quim:
Now it builds but makes no difference. Even with a ``gmt("destroy")`` it keeps trying to download ``gmt_hash_server.txt``

PaulWessel:
OK, I can test this in one hour now.  Do some googling on what Windows requires to deal with file modification times etc.  See lines 340-50 in gmt_remote.c.
Do you have an existing gmt_hash_server.txt file or do you end up in the test at line 316?

joa-quim:
The problem is that when it fails to download the code flow slips through the ``return 1`` at line 244

PaulWessel:
OK, I did not know where it would go (yet).  You want to fix it?

joa-quim:
Don't know. I don't understand the code. I noticed that I only had a ``gmt_hash_server.txt.old`` file and no ``gmt_hash_server.txt`` so the ``_utime (file, NULL);`` is actually doing nothing. And why didn't I have a ``gmt_hash_server.txt`` file? Possibly because it was removed (or renamed), but then this algo does nothing because when it comes in there is no file anymore to update.

PaulWessel:
OK, I will work on this now - done grading two labs...

PaulWessel:
Need to change things around.  It should not copy current to old before knowing if download will work.  That is why you only have old.  Just rename it back for now (remove _old).

PaulWessel:
This weekends A/C repair and consequent GMT data server outage provided some opportunities for testing.  Trying a series of fixes here:

1. Only rename hash table to old and update new after a successful download of new table.
2. If timing out on getting new hash table we simply refresh modification time on the current to delay the next hash checkup by 24 hours.
3. Impose a max 5 seconds timeout to get that tiny hash table downloaded.
4. If download fails, then set an error flag so we don't try again this session.
5. When lib curl failure, remind users they can set GMT_DATA_SERVER_LIMIT = 0 to skip remote downloads.

If server is down and users do not head that message, then the next remote file access will take a while to time-out, and the message about GMT_DATA_SERVER_LIMIT is repeated.  Not much more we can do I think.  First time users without a hash table will never get one so max wanting 5 seconds to learn why.

PR #1809, opened 2019-10-15
Add gmtinit_get_history & gmtinit_put_history to the dll

joa-quim:
I need to access them in Julia, and they were not exported to the gmt.dll


joa-quim:
Thought in something like that, but we already have ``gmtinit_conf_US`` & ``gmtinit_conf`` in the dll

PaulWessel:
Then those need to be renamed as well.  We want to be consistent.

PR #1819, opened 2019-10-16
Reset common option history during gmt subplot begin

### LINKED ISSUE #1783: Insets and subplots should have no initial history

When we begin a map inset, the history from the figure is inherited.  This can give bad side-effects.  Below are two near-identical scripts. Both plot a Mercator map with another Mercator map as inset.  First the proper way, specifying ? as the unknown map width that will fit the inset:

```
gmt begin good
	gmt basemap -JM6i -R0/80/0/50 -B -BWSen -X0.7i -Y0.3i
	gmt inset begin -DjTL+w2i/3i+o0.2i -M.4i -F+p1p
        	gmt basemap -R160/180/30/55 -JM? -Ba10fg10 -BWSne --MAP_FRAME_TYPE=plain
		echo 170 50 | gmt plot -Sc0.1i -Gred
	gmt inset end
gmt end show
```

This script gives the desired plot:

two

Now, if I simply give -JM without the ?, the script is

```
gmt begin bad
	gmt basemap -JM6i -R0/80/0/50 -B -BWSen -X0.7i -Y0.3i
	gmt inset begin -DjTL+w2i/3i+o0.2i -M.4i -F+p1p
        	gmt basemap -R160/180/30/55 -JM -Ba10fg10 -BWSne --MAP_FRAME_TYPE=plain
		echo 170 50 | gmt plot -Sc0.1i -Gred
	gmt inset end
gmt end show
```

which gives the wrong plot:

one

The reason is that the figure's -JM6i is copied into the history, hence -JM finds that before it auto-scales to fit the inset.

This exercise brings up two questions:

1. Since the inset (or subplot panel) is a _new_ plot it should not have _any_ history.  Agreed?
2. With a clean history, -JM will give an error (no history for -JM found).  This matches our GMT6 paper syntax. We should not allow projections such as -JM as shorthand for -JM?.  Agreed?

PaulWessel:

When **gmt subplot begin** is called the gmt.history file for the current figure or session may have a record of **-R -J -X -Y** settings.  Once the subplot begin starts we may or may not be given **-R -J -X -Y** arguments.  If not, then we do not want old history settings to leak into the subplot panels.  The past history is read during session creation in _GMT_Create_Session_ so at this point it is already stored in memory.  Hence, we need to reset all of that before we parse the common and specific arguments to this module.  This PR implements this change and closes #1783.

### DIFF
--- a/src/subplot.c
+++ b/src/subplot.c
@@ -689,6 +689,10 @@ int GMT_subplot (void *V_API, int mode, void *args) {
 
 	if ((GMT = gmt_init_module (API, THIS_MODULE_LIB, THIS_MODULE_NAME, THIS_MODULE_KEYS, THIS_MODULE_NEEDS, &options, &GMT_cpy)) == NULL) bailout (API->error); /* Save current state */
 	if (GMT_Parse_Common (API, THIS_MODULE_OPTIONS, options)) Return (API->error);
+	
+	if ((opt = GMT_Find_Option (GMT->parent, GMT_OPT_INFILE, options)) && !strcmp (opt->arg, "begin"))	/* Called gmt subplot begin */
+		gmt_reset_history (GMT);	/* Remove history */
+	
 	Ctrl = New_Ctrl (GMT);	/* Allocate and initialize a new control structure */
 	if ((error = parse (GMT, Ctrl, options)) != 0) Return (error);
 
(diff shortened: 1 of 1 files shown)


seisman:
There are many failing tests now.

PaulWessel:
Odd, I ran all the tests.  Let me wipe clean and do again.

PaulWessel:
OK, got the failures, will evaluate.

PaulWessel:
I had a brain fart and wanted to speed up by testing the strcmp first...but opt not set yet.  I ran the test successfully first then made that dumb change.  Changed back, works fine again for all tests.

Issue #1820, opened 2019-10-16
All animations fail on Windows


seisman:
And also ex22.sh

PaulWessel:
Cannot tell why.  The log has lots of gmtserver failures but I dont see any details as to why it fails.  Are you able to run one locally under Windows and get some clues?  I will review the code changes in movie but since it works on macOS and Linux there must be something that is not quite the same for Windows.

PaulWessel:
@joa-quim, can you test an animation and let me know if there is anything?

seisman:
Just ran the anim02.sh, it doesn't report any error, but it also don't generate the PS file.

Below is part of the verbose messages.
```
movie [DEBUG]: Found file main.sh
movie [DEBUG]: Found file main.sh
movie [DEBUG]: Found file main.sh
movie [INFORMATION]: Paper dimensions: Width = 3.5i Height = 4.167i
movie [INFORMATION]: Pixel dimensions: 252 x 301
movie [DEBUG]: Look for file angles.txt in C:/Users/seisman/.gmt
movie [DEBUG]: Look for file angles.txt in C:/Users/seisman/.gmt/cache
movie [DEBUG]: Look for file angles.txt in C:/Users/seisman/.gmt/server
movie [DEBUG]: Look for file angles.txt in C:/Users/seisman/.gmt/server/srtm1
movie [DEBUG]: Look for file angles.txt in C:/Users/seisman/.gmt/server/srtm3
movie [INFORMATION]: Create parameter initiation script movie_init.sh
movie [INFORMATION]: Create preflight script movie_preflight.sh and execute it
movie [DEBUG]: Found file angles.txt
movie [DEBUG]: Found file angles.txt
movie [DEBUG]: Found file angles.txt
movie [DEBUG]: Found file angles.txt
movie [DEBUG]: Found file angles.txt
movie [DEBUG]: Object ID 0 : Registered Data Table File angles.txt as an Input resource with geometry Non-Geographical [n_objects = 1]
movie [DEBUG]: api_begin_io: Input resource access is now enabled [container]
movie [DEBUG]: api_import_dataset: Passed ID = 0 and mode = 0
movie [INFORMATION]: Reading Data Table from File angles.txt
movie [DEBUG]: Found file angles.txt
movie [DEBUG]: Found file angles.txt
movie [DEBUG]: Source col types: (Number,Number)
[...]

PR #1822, opened 2019-10-16
When GMT_END_SHOW=off was added I forgot the export line

PaulWessel:
There is a separate fprintf before that prints the "export" part of the statement.  When I added the GMT_END_SHOW parameter I forgot to add that extra fprintf statement.


PaulWessel:
This should address issue #1820.

src/movie.c
```diff
@@ -1144,6 +1144,7 @@ int GMT_movie (void *V_API, int mode, void *args) {
 			fprintf (fp, "set GMT_SESSION_NAME=1\n");
 		else	/* On UNIX we may use the calling terminal or script's PID as the GMT_SESSION_NAME */
 			set_tvalue (fp, Ctrl->In.mode, true, "GMT_SESSION_NAME", "$$");
+		fprintf (fp, "%s", export[Ctrl->In.mode]);		/* Hardwire a Session Name since subshells may mess things up */
```
seisman: The comments are incorrect.

PaulWessel: Sorry, will fix.

PR #1829, opened 2019-10-17
Replace backslashes with slashes in paths passed to shell

PaulWessel:
While getcwd may return paths with backslashes, the shell in Windows may not like that.  Beautiful example of the left hand and right hand not talking.


seisman:
Are there any side-effects for Windows batch users? Currently, we don't have any batch scripts for testing animations.

joa-quim:
cmd accepts ``cd`` and ``move`` commands to work on paths with forward slashes. So, there should be no issues on this side.

PR #1846, opened 2019-10-19
Update meta-data for the earth_relief_xxy grids

PaulWessel:
Since they are now all based on _Tozer et al, 2019_ using Cartesian filtering on SRTM15+V2, then we need to update the attributions.  Also allow 01d and 60m to be used for the coarsest file.


seisman:
The rST files also need updates.

PR #1851, opened 2019-10-20
Report versions of gm, ffmpeg, etc.

PaulWessel:
Finally figured out how to deal with programs such as gm and ffmpeg who decided it is a **good idea** not to have a --version option, so we instead are forced to grep and awk and cut etc to get the obvious answers.


PaulWessel:
OK, shall I put it inside
	if (WIN32)
and only do the version if not Windows?

seisman:
In the top CMakLists, we have the `FindUnixCommands` directive, which can find if bash, gzip, tar et al. are available (see https://cmake.org/cmake/help/latest/module/FindUnixCommands.html).
https://github.com/GenericMappingTools/gmt/blob/da90090545e41ebfa472ca48de51f62e7f28ae81/CMakeLists.txt#L74

I believe we can check if `${BASH}` is defined.

seisman:
The Windows CI has bash, and returns:
```
*  Found GhostScript (gs)     : yes
*  Found GraphicsMagick (gm)  : no
*  Found ffmpeg               : no
*  Found open                 : yes
*  Found ogr2ogr              : yes (2.4.1)
*  Found gdal_translate       : yes (2.4.1)
```

PaulWessel:
So no gm and ffmpeg installed? The test checks if they exist before we try to get versions.  And seems to work for the GDAL tools.

seisman:
gm and ffmpeg are only installed when we run the whole tests.

PaulWessel:
OK, so it works on Windows if bash is installed.  is it likely that there will be people running cmake on Windows and not have bash? 99.9% of Windows users will install from the binary installer, no?  Is it worth adding an if-test? Perhaps cmake will stop if one of these fails due to lack of bash, so safest to add that protection?

seisman:
This is what I get on Windows. I have git bash installed, but I don't add it to my PATH, so it's invisible to cmake.
image

> OK, so it works on Windows if bash is installed. 

Strictly speaking, no. Having bash installed doesn't mean tr and cut are also available. But I think it's OK to me. If someone doesn't have bash, tr or cut, then the versions are empty.

src/CMakeLists.txt
```diff
@@ -45,30 +45,34 @@ if (NOT OGR2OGR)
 	message (WARNING "Cannot find program ogr2ogr - gmt will be unable to convert Shapefiles to GMT/OGR format")
 	set (GMT_CONFIG_OGR2OGR_MESSAGE "no" CACHE INTERNAL "OGR2OGR config message")
 else (NOT OGR2OGR)
-	set (GMT_CONFIG_OGR2OGR_MESSAGE "yes" CACHE INTERNAL "OGR2OGR config message")
+	execute_process(COMMAND bash "-c" "ogr2ogr --version | tr ',' ' ' | cut -f2 -d' '" OUTPUT_VARIABLE answer OUTPUT_STRIP_TRAILING_WHITESPACE)
```
seisman:
```suggestion
	execute_process(COMMAND ${BASH} "-c" "ogr2ogr --version | tr ',' ' ' | cut -f2 -d' '" OUTPUT_VARIABLE answer OUTPUT_STRIP_TRAILING_WHITESPACE)
```

It's better to use `${BASH}` instead of `bash`.

PaulWessel:
Can we do better than the awkward `tr ',' ' ' | cut -f2 -d' `and `head -n 1 | cut -f2 -d' '` ?

PaulWessel:
Let's wait to accept this PR until @joa-quim has tried it.

seisman:
> Can we do better than the awkward `tr ',' ' ' | cut -f2 -d' `and `head -n 1 | cut -f2 -d' '` ?

I believe I found a better way. Will try if it works.

PaulWessel:
I guess it is a question of ${AWK} is more likely to be available in Windows than tr and head.  We could do

gm help | awk '{if (NR == 1) print $2}'
ogr2ogr --version | awk -F, '{print $1}'

for instance.

seisman:
Closed as #1853 is better.

PR #1876, opened 2019-10-22
Set HAVE_GIT_VERSION to TRUE if it is a git repository

seisman:
Check if the `.git` directory exists to determine if it's a git repository.

For a git repository, set HAVE_GIT_VERSION to TRUE, and enable the `gmt_release` target.


cmake/modules/ConfigCMake.cmake
```diff
@@ -39,9 +39,14 @@ endif (NOT CMAKE_BUILD_TYPE)
 
 # Here we change it to add the git commit hash for non-public releases
 set (GMT_PACKAGE_VERSION_WITH_GIT_REVISION ${GMT_PACKAGE_VERSION})
+
+# Check if it's a git repository of not
```
seisman:
```suggestion
# Check if it's a git repository or not
```

PR #1879, opened 2019-10-23
Allow grdfilter -F?width to take a grid of filter widths

PaulWessel:
Needed to supply a spatially variable filter _width_ via a grid.  Compiles, need to test it.  We will enforce that the grid must have same dimensions and registration as the desired output grid.

### DIFF
--- a/doc/rst/source/grdfilter.rst
+++ b/doc/rst/source/grdfilter.rst
@@ -84,6 +84,9 @@ Required Arguments
     diameter *width*. This gives an isotropic filter; append /*width2*
     for a rectangular filter (requires **-Dp** or **-D0**).  By default we
     perform low-pass filtering; append **+h** to select high-pass filtering.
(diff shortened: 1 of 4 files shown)


PaulWessel:
I finished implementing the variable filterwidth filtering in grdfilter.  If the filter-width is a grid it must be of the same registration and domain as the output grid.  Not widely tested yet but given the machinery is the same for all filters I think it is OK for master commit.  Below is a test script output that filters 4 spikes with a Gaussian filter.  The top left is the spikes, the top right is the filter width grid, the bottom left is the constant filter width of 20 and the bottom right is the variable filter result.

varfilter

PaulWessel:
Hi @joa-quim or @seisman, please approve.  This is master branch with cutting edge GMT.  Test works, all seem well, and my student will be needing this for his project.

PR #1885, opened 2019-10-24
Let grdtrack -E+c continue track if next line is a continuation of previous

PaulWessel:
If we need several intermediate point to define a line then we want distance to increase continuously and we want the result to be a single segment.  Currently we get individual line segments with distance reset to zero for each segment.  The planned **+c** modifier will tied these segments together once implemented. Here we lay the groundwork for the new capability.


PaulWessel:
Tested and this works fine now.  Please approve for master, @joa-quim or @seisman.  The failure in Win compilation is due to OpenMP being turned on which exposed an unsigned variable in the OpenMP loop.  We should fix that in 6.0 first.

joa-quim:
This PR changes 63 files. Several seem really unrelated to this.

PaulWessel:
Sory, needed to merge in from 6 and master... Now 3 files

PR #1918, opened 2019-10-26
Finalize extra packing for macOS bundle

PaulWessel:
This PR finalizes how cpack works on macOS under MacPorts; this time handling the gs link.  If there is no gs -> gsc link on brew then we need to work on the script a bit more.  It would be good if @seisman on brew could

1. Temporarily set GMT_PUBLIC_RELEASE true
2. In top gmt-dev directory, run admin/build-release.sh
3. Make sure the bundle still works and contains all the goodies.
4. Examine build/add_macOS_cpack.txt to make sure all is well.

Maybe we should exchange bundles to make sure they run on systems with a different distro.


seisman:
For homebrew, all the commands are located in `/usr/local/bin` directory, but they are symlinks.

```
$ which gs gsc
/usr/local/bin/gs
/usr/local/bin/gsc
```

```
$ ll /usr/local/bin/gs /usr/local/bin/gsc
lrwxr-xr-x  1 seisman  admin    33B Oct 22 21:36 /usr/local/bin/gs -> ../Cellar/ghostscript/9.50/bin/gs
lrwxr-xr-x  1 seisman  admin    34B Oct 22 21:36 /usr/local/bin/gsc -> ../Cellar/ghostscript/9.50/bin/gsc
```

PaulWessel:
So given my gs was copied as an existing symlink, we will need the build-macos*.sh script to replace links with the real thing.  I see in bash, -L checks if a link, so then we need to do something like

```
for file in all files; do
  if [ -L $file ]; then # symblink
     echo the name the link points to
  else
    echo $file
done
```

Better suggestions?

PaulWessel:
Google says use readlink.  So we use readlink in that loop.
Please type readlink /usr/local/bin/gs and see if that works.

seisman:
```
$ readlink /usr/local/bin/gs                                                                                                                                                                       
../Cellar/ghostscript/9.50/bin/gs
```

PaulWessel:
Do you always get those ..? What if you run the readlink from another directory?

seisman:
Nothing changes.

PaulWessel:
So if we do this in the loop and /usr/local/bin/gs becomes ../Cellar/..., will cpack find things? Or do we need to replace the .. with the actual directory?

seisman:
```
$ realpath `which gs`
/usr/local/Cellar/ghostscript/9.50/bin/gs
```

I find `realpath` very useful, but it's a Linux command, and I installed it via `brew install coreutils`.

PaulWessel:
OK, I can do that too.  Perhaps that is the safest? And we document somewhere in admin that macOS maintainers will need that package?

PaulWessel:
I got this mesage:
  coreutils has the following notes:
    The tools provided by GNU coreutils are prefixed with the character 'g' by default to distinguish them from the BSD commands.
    For example, cp becomes gcp and ls becomes gls.
    
    If you want to use the GNU tools by default, add this directory to the front of your PATH environment variable:
        /opt/local/libexec/gnubin/

can you use grealpath?

PaulWessel:
One more complications.  If I blindly do this on all links, I get

	/opt/local/bin/gsc
	/opt/local/bin/gsc

and no gs.  So for macports, need to preserve that gs link.   So your gs is the executable?

PaulWessel:
I am modifying the script a bit and making three sets:

```
# 1a. List of executables whose shared libraries also are needed.
EXEPLUSLIBS="gsc gm ffmpeg"
# 1b. List of symbolic links.
EXELINKS="gs"
# 1c. List of executables whose shared libraries have been included via GDAL
EXEONLY="ogr2ogr gdal_translate"
```

Do I need a check on Macports vs HomeBrew to make changes here?  The EXELINKS will just let which tell us the path, but the other two will go through a greadpath step.

PaulWessel:
Please see if just running this produces acceptable output, i.e.

bash build-macos-external-list.sh

seisman:
This is what I get:
```
install (PROGRAMS
	/usr/local/Cellar/ffmpeg/4.2.1_1/bin/ffmpeg
	/usr/local/Cellar/ghostscript/9.50/bin/gsc
	/usr/local/Cellar/graphicsmagick/1.3.33/bin/gm
	/usr/local/bin/gs
	DESTINATION ${GMT_BINDIR}
	COMPONENT Runtime)
```

For me, both gs and gsc are executables, but gs is 13 M and gsc is only 9 Kb.

PaulWessel:
OK, looks reasonable to me, I guess you can try to run the whole thing: Set the GMT_PUBLIC_RELEASE to true and try

bash admin/build-release.sh

and see if it works or not.

seisman:
Nope, `/usr/local/bin/gs` is wrong. It's a symlink.

PaulWessel:
But /opt/local/bin/gs is a symlink for me and it copes that in as a link.

seisman:
```
-rwxr-xr-x  1 seisman  staff   266K Oct 26 19:00 ffmpeg
-rwxr-xr-x  1 seisman  staff    16K Oct 26 19:00 gm
-rwxr-xr-x  1 seisman  staff    30K Oct 26 19:00 gmt
-rwxr-xr-x  1 seisman  staff   6.0K Oct 26 18:56 gmt-config
-rwxr-xr-x  1 seisman  staff    18K Oct 21 17:04 gmt_shell_functions.sh
-rwxr-xr-x  1 seisman  staff   5.6K Oct 21 17:04 gmtswitch
lrwxr-xr-x  1 seisman  staff    33B Oct 26 19:00 gs -> ../Cellar/ghostscript/9.50/bin/gs
-rwxr-xr-x  1 seisman  staff   9.0K Oct 26 19:00 gsc
-rwxr-xr-x  1 seisman  staff   1.2K Oct 26 18:56 isogmt
```

PaulWessel:
OK, not so good. I added $distro as a variable.  Perhaps you can play with this by maybe removing the gs and adding it specially as a link to gsc in that same directory?

PaulWessel:
If your actual gs is 13 Mb (in Cellar) while gsc is only 9kB then that seems opposite to what I have (gsc is the program, gs is a link).  Very odd.

seisman:
Tried your DMG file. `gs` doesn't work:

```
$ gs
GPL Ghostscript 9.27: Can't find initialization file gs_init.ps.
```

PaulWessel:
Right.  So party can be solved by setting GS_LIB to point to it, but of course there is the entirety of /opt/local/share/ghostscript.  So would have to add that entire directory into our bundles share directory and have the init set GS_LIB.  I also have /opt/local/share/GraphicsMagick-1.3.33.  GDAL I think we already have, and there is nothing about ffmpeg.  I will try this.

PaulWessel:
Th GraphicsMagic directory only has three files and it says stuff like
This file is optional
  and the software will still function if it is removed.

So I think it is just the ghostscript directory tree.

PaulWessel:
Updated the scripts etc. Also updated the gmt-6.0.0-darwin-x86_64.dmg installer.  See if this works.  If not, see if setting export GS_LIB=/Applications/GMT-6.0.0.app/Contents/Resources/share/ghostscript (or without the ghostscript) works.

seisman:
It doesn't work. Tried different GS_LIB. Only this one can find the init file, but reports other errors:
```
export GS_LIB=/Applications/GMT-6.0.0.app/Contents/Resources/share/ghostscript/9.27/Resource/Init/
```

Errors:
```
GPL Ghostscript 9.27 (2019-04-04)
Copyright (C) 2018 Artifex Software, Inc.  All rights reserved.
This software is supplied under the GNU AGPLv3 and comes with NO WARRANTY:
see the file COPYING for details.

*** Warning: GenericResourceDir doesn't point to a valid resource directory.
               the -sGenericResourceDir=... option can be used to set this.

  ./base/gsicc_manage.c:1241: gsicc_open_search(): Could not find default_gray.icc
| ./base/gsicc_manage.c:2258: gsicc_init_iccmanager(): cannot find default icc profile
**** Unable to open the initial device, quitting.
```

PaulWessel:
Yep, I installed brew and got to the same point just now.  Seems hard to relocate.

PaulWessel:
Running

gs -sGenericResourceDir=/Applications/GMT-6.0.0.app/Contents/Resources/share/ghostscript/9.27/ t.ps

after setting

export GS_LIB=/Applications/GMT-6.0.0.app/Contents/Resources/share/ghostscript/9.27/Resource/Init

gives

  ./base/gsicc_manage.c:1241: gsicc_open_search(): Could not find default_gray.icc 

which is in another directory 

/Applications/GMT-6.0.0.app/Contents/Resources/share/ghostscript/9.27/iccprofiles/default_gray.icc

Remains difficult to get to work.

PaulWessel:
No cmake project for ghostscript, of course.  Per http://www.linuxfromscratch.org/blfs/view/svn/pst/gs.html, also depends on old patched libraries. Not sure what to do now.  Such a critical piece of software and such a pain in the neck.

PaulWessel:
Could we do stuff like setting

```
install (DIRECTORY
	/opt/local/share/ghostscript
	DESTINATION /opt/local/share
	COMPONENT Runtime)

```
if bundle is built under macports and /usr/local/share/ghostscript if brew?  Should not overwrite anything of value if the user is not using a bistro?

PaulWessel:
The Appstore says:
http://macappstore.org/ghostscript/

PaulWessel:
And finally, a simple installer of 9.27 from https://pages.uoregon.edu/koch/, which installs into /usr/local.

joa-quim:
No, GDAL install is not complete. You also need the GDAL_DATA dir and same for proj. I have this in my ``ConfigUser.cmake``
```
	# ------- The GDAL_DATA, PROJ_LIB & GHOST files location -----------------------------------------
	set (GDAL_DATA_PATH C:/programs/compa_libs/gdal_GIT/compileds/VC14_64/data)
	set (PROJ_DATA_PATH C:/programs/compa_libs/proj5_GIT/compileds/VC14_64/share/proj)
	set (GHOST_DATA_PATH C:/programs/compa_libs/ghostscript/bin)
```
which are used by ``share/CMakeLists.txt``, see lines 58-74

PaulWessel:
gs 9.50 seems to have a single lib directory /usr/local/share/ghostscript/9.50/lib so with GS_LIB it may work better than 9.27 and all those crazy hardwired dirs.

PaulWessel:
Looks like all the goal stuff is in /opt/local/share/gdal; it has no subdirectories just lots of files. Where do you place in the installer?

joa-quim:
I place them under ``share/GDAL_DATA`` and ``share/proj`` but then I have a patched proj4 with a function that also checks for the ``proj.db`` file, but not sure if that patch would work on unix too. It says

```
char *get_projlib_dir(const char *name) {
    /* Check if proj.db lieves in a share/proj dir parallel to bin/proj.dll */
    /* Based in https://stackoverflow.com/questions/9112893/how-to-get-path-to-executable-in-c-running-on-windows */
```
anyway even without that patch those files are useful but user would need to set the proj environment variable that points to them (don't remember its name right now)

PaulWessel:
Since we don't install proj exes, just the library, there seems to be no share on macOS.  At least not on my system.

joa-quim:
If you have GDAL3 it uses proj6 and then those files are needed. Especially ``proj.db`` which is where all those EPSG codes and their definitions lives.

PaulWessel:
May be better to build the release on homebrew since gs 9.50 seems to have a more sane directory structure?

PaulWessel:
Hi @joa-quim, can you post a GMT command that will exercise gdal/proj and give an error if the various data share files are not present?  I will be building a bundle via brew.

joa-quim:
Isn't that what the ``proj4.sh`` does?
Also, running grdinfo on a geotiff file should yield if it doesn't find the ``proj.db`` file

PaulWessel:
Building the bundle with brew: No libraries except libgmt and libpostscriptlight are placed in the lib folder.  E.g., not netcdf, gdal, etc.  Under macports they all get placed there.  How can it behave so differently for the same cpack command?

joa-quim:
Because it's hitting on the symlinks?

PaulWessel:
Probably, but @seisman says he has built the bundle under brew and got stuff like netcdf included.  I am building gs 9.50 from source now, under macports, and will try to build a bundle with everything.

PaulWessel:
So building with port now does not include netcdf either.  Apart from the 58 shared libs that build-macos-external-list.sh found, cpack only added the gmt and postscript libs, the gmt plugin dir. No netcdf, pcre, etc.  Maybe an error happened.  Also, would like to build gs without X11  but it adds Z11 libs now.

PaulWessel:
Well, excluding the macos list and building the bundle as I ahve done for years now yields no siupporting shared libraries.  I wonder if it gets confused with both brew and macport files and does nothing.

PaulWessel:
So after removing brew completely, rebooting, and redoing the build-release.sh step, the bundle still has no shared libraries beyond gmt and postscriptlight.  How is this possible?

PaulWessel:
Latest version seems to find all dylibs recursively and adds teh ghostscript and gdal share dirs.  This has gs 9.50 which may only take GS_LIB.  Note: Not set in the shell startup yet so if gs fails please set GS_LIB to the /Application/gmt-6.0.0.app/......./share/ghostscript/9.50/lib or whatever.

PaulWessel:
Am I the only one who thinks

/opt/local/lib/proj6/share

is one stupid place to place share information when there is opt/local/share for that?

SO where to we place this in the bundle? We would have to use GDAL_LIB if that even works for this non-GDAL, probably need to set PROJ_LIB then.  So maybe place this ih share/proj6 on the installer.

PaulWessel:
Last bit that remains is to place all the licenses for gs, gm, ffmpeg, gdal in a licenses folder in the bundle.

PaulWessel:
Should we create a licsenses directory under admin and then just place that directory in the bundle? That may be the simplest.  I added a one-liner in the startup script that directs users to that directory for license information.

PaulWessel:
Think this PR is good to be approved now, @seisman ?

PaulWessel:
share/Licenses on the bundle now lists

> FFmpeg-LGPL-License.txt		GMT-License.txt			GNU-LGPL-3.txt			GraphicsMagick-License.txt
> GDAL-License.txt		GNU-GPL-3.txt			Ghostscript-AGPL-License.txt

and when you double-click on the GMT logo it prints the gmt splash screen and adds

> Note 2: GMT may use Ghostscript, GraphicsMagick, FFmpeg, and GDAL executables; see /Applications/GMT-6.0.0.app/Contents/Resources/share/Licenses for details.

PaulWessel:
Hi @seisman, any reason this cannot be approved?

admin/build-macos-external-list.sh
```diff
@@ -1,23 +1,62 @@
...
+	top=/opt/local
+elif [ `which cmake` = "/usr/local/bin/cmake" ]; then
+	distro=HomeBrew
+	top=/usr/local
+else
+	distro=Fink
+	/sw
+fi
+# 1a. List of executables needed and whose shared libraries also are needed.
+#     Use full path if you need someting not in your path
```
seisman:
```suggestion
#     Use full path if you need something not in your path
```

admin/build-macos-external-list.sh
```diff
@@ -1,23 +1,62 @@
...
+EXEPLUSLIBS="/opt/bin/gs /opt/local/bin/gm /opt/local/bin/ffmpeg /opt/local/bin/ogr2ogr /opt/local/bin/gdal_translate /opt/local/lib/libfftw3f_threads.dylib"
+# 1b. List of any symbolic links needed
+#     Use full path if you need someting not in your path
```
seisman:
```suggestion
#     Use full path if you need something not in your path
```

admin/build-macos-external-list.sh
```diff
@@ -1,23 +1,62 @@
...
+EXELINKS=
+# 1c. List of executables whose shared libraries have already been included via other shared libraries
+#     Use full path if you need someting not in your path
```
seisman:
```suggestion
#     Use full path if you need something not in your path
```

seisman:
Could you build a bundle after merging this PR and upload it to your FTP? Then I can have more testing on the bundle.

PaulWessel:
Will do now.

PaulWessel:
OK, same ftp link as before.  Heading home - let me know if there are any issues.

PR #1923, opened 2019-10-28
Revert "GMT needs single-precision FFTW  (#1892)"

seisman:
This reverts commit 19aa6eeabb13b00d538dc5f6ce595e71c55348d8.

GMT only needs single-precision FFTW. I thought it's enough to install
the single-precision FFTW library (i.e. libfftw3-single3, fftw-libs-single),
and made the changes in PR #1892.

It turns out we still need to install libfftw3-dev or fftw3-devel, because
fftw3.h is needed when building GMT.


seisman:
The build log confirms that fftw is correctly found on Linux now.

Issue #1936, opened 2019-10-30
gm in the macOS bundle can't compare images


seisman:
I tried to compare two PS files using the gm command in the macOS bundle:
```
gm compare -density 200 -maximum-error 0.003 -highlight-color magenta -highlight-style assign -metric rmse -file test.png A.ps B.ps
```
It reported an error:
```
gm compare: Coder did not return an image (this is a bug, please report it!)
```
However, the same command works with my own gm.

My gm version is:
```
GraphicsMagick 1.3.33 2019-07-20 Q16 http://www.GraphicsMagick.org/
```

The gm in the bundle is:
```
GraphicsMagick 1.3.33 2019-07-20 Q8 http://www.GraphicsMagick.org/
```

seisman:
Q8 and Q16 are different, and Q16 is recommended (http://www.graphicsmagick.org/INSTALL-windows.html#retrieve-install-package).

But I'm not sure if the error is caused by the Q8/Q16 difference, or maybe some graphicsmagick files/libraries are not packed into the bundle.

PaulWessel:
Just got out of a seminar and now heading home.  Will look with otool again to see if anything is missing, but I think otoolr.c is pretty thorough.

seisman:
Closed by #1942

PR #1938, opened 2019-10-30
Add the gm share directory to bundle

PaulWessel:
Unclear if it is used but if they placed it in /usr/local/share it probably is needed.  Possibly related to #1936.


PaulWessel:
Any thoughts on whether this PR is necessary?

seisman:
I don't think these files are necessary.

PaulWessel:
OK, then we can close this (or do we delete then close?)

seisman:
Close, then delete the branch.

PR #1965, opened 2019-11-02
Let build-release.sh reset cmake config file if Ctrl-C is used to stop it

PaulWessel:
I've killed the build-release.sh job a few times and my original ConfigUser.cmake gets lost somehow.  This PR sets a SIGINT trap so that the cleaning up and restoring config files will take place before quitting.


PaulWessel:
I think it does no harm, and it worked on my local install, so I think it is OK to merge.

PR #1977, opened 2019-11-04
cmake: Simplify gmt_release target

seisman:
To create a gmt release tarball, we need to export the git working tree,
and remove some files and directories.

The old cmake script defines several targets: git_export_release, git_prune_dirs,
and git_prune_files. This PR merges the two intermediate targets (git_prune_dirs and
git_prune_files) into the git_export_release target.

Tested and works well.


seisman:
@PaulWessel @joa-quim Can you review this PR? I tested it locally and it doesn't change the release tarball but simplify the cmake script.

seisman:
@PaulWessel Is there any concern about this PR?

PaulWessel:
None, just forgot.

PR #1979, opened 2019-11-04
Prevent libcurl from downloading nonexistant files

PaulWessel:
If there is a typo in an earth_relief_xxy name then we still try to download from the remote server, wasting time and getting confusing error messages from the server.  This PR introduces a list of valid names and prescreens so that we can nip this problem in the bud on the client side.
Problem reported on the Discourse forum.


PaulWessel:
We need to make some deeper changes to avoid these repetitive messages.  But at least it stops the libcurl bleeding for now.

PR #1986, opened 2019-11-05
Add color selection in gmtlogo based on day of year

PaulWessel:
Curently only added Halloween and Christmas as special color periods, with default being the usual GMT logo colors.  I switch to the special color if the day is within ±4 days of the actual holiday.  We can add more colors for holidays as we figure them out.  Adding more colors is simply adding another line of strings in the structure, increment the structure size, and add an if-test for the day range.

### DIFF
--- a/src/gmtlogo.c
+++ b/src/gmtlogo.c
@@ -154,13 +154,31 @@ static float gmt_letters[GMT_N_LETTERS][2] = {	/* The G, M, T polygons */
 
 /* Specific colors for fonts, land, water, text etc */
 
-#define c_font		"51/51/51"
-#define c_grid		"150/160/163"
-#define c_land		"125/168/125"
-#define c_water		"189/224/223"
-#define c_gmt_fill	"238/86/52"
-#define c_gmt_outline	"112/112/112"
-#define c_gmt_shadow	"92/102/132"
+struct GMTLOGO_COLOR {
+	char c_font[GMT_LEN32];
+	char c_grid[GMT_LEN32];
+	char c_land[GMT_LEN32];
+	char c_water[GMT_LEN32];
+	char c_gmt_fill[GMT_LEN32];
+	char c_gmt_outline[GMT_LEN32];
+	char c_gmt_shadow[GMT_LEN32];
+};
+
+struct GMTLOGO_COLOR C[3] = {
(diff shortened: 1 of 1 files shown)


seisman:
It won't work for Chinese New Year, since the date of Chinese New Year isn't fixed in the Gregorian calendar. For example, the Chinese New Year is on Feb. 5 for 2019, and Jan. 25 for 2020. Other festivals/events may have the similar issue (e.g. world cups).

BTW, maybe add a `-T<theme>` option to specify the logo theme, so that we can have a Christmas logo even it's not Christmas.

Since the logo theme are defined by combination of six colors, we can even add modifiers to `-T`, allowing users to design any combinations of colors in the command line.

joa-quim:
Easter also changes in a non trivial way but those dates are pre-computeable in advance. We could have a list of dates for the next 10 years (or, more optimistically, a 1000th years)

PaulWessel:
Presumably there is an algorithm to figure out such things - same with Easter for instance.  I don't like the idea of a **-T** that lets users select or control the colors.  At best, I would be OK with an option to override any auto-coloring we think up and give the plain GMT standard logo colors.  I am reasonably happy with the Halloween colors but unhappy with draft 1 of Christmas.  I think for Oct 21 (GMT anniversary) we can add some extra plot calls to add a gold star and a "xxth anniversary!" text somewhere, but not sure about the rest.  Need more discussion here.

PaulWessel:
And yes, when the world cup takes place is not computable, so probably not a good idea anyway. Brazilian carnival, perhaps @leouieda ?

leouieda:
> Brazilian carnival, perhaps @leouieda?

Carnaval is also lunar and tied to Easter, actually. But what would be the color for Easter?

PaulWessel:
Not sure, purple and something? Anyhow, not sure anyone (me included) is motivated to implement an easter calendar tracker (I am sure one exists in C somewhere) to actually do these things.  So this is definitively a back burner issue.

remkos:
Sounds like WIP. Changing title.

PaulWessel:
Thanks, this has stalled for lack of (a) good color candidates for easter, Christmas, kwanza, Chinese new year, etc and (b) algorithms needed.

PaulWessel:
While a fun idea, it seems there is considerable work needed to get this to work, just for the privilege of doing a random logo on some days.  Not super helpful for users, and probably not the best use of our time.  Shall we agree to let this one die?

KristofKoch:
I can live without it.

Issue #1982, opened 2019-11-04
Bug in coast -JW-65


Esteban82:
When I plot a map centered in -65 (-JW-65/15c) some countries in the far east are painted in black. It do not occurs when I the map is center in the greenwich meridian (-JW15c) or I use the classic mode.

[Modern_Mapamundi_Bug_W0.pdf](https://github.com/GenericMappingTools/gmt/files/3806187/Modern_Mapamundi_Bug_W0.pdf)
[Modern_Mapamundi_Bug_W-65.pdf](https://github.com/GenericMappingTools/gmt/files/3806188/Modern_Mapamundi_Bug_W-65.pdf)

```
gmt begin Modern_Mapamundi_Bug_W-65
	gmt coast -Rd -JW-65/15c -Sdodgerblue2
gmt end show
```

**System information**

* Operating system: Windows 10
* GMT version (`gmt --6.0.0 64-bit version`):

PaulWessel:
I can confirm that problem on macOS as well and will have a look tomorrow.

PaulWessel:
Some more research shows that the same problem is present in classic mode (as long as you use the same resolution, i.e. -Dl) and this goes back to GMT 4 at least.  So as far as I can tell a long-lived problem.  Will need to see if it is a GSHHG problem versus a coding problem.

PaulWessel:
I get the same problem with gshhg 2.3.0.  It depends on the center meridian in the sense that the problem applies to polygon pieces near the right (east) boundary, but a we change the meridian there are other polygons that are close and turn black. So that suggests it is related to decisions made after projecting the coordinates or something with the east periodic boundary.  It has the same problem with -JH, -JN, etc but not -JQ (straight boundary).

PaulWessel:
It also applies to all resolutions but is bin-based, so maybe only affecting the polygons of the far-east bin along the periodic boundary.

PaulWessel:
I removed modern mode from the title since it is unrelated to mode.  These two commands give the exact same results:

```
gmt coast -Rd -JW-45/15c -Sdodgerblue2 -Dl -ps modern
gmt pscoast -Rd -JW-45/15c -Sdodgerblue2 -Dl > classic.ps
```

PaulWessel:
Fixed by #1995.

PR #1995, opened 2019-11-06
Do not have default paint for items not selected

PaulWessel:
This PR starts to address the bug in #1982.  The question is one of fill defaults: pscoast starts by setting these:

```
gmt_init_fill (GMT, &C->S.fill, 1.0, 1.0, 1.0);		/* Default Ocean color = white */
C->C.fill[LAKE] = C->C.fill[RIVER] = C->S.fill;		/* Default Lake/Riverlake color = Ocean color */
gmt_init_fill (GMT, &C->G.fill, 0.0, 0.0, 0.0);		/* Default Land color = black */
```

When a user typically sets land color via **-G** then the ocean fill is set to white even if **-S** is not used.  However, if **-S** is set but not **-G** (like in the users case), the land default is set to black.  Somehow, the black areas that show up in the bug apply the land fill to them.  However, as far as I know, our recursive painting, does _not_ paint areas that are not requested, but perhaps there are some exceptions when we must paint (donut situations?).
The solution seems to be to set areas not requested (i.e., land or ocean) to no_fill (-1,-1,-1) so they are never painted.  Doing so apparently fixes the bug, but a few examples now fail (2) like GMT_az_equidistant which sets land color only (as do most of our examples).  So ocean is unpainted (and hence paper white) and lakes are unpainted (since we only form land polygons for painting) but the Caspian sea (only) here fails.  Changing from -Dc to -Dl and it works, so maybe something odd with the crude resolution (`gmt coast -Rg -JE-100/40/4.5i -Bg -Dc -A10000 -Glightgray -Wthinnest -ps GMT_az_equidistant`)? Same without **-A**.

Nevertheless, this seems to be the right solution despite the new failure in GMT_az_equidistant.

### DIFF
--- a/src/pscoast.c
+++ b/src/pscoast.c
@@ -728,7 +728,7 @@ int GMT_pscoast (void *V_API, int mode, void *args) {
 
 	char *shore_resolution[5] = {"full", "high", "intermediate", "low", "crude"};
 
-	struct GMT_FILL fill[6];	/* Colors for (0) water, (1) land, (2) lakes, (3) islands in lakes, (4) lakes in islands in lakes, (5) riverlakes */
+	struct GMT_FILL fill[6], no_fill;	/* Colors for (0) water, (1) land, (2) lakes, (3) islands in lakes, (4) lakes in islands in lakes, (5) riverlakes */
 	struct GMT_SHORE c;
 	struct GMT_BR b, r;
 	struct GMT_GSHHS_POL *p = NULL;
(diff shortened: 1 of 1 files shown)


PaulWessel:
For completeness, the fill colors are now set this way:

```
	fill[0] = (Ctrl->S.active) ? Ctrl->S.fill : no_fill;
	fill[1] = fill[3] = (Ctrl->G.active) ? Ctrl->G.fill : no_fill;
	fill[2] = fill[4] = (Ctrl->C.active) ? Ctrl->C.fill[LAKE] : fill[0];
	fill[5] = (Ctrl->C.active) ? Ctrl->C.fill[RIVER] : fill[2];
```

0 is land, 1 is ocean, 2 is lake, 3 is island-in-lake, 4 is pond-in-island-in-lake, 5 is special for wide rivers.

PaulWessel:
Please give this PR a go.  I think -JE is special (special treatment in pscoast) and that bug seems unique and the failure depends on projection.

Issue #2009, opened 2019-11-08
grdgradient -n flag sends error with earth_relief_03s and earth_relief_01s


wckoeppen:
**Description of the problem**
I'm trying to create a hillshade using `grdgradient`. I'd like to use the highest-resolution relief datasets, earth_relief_03s and/or earth_relief_01s. But when I try to set the `-n` flag with either of these datasets (e.g., `-nb`):
1. I get an error that says `-n` is given more than once.
2. I see that it's trying to resample the @earth_relief_15s dataset rather than the one I've called for.

If I try to use the 15s dataset or lower resolutions, no error is produced. I can't find anything in the documents suggesting that I can't resample these two datasets. Have I missed it?

**Full script that generated the error**
```
gmt grdgradient @earth_relief_01s -R-81.0/-80.1/28.05/29.05 -A0/270 -nb -Ghillshade.nc
```

**Full error message**
```
$gmt grdgradient @earth_relief_01s -R-81.0/-80.1/28.05/29.05 -A0/270 -nb -Ghillshade.nc

earth_relief_01s: Earth Relief at 1x1 arc seconds tiles provided by SRTMGL1 (land only) [NASA/USGS].

grdsample [ERROR]: Option -n given more than once
grdsample [ERROR]: Syntax error -n option. Correct syntax:
	-n[b|c|l|n][+a][+b<BC>][+c][+t<threshold>]
	   Determine the grid interpolation mode.
	   (b = B-spline, c = bicubic, l = bilinear, n = nearest-neighbor) [Default: bicubic].
	   Append +a switch off antialiasing (except for l) [Default: on].
	   Append +b<BC> to change boundary conditions.  <BC> can be either:
	   g for geographic boundary conditions, or one or both of
	   x for periodic boundary conditions on x,
	   y for periodic boundary conditions on y.
	   [Default: Natural conditions, unless grid is known to be geographic].
	   Append +c to clip interpolated grid to input z-min/max [Default may exceed limits].
	   Append +t<threshold> to change the minimum weight in vicinity of NaNs. A threshold of
	   1.0 requires all nodes involved in interpolation to be non-NaN; 0.5 will interpolate
	   about half way from a non-NaN to a NaN node [Default: 0.5].
grdsample [ERROR]: Offending option -nb
grdblend [ERROR]: Unable to resample file @earth_relief_15s - exiting
```

**System information**

* Operating system: Linux Mint 18.2 Cinnamon 3.4.6
[...]

PR #2010, opened 2019-11-09
No need to pass -n to grdsample as already parsed

PaulWessel:
See issue #2009.  The problem is that we check if **-n** given and then we add **-n** to the command line passed to grdsample.  This is not needed since _-n_ has _already_ been parsed and we are in the same GMT session; hence we get the duplication error.


PaulWessel:
It works fine for me, so I'll merge.

Issue #661, opened 2019-04-25
psmeca plotting full moment tensors


carltape:
Thank you for the sustained support of GMT. I have made some input files for testing plotting routines for full moment tensors. Examples, including figures and input files, can be found here:
https://github.com/carltape/compearth/blob/master/momenttensor/plot/gmt/beachball_plots/
This includes the test script lune_beachballs_min.csh

```
gmtset PAPER_MEDIA Custom_3.2ix7.7i PAGE_ORIENTATION portrait MEASURE_UNIT inch BASEMAP_TYPE plain TICK_LENGTH 0.0c FRAME_PEN 2p TICK_PEN 2p
psbasemap -JXi/6.72i -R-30/30/-90/90 -Ba10f5g10:" ":/a10f5g10:" ":wesn -G200 -K -V -X0.5 -Y0.5 > lune_ngeo_iplot2_lplot1_kplot1_min.ps
psmeca beachballs_ipts1_iref1_lune_psmeca -JXi/6.72i -R-30/30/-90/90 -Sm0.45/8p -L0.5p/0/0/0 -G255/0/0 -N -O -V >> lune_ngeo_iplot2_lplot1_kplot1_min.ps
```

which was run using GMT 4.5.3 and using a modified version of utilmeca.c. The output figure is included, and it is also attached here as
[lune_ngeo_iplot2_lplot1_kplot1_min_psmeca_custom.pdf](https://github.com/GenericMappingTools/gmt/files/3115880/lune_ngeo_iplot2_lplot1_kplot1_min_psmeca_custom.pdf)

The example, partially converted for GMT 5, is:

```
gmt gmtset PS_MEDIA Custom_3.2ix7.7i PS_PAGE_ORIENTATION portrait PROJ_LENGTH_UNIT inch MAP_FRAME_TYPE plain MAP_TICK_LENGTH 0.0c MAP_FRAME_PEN 2p MAP_TICK_PEN 2p
gmt psbasemap -JXi/6.72i -R-30/30/-90/90 -Ba10f5g10:" ":/a10f5g10:" ":wesn -G200 -K -V -X0.5 -Y0.5 > test.ps
gmt psmeca beachballs_ipts1_iref1_lune_psmeca -JXi/6.72i -R-30/30/-90/90 -Sm0.45/8p -L0.5p,0/0/0 -W0.5p,0/0/0 -G255/0/0 -N -O -V >> test.ps
```

This generates the file test.ps, which I have renamed and attached as 
[...]

PR #2018, opened 2019-11-09
psmeca: Make the last few columns of input optional

### LINKED ISSUE #2016: psmeca: Last few columns of input data should be optional
This issue was reported in https://github.com/GenericMappingTools/gmt/issues/661#issuecomment-552072249.

To make it more clear, here is an example.
```
gmt psmeca -R239/240/34/35.2 -JX8c -Baf -Sa2c << END > test.ps
# lon   lat    depth strike  dip  rake mag
239.384 34.556 12.   180     18   -88  5
END
```

For **-Sa**, the format of the input  data is:
```
lon  lat  depth  strike  dip  rake  mag  newX  newY  title
```

In GMT5, the last three columns are optional and the example above works. But in GMT6,  only the last column is optional, and the example fails with an error:
```
psmeca [ERROR]: Mismatch between actual (7) and expected (9) fields near line 2 in file <Standard Input>
```

psmeca in GMT5 works because it reads the input data via `scanf` statements. However, in GMT6, the psmeca I/O was rewritten (by @PaulWessel?) to use the standard GMT I/O mechanism. 

@PaulWessel Is it possible to make the last three columns optional?

seisman:
Fixes #2016.


seisman:
I have a few failing tests locally. Need to check them

PaulWessel:
OK, let me know.

seisman:
@PaulWessel This one doesn't work. It seems the last column "10km" is read as a float value "10.0" instead of a string.
```
gmt psmeca -R239/240/34/35.2 -JX8c -Baf -Sa2c << END > test2.ps
239.384 34.556 12.   180     18   -88  5 0 0 10km
END
```

PaulWessel:
I may be able to fix that since "km" is not a valid unit, but I can imagine some users thinking that any string can be used for that title, and if they give 00101 then we _will_ read that as 101 and not as a string.  This is the problem of allowing variable record lengths with or without a string at the end...

seisman:
I may have a working solution. We can still use `GMT_COL_FIX` but only read `Ctrl->S.n_cols - 2` columns (i.e. the required columns). Then the remaining columns if any are saved in `In->text`.
Then use `sscanf(In->text, "%lf %lf %[^\n]s\n", &xynew[ix], &xynew[iy], event_title);` to read the newX, newY and title from `In->text`. Do you think it's a good way?

PaulWessel:
Yes we can do that.  In fact, in my longer posting earlier I was about to write that scheme but decided to see if the VAR would be enough.  But it cannot tell the trailing text.  I am adding a minor improvement to the parsing (it will check if the trailing chars after the last digit or single period in a string is more than one or (if one) does not match a known unit, then it is deemed text.  But as mentioned this will not catch strings that look like numbers.  So you suggestion is more foolproof.  I will still commit my fix since it applies in general and will not parse 10km as a number.

PaulWessel:
If you want to try the above, note that you will need to do this:

```
n_scanned = scanf (In->text, "%s %s %[^\n]s\n", Xstring, Ystring, event_title); 
if (n_scanned >= 2) { /* Got new x,y coordinates */
    if (ix == GMT_X) {
       gmt_scanf_arg (GMT, Xstring, GMT_IS_LON, false, &xynew[GMT_X]);
       gmt_scanf_arg (GMT, Ystring, GMT_IS_LAT,  false, &xynew[GMTY]);
   }
    else {
       gmt_scanf_arg (GMT, Ystring, GMT_IS_LON, false, &xynew[GMT_Y]);
       gmt_scanf_arg (GMT, Xstring, GMT_IS_LAT,  false, &xynew[GMT_X]);
    }
}

```
To handle proper parsing of 135:45W etc.

PaulWessel:
If you want to take a first crack at this then I can help with any pesky details.

seisman:
OK. I'll look into it later.

PaulWessel:
Maybe something like this?.  This assumes you may give an event title but not new x,y

```
if (In->text) {	/* Must examine the trailing text */
	n_scanned = scanf (In->text, "%s %s %[^\n]s\n", Xstring, Ystring, event_title);
	if (n_scanned >= 2) { /* Got new x,y coordinates and possibly event title */
		if (ix == GMT_X) {	/* Expect lon lat */
			gmt_scanf_arg (GMT, Xstring, GMT_IS_LON, false, &xynew[GMT_X]);
			gmt_scanf_arg (GMT, Ystring, GMT_IS_LAT,  false, &xynew[GMTY]);
		}
		else {	/* Expect lat lon */
			gmt_scanf_arg (GMT, Ystring, GMT_IS_LON, false, &xynew[GMT_Y]);
			gmt_scanf_arg (GMT, Xstring, GMT_IS_LAT,  false, &xynew[GMT_X]);
		}
		if (n_scanned == 2)	/* Got no title */
			event_title[0] = '\0';
	}
	else if (n_scanned == 1)	/* Only got event title */
		strncpy (event_title, In->text, GMT_BUFSIZ-1);
	else	/* Got no title */
		event_title[0] = '\0';
}
```

seisman:
@PaulWessel I followed your first solution, as the second one doesn't work if event title contains several whitespaces (e.g. `Right lateral strike`, `Right` and `lateral` will be read as the newX and newY).

I also added a test to check psmeca with different input columns. Now all tests pass.

PaulWessel:
How do you handle the difference between

lon lat title

and

A longer title

both will give n_scanned 3, no?

seisman:
I don't have a clever way to handle that, so I assume the newX and newY must be given if event title is needed. It's also consistent with older GMT4 or GMT5 behavior.

PaulWessel:
OK, I guess a clever solution would check if the resulting xynew coordinates are NaNs or not, and if they both are then change event_title to be the whole trailing text, e.g.

```
if (gmt_M_is_dnan (xynew[GMT_X]) and gmt_M_is_dnan (xynew[GMT_Y]))
	strncpy (event_title, In->text, GMT_BUFSIZ-1);
```

seisman:
For a title like "10 km", `gmt_scanf_arg` return 10.0000 and 0.000, not NaN. Why?

PaulWessel:
Odd, this is what I was fixing with #2020 and it caught the 10km title from before.

seisman:
A title "Strike slip" also gives me 0.0 and 0.0, not NaN.

PaulWessel:
I have trouble switching to this branch - it complains about conflicts etc... I would approve and then try your case with 3 word title?

seisman:
Try these commands?
```
# delete your local  meca-varcolumns branch
git branch -D meca-varcolumns
# switch to this branch
git checkout meca-varcolumns
```

PaulWessel:
OK now.  Sorry, was in middle of building the full docs.  Please give me a command or two that fails so I can debug.

seisman:
```
gmt psmeca -R239/240/34/35 -JX4c -Ba0 -Sa1c > meca.ps << EOF
239.384 34.556 12.   180     18   -88  5 10 km
EOF
```
The title is "10 km".

PaulWessel:
The FIX cols set 9 cols, and your data record has 9 including the title.  isn't the title in addition to the required cols? IF 10 km is the title then there should be 6 cols only?

PaulWessel:
Probably never mind.  Xcode does not wipe and rebuild the supplements when the source changes, so I need to rebuild the whole thing.  Your code seems to do 6 but my debugger says 9...

seisman:
@PaulWessel Your fix works well except for that newX and newY need to be reverted to 0.0 if they're NaN.

I've made the fix and updated the test script.

seisman:
I still have a question. If `-:` is used, the first two columns are "lat lon". What about the newX and newY columns? I expect to give "newY and newX", but the script below doesn't work: 

```
gmt psmeca -R239/240/34/35 -JX8c -Baf -Sa1c -C1p -: -pdf meca << EOF
34.556 239.384 12.   180     18   -88  5 34.8 239.8 10 km
EOF
```

I still need to give "newX newY". i.e. This works:
```
gmt psmeca -R239/240/34/35 -JX8c -Baf -Sa1c -C1p -: -pdf meca << EOF
34.556 239.384 12.   180     18   -88  5 239.8 34.8  10 km
EOF
```

PaulWessel:
It should by lat lon in both places I think.  I suggest we change the if test on ix to be

if (GMT->current.setting.io_lonlat_toggle[GMT_IN]) {
<do the second situation where Y is X>
else
<do the first situation where X is X

which is cleaner than the ix stuff.

PaulWessel:
Also, the checking of xynew versus zero is no good.  The day may come when you want to give 0,0 as the new coordinates and then you cannot.  Please do this instead:

```
Initialize the xynew array to GMT->session.d_NaN
Remove the reassignment of xynew back to zero
Change the test near L768 to use the gmt_M_is_dnan macro
```

seisman:
I made the change https://github.com/GenericMappingTools/gmt/pull/2018/commits/aa7fa4bdda91897b9eb268da0c1b7bf72e33371d to fix the issue with `-:`. Please help check if it's correct.



> Also, the checking of xynew versus zero is no good. The day may come when you want to give 0,0 as the new coordinates and then you cannot. Please do this instead:
> 
> ```
> Initialize the xynew array to GMT->session.d_NaN
> Remove the reassignment of xynew back to zero
> Change the test near L768 to use the gmt_M_is_dnan macro
> ```

I agree, but it's used for a long time, and I have seen many scripts using `0 0` as newX and newY.

PaulWessel:
Looks good to me.

PaulWessel:
You mean people will place 0 0 when they use -C but don't want another placement instead of not giving the coordinates? Maybe we could handle it under compatibility and say this will break in GMT 7 and warn if we get 0,0 instead of NaN NaN until then?

seisman:
> You mean people will place 0 0 when they use -C but don't want another placement instead of not giving the coordinates? 

Yes, this is exactly what the GMT4 manpage says:
```
Entries in these columns are necessary with the -C option. 
Using 0,0 in columns 8 and 9 will plot the beach ball at the longitude, latitude 
given in columns 1 and 2. The −: option will interchange the order of columns (1,2) and (8,9).
```

Maybe we could handle it under compatibility and say this will break in GMT 7 and warn if we get 0,0 instead of NaN NaN until then?

PaulWessel:
Of, a simpler time when NaNs were just bread eaten by Indians...
I think we can now handle just leaving those coordinates off entirely. Alternatively accept NaN NaN to mean no new location.  And then until GMT 7 warn about 0,0.  The test on zero is not good and GMT-like...

seisman:
Just realize that it's a bad idea to allow "title"-only input. 

The Global CMT project (the most used focal mechanism solution data source) returns:
```
-176.96 -29.25 48 7.68 0.09 -7.77 1.39 4.52 -3.26 26 X Y 010176A       
```
Our current implementation gives 
image
It means breaking many people's results.

PaulWessel:
On the road. Certainly we can handle that case?


SOEST, U of Hawaii at Manoa

seisman:
@carltape This PR fixes the psmeca input issue reported in #2016. It would be good if you can give it a try.

In the new implementation, the new coordinates and event_title are optional. For example, **-Sa** now can accept following input formats:

- `longitude  latitude  depth   strike  dip  rake  mag`
- `longitude  latitude  depth   strike  dip  rake  mag  newX newY`
- `longitude  latitude  depth   strike  dip  rake  mag  event_title`
- `longitude  latitude  depth   strike  dip  rake  mag  newX newY event_title`

carltape:
@seisman I installed this branch and confirm that it resolves the issue you were working on: psmeca allows variable number of columns, as before. Thank you!

In the process of wider testing, I found a lingering issue with psmeca. I will reopen #661

seisman:
@carltape Thanks for your reports and tests! This PS has been merged into 6.0 branch.

PR #2021, opened 2019-11-09
Migration

PaulWessel:

Adds new Resources file that discusses options for migrating older GMT scripts to be compatible with GMT 6, in particular when the old script did not use the gmt program.


doc/rst/source/datasets/dcw.rst
```diff
@@ -0,0 +1,38 @@
+The Digital Chart of the World Data (DCW)
+========================================
+
+figure:: /_images/dcw-figure.png
```
seisman:
```suggestion
.. figure:: /_images/dcw-figure.png
```

PR #2032, opened 2019-11-12
Update INSTALL.md

seisman:
- All the package distributors are providing GMT6 now. Safe to remove the note.
- Mention that the macOS bundle works for macOS >= 10.12.


PaulWessel:
So fink as well?

seisman:
Yes, fink package is available here (https://github.com/fink/fink-distributions/blob/master/10.9-libcxx/stable/main/finkinfo/sci/gmt6.info), and the install instruction is also up-to-date.

PR #2063, opened 2019-11-15
Add documentation on how to debug GMT

PaulWessel:
This is the initial cut of the start of how to get Xcode going.  It adds a new link under _Developer Resources_ on the front index page.  More will be added later. It will go in the 6.0 branch since we want to up asap anyway.  I am having a bit of trouble with some formatting:

- The automatic numbering of steps breaks after one of the figures.
- In some places, the next section is taken as a figure caption and italicized and centered.

Please help make this go away.

Also, let me know if this is too detailed.  I am working on the assumption that most of the people we want to help and learn debugging are not already Xcode wizards so I wanted to do this very slowly.  Ideally I would take it through an an actual bug that users can reproduce.  Perhaps I should have a stop 0 where we simply introduce a bug into a module, then go through now to find where things break and how to use Xcode to diagnose and fix it?


doc/rst/source/debug.rst
```diff
@@ -0,0 +1,129 @@
...
+    cmake -DCMAKE_INSTALL_PREFIX=gmt6 -DCMAKE_BUILD_TYPE=Debug -G Xcode ..
+    xcodebuild
+
+#. After this step you can launch the Xcode application and then use the File->Open menu to
+   open the project file xbuild/GMT.xcodeproj.  After it opens it may looks something like
+   this:
+
+.. figure:: /_images/xcode-1.*
+   :width: 500 px
+   :align: center
```
seisman:
```
.. figure:: /_images/xcode-1.*
   :width: 500 px
   :align: center
```
@PaulWessel Indent the whole figure directive can fix your formatting issue 1.

PaulWessel: Thanks, works fine.  Added last figure before heading home.  This takes us through all the steps to get to top of a given module.  The rest now depends on the bug.

doc/rst/source/debug.rst
```diff
@@ -0,0 +1,132 @@
...
+   "module" on line 119 and this is what it looks like (minus the cursor which is not screen-grabbed!).
+
+   .. figure:: /_images/xcode-6.*
+      :width: 500 px
+      :align: center
+
+#. The tool bar below the code has a a pause-pay button (continue to next stop point), a step-over button (execute
+   next step but do not go *into* a function, thestep-into button (execute next step which may be going into a function)
+   and the step-out button.  Step into the GMT_Call_Module function using the step-into button, then scroll down to
+   around line 10094 and place a stop point there like I did.  You are now about to call your actual C function that
```
seisman:
```
then scroll down to around line 10094 and place a stop point there like I did.
```
May need to click the "pause-play" button to make it stop at line 10094.

doc/rst/source/debug.rst
```diff
@@ -0,0 +1,132 @@
...
+   C function at the heart of the pstext module.  You can now step your way down the code, using step-over to avoid going
+   into the details of GMT sub-functions, or step-into if that is the problem.
+
+   .. figure:: /_images/xcode-8.*
+      :width: 500 px
+      :align: center
+
+
+
+See Also
```
seisman: Remove the "See Also" section?

PR #2081, opened 2019-11-17
cmake: Install openmp library when installing GMT

### LINKED ISSUE #2080: Is OMP_DLL_LOC still needed for Windows?
@joa-quim I saw following codes in the cmake script src/CMakeLists.txt:
https://github.com/GenericMappingTools/gmt/blob/50bcdb885dc30aaf99f043e3d12aeebb814e0a2a/src/CMakeLists.txt#L246-L251

Is OMP_DLL_LOC still necessary for recent cmake releases? If not, I think we can remove these codes.

seisman:
- `set (CMAKE_INSTALL_OPENMP_LIBRARIES TRUE)` installs the `vcomp140.dll` library to the bin directory. 
- `set (CMAKE_INSTALL_SYSTEM_RUNTIME_COMPONENT "Runtime")` specifies the COMPONENT Runtime.

Fixes #2080.

@joa-quim The CI build shows that the vcomp140.dll library is copied to the bin directory correctly.  Please check if it works for you locally, and also check if the the "Runtime" component is correctly set.


joa-quim:
Good try. It works for OMP but fails for the runtime. It does not install ``msvcp140.dll``, only the ``vcruntime140.dll``

joa-quim:
Sorry, I had commented also the line ``include (InstallRequiredSystemLibraries)``. Uncommented it now and all is installed correctly.

seisman:
Glad to know that it works.

Issue #2099, opened 2019-11-20
Modern mode is fooled by PS_MEDIA setting


PaulWessel:
A modern mode plot should not listen to **PS_MEDIA** settings unless we are making a PostScript plot.  This example demonstrates the problem:

```
gmt begin map
  gmt set PS_MEDIA letter
  gmt coast -JU20/9i -BWSen -B2mf1mg2m -R-64.34/-64.19/44.43/44.56 -Df -Gtan -Wthinnest
gmt end show
```

If we comment/remove the gmt set command it works correctly.

seisman:
Closed by #2100.

Issue #497, opened 2019-03-12
Incorrect cropping in modern mode for US units


seisman:
**Description of the problem**

The following commands should generate three identical plots. However, in my computer, map1.png is not correctly cropped. map2.png and map3.png are also slightly different. See attached figures below. 

```bash
gmt begin map1 png
gmt coast -Rg -JH15c -Baf -W1p -A10000
gmt end

gmt coast -Rg -JH15c -Baf -W1p -A10000 -png map2

gmt pscoast -Rg -JH15c -Baf -W1p -A10000 > map3.ps
gmt psconvert -A -P -Tg map3.ps

```

**map1.png:**
![image](https://user-images.githubusercontent.com/3974108/54171451-b5d1aa80-4450-11e9-8dba-54cba909e00d.png)

**map2.png:**
![image](https://user-images.githubusercontent.com/3974108/54171459-bff3a900-4450-11e9-85d1-d5e8462f2340.png)

**map3.png:**
![image](https://user-images.githubusercontent.com/3974108/54171477-cbdf6b00-4450-11e9-9c43-77e5d3eeadc3.png)

**System information**

* Operating system: macOS
* Version of GMT: 6.0.0 latest
* Version of gs: 9.26, installed via homebrew

PaulWessel:
For me, 1 and 2 are identical. 3 is different because classic and modern has different behavior as to which GSHHG data set they pick.  Classic sets **-Dl** (low) by default while modern elects the most reasonable resolution given scale.

seisman:
psconvert debug information for 1:
```
end [DEBUG]: End Workflow.  Session Name = 50080. Directory /Users/seisman/.gmt/sessions/gmt6.50080 removed.
end [DEBUG]: No figure file /Users/seisman/.gmt/sessions/gmt6.50080/gmt.figures - nothing to do
end [DEBUG]: Got session name as map1 and default graphics formats as png
end [INFORMATION]: Process GMT figure queue: 1 figures found
end [INFORMATION]: Processing GMT figure #0 [map1 png ]
end [DEBUG]: psconvert: /Users/seisman/.gmt/sessions/gmt6.50080/gmt_0.ps- -Tg -Fmap1 -A
end (gmtlib_free_tmp_arrays): tried to free unallocated memory
psconvert [DEBUG]: Complete partial PS file /Users/seisman/.gmt/sessions/gmt6.50080/gmt_0.ps-
[...]

Issue #106, opened 2018-10-01
Modern mode: specify paper size and not cropping the figure


seisman:
In modern mode, figures are by default cropped to the minimum required by the image contents, using `psconvert -A`. 

Sometimes, we may want to plot a figure on A4 paper for printing. In this case, we need to specify paper size and set `psconvert` to not crop the figure.

This is what I try:
```
gmt begin map png
gmt set PS_MEDIA=A4 PS_CONVERT=""
gmt basemap -R0/10/0/10 -JX10c -Bafg 
gmt end
```
![map](https://user-images.githubusercontent.com/3974108/46311003-f0284900-c58e-11e8-816b-d625f58e303c.png)
However, you'll see that the right part of the figure is missing. It seems that the origins for the plot are somehow in the middle of the paper, different from the behavior in classic mode. Is it desired or a bug?

PaulWessel:
Hm, I print PDF figures all the time.  The printer knows its paper size and off it goes.  I can even scale the figure if I want to or let it be exactly as is.  it tends to center the figure on the paper. Anyway, what you got is probably not what we want but I think you are trying to defeat how GMT is doing things.

joa-quim:
I've not tested the modern mode much on Win and suspect there are issues in it. Just to check, do you also get this error?
```
gmt basemap -R0/10/0/10 -JX10c -Bafg --PS_MEDIA=A4 --PS_CONVERT="" --MAP_ORIGIN_X=2i --MAP_ORIGIN_Y=2i map pdf
gmt [ERROR]: Shared GMT module not found: basemap
```

seisman:
@joa-quim I think the correct one-line command should be:
```
gmt basemap -R0/10/0/10 -JX10c -Bafg --PS_MEDIA=A4 --PS_CONVERT="" --MAP_ORIGIN_X=2i --MAP_ORIGIN_Y=2i -pdf map
```
Please note the dash symbol before `pdf`.

PaulWessel:
Note this gives what you want but PS:

`gmt basemap -R0/10/0/10 -JX10c -Bafg --PS_MEDIA=A4 --PS_CONVERT="" --MAP_ORIGIN_X=2i --MAP_ORIGIN_Y=2i -ps map`

joa-quim:
Right. You can see that have not been practicing modernity. But it respects the X,Y origin. I mean, origin is at 2,2 inch

seisman:
Two solutions I just find.
```
[...]

PR #2100, opened 2019-11-20
Ignore PS_MEDIA unless classic mode or PostScript format

PaulWessel:
Unless we are making a PostScript plot we must ignore any **PS_MEDIA** settings. Closes issue #2099.

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -6731,7 +6731,6 @@ struct PSL_CTRL *gmt_plotinit (struct GMT_CTRL *GMT, struct GMT_OPTION *options)
 	if (gmt_M_compat_check (GMT, 4) && GMT->current.setting.ps_copies > 1) PSL->init.copies = GMT->current.setting.ps_copies;
 	PSL_setdefaults (PSL, GMT->current.setting.ps_magnify, GMT->current.setting.ps_page_rgb, GMT->current.setting.ps_encoding.name);
 	GMT->current.ps.memory = false;
-	gmt_M_memcpy (media_size, GMT->current.setting.ps_page_size, 2, double);
 
 	GMT_Report (GMT->parent, GMT_MSG_DEBUG, "Running in PS mode %s\n", ps_mode[GMT->current.setting.run_mode]);
 	if (GMT->current.setting.run_mode == GMT_MODERN) {	/* Write PS to hidden gmt_#.ps- file.  No -O -K allowed */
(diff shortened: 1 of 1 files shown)


seisman:
@PaulWessel This PR may break animations. Below is the movie_frame.sh of anim02.
```
#!/usr/bin/env bash
# Main frame loop script
export GMT_SESSION_NAME=$$
export GMT_END_SHOW=off
# Include static and frame-specific parameters
source movie_init.sh
source movie_params_$1.sh
mkdir ${MOVIE_NAME}
cd ${MOVIE_NAME}
gmt begin
# 	Set output PNG name and plot conversion parameters
	gmt figure ${MOVIE_NAME} png E${MOVIE_DPU},A+n+r
	gmt set PS_MEDIA 3.5ix4.167i DIR_DATA /Users/seisman/Gits/gmt/gmt/doc/examples/anim02,/Users/seisman/Gits/gmt/gmt/doc/examples/anim02/anim_02
	width=`gmt math -Q ${MOVIE_WIDTH} 0.5i SUB =`
	gmt grdimage @tut_relief.nc -I+a${MOVIE_COL0}+nt2 -JM${width} -Cmain.cpt 		-BWSne -B1 -X0.35i -Y0.3i --FONT_ANNOT_PRIMARY=9p
	gmt plot -Sc0.8i -Gwhite -Wthin <<< "256.25 35.6"
	gmt plot -Sv0.1i+e -Gred -Wthick <<< "256.25 35.6 ${MOVIE_COL1} 0.37i"
gmt end
# Move PNG file up to parent directory and cd up one level
mv -f ${MOVIE_NAME}.png ..
cd ..
# Remove frame directory and frame parameter file
rm -rf ${MOVIE_NAME}
rm -f movie_params_$1.sh

```

PaulWessel:
OK, will look later today - full day here.  movie may be an exception to that rule.

PaulWessel:
Let's review this issue.  It started because a bad PS_MEDIA setting ended up cropping a plot which seemed very unseemly for modern mode.  However, we did not particularly discuss the pros/cons on this.  Given modern mode is meant to be simple, we don't want to crop valid plots and confuse the user.  Also, if the user wants space around  a plot there is the PS_CONVERT and gmt begin and figure settings you can use.  If the user wants an A4 plot however, then right now the option is to do a PS plot with PS_MEDIA=a4 then convert that manually to PDF, for instance.  Then, there is the case of movie making where we definitively need to specify a page size.  So what do you think? If we revert #2100 then we are back to rouge PS_MEDIA settings.  Or are we? Do we not start a sessions with a clean slate? If someone sets PS_MEDIA in a modern mode, should we really disallow that?

seisman:
There are some related discussions in issues #106 and #497.

> If the user wants an A4 plot however, then right now the option is to do a PS plot with PS_MEDIA=a4 then convert that manually to PDF, for instance. 

Currently it's the only way to have an A4 plot. In #106, I thought I could have an A4 plot by setting PS_MEDIA to A4 and disable the cropping by setting PS_CONVERT to an empty string. However, it gave me a wrong figure, because the plot origin in modern mode is 5 inch by 5 inch away from the lower-left corner of the paper. 

> Do we not start a sessions with a clean slate? 

I don't think it's a good idea. Users may have their own gmt.conf files for specific purposes. 

> If someone sets PS_MEDIA in a modern mode, should we really disallow that?

Are there any technical difficulties allowing PS_MEDIA in modern mode? If PS_MEDIA is set in modern mode, can we just move the plot origin 1 inch by 1 inch (i.e. the origin for classic mode) and disable auto-cropping?

PaulWessel:
We could try to do that, i.e., use PS_MEDIA if set and undo the 5 inch slop.  Since we committed #2100 perhaps you know how to revert that - a bit easier than to redo stuff.

joa-quim:
Agree. If, for unknown reason, user sets the paper size then we should try to cope with it. Fear is that setting is from GMT < 6 habits.

PaulWessel:
We can better explain in what situations PS_MEDIA is useful.  Hi @seisman, could you help with the reverting?

seisman:
#2113 reverts #2100 by `git revert commit-hash`.

#2101 is kept and we can modify it later.

PR #2105, opened 2019-11-20
Use codesign to identify bundle to GateKeeper

PaulWessel:
Not out of the woods yet on this given all the 3rd party things we dont build ourselves.  This will be harder down the road.

Maybe this could go inside an if test checking if the user running the script is pwessel so that others can at least build the bundle for testing?


seisman:
The signed bundle doesn't work for me, so keep this PR unmerged before we find a working solution?

PaulWessel:
That is OK< but I also added an if-test so it only tries to sign if I am building.

PR #2106, opened 2019-11-20
cmake: Cleanup cmake codes for testing batch scripts on Windows

seisman:
Currenly, we don't have a working mechanism to test batch scripts on
Windows, and the variable "DO_EXAMPLES_WIN_BATCH_DISABLED" is never
used.

### DIFF
--- a/doc/examples/CMakeLists.txt
+++ b/doc/examples/CMakeLists.txt
@@ -105,25 +105,12 @@ if (UNIX AND DO_ANIMATIONS)
 endif (UNIX AND DO_ANIMATIONS)
 
 # run examples (test)
-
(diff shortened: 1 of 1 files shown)


joa-quim:
I don't know. We don't but it would be nice to have (I rarely run bash scripts) and if this is removed it will be harder to restore it.

seisman:
You can always call the do_examples.bat script in the doc/examples directory to run all the batch example scripts. If someday we have a solution for testing batch scripts, we can add it back in seconds.

joa-quim:
> we can add it back in seconds.

If we still remember that it ever existed.

seisman:
Take it easy. I can rewrite it from scratch.

Issue #2017, opened 2019-11-09
filter1d/mfilter.sh fails due to the missing remote file


seisman:
The test script uses the data http://scrippsco2.ucsd.edu/assets/data/atmospheric/stations/in_situ_co2/weekly/weekly_in_situ_co2_mlo.csv.

It seems something wrong with their server. The URL was redirected to (note the missing slash after `edu`)

https://scrippsco2.ucsd.eduassets/data/atmospheric/stations/in_situ_co2/weekly/weekly_in_situ_co2_mlo.csv

and of course the file is not there.

There is nothing we can do here. Just open a new issue to remind us the failure. Maybe we should change to another data file for testing purpose?

PaulWessel:
I send them a note - hopefully they can fix it quickly.

seisman:
The test is still failing. Maybe they changed the data file format. Now the file starts with 
```
% Weekly average baseline air CO2 data
% From archive dated 05-Sep-2016 09:05:56
%
1958-03-29, 316.19
1958-04-05, 317.31
1958-04-12, 317.69
1958-04-19, 317.58
1958-04-26, 316.48
1958-05-03, 316.95
1958-05-17, 317.56
1958-05-24, 317.99
1958-07-05, 315.85
1958-07-12, 315.85
1958-07-19, 315.46
```

PaulWessel:
It is always a day of joy when formats change to something that violates the ISO standard.  Now we can add back in awk, etc.  I will waste my time asking them to consider a standard format but since they did not respond to my first email there is probably not much hope they will bother.

PaulWessel:
So GMT actually handles the yyyy-mm-dd string with the missing **T**, and we can use --IO_HEADER_MARKER=% to overcome those comments.  So perhaps this can work well with just that change.

PaulWessel:
I am implementing this now.

seisman:
Fixed by #2128.

seisman:
mfilter.sh is failing again, because they changed the data format.

PaulWessel:
Hm, looks like they are back to what it was before they started messing around.

PR #2109, opened 2019-11-21
test: Fix URL in mfilter.sh

seisman:
The old http URL is still broken, but the https URL works. Probably they're migrating
their website to https.

Closing #2017.


PaulWessel:
Great, they never responded to my email.

PR #2116, opened 2019-11-22
Fix a wrong comment

seisman:
5 inch = 12.7 cm


PaulWessel:
Think I changed my mind halfway through that code comment,

seisman:
We changed it to 1 meter and then changed it back to 5 inch due to a bug of gs 9.27.

PR #2120, opened 2019-11-22
Honor PS_MEDIA if set, and handle the 5 inch margin

PaulWessel:
This PR addresses the #2100 fix by allowing **PS_MEDIA** to be set in modern mode even for non-PostScript formats, in which case we no longer crop and instead honor that setting.  The new solution applies the **PS_MEDIA** setting in effect when the plots starts, and when gmt end comes along to finish the figures we determine if the media size was meant to be automatic or fixed.  If automatic then we crop in psconvert **-A**, otherwise we leave the size as specified.


PaulWessel:
See if you can break this with animation of similar, @seisman, but it looks promising to me, and tests work.

seisman:
It doesn't work if I already have a gmt.conf with PS_MEDIA=a4, for example:
```
gmt defaults -D > gmt.conf
gmt begin map
gmt coast -Rg -JH15c -Baf -W1p -Xc -Yc
gmt end show
```

Is it intentional?

PaulWessel:
There were more checks dealing with this than I remember.  This should fix most of those - but the strange GMT_App_F_stand+_iso+.sh now fails.

PaulWessel:
GMT_App_F_stand+_iso+.s fails because gmtest.in calls gmt set -Du before each test script. When I run that I get

```
#
# GMT 6.0.1 Defaults file
#
# Projection Parameters
#
PROJ_LENGTH_UNIT               = inch
#
# PostScript Parameters
#
PS_CHAR_ENCODING               = Standard+
PS_MEDIA                       = letter
#
# Calendar/Time Parameters
#
TIME_WEEK_START                = Sunday
```


I.e., these are the defaults that differ from SI.  Well, now PS_MEDIA is set to letter and we must honor that, so when that script tries to make two PNGs to include into the plot it sees the PS_MEDIA and realizes these are not PS plots so we must honor PS_MEDIA and it goes wrong.

I can fix this by deleting gmt.conf at the start and then call gmt set -Du after the two PNG figures have been made.  It is a strange script that has a modern session with two figure calls first, then another modern session where we want the media to be set.  I think this is a very weird and special case and my solution is OK.

doc/rst/source/gmt.conf.rst
```diff
@@ -1151,10 +1151,13 @@ fonts can be found in the :doc:`gmt` man page.
 .. _PS_MEDIA:
 
 **PS_MEDIA**
-    Sets the physical format of the current plot paper [a4 (or letter)].
-    The setting only applies to PostScript plots (i.e., classic mode or when using
-    modern mode and selecting PostScript outout).  The following formats (and their
-    widths and heights in points) are recognized:
+    Sets the physical size of the current plotting paper. This is required
```
seisman: This paragraph is a little hard to understand. Could you rephrase it?

PaulWessel: OK, done

PaulWessel:
With the latest fix all tests pass for me.  PS_MEDIA description updated and now applies to any media if given, but is required for PostScript.

PR #2121, opened 2019-11-22
doc: Change link name from "GMT Defaults" to "GMT Default Settings (gmt.conf)"

seisman:
screenshot:
image


joa-quim:
It is a bit long. We can drop the word ``GMT`` as well as the ``File`` in ``Grid File Format Specifications``

seisman:
OK.

PR #2127, opened 2019-11-22
Better check for .extension in begin

PaulWessel:
The initial check did not care about the position of the period.  Addresses #2082.


seisman:
We decided not to strip off any extensions and let users decide the names they like. This PR will be closed without merging.

PR #2131, opened 2019-11-23
doc: Clarify difference between png and PNG

seisman:
For a long time, I thought the diffrence between png and PNG is that
PNG support tranparency while png doesn't. But that's not true.
The only difference is that the background is opaque (white) or transparent.

In this PR, I make small changes to clarify that.

### DIFF
--- a/doc/rst/source/begin.rst
+++ b/doc/rst/source/begin.rst
@@ -73,27 +73,19 @@ Supported Graphic Formats
 
 .. _tbl-formats:
 
-    +--------+-----------------------------------------+
-    | Format | Explanation                             |
-    +========+=========================================+
-    |  bmp   | Microsoft Bit Map                       |
-    +--------+-----------------------------------------+
-    |  eps   | Encapsulated PostScript                 |
-    +--------+-----------------------------------------+
-    |  jpg   | Joint Photographic Experts Group Format |
-    +--------+-----------------------------------------+
-    |  pdf   | Portable Document Format [Default]      |
-    +--------+-----------------------------------------+
-    |  png   | Portable Network Graphics (opaque)      |
-    +--------+-----------------------------------------+
-    |  PNG   | Portable Network Graphics (transparent) |
-    +--------+-----------------------------------------+
-    |  ppm   | Portable Pixel Map                      |
-    +--------+-----------------------------------------+
-    |   ps   | Plain PostScript                        |
-    +--------+-----------------------------------------+
-    |  tif   | Tagged Image Format File                |
-    +--------+-----------------------------------------+
+====== ===================================================
+Format Explanation
+====== ===================================================
+bmp    Microsoft Bit Map
+eps    Encapsulated PostScript
+jpg    Joint Photographic Experts Group Format
+pdf    Portable Document Format [Default]
+png    Portable Network Graphics (opaque background)
+PNG    Portable Network Graphics (transparent background)
+ppm    Portable Pixel Map
+ps     Plain PostScript
+tif    Tagged Image Format File
+====== ===================================================
 
 Examples
 --------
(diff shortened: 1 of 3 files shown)


joa-quim:
Transparency is more than one thing. It can be a 4th layer with transparency or objects lay down over others and a certain transparency. Oddly I get transparency with any raster format, but only -TG sets a transparency layer. However, it also is having its issues. The images below are adapted from a Julia example. First is done with **-Tg** and second with **-TG**. There is an issue with this latter because it becomes completely transparent at 50%, not 100. 

The julia script
```
rect = [0.5 0.5; 0.5 9; 2.5 9; 2.5 0.5; 0.5 0.5];
plot([1.5 1.5; 1.5 8; 8 8; 8 1.5; 1.5 1.5], region=[0 10 0 10], fill=:green)
text!(mat2ds([0.2 5], ["oOoOoOoOoOoOoOoOo"]), region=[0 10 0 10], font=28, justify=:LM)
plot!(rect, lw=0.5, fill=:blue)
plot!([3 0].+rect, lw=1, ls="--", fill=:blue, transparency=30)
plot!([6 0].+rect, lw=2, lc=:red, fill=:blue, transparency=70, show=true, savefig="transp_1.png")
```
and the images. Let me repeat, even if I use jpeg (``savefig="transp_1.jpg``) I still get transparency.
Attached is also the PS file.

transp_1

Damn it, the fig bellow is not what I see when I open this image on Windows.

transp_2



NOTE, this file ``transp.zip``, is actually a ``transp.rar`` but I had to rename it to cheat Github.
[transp.zip](https://github.com/GenericMappingTools/gmt/files/3882722/transp.zip)

PaulWessel:
So this is probably a bug then: If I run

gmt psconvert -Tg transp.ps -A -Fno

you expect it to not be transparent, but because psconvert detects there are transparency commands in the postscript:

psconvert [INFORMATION]: PS file with transparency must be converted to PDF before creating PNG

we first compute a PDF (which will have transparency) and then we convert the PDF to PNG which inherits the transparency.

The first plot is true transparency (opaque pixels) as seen in Photoshop:
transp
while the second approach (-Tg) also gives transparency inherited from the PDF but has opaque background pixels:
notrans

I assume the solution you want is to only compute the intermediate PDF if -TG was given, otherwise go directly to the given driver?

joa-quim:
I don't think there is any bug. I was being fooled by a buggy windows image viewer. This is what I see with the default viewer
Capture2

but any other viewer that I tried showed the correct thing. 

What is not correct is saying that **-TG** only sets the background to transparent. In fact what it does is ALWAYS create a transparency layer. You can check that with Matab's ``imfinfo`` function. For example in the case of this rectangles example the alpha layer is this

Capture

One can also confirm with Matlab that even this, that has no background, has an alpha layer. We don't see its effect because the transparency is 100%, but it's in the file.
```
gmt coast -R14/21/16/20 -JM10 -Gbrown -PNG map
```

PaulWessel:
I guess there are no circumstances where the user uses transparency options in GMT but dont want transparency in the final product? So we should continue to do what we do, i.e., make that intermediate PDF whenever we discover PS transparency.

joa-quim:
Yes, we should continue doing as now but the description is not correct. Instead of
```
png    Portable Network Graphics (opaque background)
PNG    Portable Network Graphics (transparent background)
```
we must have
```
png    Portable Network Graphics
PNG    Portable Network Graphics (with transparency layer)
```

and somewhere also explain that **-t** gives transparency to all raster formats but only PNG ads a alpha layer that can be used if the image is to be combined with other images in image processing softwares.

PaulWessel:
OK, that seems pretty clear to me. Any problems with this @seisman? If not I can update the text.

PR #2135, opened 2019-11-23
doc: Fix the description of psmeca and pscoupe -Sm|d|z

seisman:
As I understant it, a full seismic moment tensor can be decomposed into three
parts: isotropic (ISO) + double couple (DC) and compensated linear vector dipole (CLVD).
Of course, there are many different ways to decompose a moment tensor, but the
ISO + DC + CLVD decomposition is the most simple/used one.

The old documentation of psmeca says:

```
Use -Sm to plot the Global CMT seismic moment tensor with zero trace.
Use -Sd to plot only the double couple part of moment tensor.
Use -Sz to plot the anisotropic part of moment tensor (zero trace).
```

For a long time, I was confused about the difference between **-Sm** and **-Sz**. After reading the source codes and some tests, I think the above documentation is misleading and even wrong.
I believe the correct description is:

- **-Sm** plots the full moment tensor (ISO + DC + CLVD)
- **-Sz** plots the anisotropic part of moment tensor (DC + CLVD)
- **-Sd** plots the double couple part of the moment tensor (DC)

For the Global CMT solution, the ISO part is always zero (i.e. with zero trace),
**-Sm** and **-Sd** gives the same result.

This PR fixes the description of **-Sm|d|z**.

@carltape I'd appreciate it if you could help check if my understanding is correct.

### DIFF
--- a/doc/rst/source/supplements/seis/explain_meca_-S.rst_
+++ b/doc/rst/source/supplements/seis/explain_meca_-S.rst_
@@ -80,16 +80,15 @@
 
 **-Sm\|d\|z**\ *scale*\ [**+f**\ *font*][**+j**\ *justify*][**+o**\ *dx*\ [/*dy*]]
 
-    Seismic moment tensor (Global CMT, with zero trace). *scale* adjusts
+    Seismic moment tensor. *scale* adjusts
     the scaling of the radius of the "beach ball", which will be
     proportional to the magnitude. *scale* is the size for magnitude = 5
     (that is scalar seismic moment = 4.0E23 dynes-cm).
     **-T**\ *0* option overlays best double couple transparently.
-    Use **-Sm** to plot the Global CMT
-    seismic moment tensor with zero trace. Use **-Sd** to plot only the
-    double couple part of moment tensor. Use **-Sz** to plot the anisotropic
-    part of moment tensor (zero trace). The color or shade of the
-    compressive quadrants can be specified with the **-G** option. The color
+    Use **-Sm** to plot the full seismic moment tensor (i.e. isotropic + double couple + CLVD parts).
+    Use **-Sd** to plot only the double couple part of moment tensor.
+    Use **-Sz** to plot the anisotropic part of moment tensor (zero trace, i.e. double couple + CLVD parts).
+    The color or shade of the compressive quadrants can be specified with the **-G** option. The color
     or shade of the extensive quadrants can be specified with the **-E** option.
     Append **+f**\ *font* to change the font of the text string;
     append **+j**\ *justify* to change the text location relative to the beachball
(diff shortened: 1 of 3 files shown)


carltape:
@seisman I will try to respond soon. It is not a trivial response, partly because I need to familiarize myself with some of the psmeca options that I have not used.

seisman:
@carltape Thanks. Please let me know if I made any mistakes here.

carltape:
@seisman Thanks for your pursuit. I created another example to check how -Sm,-Sz,-Sd behaves with a reference set of full moment tensors. The figure file can found from [this directory](https://github.com/carltape/compearth/tree/master/momenttensor/plot/gmt/beachball_plots)
I have also attached the file here:
[lune_hammer_iplot2_lplot1_kplot1_gmt600pr2067_psmeca_flags.pdf](https://github.com/GenericMappingTools/gmt/files/3929062/lune_hammer_iplot2_lplot1_kplot1_gmt600pr2067_psmeca_flags.pdf)

The left plot is for -Sm, the middle plot is for -Sz, and the right plot is for -Sd. The left plot looks good, as we previously established. The middle plot also looks good: for a fixed lune longitude, the beachball should be the same. The two isotropic points are not deviatoric, so technically these could be absent. (I think the behavior here is fine.)

In the right plot, all full moment tensors in the lune interior have the same closest-DC as shown. However on the lune boundary we see DCs with orientations different from those in the interior. The input moment tensors on the boundary are all crack tensors and hence have a rotational symmetry axis: On the left side, it is a vertical T axis; on the right side, it is an east-west P axis. There is no unique closest-DC for these crack tensors. If a DC on the right lune boundary is rotated 90 deg about the P axis, then it agrees with the DCs in the interior, and since the P axis was a rotational symmetry for the crack tensor, the rotation does not change the distance from the DC to the crack. A DC on the left boundary can likewise be rotated 90 deg about the T axis.

In summary, the plotting behavior is correct for this example.

Now for your original question, which is about the documentation. My recommendation would be this:

```
-Sm plots the (full) moment tensor
-Sz plots the deviatoric part of the moment tensor (zero trace)
-Sd plots the closest double couple moment tensor (zero trace and zero determinant)

Global CMT moment tensors are deviatoric, so -Sm and -Sz will produce the same result,
but -Sd will not, unless the input moment tensor is already a double couple.
```

As an aside, the use of percentages of components of moment tensors (like ISO, DC, CLVD) is widely used in the literature, but I don't think it's the right concept to be promoting. I have some notes available on [this page](https://sites.google.com/alaska.edu/carltape/home/research/beachball), which expand upon some published material.

seisman:
@carltape Thanks for your detailed explanation and test. I've opened PR #2215 to improve the description and also include your new test into GMT.

Issue #2104, opened 2019-11-20
Scripts for GMT4-compatibility don't work


seisman:
We provide several scripts to allow users to call modules like "grdimage" without the "gmt" prefix.
- gmt_aliases.csh
- gmt_functions.sh
- gmt_links.sh

These scripts don't work because `gmt --show-modules` now returns modern mode module names. Thus, these scripts provide "basemap" instead of "psbasemap".

PaulWessel:
Might we need a gmt --show-classic-modules option to solve this problem?

seisman:
I think so.

seisman:
Closed by #2137.

PR #2137, opened 2019-11-23
Add new --show-classic option to gmt

PaulWessel:
This is needed so that the backwards compatibility helper scripts gmt_aliases.csh, gmt_functions.sh and gmt_links.sh can do their things for the classic (not modern) functions.
Closes issue #2104.

### DIFF
--- a/doc/rst/source/gmt.rst
+++ b/doc/rst/source/gmt.rst
@@ -85,6 +85,9 @@ If no module is given then several other options are available:
 **--show-citation**
     Show the citation for the latest GMT publication.
 
+**--show-classic**
+    List classic module names on stdout and exit.
+
 **--show-cores**
     Show number of available cores.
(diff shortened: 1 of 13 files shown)


seisman:
I can see some modern-only modules like *begin*, *end*, *movie*.

PaulWessel:
Weird, I just tested that and we are skipping them.  You are doing gmt --show-classic right?

seisman:
```
$ gmt --show-classic
psbasemap
begin
blockmean
blockmedian
blockmode
clear
psclip
```

PaulWessel:
OK fixed, but not getting any supplement names yet.

PaulWessel:
Should work now - tested and works for me.  We exceeded 32 bytes in the new function name that prints the classic forms.

PR #2145, opened 2019-11-24
Better pscoast check for -E region

### LINKED ISSUE #2142: gmt pscoast -= return -R0/0/0/0
By a change I typed `gmt pscoast -=` in the terminal, and get `-R0/0/0/0` returned.
```
$ gmt pscoast -=   
-R0/0/0/0
```

What's this? Is it some hidden feature or a bug?

PaulWessel:
There was no check that **-E** was needed to get the region. Discovered when this command was run

`gmt pscoast -=
`

This closes #2142. Also, since **-=**_list_ is an implemented yet undocumented option, it should at least receive the minimum parsing check of making sure _list_ was given and that it is a valid file.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -12663,7 +12663,7 @@ struct GMT_CTRL *gmt_init_module (struct GMTAPI_CTRL *API, const char *lib_name,
 	if (options && strstr (mod_name, "legend") && (opt = GMT_Find_Option (API, 'D', *options)) && strchr ("jJg", opt->arg[0])) /* Must turn jr into JR */
 		required = "JR";
 
-	if (options && gmt_M_compat_check (GMT, 5) && !strncmp (mod_name, "pscoast", 7U) && (E = GMT_Find_Option (API, 'E', *options)) && (opt = GMT_Find_Option (API, 'R', *options)) == NULL) {
+	if (options && !strncmp (mod_name, "pscoast", 7U) && (E = GMT_Find_Option (API, 'E', *options)) && (opt = GMT_Find_Option (API, 'R', *options)) == NULL) {
 		/* Running pscoast -E without -R: Must make sure any the region-information in -E is added as args to new -R.
 		 * If there are no +r|R in the -E then we consult the history to see if there is an -R in effect. */
 		char r_code[GMT_LEN512] = {""};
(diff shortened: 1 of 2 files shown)


seisman:
So `gmt pscoast -EUS` returns `-R172.43611145/293.051109314/18.9098587036/71.3906860352`? Is it a feature? If yes, I don't see where this feature is documented. And it doesn't work in modern mode.

```
gmt begin map
R=`gmt coast -EUS`
echo $R
gmt end
```

PaulWessel:
Under the explanation for -E it says " If neither -J nor -M are set then we just print the -Rwesn string.".
I suspect under modern mode we automatically add a default -JQ setting in gmt_init_modules so there is never a case where -J is not given in modern mode.  We would have to make a check that if coast and no -R -J and -E then no -R -J is required.

PaulWessel:
How to make this feature work in modern mode:  If you give 

`gmt coast -EFR`

then you are not asking for painting/clipping (-G) or drawing (-W) or dumping (-M).  In that case, no -R -J should be added and the result should be printing of a -R string.
Otherwise there will be at least one of -W, -G, -M and region and projection are needed.

PaulWessel:
Should work now, @seisman.

seisman:
```
gmt begin
gmt coast -EUS
gmt end
```
prints `-R0/0/0/0`.

PaulWessel:
Hm, will check again. 

gmt coast -EUS -pdf map
-R172.43611145/293.051109314/18.9098587036/71.3906860352

and thought the other case ran too.

seisman:
You one-liner command also gives me `-R0/0/0/0`.

PaulWessel:
It is a bit puzzling.  Your case (now) works for me, as does the classic `gmt coast -EUS -pdf map`.  However, `gmt coast -EUS -pdf map` gives the zeros on the command line but works inside Xcode.  I am not sure if it gets mixed up with which shared libgmt it finds...

PaulWessel:
Do you still get 0/0/0/0 after the latest commit, @seisman?

seisman:
The "begin" & "end" script gives the correct -R, but the one-liner still gives -R0/0/0/0.

PaulWessel:
I suggest we declare victory on this issue.  Both the classic example and the modern mode example works correctly.  The one-liner fails, but it is not clear to me that a special syntax (one-liners) created to simplify making a one-command plot should be used to extract a region.  We are even passing -pdf to specify a format so it is very strange to start with.  I think we can live with that, no?

PR #2146, opened 2019-11-24
Failed to pass psconvert args for some cases

PaulWessel:
When specifying background color fill via **PS_CONVERT** settings we must check if the **+n** modifier is set (no cropping) so that in cases where we are not allowed to crop via **-A** (which is when a fixed media size has been specified) we are still allowed to pass the **-A** arguments to psconvert when gmt end converts the plots to the final products.  As it were, args like **-A+g**_gray_**+n** did not get passed because a **PS_MEDIA** setting was in effect (e.g., in movie) but the **-A** did not want to clip (hence **+n**), just set the fill (**+g**).

The two PS files are updated as a consequence of #2141.

### DIFF
--- a/doc/examples/anim02/anim02.ps
+++ b/doc/examples/anim02/anim02.ps
@@ -1,17 +1,16 @@
 %!PS-Adobe-3.0
-%%BoundingBox: 0 0 252 300
-%%HiResBoundingBox: 0 0 252.0000 300.0240             
-%%Title: GMT v6.0.0_1ded948 [64-bit] Document from grdimage
+%%BoundingBox: 0 0 252 301
+%%HiResBoundingBox: 0 0 252.0000 300.0240
+%%Title: GMT v6.0.1_3015954_2019.11.23 [64-bit] Document from grdimage
 %%Creator: GMT6
 %%For: unknown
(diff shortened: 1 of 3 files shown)


seisman:
I thought this should work, but actually not. I have to use `A+glightred+n` to paint the BoundingBox.
```
gmt begin map pdf
gmt set PS_MEDIA a4 PS_CONVERT A+glightred
gmt coast -Rg -JH15c -Baf -W1p
gmt end show
```

PaulWessel:
The **-A** option initially only dealt with cropping, but we then added modifiers that not all deal with cropping, such as painting the bounding box. To not ask for cropping but paint we added **+n**.  None of this is ideal, and I wonder if we can avoid +n altogether since if **+g** should only be applied after we determine the bounding box anyway, whether given or via crop.

seisman:
Classic mode still needs **+n**, right?

PaulWessel:
Yes, and movie -G also issues +n.

PaulWessel:
I think wether or not we can come up with a scheme that makes **+n** obsolete is separate from the bug fixes in this PR.  I think these are OK and fixed a problem.

Issue #2143, opened 2019-11-24
Update gmtdefaults usage 


seisman:
The usage and manapge says the **-D** is optional, but `gmt defaults` alone only print its usage. So **-D** is required. Right?

I'd like to fix it after you confirm that it's a bug.

PaulWessel:
More than one thing here.  I think the **-Du**|**s** derives from the olden days when we wrote inches or cm as length units for things like **MAP_ORIGIN_X**, but now we always write points.  E.g.,

```
gmt defaults -Du > u
gmt defaults -Ds > s
diff u s

```
returns nothing.  So what is the point of **-D** at all?  Maybe better to remove it (but accept it as backwards compatibility) and not have any options at all.

seisman:
This is what I get:
```
diff u s
121c121
< PROJ_LENGTH_UNIT               = inch
---
> PROJ_LENGTH_UNIT               = cm
127c127
< PS_CHAR_ENCODING               = Standard+
---
> PS_CHAR_ENCODING               = ISOLatin1+
135c135
< PS_MEDIA                       = letter
---
> PS_MEDIA                       = a4
150c150
< TIME_WEEK_START                = Sunday
---
> TIME_WEEK_START                = Monday
```

PaulWessel:
I think that was the intention but clearly it depends on gmt.conf setting somewhere.  If you have a gmt.conf set to US then they are identical.  So a bug I would say.

PaulWessel:
If there is a gmt.conf that has all those 4 settings already, then regardless what they are the gmt defaults **-D** will read those and write out the same thing, hence my diff gave nothing.  If the gmt.conf misses one or more of those, then the output will be different but only for that one.  So not sure what is a good solution here, other that to say **-D** is required.  I think the real fix would be to either ignore the gmt.conf (but that is processed during session creation) or re-initialized to either US or SI depending on -D and then print that out.

PR #2150, opened 2019-11-24
Allow gmtdefaults to run without args

PaulWessel:
gmt defaults without arguments are supposed to print the users current defaults, but not working.  Also, **-D** should print the system defaults (SI or US) but it did not reset to those values before reporting . This PR fixes both issues.  Closes issue #2143.

### DIFF
--- a/doc/rst/source/gmtdefaults.rst
+++ b/doc/rst/source/gmtdefaults.rst
@@ -21,7 +21,7 @@ Description
 
 **defaults** lists all the GMT parameter defaults if the option
 **-D** is used. There are three ways to change some of the settings: (1)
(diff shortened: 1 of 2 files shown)


seisman:
The gmtdefaults manpage says:
```
Your currently active defaults come from the gmt.conf file in the current working directory, if present; else from the gmt.conf file in your home directory, if present; else from the file ~/.gmt/gmt.conf if present; else from the system defaults set at the time GMT was compiled.
```
It should also be updated for modern mode.

PaulWessel:
Does this help?

seisman:
Looks good.

PR #2160, opened 2019-11-26
Add option -J to surface.

joa-quim:
Like grdedit -J, the info passed here via -J is only used to set referencing information the grid's metadata. Ofc, this works only for formats that support it, namely *netCDF* and *GeoTIFF*.
Ideally one should be able to fetch this info when reading .gmt files (and future OGR formats) and use it directly. And other grid producers should have this added too.

### DIFF
--- a/doc/rst/source/surface.rst
+++ b/doc/rst/source/surface.rst
@@ -17,6 +17,7 @@ Synopsis
 |SYN_OPT-R|
 [ |-A|\ *aspect_ratio*\ \|\ **m** ]
 [ |-C|\ *convergence_limit*\ [%] ]
+[ |-J|\ *parameters* ]
 [ |-L|\ **l**\ *lower* ] [ **-Lu**\ *upper* ]
 [ |-M|\ *max_radius*\ [**u**] ]
 [ |-N|\ *max_iterations* ]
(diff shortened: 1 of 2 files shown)


seisman:
@joa-quim Just to confirm that, this PR is targeted for master branch, not 6.0 branch?

PaulWessel:
Hm, it could go 6.0 if we think of it as an omission.

joa-quim:
This time it went to master on purpose because it's a new feature (and xyz2gtrd, triangulate, etc need it too). But since you bring this, I think we should discuss the 6.0 *vs* master now. It only makes sense to keep committing to 6.0 if we intend to release a 6.0.1 only with bug fixes and later (?) a 6.1.0. 
I think this is too much of releases in a short time. For me, we should stop committing in 6.0 and probably make 6.1.0 the next release, or then call it 6.0.1 and make the little sin of including the current new features in master.

PaulWessel:
See my recent (5 minutes ago) posting in the Cheese Shop.  We must do 6.0.1 I think, etc.

joa-quim:
Going home now

seisman:
We should try out best to follow the [Semantic Versioning](https://semver.org/) rules, i.e. bug-fixes only for 6.0.1 release, and new features for 6.1.0 release.

joa-quim:
Discourse seems to be down

PaulWessel:
Yes, discourse has a plugin that monitors if the admin person responsible for the installation is on leave and it determined that since Colby has the week off that it was a good time to act up...  I am checking with IT if they are aware of what to do to restart, etc.  gmtserver is otherwise up.

PR #2167, opened 2019-11-27
Allow -Ls<string> to specify a fixed label

PaulWessel:
It is often desirable to place a fixed label on all movie frames but there was no option to do that, meaning you would have to add a gmt text call somewher.  Now we can use **-Ls**_string_[_modifiers_].


seisman:
@leouieda Any idea why the Travis CI is blocking every PR to the master, even though we have completely disabled it?

PaulWessel:
No, not sure, maybe @leouieda knows?

leouieda:
We have setup branch protection rules (https://github.com/GenericMappingTools/gmt/settings/branches) to avoid merging PRs accidentally that break everything. `master` had Travis as a requirement still. I removed it and added Azure instead. So merging will be "blocked" if Azure builds fail (admins can still overwrite this but there is a box to tick). The same is enabled for the `6.0` branch. You can also specify which specific Azure builds need to pass if there is some that are known to fail.

PR #2170, opened 2019-11-27
Long verbose message if we cannot find the standard font listing

PaulWessel:
Added long verbose message if _psl_init_fonts_ function does not find any custom fonts.

### DIFF
--- a/src/postscriptlight.c
+++ b/src/postscriptlight.c
@@ -3159,8 +3159,7 @@ static int psl_init_fonts (struct PSL_CTRL *PSL) {
 		PSL->internal.N_FONTS = i;
 	}
 	else {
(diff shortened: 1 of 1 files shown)


src/postscriptlight.c
```diff
@@ -3158,6 +3158,10 @@ static int psl_init_fonts (struct PSL_CTRL *PSL) {
 		fclose (in);
 		PSL->internal.N_FONTS = i;
 	}
+	else {
+		PSL_message (PSL, PSL_MSG_LONG_VERBOSE, "No PSL_custom_fonts.txt found\n");
```
joa-quim: I'm afraid this can do more harm than good. ``PSL_custom_fonts.txt`` is not a file users normally have or should have. So seeing a warning that something was not found makes think in a somewhat broken installation.

PaulWessel: it is not a warning, it is information under -Vl.

PR #2171, opened 2019-11-27
doc: Update image examples

seisman:
- Update the description of -G option
- Update some examples to use remote cache files

### DIFF
--- a/doc/rst/source/image.rst
+++ b/doc/rst/source/image.rst
@@ -35,8 +35,6 @@ Synopsis
 Examples
 --------
 
-.. include:: explain_example.rst_
-
 .. include:: oneliner_info.rst_
 
 To plot the remote image needle.jpg, scaling it be 7 cm wide (height is scaled
--- a/doc/rst/source/image.rst
+++ b/doc/rst/source/image.rst
@@ -52,12 +52,12 @@ To only plot its red band as gray shade, use::
 
     gmt image @needle.jpg+b0 -Dx0/0+w7c -pdf plot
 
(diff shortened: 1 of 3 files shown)


seisman:
@PaulWessel Is there a one-bit raster image in the remote server, to be used in the example?

PaulWessel:
No but I just made one from vader:

vaderbits

I can add it to the cache.  It is a 1-bit PNG.

seisman:
Looks good. I think it can be used in many examples.

joa-quim:
I'm curious on what an example using -G will do. I always get confused when I see ``1-bit`` image. And that so because the black-n-white (BW) images are often made of one byte images but where only the first bit is used. For example in Matlab *logicals* can have only 0 or 1 values but they use a one byte type variable. And ML has the packbits function to use true 1-bit.

Will -G work for both true 1-bit and BW images?

seisman:
@PaulWessel Your 1-bit png image doesn't work with the example in the image manpage:
```
gmt psimage 1_bit.ras -Gbrown+b -Gred+f -Dx0/0+w1c+n5 > image.ps
```

PaulWessel:
I think @joa-quim mentioned this that GDAL will return 1-bit images as bytes.  The original code case probably assumes we have 8 bits inside each byte.  This was how it was when we only supported Sun raster files and we read those files ourselves.
I added vader1.png to the cache so should be updated on the server soon.

PaulWessel:
See #2186 and hopefully everything is addressed.

PR #2178, opened 2019-11-28
Remove obsolete syntax -E+r|R<inc> for DCW data

### LINKED ISSUE #2177: coast -E+r|R is not explained in the manpage

seisman:
Closes #2177.

### DIFF
--- a/src/gmt_dcw.h
+++ b/src/gmt_dcw.h
@@ -25,13 +25,13 @@
 
 /*!
  * \file gmt_dcw.h
- * \brief Definitions for using the DCW 
+ * \brief Definitions for using the DCW
  */
 
 #ifndef GMT_DCW_H
 #define GMT_DCW_H
 
-#define DCW_OPT "<code1,code2,...>[+l|L][+g<fill>][+p<pen>][+r|R[<incs>]]"
+#define DCW_OPT "<code1,code2,...>[+l|L][+g<fill>][+p<pen>]"
 
 enum GMT_DCW_modes {
 	GMT_DCW_REGION	= 1,


seisman:
pscoast is the only module which uses DCW data directly. I also don't have a solution.

PaulWessel:
The **-R** parser (parse_R_option) also deals with DCW data since **-R**FR**+r**2 means reading the France polygon and rounding to multiples of 2.  But with **-R** being a common option I am not sure what module could be given the duty of reporting **-R**.  But there is this simple solution:

Since every module that accepts **-R** can do what pscoast **-E**... does, it is probably no longer important to get a **-R** string via pscoast to pass to another module.  The only reason is backwards compatibility and we can do that by accepting this PR.  Most users will rather do

`grd grdimage -RFR+r2 ...`

than

```
R=`gmt pscoast -EFR+2`
gmt grdimage $R ...
```

So shall I just accept your patch and we move on?  Are there any circumstances where we must give the user a **-R** string like that? I think the only other modules that return **-R** strings are gmtinfo and grdinfo.

PaulWessel:
I've just implemented **+r|+R**_incs_ in gmtinfo and one issue comes up.  I would like to declare a bug in **+r**.  The rounding to nearest multiple of _dx_ is nice, but in the case that John brought up, the min is zero, and no **+r** will change that to less than zero.  **+R** will, but then we don't get nice rounded values.  I currently do stuff like this (e.g., for west):

`west  = floor (west  / Ctrl->I.delta[XLO]) * Ctrl->I.delta[XLO];`

I think it should instead be something like this:

```
new_west  = floor (west  / Ctrl->I.delta[XLO]) * Ctrl->I.delta[XLO];
west = (west == new_west) ? new_west - Ctrl->I.delta[XLO] : new_west;
```
I.e., the rounded region can never equal the given region if it already is an exact multiple.
Alternatively, this is a new feature and we need another modifier letter that is unique in both -I and -R.  They don't use many modifiers so something like +e for extend could be used.  So that would mean

**+r**_incs_: Convert region to multiples of given incs, rounded outward.
**+e**_incs_: Same, but any limit already a multiple of inc gets shifted by one more inc
**+R**_incs_: Simply add the incs to the determined limits.

Thoughts?

seisman:
In the worst case, users can use `gmt coast -EFR -M | gmt info -Ie` to have the -R string.

PaulWessel:
True, but isn't OK to just leave the capability there quietly and remove any trace of it from the documentation? We can even add a compatibility check around it so if 7 then complain.

seisman:
Yes, I agree.

PaulWessel:
OK, I will accept your change, but do you also agree with +r has a bug or should we add +e? I can see the usefulness of all three ways so perhaps adding new +e is the best solution (in master)?

seisman:
I missed that comment. I think +e is necessary.

seisman:
Give @joa-quim more time before merging this PR.

PaulWessel:
OK, will wait for @joa-quim but starting a new feature request for **+e**.  There are some questions to discuss there as well.

joa-quim:
I think it's OK to merge this. This syntax is documented in GMT5.4.5 bu the ``DCW_OPT`` define is only used in pscoast, so if deprecated it should not be documented.

seisman:
OK, merging.

PR #2183, opened 2019-11-29
Detect when images are of 1-bit type and add a cmap

joa-quim:
With this PR, from Julia I get

```
julia> I = gmtread("C:\\v\\vader.png");

julia> I.colormap
8-element Array{Int32,1}:
   0
   0
   0
   1
 255
 255
 255
   1
```
but I'm not sure if this translate to the right GMT cpt needed for -G option work.

### DIFF
--- a/src/gmt_gdalread.c
+++ b/src/gmt_gdalread.c
@@ -538,8 +538,18 @@ GMT_LOCAL int populate_metadata (struct GMT_CTRL *GMT, struct GMT_GDALREAD_OUT_C
 		Ctrl->ColorMap[j++] = -1;
 	}
 	else {
-		Ctrl->ColorMap = NULL;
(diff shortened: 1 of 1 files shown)


PaulWessel:
I think you added obsolete variables in this commit.  Remember that in gmt_resources.h we have
```
#if def GMT_BACKWARDS_API
	int *ColorMap; 
	int nIndexedColors; 
#endif
```

The approved names are 

```
	int *colormap;                  /* Array with color lookup values */
	int n_indexed_colors;           /* Number of colors in a paletted image */

```

PaulWessel:
Sorry, never mind, yours are in the Ctrl structure, not the GMT_IMAGE one.  Anyway, found a few problems related to BACKWARDS_API of mine so fixing those now.

PR #2190, opened 2019-11-30
Fix label attributes in psevents and transparency in psxy

PaulWessel:
In testing psevents there were a few issues that needed addressing:
1. There was no way to have a different duration for the symbol label. This was addressed with a missing modifier **+l**dt for **-Et**.  The default is still the same duration.
2. The label plotting was simplified since there is only the issue of transparency during the rise and the fade phases.
3. When psxy was called to plot a transparent symbol it failed to reset transparency at the end, thus messing up the transparency of any overlay with transparency that followed.

Documentation has been updated and all tests pass as before (since the tests did not check the above).


PaulWessel:
Not to merge yet since still some issues I need to fix.

PaulWessel:
Good to go, hopefully @joa-quim or @seisman are working today.

PR #2193, opened 2019-12-01
Deprecate modifiers +p- and +g- for vector head attributes

PaulWessel:
The trailing - is not needed to turn off vector head outline and fill, respectively, but we will accept them as backwards compatible versions of **+p** and **+g** (i.e., no arguments turns off the outline and fill).


PaulWessel:
Well, it is the kind of thing we want to remove from the record as soon as possible, hence 6.  Looks like we will do a 6.1 early next year anyway.

Issue #2196, opened 2019-12-02
GMT_Encode_Options and modern mode troubles


PaulWessel:
The GMT mex was built and tested for classic mode.  Experiments by @joa-quim via Julia with modern mode has shown a few issues, which are also shared in GMT mex.  Here I list gmt mex commands:

1. `gmt ('begin') `does initiate a modern mode session, and API->GMT.current.settings.run_mode is set to **GMT_MODERN**. However, when the next module is called, here `gmt('plot', xy)` (xy is a MATLAB matrix),  and the persistent MATLAB API pointer is recycled, the _run_mode_ is back to **GMT_CLASSIC**.  So we don't know we are being called in modern mode; not sure where it gets reset yet.
2. The various checks in _GMT_Encode_Options_ are using classic module names, but we are running **plot**.  I have fixed this to always obtain the classic names, so that problem is solved.
3. Next is the issue of **KEYS**.  I notice many of the _PostScript_ modules have **KEYS** in the format _classic@modern_, and if modern mode we get the _modern_ keys after the @ character, else we get the _classic_ part.  This failed because the run_mode parameter was not set yet.  I can probably set that one one way or the other to extract the correct keys.  A simple ways is to compare if given and classic module name differs, but that will fail if someone uses the name psxy in modern mode (a bad idea anyway, but seems fragile).
4. Yet, many _PostScript_ producing modules do **not** have _classic@modern_ strings and I don't understand why. Some, like pscoast, can do many things and those cases are explicitly handled by tests in GMT_Encode_Options. But, for instance, psxy/plot is only a _PostScript_ producer so why does it not have the modern keys that specifically exclude the _PostScript_ key? As it is, _GMT_Encode_Options_ determines it needs an output filename and we end up with the command `gmt plot -<? ->?`, where the input ? will be linked to data from MATLAB but the output will go crazy since we don't write _PostScript_ to stdout in modern mode.  Again, if we know it is modern mode then all the PostScript keys should just be skipped.

I will try to determine we are modern mode and see if adding modern keys to psxy takes us a step further.

PaulWessel:
[...]

PR #2198, opened 2019-12-02
Fix API

seisman:
See #2196 for context.  This PR addresses those concerns in GMT; a smaller PR will also be submitted for gmtmex.


PaulWessel:
OK, this should take care of any case where an virtual input grid needs to be read twice: once to determine the region and then again as input to the original module.

PaulWessel:
I guess @joa-quim has to do the review on this.

Issue #2207, opened 2019-12-03
gmtinit_round_wesn() can be improved


joa-quim:
This Julia script illustrates the issues
```
gmtbegin("lixo1.png");
      plot([560 4480; 611 4531], lw=1, frame=(axes=:WSen, annot=:a, grid=2, title="(560 4480; 611 4531)"), R=:a);
gmtend("show")
```
![lixo1](https://user-images.githubusercontent.com/537321/70084912-21732d80-1607-11ea-9592-849f9015d805.png)

Other examples
![lixo2](https://user-images.githubusercontent.com/537321/70084986-4bc4eb00-1607-11ea-9987-672298210a2f.png)

![lixo3](https://user-images.githubusercontent.com/537321/70084998-51bacc00-1607-11ea-81ce-674c0828eeef.png)

PR #2208, opened 2019-12-03
Improve the BB rounding algorithm for the Cartesian case.

joa-quim:
This addresses #2207 The figs here are comparable one-to-one with those in that issue. The idea is to always round to the next fifth of a *decade*, where the *decade* is expanding increment calculated with log10.

lixo11

lixo22

lixo33


joa-quim:
Are we still on time to make this the default behavior (instead of ``-Rexact``)?

PaulWessel:
I think nobody will notice.  Sorry, this dropped off. Chat at AGU.

PR #2219, opened 2019-12-06
Add new share/tools script to get remote files

PaulWessel:
It is convenient for users to download all the cache and server DEMs prior to trips where internet connectivity may be poor or lacking.  gmt_getremote.sh downloads all the files not already on the system.

Still need a few documentation-related things and possibly Cmake somewhere?

### DIFF
--- a/share/tools/gmt_getremote.sh
+++ b/share/tools/gmt_getremote.sh
@@ -46,6 +46,6 @@ if [ ${cache} -eq 1 ]; then
 fi
 if [ ${data} -eq 1 ]; then
 	echo "Download all remote DEM grids from ${SERVER} not in ~/gmt/server (be very patient)" >&2
-	awk 'NF==3 && $1~/earth/ {print "@"$1}' /tmp/gmt_hash_server.txt | xargs gmt which -Gu
+	awk 'NF==3 && $1~/earth_relief/ {print "@"$1}' /tmp/gmt_hash_server.txt | xargs gmt which -Gu
 fi
 rm -f /tmp/gmt_hash_server.txt
--- a/share/tools/gmt_getremote.sh
+++ b/share/tools/gmt_getremote.sh
@@ -45,7 +45,7 @@ if [ ${cache} -eq 1 ]; then
 	awk 'NF==3 && $1!~/^earth_relief/ {print "@"$1}' /tmp/gmt_hash_server.txt | xargs gmt which -Gc
 fi
 if [ ${data} -eq 1 ]; then
-	echo "Download all remote DEM grids from ${SERVER} not in ~/gmt/server (be very patient)" >&2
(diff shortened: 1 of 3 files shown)


share/tools/gmt_getremote.sh
```diff
@@ -0,0 +1,51 @@
...
+SERVER=`gmt get GMT_DATA_SERVER`
+curl -sk ${SERVER}/gmt_hash_server.txt > /tmp/gmt_hash_server.txt
+if [ $? -ne 0 ]; then
+	echo "Error: curl failed with error $?" >&2
+	exit 1
+fi
+
+if [ ${cache} -eq 1 ]; then
+	echo "Download all cache files from ${SERVER} not in ~/.gmt/cache (be patient)" >&2
+	awk 'NF==3 && $1!~/earth/ {print "@"$1}' /tmp/gmt_hash_server.txt | xargs gmt which -Gc
```
seisman:
```suggestion
	awk 'NF==3 && $1!~/^earth_relief/ {print "@"$1}' /tmp/gmt_hash_server.txt | xargs gmt which -Gc
```
It's better to check if the first column starts with `earth_relief`.

share/tools/gmt_getremote.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+if [ ${data} -eq 1 ]; then
+	echo "Download all remote DEM grids from ${SERVER} not in ~/gmt/server (be very patient)" >&2
```
seisman:
```suggestion
	echo "Download all remote DEM grids from ${SERVER} not in ~/.gmt/server (be very patient)" >&2
```

share/tools/gmt_getremote.sh
```diff
@@ -0,0 +1,51 @@
...
+	awk 'NF==3 && $1~/earth/ {print "@"$1}' /tmp/gmt_hash_server.txt | xargs gmt which -Gu
```
seisman: I think this command also downloads the 03s and 01s DEMs. Is it intended?

PaulWessel: Hm, no.  Let's not include those - way too much.  Or add special srtm1s and srtm3s targets to the script.

PaulWessel: Wait, there is no earth_relief_01s.grd entry in the hash table.  Don't think we need to do anything about this.

seisman: Yes, you're right. Please also change this line to `$1~/^earth_relief/`

PaulWessel: OK, done.

share/tools/gmt_getremote.sh
seisman: Don't forget this one.

PR #2263, opened 2019-12-09
If transparency is <= 1, give warning about using percentage

PaulWessel:
If user gives **-t**0.3 it is likely they forgot to specify transparency in percentage (i.e., **-t**30) but we don't know that, so we just warn.


joa-quim:
Or we could accept transparency both in [0 1] or [0 100]. No one would really mean 0.3 of transparency in a 0 100 range.

PaulWessel:
As we discussed, if we incorrectly assume 0.9 is an error and should be -t90 then we prevent everybody from actually selecting 0.9% transparency.  So I think a warning is all we can do, which is still an improvement.

joa-quim:
... he could always do t0.009

PR #2286, opened 2019-12-16
Allow both -i and -o to specify an open-ended list to end of record

PaulWessel:
If a user wants to shuffle the order of cols 0 and 1 but otherwise read the entire record, they have to list everything that follows the **-i**1,0 list to ensure all numerical columns are added.  She also needs to add ,t to ensure the trailing text is added.  Likewise for output.  This is tedious and requires you to first determine how many columns are present.  This PR allows open-ended ranges to be given, such as

```
-i1,0,4:,t [read col 1, 0, then 4 to the end or external record, plus trailing text]
-o3:,t1  [Write cols 3 to end of internal record, plus 2nd word of trailing text]
```

This adds new capability to GMT.  Also while not pretty (yet), this means **-:** i can be specified as **-i**1,0,2:

### DIFF
--- a/doc/rst/source/explain_-icols_full.rst_
+++ b/doc/rst/source/explain_-icols_full.rst_
@@ -9,6 +9,7 @@
     column, optionally add any of  the following: **+l** takes **log10** of
     the input values first; **+s**\ *scale*, subsequently multiplies by a
     given  scale factor [1]; **+o**\ *offset*, finally adds a given offset [0].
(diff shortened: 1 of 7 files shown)


joa-quim:
Should be committed on master, not 6.0.
In fact, everything should go to master now.

PaulWessel:
OK, but it goes to master a minute after merged to 6.

Issue #2307, opened 2019-12-20
Annotations for automatically generated CPT


seisman:
See the script and output below:
```
gmt begin map png
gmt grdimage @earth_relief_20m -JH15c -Rg -Baf -BWSen
gmt colorbar
gmt end show
```

![image](https://user-images.githubusercontent.com/3974108/71283463-4f0de580-232e-11ea-952a-13d0073c86d6.png)

PaulWessel:
So the bar is scaled to fit the data range, and when that happens we normally want to use -B to get sane intervals.  On the other hand, the default is to annotate the boundaries as given in the CPT.  I guess we need to know if the cpt was stretched and hence we should instead default to -B.

PaulWessel:
In terms of detecting that we have this case, perhaps we can do so by examining the z-levels? I.e., if they are not equidistant and not integers? but may not be what users want.  Seems like we need to remember that we automatically stretched the CPT to fit a data range, not a user-specified range? If someone did `gmt makecpt -Cgeo -T-9000/9000` then that is different from just using a grids min/max. The autostretch happens in grd2kml, grdimage, grdvector and grdview.  These have a **-C**...[**+i**dz] that can be used to round the CPT rather than use the raw min/max. Perhaps if _dz_ == 0 (not given) we will want to pass that information to the CPT so when it is used (by colorbar) we know we should do -B if no -B was given?
Alternatively, we could just decide that in modern mode all colorbar calls without **-B** gets **-Baf**.

seisman:
> Alternatively, we could just decide that in modern mode all colorbar calls without **-B** gets **-Baf**.

No, users may write their own CPT and want to annotate it in the same as what CPT does.

> The autostretch happens in grd2kml, grdimage, grdvector and grdview. These have a **-C**...[**+i**dz] that can be used to round the CPT rather than use the raw min/max. Perhaps if _dz_ == 0 (not given) we will want to pass that information to the CPT so when it is used (by colorbar) we know we should do -B if no -B was given?

I think this is a better solution.

PR #2311, opened 2019-12-21
Supply -Baf when colorbar plots a stretched CPT

PaulWessel:
Tools like _grdimage_ may produce a stretched CPT that fits the z-range of a grid exactly.  In those cases we cannot let psscale _use_ the CPT levels as annotation and tick intervals but must override via **-Baf**.  This PR does so by adding a comment keyword **STRETCHED** to such CPTs as they are written and then we watch for that word upon reading.  Closes issue #2307.

### DIFF
--- a/doc/rst/source/colorbar_common.rst_
+++ b/doc/rst/source/colorbar_common.rst_
@@ -33,7 +33,7 @@ Optional Arguments
     provided (classic mode only), the default is to annotate every color level based on the
     numerical entries in the CPT (which may be overridden by ULB
     flags in the CPT). The exception to this rule is for CPT files that were scaled to fit the range
-    of a grid exactly and thus have arbitrary color levels; these will trigger an automatic -Baf setting.
+    of a grid exactly and thus have arbitrary color levels; these will trigger an automatic **-Baf** setting.
     To specify custom text annotations for
     intervals, you must append ;\ *annotation* to each z-slice in the CPT.
     For standard **-B** operations, see below.
--- a/doc/rst/source/colorbar_common.rst_
+++ b/doc/rst/source/colorbar_common.rst_
@@ -30,9 +30,11 @@ Optional Arguments
     the right of a vertical bar), except when using the **+m** modifier of the **-D** option. As an
     option, use the y-axis label to plot the data unit to the right of a
     horizontal bar (and above a vertical bar). If **-B** is omitted, or no annotation intervals are
-    provided, the default is to annotate every color level based on the
+    provided (classic mode only), the default is to annotate every color level based on the
     numerical entries in the CPT (which may be overridden by ULB
-    flags in the CPT). To specify custom text annotations for
+    flags in the CPT). The exception to this rule is for CPT files that were scaled to fit the range
+    of a grid exactly and thus have arbitrary color levels; these will trigger an automatic -Baf setting.
+    To specify custom text annotations for
     intervals, you must append ;\ *annotation* to each z-slice in the CPT.
     For standard **-B** operations, see below.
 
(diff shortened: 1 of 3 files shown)


doc/rst/source/colorbar_common.rst_
```diff
@@ -30,9 +30,11 @@ Optional Arguments
     the right of a vertical bar), except when using the **+m** modifier of the **-D** option. As an
     option, use the y-axis label to plot the data unit to the right of a
     horizontal bar (and above a vertical bar). If **-B** is omitted, or no annotation intervals are
-    provided, the default is to annotate every color level based on the
+    provided (classic mode only), the default is to annotate every color level based on the
     numerical entries in the CPT (which may be overridden by ULB
-    flags in the CPT). To specify custom text annotations for
+    flags in the CPT). The exception to this rule is for CPT files that were scaled to fit the range
+    of a grid exactly and thus have arbitrary color levels; these will trigger an automatic -Baf setting.
```
seisman:
```suggestion
    of a grid exactly and thus have arbitrary color levels; these will trigger an automatic **-Baf** setting.
```

src/gmt_support.c
```diff
@@ -7787,9 +7790,11 @@ struct GMT_PALETTE *gmt_get_palette (struct GMT_CTRL *GMT, char *file, enum GMT_
 		/* Stretch to fit the data range */
 		/* Prevent slight round-off from causing the min/max float data values to fall outside the cpt range */
 		if (gmt_M_is_zero (dz)) {
+			struct GMT_PALETTE_HIDDEN *PH = gmt_get_C_hidden (P);
 			GMT_Report (GMT->parent, GMT_MSG_LONG_VERBOSE, "Auto-stretching CPT file %s to fit data range %g to %g\n", master, zmin, zmax);
 			noise = (zmax - zmin) * GMT_CONV8_LIMIT;
 			zmin -= noise;	zmax += noise;
+			PH->auto_scale = 1;	/* Flag for colorbar to supply -B af if not given */
```
seisman:
```suggestion
			PH->auto_scale = 1;	/* Flag for colorbar to supply -Baf if not given */
```

seisman:
It works well, but **STRETCHED** doens't sound a good keyword. Most CPTs are streched version of master CPTs, right?

src/gmt_support.c
```diff
@@ -7292,6 +7293,8 @@ struct GMT_PALETTE * gmtlib_read_cpt (struct GMT_CTRL *GMT, void *source, unsign
 		}
 		else if ((h = strstr (line, "CYCLIC")))	/* CPT should wrap around */
 			X->is_wrapping = 1;
+		else if ((h = strstr (line, "STRETCHED")))	/* CPT was stretnch to exact min/max with no dz rounding */
```
seisman:
```suggestion
		else if ((h = strstr (line, "STRETCHED")))	/* CPT was stretched to exact min/max with no dz rounding */
```

PaulWessel:
I agree STRETCHED is dumb, will change to something better. But in the end, it is just for internal consumption.

seisman:
The annotations are still messy when **-C+i** is used. Try the script below. 
```
gmt begin map png
gmt grdimage @earth_relief_20m -JH15c -Rg -Baf -BWSen -C+i1000
gmt colorbar
gmt end show
```
Perhaps we should add **-Baf** for all automatically generated CPTs.

PaulWessel:
Hm, I will debug this later -I must now have understood my implementation of the +idx.  If so then yes we would have to do it for all stretched CPTS.

PaulWessel:
Fixed, will commit after tests complete.

PR #2325, opened 2019-12-22
Fix formatting plus missing -m option

PaulWessel:
The x2sys_init module has a **-m** option that was missing from the documentation.


seisman:
I thought GMT can automatically detect multi-segment tables since GMT5. Why does x2sys_init still need **-m**?

PaulWessel:
Yes, had to read x2sys.c to learn that the -m flag is skipped anyway.  So now I have removed its mention and use.

Issue #2327, opened 2019-12-22
makecpt in externals produce wrong CPTs when master has hinge


joa-quim:
```
julia> cpt = makecpt(" -T-10000/2000/100 -Cgeo");
julia> psscale(cpt, B=:af, show=1, D="x8c/1c+w12c/0.5c+jTC+h", fmt=:png)
```
![GMTjl_tmp](https://user-images.githubusercontent.com/537321/71324905-cf724900-24dc-11ea-91b5-1e341e2f14b2.png)

Checked and Matlab displays the same behavior, which doesn't rule out the possibility that error in gmtmex and the Julia wrapper mimicked it.

PaulWessel:
This section in gmtmex_parser.c looks suspicious:

```
		if (mxIsNaN (hinge[0])) {
			P->has_hinge = 1;
			P->mode &= GMT_CPT_HINGED;
		}
```

Should be` !mxIsNaN (hinge[0])`, no? If NaN then no hinge, else there is a hinge at that value.
And should not P->hinge be set to hinge[0] if true? (default to zero so probably not the issue in this example, but the if-test is.

seisman:
GMT command line gives the same result.
```
gmt begin map png
gmt makecpt -T-10000/2000/100 -Cgeo
gmt colorbar -Baf -Dx1c/1c+w10c+h
gmt end show
```
![image](https://user-images.githubusercontent.com/3974108/71326752-8b2b7c80-24cd-11ea-99d8-2defd23bc8c4.png)

PaulWessel:
Problem is makecpt which is not doing the same thing as the auto-stretching (which does honor the hinge).  Still think the mex code is wrong but that is not the problem here though.  Will have to see where to fix it.

PaulWessel:
Fixing the lack of hinge-implementation in sample_cpt.  Getting a few failures:
	125 - GMT_tut_17.sh (Failed)
	407 - grdmath/ops.sh (Failed)
	458 - makecpt/logcpt.sh (Failed)

Looking at tut 17 we have the case that we now actually are honoring the hinge.  Top is the new plot, bottom is the old:

![GMT_tut_17-new](https://user-images.githubusercontent.com/26473567/71329676-eceae700-24cb-11ea-8c9c-38d4e6d96fd4.png)

![GMT_tut_17](https://user-images.githubusercontent.com/26473567/71329677-f3795e80-24cb-11ea-8a2b-abef6008c214.png)

[...]

PR #2334, opened 2019-12-23
Let makecpt honor the hinge during resampling

PaulWessel:
See #2327 for context.  Unlike the auto-stretching that occurs when a master CPT is given to grdimage (stretched to fit the data range, considering any hinge), makecpt did not have any implementation that dealt with the hinge.  This PR implements the algorithm that treats the two sides of the CPT differently.  I will update the three PS failures due to the change before committing the PR after approval.


PaulWessel:
**We have an aliasing issue when we do makecpt and request a discrete output table.**

Consider master CPT earth which has 160 z-slices.  Appendix M 1a asks for only 8 discrete colors.  There is no averaging here, so we simply sample the values of the z-slice closest to each of the 8 points.  We normally call this _aliasing_.  In fixing the hinge problem I noticed we pick the half-way point in a slice to get the color.  I.e., if the input table is continuous then we pick the z-value at the middle of the output slice and determine the nearest input slice to "sample" the colors from the slice.  The alternative would be to just pick the start output z-value and find its nearest input slice.  Below you seen the difference.  The top is the new output by picking the lower (this accentuates the hinge color change) while the lower is the old output that picks the mid-point of the z-slice (this picks a color far from the hinge color change in most cases, especially when n_in >> n_out).  You can see the new scheme picks the strong green just above sea level while the old would pick a color much higher up in the input table:

NEW SCHEME:
new

OLD SCHEME:
old

If we retain the old scheme then we have 3 failures, while the new scheme adds several of these M appendix plots plus a few others.  Either scheme introduces some sort of aliasing, but perhaps a hinge should yield a clear difference across it for the discrete CPT as well, so I think the new scheme is better.  Comments, @joa-quim and @seisman ?

I also think we should be more clear in saying that selecting discrete CPTs from continuous tables is a **bad idea**.  Some of this comes from old GMT4 usage where we would run makecpt with _zmin/zmax/zinc_, but since GMT 5 we avoid the _zinc_ unless you want a discrete table.

joa-quim:
The fact that the new scheme respects the hinge color transition makes a strong point in its favor.

PaulWessel:
What say you, @seisman?  Maybe like @joa-quim says if we have to alias we might as well pick an alias that also accentuates the hinge?

seisman:
Yes, I also favor the new scheme.

PaulWessel:
OK, I have implemented new aliasing scheme and update the PS files that changed.  Test pass again now.  Please review.

PR #2343, opened 2019-12-25
movie_frame.XX creates the image directly in parent dir.

joa-quim:
This PR changes the movie_frame.xxx to create the image directly in the parent directory instead of creating it in the subdir and move it down. On Windows this avoids the annoying message ``1 file(s) moved.`` that was printed for each frame.

On the other hand **-V** was doing nothing. Now it prints the processed frame number but without changing line. 

In Julia it now nicely reports

```
julia> movie("main_sc.jl", pre="pre_sc.jl", C="7.2ix4.8ix100", N=:anim04, T="flight_path.txt", H=2, L="f+o0.1i", F=:mp4, A="+l+s10", V=true, Vd=1)
        movie main_script.bat -V -C7.2ix4.8ix100 -Nanim04 -Tflight_path.txt -A+l+s10 -Fmp4 -H2 -Lf+o0.1i -Sbpre_script.bat
movie [WARNING]: Frame 59 of 60 completed [100.0 %]
```

### DIFF
--- a/src/movie.c
+++ b/src/movie.c
@@ -1707,13 +1707,12 @@ int GMT_movie (void *V_API, int mode, void *args) {
 			status[k].completed = true;	/* Flag this frame as completed */
 			n_cores_unused++;		/* Free up the core */
 			percent = 100.0 * n_frames_completed / n_frames;
-			GMT_Report (API, GMT_MSG_VERBOSE, "Frame %*.*d of %d completed [%5.1f %%]\r", precision, precision, k, n_frames, percent);
+			GMT_Report (API, GMT_MSG_LONG_VERBOSE, "Frame %*.*d of %d completed [%5.1f %%]\n", precision, precision, k, n_frames, percent);
 		}
 		/* Adjust first_frame, if needed */
 		while (first_frame < n_frames && status[first_frame].completed) first_frame++;
 		if (n_frames_completed == n_frames) done = true;	/* All frames completed! */
 	}
-	GMT_Report (API, GMT_MSG_VERBOSE, "\n");		/* Last was a '\r' */
 	/* END PARALLEL EXECUTION OF FRAME SCRIPTS */
 
 	gmt_M_free (GMT, status);	/* Done with this structure array */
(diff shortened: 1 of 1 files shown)


joa-quim:
Yes, seeing ``movie [WARNING]:`` is not ideal but I don't want to use **-Vl** that prints a lot more info than we want when only wishing to see the percentage of the job done so far. We want a kind of progress bar and **V** by itself prints noting.

PaulWessel:
We can rethink our whole verbosity situation.  I think most users expect to see WARNINGS if they have selected some option combination that raises a flag of some sort.  Users do not expect to have to set -V to learn that their options are bad.  To that effect, we could simply raise the default verbosity level to GMT_MSG_VERBOSE so that anyone wanting to see nothing must use -Vq.  We have

> 	     q - Quiet, not even fatal error messages.
> 	     n - Normal verbosity: only error messages.
> 	     c - Also produce compatibility warnings [Default when no -V is used].
> 	     v - Verbose progress messages [Default when -V is used].
> 	     l - Long verbose progress messages.
> 	     d - Debugging messages.

Currently -Vn is the default.  If -Vv was the default then warnings would show by default. Then, using -Vl to see long verbose would be fine for movie, and perhaps we can push some useless information messages into the debug realm instead.

joa-quim:
Yes, ``-Vn`` should be the default. I mean, warnings should always be printed unless ``-Vq``. But deprecation warnings ``-Vc`` are more a punishment than a real warning and those would pop up is ``-Vv`` is to become the default.

However, I still think we need a progress type info under ``-Vv`` like the case of my change here or, from memory, the convergence in ``surface``. For all modules that take some time ``-V`` should print a time indicator, and preferably nothing more.

PaulWessel:
You mean **-Vv** should be the default.  It is currently **-Vn** (errors messages).  We could move **-Vc** to come _after_ **-Vl** but before **-Vc** so that you could choose to find out if your scripts use deprecated syntax.  Since we are still supporting it we could remove the punishment and only keep the opportunity to learn via **-Vc**.

joa-quim:
>  We could move -Vc to come after -Vl but before -Vc 

you mean after **-Vv** but before **-Vl**
Yes, that sounds good. Have to go now for a while.

PR #2377, opened 2020-01-02
The -o docs incorrectly talked about scaling

PaulWessel:
Only **-i** has the scaling modifiers.  However, parsing of **-o** silently ignored any modifiers.  Now we give an error.  Also, improved the documentation for **-i** and **-o**.


PaulWessel:
I was wondering that myself yesterday when I thought I needed it, but it has to do with complications arising during the various output schemes.  I convinced myself it is not worth trying to implement (at this point) so better not have the wrong documentation.

seisman:
OK to merge.

Issue #2382, opened 2020-01-02
Documentation of grdedit -h -i


seisman:
It's not clear whether grdedit supports **-h** and **-i**.

The synopsis says yes, but both `THIS_MODULE_OPTIONS` and the long usage don't include the **-i** option, and the manpage also doesn't list them.

PR #2384, opened 2020-01-02
Clarify that grdedit can use -h -i

PaulWessel:
Code did not enable **-i** and the docs where shaky.  Closes issue #2382.


joa-quim:
You keep committing to 6.0
The movie KEYs were already fixed in master.

PaulWessel:
It automatically rebuilt the module.[c]h file when I dealt with grdedit.  The move bug was in 6 though, not master.

joa-quim:
But since there will be no release cut off from the 6.0 branch, we should stop committing to it, which will later go to master via merge. Much better to only commit directly to master

PR #2391, opened 2020-01-03
Add instructions for building with macports

### LINKED ISSUE #2388: Add instructions for building with macports

Since some of the developers and others are still using macports for building from source, we should have an entry in the BUILDING.md file.  One reason is that installing things like gdal is nontrivial on macports since you need to specify a list of variants.  By default, the gdal installation does not support netCDF, for instance, and I need to run

` sudo port install gdal +curl +geos +hdf5 +netcdf +openjpeg
`

to regain basic GMT capabilities.  The common man or woman won't know that without documentation.

seisman:
Fixes #2388.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -14,7 +14,8 @@ For general users:
   * [Fedora](#fedora)
   * [Archlinux](#archlinux)
   * [FreeBSD](#freebsd)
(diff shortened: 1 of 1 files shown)


seisman:
I don't test it, but it should work.

Issue #2409, opened 2020-01-05
-S under Azimuthal Equidistant projection


yangleir:
When add `-Scolor` for pscoast under the Azimuthal Equidistant projection, the map shows a bug that some land is filled with the color set by -S.
If we add a optionally horizon parameter like  `-JE180/50/170/4.5i`, the map is changed to normal. But if the horizon is set to 180 (default), the map is broken again. So I guess it is the default horizon affected the -S.
The value of the longitude in `-JE` also related to the bug. If it was set to a 10 like `-JE10/50/170/4.5i` (with -S), the map shows bug, too. So it is a bit confused.

**Description of the problem**

**Full script that generated the error**

```
gmt pscoast -Rg -JE180/50/4.5i -Bg -Dc -A1000 -Glightgreen -Sgray > GMT_az_equidistant.ps
gmt pscoast -Rg -JE180/50/170/4.5i -Bg -Dc -A1000 -Glightgreen -Sgray > GMT_az_equidistant.ps
```

**Actual outcome**

![GMT_az_equidistant](https://user-images.githubusercontent.com/9152905/71777395-4bd75280-2fda-11ea-9922-f2503b94d0e4.png)

**System information**

* Operating system:MINGW64_NT-10.0-17763
* GMT version (`gmt --version`):6.0.0

PaulWessel:
Hi @yangleir, I have addressed this particular problem and it has been merged into the repo.  Thank you for your report.

yangleir:
Hi @PaulWessel ,Thanks l so much. You and your team always fix bugs so fast.

Issue #144, opened 2018-10-24
Bug in Azimuthal Equidistant projection (-JE) 


longgangfan:
**Description of the problem**
When  I run the code below:
(1)`pscoast -JE-120/30/4.5i -Rg -Glightgray -B30/30 -Dc -A10000 -Wthinnest -P >test.ps`
![test](https://user-images.githubusercontent.com/11954551/47424333-05714b80-d7ba-11e8-8464-18adf9c15400.jpg)
It seems all right, but if I change the Longitude and latitude of the projection center or use "-S", it will output a image with some area overlapped like below.
(2)`pscoast -JE120/30/4.5i -Rg -Glightgray -B30/30 -Dc -A10000 -Wthinnest -P >test.ps`
![test](https://user-images.githubusercontent.com/11954551/47424607-9a744480-d7ba-11e8-97d5-f38ea3721a3a.jpg)
(3)`pscoast -JE-120/30/4.5i -Rg -Glightgray -Slightblue -B30/30 -Dc -A10000 -Wthinnest -P >test.ps`
![test](https://user-images.githubusercontent.com/11954551/47424860-3f8f1d00-d7bb-11e8-8f99-5d0590465bd8.jpg)

I test this bug on macos 10.14 (homebrew version gmt 5.4.4) and ubuntu 16.04 (github master branch), it can be reproduced on both of them.
If I set the horizon to 179.9999, the output seems to be right.
(4)`pscoast -JE-120/30/179.9999/4.5i -Rg -Glightgray -Slightblue -B30/30 -Dc -A10000 -Wthinnest -P >test.ps`
![test](https://user-images.githubusercontent.com/11954551/47425398-ca244c00-d7bc-11e8-89c0-7977a512977d.jpg)

**System information**

* Operating system: ubuntu 16.04/macos mojave
* Version of GMT: 6.0.0_93597c0/5.4.4

joa-quim:
This is a duplication of the SVN https://gmt.soest.hawaii.edu/issues/872 issue.

longgangfan:
> This is a duplication of the SVN https://gmt.soest.hawaii.edu/issues/872 issue.

Sorry for my duplication.

joa-quim:
It's OK. We have to port the remaining bugs in the Track system to here anyway, so we may well start with this one.

PaulWessel:
[...]

PR #2411, opened 2020-01-06
When clobbering the background we only want to plot land on top

PaulWessel:
See issue #2409 for background.  The **pscoast** code would set _clobber_background_ to true (since both land and ocean fill are selected) but then would initialize _start_direction_ and _stop_direction_ to -1 and +1 and select _level_to_be_painted_ as 0 for both directions.  Not sure how this worked before but clearly it fails catastrophically from time to time, like in #2409.
However, after attempting a fix I get several failures related to **pscoast**:

```
	 77 - GMT_hammer.sh (Failed)
	150 - anim08/anim08.sh (Failed)
	151 - ex01/ex01.sh (Failed)
	169 - ex19/ex19.sh (Failed)
	170 - ex20/ex20.sh (Failed)
	551 - pscoast/pscoast_JA.sh (Failed)
	552 - pscoast/pscoast_JE.sh (Failed)
	554 - pscoast/pscoast_JE3.sh (Failed)
	557 - pscoast/pscoast_JS.sh (Failed)
	629 - pssolar/shading.sh (Failed)
```

However, I cannot actually see any real failures. It seems the new PS are the correct one and the old ones had (due to the above bug) drawn some features twice and there are slightly different polygon renders.  So @seisman and @joa-quim, please run tests and see what you think of the "failures" and if you agree that these are not failures but should be PS updates due to the code fix.


seisman:
I have one more failure (GMT_TM.sh), but all the new PS seem correct to me.

seisman:
BTW, #144 also reported two -JE failings. The second one is also fixed by this PR.

PaulWessel:
Sorry, I left off GMT_TM.sh - I have that one as well.  Since we flagged failing tests we never see them - I think we may have some with -JE that perhaps now work?

seisman:
Yes, that's possible.

PaulWessel:
pscoast/pseudo.sh is the only coastline test flagged with GMT_KNOWN_FAILURE. Would you know how I might run those tests without editing all the tests and (temporarily) removing that tag?  I will do it manually for pseudo for now.

seisman:
I think you can edit the file `test/gmtest.in`, and always set `known2fail` to 0.

PaulWessel:
No, pseudo.ps has other issues not related to pscoast but to **-p** and positioning.
As for this PR perhaps we will wait for @joa-quim to run tests as well and see if he agrees with us that the new PS is the correct one - I can them add those to the PR.

joa-quim:
Curious. Common failing tests I have only

```
        150 - anim08/anim08.sh (Failed)
        551 - pscoast/pscoast_JA.sh (Failed)
        552 - pscoast/pscoast_JE.sh (Failed)
```
I still have several failures like. Don't get it why the same does not happen on CI
```
psxy [ERROR]: Could not find custom symbol C:C:/programs/WinGit/progs_cygw/GMTdev/gmt5/master/test/psxy/comet
```

Regarding the failures, 551 & 552 differ only on the rim (a very thin purple line), so the new PS as as good as the previous. 150 differs in the rim and in some islands. Strange but again, as good as the old one.

PaulWessel:
OK, given that this is clearly a bug in pscoast I think it is reasonable to update the PS following that change.  I will compute the new PS and add to this PR and then ask for your reviews.

PaulWessel:
OK, PS files updated, please approve this PR, @joa-quim or @seisman.

Issue #2412, opened 2020-01-06
CPT issues related to hinges


seisman:
Script below gives different results after #2327 was merged:
```
gmt makecpt -Cpolar -T100/300/20 > mycpt1.cpt
gmt makecpt -Cpolar -T100/300/20 -Z > mycpt2.cpt
gmt psscale -Cmycpt1.cpt -Baf -Dx0/0+w10c+h -P -K > map.ps
gmt psscale -Cmycpt2.cpt -Baf -Dx0/0+w10c+h -Y2c -P -O >> map.ps
gmt psconvert -A -P -Tg map.ps
```

**Before:**

![image](https://user-images.githubusercontent.com/3974108/71831770-9529af80-3077-11ea-8e29-7a629859c08a.png)

**After:**

![image](https://user-images.githubusercontent.com/3974108/71832051-1f721380-3078-11ea-98c2-0cdc42ba81cb.png)

Issues:

1. There is no way to generate a "polar" CPT like before.
2. The new continuous CPT looks weird.

For the first issue, possible solutions are:

1. Add an option to allow ignoring hinge
2. For master CPTs like *polar*, *red2green*, is the hinge really useful? Should we remove hinges from these CPTs?

PaulWessel:
I think those CPTs do have a hinge, but I agree that it being hardwired to 0 is not useful in all cases.  The before/after effect is of course the fix to actually do what we said we were doing regarding resampling.  Possible options:

1. Reset the hinge if selected range does not contain the actual hinge.  It is unlikely people want the all-red CPTs above, they probably want 200 to be the new hinge.
2. Offer a modifier **+h**[_z_] to the cptfile that resets the hinge to z (if given) or removes the hinge if not given.

I guess option 2 is more flexible and also can handle option 1 without making that the default action.

seisman:
If I understand it correctly, the old behavior is to ignore the hinge, and we declared it as a bug and fix it in #2334. However, it means the new behavior will break many old scripts.

PaulWessel:
Yes, but clearly the intent of the hinge was to honor it, so there was a but.  However, we are free to define what should happen when the requested range does not include the hinge, and we could decide that the only sensible thing to do is to ignore the hinge in those cases.  Unlike CPTs for bathymetry (where hinge = 0 is a good thing), we could also decide that hinge for polar etc does not make sense since 0 is not a universal value for hinges.

joa-quim:
[...]

PR #2416, opened 2020-01-07
Update how CPT hinges are handled

PaulWessel:
Separate between soft and hard hinges.  Hard hinges are data driven like those for bathymetry or magnetics, while soft hinges are ignored unless **+h**[_hinge_] is given.  Explain this better in the CPT documentation.

- [x] Fix the etopo1 CPT to have a hinge at 0, not some tiny nonzero number
- [x] Fix the srtm CPT to also have hinge at 0 but stay blue for a short while afterwards
- [x] Add optional modifier **+h**[_hinge_] to turn soft hinges into hard hinges
- [x] Remove the value for the HINGE given in the CPTs since they should (now) all be zero in the normalized -1/1 range
- [x] Remove change HINGE to HARD_HINGE for the bathy.topo and mag CPTs
- [x] Change remaining HINGE to SOFT_HINGE
- [x] Add triangles on GMT_App_M_1?.ps plots to indicate which CPTs have hard or soft hinges


PaulWessel:
Notes to self: the normalized hinge value is always 0.  When we internally normalize a CPT for 0-1 in gmt_sample_cpt the normalized hinge will be at 0.5.  The user-supplied hinge will be in user coordinates [0].

PaulWessel:
I believe this PR is ready for review.  I have improved the documentation pertaining to CPTs by explaining the new soft and hard hinges, linking to the actual CPT plots in the appendix, adding triangles to indicate which ones have soft or hard hinges, and introduce two include rst_files that programs that produce (create_cpt.rst_) or use with grids (use_cpt_grd.rst_) includes to spell out what **-C** does.  I hope to later do one for the remaining tools that read CPTs (plot, etc) but these are still too specific to the module.  In the process I had to update a few PS since we basically reverted the hard hinges from many CPTS such as polar, split, etc.  For instance, this demonstrates that polar no longer has a hard hinge:

```
gmt makecpt -Cpolar -T-1/3
-1	blue	1	white
1	white	3	red
```

but it can be activated by forcing a hinge:

```
gmt makecpt -Cpolar+h -T-1/3
-1	blue	0	white
0	white	3	red
```

and instead putting the hinge at 2:

```
gmt makecpt -Cpolar+h2 -T-1/3
-1	blue	2	white
0	white	1	red
```

remkos:
There are a few things I do not understand and may need to be clarified:
* Is "normalised units" the same as "units in the CPT file"?
* Is <value> in +h<value> in normalised units, in the units of the CPT file, or scaled units? This is not at all indicated.
* Why are CPTs with a (soft) hinge said to be normalised to -1/1 whereas others (if they still exist) normalised to 0/1? Is this information relevant to the user? Isn't that an internal issue?
* It is not evident that in case you do not make discrete intervals out of a CPT that is continuous, the hole matter about hinges is totally irrelevant. In that case every value in the CPT file is a hinge.
* What are +U and +u supposed to do? It is said to be a conversion from meters to "another unit". To Watts maybe? How do I specify that unit? And the CPTs are dimensionless; they are not in meters. I really do not get the point of this.

Personally, I think this is making things way too complicated with those hinges. I think users should *not be allowed* to set the intervals on output CPT files, i.e. changing from continuous to discrete or vice versa; then the whole issue becomes moot. If a CPT has, e.g., 16 discrete colours, like `panoply` then it should stay like that. It's how the colour map was designed, with every value in the CPT file being a "hinge". Yes, the user can alter the interval to their wishes, but not change whether it is discrete or continuous.

Conversely, a colour map like `polar` is meant to be continuous, with a hinge in the middle, and that should stay like that. If there is a real need for the polar-like colour map with say 16 discrete intervals, it should be made. Yes, that would extend the number of colour maps provided, e.g., adding polar16, but that be it.

True, the current implementation provides a lot of possibilities, but at the detriment of clarity, and opening the door to a lot of user-misuse.

PaulWessel:
Thanks for the feedback.  Here are some background before I address your specific concerns: In GMT 6 we added HINGE to several CPTs and in the process introduced backwards incompatibilities.  The discussion in #2412 captures those problems.  We realized there is a difference between what I now call a _hard_ hinge (e.g., sea-level, specified by a feature of the intended data) versus a _soft_ hinge (i.e., the white point in polar, which is not tied to a data feature _per se_).  The 6.0 broke the ability to make a polar-like CPT that goes from 0 to 100 with a white point at 50, for instance.  The solution was to restore backwards compatibility but recognize this second group of hinge-worthy CPTs by flagging them as having soft hinges that the user can activate if they so desire via **+h**[_hinge_] (e.g., when not specifying a symmetric range about zero or not selecting zero as their data hinge).  OK, some answers:

> Is "normalised units" the same as "units in the CPT file"?

Yes, and your feedback makes me realize that there is no point exposing the user to the issue of normalized _z_-values and the -1/+1 versus 0/1.  It is irrelevant to them.

> Is in +h in normalised units, in the units of the CPT file, or scaled units? This is not at all indicated.

The Cookbook makes this distinction I think by contrasting the normalized _z'_ values with the users _data_ values.  So the hinge is in user data values.  I will make this more clear in the documentation for **-C**.

> Why are CPTs with a (soft) hinge said to be normalised to -1/1 whereas others (if they still exist) normalised to 0/1? Is this information relevant to the user? Isn't that an internal issue?

Yes, internal issue and I will remove that.  It may be placed as comments for gurus who need to make new master CPTs in the future.

> It is not evident that in case you do not make discrete intervals out of a CPT that is continuous, the hole matter about hinges is totally irrelevant. In that case every value in the CPT file is a hinge.

Yes, discrete CPTs have hinges everywhere; this distinction will be clarified.  Nothing new here though.  What we should add is that selecting discrete nodes may introduce aliasing of the original color table; I think we already warn when users create some of these.

> What are +U and +u supposed to do? It is said to be a conversion from meters to "another unit". To Watts maybe? How do I specify that unit? And the CPTs are dimensionless; they are not in meters. I really do not get the point of this.

These modifiers goes back to at least 5.2 or before. Their documentation has been sketchy and only present in the cookbook.  The initial reason for these modifiers was to have the ability to use a CPT scaled to meters (e.g., for a topography grid) but be able to use it with **psscale** to make a km CPT without having to make a second CPT just for that purpose.  Furthermore, the same modifiers have also been available to grids from that time as well, converting the _x_ and _y_ values in Cartesian grids between different _distance_ units (e|k|m|M|n|u|s).  So I need to emphasize _distance_ unit more clearly.  Thus, **+U**|**u** is a shared and long-lived mechanism for both CPTs and grids.

I will implement the changes I mentioned next.

PaulWessel:
I have update the documentation discussion of CPTs in the cookbook and the two new include rst_ files regarding hinges.

seisman:
See the script below. The output seems incorrect.
```
gmt begin map png
gmt makecpt -Cpolar -T100/300/20
gmt colorbar -Baf -Dx0/0+w10c+h -Y2c

gmt makecpt -Cpolar -T100/300/20 -Z
gmt colorbar -Baf -Dx0/0+w10c+h -Y2c

gmt makecpt -Cpolar+h -T100/300/20
gmt colorbar -Baf -Dx0/0+w10c+h -Y2c

gmt makecpt -Cpolar+h -T100/300/20 -Z
gmt colorbar -Baf -Dx0/0+w10c+h -Y2c

gmt makecpt -Cpolar+h200 -T100/300/20
gmt colorbar -Baf -Dx0/0+w10c+h -Y2c

gmt makecpt -Cpolar+h200 -T100/300/20 -Z
gmt colorbar -Baf -Dx0/0+w10c+h -Y2c

gmt makecpt -Cpolar+h150 -T100/300/20
gmt colorbar -Baf -Dx0/0+w10c+h -Y2c

gmt makecpt -Cpolar+h150 -T100/300/20 -Z
gmt colorbar -Baf -Dx0/0+w10c+h -Y2c

gmt end show
```

image

PaulWessel:
Yes, I see the discrete output tables are wrong - it looks fine with most of the continuous. I have reverted this to WIP and will work on this a bit later.

PaulWessel:
Unrelated to the script above but: Should we impose that all the GMT master CPTs be continuous?  It is really treacherous to have users create discrete tables by resampling discrete CPTs, since the number of slices in the master vs the ones requested cause really problems.

joa-quim:
At lost not the ``mag.cpt``. It's meant to be discrete.

PaulWessel:
But would you want Lester to run
gmt makecpt -Cmag -T300/700/100
on it?

joa-quim:
If it's going to be used in a FAA map maybe that's fine. What I want is to continue to use ``gmt grdimge mag.grd -Cmag`` with no need to re-create a discrete CPT everytime.

PaulWessel:
I guess what I think we would like to avoid is bad user decisions in making discrete tables that badly alias the original tables, especially if they are discrete to start with.

joa-quim:
You can bug them with a message saying "Resampling a discrete CPT is really a BAD idea".

remkos:
> I guess what I think we would like to avoid is bad user decisions in making discrete tables that badly alias the original tables, especially if they are discrete to start with.

That comes back to what I said before. Maybe we should not even *allow* to change the intervals from discrete CPTs. Just like mag.cpt, they are designed a certain way, and users should not be allowed to meddle with it.

I'm also not in favour of making discrete CPTs out of continuous ones, but that is way less dangerous.

remkos:
> Thanks for the feedback. Here are some background before I address your specific concerns:

Thanks. That explanation and the changes to the rst files explained a lot.

PaulWessel:
I agree with that, @remkos.  The stretching of discrete CPTs to some new range and discrete interval is full of pitfalls and can only serve to make GMT results look ugly.  As you say, the making of discrete from continuous is at least clearer where the user basically samples a continuum at discrete points.  Some applications do need discrete CPTs so I think we should allow the latter, but I would be happy to put my foot down on allowing any discrete CPT to be used as the other dynamically scalable CPTs.

remkos:
Yes, @PaulWessel, I think we can avoid a lot of pitfalls by allowing one of the following:
- [x]  Use a CPT file (discrete or continuous) directly without scaling.
- [x] Use a CPT file (discrete or continuous) with scaling but no specified intervals. In this case the soft hinges are trivial, the hard hinges a bit more complicated but not extensively so. As you wrote in one of the rst changes above: it makes one CPT into a concatenation of two CPTs with independent scaling.
- [x] Convert a continuous CPT without hinges to a discrete one by using three parameters in the -C option.

The following are already more complicated, with decreasing benefit:
- [x] Convert a continuous CPT with soft hinges to a discrete one. Now the soft hinge needs to coincide with the middle of an interval (in which case the interval gets that colour) or the edge of it (in which case the hinge can be ignored). Either way, the interpolation is per usual. If the hinge is at neither of those positions, the interpolation should still follow the usual interpolation rules, but I can only expect strange results. This can still be allowed.
- [ ] Convert a continuous CPT with hard hinges to a discrete one. This is definitely prone to complications. The hard hinge is likely to have to appear, say, at 0 meters sea level. Now this value needs to be at the edge of an interval. And there need to be an integer number of intervals on either side of it. Nothing else should be allowed. Or it should not be allowed altogether.
- [ ] Convert a discrete CPT with or without hinges to a continuous one. This is never a good idea! Should not be allowed.
- [ ] Convert a discrete CPT with or without hinges to a discrete one with different intervals. Same as above. I would disallow this.

remkos:
For the cases we do not allow `grd2cpt` or `makecpt` should simply say: "the input color map you specified was not designed to be altered in the way you intend".

PaulWessel:
Hel @seisman or @joa-quim.  I have trouble with the -j docs.  We define this in common_SYN_OPTs.rst_:

`.. |SYN_OPT-j| replace:: **-j**\ :ref:`flags <-distcalc_full>`
`

Then we have in explain_distcalc_full.rst_ this line:

`.. _-distcalc_full:`

However, when building I get lots of these:

`source/common_SYN_OPTs.rst_:43: WARNING: undefined label: -distcalc_full (if the link has no caption the label must precede a section header)`

Looking a similar cases such as explain_-icols_full.rst_ the only thing I can see is the lack of a - in the distcalc file name.  Does that matter or is there something else going on?

seisman:
@PaulWessel I just built this branch and didn't see warnings. Maybe you need to cleanup the build directory and rebuild it.

PaulWessel:
I've done quite a bit of work on this branch now, addressing previous weaknesses.  Here is a summary of changes:

- Clarity on how hinges work.  If a hinge is outside the desired range then the hinge will be deactivated.  This make complete sense for CPTs with soft hinges.  For hard hinges I now give stern warnings that this is bad, but there are other more serious issues with this - see more below.
- I have improved how making CPTs from lists of colors work (**-C**_color1_,_color2_,... and **-T**).  In fact, I added a new section to the cookbook under CPTs which covers the four different scenarios and includes a new illustration (_GMT_colorlist.sh_).
- I have added a new test script based on @dongdong's failing script above [_cpthinges.sh_].  This now works.
- I have cleaned up the purpose of **-Z**.  It was a leftover option from the old days to create continuous CPTs. However, since we now refuse to make a continuous CPT from a discrete one, the only purpose of **-Z** is to create continuous CPTs from lists of colors when the _z_-nodes are not equidistant.  This is explained in the new section.
- Following on with **-Z**, I have cleaned up scripts that did things like **-T**_min_/_max_/_inc_ **-Z** since it is meaningless when sampling continuous master tables.  It should just be **-T**_min_/_max_.  Backwards compatible though.
- I have converted several old, discrete master CPTs to continuous CPTs.  The only remaining discrete CPTs are _ibsco_, _mag_, _panoply_, part of _globe_, and the categorical CPTs _categorical_ and _paired_.  These will stay as they are since they are not really under our jurisdiction.

We should discuss and decide what should happen when a user uses a master CPT with a hard hinge at zero and the data range does not include the zero.  This is not an unusual case: Consider something like a map of deep Africa: `gmt grdimage @earth_relief_10m -R20/30/10/20 -JM6 -B -pdf map`.  Here there is only positive topography so we are violating the hinge-rule.  Yet, this should not lead to an error and stop, nor should it lead to a shoreline color up in the air:

map

Remember, this is our default CPT for this sort of thing.  One possibility is to determine which half of the CPT should be excluded, then drop that half and only use the other half.  In the Africa case we would end up with

map

I think that is a much better solution for these auto-CPT selections with hinges. Similarly, a map of just a piece of the ocean should likewise not end up with a fake shoreline but only use the lower half of the CPT.  One variant of this solution would be to always include the hinge in the CPT, i.e., either start at 0 and up to max for topography and start at min and end at 0 for bathymetry.

Let me know what you think, @GenericMappingTools/core.

seisman:
> If a hinge is outside the desired range then the hinge will be deactivated.

I agree, but when the hinge is exactly at the limits of a range, the hinge should still be activated. See the script below. In this case, I expect to have a white-to-red CPT from 0 to 10. 

```
$ gmt makecpt -Cpolar+h0 -T0/10 -V
makecpt [WARNING]: gmt_stretch_cpt: CPT hard hinge is outside actual data range - hinge is ignored and your result is likely to be terrible.
0	blue	5	white
5	white	10	red
B	black
F	white
N	127.5
```

PaulWessel:
Yes, that should be doable - will have a look.

seisman:
As for the hard hinges, both your two solution sound reasonable to me, and I prefer the first one.

doc/rst/source/cookbook/features.rst
```diff
@@ -1128,45 +1128,88 @@ Master (dynamic) CPTs
...
-Dynamic CPTs will need to be stretched to the user's preferred range, and there
+and all hard hinges occur at data value *z = 0*.
+Other CPTs may instead have a *soft* hinge which indicates a natural hinge or transition
+point in the CPT itself, unrelated to any natural data set *per se*. These CPTs
+are flagged by the special comment
+
+| ``# SOFT_HINGE``
+
+CPTs with soft hinges behave as regular (non-hinge) CPTs *unless* the user
+appends **+h**\ [*hinge*] to the CPT name.  This modifer will convert the soft
```
seisman:
```suggestion
appends **+h**\ [*hinge*] to the CPT name.  This modifier will convert the soft
```

doc/rst/source/cookbook/features.rst
```diff
@@ -1128,45 +1128,88 @@ Master (dynamic) CPTs
...
 The CPTs distributed with GMT are *dynamic*.  This means they have several
 special properties that modify the behavior of programs that use them.
-All dynamic CPTs are normalized in one of two ways: If a CPT was designed
+Dynamic CPTs comes in a few different flavors: Some CPTs were designed
 to behave differently across a *hinge* value (e.g., a CPT designed specifically
 for topographic relief may include a discontinuity in color across the
-coastline at *z = 0*), then the CPT's *z*-values will range from -1, via 0
-at the hinge, to +1 at the other end.  The hinge value is specified via the special
-comment
+coastline at *z = 0*), and when users select these CPTs they will be stretched
```
seisman: Describe how it behaves if *hinge* is outside the data range.

doc/rst/source/cookbook/features.rst
```diff
@@ -1128,45 +1128,88 @@ Master (dynamic) CPTs
...
+node points in the CPT by providing a comma-separated list of values (**-T**\ 0,4,5.5,6). Now, the constant
+color intervals have unequal ranges, illustrated by red (*z = 0-4*), yellow (*z = 4-5.5*) and purple (*z = 5.5-6*).  Finally, we
+create a continuous color table (d) with arbitrary nodes by giving **-T**\ 0,2,6 and adding **-Z**;
+the latter option forces a continuous CPT pinned to a given list of node values.  Now, the colors
+continuously change from red (at *z = 0*) via yellow (at *z = 2*) to purple (at *z = 6*).
+Modules that obtain the *z*-range indirectly (e.g., :doc:`/grdimage`) may use the exact data range
+to set the quivalent of a **-T**\ *min/max* option.  You may append **+i**\ *dz* to the
+color list to have the *min* and *max* values rounded down and up to nearest multiple of *dz*, respectively.
+
+.. figure:: /_images/GMT_colorlist.*
```
seisman: It would be more clear to put the labels (a)-(d) on the left, and have the options (e.g. `-Cred,yellow,purple -T0/6`) on the right.

src/gmt_api.c
```diff
@@ -7145,7 +7146,8 @@ void *GMT_Read_Data (void *V_API, unsigned int family, unsigned int method, unsi
 				size_t len = strlen (file), elen;
 				char *ext = (len > 4 && strstr (file, ".cpt")) ? "" : ".cpt", *q = NULL;
 				elen = strlen (ext);
-				if ((q = gmtlib_file_unitscale (file))) q[0] = '\0';	/* Truncate modifier */
+                if ((q = gmtlib_file_unitscale (file))) q[0] = '\0';    /* Truncate modifier */
```
seisman: Incorrect indentation due to mixed tabs and whitespaces?

src/gmt_support.c
```diff
@@ -7151,7 +7173,7 @@ void gmtlib_free_palette (struct GMT_CTRL *GMT, struct GMT_PALETTE **P) {
 /*! Adds listing of available GMT cpt choices to a program's usage message */
 int gmt_list_cpt (struct GMT_CTRL *GMT, char option) {
 	gmt_message (GMT, "\t-%c Specify a colortable [Default is %s]:\n", option, GMT_DEFAULT_CPT_NAME);
-	gmt_message (GMT, "\t   [Notes: R=Default z-range, H=Hinge, C=colormodel]\n");
+	gmt_message (GMT, "\t   [Legend: R = Default z-range, H = Hard Hinge, S = Soft Hinge C = Colormodel]\n");
```
seisman:
```suggestion
	gmt_message (GMT, "\t   [Legend: R = Default z-range, H = Hard Hinge, S = Soft Hinge, C = Colormodel]\n");
```

seisman:
Found a user error not caught by GMT. I accidentally appended **+h** to the **-T** option and I got NaNs:
```
$ gmt makecpt -Cpolar -T-10/5+h
NaN	blue	NaN	white
NaN	white	NaN	red
B	black
F	white
N	127.5
```

src/gmt_api.c
```diff
@@ -7145,7 +7146,8 @@ void *GMT_Read_Data (void *V_API, unsigned int family, unsigned int method, unsi
 				size_t len = strlen (file), elen;
 				char *ext = (len > 4 && strstr (file, ".cpt")) ? "" : ".cpt", *q = NULL;
 				elen = strlen (ext);
-				if ((q = gmtlib_file_unitscale (file))) q[0] = '\0';	/* Truncate modifier */
+				if ((q = gmtlib_file_unitscale (file))) q[0] = '\0';    /* Truncate modifier */
+                else if ((q = strstr (file, "+h"))) q[0] = '\0';    /* Truncate +h modifier */
```
seisman: The indention of this line needs fixed.

src/gmt_api.c
PaulWessel: Yes, Xcode messes up my code.

doc/rst/source/cookbook/features.rst
PaulWessel: Done.

src/gmt_api.c
PaulWessel: Done

seisman: No, it's another line at line 7150.

seisman:
As for the documentation, there are several subsections about CPTs, and the "general features" chapter is also too long. Maybe it's better to have a separate chapter for CPTs.

PaulWessel:
Yes perhaps, we should go through and see if the partitioning is helpful or if a better layout would work better.  Will focus on fixing the last hinge-related issues first.

seisman:
Yes, we can work on the cookbook after this PR is merged.

PaulWessel:
I have implemented the fixes we discussed.  Basically, the policy on hinges becomes this:

- Hard hinge: If selected range does not include the hinge then we adjust the range so it does.  The unused side of the CPT is discarded.
- Soft hinge: If selected range does not include the hinge then we ignore the hinge.

I only had to fix one example in th tutorial (tut_19) where a hard hinge CPT was used to plot relief in the Rockies and the water colors were being used.  Now switched to cpt DEM2 and all is well.  Let me know if any other oddities remain.

seisman:
If the _hinge_ is exactly at the range limits, it should raise a warning. See the script below:
```
gmt makecpt -Cpolar+h0 -T0/10 -V
makecpt [WARNING]: gmt_stretch_cpt: CPT hard hinge is outside actual data range - range adjusted to start at hinge 0 and below-hinge CPT ignored.
0	white	10	red
B	black
F	white
N	127.5
```
---

*soft hinge* should be ignored, right? In the script below, the lower limit should be 0.
```
gmt makecpt -Cpolar+h-1 -T0/10
-1	white	10	red
B	black
F	white
N	127.5
```

---

Should add a paragraph in the cookbook, explaining the behavior when hinge is outside the data range in the cookbook. 

---

A unchecked user error:
```
$ gmt makecpt -Cpolar -T-10/5+h
NaN	blue	NaN	white
NaN	white	NaN	red
B	black
F	white
N	127.5
```

PaulWessel:
Just so I understand, your first issue is the nature of the message (says outside but should say at the hinge or something?).  The other two I understand and agree.

seisman:
> Just so I understand, your first issue is the nature of the message (says outside but should say at the hinge or something?). The other two I understand and agree.

No, it shouldn't raise any warnings. Sorry, I had a typo in my previous comment.

PaulWessel:
Not sure about that.  The combination of hinge and -T here causes us to throw away "half" the CPT.  It is only at -T-eps/10 that we will actually use the lower part of the CPT.

PaulWessel:
The **-T** parsing should probably have been fixed separately, but anyway.

seisman:
For a hard-hinge master CPT, if the **-T** range doesn't include 0, then the upper or lower limit of the new CPT is always 0. Is it intended?
```
$ gmt makecpt -Crelief -T500/1000
0	70/120/50	62.5	120/100/50
62.5	120/100/50	125	146/126/60
125	146/126/60	250	198/178/80
250	198/178/80	375	250/230/100
375	250/230/100	500	250/234/126
500	250/234/126	625	252/238/152
625	252/238/152	750	252/243/177
750	252/243/177	875	253/249/216
875	253/249/216	1000	white
B	black
F	white
N	white
```

PaulWessel:
Yes, it was intentionally implemented that way.  However, it does not have to be that way.  I just thought if you are using a hard hinge CPT that you should not be able to turn the color at _z_ = +eps into the color at 2000 meters elevation.  Remember, these hard CPTs are tied to specific data sets and have hinge values in those units.  Many of them have light/green colors just above the shoreline, and if we allowed _z_low_ to be 2000 meters then those colors would suddenly be painted at high elevations.  That is why I chose this implementation.

PaulWessel:
Hi @joa-quim, see the last exchange between me and @seisman.  Should we extend the range to include the hinge (like I have done) or do we simply chop off the unused half of the CPT and treat the remainder as a regular CPT and sample it accordingly?  I.e., the hard hinge feature dies if not in the selected range.

joa-quim:
I'm still a bit hesitant but I think I prefer the way you made it. It might be more surprising to user but gives consistent results either if the hinge is within the selected range or not.

joa-quim:
But we need a midway solution. See Dongdong's example and imagine the palette will be used in ``psscale``. We don't want the 0-500 colors to be plotted. They were not asked. So, yes, compute it like that but don't output the non-asked chunk.

PaulWessel:
OK, can easily do that.  So basically the hard hinge rules are:

If user range straddles the hinge, resample the two halves separately (like always)
If hinge is outside then discard one of the two CPT halves and resample as if there were no hinge.

For soft hinge the difference is we do not discard the other half and simply just remove the hinge and use the whole CPT.

I just ran the tests and no ill effects.  And as for @seisman's example:

```
gmt makecpt -Crelief -T500/1000
500	70/120/50	531.25	120/100/50
531.25	120/100/50	562.5	146/126/60
562.5	146/126/60	625	198/178/80
625	198/178/80	687.5	250/230/100
687.5	250/230/100	750	250/234/126
750	250/234/126	812.5	252/238/152
812.5	252/238/152	875	252/243/177
875	252/243/177	937.5	253/249/216
937.5	253/249/216	1000	white
B	black
F	white
N	white
```

I still do no like the idea of applying colors intended for near-coastal regions to elevations at much higher levels.  That was prevented by my initial implementation of including the hinge in the output.
But perhaps that is not such a good argument.  Even in my implementation the high-mountain whites are coming down (like here) to 1000 m.  So perhaps stick with the new plan?

joa-quim:
But now you changed it. What I meant was to leave as in first Dongdong's example (and as you had it) but just not write the not asked part. The cpt should start at
```
500	250/234/126	xx	rr/gg/bb
...
```

seisman:
From the view of writing documentations, I expect to have a few simple rules about hinges. Currently, soft hinge and hard hinge behaves differently, making it a little difficult to document (it also means difficult to understand for users).

Here are my thoughts about hinges (sorry to propose new things after you already did a lot of coding). CPT with hinges should follow simple and unified rules: 

1. If *zmin<hinge<zmax*, resample the negative part to *zmin*-*hinge*, and positive part to *hinge*-*zmax*
2. If *hinge<zmin*, resample positive part to *hinge*-*zmax*, but only output *zmin*-*zmax*
3. If *hinge>zmax*, resample negative part to *zmin*-*hinge*, but only output *zmin*-*zmax*
4. the only difference between hard hinge and soft hinge is, hard hinge implies **+h**0, but the user hinge can be overwritten by **+h***hinge*.

The benefits of such rules:
1. the behaviors of soft and hard hinge are more consistent, as long as users know that hard hinge implies **+h**0
2. From the coding part, the codes may be simplified. For a hard hinge CPT, we only need to set the user hinge to 0 during initialization, then we no longer care if it's soft or hard
3. It's more flexible than the current implementation. User who add **+h** explicitly should be responsible for their own choises.

PaulWessel:
I am not terribly happy about this.  However, the good news is the whole notion of soft and hard hinges are only in master and all can change, so happy to discuss this.  In essence your proposal makes just a minor distinction between hard and soft (just presetting **+h**0 so hinge is active versus soft hinges that are just dormant) but that can all change with any **+h**_z_ value.  So CPTs specifically designed for bathymetry/topography can be changed to have a hinge around a temperature of 1000C to show temperature variations across 1000 instead of topographic relief around zero.  I am just uncomfortable with this.  If you look at tutorial plot 19 in 6.0 you will see ocean colors painting land.  This no longer happens due to rule (2) above.  However, allowing **+h**1000 with our current hard-hinge CPTs would bring those ocean colors back in.

joa-quim:
- I agree with rules 1-3 (that's what I've been trying to say)
- I'm very lost with hinges. Are hard & soft hinges different or not. hinge = 0 is not a distinction

PaulWessel:
I am also OK with rules 1-3.  The issue I have is with hard versus soft hinges.  A hard hinge means what it says, it is inflexible because it is tied to a physical feature (shoreline).  A soft hinge is _independent_ of data and instead reflects a point of rapid color change in the CPT that could be tied to a data value hinge, and the user can do so via **+h**_z_.  I am fine with the idea of using _polar_ to show deviation in seismic velocities from 6.0 km/s, for instance, via `gmt makecpt -Cpolar+h6 -T4/8`.  But for a hard hinge I don't think  `gmt makecpt -Cgeo+h1000 -T2000/4000` should be allowed.  It violates the intent of that particular CPT and potentially makes stupid GMT maps.  Perhaps we could get @remkos and @leouieda to weigh in on these issues.

seisman:
With rules 1-3, `gmt makecpt -Cgeo -T2000/4000` give what you want exactly, right? Like what I said, users who explicitly use **+h1000** are responsible for their own choices.

joa-quim:
If we can change hard hinges ... they are no hard anymore and then we should only have *hinges*

PaulWessel:
Yes.  We either have hard vs soft hinges, or just hinges.  The command above gives (with rule 2):

```
gmt makecpt -Cgeo -T2000/4000
2000	0/97/71	2020.40816327	16/122/47
2020.40816327	16/122/47	2204.08163265	232/215/125
2204.08163265	232/215/125	2489.79591837	161/67/0
2489.79591837	161/67/0	2693.87755102	100/50/25
2693.87755102	100/50/25	3142.85714286	110/109/108
3142.85714286	110/109/108	3632.65306122	255/254/253
3632.65306122	255/254/253	4000	255/254/253
```

but `gmt makecpt -Cgeo+h3000 -T2000/4000 -V` gives

```
makecpt [WARNING]: Cannot override hard hinges in CPT geo.cpt with +h. Modifier is ignored.
makecpt [WARNING]: gmtsupport_validate_cpt: CPT hard hinge is outside actual data range - range adjusted to start at hinge 2000 and below-hinge CPT ignored.
```

joa-quim:
So we are not talking the same rules. ``0/97/71`` is color of z = 0 in *geo*. Not ``z = 2000``. 
This is not what you implemented at the time of https://github.com/GenericMappingTools/gmt/pull/2416#issuecomment-574978063

PaulWessel:
Well, doing otherwise takes us to **-G**:

```
-G Truncate incoming CPT to be limited to the z-range <zlo>/<zhi>.
	   To accept one of the incoming limits, set that limit to NaN.
```

My first implementation stretched zmin or zmax to reach the hinge at 0.  The above is the case where we leave zmin/zmax along and just toss the other half of the CPT.  To strictly interpolate 2000/4000 to mean just that section of geo has not been done (but possible via -G).

PaulWessel:
Should have said that CPTs with a hard hinge, like any CPT, are stretchable, hence the color at the users 2000 will be the color from the hinge.  Otherwise the CPT would be completely static and unchangeable.  You then run

gmt makecpt -Ggeo > mine.cpt

and use that.

PaulWessel:
I would like this to get resolved so we can plan 6.1

CPTs with hard hinges and ranges (e.g., geo with range RANGE = -8000/8000) can obviously be stretched and resampled: We may want to go from -12000 to +9000 meters, for instance.  The range of the CPT is never fixed; it is just a default range when none is given in makecpt.  Hence, commands like `gmt makecpt -Cgeo -T2000/4000` cannot simply pull out colors from the original 2000/4000 subset of the -8000/8000 range.  That would mean we could not use this cpt outside the -8/8 km range.  Instead, we resample the positive elevation CPT (hinge 0 to 8000) and compress it to fit the desired 2000/4000 meter output.  So yes, the first color at 2000 is the one at 0 in the original CPT and the last color at 4000 is the color at 8000 in the original CPT.

There may be cases where you want to extract only a subset of the CPT _colors_.  We added **-G** for that, but this is a very special operation that few people would ever need.

Thus, it seems the final decision we are stuck at has to do with the meaning of a hard hinge.  I am arguing that it should be immutable: It is what it is, and the designer of the CPT wants it to be a fixed boundary tied to the physical properties it was intended for (topographic relief).  In contrast, a _soft_ hinge is mutable; you can change it with **+h**_z_.  If it were not so and **+h** could apply to the hard hinges as well, then the only difference there would be between hard and soft is this:

hard: There is _always_ a hinge (no **+h** needed to activate it) and we stretch the output differently, except when user range excludes the hinge where we discard the unused half of the CPT, then sample normally.
soft: It requires a **+h** modifier to activate the hinge.  After that it behaves as a hard hinge.

joa-quim:
Re-reading this again I have no strong opinion on the two alternatives, but the second (to let a hard hinge soften a bit and move) gives more freedom without increasing confusion. The last paragraph

``hard: There is always a hinge (no +h needed to activate it) and we stretch the output differently, except when user range excludes the hinge where we discard the unused half of the CPT, then sample normally.
soft: It requires a +h modifier to activate the hinge. After that it behaves as a hard hinge.``

is what should go into the docs.

PaulWessel:
OK, I can live with that.  After all, it takes an unprecendented act on part of the user to move the geo hinge away from 0 via **+h**, so not likely to happen by accident.  I will implement the movable hard hinge and see how it goes.

PaulWessel:
Hi @joa-quim and @seisman.  Let me know if this is the expected behavior.  If you agree then I can adjust the docs and request a review.

t

seisman:
Yes, this is exactly what I expected.

joa-quim:
Yes, that's how I understood it. I assume that an equivalent of case d) for soft hinges, e.g. *-Cpolar+h -T10/20* will give the same cpt as i) but with [10 20] limits.

PaulWessel:
Yes:
```
gmt makecpt -Cpolar+h -T10/20 
10	white	20	red
```

PaulWessel:
Hi @seisman, if you are happy with the latest merge in this PR please review and then we can fix any issues and merge this back to the mothership.

seisman:
A crash by a user error:
```
$ gmt makecpt -Cwhite,blue -T0,1
makecpt [ERROR]: Mistmatch between number of entries in color list and z list
gmt(56804,0x106bcf5c0) malloc: *** error for object 0x7fd9d8700d40: pointer being freed was not allocated
gmt(56804,0x106bcf5c0) malloc: *** set a breakpoint in malloc_error_break to debug
[1]    56804 abort      gmt makecpt -Cwhite,blue -T0,1
```

seisman:
@PaulWessel Remember to click the "Squash and merge" button when merging. Don't use the first one.

image

PaulWessel:
Thanks, done.

PR #2417, opened 2020-01-07
Add a new map projection 

yangleir:


yangleir:
This is just a request. Maybe I got the wrong place. There is no relation to the #2417.

PR #2419, opened 2020-01-07
Add missing dependencies to CMake rules for gmt_dimensions.h

remkos:

This adds the proper dependencies to the rules to make gmt_dimensions.h
This came to bare in trunk when gmt_ellipsoids.h was extended but its dimension was not updated.

This commit needs also to be merged with trunk.


remkos:
It looks like the failures for Linux should be ignored. They are just the result of errors in the checking process.

PR #2422, opened 2020-01-07
Test that values passed to -S option make sense

joa-quim:
Do not accept below *below* > *low* or *above* < *high* or *between* not between *low* and *high*


joa-quim:
Yep, that's why one day we'll have to start one of those fuzzy testers.

PR #2424, opened 2020-01-08
grdfilter area weight wrong but does not matter

PaulWessel:
The area of grid cells calculation in grdfilter used a wrong conversion from degree to radians (scaled longitude increment by **R2D** instead of **D2R**), but this affected all cells equally and hence the constant cancelled out in the _sum_wz/sum_w_ step.
I checked the similar calculation in gmt_cell_area (used in grdmath) and it has correct conversion.  I double-checked this by running 

```
gmt grdmath -Rg -I30m AREA = a.nc --PROJ_ELLIPSOID=1
gmt grdmath -f0-1f a.nc SUM 1e6 MUL = s.nc
```

and s.nc said total area was 12.56 (=4*pi).  The 1e6 above is there since we want to report in km^2.


remkos:
It makes mathematical sense that in this case the scaling error did not matter, because it was done on longitude. Had it been latitude and then sines or cosine taken would be a different story altogether.
Merging.

PR #2430, opened 2020-01-09
Ensure column scaling is performed correctly with -i

### LINKED ISSUE #2429: Record parsing error for some -i selection
See [the forum](https://forum.generic-mapping-tools.org/t/gmtselect-i-o-option/182/5) for background.  In that case, `-i5+o0.00005,6+o0.00005,0-7` ended up only offsetting column 5 by 0.0005 but not column 6.  Also, the repeating of columns 5 and 6 in 0-7 would reapply the offsets despite not requested there (and still do it incorrectly for column 6).

Having looked at the case I wish to implement a simpler row parsing in _gmtio_ascii_input_. We should first convert the actual data columns into an internal buffer, then create the logical input record from these values and the order and translations specified via the **-i** selections.

PaulWessel:
See #2429 for background.  This PR fixes this problem by redesigning our _gmt_convert_col_ macro to _not_ change its argument but instead to return the (possibly translated) value.  That way the original physical column value is kept as is and may be translated or not translated as **-i** dictates repeatedly.  Closes #2429.

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -198,7 +198,7 @@ typedef enum {
 } SwapWidth;
 
 /*! Macro to apply columns log/scale/offset conversion on the fly */
-#define gmt_convert_col(S,x) {if (S.convert) x = ((S.convert & 2) ? log10 (x) : x) * S.scale + S.offset;}
+#define gmt_convert_col(S,x) ((S.convert) ? ((S.convert & 2) ? log10 (x) : x) * S.scale + S.offset : x)
 
 /* These functions are defined and used below but not in any *.h file so we repeat them here */
 int gmt_get_ogr_id (struct GMT_OGR *G, char *name);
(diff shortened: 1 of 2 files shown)


seisman:
Are we still maintaining GMT 5? If yes, we probably need to also apply this commit to 5.4 branch.

PaulWessel:
Hm, I know we have a 5.4 branch but I have not kept it up.  I prefer to say that 5.4.? was the last GMT 5 version.

seisman:
The latest GMT5 version is 5.4.5. We definitely won't release 5.5.x, but are we going to have 5.4.6, just like what you did to GMT4?

joa-quim:
GMT5 is done. No more updates. Would be a pain going back to VS2013, old GDAL etc.

seisman:
OK, I agree.

Issue #2436, opened 2020-01-12
Static build doesn't work


seisman:
Tried to have a static build by setting BUILD_SHARED_LIBS to FALSE. The build is succesful, and `gmt` can print the splash screen, but `gmt coast` says `No module named coast was found.`

PaulWessel:
tool -L gmt says what?

seisman:
```
$ otool -L gmt
gmt:
	/usr/local/opt/netcdf/lib/libnetcdf.15.dylib (compatibility version 15.0.0, current version 15.0.0)
	/usr/lib/libcurl.4.dylib (compatibility version 7.0.0, current version 9.0.0)
	/usr/local/opt/gdal/lib/libgdal.20.dylib (compatibility version 26.0.0, current version 26.2.0)
	@rpath/libpcre.1.dylib (compatibility version 4.0.0, current version 4.11.0)
	/usr/local/opt/fftw/lib/libfftw3f.3.dylib (compatibility version 9.0.0, current version 9.8.0)
	/usr/local/opt/fftw/lib/libfftw3f_threads.3.dylib (compatibility version 9.0.0, current version 9.8.0)
	/System/Library/Frameworks/Accelerate.framework/Versions/A/Accelerate (compatibility version 1.0.0, current version 4.0.0)
	/usr/local/opt/glib/lib/libglib-2.0.0.dylib (compatibility version 6201.0.0, current version 6201.4.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.250.1)
```
The file size of the gmt executable is ~4.6 Mb.

PaulWessel:
So is there static libraries being built. All that stuff in libgmt should now be in gmt.exe, no? But that cannot be the case.  Are there any libgmt* files being produced?

seisman:
I think the static libraries are already included in the gmt executable. That's why the gmt.exe is as large as 4.6 Mb.

The static libraries are still being built in the lib directory. Below are what I have in my lib directory

```
.
├── gmt
│   └── plugins
│       └── supplements.so
├── libgmt.6.1.0.dylib
├── libgmt.6.dylib -> libgmt.6.1.0.dylib
├── libgmt.a
├── libgmt.dylib -> libgmt.6.dylib
├── libpostscriptlight.6.1.0.dylib
[...]

PR #2438, opened 2020-01-12
Allow setting BUILD_SHARED_LIBS to false for static build

PaulWessel:
Creating a static GMT failed to produce working executables.  There were several issues to address:

1. The gmt_core_module.c produced by gmt_make_module_src.sh had a function _gmt_core_module_lookup_ that we probably had never tested before.  It was furthermore out of date, not knowing about modern module names.  This has now been addressed and modules are again being found.
2. The other problem was the reliance on a static array in gmt_init.c.  For reasons not entirely clear to me, when supplemental modules were called and these again called functions in the static library, that static array (hash codes to handle parsing of GMT defaults) was reinitialized to zeros and no GMT defaults could be found.  Making it a structure under the API private pointer means it is initialized once by GMT_Create_Session and does not need to me modified again during that session.
3. The problem above is likely related to the fact that only the GMT core collection of modules are in the static library.  The supplemental library (and any other GMT-compatible external modules, such as the MB-system plugins) remain as shared libraries accessed via the plugins (*.so files).

I have build both static and shared and run all tests and all seems well.  It is important to give it a spin under Windows since shared/static behaves differently and I cannot guarantee that my solution works for Windows.  Hence, need feedback from @joa-quim and @seisman on this.

Otherwise, closes issue #2436.


seisman:
@PaulWessel Please remove the file "src/init" in this branch.

seisman:
GMT still can't find supplemental modules, and the Linux and Windows builds fail probably due to some inconsistent compiler options. See the CI reports in https://github.com/GenericMappingTools/gmt/pull/2440 for details.

joa-quim:
I’m away from the computer till Tuesday so can’t revisit but I have the impression that the supplements rely exclusively on dlopen and that cannot be made static.  

Sent from my iDedo

PaulWessel:
So I had to delete several builds that had libgmt.dylib and reboot _before_ my static build would run and find the supplements.  It would run fine in Xcode debug but then the same code would report not finding the supplements from the command line...I think the code is correct (for macOS for sure) but boy, Apple really is doing something odd under the hood with shared libraries and links.

PaulWessel:
@joa-quim, the supplemental and all other external libs will remain shared forever.  The static is _just_ about the core.  We have no way of building static with libraries we don't even maintain (e.g., mbsystem.so).

PaulWessel:
Sure this is not OK? All checks passed.

seisman:
The CI in this PR are building shared libraries. See #2440 for static builds.

PaulWessel:
Hi @seisman, I guess this PR can be deleted once #3133 is merged?

seisman:
Yes, we probably need to rethink about it if someone wants a static build in the future.

Issue #2439, opened 2020-01-12
Add progress bar option to movies


PaulWessel:
**Description of the desired feature**

While **movie** offers automatic labeling of frame number, elapsed time, or real time via the **-L** option, it is often of interest to add a progress bar that indicates how far into the animation we have reached. Having given this some thought, here are some proposed progress bars (showing a snapshot at frame 185 of 555, or a progress of 33%):

![progressbar](https://user-images.githubusercontent.com/26473567/72216636-d0890b80-34c7-11ea-9031-24f02fdc2c03.png)

The illustration shows three circular and three linear progress indicators.  I imagine a currently available option like **-E** or **-K** in **movie** could follow a syntax like this:

**-E**[_just_][**+w**_width_][**+p**_pen1_[/_pen2_]][**+g**_fill1_[,_fill2_]][**+**_offset_][**+a**[**e**|**f**|**p**|**c**_col_]][**+s**_kind_]

where most of these modifiers are familiar from the placement of other types of embellishments (e.g., _just_ for positioning, _width_ for bar/circle width, _offset_ to move away from the anchor point). There may be 1-2 pens and 1-2 fills to specify (the progress bars typically have two colors, either via lines or via fills), and **+a** controls if there is any annotation and if so if we should display **e**lapsed time, **f**rame counter, **p**ercentage, or the time from **c**olumn _col_ (this is similar to codes in **-L**). Finally, **+s** selects which progress indicator from this list (a-f).  There should be reasonable defaults for most settings and even just **-E** should plot a reasonable progress bar [we need to pick a default bar - maybe a)?].

Since all such indicators consist of a constant part (e.g., the background line, the filled circle) and a changing part (the inner line, the wedge or partial arc, the percentage text) I can implement it efficiently.  As for labels via **-L** the **-E** option would be repeatable.

Let me know if @GenericMappingTools/core or anyone else have any comments or suggestions.

seisman:
It's really cool! 

For the two pens and fills, better to use the same separator, i.e.  [**+p**_pen1_[/_pen2_]][**+g**_fill1_[/_fill2_]].

PaulWessel:
problem is things like the colors 233/14/15 and 13/155/200, it would look odd I think with **+g**233/14/15/3/155/200, no?  Alternative is a second **+g**-like option, say **+G**.  Same for second pen **+P.**

seisman:
You're right. **+P** and **+G** seem OK to me.

[...]

PR #2442, opened 2020-01-12
Implementing movie progress indicators

PaulWessel:
See #2439 for background.  I am implementing those six types with a) as the default.


PaulWessel:
I have implemented the six types of movie progress indicators (a-f) and documented them in the movie module (with picture).  Tested reasonably well.  Most have reasonable defaults, so one for instance can add just **-Pa** and get a good result.  The doc script GMT_movie_progress.sh generates the default looks of all 6 types and that file is included in the movie.rst documentation.  As you may know, these features need to be added during the last steps of each frame, hence movie communicates with the main script via environmental variables.

PaulWessel:
Here is a movie that shows the six progress indicators in action at the same time:
[Progress indicators](https://www.dropbox.com/s/sns67ab10sr2qqt/GMT_movie_progress.mp4?dl=0)

doc/scripts/GMT_movie_progress.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+#
+# Makes a plot of the six movie progress indicators avalable
+#
+# Make dummy map script for "movie"
+cat << EOF > map.sh
+gmt begin
+	gmt set MAP_VECTOR_SHAPE 0.75
+	gmt psbasemap -R0/10/0/5 -JX7.6i/3.4i -Bafg -BWSrt+gbeige
```
seisman:
```suggestion
	gmt basemap -R0/10/0/5 -JX7.6i/3.4i -Bafg -BWSrt+gbeige
```

seisman:
The label positions are incorrect for SI units.
image

PaulWessel:
Yes, found that last night too.  Need to track down where that is happening.

PaulWessel:
Fixed. Note that movie -CHD etc will create slightly different canvas sizes for inch and cm (9.6x5.4" vs 24x13.5 c), but these both rasterize to the same number of pixels.  I updated the PS after running the tests which run in inch mode.

seisman:
**+ae** doesn't work well.

image

PaulWessel:
Seems OK.  For that particular case we are 0.41 seconds into the movie and we don't display fractional seconds, hence 00.  Use -M30 and you will be 01.  There is the +sscale modifier to scale frame to time (default uses frames per sec) but I need to move it until after the mode is set...

PaulWessel:
Thanks.  I improved things further with good default values for everything so **-P**? is all you need to get started.  Detected a few bugs in our movie scripts (4,8), added progress indicators to 3,7, simplified movie **-T** a bit (no need for +w if you imply it via **-L**, **-P**), updated docs for movie with all the defaults, etc.  I have run all the tests and built the docs with updated animations and all look well to me.  I will merge in 45 minutes unless you notice anything odd.

Issue #2450, opened 2020-01-15
Improve font switch @%fontno%


seisman:
A few "issues" with font switch @%fontno%:

1. Doesn't check if *fontno* is in the allowed range
2. Doesn't accept *font name*.

A simple script to reproduce it:
```
echo 5 5 'TEXT @%1%TEXT@%% @%Helvetica-Bold%TEXT@%%' | gmt text -R0/10/0/10 -JX10c -Baf -F+f20p -png map
```

PR #2454, opened 2020-01-16
Check that @%font% is a valid number of font name

PaulWessel:
We did not check. Now we do, and we also allow font names there.  A bad font number or name is replaced with 0 (Helvetica).  Closes issue #2450.


PaulWessel:
Well, they are coming from the postscriptlight library where the problem is being detected.

PaulWessel:
It would be a lot of work to have this error propagate all the way back to GMT in order to format it better.  Not worth it - at least now you get the message!

PR #2456, opened 2020-01-16
Set logo default with to 2i if not given

PaulWessel:
But I think we should change to SI, or use 144 points.  Since changing it from 2 inches would be a backwards incompatibility I suggest we list 144 point as the default.  OK?

### DIFF
--- a/doc/rst/source/gmtlogo.rst
+++ b/doc/rst/source/gmtlogo.rst
@@ -39,7 +39,7 @@ To append a GMT logo overlay in the upper right corner of the current map, but
 scaled up to be 6 cm wide and offset by 0.25 cm from the border, try::
 
     gmt begin map
-    gmt ...<plot the map using -R J>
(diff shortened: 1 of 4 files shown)


doc/rst/source/gmtlogo.rst
```diff
@@ -31,16 +31,16 @@ Examples
...
     gmt logo -pdf logo
 
 To append a GMT logo overlay in the upper right corner of the current map, but
-scaled up to be 3 inches wide and offset by 0.1 inches from the border, try::
+scaled up to be 6 cm wide and offset by 0.25 cm from the border, try::
 
     gmt begin map
-    gmt ...
-    gmt logo -DjTR+o0.1i/0.1i+w3i
+    gmt ...<plot the map using -R J>
```
seisman:
```suggestion
    gmt ...<plot the map using -R -J>
```

PR #2461, opened 2020-01-18
polar: Fix several bugs and improve documentations

seisman:
The two examples in the manpage don't work due to two bugs.

The first one, when takeoff-angle>90, symbol_size2 isn't changed, i.e.
symbol_size2=0, and symbols aren't plotted. It doesn't make sense.

The second one: I don't know much about HYPO71 format, but from the
codes and the example, the last column should be something like `ipu0`,
and the phase polarity is determined by its third character. Thus, when
read the last column, we should use `%s` instead of `%c`.

The original PS `seis_09.ps` is also updated, due to the bug 1.
When takeoff-angle=90, the symbol_size2 is set to 0.
The new PS is consistent with the output of GMT 4.5.18.

### DIFF
--- a/doc/rst/source/supplements/seis/polar.rst
+++ b/doc/rst/source/supplements/seis/polar.rst
@@ -47,7 +47,7 @@ Examples
 
 ::
 
-    gmt polar -R239/240/34/35.2 -JM8c -N -Sc0.4 -h1 -D39.5/34.5 -M5 -pdf test << END
+    gmt polar -R239/240/34/35.2 -JM8c -N -Sc0.4 -h1 -D239.5/34.5 -M5 -pdf test << END
     #stat azim ih pol
     0481 11 147 c
     6185 247 120 d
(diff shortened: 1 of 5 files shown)


nukenine:
Hello, I ran into this error, as my `GMT` version was older and upgrading allowed me to put in take-off angles that are greater than 90°. 

However, do I need to adjust my station azimuths somehow to account for the fact that my take-off angles are >90°, or can I keep my station azimuths the same? 

Additionally, is `pspolar` projecting the station locations down to the lower focal half-sphere, or it plotting the upper focal half-sphere when it sees take-off angles >90° ?

seisman:
> However, do I need to adjust my station azimuths somehow to account for the fact that my take-off angles are >90°, or can I keep my station azimuths the same?

You can keep you station azimuths the same. `pspolar` adds azimuths by 180 if take-off angle are >90°.

> Additionally, is `pspolar` projecting the station locations down to the lower focal half-sphere, or it plotting the upper focal half-sphere when it sees take-off angles >90° ?

All polarity observations are plotted on the lower focal hemisphere. For take-off angles>90°, the antipodes are plotted.

PR #2463, opened 2020-01-18
psmeca & pscoupe: Add -S+a<angle> to specify angle of event label

seisman:
Add one more modifier +a<angle> to specify angle of event label.

### DIFF
--- a/src/seis/pscoupe.c
+++ b/src/seis/pscoupe.c
@@ -8,7 +8,7 @@
  *--------------------------------------------------------------------*/
 /*
 
-pscoupe will read focal mechanisms from input file and plot beachballs on a cross-section. 
+pscoupe will read focal mechanisms from input file and plot beachballs on a cross-section.
 Focal mechanisms are specified in double couple, moment tensor, or principal axis.
 PostScript code is written to stdout.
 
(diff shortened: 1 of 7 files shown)


src/seis/pscoupe.c
```diff
@@ -746,18 +747,20 @@ GMT_LOCAL int parse (struct GMT_CTRL *GMT, struct PSCOUPE_CTRL *Ctrl, struct GMT
 						break;
 				}
 
-				if ((strstr (opt->arg, "+f")) || strstr (opt->arg, "+o") || strstr (opt->arg, "+j")) { 
-					/* New syntax: -S<format><scale>+f<font>+o<dx>/<dy>+j<justify> */
+				if ((strstr (opt->arg, "+a")) || (strstr (opt->arg, "+f")) || strstr (opt->arg, "+o") || strstr (opt->arg, "+j")) {
```
seisman: @PaulWessel Here I use `strstr` four times to check if `opt->arg` contains any of the modifiers. Is there a utility function in GMT which can do the same thing but shorter?

PaulWessel: We have (in gmt_support) _gmt_validate_modifiers_ which makes sure only the listed modifiers appears and gives an error if it finds a +modifier not listed, and _gmt_first_modifier_ which returns pointer to the part of the string where the first modifier listed appears.  Perhaps you could use that one:

```
if ((first = gmt_first_modifier (GMT. opt->arg, "afjo"))) { /* Know we got one of these modifiers
   	if (gmt_get_modifier (first, 'o', string))	/* Offset of refpoint */
		if (gmt_get_pair (GMT, string, GMT_PAIR_DIM_DUP, I->off) < 0) n_errors++;
...
```

seisman: OK. Will try after merging this PR.

PR #2488, opened 2020-01-22
Clean up julia scripts

seisman:


joa-quim:
Deleting these scripts here it's trowing them away for good.

seisman:
@joa-quim Probably you can move these scripts to GMT.jl since they are never tested in the main GMT repository.

joa-quim:
If I wanted them in GMT.jl they would there already. Those scripts were written because I needed them (troubles with bash on Win). 

Let me be clear. I'm against deleting them.

seisman:
OK. Let's keep them.

PR #2504, opened 2020-01-23
Clarify -h option and header records

PaulWessel:
When **-h** is set then the first _n_ records are taken to be header records.  Only after they have been read/skipped does the rule of what to do with blank lines and records starting with # take effect.


seisman:
Merge or close?

PR #2583, opened 2020-01-28
GDAL librarify

joa-quim:
Add structure to call GDAL programs via its C API.
Current implementation already allows calling the ``gdal_grid`` program using a temporary **-A** option of ``nearneighbor``. An annoying issue is that, so far, simple input data must be passed in ascii via the GDAL's VRT mechanism. 

As example, put the following into a ``lixo.csv`` file
```
2.53857 5.16657 0
2.48365 6.26811 1
8.68883 4.55983 2
4.11104 7.78623 3
1.79704 6.53027 4
7.17493 3.81713 5
3.41052 8.18161 6
8.35062 1.43348 7
8.1706 4.46765 8
5.27815 1.15172 9
```
create a ``lixo.vrt`` file with
```
<OGRVRTDataSource>
    <OGRVRTLayer name="lixo">
        <SrcDataSource>lixo.csv</SrcDataSource>
        <GeometryType>wkbPoint</GeometryType>
        <GeometryField encoding="PointFromColumns" x="field_1" y="field_2" z="field_3"/>
    </OGRVRTLayer>
</OGRVRTDataSource>
```
and run

```
gmt nearneighbor lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -A"-a nearest"
```
to get the same as the ``nat_nn.sc`` test (but, just realized, UD flipped)

GMTjl_tmp


PaulWessel:
Since you are working in a WIP branch, do we really need to approve commits?

PaulWessel:
Ok, but later this week.  We said we would release 6.1 soon (February) so we should discuss a checklist with @serisman and see if there are specific things we need to complete first.

joa-quim:
@PaulWessel please have a go with this. I added a **-Ns** option to ``nearneighbor`` that does the same as **-Nn** but using GDAL.
Also added a wrapper module called ``gmtgdal`` which is a wrapper to the ``gmt_gdal_librarified()`` functions. The name is a bit poor and if ofc open to discussion. As well as its functionality, which for now is

```
usage: gmt gmtgdal <infile> -A<prog> [-F<gd opts>] [-G<outgrid>] [-M[+r+w]] [-W<outds>] [-R<xmin>/<xmax>/<ymin>/<ymax>[+r][+u<unit>]] [-V[<level>]]
           Specify input file name
        -A Specify the GDAL program name.
        -F list of GDAL options for the selected program in -A wrapped in double quotes.
        -G Name of output grid.
        -M Means to both read and write with GDAL. Use +r to only read or +w to write.
        -W Name of output dataset when writen by GDAL.
```
Only the GDAL's ``gdal_grid, gdalinfo`` and ``gdal_dem`` programs are  *usable* (and latter very incomplete)

PaulWessel:
OK, trying nearneighbor:

```
gmt nearneighbor lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -Ns
nearneighbor [ERROR]: Distance units must be one of d|m|s|e|f|k|M|n|u
```

What am I doing wrong?
Also, we cannot have people building vrt files - that would have to happen automatically if -Ns.

joa-quim:
Ok, do
```
gmt nearneighbor lixo.cvs -R0/10/0/10 -Gjunk.nc -I0.05 -Ns
```
``nearneighbor`` does not know how to read *.vrt* files.

If you were using ``gmtdal``, then
```
gmt gmtgdal lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest" -M+r
``` 
should work too.

PaulWessel:
I switched to the GDALlibrify branch, rebuilt everything, and

gmt nearneighbor lixo.csv -R0/10/0/10 -Gjunk.nc -I0.05 -Ns
nearneighbor [ERROR]: Distance units must be one of d|m|s|e|f|k|M|n|u
pwessel@MacBeth:~/UH/RESEARCH/CVSPROJECTS/GMTdev/gmt-dev/dbuild-> gmt gmtgdal lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest" -M+r

ERROR: No module named gmtgdal was found.  This could mean one of four cases:
Shared libraries must be in standard system paths or set via environmental parameter DYLD_LIBRARY_PATH.

Do you have any other settings in cmake that needs to be applied?

joa-quim:
Ah, forgot about it. I am using the ``set (EXTRA_MODULES gmtgdal.c)`` in ``ConfigUser.cmake``. With it, any of
```
gmt gmtgdal lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest" -M+r
and
gmt gmtgdal lixo.csv -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest"
```
work. But something screwed meanwhile with ``nearneighbor -Ns``. Will have to debug and fix.

joa-quim:
Wait, don't know why I thought it hadn't work. It does
```
$ gmt nearneighbor lixo.csv -R0/10/0/10 -Gjunk.nc -I0.05 -Ns -V
nearneighbor [INFORMATION]: Reading Data Table from File lixo.csv
nearneighbor [INFORMATION]: Cartesian input grid
nearneighbor [INFORMATION]: Cartesian input grid
nearneighbor [INFORMATION]: gdal options used: -ot Float32 -txe 0.000000 10.000000 -tye 0.000000 10.000000 -outsize 201 201 -a nearest:radius1=0.000000:radius2=0.000000:nodata=NaN -of MEM
nearneighbor [INFORMATION]: Writing grid to file junk.nc
```

joa-quim:
I'm assuming you have
```
#if defined(HAVE_GDAL) && (GDAL_VERSION_MAJOR >= 2) && (GDAL_VERSION_MINOR >= 1)
```

PaulWessel:
GDAL 3.0.3, released 2020/01/08
Not sure about your if test.  Yes MAJOR > 2 but MINOR = 0 so fails your test.  Not what you meant.  Maybe

#if defined(HAVE_GDAL) && ((GDAL_VERSION_MAJOR > 2) || (GDAL_VERSION_MINOR >= 1))

assuming there is no gdal 1 anymore.

joa-quim:
No, there are plenty of 1.11 out there GDALs (guess whose linux distro). Made it
```
#if defined(HAVE_GDAL) && ((GDAL_VERSION_MAJOR >= 2) && (GDAL_VERSION_MINOR >= 1)) || (GDAL_VERSION_MAJOR >= 3)
```
update and rebild.

PaulWessel:
OK, near neighbor ran now and gives the flipped image.  No solution to that yet?

PaulWessel:
gmt gmtgdal lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest" -M+r
0   libgmt.6.dylib                      0x000000010a496846 gmt_memory_func + 502
1   ???                                 0x00000000001300b0 0x0 + 1245360
Stack backtrace:
0   libgmt.6.dylib                      0x000000010a3d1fa3 sig_handler + 515
1   libsystem_platform.dylib            0x00007fff65fe842d _sigtramp + 29
2   ???                                 0x0000000000000000 0x0 + 0
3   libgmt.6.dylib                      0x000000010a8e1518 GMT_gmtgdal + 136
4   libgmt.6.dylib                      0x000000010a3fe354 GMT_Call_Module + 1092
5   gmt                                 0x000000010a3c423e main + 1086
6   libdyld.dylib                       0x00007fff65def7fd start + 1
pwessel@MacBeth:~/UH/RESEARCH/CVSPROJECTS/GMTdev/gmt-dev/dbuild-> ll lixo.vrt
-rw-------  1 pwessel  staff   271B Feb  8 09:41 lixo.vrt
pwessel@MacBeth:~/UH/RESEARCH/CVSPROJECTS/GMTdev/gmt-dev/dbuild-> gmt gmtgdal lixo.csv -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest"
0   libgmt.6.dylib                      0x000000010b84e846 gmt_memory_func + 502
1   ???                                 0x00000000001300b0 0x0 + 1245360
Stack backtrace:
0   libgmt.6.dylib                      0x000000010b789fa3 sig_handler + 515
1   libsystem_platform.dylib            0x00007fff65fe842d _sigtramp + 29
2   ???                                 0x0000000000000000 0x0 + 0
3   libgmt.6.dylib                      0x000000010bc99518 GMT_gmtgdal + 136
4   libgmt.6.dylib                      0x000000010b7b6354 GMT_Call_Module + 1092
5   gmt                                 0x000000010b77a23e main + 1086
6   libdyld.dylib                       0x00007fff65def7fd start + 1
pwessel@MacBeth:~/UH/RESEARCH/CVSPROJECTS/GMTdev/gmt-dev/dbuild->

joa-quim:
It's not flipped. Same as
```
gdal_grid lixo.vrt -ot Float32 -txe -0.025000 10.025000 -tye -0.025000 10.025000 -outsize 201 201 -a nearest -co ZLEVEL=5 -co FORMAT=NC4 -of netCDF junk.nc
```

PaulWessel:
You are correct.  I was looking at the flipped one on this page and not on the original.

joa-quim:
Does this crash too
```
gmt gmtgdal lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest" -M+r+w
```

**+r** means read with GDAL and **+w** write with GDAL too (does not use the GMT machinery)

PaulWessel:
Yes, same crash.

joa-quim:
OK, added gmtgdal.c to CMakeList to see if cppcheck detects something.

PaulWessel:
OK, but probably quicker to just run cppcheck locally.  I ran 
cppcheck --error-exitcode=1 --std=c99 --force --quiet gmtgdal.c 
and nothing.

PaulWessel:
I know.  You are calling

	struct GMT_GDALLIBRARIFIED_CTRL *st = gmt_M_memory (GMT, NULL, 1, struct GMT_GDALLIBRARIFIED_CTRL);;


way before GMT is even allocated.  Now, you may not use GMT inside there but I do by virtue of 

#ifdef MEMDEBUG
	if (GMT->hidden.mem_keeper->active)
		gmt_memtrack_add (GMT, where, tmp, prev_addr, nelem * size);
#endif

joa-quim:
Thanks. Too much blind copy-paste. Fixed

PaulWessel:
Now those commands run.  Some feedback to help you strengthen things:

```
gmt gmtgdal lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest" -M+r
gmtgdal (gmt_grdio.c:2680(gmtgrdio_free_grid_ptr)): Wrongly tries to free item
gmt [ERROR]: Memory not freed first allocated in gmt_gdal_librarified.c:42(breakMe) (ID = 29): 0.500 kb [512 bytes]
```

Maybe you are freeing the wrong thing? One is left unfreed while another is wrongly freed.

```
gmt gmtgdal lixo.vrt -R0/10/0/10 -Gjunk.nc -I0.05 -F"-a nearest" -M+r+w
Warning 1: The dataset has several variables that could be identified as vector fields, but not all share the same primary dimension. Consequently they will be ignored.
Warning 1: dimension #1 (x) is not a Longitude/X dimension.
Warning 1: dimension #0 (y) is not a Latitude/Y dimension.
gmt [ERROR]: Max total memory allocated was 5.594 kb [5728 bytes]
gmt [ERROR]: Single largest allocation was 3.000 kb [3072 bytes]
gmt [ERROR]: MEMORY NOT FREED: 0.500 kb [512 bytes]
gmt [ERROR]: Items allocated: 26 reallocated: 4 freed: 25
gmt [ERROR]: Items NOT PROPERLY FREED: 1
gmt [ERROR]: Memory not freed first allocated in gmt_gdal_librarified.c:42(breakMe) (ID = 29): 0.500 kb [512 bytes]
```

Since only developers are doing the MEMDEBUG it does not matter that these say ERROR but I think I will change to WARNING in master anyway.  I dont know if you can control the warning levels for the gdal which is where those warnings come from.  What do those mean anyway?

joa-quim:
I have no control on the GDAL warnings (or must search more). They are a annoying.
This one happens  always (even from gdal_grid)
```
ERROR 3: An attempt to read an undefined ID from virtual variable collection was made
```
and the ``ERROR 1`` they only show up if the file (*junk.nc*) already exists. GDAL also has the nasty feature of always naming the *x,y* coordinates as *long,lat* regardless of if data is geog or not.

PaulWessel:
The VRT file: If the input is always points then we should be able to create that VRT on the fly in /tmp, add in the CSV file in that field, and pass that VRT file.  I cannot imagine we want to open GMT users up to making these files.

joa-quim:
No, much better. See the ``gdal_vector()`` function. It uses the MEM driver to create a GDAL dataset from the data we read or generate. So no need to create VRTs under the hood. However, it's interesting to maintain the capability of reading OGR formats. To do that one must use the **-M+r** option. The ``lixo.vrt`` is just a testing case for that.

PaulWessel:
See new #2697 which will affect gmtgdal for sure.
Since your -Ns option is much faster and more stable than the mess in triangulate -Qn that we call if -Nn, I think we should remove the older stuff since it is buggy.  I see no need to keep that as "backwards compatible" for anyone not building with GDAL since those users must be very few anyway.

Related question is grdfill -An.  This is Eric Xu's algorithm to fill in an existing grid based on NN.  Does the GDAL version allow this, or would we need to supply the data in the grid as input (x,y,z) to the gdal version?  Again, that one may be faster and may not require that radius argument.

joa-quim:
Yes, I wanted to bring the **-Nn** *vs* **-Ns**. Maybe we may want to keep the **-Qn** for future but if the GDA solution is to be used then it should take the **-Nn** option.

Regarding ``grdfill``, I'm afraid we always need to pass a *(x,y,z)* to GDAL. In Mirone I have a fill gaps option that finds the gaps, cuts a sub-region around them, converts it to x,y,z and feeds the interpolation routine. This is the most efficient (but worky) way. Note also that using the GMT machinery to write grids implies making a copy of the z array (need to fetch it from the GDAL owned memory that is deeply hidden in the C++ object).

PaulWessel:
grdfill also has that machinery (it finds and loops over the holes, extends by 2 rows/cols, and for instance calls surface if -As is given - thus it could just as easily call the GDALified function.  We have three altos right now: -Ac fills with a constant value, -An fills with NN, and -As fills via spline.

joa-quim:
Good then. And no need to worry with radius. The default (radius = 0) is to fill all nodes.

joa-quim:
OK, I've swapped **-Ns** and **-Nn** meaning now **-Nn** uses the GDAL code.

PaulWessel:
I merged in master changes.

PaulWessel:
OK, please commit and I can rip out the experimental triangulate stuff.

joa-quim:
It's committed

PaulWessel:
OK, I see. OK for me to rip out that code, then?

joa-quim:
Yes, sure.

PaulWessel:
OK, made my changes and updated a few things.

PaulWessel:
Since your gdal_vector calls GMT_Read_Data I think you need to add support for -i, -h, -q.  All you need is to add them to the OPTIONS and the usage part and the rest is automatic.  Plus man page.

joa-quim:
It already has. So it misses -q

#define THIS_MODULE_OPTIONS "->RVbhi"

joa-quim:
OK, the wrapper program name. I named it ``gmtgdal`` as a lower case version of ``GMTGDAL``. It's true that *gmt-proggys* tend to use table data (but several do not), whilst *grd-proggys* deal with grids, but several deal with grids and do not have *grd* in the name.

So far ``gmtgdal`` deals almost exclusively with rasters (grids and images), so a ``grdgdal`` name would be more appropriated. On the other hand, the *gmt* prefix has own advantages. We can call them as
```
gmt gdal ...
```
and at a certain point I think it would be useful to make it write OGR formats out of our table data files (thus avoiding the need of system calls to ``ogr2ogr``).

All of these to say. I have no strong opinion on a ``gmtgdal`` *vs* ``grdgdal`` *vs* ``someothername``

Opinions?
(@seisman)

joa-quim:
ping @PaulWessel  @seisman

PaulWessel:
A couple of thoughts: I think for the command line it does not matter.  However, I know for the API it is better if a module only returns one type of thing.  We are even considering to break some old modules into several to be simpler, and maybe we will do that with modules that returns different things depending on options.  Going forward with new modules, I think we should try hard to only make them produce one thing.  So I think **grdgdal** sounds like a much better name for now, and if you end up making a **gmtgdal** that deals with tables down the line then that just is another well-named module.

So I think **grdgdal** for the current module would be better.

joa-quim:
Right. Agree with that. A **gmtgdal** might (or not) see the light of day further down the road.

Issue #2596, opened 2020-01-31
Running a lint checker nightly in the CI


PaulWessel:
**Description of the desired feature**

A linter can find issues that the compiler with warnings misses.  I am just playing with [cppcheck](http://cppcheck.sourceforge.net) and it is finding things that I am surprised the compilers is not warning us about already (well, at least not mine).  Just as a single example, it printed 

```
inset.c:287:10: error: Buffer is accessed out of bounds: Bopts [bufferAccessOutOfBounds]
  fgets (Bopts, PATH_MAX, fp);
```

and it does that because Bopts is declared

`char Bopts[GMT_LEN256] = {""};`

Duh!  Might it be possible to run ccpcheck once a night on all *.[ch] files and have it fail if there is output to stderr?  I started this job (still running, so fairly slow and thorough apparently) locally:

```
find . -name '*.[ch]' > /tmp/t.lis
while read file; do cppcheck --force $file 2>> /tmp/log; done < /tmp/t
```

if /tmp/log is not empty then something was reported.  There are different severity levels so this may need to be tuned.

The cppcheck man page offers styled output and extra checks; have not read it yet.

seisman:
Close this issue?

PaulWessel:
Yes, thanks for your help!

seisman:
FYI, `cppcheck --max-configs=1 src/*.c` is fast (~3 min) for local check.

PR #2600, opened 2020-01-31
Run cppcheck in CI jobs

seisman:
For scheduled CI jobs, check all C source codes (it may take one hour);
For PRs, check changed files only (very quick for small changes).

Address #2596.


seisman:
Currently, it runs `cppcheck --error-exitcode=1 --force`. We may add more options later.

Issue #2346, opened 2019-12-26
Rethinking default verbosity and the hierarchical order of verbosities


PaulWessel:
**Description of the desired feature**

Here are the various verbosity levels offered by GMT:
```
-V Change the verbosity level (currently v).
   Choose among 6 levels; each level adds more messages:
     q - Quiet, not even fatal error messages.
     n - Normal verbosity: only error messages.
     c - Also produce compatibility warnings [Default when no -V is used].
     v - Verbose progress messages [Default when -V is used].
     l - Long verbose progress messages.
     d - Debugging messages.
```
Currently, it says default verbosity level is **-Vc** which only prints error messages and compatibility warnings.  I think most users find it a bit odd that if their selection of options produces a case _we_ deem worthy of a warning message (e.g., we must do great circle interpolation despite users wanting ellipsoidal) they do not actually see that message unless they specify **-V**.  Thus, the warning _we think_ they should see is not visible unless they anticipate the problem in the first place.  I think this needs to change.  The current decision derives from a past situation when we had a mix of warnings and informational messages at the same level.  Now, users can set their own verbosity default level via gmt set **GMT_VERBOSE**, but out of the gate that is set to compat.  Finally, the description of **-Vv** is wrong since it does not report progress messages, only warnings.  As you know, all messages a prepended with the scary upper case words ERROR, WARNING, INFORMATION, COMPAT, and DEBUG and we do not want to say `WARNING: Reading file test.txt`, for instance.  Harmless but informative messages need to be under **-Vl**.  The users must _choose_ to get that extra information.

Since we correctly parse any deprecated option syntax it is also annoying to people that we constantly badger them with these "get with the program" messages.  I think users should be able to select such badgering if they are motivated to change their scripts, but not if they don't want to.

**Proposal:**

My proposal has two parts:

1. Slightly adjust the order of the verbosity (and use some new mnemonic letters, yet keep the old letters for backwards compatibility).
2. Let the default verbosity level be **-Vw** (error messages _and_ warnings).

We would then have this situation:
```
-V Change the verbosity level (currently w).
   Choose among 6 levels; each level adds more messages:
     q - Quiet, not even fatal error messages.
     e - Error messages only.  (note: We will also accept -Vn)
     w - Warning and Error messages [Default] (note: we also accept -Vv).
     i - Informational messages (note: Also accept -Vl).
     c - Compatibility warnings are added.
     d - Debugging messages [meant for developers only].
```
The biggest changes here are

[...]

PR #2608, opened 2020-02-01
Set implicit time unit when parsing -T

PaulWessel:
See warning report number 4 under #2346. We rely on **TIME_UNIT** when no unit found but if so we must duplicate that unit back to the structure. The example runs correctly but the warning flagged that we did not copy over the implied time unit.
Review also showed that users could simply append the unit in the -T option:

`gmt math -T2011-02-01T/2011-02-28T/1d TROW 200 ADD = /tmp/shit`

instead of the more awkward

`gmt math -T2011-02-01T/2011-02-28T/1 --TIME_UNIT=d TROW 200 ADD = /tmp/shit`

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -15934,15 +15934,13 @@ unsigned int gmt_parse_array (struct GMT_CTRL *GMT, char option, char *argument,
 		gmt_set_column (GMT, GMT_IN, tcol, GMT_IS_ABSTIME);	/* Set input column type as time */
 		/* Set output column type as time unless -fo has been set */
 		if (!GMT->common.f.active[GMT_OUT]) gmt_set_column (GMT, GMT_OUT, tcol, GMT_IS_ABSTIME);
-		if (has_inc) {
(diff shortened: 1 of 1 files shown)


PaulWessel:
The unit fix fixes warnings from commands like

`gmt earthtide -T2019-01-02T -S`

PaulWessel:
I don't understand what lint is complaining about on this (and similar) lines in gmt_support.c:

```
	/* Determine if there are colorizing options applied, i.e. [:F<rgb>B<rgb>] */
	len = (int)strlen (fill->pattern) - 1;
	for (i = 0, pos = -1; i < len && fill->pattern[i] && pos == -1; i++)
		if (fill->pattern[i] == ':' && i < len && (fill->pattern[i+1] == 'B' || fill->pattern[i+1] == 'F')) pos = i;
```

It says we exceed the pattern array for i = 1000 or something but len surely still stop that?

PaulWessel:
I am merging this one but the lint jobs will continue to fail on us until we understand what is up with that type of message.

PR #2628, opened 2020-02-02
Eliminate *unit* from documentation

PaulWessel:
We were very inconsistent with indicating that units can be appended in the documentation.  Basically, you can append them anywhere you specify distances or dimensions, so better to remove the partial documentation altogether.


doc/rst/source/changes.rst
```diff
@@ -905,7 +905,7 @@ changes to existing syntax will be backwards compatible:
 
 *  :doc:`psconvert` (apart from the name change) has several new features, such as
    reporting dimensions of the plot when **-A** and **-V** are used,
-   scaling the output plots via **-A+s**\ [**m**]\ *width*\ [*unit*][/\ *height*\ [*unit*]],
+   scaling the output plots via **-A+s**\ [**m**]\ *width*[/\ *height*],
```
seisman:
```suggestion
   scaling the output plots via **-A+s**\ [**m**]\ *width*\ [/*height*],
```

doc/rst/source/cookbook/features.rst
```diff
@@ -1882,7 +1882,7 @@ projections you may wish to specify the domain using the lower-left and upper-ri
 instead (similar to how the **-R** option works), by adding **+r**\ .  Some optional modifiers are available:
 
 #. Set inset size.  If you specified a reference point then you must also specify the inset dimensions with the
-   **+w**\ *width*\ [*unit*][/*height*\ [*unit*]], where *height* defaults to *width* if not given.
+   **+w**\ *width*[/*height*], where *height* defaults to *width* if not given.
```
seisman:
```suggestion
   **+w**\ *width*\ [/*height*], where *height* defaults to *width* if not given.
```

PaulWessel:
Sorry, if there are more, how can I (now) find those places without viewing the HTML?

seisman:
There is no easy way to know if they work or not. I usually copy the reST codes and check in http://rst.ninjs.org/.

PaulWessel:
Seems to be specifically having `*[` where the `*` ends an italicized word.

doc/rst/source/mapproject.rst
```diff
@@ -133,7 +133,7 @@ Optional Arguments
 
 .. _-F:
 
-**-F**\ [*unit*]
+**-F**
```
seisman: Maybe we should keep this one.

PaulWessel: Yep, and a few others.  Almost done.

PR #2638, opened 2020-02-03
Clarify array generation

PaulWessel:
Added some simple examples so users can see how it is specified.  Also added reference for grdinterpolate.


doc/rst/source/grdinterpolate.rst
```diff
@@ -1,5 +1,6 @@
 .. index:: ! grdinterpolate
 .. include:: module_core_purpose.rst_
+.. include:: /explain_array.rst_
```
seisman: I have already included this file after `include:: explain_help.rst_` in my previous PR.

PR #2643, opened 2020-02-03
Improve formatting of pssolar output

anbj:


PaulWessel:
@joa-quim, any special reason for the particular formatting of `pssloar -I `output?  Could we split the lon/lat line into twosome with azimuth/elevation, so everything is `<something> = <value>` without all that leading whitespace?

PR #2653, opened 2020-02-04
Let multiple mode message be information

### LINKED ISSUE #2651: histogram warning
```
gmt histogram -T2.5 -W1p -pdf map << EOF
1
2
3
4
5
6
7
8
9
EOF
```

Warning:
```
histogram [WARNING]: 5 multiple modes found
```

Change it to INFO? What does it mean?

PaulWessel:
Various histogram or filtering operations involving mode estimation may report it found more than one mode.  We let this be flagged as INFORMATION.  Closes #2651.

### DIFF
--- a/src/pshistogram.c
+++ b/src/pshistogram.c
@@ -408,7 +408,7 @@ GMT_LOCAL int get_loc_scl (struct GMT_CTRL *GMT, double *data, uint64_t n, doubl
 	/* Get mode */
 
 	gmt_mode (GMT, data, n, j, 0, 0, &n_multiples, &stats[2]);
-	if (n_multiples > 0) GMT_Report (GMT->parent, GMT_MSG_WARNING, "%d multiple modes found\n", n_multiples);
+	if (n_multiples > 0) GMT_Report (GMT->parent, GMT_MSG_INFORMATION "The histogram approximation to the pdf has %d multiple modes\n", n_multiples);
 
 	/* Get MAD for L1 */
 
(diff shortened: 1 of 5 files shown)


src/pshistogram.c
```diff
@@ -408,7 +408,7 @@ GMT_LOCAL int get_loc_scl (struct GMT_CTRL *GMT, double *data, uint64_t n, doubl
 	/* Get mode */
 
 	gmt_mode (GMT, data, n, j, 0, 0, &n_multiples, &stats[2]);
-	if (n_multiples > 0) GMT_Report (GMT->parent, GMT_MSG_WARNING, "%d multiple modes found\n", n_multiples);
+	if (n_multiples > 0) GMT_Report (GMT->parent, GMT_MSG_INFORMATION, "The histogram approximation to the pdf has %d multiple modes\n", n_multiples);
```
seisman: Is "to the pdf" correct?

PaulWessel: Wel, not really.  A pdf should integrate to 1 but our histogram is counts.  Maybe we should say "data distribution" instead?

seisman: I have no idea what do "modes" or "data distribution" mean.

PaulWessel: Lots of data have a single mode (i.e., peak), perhaps well approximated by a Gaussian distribution.  But often, real data may have more than one mode, and a histogram may then have several local peaks and not exhibit a smooth single-peak shape.  The GMT function that tries to determine the statistics (mean, median, mode) will detect this and report how many modes it found.  The pdf (probability density function) is the theoretical distribution of the phenomena that your data are approximating, but the pdf must sum to 1 (100%) while the histograms are counts.  If you divide the histogram counts by the total count then you get an approximation to the pdf.
I think I will just say "the histogram has multiple (%d) modes".

seisman: For my example in #2651, I just want to plot a histogram and don't care about any statistics. Why does GMT calculate and report these modes?

PaulWessel: We are trying to be helpful.  You won't see those messages now unless you add -V.

seisman: OK.

Issue #2646, opened 2020-02-03
GMT can't generate plots if username has whitespaces


seisman:
See the screenshot below:
![image](https://user-images.githubusercontent.com/3974108/73663125-f9399680-466a-11ea-972e-44856963ec49.png)

PaulWessel:
I can see the quoted filename is being passed in the call to psconvert.  However, in psconvert we strip the quotes from the file name via _noquote_name_.  We do add quotes back when calling gs, but I suspect some earlier step is calling _gmt_getdatapath_ (which gives all the "Look for file .." messages and clearly there we already have lost it.
Might be obvious if @joa-quim could replicate this in a dummy account with spaces in the name.

seisman:
@PaulWessel You can also replicate it by creating a directory and setting it to HOME, e.g.
```
mkdir WHITE\ SPACE
cd WHITE\ SPACE
export HOME=`pwd`
gmt basemap -R-10/10/-10/5/0/10 -JX10c/3c -Baf -pdf map -Vd
```

PaulWessel:
No, ran fine for me.  Temp files still written to /Users/pwessel/.gmt/sessions.  Maybe HOME not working for macOS?

joa-quim:
Weird again. From *cmd* it errors but not clear why
```
gmt [DEBUG]: Shared Library # 1 (supplements). Path = C:/programs/GMTdir/gmt_plugins/supplements_w64.dll
gmt [DEBUG]: Shared Library # 2 (mbgmt). Path = C:/programs/GMTdir/gmt_plugins/mbgmt_w64.dll
begin [INFORMATION]: Creating a workflow directory "C:/v/spa co"/.gmt/sessions/gmt6.40852
gmt_mkdir (intermediate) error: Invalid argument
begin [ERROR]: Unable to create a workflow directory : "C:/v/spa co"/.gmt/sessions/gmt6.40852
```
trying to debug it from Julia, the firs step worked. Only the ``psconvert`` errored for the same reason.
```
julia> gmtend()
psconvert [ERROR]: Error for input file: No such file ('c:/v/spa)
```

joa-quim:
The problem is that ``gmt_mkdir()`` (and probably others) aren't able to deal with paths that start with a ``"`` and ``API->session_dir`` has them. Tried to get rid of it but id I do it in API->session_dir in ``gmt_manage_workflow()`` then I have a crash at the end. 
[...]

PR #2656, opened 2020-02-04
Address complications for filenames with spaces in psconvert and show

PaulWessel:
See #2646 for background.  The solution involved several steps:

- The filename argument to psconvert **-F** needed to be de-quoted.  This is necessary because when we build the gs command we always surround the final output name with quotes.
- _GMT_Create_Options_ needs to consider both single and double quotes when converting a single text string commands that contain filenames in quotes to linked options.
- The gmt end **show** code also needed to be mindful of the presence of quotes in file names.

With these changes, GMT handles filenames with spaces provided given in quotes; these two commands work equally well for me (macOS):

```
gmt basemap -R-10/10/-10/5/0/10 -JX10c/3c -Baf -pdf 'My map'
gmt basemap -R-10/10/-10/5/0/10 -JX10c/3c -Baf -pdf "My map"
```

Hopefully this addresses all of #2646 but that needs to be determined.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -12951,7 +12951,10 @@ GMT_LOCAL int set_modern_mode_if_oneliner (struct GMTAPI_CTRL *API, struct GMT_O
 			}
 		}
 		if (opt->next && opt->next->option == GMT_OPT_INFILE) {	/* Found a -ext[,ext,ext,...] <prefix> pair */
-			snprintf (session, GMT_LEN128, "%s %s", opt->next->arg, figure);
+			if (strchr (opt->next->arg, ' '))
+				snprintf (session, GMT_LEN128, "\'%s\' %s", opt->next->arg, figure);
+			else
+				snprintf (session, GMT_LEN128, "%s %s", opt->next->arg, figure);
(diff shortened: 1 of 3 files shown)


PaulWessel:
The linter fails on lines like this:
src/gmt_init.c:10799]: (error) sprintf format string requires 2 parameters but only 1 is given
since it cannot tell that we are in fact passing 2 characters.  Perhaps we can rewrite the GMT_def macro so that we can have that comma that is missing.

PaulWessel:
Since the _GMT_def_ macro is more like a text substitution that provides a multiply symbol (*), the multiplier to the argument actually given in _sprintf_, then the comma, and then a unit, perhaps if we made two macros:
_GMT_def_scale_ and _GMT_def_unit_ and do

`snprintf (value, GMT_LEN256, "%g%c", GMT->current.setting.map_annot_offset[GMT_PRIMARY] * GMT_def_scale(GMTCASE_MAP_ANNOT_OFFSET_PRIMARY), GMT_def_unit(GMTCASE_MAP_ANNOT_OFFSET_PRIMARY));`

then presumably the linter can figure this out?

seisman:
I think so.

PR #2665, opened 2020-02-04
Handle quotes in API->session_dir

PaulWessel:
Avoid quotes in name when used to build a path.  Hopefully helping with #2646, @joa-quim.

### DIFF
--- a/src/common_string.c
+++ b/src/common_string.c
@@ -26,6 +26,7 @@
  *  gmt_chop                Chops off any CR or LF at end of string
  *  gmt_chop_ext            Chops off the trailing .xxx (file extension)
  *  gmt_get_ext             Returns a pointer to the tailing .xxx (file extension)
+ *  gmt_strdup_noquote		Duplicates a string but removes any surrounding single or double quotes
  *  gmt_strstrip            Strip leading and trailing whitespace from string
  *  gmt_strlshift           Left shift a string by n characters
  *  gmt_strrepc             Replaces all occurrences of a char in the string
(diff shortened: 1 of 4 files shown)


joa-quim:
Nope, same thing

```
begin [INFORMATION]: Creating a workflow directory "C:/v/spa co"/.gmt/sessions/gmt6.40852
gmt_mkdir (intermediate) error: Invalid argument
begin [ERROR]: Unable to create a workflow directory : "C:/v/spa co"/.gmt/sessions/gmt6.40852
```

PaulWessel:
wow, so this actually fails:

```
if (strchr ("\'\"", API->session_dir[0]) && strchr ("\'\"", API->session_dir[end])) start++, API->session_dir[end] = '\0';	/* Remove quotes */
snprintf (dir, PATH_MAX, "%s/gmt%d.%s", &API->session_dir[start], GMT_MAJOR_VERSION, API->session_name);
API->gwf_dir = strdup (dir);
if (start) API->session_dir[end] = API->session_dir[0];	/* Restore quotes */
```

How can that be - what am I missing.  It says if your session dir starts with a quote and ands with a quote, skip them when making the workflow directory.

joa-quim:
I don't get this ``if (strchr ("\'\"", API->session_dir[0])``. *strchr* searches for a char but you are seeking two: *\'* and *\"*
We have a ``gmt_strrep()`` but I couldn't make it work to replace a double quote by a nothing.
``gmt_strrep(str, "\"", "")`` replaced ``"`` by the char *s*

PaulWessel:
The strchar is looking for a single char (API->session_dir[0]) in a string of two characters (" and ').

joa-quim:
It fails because ``API->session_dir`` is already
"C:/v/spa co"/.gmt/sessions

Like I mentioned before. We should not store strings with double or single quotes. Only add them when passing to 3rd party programs.

joa-quim:
BTW, it was here when I tried ``gmt_strrep()`` and didn't solve this issue.

PaulWessel:
OK, let me work on this.

PaulWessel:
See how this one goes.  I promoted your noquote function from psconvert to common_strings and called it gmt_strdup_noquote.  It should get further and then we can see where else we have trouble.

joa-quim:
YES, now it finally works.

PaulWessel:
Please run tests etc before you approve the PR.

joa-quim:
These are failing and didn't use to, but are unrelated
```
ex47/ex47.sh
psscale/hardsoft.sh
```

PaulWessel:
I get some failures sometimes and if I just rerun those with

ctest -R ex47

then they work OK.  Is that the case for you? If so please approve the PR, else let me know what fails.

PR #2666, opened 2020-02-04
Give suggestion on what to do when "Subplot information file already exists"

joa-quim:


joa-quim:
Right about the period but didn't add Python because it doesn't work that way. I mean, one doesn't run *gmt cmd ....* in it. But added anyway.

Issue #2587, opened 2020-01-29
Simplify the CMake configuration file


PaulWessel:
**Description of the desired feature**

Right now the cmake/ConfigUserTemplate.cmake has 311 lines. The vast majority of these are not to be changed by the user.   It is easy to get lost in this large file.  Also, once you have made your copy to ConfigUser.cmake and we change the template, most users will not remember to update their config file.  Might it be better to use CMake's [include statement](https://cmake.org/cmake/help/latest/command/include.html) and instead isolate the few parameters that users may change into a ConfigGMTTemplate.cmake file that users can copy to ConfigGMT.cmake which is then included by our ConfigUser.cmake (renamed from ConfigUserTemplate.cmake)?  Maybe we could use CMake to tell [check if a file exists](https://cmake.org/cmake/help/v3.5/command/if.html) and then include the users GMT configuration file from their home directory instead of creating things in cmake?

seisman:
What about splitting the parameters into two files: ConfigUserTemplate.cmake and ConfigUserAdvancedTemplate.cmake? 

Following parameters can be put in the ConfigUserTemplate.cmake file:

- CMAKE_INSTALL_PREFIX
- GMT_INSTALL_MODULE_LINKS
- GSHHG_ROOT
- COPY_GSHHG
- DCW_ROOT
- COPY_DCW

PaulWessel:
Yes, that is a good idea.  There are several more advanced parameters that could go there that is separate from the scripting part of the file.  I just think it is easier to digest a parameter file(s) instead of having the assignments embedded in a CMake script.

PaulWessel:
Not sure if the COPY_GSHHG and COPY_DCW are necessary at that level.  One users have downloaded these they should not need to duplicate them - that is for our bundle, etc?

PR #2667, opened 2020-02-05
Simplify CMake Configuration

PaulWessel:

See #2587 for background.  This PR splits out a tiny file (GMTSettingsTemplate.cmake) from the long ConfigUserTemplate.cmake with just the four most important configuration settings:

`CMAKE_INSTALL_PREFIX, GMT_INSTALL_MODULE_LINKS, GSHHG_ROOT, DCW_ROOT`

They still have defaults and are all commented out, but the user can now make a copy from that template and just edit a small file with just a few lines rather than the 300+ line ConfigUser.cmake file.  That file still has advanced tweakings that is only useful for gurus and package maintainers.  I updated BUILDING.md as well.

The new scheme worked for me.  I have some questions that we can sort out during the PR review:

1. Do we want to move the detailed Windows instructions below to the more advanced file instead:
```
set (CMAKE_C_FLAGS "/D_CRT_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_DEPRECATE ${CMAKE_C_FLAGS}")
set (CMAKE_C_FLAGS "/D_CRT_NONSTDC_NO_DEPRECATE /D_SCL_SECURE_NO_DEPRECATE ${CMAKE_C_FLAGS}")
```
It seems too advanced for the common Windows man.
2. Do you want another name than GMTSettingsTemplate? It looks like there has to be a file named ConfigUser.cmake so I thought GMTSettings.cmake is OK, but it could be just GMT.cmake.
3. Anything else to consider?  I dont like the ConfigUser.cmake file because it mixes assignments and programming.  The parameters should be decoupled from the cmake script I think.  This is a first step, but ideally there could be a ConfigAdvancedTemplate.cmake as well.


cmake/ConfigUserTemplate.cmake
```diff
@@ -27,19 +35,16 @@
...
-# ============================================================================
-# Basic setup begins here.  All settings are optional.  In most cases, setting
-# CMAKE_INSTALL_PREFIX should be all you need to do in order to build GMT with
-# reasonable defaults enabled.  Note: If you need to specify directory names
-# with spaces (e.g., on Windows) then you must put them in quotes.
-# ============================================================================
-
-# Installation path (usually defaults to /usr/local) [auto]:
-#set (CMAKE_INSTALL_PREFIX "prefix_path")
+if (EXISTS ${GMT_SOURCE_DIR}/cmake/GMTSettings.cmake)
```
seisman: The if statement should be moved to `cmake/modules/ConfigCMake.cmake`.

PaulWessel:
Why there? The contents came from ConfigUser.cmake, no?

seisman:
> Why there? The contents came from ConfigUser.cmake, no?

If someone copies `GMTSettingsTemplate.cmake` to `GMTSettings.cmake`, but doesn't copy `ConfigUserTemplate.cmake`, then the statements in `ConfigUserTemplate.cmake` have no effects.

PaulWessel:
OK, I guess I am slowly realizing that all the lines in ConfigUserTemplate.cmake are commented out.  So how is ConfigUser.cmake called? Is it a required name that Cmake looks for, and read if found?

seisman:
It's included at the very top of `cmake/modules/ConfigCMake.cmake`.

PaulWessel:
Makes total sense - thanks.  Only remaining question from my initial thinking is we we should look in the users home dir **if** there is no GMTSettings.cmake in the cmake dir.

PaulWessel:
i.e., adding elseif branch:

```
if (EXISTS ${GMT_SOURCE_DIR}/cmake/GMTSettings.cmake)
	include (${GMT_SOURCE_DIR}/cmake/GMTSettings.cmake)
elseif (EXISTS $ENV{HOME}/GMTSettings.cmake)
	include ("$ENV{HOME}/GMTSettings.cmake")
endif ()
```

seisman:
> This PR splits out a tiny file (GMTSettingsTemplate.cmake) from the long ConfigUserTemplate.cmake with just the four most important configuration settings:
> 
> `CMAKE_INSTALL_PREFIX, GMT_INSTALL_MODULE_LINKS, GSHHG_ROOT, DCW_ROOT`
> 

I still want to include `COPY_DCW` and `COPY_GSHHG` to this template. The reason is, I think most users just download the three tarballs (GMT, dcw and gshhg) to a temporary directory, extract the tarballs, and the follow the building instruction to build and install them. After that, it's likely they will delete these tarballs and directories, and then find that GMT can't find GSHHG data anymore, because `COPY_GSHHG` is false by default, and GSHHG data are deleted. 

> 1. Do we want to move the detailed Windows instructions below to the more advanced file instead:
> 
> ```
> set (CMAKE_C_FLAGS "/D_CRT_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_DEPRECATE ${CMAKE_C_FLAGS}")
> set (CMAKE_C_FLAGS "/D_CRT_NONSTDC_NO_DEPRECATE /D_SCL_SECURE_NO_DEPRECATE ${CMAKE_C_FLAGS}")
> ```

Yes for me, maybe in the "Debugging" section, although I have no idea what these C flags are.

> 2. Do you want another name than GMTSettingsTemplate? It looks like there has to be a file named ConfigUser.cmake so I thought GMTSettings.cmake is OK, but it could be just GMT.cmake.

Maybe `ConfigUserTemplate.cmake` for the tiny template and `ConfigUserAdvancedTemplate.cmake` for the longer one?

> Only remaining question from my initial thinking is we we should look in the users home dir **if** there is no GMTSettings.cmake in the cmake dir. The reason is if you are installing from a tarball then each time to extract you need to duplicate those parameter files again. Having one in ~ would let you just have one.

I don't like it. If we really need a global config file, it should be in `~/.gmt/`.

PaulWessel:
> I still want to include `COPY_DCW` and `COPY_GSHHG` to this template. The reason is, I think most users just download the three tarballs (GMT, dcw and gshhg) to a temporary directory, extract the tarballs, and the follow the building instruction to build and install them. After that, it's likely they will delete these tarballs and directories, and then find that GMT can't find GSHHG data anymore, because `COPY_GSHHG` is false by default, and GSHHG data are deleted.

Those are very good points - I agree and have updated that file.

> > 1. Do we want to move the detailed Windows instructions below to the more advanced file instead:
> 
> Yes for me, maybe in the "Debugging" section, although I have no idea what these C flags are.
> 
Lets see what @joa-quim has to say.  They seem advanced to me.

> > 1. Do you want another name than GMTSettingsTemplate? It looks like there has to be a file named ConfigUser.cmake so I thought GMTSettings.cmake is OK, but it could be just GMT.cmake.
> 
> Maybe `ConfigUserTemplate.cmake` for the tiny template and `ConfigUserAdvancedTemplate.cmake` for the longer one?
> 
That is good, as it retains the original name for the file should edit.

> > Only remaining question from my initial thinking is we we should look in the users home dir **if** there is no GMTSettings.cmake in the cmake dir. The reason is if you are installing from a tarball then each time to extract you need to duplicate those parameter files again. Having one in ~ would let you just have one.
> 
> I don't like it. If we really need a global config file, it should be in `~/.gmt/`.

That is fine too.  I guess what I dont like about having things inside cmake is that if you wipe the directory and start over then your config files are lost.  Having them outside the working copy of the repo seems handy. And, my suggestions would only kick in if there are no file in cmake that is found first.

anbj:
> > I still want to include `COPY_DCW` and `COPY_GSHHG` to this template. The reason is, I think most users just download the three tarballs (GMT, dcw and gshhg) to a temporary directory, extract the tarballs, and the follow the building instruction to build and install them. After that, it's likely they will delete these tarballs and directories, and then find that GMT can't find GSHHG data anymore, because `COPY_GSHHG` is false by default, and GSHHG data are deleted.
> 
> Those are very good points - I agree and have updated that file.
> 

Why not set `COPY_DCW` and `COPY_GSHHG` to true by default?

PaulWessel:
By default would mean we set those two items to TRUE in the modules/COnfigCMake.cmake, no?  And then set things to FALSE in the commented lines in ConfigUser.cmake?  That is the only way it because the default.
Or do you just want to remove the comments and make setting them to TRUE the default in the template?

joa-quim:
Leave those Windows CFLAGS where they are. They prevent a flood of warnings because the MS considered safe versions of functions are not being used.

seisman:
> By default would mean we set those two items to TRUE in the modules/COnfigCMake.cmake, no? And then set things to FALSE in the commented lines in ConfigUser.cmake? That is the only way it because the default.

I think it's better.

> Leave those Windows CFLAGS where they are. They prevent a flood of warnings because the MS considered safe versions of functions are not being used.

These C_FLAGS are only mentioned in the BUILDING.md. I think we should  add it to the ConfigUserAdvancedTemplate.cmake (of course, commented by default).

PaulWessel:
OK, I have made the GSHHG and DCW copy TRUE by default - users can override.  Also, I have moved the specific warning-suppressions for Windows to the Advanced template, commented out.

cmake/modules/ConfigCMake.cmake
```diff
@@ -17,7 +17,15 @@
 #
 include ("${CMAKE_SOURCE_DIR}/cmake/ConfigDefault.cmake")
 
-# A "ConfigUser.cmake" in the source tree overrides the defaults.
+# A "GMTSettings.cmake" in the source tree overrides the basic defaults.
+if (EXISTS ${GMT_SOURCE_DIR}/cmake/GMTSettings.cmake)
+	include (${GMT_SOURCE_DIR}/cmake/GMTSettings.cmake)
+else ()	# Default is to copy the support data into the build shared dir
+	set (COPY_GSHHG TRUE)
+	set (COPY_DCW TRUE)
```
seisman: Move these two lines to `cmake/ConfigDefault.cmake`? It seems most default values are set in that file.

cmake/modules/ConfigCMake.cmake
```diff
@@ -17,7 +17,15 @@
 #
 include ("${CMAKE_SOURCE_DIR}/cmake/ConfigDefault.cmake")
 
-# A "ConfigUser.cmake" in the source tree overrides the defaults.
+# A "GMTSettings.cmake" in the source tree overrides the basic defaults.
+if (EXISTS ${GMT_SOURCE_DIR}/cmake/GMTSettings.cmake)
```
seisman: Remove this include statement. There is no `cmake/GMTSettings.cmake` now.

cmake/modules/ConfigCMake.cmake
```diff
@@ -17,7 +17,15 @@
...
+	include (${GMT_SOURCE_DIR}/cmake/GMTSettings.cmake)
+else ()	# Default is to copy the support data into the build shared dir
+	set (COPY_GSHHG TRUE)
+	set (COPY_DCW TRUE)
+endif ()
+
+# A "ConfigUser.cmake" in the source tree overrides the advanced defaults.
```
seisman: Include "ConfigUser.cmake" first, then "ConfigUserAdvanced.cmake".

.gitignore
```diff
@@ -3,6 +3,7 @@ build
 ?build-mp
 .*~
 cmake/ConfigUser.cmake
+cmake/GMTSettings.cmake
```
seisman:
```suggestion
cmake/ConfigUserAdvanced.cmake
```

BUILDING.md
```diff
@@ -296,25 +296,29 @@ Extract the files and put them in a separate directory (need not be where you ev
 
 GMT can be built on any platform supported by CMake. CMake is a cross-platform,
 open-source system for managing the build process. The building process is
-controlled by two configuration files in the `cmake` directory:
+controlled by three configuration files in the `cmake` directory:
```
seisman:
```suggestion
controlled by two configuration files in the `cmake` directory:
```
Only two configuration files, right?

joa-quim:
I don't understand why you call them advanced. They aren't. Simply protect from a flood of warning messages.
I don't agree  with this change.

BUILDING.md
PaulWessel: Well, as written up (From Florian), he counts the CMakeDefaults as one of them.  He then says NOT to edit that one.  So that leaves 2 more, for a total or 3.  We could only two for users to make changes to or something.

PaulWessel:
> I don't understand why you call them advanced. They aren't. Simply protect from a flood of warning messages.
> I don't agree with this change.

Well, then we uncomment them by default - I don't really care.  They still dont belong in the basic settings and should stay in the advanced settings.  Having those two very long and scary statements in the BUILDING.md and in the basic configuration setting is a no no.

joa-quim:
if anyone building GMT on Windows get scared with such a thing ... better to buy a dog.

seisman:
For those two MSVC flags, we probably can make them the default C_FLAGS, just like what we're doing here:
https://github.com/GenericMappingTools/gmt/blob/4eeea419d7164e1a16310046146d3f3dcf2bc8a0/cmake/modules/ConfigCMake.cmake#L206-L210

BUILDING.md
seisman: OK with 3.

joa-quim:
Yes, that seems more reasonable.

PaulWessel:
yes, fine with me, please make the change (I have meetings coming up).

seisman:
@PaulWessel @joa-quim I've made some changes to this PR. Please see if you're OK with it.

PaulWessel:
Looks good to me.

seisman:
@joa-quim Is it OK to approve and merge?

Issue #2466, opened 2020-01-19
Four failing tests on macOS CI


seisman:
The macOS CI reports four failing tests recently, although they all work well for me locally. The four tests are:

```
	253 - geodesy/case_largeR_noW.sh (Failed)
	254 - geodesy/case_largeR_withW.sh (Failed)
	255 - geodesy/case_smallR_noW.sh (Failed)
	256 - geodesy/case_smallR_withW.sh (Failed)
```

However, the CI doesn't give any useful information for debugging.

PaulWessel:
Hm, wonder about Ax= b and eigenvalues and lapack issues?

PR #2668, opened 2020-02-05
Fix four failing geodesy tests

seisman:
Address #2466.

Four geodesy tests fail recently. I can't reproduce these failures locally, so I have to rely on the CI for debugging.

It turns out the command `data=$(gmt which -G @wus_gps_final_crowell.txt -Vd)` fails. I added some debugging statements (see the changed files for details) and the CI debugging messages show that the command may crash at line 217 (i.e. the `strncpy` function).

https://github.com/GenericMappingTools/gmt/blob/13ff05a0bacb5b94e66a4a2351ab2d9a20cc31e2/src/gmtwhich.c#L213-L218

Please wait ~30 minutes for the CI reports for this PR.


PaulWessel:
OK, we can see what it says. If the first if-test is true then _L_ is pointing to the last / in the path. If there is nothing after that slash then `L[0] == '/'` and `L[1] = '\0'`.  Then the question is should _strncpy_ fail if we give a source string that is ""?
Anyway, I dont think we have changed anything related to this code in a while, but we did change the quote business should the paths involved on the CI have spaces in them (which I doubt).

seisman:
The CI reports:
```
gmtwhich [INFORMATION]: Writing Data Table to Standard Output stream
gmtwhich [DEBUG]: GMT_Begin_IO: Output resource access is now enabled [record-by-record]
gmtwhich [DEBUG]: Local file /Users/runner/.gmt/server/gmt_hash_server.txt found
gmtwhich [DEBUG]: File /Users/runner/.gmt/server/gmt_hash_server.txt less than 24 hours old, refresh is premature.
gmtwhich [DEBUG]: Look for file wus_gps_final_crowell.txt in /Users/runner/.gmt
gmtwhich [DEBUG]: Look for file wus_gps_final_crowell.txt in /Users/runner/runners/2.164.7/work/1/s/test/geodesy
gmtwhich [DEBUG]: Look for file wus_gps_final_crowell.txt in /Users/runner/.gmt/cache
gmtwhich [DEBUG]: Found file /Users/runner/.gmt/cache/wus_gps_final_crowell.txt
gmtwhich [DEBUG]: Look for file wus_gps_final_crowell.txt in /Users/runner/.gmt
gmtwhich [DEBUG]: Look for file wus_gps_final_crowell.txt in /Users/runner/runners/2.164.7/work/1/s/test/geodesy
gmtwhich [DEBUG]: Look for file wus_gps_final_crowell.txt in /Users/runner/.gmt/cache
gmtwhich [DEBUG]: Found file /Users/runner/.gmt/cache/wus_gps_final_crowell.txt
gmtwhich [DEBUG]: Before gmt_rename_file /Users/runner/.gmt/cache/wus_gps_final_crowell.txt wus_gps_final_crowell.txt
gmtwhich [DEBUG]: Copying /Users/runner/.gmt/cache/wus_gps_final_crowell.txt -> wus_gps_final_crowell.txt
gmtwhich [DEBUG]: After gmt_rename_file /Users/runner/.gmt/cache/wus_gps_final_crowell.txt wus_gps_final_crowell.txt
ERROR: gmtest:54
memtrack errors: 0
exit status: 1
ERROR: case_largeR_noW.sh:9
memtrack errors: 0
exit status: 1
```

PaulWessel:
So no crash but failed test at that line you mentioned, but running fine locally.  I would think it is an issue in the CI only since no failures otherwise?

seisman:
I added a debug statement after the `strncpy` function. The debug message was not shown in the CI reports. So I think the `strncpy` function crashes due to some unknown reasons.

Actually, the failures are intermittent. They always fail in the full tests, but pass if the tests are run via `ctest -R geodesy`.

PaulWessel:
Since assuming that strncpy has no bugs is a safe bet, the most likely problem is that we are somehow corrupting the pointer passed in by _earlier_ actions. I will stare at it again.

PaulWessel:
Looking at [this](https://stackoverflow.com/questions/33276068/using-strncpy-where-destination-contains-the-source) I think our problem is the last entry there: Both our source and destination is basically path, so the answer is to use memmove instead.  I am trying

memmove (path, &L[1], strlen (&L[1])+1);	/* Report the file in the local directory now */

since the strlen should return the length of the filename and the +1 should add the terminating zero byte.  Lets see what happens in the CI.

seisman:
I believe your fix works.

PaulWessel:
Good, we probably should review all strncpy and strcpy lines for similar.  This actually reminds me of a strcpy crash which was like that and the Apple C lib changed and made it a SEGV violation instead of just undefined behavior.  Every gmt command SEGV'ed immediately.  I think it was GMT 4.4.x or something and we had to update very quickly.

seisman:
I think this PR is OK to merge. Two failing tests are unrelated.
```
	375 - grdgdal/gdal_nn.sh (Failed)
	500 - nearneighbor/nat_nn.sh (Failed)
```

PaulWessel:
We have lots of strncpy calls.  Just scanning through these all look suspicious, but there may be more since I did not check if src or dst is a pointer derived from the other (like the gmtwhich case):

./gmt_io.c:     strncpy (&scopy[j+1], &p[1], GMT_LEN64-1);
```
./gmt_io.c:             strncpy (GMT->current.io.curr_trailing_text, &GMT->current.io.curr_text[start_of_text], GMT_BUFSIZ-1);
./mgd77/mgd77.c:                                for (rec = 1; rec < count[0]; rec++) strncpy (&text[rec*count[1]], text, count[1]);     /* Replicate one string */
./mgd77/mgd77.c:                                        if (rec_in > rec) strncpy (&text[rec*count[1]], &text[rec_in*count[1]], count[1]);      /* Must shuffle text records */
```

A bit odd that the compiler does not point any of this out.  Not sure what to do. We could write a strncpy function that determines from src how many chars and then checks if there is pointer overlap with dst...

PR #2671, opened 2020-02-05
Compute require colatitude when setting a pole a point for small circle

PaulWessel:
project can draw a small circle, but if you specify the projection via a pole by **-T** and a point by **-C**, it is complicated to draw the small circle through C because you do not know its colatitude needed in **-G** without computing it first separately (e.g., with mapproject).  This PR adds a new modifier **+c** to **-G** that will automatically supply the required colatitude to active the desired small circle.

### DIFF
--- a/doc/rst/source/project.rst
+++ b/doc/rst/source/project.rst
@@ -14,7 +14,7 @@ Synopsis
 
 **gmt project** [ *table* ] |-C|\ *cx*/*cy* [ |-A|\ *azimuth* ]
 [ |-E|\ *bx*/*by* ] [ |-F|\ *flags* ]
-[ |-G|\ *dist*\ [/*colat*][**+h**] ]
+[ |-G|\ *dist*\ [/*colat*][**+c**\|\ h**] ]
 [ |-L|\ [**w**\|\ *l\_min*/*l\_max*] ]
 [ |-N| ] [ |-Q| ] [ |-S| ]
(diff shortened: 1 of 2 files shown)


doc/rst/source/project.rst
```diff
@@ -147,11 +147,13 @@ Optional Arguments
 
 .. _-G:
 
-**-G**\ *dist*\ [/*colat*][**+h**]
+**-G**\ *dist*\ [/*colat*][**+c**\|\ h**]
```
seisman:
```suggestion
**-G**\ *dist*\ [/*colat*][**+c**\|\ **h**]
```

doc/rst/source/project.rst
```diff
@@ -147,11 +147,13 @@ Optional Arguments
...
     Generate mode. No input is read. Create (*r*, *s*, *p*) output
     points every *dist* units of *p*. See **-Q** option. Alternatively,
     append **/**\ *colat* for a small circle instead [Default is a
-    colatitude of 90, i.e., a great circle]. Use **-C** and **-E** to
+    colatitude of 90, i.e., a great circle]. If setting a pole with **-T**
+    and you want to small circle to go through *cx*/*cy*, append **+c** to
```
seisman: "you want to small circle to go through" doesn't read well.

PaulWessel: ops, fixed.

PR #2698, opened 2020-02-09
Only run cppcheck in scheduled jobs

seisman:
cppcheck is useful but slow.
This PR remove the cppcheck stuff from PR jobs. cppcheck will be run
nightly.


PaulWessel:
Maybe skip this one as well which is not our code but generates lots of these:
src/mergesort.c:276:10: error: Pointer addition with NULL

PR #2699, opened 2020-02-09
Let cppcheck ignore src/mergesort.c

seisman:
See https://github.com/GenericMappingTools/gmt/pull/2698#issuecomment-583792581


seisman:
We now move cppcheck to scheduled jobs. It checks all *.c and *.h files no matter they change or not.

PR #2701, opened 2020-02-09
Add acknowledgment of software GMT relies on

PaulWessel:
The usual list is now added to the end of the README in a new  acknowledgment section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,3 +88,21 @@ Copyright (c) 1991-2020 by [the GMT Team](AUTHORS.md).
 GMT is released under the
 [GNU Lesser General Public License](http://www.gnu.org/licenses/lgpl.html)
 version 3 or any later version. See [LICENSE.TXT](LICENSE.TXT) for full details.
+
+## Acknowledgment
+
+GMT relies on several other Open Source software libraries and programs for its
+operation.  We gratefullly acknowledge the importance to GMT of these products.
+GMT may be linked with these libraries (some are optional):
+[Geospatial Data Abstraction Library (GDAL)](https://gdal.org),
+[Network Common Data Form (netCDF)](https://www.unidata.ucar.edu/software/netcdf/),
+[Perl Compatible Regular Expressions (PCRE)](https://www.pcre.org), 
+[Fastest Fourier Transform in the West (FFTW)](http://www.fftw.org),
+[Linear Algebra Package (LAPACK)](http://www.netlib.org/lapack/),
+[Basic Linear Algebra Subprograms (BLAS)](http://www.netlib.org/blas/), and
+[ZLIB](https://www.zlib.net). GMT may call these executables:
+GDAL (ogr2ogr, gdal_info), [Ghostscript](https://www.ghostscript.com),
+[FFmpeg(https://www.ffmpeg.org)],
+[dxg-open](https://www.freedesktop.org/wiki/Software/xdg-utils/), and
+[GraphicsMagick](http://www.graphicsmagick.org).
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -88,3 +88,21 @@ Copyright (c) 1991-2020 by [the GMT Team](AUTHORS.md).
...
+GMT may be linked with these libraries (some are optional):
+[Geospatial Data Abstraction Library (GDAL)](https://gdal.org),
+[Network Common Data Form (netCDF)](https://www.unidata.ucar.edu/software/netcdf/),
+[Perl Compatible Regular Expressions (PCRE)](https://www.pcre.org), 
+[Fastest Fourier Transform in the West (FFTW)](http://www.fftw.org),
+[Linear Algebra Package (LAPACK)](http://www.netlib.org/lapack/),
+[Basic Linear Algebra Subprograms (BLAS)](http://www.netlib.org/blas/), and
+[ZLIB](https://www.zlib.net). GMT may call these executables:
+GDAL (ogr2ogr, gdal_info), [Ghostscript](https://www.ghostscript.com),
+[FFmpeg(https://www.ffmpeg.org)],
```
seisman:
```suggestion
[FFmpeg](https://www.ffmpeg.org),
```

README.md
```diff
@@ -88,3 +88,21 @@ Copyright (c) 1991-2020 by [the GMT Team](AUTHORS.md).
...
+[dxg-open](https://www.freedesktop.org/wiki/Software/xdg-utils/), and
```
seisman:
```suggestion
[xdg-open](https://www.freedesktop.org/wiki/Software/xdg-utils/), and
```

README.md
```diff
@@ -88,3 +88,21 @@ Copyright (c) 1991-2020 by [the GMT Team](AUTHORS.md).
...
+
+## Acknowledgment
+
+GMT relies on several other Open Source software libraries and programs for its
+operation.  We gratefully acknowledge the importance to GMT of these products.
+GMT may be linked with these libraries (some are optional):
+[Geospatial Data Abstraction Library (GDAL)](https://gdal.org),
+[Network Common Data Form (netCDF)](https://www.unidata.ucar.edu/software/netcdf/),
+[Perl Compatible Regular Expressions (PCRE)](https://www.pcre.org), 
+[Fastest Fourier Transform in the West (FFTW)](http://www.fftw.org),
```
joa-quim: We cannot say that we rely on FFTW (we don't). FFTW is GPL, we don't rely on GPL libs.

PaulWessel: Agree that we dont rely on it, we may use it.  I dont understand the point about GPL and that we dont rely on GPL libs.  What do you mean?

joa-quim: If we say we rely on a GPL lib, GMT has to be GPL.

PaulWessel: what is the approved verb then? use?

PaulWessel: Please help phrase this so we can get it done.

joa-quim: I think it will do if we add an [optional] at the end of all of them but netCDF.

PaulWessel: OK, I marked most as optional.  OK with you?

PaulWessel:
See if you have comments on this, @joa-quim and @seisman.  The GMT splash screen now says something like this (my build):

```
	GMT - The Generic Mapping Tools, Version 6.1.0_3630ae8-dirty_2020.02.09 [64-bit] [8 cores]
	[Linked with netCDF, GDAL, PCRE, FFTW, LAPACK, ZLIB]
	(c) 1991-2020 The GMT Team (https://www.generic-mapping-tools.org/team.html).
```

README.md
joa-quim: > OK, I marked most as optional. OK with you?

It's not me. I don't give a shit on it. It's the GNU zealots that I'm concerned with.

PaulWessel: It seems the standard practice is to ignore this stuff until a zealots starts attacking.  Then one can add both GPL and LPGL licenses to GMT and users can select which one, apparently.
Seems OK to you @seisman?

seisman: OK to me.

PR #2715, opened 2020-02-12
Tune docs and add a test.

joa-quim:
The test needs an orig PS


seisman:
@joa-quim You need to run `gmt_make_module_src.sh` again to fix the CI failure.

joa-quim:
I don't understand what the diff is complaining about. The lines are exactly equal.
The ``gmt_make_module_src.sh`` doesn't work on Windows. it has several ``/tmp/...`` and gives lots of errors like
```
: Invalid argument
grep: movie: No such file or directory
```

seisman:
Since you changed "gmtgdal" to "grdgdal", the order of the modules changes.

seisman:
> The `gmt_make_module_src.sh` doesn't work on Windows. it has several `/tmp/...` and gives lots of errors like
> 
> ```
> : Invalid argument
> grep: movie: No such file or directory
> ```

Perhaps we should use `$TMPDIR` instead of `/tmp`.

PaulWessel:
Those scripts certainly were not written with Windows in mind. Maybe you can post the full output of running them with bash -xv ?

PR #2719, opened 2020-02-12
Replace /tmp with $TMPDIR

seisman:
`$TMPDIR` may work well on Windows.


seisman:
@joa-quim Please check if it works for you.

joa-quim:
Nope. No difference. Same errors.

PaulWessel:
What are the errors?

joa-quim:
Those that I reported in #2715
```
: Invalid argument (many of these)
grep: movie: No such file or directory
```

seisman:
This PR may still not work well on Windows, but at least it's a good improvement for Linux/macOS. Merge it?

PaulWessel:
Yes, clearly better.  If @joa-quim can run the full bash -xv on the script we can see what is going on better - the errors above do not help me.

PaulWessel:
So please merge.

PR #2720, opened 2020-02-12
Tracking option to grdtrack

PaulWessel:

I have added a new option **-F**[**+n**][**+z**_z0_] to grdtrack to be used in conjunction with **-C** and a single input grid.  It will examine all the cross-profiles and determine three points on each profile: The center peak and the left/right rise points (when first exceeding _z0_).  This is useful for tracing the center ridge line of an elongated feature and possibly its width.  Below you see a smoothed version of the Hawaii-Emperor Chain bathymetry, a rough predicted trail from a plate motion model (red) used to create the crossings (yellow lines), and the white center dots identified by **-F**.  The line connecting the white dots are a reasonable median line representation of the chain.  Green and red dots indicate the left/right extent above 250 m.

smooth

Sorry, I also managed to run the tests and update the nat_nn.ps file as part of this.

### DIFF
--- a/doc/rst/source/grdtrack.rst
+++ b/doc/rst/source/grdtrack.rst
@@ -111,7 +111,8 @@ Optional Arguments
     *length* sets the full length of each cross-profile, while *ds* is
     the sampling spacing along each cross-profile. Optionally, append
     **/**\ *spacing* for an equidistant spacing between cross-profiles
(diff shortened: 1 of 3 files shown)


PaulWessel:
_zo_ is a threshold.  You want to find the first and last point on the profile that rises above the noise floor of _z0_.

PR #2723, opened 2020-02-12
grdinfo option to scan for maxima along rows or cols

PaulWessel:

The new **-E**[**x**|**y**][**+l**|**h**] option will examine each column (**-Ex**) or row (**-Ey**) in the given grid and, for each vector, determine the location and value of the maximum (or minimum via **+l**).  We then report the resulting series of (x,y,z) coordinates across all the columns (or rows).


joa-quim:
Following the *complicate it* tradition, you may want to add a **+a<val>|+b<val>** to only search **a**bove or **b**ellow a threshold. Useful to limit search under water.

PaulWessel:
Hm, can see this with a threshold of 0 for sea level.  We have things like this elsewhere, e.g., max but only among the negative values or min but only among the positive values.

PaulWessel:
Now you can get minimum of the positives or maximum of just the negatives, like in filter1d and grdfilter.

PR #2755, opened 2020-02-17
Let all gallery examples be run using SI Default settings

### LINKED ISSUE #2752: Use SI defaults for all examples
Recently, @PaulWessel updated all examples to SI units (cm or point instead of inch). However, the original PS files are still generated using US defaults, and in the gmtest script, we have 
```
gmt set -Du FORMAT_TIME_STAMP "Version 6" GMT_GRAPHICS_FORMAT ps
```

We should regenerate all PS files using SI defaults, and change `gmt set -Du` to `gmt set -Ds`.

PaulWessel:
Now that we have modified all examples scripts to use SI units (at least mostly), we wanted to remove the forced US setting and instead use the default SI setting for various defaults (which includes paper size A4).  After rerunning all examples this way, we needed to revert a few changes to **-X -Y** here and there since part of a few figures (11, 22, 39) had fallen off the paper (regardless of SI/US).  Also, ex09 needed to specify the wiggle scale using a plot unit (now 750c), and the star in ex15 was not visible so now slightly larger and red.  I made ex23 25 cm wide for clarity. Updated the wiggle rst to use similar language as elsewhere regarding appending units, and fixed a formatting error in mapproject.rst. Closes #2752.


PaulWessel:
Note: Apart from the changes in ex09, 15, and 23, the changes do not affect the PNGs produced for the docs.  Most changes only affect positioning on a paper page for the PostScript plots.

seisman:
ex18 needs to be updated.

PaulWessel:
I did 18, 45, and anim01.  What is missing in 18?

seisman:
sorry, I missed that file.

PR #2769, opened 2020-02-19
doc: Use earth_relief_02m instead of remote cache file USEast_Coast.nc

seisman:
`USEast_Coast.nc` is generated by `grdcut` the earth_relief_02m file.


doc/examples/anim04/anim04.sh
```diff
@@ -17,15 +17,16 @@ cat << EOF > pre.sh
...
-	gmt grdgradient @USEast_Coast.nc -A90 -Nt1 -Gint_US.nc
+	gmt grdcut @earth_relief_02m -R-100/-70/10/50 -GUSEast_Coast.nc
+	gmt grdgradient USEast_Coast.nc -A90 -Nt1 -Gint_US.nc
 	gmt makecpt -Cglobe -H > globe_US.cpt
 gmt end
 EOF
 # 2. Set up the main frame script
 cat << EOF > main.sh
 gmt begin
 	gmt grdimage -JG\${MOVIE_COL0}/\${MOVIE_COL1}/160/210/55/0/36/34/\${MOVIE_WIDTH}+ \
```
seisman: Note the trailing `+` in the -J option. Is it an obsolete syntax?

PaulWessel: Yes it is.  Means the max dimension so +du in new syntax.

seisman: I've changed it to the new syntax.

PR #2779, opened 2020-02-20
Remove unused structure element in grdinterpolate

PaulWessel:
Left-over from cut/paste probably.

### DIFF
--- a/doc/rst/source/grdinterpolate.rst
+++ b/doc/rst/source/grdinterpolate.rst
@@ -111,30 +111,22 @@ Examples
 --------
 
 To extract a single, new 2-D layer from the temperature.nc 3-D cube for level 3400
-using a cubic spline, try
-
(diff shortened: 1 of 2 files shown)


seisman:
@PaulWessel Check the output of `gmt grdinterpolate`. Some small issues:

1. -V is missing
2. -Z should be shown before -^

PaulWessel:
Unrelated, the explain_array.rst_ that we use to explain -T in many programs (including grdinterpolate) does not format the examples as I thought.  The trailing :: we use before giving an example results in a single colon being plotted.  Is this normal?

seisman:
Yes, see https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks.

PaulWessel:
Thanks.

PR #2782, opened 2020-02-21
admin: Simplify the image_optimize.sh script

seisman:
Use `pngquant --force --ext .png` to overwrite the original PNG files.


seisman:
ping @PaulWessel

Issue #2783, opened 2020-02-21
Specifying 2-D input grids to grdinterpolate


PaulWessel:
**Description of the desired feature**

The grdinterpolate module has not yet been released but will be soon (in 6.1.0).  Thus, this is a good time to discuss some aspects of it before we release it, as changing things later is more of a hassle.

grdinterpolate can sample a 3-D data cube along the 2-D plane normal to the time axis (or depth; will just call it time in this post).  The 3-D cube can be specified in two ways:

1. A 3-D netcdf file [Times are obtained from the data cube metadata]
2. A list of 2-D grids [Times are obtained from **-Zi**, either equidistant setup of a list]

While this may be fine if you have a manageable number of grids and they are simply spaced (e.g., imagine `gmt grdinterpolate temperature_*.grd -Zi0/1000/100`) for feeding in 11 grids at different depths 0, 100, 200, etc., it may be messier if you have grids at different times.  For that we added the option **-Zi**_times_, so you could do `gmt grdinterpolate INSAR*.grd -Zidates.txt`.

I can imagine that it may also be convenient to have a single list file that contains the times _and_ the file names per record:

```
2012-12-12T04:33 firstfile.grd
2012-12-16T14:03 nextfile.grd
...
```

and then pass that file to grdinterpolate somehow.  This also has the benefit of ordering the files properly and in only one place (versus making sure your input file order and level list match).  If you chose poor filenames then ls *.grd may not yield the desired order.

Given different users may have different needs., perhaps we should add a third way to enter the data:

3. A list file with times and corresponding file names given via a new option **-L**_listfile_.

Let me know what you think, hopefully David and Eric can comment on what is useful for them.

dsandwell:
Our files are not always equally spaced in time so the -Zidates.txt would be good.  The file names we use will always be listed in the proper order because the time is YYYYDDD where DDD  is the day of the year.  We could easily have YYYY.YYYYYY where the values behinf the decimal point are fractions of a year.

PaulWessel:
OK, since your needs are covered by the current scheme, we can always reconsider a **-L** option later if the need arises from some other quarter.

PR #2784, opened 2020-02-21
Let grdinterpolate also do 1-D profiles through the data cube

PaulWessel:

Initially, grdinterpolate was set up to interpolate a 3-D data cube at planes normal to the 3rd axis (usually time or depth), such that the planes could be saved as new grids.  It is also useful to sample such grid along lines that are parallel to the 3rd axis going through the grids at specified (fixed) lon,lat point(s).  This PR adds the new option **-S**_x/y_|_pointfile_[**+h**_header_] to do this.  Because **-S** changes the output type form grid to table I have added the usual common options that accompany table i/o (**-b**, **-i,** etc).  We pass the input points to the given grid layers to grdtrack and thus first just sample the layers at their levels.  If **-T** is used we then interpolate these series via sample1d.

I have tested it on some synthetic data and made sure **-S** works for a fixed point and a list of points, and that **-T** works with this as well.  It would be helpful if Dave and Eric could build this branch and try it out.  If you have a bunch of grids (*.grd), a table with the times (dates.txt), and a file with desired locations of the time-series (points.txt), you can try

`gmt grdinterpolate *.grd -Zidates.txt -Spoints.txt > profiles.txt`

To save to separate profile files you need to add **-G**profile_%d.txt or similar.


PaulWessel:
As per #2783 I have also implemented a grid slicing through the 3-D cube.  It is limited to a vertical slice but the surface expression (set with **-E**) can be complicated and sampling can be customized.  For 3-D layers with variable spacing you must also sample along the 3rd dimension via **-T**.  Added new modifiers **+p** (force sampling along a parallel if the line is so aligned) and **+g** (report distance as degrees longitude or latitude instead of great circle distances if line is so aligned ); the +g modifier is also added to grdtrack but +p is left off since grdtrack has **-Ap**.  Also, since grdinterpolate does not have **-A** I added modifier **+x** for lo**x**odrome.

Summary: **-S** drills down the cube at the given point(s) and produces a set of time/space series, while -E creates a vertical slice 2D grid from the 3D cube.

PaulWessel:
Hi @seisman, just checking if you are happy with this PR so I can merge it?  We can then finalize the -JP stuff from #2785.

seisman:
I didn't find time to try this PR, but the description of **-E** and **-S** looks good to me. It would be good if you can add a few tests.

PaulWessel:
OK, will do that.

PaulWessel:
The obvious examples are the ones from the seismic data set and I kind of want to fix the crazy #2785 stuff first so I can make tests that does not require awking those grids to make radial distances...  I think I will work on that first, so we can hold this one until ready.

PaulWessel:
Sorry, meant implementing #2791.

PR #2792, opened 2020-02-23
Reduce useless messages from gmtconnect

### LINKED ISSUE #2747: gmtconnect warning
Oddly this one fails in Julia (does not connect) but works on Matlab. However, in both cases it prints the warning
```
gmtmex('gmtconnect -T1.1', [0 0; 1 1], [1.1 1.1; 2 2])
gmtconnect [WARNING]: Trouble: Pass 1 found 1 while pass 2 found 2 connections!
```
But this one is silent
```
gmtmex('gmtconnect -T1.5', [0 0; 1 1], [1.1 1.1; 2 2])
gmtconnect [WARNING]: Trouble: Pass 1 found 1 while pass 2 found 2 connections!
```
something is weird. Even **-T0.1** should be enough, but
```
gmtmex('gmtconnect -T0.1', [0 0; 1 1], [1.1 1.1; 2 2])
gmtconnect [WARNING]: Trouble: Pass 1 found 0 while pass 2 found 1 connections!
gmtconnect [WARNING]: Trouble: Pass 1 found 0 while pass 2 found 1 connections!
```

PaulWessel:
gmtconnect does a first pass to determine which segments will be joined so that we can build a text string with these connection for the segment header.  The second pass then does the connections.  There is a smalll difference in how the two passes operate that makes the number of passes sometimes report different numbers of connections.  This is more of a debug message than anything else and hopefully this can be completely fixed soon.  Thus, I have moved this mesage from warning to debug so that id does not bother people.
Closes #2747.

### DIFF
--- a/src/gmtconnect.c
+++ b/src/gmtconnect.c
@@ -586,8 +586,8 @@ int GMT_gmtconnect (void *V_API, int mode, void *args) {
 				dd[SEG_I][END_A] = dd[SEG_J][END_B] = DBL_MAX;	/* Flag as single line segment so two ends are not used to connect them */
 				/* Only compute a distance for this single segment if it has more than 2 points.  2 point cannot connect back on itself as it would make a polygon with zero area */
 				dd[SEG_I][END_B] = dd[SEG_J][END_A] = (segment[iseg].n < 3) ? DBL_MAX : gmt_distance (GMT, segment[iseg].x_end[END_A], segment[iseg].y_end[END_A], segment[iseg].x_end[END_B], segment[iseg].y_end[END_B]);
-    			nearest_end[SEG_I][END_A] = nearest_end[SEG_J][END_A] = END_B;	/* Duplicate the nearest ID info since it is a single line segment compared to itself */
-    			nearest_end[SEG_J][END_B] = nearest_end[SEG_I][END_B] = END_A;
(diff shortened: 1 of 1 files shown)


src/gmtconnect.c
```diff
@@ -586,43 +586,43 @@ int GMT_gmtconnect (void *V_API, int mode, void *args) {
 				dd[SEG_I][END_A] = dd[SEG_J][END_B] = DBL_MAX;	/* Flag as single line segment so two ends are not used to connect them */
 				/* Only compute a distance for this single segment if it has more than 2 points.  2 point cannot connect back on itself as it would make a polygon with zero area */
 				dd[SEG_I][END_B] = dd[SEG_J][END_A] = (segment[iseg].n < 3) ? DBL_MAX : gmt_distance (GMT, segment[iseg].x_end[END_A], segment[iseg].y_end[END_A], segment[iseg].x_end[END_B], segment[iseg].y_end[END_B]);
-    				nearest_end[SEG_I][END_A] = nearest_end[SEG_J][END_A] = END_B;	/* Duplicate the nearest ID info since it is a single line segment compared to itself */
```
seisman: The indentations don't look right.

PaulWessel: Seems like Xcode editor likes to add spaces not tabs.  Need to see if there is a setting for this.  Also, would be nice if this was automated so that indentations were always one tab per level.

PR #2796, opened 2020-02-24
Fix some double-semicolons ;; to single-semicolon ;

seisman:
Double semicolon `;;` doesn't mean anything in C. I believe they're
typos.


joa-quim:
Guys, these type of fixes do not need PRs. Just apply them directly. 

Sent from my iDedo

Issue #2791, opened 2020-02-23
Plotting Earth cross-sections without going crazy


PaulWessel:
Seems to me GMT is not doing a great job here - but maybe I am not used to making these types of plots.  Seismologists like @seisman may have more experience with this.  To plot just the mantle and exclude the core, I know Earth radius is ~6370 km and core is at a depth of ~ 2890 km.  OK, so I want a windshield-wiper plot where the (not plotted) center of the Earth is at the projection center and the two arcs should be at r = 6370 (surface) and r = 6370-2890 = 3480 (core-mantle boundary).  So looks like this works: `gmt basemap -R0/180/3480/6370 -B -JP15c+z -png map` :

![map](https://user-images.githubusercontent.com/26473567/75101756-d4f41a00-5585-11ea-852e-93742a4c7be5.png)

**Problem 1:** Why do I have to do that math manually?  I want to use 0 and 2890 as my depths (z-value).

Lets say this is a slice along equator, so those degrees are degree longitude.  Then, the plot is OK if looked at from the N pole (since longitudes increase counterclockwise).  However, if I want to look at this from the S pole, then I need to do more work, and after some trial/error I select `gmt basemap -R0/180/3480/6370 -B -JPa15c/90+z -png map`:

![map](https://user-images.githubusercontent.com/26473567/75101794-8004d380-5586-11ea-9d27-2449312652ca.png)

**Problem 2:** This might be hard to find out for laypeople.  OK, that is fine I guess, perhaps fixable in the documentation.

Now I want to display seismic velocities in this crossection.  I take the S362ANI_kmps.nc and extract data from 25 to 2890 in steps of 5:

gmt grdinterpolate S362ANI_kmps.nc -E0/0/180/0+i1d -T30/2890/5 -Gslice.nc -Ap

That grid now has x from 0/180 and y from 25 to 2890.  How to plot as above?  Trying `gmt grdimage slice.nc -B -JP15c+z -png map` brings no joy:

![map](https://user-images.githubusercontent.com/26473567/75101895-0968d580-5588-11ea-99d2-0935a25fb189.png)

since there is no info on radius range or depth range, and the depth of 25 (start of grid) results in a tiny cavity at the center with ugly annotations.  OK, maybe need to use the **-R**0/180/3480/6370 from before? No, that gives 

> grdimage [WARNING]: Your grid y's or latitudes appear to be outside the map region and will be skipped.

which of course is true.  Hacking away I do

```
[...]

PR #2799, opened 2020-02-24
Polar upgrade

PaulWessel:

This PR changes the polar projection syntax in a backwards compatible way.  It used to be

**-Jp**[**a**]_scale_[/_origin_][**+r**|**z**]

and is now

**-Jp**_scale_[**+a**][**+o**_origin_][**+r**|**z**]

(same for **-JP**).  Unfortunately, I had forgotten to commit some minor changes related to error return codes in a few other files as well as addressing some Coverity fussing, so they are here too.  I think they are pretty harmless.


joa-quim:
Later. Now we have to take care of these beauties. 


BA8B49D3-6ED4-49F5-B747-4A0B5FC69AA8

doc/rst/source/explain_-J_full.rst_
```diff
@@ -273,12 +273,12 @@ projection names stands for Equal-Area and Conformal, respectively):
...
 
-    **-Jp**\ [**a**]\ *scale*\ [*/origin*][**+r**\|\ **z**] or **-JP**\ [**a**]\ *width*\ [*/origin*][**+r**\|\ **z**]
+    **-Jp**\ *scale*\ [**+a**]\ [**+o**\ *origin*][**+r**\|\ **z**] or **-JP**\ *width*\ [**+a**]\ [**+o**\ *origin*][**+r**\|\ **z**]
     (Polar coordinates (*theta*, *r*))
 
-    Optionally insert **a** after **-Jp** [or **-JP**] if *theta* is azimuths CW
+    Optionally add **+a** if *theta* is azimuths CW
     from North instead of directions CCW from East [Default]. Optionally
-    append /*origin* in degrees to indicate an angular offset [0].
+    append **+o**\ *origin* in degrees to indicate an angular offset [0].
```
seisman: I'm often confused about the definition of "offset". It seems the offset is CW by default, and CCW for **+a**. Usually I have to try both positive and negative values to find the one I need. It would be better if the description can be more clear if possible.

seisman:
It's not a bug. I gave invalid values by mistake. See below. The error messages are clear, but the 2nd error message (Internal Failure = i) may be improved/removed.
```
$ gmt basemap -JP5c  -R0/3600/0/1 -Bxa45f -pdf map  
basemap [ERROR]: Internal Failure = GMT_MAP_EXCEEDS_360
basemap [ERROR]: Internal Failure = i
psbasemap (GMT_psbasemap): Map region exceeds 360 degrees
```

PaulWessel:
Hm, I'll pass.  Note that this PR does not include what we discussed over the weekend.

seisman:
We also need to update the cookbook/coordinate_transformations.rst and maybe some example scripts.

doc/rst/source/explain_-J_full.rst_
PaulWessel: OK, relearning which way it goes and will update the docs shortly.

joa-quim:
OK, I also find the **+r** description a bit cryptic ``Finally, append +r if r represents elevations in degrees``. What is *elevation* in this context?

PaulWessel:
So the docs says

`Finally, append +r if r represents elevations in degrees (requires south >= 0 and north <= 90)`

so it is not that cryptic: It means that instead of radial distance you have elevation values from 0 (horizontal) to 90 (vertical), i.e., your data is _azimuth,elevation_ which are both angular measurements. I think 0 (horizontal means at the outer perimeter and 90 means center of projection.  So if you give north < 90 you get a donut.

joa-quim:
Right, the ``(requires south >= 0 and north <= 90)`` bit helped, but the donut example should be added to description as it makes things definitively more clear.

PaulWessel:
We can add improved language and examples when we take the next step with the #2791 proposal.

PR #2803, opened 2020-02-25
Update to -JP new syntax

seisman:
All tests still pass.


PaulWessel:
Check if the polarpanels.sh works.

seisman:
That's weird. I got different output, but the ctest run passed.

seisman:
`-JPa?+o45` mixes old and new syntaxes. It shouldn't work, but it works.

I've updated it to the new syntax `-JP?+a+o45`.

PaulWessel:
It works because checking for the leading 'a' does not set up a separate branch if found.  Perhaps it should so such things would be errors.

seisman:
Yes, it should report an error.

PR #2816, opened 2020-02-27
Implementing new polar projection modifiers and capabilities

PaulWessel:

The polar (r-theta cylindrical) Cartesian projection is being improved and enhanced to handle more situations.  The new syntax will be completely backwards compatible with the old syntax (only showing upper-case for map width here) which is:

**-JP**[**a**]_width_[/_base_][**r**|**z**]

1. The **a** switches from CCW angles to CW azimuth.
2. The optional /_base_ adds an angular rotation [0].
3. The trialing **r** means we have elevations (enforces _y_-range to 0-90 and adds degree symbol.
4. The trailing **z**  annotates _n - r_ instead or _r_.

The new syntax looks like this:

**-JP**_width_[**+a**][**+f**[**e**|**p**|_radius_]][**+r**_offset_][**+t**_rotation_][**+z**[**p**|_radius_]]

1. **+a** switches from CCW angles to CW azimuths.
2. **+f** flips radial direction. Append **e** for elevations (enforces _y_-range to 0-90 and adds degree symbol), **p** for planetary radius (determined from current ellipsoid in m or km as needed), or a specific radius.
3. **+r** adds a radial offset [0].
4. **+t** adds an angular rotation [0].
5. **+z** annotates _n - r_ instead or _r_.  Alternatively, annotate _radius - r_ where _radius_ is either planetary radius or specified radius.

The main improvements in functionality is in specifying depths in **-R** options and allow for a separate parameter that indicates the radius of the sphere, and more control of radial annotations.  The mapping portion of the changes have been implemented and is working (r,theta or z,theta to x,y), but there are a few issues pertaining to annotation justification.  Below if the output of the test polarcases.sh:

polarcases

The remaining issues to work on before approving this PR is

1. The **+f** annotations (except **+fe**) has some odd justification code changes both for the angles (e.g., middle-top plot, as well as for radial annotations.
2. The brand new radial offset mechanism (**+r**), with or without **+f**, needs more work.
3. The +**zp**|_radius_ has not been implemented yet nor documented.


seisman:
Can we switch the modifiers **+r** and **+o**? Here are my reasons:
- It's a r-θ plot. It's more intuitive to have **+r** controlling the radius offset. Moreover, **o** also looks like θ.
- In the current implementation, **+r***rotation* is explained as "adding an angular rotation". By default, theta is CCW angles from East, and **+r** rotates the plot CW. If **+a** is used, theta is CW azimuths from North, and **+r** rotates the plot CCW. I find the rotation very non-intuitive. 
Instead, we can say that **+r** set the angle (or *base* as used in the old description) of the *origin*.  **+r30** means that East=30°, and **+r30+a** means North is 30°. Such an explanation is more intuitive to me.

So maybe **+r***roffset* and **+o***origin*?

PaulWessel:
Yes, we can do that.  Hopefully later tonight.

PaulWessel:
We are saying radius and theta, so maybe +r and +t is better ?

seisman:
I'm OK with **+t**.

PaulWessel:
Some progress., but extending circle to 300 degrees shows more issues for the new cases, now as new test polarcircles.sh:

polarcircles

PaulWessel:
More progress.  I have written a new separate function for GMT_POLAR that determines justification and text angle of annotations.  The result for the polarcirles.sh test is below; as far as I can tell, everything is correct, including the new scheme **+zp** for annotating radial distances when giving **-R** in depths:

polarcircles

Moving to smaller regions, the polar cases.sh test output is below.  I see one problem: The radial allocation for **+f** at the apex is given the wrong justification code (it is correct for the non **+f** in lower left corner).  Also note that the new implementation has the inner and outer circle annotation match in justification.  Specifically the 0-degree annotation on the top rows have its base against the outline, whereas prior annotations would have that particular annotation be rotated 180 degrees.  For this reason, a small handful of tests fail.  We may decide to declare those successes and update the PostScript originals:

polarcases

It would be helpful if @seisman and @joa-quim tried this branch and ran all the examples and examine the failures and let me know what they think of the new justification.  I think we are close to wrapping these improvements up.

joa-quim:
The new justification is clearly better.
Note also that the apex case (**+f**) shows another, and probably related, issue. The ``1`` tick not actually in the apex but a bit inward offset. That is not the case of the LL plot.

PaulWessel:
The problems has to do with the code that finds the intersection of the tick and the boundary.  For +f when things are reversed, the angle at the apex switches 180 for some reason and that leads to the wrong justification code.  Will have to debug to see why this is the case.

PaulWessel:
I think this is the only issue (annotation with +f where the E and W sides meet at the apex).  Everything else looks good to me.  However, need to examine the 4-5 failures due to the changes.

seisman:
The failures look good to me.

PaulWessel:
OK, will try to work on the **+f** apex issue.  Have many letters to write this weekend, plus NSF reports...

joa-quim:
Yes,  I looked at the failures too and they are due only to how the ``0`` deg are labeled

PaulWessel:
I found the special case code (gmt_get_angle) that needed to do a flip when the **+f** was in effect.  Now the plot looks correct:

polarcases

PaulWessel:
Will rerun all tests and update changed originals then.

PaulWessel:
OK gang, all updated and looking sharp.  Taken the WIP off the title.

PaulWessel:
I just need to update the docs for **+z**[**p**|_radius_].  Forgot.

doc/rst/source/explain_-J_full.rst_
```diff
@@ -273,15 +273,18 @@ projection names stands for Equal-Area and Conformal, respectively):
 
 .. _-Jp_full:
 
-    **-Jp**\ *scale*\ [**+a**]\ [**+o**\ *origin*][**+r**\|\ **z**] or **-JP**\ *width*\ [**+a**]\ [**+o**\ *origin*][**+r**\|\ **z**]
+    **-Jp**\ *scale*\ [**+a**]\ [**+f**\ [**e**\|\ **p**\|\ *radius*]][**+r**\ *offset*][**+t**\ *rotation*][**+z**] or **-JP**\ *width*\ [**+a**]\ [**+f**\ [**e**\|\ **p**\|\ *radius*]][**+r**\ *offset*][**+t**\ *rotation*][**+z**]
```
seisman: **+z** syntax needs updates.

doc/rst/source/explain_-J_full.rst_
```diff
@@ -273,15 +273,20 @@ projection names stands for Equal-Area and Conformal, respectively):
...
-    from North instead of directions CCW from East [Default]. Optionally
-    append **+o**\ *origin* in degrees to indicate a CW (or CCW if **+a** iset set) angular offset [0].
-    Finally, append **+r** if *r* represents elevations in degrees (requires *south* >= 0
-    and *north* <= 90), or append **+z** if you want to annotate depth rather than
-    radius [Default]. Give *scale* in UNIT/r-unit.
+    Give *scale* in UNIT/r-unit.  Optionally add **+a** if *theta* is azimuth CW
+    from North instead of direction CCW from East [Default].
+    Select **+f** to flip the radial direction to point inwards, and append **e** to indicate
+    that *r* represents elevations in degrees (requires *south* >= 0 and *north* <= 90),
+    **p** to select current planetary radius as maximum radius [*north*], or *radius*
```
seisman:
```suggestion
    **p** to select current planetary radius (determined by :term:`PROJ_ELLIPSOID`) as maximum radius [*north*], or *radius*
```

doc/rst/source/explain_-J_full.rst_
```diff
@@ -273,15 +273,20 @@ projection names stands for Equal-Area and Conformal, respectively):
...
+    to specify such a radius directly.
+    Append **+r**\ *offset* to include a radial offset [0].
+    Append **+t**\ *rotation* in degrees to indicate a CCW angular offset [0].
```
seisman: > Append **+t**\ *rotation* in degrees to indicate a CCW angular offset [0].

It can be CW or CCW depending on the **+a** modifier. Maybe say:

> Append **+t***theta* in degrees to indicate the angle of the East direction or North direction if **+a** is used [0].

test/subplot/polarcases.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+# Test new implementation of -JP modifiers
+gmt begin polarcases ps
+	gmt set FONT_ANNOT_PRIMARY 7p MAP_TICK_LENGTH_PRIMARY 2p MAP_ANNOT_OFFSET_PRIMARY 2p FONT_TITLE 12p
+	gmt subplot begin 4x3 -Fs2i -BWSNE -Bafg -M3p
+		gmt basemap -R0/120/0.5/1 -JP? -c -B+t"Default"
+		gmt basemap -R0/120/0.4/1 -JP?+f -c -B+t"\053f"
```
seisman: Why use `\053`? Using `+` is OK, right?

test/subplot/polarcases.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
seisman: Tests should be grouped by topic. These tests are not for testing `subplot` and should be moved to other directories, e.g. projection.

src/gmt_init.c
```diff
@@ -6527,7 +6527,7 @@ void gmtlib_explain_options (struct GMT_CTRL *GMT, char *options) {
 			gmt_message (GMT, "\t       Specify region in oblique degrees OR use -R<>r\n");
 			gmt_message (GMT, "\t       Upper-case A|B|C removes enforcement of a northern hemisphere pole.\n");
 
-			gmt_message (GMT, "\t   -Jp|P<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<rotation][+z[p|<radius>]] (Polar (theta,radius))\n");
+			gmt_message (GMT, "\t   -Jp|P<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<origin][+z[p|<radius>]] (Polar (theta,radius))\n");
```
seisman:
```suggestion
			gmt_message (GMT, "\t   -Jp|P<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<origin>][+z[p|<radius>]] (Polar (theta,radius))\n");
```

src/gmt_init.c
```diff
@@ -7608,9 +7608,9 @@ void gmt_syntax (struct GMT_CTRL *GMT, char option) {
 					gmt_message (GMT, "\t  <scale> is <1:xxxx> or %s/degree, or use <width> in %s\n", u, u);
 					break;
 				case GMT_POLAR:
-					gmt_message (GMT, "\t-Jp<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<rotation][+z[p|<radius>]] OR -JP<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<rotation][+z[p|<radius>]]\n");
+					gmt_message (GMT, "\t-Jp<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<origin][+z[p|<radius>]] OR -JP<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<rotation][+z[p|<radius>]]\n");
```
seisman:
```suggestion
					gmt_message (GMT, "\t-Jp<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<origin>][+z[p|<radius>]] OR -JP<scale>|<width>[+a][+f[e|p|<radius>]][+r<offset>][+t<origin>][+z[p|<radius>]]\n");
```

doc/rst/source/proj-codes.rst
```diff
@@ -85,8 +85,8 @@ The projection codes for the GMT-style and the PROJ-style are tabulated below.
    * - **-JOc**\ |lon0|/|lat0|/|lonp|/|latp|/\ *width*
      - **-Jomercp/**\ |lon0|/|lat0|/|lonp|/|latp|/\ *scale*
      - Oblique Mercator, 3: origin and pole :ref:`... <-Jo_full>`
-   * - **-JP**\ *width*\ [**+a**]\ [**+o**\ *origin*][**+r**\|\ **z**]
-     - **-Jpolar**\ *scale*\ [**+a**]\ [**+o**\ *origin*][**+r**\|\ **z**]
+   * - **-JP**\ *width*\ [**+a**]\ [**+f**\ [**e**\|\ **p**\|\ *radius*]][**+r**\ *offset*][**+t**\ *rotation*][**+z**\ [**p**\|\ *radius*]]
+     - **-Jpolar/**\ [**+a**]\ [**+f**\ [**e**\|\ **p**\|\ *radius*]][**+r**\ *offset*][**+t**\ *rotation*][**+z**\ [**p**\|\ *radius*]]
```
seisman:
```suggestion
   * - **-JP**\ *width*\ [**+a**]\ [**+f**\ [**e**\|\ **p**\|\ *radius*]][**+r**\ *offset*][**+t**\ *origin*][**+z**\ [**p**\|\ *radius*]]
     - **-Jpolar/**\ [**+a**]\ [**+f**\ [**e**\|\ **p**\|\ *radius*]][**+r**\ *offset*][**+t**\ *origin*][**+z**\ [**p**\|\ *radius*]]
```

PaulWessel:
Noticed a few things:

1. Radial annotations did not set the proper format.  So if increments were 0.2, we had 0, 0.2, ... , 1 and not 0.0 and 1.0.  This is now fixed.
2. When there is no inner radius then the W and E radii join at the center, and the annotations at that point spill onto the map.  See polarcircles.sh for the lower left example.  I am working on that now.

PaulWessel:
Just to clarify; see below.  I think the only reasonable solution for this case is not to annotate at the apex.  Agree?  I think that rule only applies to the apex.  The fact that in this example the two 0.2 labels overprint is more of a user concern.  They can turn of W or E, for instance, but the remaining 0.0 will still plot badly.

map

seisman:
I agree. The ticks at the apex are also not good.

image

PaulWessel:
Ah, but this whole thing depends on the angle at the apex. We have several examples where the west/east limbs are just 90 degrees in longitude apart and then there is no problems at all.  I think if the angle at the apex is < 180 then we turn off those labels, otherwise they are OK.

seisman:
Agreed.

PaulWessel:
A bit more thinking is required.  See plot with angles going from 0/90 lower right to 0/170 upper left. Based on that perhaps the rule should be a bit more sophisticated:

1. If (_east-west_) == 180 then we annotate the apex since if selected it will overprint exactly (or if **-BWE** then we can make sure we only print it once).
2. If (_east-west_) < 135 the we can annotate the apex for both limbs.
3. Otherwise we suppress apex annotation entirely.

Users may need to wisely choose annotation font size since a too big font (like in this example) may cause trouble even with this rule.  But we cannot really compute the optimal angle cutoff (well, we could try, but that can be added later if it is worth it).

bad

PaulWessel:
One more rule.  If only W or E has annotations (e.g., **-BWe**) then there is no risk of overprinting at all.  So only the **-BWE** case is of concern.

PaulWessel:
No failures, apex cases look fine.  Perhaps OK to accept this one now, @seisman?

PaulWessel:
I could not resist trying to find the optimal range for turning off the apex label.  See sketch: When alpha > beta we cannot place the labels.  Dashed line is bisector of those two other lines.  Changing annotation font size, tick length, and annotation offset all affects alpha.  I approximate the label by the rectangle whose height is approximated by the font size and our ratio of how tall the letter N is relative to the full font size.

IMG_1048

PR #2817, opened 2020-02-28
Literally include scripts in the cookbook to avoid repeating them

seisman:
Besides what the title says, this PR also updates some plotting scripts.

Some scripts use two commands to make plots: first "gmt set",
then one-liner "gmt coast".
I believe it's a bad practice, as it generates a "gmt.conf" file
in the current directory and affects subsequent scripts.
It also generates a PS file, but we're trying best to avoid PS in the cookbook.

These scripts are rewritten using "begin" & "end" modern mode.


seisman:
BTW, there are still many one-liner scripts, which can't be literally included in the cookbook, because these scripts use "-ps" but we want to show "-pdf" in the cookbook.

seisman:
Ping @PaulWessel

PR #2830, opened 2020-03-01
Add slicing test for grdinterpolate

PaulWessel:

We use curl to download the 3-D seismic velocity grid from IRIS, sample it along a slice through the Equator, and plot some polar projection maps.


PaulWessel:
Closes #2791.

test/grdinterpolate/slices.sh
```diff
@@ -0,0 +1,22 @@
...
+		gmt grdimage vs.nc -JP?+fp -c  -BWESn
+		gmt grdimage vs.nc -JP?+fp+a+t90 -c -BWESn
+		gmt grdimage vsv.nc -JP?+fp -c  -BWESn
+		gmt grdimage vsv.nc -JP?+fp+a+t90 -c -BWESn
+		gmt grdimage vsv.nc -JP?+fp+zp+a+t90 -c -BWESn
+		gmt grdimage vsh.nc -JP?+fp -c  -BWESn
+		gmt grdimage vsh.nc -JP?+fp+a+t90 -c -BWESn
+		gmt grdimage vsh.nc -JP?+fp+zp+a+t90 -c -BWESn
+	gmt subplot end
+gmt end show
```
seisman: We should try our best to not add large files to the repository. No need to plot the vsv and vsh here. They are almost the same as vs.

PaulWessel: True, but nothing is added to the repo here?  Everything is derived locally from the one file that is curled in.

seisman: The PS file is almost 2 MB.

PaulWessel: OK, I see what you are getting at. True, I will remove some panels and dumb down the interpolation a bit.

joa-quim:
I mentioned this idea already but apparently got unnoticed.
If we store the PNGs instead of the PSs as original we can save quite some space and speed up the tests because the originals would be already rasterized.

But the repo size would increase because the old PS are there already.

PaulWessel:
Yes, there are many things we wish we now could remove from the repo.  Seems like there has to be a way to surgically do this...

seisman:
> If we store the PNGs instead of the PSs as original we can save quite some space and speed up the tests because the originals would be already rasterized.

Storing PNG may not be a good idea. If we make a small change to a PNG file and commit it, git will store both the old and the new copy in the git pack, because git can not diff a binary file.

PaulWessel:
Is there any mechanism to use git to checkout a repo but exclude a directory in it? I.e., not checking out the test dir?

PaulWessel:
subversion had that via a depth setting.

seisman:
I don't think it's possible with git. But most users can have a shallow clone
```
git clone --depth 50 https://github.com/GecgccnericMappingTools/gmt
```
I think it's only ~300 MB.

seisman:
This is what I have with a shallow clone
```
114M	./test
480K	./cmake
 40K	./ci
164K	./admin
 28K	./.github
 67M	./doc
 97M	./.git
 14M	./share
 15M	./src
307M	.
```

PR #2839, opened 2020-03-02
Start expanding the psxy|z option -Z to read a file

joa-quim:


PaulWessel:
I see a **+d** modifier.  May I remove that and just check **-Z**_arg_ if arg passes the gmt_access (R_OK) check instead?

PaulWessel:
We do that in other places expecting a value or a file, e.g., -I for intensities.

joa-quim:
Sure. That **+d** was temporary anyway. My idea was transform the old **l|f** into **+f** & **+f** but then **+f** could not be used for file. Keeping the **l|f** facility for files would be nice too.

PaulWessel:
And I see I failed to convert the l and f to +l and +f, so will have to do that yet be backwards compatible.

PaulWessel:
OK, first cut, compiles.  Note a few changes:

1. Implemented the **+l** and **+f** modifiers, backwards compatible
2. Set W.active and/or G.active like the other **-Z** stuff
3. Not setting change and commented out stuff checking for change == 4 for now.

Please test.  The only difference between **-Z**_value_ and **-Z**_file_ should be that we supply a new value per record, not the rest of the logic.  I commented out another line where Z.active was used.  Please test first and let me know.

PaulWessel:
Also updated plot man page (but not psxy yet).  Once all works we must duplicate this in psxyz.

joa-quim:
Break somewhere

```
julia> plot(neigh, C=C, choro=rand(size(neigh)[1],1), name="run2.png", title="Run2", show=true, Vd=1)
        psxy -JX12c/0 -Baf -BWSen+tRun2 -R562000/612000/4482000/4532000 -C -ZC:\TMP\GMTjl_Zvals.dat -P -K > C:\TMP\GMTjl_tmp.ps
psxy [ERROR]: Could not decode C:\TMP\GMTjl_Zvals.dat, return NaN.
```
but the file ``C:\TMP\GMTjl_Zvals.dat`` is fine.

PaulWessel:
OK, will try on the command line first here.  Do you have a handy test data set I can use?

joa-quim:
You can use this small files

PaulWessel:
No files?

joa-quim:
By mail they didn't want to go

[Z.zip](https://github.com/GenericMappingTools/gmt/files/4278394/Z.zip)

PaulWessel:
OK, got it to work I think.

joa-quim:
Getting crashes on julia/debug mode. First was *outline_setting* used without being initialized. Done that but now the same with *change*. Isn't this a sign of something?

PaulWessel:
OK, the outline_setting is now fixed as should the rest.  WOrks on command line.

joa-quim:
You only changed *outline_setting * but next come a crash for the same reason but with *change*.
Command line or via externals should be exactly the same. The difference is the compiler.

PaulWessel:
Where is change used when Zin != NULL?

joa-quim:
Fixed a typo. Update before continuing

joa-quim:
With *change* initialized to 0 now it works.

PaulWessel:
OK, expanding to psxyz and man pages.

joa-quim:
But we can't have this for externals, right (they can't be accessed)?
```
				if (gmt_not_numeric (GMT, &opt->arg[j]) && !gmt_access (GMT, &opt->arg[j], R_OK)) {	/* Got a file */
					Ctrl->Z.file = strdup (&opt->arg[j]);
					n_errors += gmt_M_check_condition (GMT, Ctrl->Z.file && gmt_access (GMT, Ctrl->Z.file, R_OK),
					                                   "Option -Z: Cannot read file %s!\n", Ctrl->Z.file);
				}
```

PaulWessel:
gmt_access works fine with externals since a memory given can always be accessed.  see gmt_access definition in gmt_io.c

PaulWessel:
Basic test of psxyz seems to work too.

PaulWessel:
Added a test each for psxy and psxyz. Please check all and if good to go then perhaps we can ask for approvals.

joa-quim:
Cool, it now works from Julia without resorting to the trick of saving a temp file.

joa-quim:
Tests pass. Ready to go.

Issue #2838, opened 2020-03-02
New failing tests


seisman:
Linux:

-  grdmask/polarhole.sh

Windows:

- grdmask/polarhole.sh
- psbasemap/polarapex.sh

PaulWessel:
Are they tiny things? I updated those PS over the weekend from how they looked on macOS.

seisman:
grdmask/polarhole.sh fails due to the different orientation of an annotation. See the diff file below.
![image](https://user-images.githubusercontent.com/3974108/75706668-8f250900-5c8b-11ea-981f-86772f2b6eeb.png)

psbasemap/polarapex.sh fails because it doesn't have a shebang at the beginning. I'll fix it in the master branch directly.

PaulWessel:
Hm, presumably tiny round-off affecting the logic in the new gmtlib_polar_prepare_label in gmt_support.c.  For instance, we have things like

```
case 1:
	if (angle >= 90.0 && angle < 270.0) *justify = 7, *text_angle = angle - 180;
	else  *justify = 5, *text_angle = angle;
```

and if angle is 90.000000001 or 269.9999999 then we may trigger the first test even though the label clearly belongs in the second category.  So probably should be

```
case 1:
	if ((angle-GMT_CONV4_LIMIT) >= 90.0 && (angle+GMT_CONV4_LIMIT) < 270.0) *justify = 7, *text_angle = angle - 180;
	else  *justify = 5, *text_angle = angle;
```

Do you want to test that sort of solution on Linux?

seisman:
No, it doesn't work.

PaulWessel:
OK, could you please attach the pdf (or a PNG version of it) so I can see it better (and not the change PNG).

seisman:
[polarhole.pdf](https://github.com/GenericMappingTools/gmt/files/4278762/polarhole.pdf)

PR #2842, opened 2020-03-03
Fix angle alignment and restore pen settings

PaulWessel:
This PR addresses #2838 by allowing for some slight round-off in angle values near the cut-off limits used in the choices for justifications.  A few PS originals needed updates as well.

It also corrects some mistakes I introduced in #2839 that caused several checks to fail.


PaulWessel:
Before I merge, might @seisman check if this works the same on Linux and the others now?

seisman:
OK. I'll enable CI tests for this branch later.

seisman:
All CI tests passed. Good to merge.

PR #2848, opened 2020-03-04
New -B modifier +i for internal annotations

PaulWessel:
For some projections, such as azimuthal or polar with 360 longitude range or global projections with 180 latitude range and where the poles are points, selected annotations may be excluded since the required axis along which to plot them do not exist.  This new modifier will force additional annotations inside the map frame along the specified parallel of meridian.

### DIFF
--- a/doc/rst/source/cookbook/options.rst
+++ b/doc/rst/source/cookbook/options.rst
@@ -286,7 +286,7 @@ tick, and gridline intervals, axes labels, and annotation units.
 
 The Frame settings are specified by
 
--  **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+n**][**+o**\ *lon/lat*][**+t**\ *title*]
+-  **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+i**\ [*val*]][**+n**][**+o**\ *lon/lat*][**+t**\ *title*]
 
 Here, the optional *axes* dictates which of the axes should be drawn
 and possibly annotated.  By default, all 4 map boundaries (or plot axes)
@@ -303,6 +303,10 @@ corner and the order goes counter-clockwise.  Append **+b** to draw the outline
 of the 3-D box defined by **-R**; this modifier is also needed to display
 gridlines in the x–z, y–z planes.  You may paint the
 map canvas by appending the **+g**\ *fill* modifier [Default is no fill].
+Use **+i** to annotate an internal meridian or parallel when the axis that normally
+would be drawn and annotated does not exist (e.g., azimuthal map with 360-degree range
+has no latitude axis, and a global Hammer map has no longitude axis);
+optionally append the parallel or meridian [0].
 If gridlines are specified via the Axes parameters (discussed below) then
 by default these are referenced to the North pole.  If, however, you wish
 to produce oblique gridlines about another pole you can append **+o**\ *lon/lat*
--- a/doc/rst/source/explain_-B_full.rst_
+++ b/doc/rst/source/explain_-B_full.rst_
@@ -3,7 +3,7 @@
 **-B**\ [**p**\|\ **s**]\ *parameters*
     Set map Frame and Axes parameters. The Frame parameters are specified by
 
-    **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+n**][**+o**\ *lon/lat*][**+t**\ *title*]
+    **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+i**\ [*val*]][**+n**][**+o**\ *lon/lat*][**+t**\ *title*]
 
     where *axes* selects which axes to plot. By default, all 4 map boundaries
     (or plot axes) are plotted (named **W**, **E**, **S**, **N**). To customize,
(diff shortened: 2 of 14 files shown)


PaulWessel:
So far I have implemented the case for azimuthal/polar internal annotations.  First case is a polar (r-theta) map (`gmt basemap -R0/360/0/10 -JP6i -Bxafg30 -Byafg -B+i90 -png map1`):

map1

Next up is a S polar map (`gmt basemap -R0/360/-90/-30 -JA0/-90/6i -Bxafg30 -Bya15g15+f -B+i-60 -png map2`): 

map2

And finally a N polar maps (`gmt basemap -R0/360/30/90 -JA0/90/6i -Bxafg -Bya15g30+f -B+i60 -png map3`):

map3

Given this, time to ask some questions of @joa-quim and @seisman before I move along:

1. The annotations are offset from the exact location by MAP_ANNOT_OFFSET.  This seems reasonable to me; what you you think?
2. There are no tick marks or other symbols being placed at the exact (lon/lat) spots. Is it important to mark the spots? Since there is no axis drawn, it seems odd to draw ticks. Could place a small grid cross there?
3. There is no painting to white out gridlines that go through the annotations.  One possibility is to say if **+g**_fill_ was used in the frame **-B** (to paint the inside of the map) then perhaps it makes sense to paint the bounding box for each label that way.  This means you could use **+g**_white_ to get the effect as well.
4. I have included the central annotation (r = 0) but excluded the perimeter annotation.  OK with this?

joa-quim:
1 - I would rather not have the MAP_ANNOT_OFFSET but perhaps have it with respect to the theta circles. See [this Matlab plot](https://www.mathworks.com/matlabcentral/mlc-downloads/downloads/e5ac47b2-4a80-11e4-9553-005056977bd0/ccdd7449-1c32-43a8-a5a2-f833b3906d9d/images/screenshot.jpg)

2 - Not sure what's better. Perhaps no ticks.

3 - Crossing annotations with lines is not pretty but we cannot paint to avoid it because it would also be painting everything else that is bellow.

4 - Perhaps only annotate r_min if r_min is not zero (donuts)?

PaulWessel:
One thought is that if you do draw the gridlines then having a corner justification and annotation offset would allow the annotations to miss the gridlines for most cases.

PaulWessel:
By using corner justification we avoid the problem with gridline overprinting the annotations:

map3

I think this beats the other arrangement above, no? I like it and I am coding it (!).

PaulWessel:
Work that remains to be done:

1. Make it also work for the polar r-theta when **+f**, **+x**, **+a**, or **+t** are used, and add the azim_radial.sh test
2. Implement for parallel as well and add test global_parallel.sh.

seisman:
It seems **+i** is a modifier for frame settings, why is that?
In your first example, `gmt basemap -R0/360/0/10 -JP6i -Bxafg30 -Byafg+i90 -png map1`, does this one make more sense? **a** in **-By** controls the annotation intervals, and **+i90** controls the locations.

PaulWessel:
The new feature is an _internal axis_ separate from the axes specified via -BWSNEZ.  That is why it should go with the frame settings.  It is not an axis setting.

seisman:
OK, I take your point.

PaulWessel:
Wrapped up radial annotations for polar r-theta projection, with new test azim_radial.sh. Thus only the longitude annotations for global projections remain. However, two new failures show up: GMT_App_F_stand+_iso+.sh I have seen before and fixed but cannot recall what the hell it is again - must rediscover.  The other is GMT_eckert4.sh which makes no sense.  It runs fine by itself but does not produce anything when run via ctest.  All the new code is under an if-test that is not triggered by that script so a mystery.  If anyone has any hints please let me know.

azim_radial

joa-quim:
Just noticing one thing. For the azimuthal example the annotations are above the line they are annotating but for the r-theta they are below the line.

Question. Being **-i** a new axis can we draw only de annotations but not the axis (like in the Matlab example)?

seisman:
`GMT_eckert4.sh` fails because I used `-pdf` for the one-liner command. I've fixed it and committed directly to the master branch.

PaulWessel:
Answering @joa-quim:

1. I have fixed the azimuthal so it is the same layout as the r-theta cases
2. No axis is drawn, just annotations, and if gridlines are not activated or goes through the selected meridian then I draw small grid crosses.

PaulWessel:
OK, the longitude annotations are much easier and looks like below.  Let me know if you think I should skip the first annotation as well (I am skipping the last).

azim_global

PaulWessel:
BTW, here is the updated azim_polar.sh output:

azim_polar

PaulWessel:
I have removed the WIP and opening this for review.

seisman:
The *internal* annotations don't use minus symbols correctly.
image

joa-quim:
:scissors: to long minus. They waste valuable space.

PaulWessel:
Seriously!?  There is minus and there is hyphen.  We have spent lots of time getting the character sets stuff working correctly.  Not changing that.

joa-quim:
I know we won't be changing it but you may remember I never liked the original change. Those math minus are just too long. The hyphen char looks much better. But not arguing we should go back.

seisman:
Something wrong with the CI. No need to wait for it.

PR #2852, opened 2020-03-06
Simplify gmt5syntax with "gmt --show-classic"

seisman:
Use `gmt --show-classic` to list all the module names.


seisman:
`chomp` is a built-in Perl function to remove trailing characters (`\n` by default) of a string or an array of string.

seisman:
`gmt --show-classic` output module names line by line, so the array would be 
```
("psbasemap\n", "psxy\n", ...)
```
Have to use `chomp` to remove the trailing `\n`.

PaulWessel:
Forgot it was a Perl script...

PR #2857, opened 2020-03-06
Correct keys settings for files to be read via options

PaulWessel:
A few keys were missing flag to skip modifiers, and some modules missed a key entirely.  Specifically:

```
originator: FD( should be FD(= since there are optional modifiers
psxy[z]: ZD( should be ZD(= since there are optional modifiers
sample1d: Added missing TD(
gmtvector: Remove AD( key since -A does not take a file
surface: DD( should be DD(= since there are optional modifiers
```

I tested and psxy works in Matlab with **-Z**[**+f**|**l**] with data given via memory works fine. @joa-quim please try Julia.

A remaining concern is psrose whose **-E**[**m**|[**+w**]_file_] is _either_ input or output: **-E**_file_ will read from _file_ (so key **ED(** is good), but **-E+w**_file_ writes to _file_.  I do not think we can have the same key (**-E**) twice so this is bad design; also **-F** was missing from usage so we are out of more letters.  Proposed solution:

Specify the KEY as **ED-** and set it to **ED(** or **ED)+w** depending on the **-E** option.  This remains full backwards compatibility.


PaulWessel:
I added the fix for psrose indicated above.

Issue #2749, opened 2020-02-17
Custom fonts, ex31, and better handling


PaulWessel:
This is not a bug but I noticed when running the tests I get this warning from ex31:

`gmt [WARNING]: Representation of font type not recognized. Using default.`

Now, when running it manually I do not get that error. Yet, the plot that is generated does not use the custom fonts LinBiolinum* and LinLibertine* ( you can see the text, using Courier I think, exceeds the legend).

![ex31](https://user-images.githubusercontent.com/26473567/74615329-dd4edf80-50c3-11ea-9920-63e71e92c85a.png)

The nightly build of the documentation and gallery show the correct fonts.  Some searching led me to rediscover this line in docs/scripts/CMakeList.txt:

-C-sFONTPATH="${GMT_SOURCE_DIR}/doc/examples/ex31/fonts

being passed to **psconvert** when it makes all the PNG plots needed for the docs.  While that is OK, I think perhaps we should pass `C-sFONTPATH="${GMT_SOURCE_DIR}/doc/examples/ex31/fonts` to **gmt begin** so that running the script stand-alone will produce the correct answer.  But of course `${GMT_SOURCE_DIR}` is not a defined variable when you just run the script.  Testing alternatives, I changed the first few lines of the script to

```
FDIR=/Users/pwessel/UH/RESEARCH/CVSPROJECTS/GMTdev/gmt-dev/doc/examples/ex31/fonts
gmt begin ex31 png C-sFONTPATH=${FDIR}
```

That worked, but is not particularly nice.  Can we think of a better way to handle these cases?   My first attempt failed because I was not in the ex31 directory when running the script so any relative path would fail.  It would also be nice to say

`gmt begin ex31 png C-sFONTPATH=fonts`

meaning the subdirectory fonts in the current directory.  But that would also only work if running from that directory since we need to detect the current directory.  Perhaps custom fonts should be always looked for in ~/.gmt/fonts?  If such a directory exists perhaps **psconvert** should automatically append the **-C** option?  Alternatively, using custom fonts requires adding them to PSL_custom_fonts.txt so if that file exists we may wish to add that directory?  Or perhaps the PSL_custom_fonts.txt file should have an optional `#FONTDIR=directory` comment that we can parse and use in **psconvert**?

[...]

PR #2858, opened 2020-03-07
ex31: Set PS_CONVERT to make the example portable

seisman:
Partially address #2749.

With the following changes, users can run the example outside the ex31 directory.
```
# Path to current directory
EX31DIR=$(dirname $0)

gmt set PS_CONVERT="C-sFONTPATH=${EX31DIR}/fonts"
```
If you think the change is good, I'll work further to clean up some stuff.


seisman:
ping @PaulWessel

Issue #2795, opened 2020-02-24
block* and -G from externals must always return grids


joa-quim:
When calling from externals with -G we are expecting to always get grids, but if, for example, condition
``"No data records found; no output produced`` is triggered we return error and no grids. This in itself will trigger calls to the external ``error()`` function, which is ... not good. What should happen is the return of empty grids (that is, all NaNs)

PaulWessel:
Yes, need elegant solution for this.  E.g.,

```
if (n_read == 0) {	/* Blank/empty input files */
	GMT_Report (API, GMT_MSG_WARNING, "No data records found; no output produced\n");
	GMT_External_Cleanup (API, GMT_IS_GRID, file);  /* Possibly other args but -R should be set already */
	Return (GMT_NOERROR);
}
```

where GMT_External_Cleanup is in charge of creating an empty grid or a no-record data set etc and call GMT_Write_Data on that if API->external is true.  in blockmean there may be several grids requested so there may need to be a loop there (module specific).  Perhaps think about where else we may run into this and take note of how GMT_External_Cleanup needs to work.

PaulWessel:
Fixed. closing

PR #2861, opened 2020-03-08
Increase information encoded in virtual file names

### LINKED ISSUE #2476: grdmath for externals seems to only accept input grids, not lines
```
julia> G = gmt("grdmath -R60/95/6/24 -I0.1 ? LDIST", mat2ds([73 24; 89 6]))
ERROR: grd_init: input (Array{GMT.GMTdataset,1}) is not a GRID container type
```
bu it works if I first save the the line into a file and next do

```
julia> G = gmt("grdmath -R60/95/6/24 -I0.1 line.dat LDIST")
```

### LINKED ISSUE #2860: Issues related to API and backwards compatibility

We made some changes to the API after it was first introduced in 5.  To avoid any backwards compatibility issues we added code like this, e.g. at the end of the _struct GMT_DATASEGMENT_ definition:

```
#ifdef GMT_BACKWARDS_API
	double **coord;
#endif
```

since we had briefly used the variable _coord_ for what we now uniformity call _data_.  There are a few like that in gmt_resources.h. I do not know if there are anyone whose code would break if we removed this.  However, for 6.x there may be new things that requires a backwards compatibility.  The GMT_STR16 (value 16) was a dumb name and I will need to increase it to 32 and introduce a new constant called GMT_VF_LEN = 32 instead.  The GMT_STR16 will be kept but set to 32 as well.

Question: Perhaps GMT_BACKWARDS_API need to be more version-specific. Maybe it should be GMT_API_V5 and my new change could go under GMT_API_V6.0 or something?

PaulWessel:

When memory resources are passed from one module to another, or when GMT modules are called from external interfaces (MATLAB, Julia, Python) we create a virtual file name that represents the memory container and whose encoded object ID allows us to find the memory.  These names have the format `@GMTAPI@-######`.  To prepare for a simpler scheme down the road I have extended the format of the virtual file names to `@GMTAPI@-S-D-F-A-G-M-######`, where the extra letters convey information about primary/secondary data, input/output, what data type, the geometry, and if a dummy messenger container for output.  Currently, this information is encoded in the virtual file name but only the object ID is decoded (as before).  Over time I hope to decode some of the other information for data validation.

In the process I fixed two issues: grdmath can now deal with dataset input for operators such as LDIST and INSIDE, and block*m modules can now return empty NaN grids if no data points fell inside the region.  This PR closes #2476 and #2795.

Finally, I have abandoned the oldest functions and parameters used in the initial 5.2 API; This closes #2860.  Three test programs have been removed since no longer testing anything useful.


PaulWessel:
Hope @joa-quim can give this a try with Julia for the grdmath and blockm* cases.

joa-quim:
Looks like I'll have to revisit my MB built (a potential big pain). With this branch I get
```
julia>
[ Info: Precompiling GMT [5752ebe1-31b9-557e-87aa-f909b540aa54]
gmt [ERROR]: Unable to open GMT shared mbgmt library: Error <127>: The specified procedure could not be found.
```

PaulWessel:
Might be worth it.  Some months ago we added gmt_core_module.c etc functions to report both classic and modern names.  Perhaps some naming changes is behind your error.

joa-quim:
I suspect more that I was using some GMT5 API thing that now is gone. I don't get that error with master branch.

PaulWessel:
Maybe thinks like header->nx instead of n_columns?

joa-quim:
Maybe, but those should raise compiler errors.

PaulWessel:
True. When checking what special gmt_supplements_* function we have I find this:

```
nm -g supplements.so | grep gmt_supplements
00000000000063d0 T _gmt_supplements_module_classic_all
00000000000065b0 T _gmt_supplements_module_group
0000000000006510 T _gmt_supplements_module_keys
0000000000006360 T _gmt_supplements_module_list_all
0000000000006160 T _gmt_supplements_module_show_all
```

So I think your mb build should have the same *module* calls.  They have changed a bit since our structure now contains both classic and modern names.

joa-quim:
Well, I was not rebuilding the mb supplement. Now I'm getting many compiling errors. Let's see if it simple to clear them.

joa-quim:
Seems we changed the ``gmt_init_module()`` API.It now has one more argument.

PaulWessel:
That was many moons ago though.

joa-quim:
We removed something related to pthread in CMake scripts didn't we? I have tons of these
```
c:\progs_cygw\mb-system\include\iowrap-posix.h(81): fatal error C1083: Cannot open include file: 'pthread.h': No such file or directory
```

joa-quim:
What happens is that I have this
```
	set (EXTRA_INCLUDE_GMTSUPPL "mb/suppl_add_mb.txt")
```
but now the use of ``EXTRA_INCLUDE_GMTSUPPL`` has been removed from the CMakeList and my MB build it broken.

PaulWessel:
Maybe @seisman can advice on the pthreads and EXTRA_INCLUDE_*, but for non-GMT shared libraries they should build separately and only link with the GMT libraries, no.  E>g., see gmt-custom for building a separate shared lib the needs GMT.

PaulWessel:
For Linux/macoS the mbsystem installer builds the shared library, so this is a Win only thing.  You should not need any Cmake stuff really, and just have a batch script or similar that does its thing?

PaulWessel:
BTW, I need to so some work on gmt-custom it seems.

seisman:
> We removed something related to pthread in CMake scripts didn't we? I have tons of these
> 
> ```
> c:\progs_cygw\mb-system\include\iowrap-posix.h(81): fatal error C1083: Cannot open include file: 'pthread.h': No such file or directory
> ```

Yes, pthread was removed in #1937. If mbsystem needs pthread, you probably need to add the cmake snippet back to the CMakeLists.txt of mbsystm.



> What happens is that I have this
> 
> ```
> 	set (EXTRA_INCLUDE_GMTSUPPL "mb/suppl_add_mb.txt")
> ```
> 
> but now the use of `EXTRA_INCLUDE_GMTSUPPL` has been removed from the CMakeList and my MB build it broken.

We changed `EXTRA_INCLUDE_GMTSUPPL` to `EXTRA_MODULES_SUPP` in #2691.

PaulWessel:
I think the main thing is that any supplemental requirements shall go into the setup or CMakeLists.txt for the supplemental library, not GMT itself.

PaulWessel:
FYI, the gmt-custom was moved to git and renamed custom-supplements.  I will update it for GMT6 now.

joa-quim:
The MB Win build would be an act of pure masochism if I didn't need it for classes. Everything works to break it all the time.
It doesn't go into GMT. Only to my private build of GMT.

PaulWessel:
OK, I will be updating custom-supplements, hopefully with @seisman's help, to make it simpler to create custom supplements.  Not clear to me if this will help you or not, though.
OK for me to merge this PR?

joa-quim:
No, it's not to build a custom supplement that I'm after but I made it build again the MB+GMT. Will need to test it but not now.

Meanwhile the *grdmath* issue became a little worse. This now crashes Julia
```
G = gmt("grdmath -R60/95/6/24 -I0.1 ? LDIST", mat2ds([73 24; 89 6]))
```

joa-quim:
But go ahead and merge this. We'll deal with the other issue after.

PaulWessel:
Does work for gmt mex though.  Perhaps your *.jl code needs an update since I made changes in gmt.h and gmt_resources.h, including changing the GMT_STR16 to GMT_VF_LEN 32?

joa-quim:
Don't know what but this branch broke GMT.jl

PaulWessel:
I am pretty sure you have told me in the past that you need to make Julia changes if I change anything in gmt.h and gmt_resources.h.  Did you do those?

joa-quim:
Found it. It was the change in ``GMT_RESOURCE``.

PR #2863, opened 2020-03-08
Improve gmt_make_module_src.sh

PaulWessel:
A bit more documentation since this is the function that external supplements developers need to run to create the required module_lib C and H glue.  I know allow a custom tag to be given instead of an error, and I only build the RST stubs for core and supplements.  This updated script needs to replace the out-of-date gmt_make_custom_code.sh distributed with GMT in the share/tools directory.

@seisman, could you help me to have src/gmt_make_module_src.sh be properly installed in share/tools? I have removed the gmt_make_custom_code.sh from the share/tools and removed it from CMakeList.txt but there needs to be an install command somewhere (in src) to place it in the right dir with executable permission.


PaulWessel:
Works fine for GMT as before. Please approve.

Issue #2875, opened 2020-03-09
Support asymmetrical grid ticks as in navigational charts


PaulWessel:
**Description of the desired feature**

Currently GMT plots symmetric crosses.  For navigational charts it is customary to for meridians to have asymmetrical ticks that point away from Greenwich and for parallels to have have ticks pointing away from Equator.  That way, users know from tick orientation what hemisphere they are in.  See Discourse forum [discussion](https://forum.generic-mapping-tools.org/t/asymmetric-grid-tick-marks/291) for context.

I suggest this can be implemented by letting **MAP_GRID_CROSS_SIZE_PRIMARY**|**SECONDARY** accept a _negative_ size.  Then, we implement the above, else for positive sizes we do symmetric crosses.  For zero we instead do gridlines.

KristofKoch:
Wow, that was quick. I'll give it a try when I solved my problem with [installing GMT from source on MacOS](https://forum.generic-mapping-tools.org/t/install-gmt-from-source-on-macos-cmakelists-txt/292).

KristofKoch:
Already mentioned in #2877 but to keep things in order:

0°/180° longitude and 0° latitude remain symmetrical with all ticks extending on both sides of the grid line. See below example taken from ONC M-2:

![onc_m-2](https://user-images.githubusercontent.com/45796239/76698702-c396bd00-66ae-11ea-8958-7a6f01f9be70.jpg)

seisman:
Closed by #2877.

PR #2877, opened 2020-03-10
Let a signed grid cross size mean centered or asymmetrical grid ticks

PaulWessel:
See #2875 for context. Hope @KristofKoch can give it a test. Alternate giving a small x-grid interval and course y-interval to get the effect.  E.g.,

```
gmt psbasemap -R-5/5/-5/5 -JM6i -P -Bafg1 -K > t.ps
gmt psbasemap -R-5/5/-5/5 -JM6i -P -Bxg1 -Byg5m -O -K --MAP_GRID_CROSS_SIZE=-6p  >> t.ps
gmt psbasemap -R-5/5/-5/5 -JM6i -P -Byg1 -Bxg5m -O --MAP_GRID_CROSS_SIZE=-6p  >> t.ps
```


PaulWessel:
Hi @KristofKoch, any luck it testing this branch?

KristofKoch:
Hi @PaulWessel, please give me another 36h … I'm currently at work in my day job. Will give it a try as soon as I can.

PaulWessel:
No worries. At my day-job too!

PaulWessel:
Thinking about possible improvements to this.  Clearly, as implemented, it is assumed that there are gridlines present to support the ticks.  So in an ideal world this would be one set of -Bs and -Bp to do the course and fine ticks.  But your examples has three levels: course grid lines, finer longer ticks, and closely-spaced short ticks.  We have no way of doing three settings (yet).

PaulWessel:
One option is something like this script (which does not do what I want yet, but I will explain below):

```
gmt begin map
	gmt basemap  -Bpg2m -Bsg10m --MAP_GRID_CROSS_SIZE=-3p/-6p
gmt end show
```

Here we are saying that the longer, secondary ticks should occur every 10 minutes and the primary, shorter ticks should be placed every 2 minutes.  However, this is all under the assumption that these ticks are only drawn along the earlier 1x1 degree gridlines.  As it is now, we would just get lots of ticks every 10m and 2m.  A solution would be to remember the previous gridline interval (keep track of last gridline interval used) and then when cross sizes are _negative_ we use the previous _gridline_ interval with the specified _gridcross_ interval(s) to accomplish the desired plot.  In the process I would let **MAP_GRID_CROSS_SIZE** accept the syntax primary/secondary for cross sizes.  Currently it expects one item and sets both, but if given two separated by a slash we could assign the two separately.  This is a script that gives what I think you want (the hard way):

```
gmt begin map png
	gmt basemap  -Bpxg1 -Bpyg2m --MAP_GRID_CROSS_SIZE_PRIMARY=-3p -Bsyg1 -Bsxg2m --MAP_GRID_CROSS_SIZE_SECONDARY=-3p
	gmt basemap  -Bpxg1 -Bpyg10m --MAP_GRID_CROSS_SIZE_PRIMARY=-6p -Bsyg1 -Bsxg10m --MAP_GRID_CROSS_SIZE_SECONDARY=-6p
gmt end show
```

with result

map

KristofKoch:
Hi Paul,

that virus is wreaking havoc at work … but finally:

I like your proposed new syntax. Is there a way to get it to work for even more (many!) sets of tick marks?

Something like `-Bg1m/g5m/g10m/g30m --MAP_GRID_CROSS_SIZE=-2p/-3p/-6p/-9p` 

And as I'm already dreaming things up – is there a way to get tick marks extend both sides of the grid line (not a cross)?

I don't know if "=" is a good choice in terms of parsing but in my mind it would be something like "-" = one side, "=" = both sides of the grid line. So this:

    -Bg1m/g5m/g10m/g30m --MAP_GRID_CROSS_SIZE=-2p/-3p/=6p/=9p

would lead to something like this:

grids

where you have the single side ticks (1' & 5') mixed with double-sided tick marks (10' & 30'). 

The reason why I don't want a cross are tick marks at high latitudes with Lambert conic conformal or polar stereographic projections. Here the latitude grid lines bend so much that the crosses become visible.

KristofKoch:
And while you are on it – any chance to have the tick marks 0°/180° longitude and 0° latitude extend both ways?

Here the example for 0° longitude and 0° latitude taken from ONC M-2:
onc_m-2

And the example for 180° longitude and 0° latitude taken from ONC M-16:
onc_m-16

I guess there are users who are more interested in the looks than in the actual usability case so maybe they might not like to have them extend both ways. Another option?

KristofKoch:
And another possible improvement … maybe there is a chance to specify pen thickness on a per tick mark basis as well? The 1' ticks having a thinner pen than the 30' ticks? 

Maybe I can help by writing the documentation for those features?

PaulWessel:
Hi @KristofKoch, I will think about this. I don't think we will allow a 3rd tick/pen in one go, so since these maps count for 0.001% of GMT maps it is OK if you have to run a few commands to get it - you can always write a function of alias for something like that if you need it often.

KristofKoch:
Hi @PaulWessel, I agree with you that more than two ticks are rarely needed and can easily achieved if desired. None the less I take pride in being responsible for 0.001% of all GMT maps ;-)

PaulWessel:
Hi @KristofKoch. A small update: If grid cross sizes are specified with a leading + then they become symmetrical ticks, a leading - still means asymmetrical ticks, and no leading sign means regular grid cross.  Please give this a try and see if you can demonstrate to me how it may fail for high latitudes, then post a short basemap script here so I can test further.  I suspect I need to do more work to cover the case you mention.

KristofKoch:
Hi @PaulWessel, thank you for working on this. I gave 6.1.0_7c0ba1d_2020.03.19 a test. I found a few issues. For ease I just used lon or lat ticks. 

**Pen stroke:**

It gets funny if I use a pen thicker than the grid lines:

```
gmt begin MAP_GRID_PEN_PRIMARY
  gmt basemap  -Bpxg1 -Bpyg5m --MAP_GRID_CROSS_SIZE_PRIMARY=-3p --MAP_GRID_PEN_PRIMARY=3p 
gmt end show
```

MAP_GRID_PEN_PRIMARY

**High latitudes:**

There seems to be no difference between `--MAP_GRID_CROSS_SIZE_PRIMARY=+100p` (which are supposedly symmetric ticks) and `--MAP_GRID_CROSS_SIZE_PRIMARY=100p` (which are supposed to be crosses):

```
gmt begin symmetric
  gmt basemap -R-10/10/68/72 -JS0/90/60/6i  -Bafg10
  gmt basemap  -Bpyg5 -Bpxg5d --MAP_GRID_CROSS_SIZE_PRIMARY=+100p --MAP_GRID_PEN_PRIMARY=red
gmt end show
```

symmetric

```
gmt begin cross
  gmt basemap -R-10/10/68/72 -JS0/90/60/6i  -Bafg10
  gmt basemap  -Bpyg5 -Bpxg5d --MAP_GRID_CROSS_SIZE_PRIMARY=100p --MAP_GRID_PEN_PRIMARY=red
gmt end show
```

cross

There is a good chance I'm using the `-B` syntax wrong. How do I get rid of the lon ticks which are peaking out of the lat grid line?

Looking at asymmetric ticks I have a similar problem as with the two above. There still is a lon tick which escapes the grid line:

```
gmt begin asymmetric
  gmt basemap -R-10/10/68/72 -JS0/90/60/6i  -Bafg10
  gmt basemap  -Bpyg5 -Bpxg5d --MAP_GRID_CROSS_SIZE_PRIMARY=-100p --MAP_GRID_PEN_PRIMARY=red
gmt end show
```

asymmetric

Is it me again not using `-B` properly?

PaulWessel:
Thanks @KristofKoch I have made many changes; this feature only applies to modern mode:

1. Gridline ticks are considered a form of embellishment on gridlines.  Thus, you must first lay down the single (primary) set of gridlines you wish to embellish.
2. Next, you can specify the actual grid tick spacing you want and we will use the saved gridline spacing to do the placement correctly.
3. I now draw ticks, not crosses.

As an example, this test:

```
gmt begin asymm_grid ps
	gmt basemap -R-2/2/-2/2 -JM6i -Bafg1 -B+f
	gmt basemap -Bpg10m -Bsg2m --MAP_GRID_CROSS_SIZE_PRIMARY=+8p --MAP_GRID_CROSS_SIZE_SECONDARY=-4p --MAP_GRID_PEN_SECONDARY=default,black
gmt end show
```

produces the plot below.  So one call for coarse gridlines, one call to do two sets of ticks.

asymm_grid

Please give it a spin for other latitudes and projections.

KristofKoch:
Hi @PaulWessel these are big improvements. Awesome! I gave 6.1.0_db366d6_2020.03.20 a test and in generally behaves as expected with the "new" syntax. However a small problem remains: At high latitudes the ticks still peak out from the bent latitude grid lines. The placing is logically sound. I want a tick every 1°, I get one. Exaggerated example:

```
gmt begin asymmetric_high_lat
  gmt basemap -R-10/10/78/87 -JS0/90/60/6i  -Bafg5
  gmt basemap -Bpg1d --MAP_GRID_CROSS_SIZE_PRIMARY=-100p --MAP_GRID_PEN_PRIMARY=red
gmt end show
```

asymmetric_high_lat

They even make the step down at N85° in lon density. Great! But unfortunately some ugliness when plotted at a lat/lon grid line junction. Maybe there is an easy way to omit them there?

And another thing – any chance to get the functionality in classic mode as well? My workflow relies heavily on just generating those layers with updated data and then sandwich it together with all the other preproduced and stored layers in a second step to keep generation time low.

PaulWessel:
Hi @KristofKoch.  It makes sense not to draw the ticks exactly at the gridlines since the latter are curved and the ticks are straight.  I think that is what you mean.  I can probably fix that.

As for classic: It cannot do the remembering.  However, you should be able to do it the way we started off with, where you alternate course and detailed ticks.

KristofKoch:
Hi @PaulWessel – good evening to Hawai'i. That's exactly what I meant with my convoluted sentence above. Don't draw the ticks if placed exactly at the grid lines.

PaulWessel:
OK, should not place ticks on gridlines now.  I have no new answer regarding classic vs modern for the memory yet.

PaulWessel:
OK @KristofKoch, should now work for classic mode too, but since it writes the temp file to /tmp with a fixed name you cannot run multiple jobs in parallel (unless you use classic isolation mode).

KristofKoch:
That's really awesome. Gave 6.1.0_45ff494_2020.03.20 a test and it works beautifully. Thank you very much, @PaulWessel! 

I learnt the hard way that isolation mode is the way to go when running classic mode jobs in parallel … don't ask.

PaulWessel:
Hi @seisman, although @joa-quim approved this, that was a while ago and things have changed quite a bit (for the better I think).

KristofKoch:
Well, I did find a problem while experimenting with classic and modern mode … 

```
gmt psbasemap \
  -Rg \
  -JG8.1703/49.9816/15c \
  -Bg -P -K \
  > earth_classic.ps

gmt psbasemap \
  -R -J \
  -Bpg5d -Bsg1d \
  --MAP_GRID_CROSS_SIZE_PRIMARY=+5p \
  --MAP_GRID_CROSS_SIZE_SECONDARY=-3p \
  --MAP_GRID_PEN_PRIMARY=default,red \
  --MAP_GRID_PEN_SECONDARY=default,blue \
  -O \
  >> earth_classic.ps
```

```
gmt begin earth_modern
  gmt basemap -Rg -JG8.1703/49.9816/15c -Bg
  gmt basemap -Bpg5d -Bsg1d --MAP_GRID_CROSS_SIZE_PRIMARY=+5p --MAP_GRID_CROSS_SIZE_SECONDARY=-3p --MAP_GRID_PEN_PRIMARY=default,red --MAP_GRID_PEN_SECONDARY=default,blue
gmt end show
```

crop_globe

You are looking at a crop from the North Pole. Same for both classic and modern mode. The ticks run without gridline from N85° to N90°.

KristofKoch:
@PaulWessel you are pushing commits faster than I can build the GMT versions ;) I ran above test in 6.1.0_45ff494_2020.03.20 but problem persists in  6.1.0_9b1b886_2020.03.20.

PaulWessel:
Forgot about the polar circles.  Will fix later today. Thanks for the testing!

KristofKoch:
Just noticing it myself in my example above: the latitude ticks are pointing towards the Equator instead away from it. Longitude ticks are correct. 

earth_modern

PaulWessel:
Will fix that too.

KristofKoch:
Same tick-facing-wrong-way-problem slipped me through further up in the `-JS` example https://github.com/GenericMappingTools/gmt/pull/2877#issuecomment-602005297

KristofKoch:
@PaulWessel thank you for your efforts. I really appreciate it. But please don't forget to enjoy your weekend!

PaulWessel:
About to go on a hike enjoying some vitamin D provided by sun.  I fixed the tick orientation and the polar caps.  See what you think.  I could all for N-S ticks along the polar cap - that is currently not happening since it is not at a latitude that is a multiple of the parallel spacing.  Let me know.

KristofKoch:
Thank you, @PaulWessel. I hope you had a great time hiking. I gave 634a89a a test and the problem with the polar caps and the direction of the N-S ticks has vanished in my testing. 

Is there a chance to get symmetric ticks at 180° longitude similar to 0° longitude as seen in this crop of the current FAA Sectional Chart? 

180deg-lon

Concerning on the behavior at the polar caps: I'm not quite sure what you meant by

> I could all for N-S ticks along the polar cap […]

Maybe it's my lack of English or did something got mangled by autocorrect? 

My thoughts for discussion about the behavior at the polar caps:

* **Show E-W tick marks at the four cardinal grid lines - yes or no?** I lean towards yes as there is not much distortion in latitude as there is in longitude. But I see a strong point in leaving them out as the whole polar cap thing serves as a simplified depiction for the pole area.
* **Show N-S ticks at the "polar circle" regardless if it is a multiple of parallel spacing?** Mh, a tricky decision in my  eyes.
  * if shown the adjacent part south of the circle can still be used for navigation.
  * if shown it can get very crowded up to the point which makes it totally unusable. Hard to balance "tick density" for large areas while maintaining usability. 
  * if not shown it makes for a much cleaner look and goes in sync with the heavy simplification thought for the polar area.
  * maybe show only the primary ticks? Cleans up the clutter by too closely spaced secondary ticks while still maintaining usability?
  * make the behavior another option? (I hear you cursing me)

What are your thoughts?

KristofKoch:
66129f4 does what it promises. Thanks!

KristofKoch:
A thought on the polar cap issue:

My idea with showing only primary ticks was under the assumption that the primary ticks are those with the wider spacing – a look in the docs showed that the convention seems to be the other way around. I think you understand what I mean – just show the wider spaced ticks.

PaulWessel:
Will get to this after family FaceTime with remote kids.

PaulWessel:
I will tick the polar cap parallel(s).  For the coarse gridlines toward the poles I think we only do the primary grid ticks, but I will experiment.  The grid line spacing up there is controlled by MAP_POLAR_CAP.

KristofKoch:
The full set of ticks at the cap parallels? Or a reduced (primary ticks only) set?

PaulWessel:
Full set of ticks on the polar cap - like the other parallels, but only coarser on the meridians.  However, perhaps it may make more sense to do just coarser ticks on both.

I write courser because primary and secondary labels is a bit backwards in the sense that for time axis the smallest spacing is called primary and the coarser spacing is called secondary.  So above, we may drop to just secondary since space is getting limited.  I could also automatically pick the coarsest of the two to avoid any confusion.

KristofKoch:
I'm not sure if there is a "perfect" solution for the cap parallels  – If I look at the full globe (`-JG`) example above the full set would look ugly as it would be only a thick blue ring. Using `-JS` the whole thing might look completely different. 

Thank you for clarification where the primary and secondary comes from … I was quite a bit surprised to learn this in the docs. I agree for geo purposes it "feels backwards".

KristofKoch:
Toying around with this a bit further I agree with you that it may make most sense to do just the coarser ticks on both. Because if there is enough space for the denser ticks I can as well move `MAP_POLAR_CAP` further towards the pole as there is no need for the simplification at the current latitude. 

The "auto-pick-coarsest-ticks" would be a nice thing to have as well. Might also suppress some questions in the forums.

PaulWessel:
Need feedback from @joa-quim and @seisman on this:

The concept of primary and secondary annotations/ticks/gridlines originated with time-axes where it is often required to annotate more than one unit.  In these cases, we use the **primary** term to mean the annotation closest to the axis and the **secondary** term to be the one further out.  For time axes this always meant that primary intervals were _smaller_ than secondary intervals:

GMT_-B_time1

For geographic maps with two sets of annotations (rarely used) we actually do the same:

GMT_-B_geo_2

Thus, for the case at hand (grid ticks) we must stay with the same scheme.  Hence, for grid-ticks associated with polar caps I will use the coarser secondary gridline spacing, _assuming_ it is the coarser.  However, @KristofKoch example above switched the two and he used the coarser as primary and the detailed as secondary.  This can lead to confusion. Possible remedies:

1. We can impose the syntax rule that primary interval < secondary interval and give an error if that is not followed.
2. We can simply assume the user did what we say and plot normally.  In @KristofKoch example he may be surprised to find his smaller interval being used within the polar caps.
3. We can check which interval is the smallest and if backwards we simply use an index array PS[2] that by default is {0,1} but can be changed to {1,0} in this case and we use the index array to access the two settings for intervals, pen, etc. I.e., in a loop k = 1,2 over primary, secondary we use the index PS[k] to get pen or interval instead of the current [k].

I think option 3 is the best since it follows our intention and also handles any user mixup.  We may wish to give a warning when this happens.  Comments?

joa-quim:
> The concept of primary and secondary annotations/ticks/gridlines originated with time-axes

Historical correction. This concept was born when I, on board of French ships, saw their maps with double annotations made with their *Caraibes* Sun software and started bugging you to have the same in GMT (much like Kristof now).

I have not been following this matter closely (busy with GEOS) but yes, option 3 is the more versatile one.

PaulWessel:
Yes, that is true for maps.  I don't remember if geo came before time.  Time happened around 2004 (for GMT 4I think).
Looking at implementing option 3.

joa-quim:
Made a 1 month cruise in 1998 on board of *l'Atalant*. Probably from around that time.

PaulWessel:
Hi @KristofKoch, please check this out once more.  The changes are:

1. If you give primary as the larger interval we will swap with secondary, since GMT logic dictates this order.  This switches the **-Bp** and **-Bs** settings but does not change other things.  You get a warning and probably the wrong pens and tick settings.
2. The secondary tick information is used on the polar cap parallels and the polar cap meridians.

KristofKoch:
Wow – @PaulWessel that's impressive work. Thank you very much! I feel it is about time to send another beer (maybe a keg would be more adequate) your way for all the work I'm causing. 

1. 1538dd2 does as you say: `-Bp` and `-Bs` switch but the pen settings remain the same and I get a warning message. 
   * Im not sure if this causes havoc with some existing scripts as it wasn't a problem before this implementation when `-Bp` was coarser than `-Bs`. 
   * For me the warning message leaves a bit to be desired – maybe you can add the hint that secondary should be coarser than primary and that pens are not switched? Or switch the pen definition as well to get an expected result? Switching pens as well would maintain compatibility with existing scripts with erroneous `-B` usage which worked until today. I know what we are talking about here but I think if I get this message in a couple of years with a plot where the pens are wrong I'll have little to no idea where to start looking.
2. Works as expected. Awesome!

PaulWessel:
I am trying to avoid having to swap ~10 different parameters that have _PRIMARY and _SECONDARY in them - see gmt.conf.rst for all those parameters (pens, tick lengths, colors, etc., etc.).
However, I am OK with just warning and explain what GMT expects but not swap anything.  That will let old scripts run but the new polar tick stuff will be wrong so you will want to change new scripts.

KristofKoch:
@PaulWessel I see your point. I think your proposal is the more sensible way. 

Warn and explain always - even if polar cap is not involved - but don't swap anything. Thus old scripts keep working (and get a hint on how to improve them) and there should no new scripts besides my few test scripts yet. So win-win?

PaulWessel:
Hi @KristofKoch, I have improved the documentation for -B in usage and rst files to clarify that primary sets the more frequent annotations while secondary sets the less frequent intervals.  This was largely missing except in the context of time axes.

I have also changed my test above to just warn and explain, with no forced swapping.  This should alley fears of breaking backwards compatibility.  I think this PR is good to merge now, yes?

PaulWessel:
I am answering my question: Yes good to merge. Happening now.

KristofKoch:
Good morning from the other side of the blue marble, looks good to me as well. @PaulWessel , thank you for your efforts! This makes my life a lot easier.

PR #2881, opened 2020-03-10
Add GMT_MSG_NOTICE verbosity level for special notices

PaulWessel:
These are messages that the developers need to convey regardless of verbosity (except for **-Vq**).


seisman:
@PaulWessel You need to run `gmt_make_enum_dicts.sh` to update the header files.

PR #2883, opened 2020-03-10
cmake: Move GMTMEX installtion to admin/add_windows_cpack.txt

seisman:
GMTMEX is only used when building Windows binaries.
It's better to move them to the admin/add_windows_cpaak.txt file.


seisman:
@joa-quim Are you OK with the PR.

joa-quim:
OK. Look harmless.

PR #2884, opened 2020-03-10
cmake: Eliminate unused variable GMT_OPENMP

seisman:
In `cmake/ConfigUserAdvanced.cmake`, users can set `GMT_ENABLE_OPENMP`
to enable GMT OpenMP support. But when it's set OFF, the cmake
message says "User variable GMT_OPENMP not set: disabling OpenMP."

Variable **GMT_OPENMP** is defined in src/CMakeLists.txt`, but not used
elsewhere. This PR eliminates the variable and fix the cmake message.


seisman:
ping @PaulWessel

PR #2902, opened 2020-03-14
Update psconvert with new ghostscript option

PaulWessel:
We must now pass the new -dALLOWPSTRANSPARENCY option when transparency is required.  Has no effect on gs versions < 9.51; in fact, until bugs in 9.51 is fixed this option has no effect whatsoever!

### DIFF
--- a/src/psconvert.c
+++ b/src/psconvert.c
@@ -1507,7 +1507,7 @@ int GMT_psconvert (void *V_API, int mode, void *args) {
 #endif
 	/* Define the 4 different sets of GS parameters for PDF or rasters, before and after SCANCONVERTERTYPE=2 added in 9.21 */
 	/* 2018=09-18 [PW]: Removed -DSAFER since it now prevents using the Adobe transparency extensions.  All our test pass with no -DSAFER so I have removed it */
-	/* 2020-03-14 [PW]: Added -dALLOWPSTRANSPARENCY siince 9.51 onwards requires it to do transparency.  It is simply ignored by older gs versions */
+	/* 2020-03-14 [PW]: Added -dALLOWPSTRANSPARENCY since 9.51 onwards requires it to do transparency.  It is simply ignored by older gs versions */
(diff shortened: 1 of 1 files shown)


joa-quim:
Not sure they considered this transparency issue a bug. And, does this new option have the potential to avoid the trick of having to go via pdf to get a transparent png?

PaulWessel:
Chris apologized for not listing this issue in their "Breaking compatibility" release statement.  But there is independently a bug in pdfwrite in 9.51 they are working on.

PaulWessel:
We will check once 9.51 has been fixed to see if -dALLOWPSTRANSPARENCY might take us directly from PS to PNG.  But I suspect not since it is specifically a PDF feature.

seisman:
I downloaded [ghostscript's example file](http://git.ghostscript.com/?p=ghostpdl.git;a=blob;f=examples/transparency_example.ps) and tried to convert it to PNG and PDF with the following commands (gs 9.51):
```
gs -dNOSAFER -dBATCH -dNOPAUSE -sDEVICE=png16m -dGraphicsAlphaBits=4 -dALLOWPSTRANSPARENCY -sOutputFile=examples_transparency_example.png examples_transparency_example.ps
gs -dNOSAFER -dBATCH -dNOPAUSE -sDEVICE=pdfwrite -dALLOWPSTRANSPARENCY -sOutputFile=examples_transparency_example.pdf examples_transparency_example.ps
```
I can see transparency in the PNG file but not PDF file. This is a known gs bug for pdfwrite.

I also tried the same commands to `admin/PS_orig_for_gs.ps`, but I can't see transparency in both PNG and PDF, possibly because gs has their own way to set transparency, and `-dALLOWPSTRANSPARENCY` only works if GMT also following their rules.

PaulWessel:
I guess we may need to research the PDF 1.4 specifications.  They are using odd things like .pushpdf14devicefilter and .setopacityalpha, etc.  We would need to study [their docs](https://www.ghostscript.com/doc/current/Language.htm#Additional_operators) and replace the PostScript commands our _PSL_settransparency_ and _PSL_settransparencymode_ functions produce.  If this allows transparent PNG directly from PS then we should do this instead of wasting CPU on making an intermediary PDF file.

PaulWessel:
We actually do deal a bit with this.  Our definition of PSL_trans is

```
/PSL_transp {						% Set transparency
  /.setopacityalpha where {pop .setblendmode .setopacityalpha}{		% Using ghostscript
  /pdfmark where {pop [ /BM exch /CA exch dup /ca exch /SetTransparency pdfmark}	% Or Adobe
  {pop pop} ifelse} ifelse				% Or skip if neither are supported
```
}!

So depending on what tools is processing our PS (Adobe Distiller or ghostscript) we try to issue the right transparency lingo.  But we are not adding some of those things mentioned above.

seisman:
I've opened a new issue in #2911 to better track it.

Issue #2897, opened 2020-03-14
Let tests detect failure of ghostscript conversion


PaulWessel:
**Description of the desired feature**

We have suffered through a long series of ghostscript updates and bugs the last couple of years.  Apparently, the **ghostscript** team does not have a good system of testing that their changes are not breaking things that work.  Recently, they released version 9.51 and at first go it seems fine, but then we learned that the transparency feature is broken.  Our tests do not actually look for that (it compares an original and a newly created PostScript plot via conversion to PNG and pixel-wise comparison, but that conversion does not honor transparency anyway).

Since we know when **ghostscript** gets updated, there is no reason to include this type of check in our standard tests. However, it would help prevent the release of a broken **ghostscript**.  Perhaps a simple script in admin we can run as needed that will create two identical plots except one has transparency, then convert those to PNG with psconvert and then use gmt compare?

seisman:
It sounds a good idea to me.

seisman:
Closed by #2905.

Issue #2903, opened 2020-03-14
Check status of ghostscript before releasing 6.1.0


PaulWessel:
Once again a **ghostscript** release (9.51) comes with bugs affecting GMT, in this case a loss of transparency.  The problem has already been [reported](https://bugs.ghostscript.com/show_bug.cgi?id=702219) and hopefully distributions will get a patch or there will be a 9.52 soon.  Whatever the situation is, when GMT 6.1.0 is released we may need to warn against 9.51 or add a warning in **psconvert** when users request transparency.

A secondary issue in 9.51 was the unannounced breaking of backwards compatibility by now requiring the option [-dALLOWPSTRANAPARENCY](https://www.ghostscript.com/doc/9.51/Use.htm) to activate the PDF transparency machinery.  GMT enabled this in #2902 but due to the above bug it has no effect yet.

joa-quim:
I can see transparency directly on the PS files when using the, now abandoned by Artifex, gsview6.0.

PaulWessel:
(.52 has been released but apparently not addressing the bug in transparency reported above.

PaulWessel:
Update: Per Chris Liddel, next gs 9.53 is slated for September.  They do one very 6 months apparently.  So 9.50 for a while for us.

remkos:
Have you tried the solution that is posted on the ghostscript website, i.e. using -dALLOWPSTRANSPARENCY ?
See: https://www.ghostscript.com/doc/9.52/News.htm (Section 1.1.)
And: https://www.ghostscript.com/doc/9.52/Use.htm#ALLOWPSTRANSPARENCY

Although, reading https://www.ghostscript.com/doc/9.52/Language.htm#Transparency, I'm not so certain this will do it.

PaulWessel:
Thanks @remkos for checking.  We will probably need to look at the transparency setup more carefully going forward, but plate is full right now....

seisman:
FYI, ghostscript 9.53 is released.

seisman:
One of the biggest changes is that the ghostscript version is 9.53.0, not 9.53. It may break some codes.

PaulWessel:
I think we only check gs --version output in psconvert where we take the string returned and do this:

`		int n = sscanf (cmd, "%d.%d", &gsVersion.major, &gsVersion.minor);`

[...]

PR #2905, opened 2020-03-14
Add amin/gs_check.sh to test new ghostscript versions for compliance

PaulWessel:
Given that ghostscript versions often are released with bugs (see #2903), we add this script so that prior to a new GMT release we can run this script with the very latest ghostscript version and see if basic features like fonts, patterns, and transparency are still converted correctly.


PaulWessel:
This PR closes #2897.

admin/gs_check.sh
```diff
@@ -0,0 +1,57 @@
...
+
+if [ ! -d cmake ]; then
+	echo "Must be run from top-level gmt directory"
+	exit 1
+fi
+if [ ! -d build ]; then
+	mkdir -p build
+fi
+# Do the work in the build directory
+cd build
```
seisman: Echo something here so that others know that the script outputs everything in the build directory.

PaulWessel: OK will do

PaulWessel:
None of the checks seem to finish the last few days...

PR #2924, opened 2020-03-16
Better documentation, checking for einterpolators

PaulWessel:
Mostly improvements by comments and usage messages, plus a new convenience function to combine mode and type into a single value used in a few places.

Outstanding question: If a user has a separate _lon_, _lat_[, _distance_] table and wishes to get a vertical sliced grid, she is out of luck: She must use **-E** to set that profile up.  Might we relax that and allow **-E**_table_ as long as we check that the file is equidistant?  We can accept just _lon_, _lat_ table as well and add the distances if need be.

### DIFF
--- a/src/gmt_prototypes.h
+++ b/src/gmt_prototypes.h
@@ -351,6 +351,7 @@ EXTERN_MSC bool gmt_this_alloc_level (struct GMT_CTRL *GMT, unsigned int alloc_l
 
 /* gmt_support.c: */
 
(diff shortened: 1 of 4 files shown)


seisman:
To my understanding, currently grdinterpolate -E only supports cross-sections along great circle, parallels, and meridians, right? If that's true, then I can see the support of *lon, lat, distance* table is useful for some users.

PaulWessel:
Yes, it would be sad if we cannot support that.  I will implement this.

PR #2926, opened 2020-03-16
No longer check for sphinx version

PaulWessel:
Updated template supports v1 and v2.


seisman:
I forgot to mention that, although PR #2925 fix the stylings of option, it also introduces some tiny styling issues of other elements. So we may still need to use sphinx 1.8.5 when making a release.

PaulWessel:
OK, closing this branch. How to delete it?

PR #2927, opened 2020-03-16
Let grdinterpolate -E accept a user profile file

PaulWessel:
Instead of having always to create the profile from coordinates via **-E**_lines_, allow users to supply a single segment with equidistant spacing via **-E**_file_ as well.  I added a new test (_slices_file.sh_) to make sure it works.

### DIFF
--- a/test/grdinterpolate/slices_file.sh
+++ b/test/grdinterpolate/slices_file.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+# Test grdinterpolate slicing along equator through a 3-D grid
+# Getting the file directly from IRIS.
+# This script tests ability to supply our own profile file
+
+gmt begin slices_file ps
+	gmt set PROJ_ELLIPSOID sphere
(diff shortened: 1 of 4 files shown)


test/grdinterpolate/slices_file.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# Test grdinterpolate slicing along equator through a 3-D grid
+# Getting the file directly from IRIS.
+# This script tests ability to supply our own profile file
+
+gmt begin slices_file ps
+	gmt set PROJ_ELLIPSOID sphere
+	curl -s http://ds.iris.edu/spudservice/data/17996723 -o S362ANI_kmps.nc
+	gmt math -T0/180/0.5 0 = line.txt
+	gmt grdinterpolate S362ANI_kmps.nc?vs -Eline.txt -T30/2890/10 -Gvs.nc
```
seisman: Where is the line.txt file?

seisman: Oh, I found it.

PR #2939, opened 2020-03-17
Use bold font to call attention to Note:

PaulWessel:
Being consistent across the docs, and also capitalize the following sentence.


doc/rst/source/gmtregress.rst
```diff
@@ -175,7 +175,7 @@ Optional Arguments
 
 .. include:: explain_array.rst_
 
-Note:
+**Note**:
```
seisman: Not for section titles.

seisman:
```suggestion
Note:
```

PaulWessel: OK, taht is the only one - I checked

PR #2955, opened 2020-03-22
Add DOT as dot-product operator in grdmath

PaulWessel:
Take 2-D or 3-D dot products between vector(s) (A,B) and the grid node coordinates.  Added earth_shade.sh test script as well and fixed wrong number of operators in the documentation.

### DIFF
--- a/src/grdmath.c
+++ b/src/grdmath.c
@@ -27,7 +27,7 @@
 
 /* Notes: To add a new operator:
  * 1) Just add one more entry at the end of the array of operator names at top of GMT_grdmath function.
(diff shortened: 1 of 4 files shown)


src/grdmath.c
```diff
@@ -25,6 +25,15 @@
  * Version:	6 API
  */
 
+/* Notes: To add a new operator:
+ * 1) Just add one more entry at the end of the array of operator names at top of GMT_grdmath function.
+ * 2) Add more more entry at the end with the specifics in init_operators in grdmath_init function.
```
seisman:
```suggestion
 * 2) Add one more entry at the end with the specifics in init_operators in grdmath_init function.
```

PR #2956, opened 2020-03-22
Warn if remote download is turned off

### LINKED ISSUE #2954: GMT_DATA_SERVER_LIMIT=0 doesn't work?
The documentation says remote file download can be turned off by setting GMT_DATA_SERVER_LIMIT to 0. I want to see what will happen if remote file download is turned off.

Here are the commands I use:
```
# remove the cached file
rm ~/.gmt/cache/ridge_07.txt

# turn off remote file download
gmt set GMT_DATA_SERVER_LIMIT 0

# see what will happen if I want to use a remote file that doesn't exist locally
gmt which -Gc @ridge_07.txt
```

This is my output:
```
/Users/seisman/.gmt/cache/ridge_07.txt
```

I expect to see a warning like "remote file cann't be downloaded because ..." or similar.

PaulWessel:
This value or no download should both result in a message when people try to download files if they do not exist in cache.  Closes #2954.

### DIFF
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -402,7 +402,8 @@ Three classes of files are given special treatment in GMT.
    the default setting for :term:`GMT_DATA_SERVER`.  Alternatively, configure the CMake
    parameter GMT_DATA_SERVER at compile time.
 #. If your Internet connection is slow or nonexistent (e.g., on a plane) you can also
-   set the size of the largest datafile to download via :term:`GMT_DATA_SERVER_LIMIT` to be 0.
+   limit the size of the largest datafile to download via :term:`GMT_DATA_SERVER_LIMIT` or
+   you can temporarily turn off such downloads by setting :term:`GMT_AUTO_DOWNLOAD` off.
 
 The user cache (:term:`DIR_CACHE`) and all its contents can be cleared any time
 via the command **gmt clear cache**, while the server directory with downloaded data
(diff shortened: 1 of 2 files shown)


seisman:
I can't download any remote files now, even though I don't change the default GMT_DATA_SERVER_LIMIT value.

PaulWessel:
Sorry, I confused myself.  0 means unlimited, so gmt set GMT_DATA_SERVER_LIMIT unlimited or gmt set GMT_DATA_SERVER_LIMIT 0 is the same thing.  To turn off download, use gmt set GMT_AUTO_DOWNLOAD off.

PaulWessel:
Changed title since I did add a new message.

seisman:
There are two places in gmt_remote.c says:
```
You can turn remote file download off by setting GMT_DATA_SERVER_LIMIT = 0
```

seisman:
The cookbook/feature.rst says
```
 #. If your Internet connection is slow or nonexistent (e.g., on a plane) you can also
    set the size of the largest datafile to download via :term:`GMT_DATA_SERVER_LIMIT` to be 0.
```

seisman:
> There are two places in gmt_remote.c says:
> 
> ```
> You can turn remote file download off by setting GMT_DATA_SERVER_LIMIT = 0
> ```

Don't forget to update this.

PaulWessel:
Things were written before they were implemented...

seisman:
```
gmt begin map
gmt set GMT_AUTO_DOWNLOAD off
cat > t.dat << EOF
1 1
2 2
EOF
gmt plot t.dat -R0/10/0/10 -JX10c -Baf -Sc0.2c
gmt end show
```
I'm getting the following errors/warnings even though I don't use any remote files.
```
plot [ERROR]: Remote download is currently deactivated
plot [ERROR]: Remote download is currently deactivated
```

PR #2958, opened 2020-03-22
Prevent resampling of curves when Bezier spline is selected

### LINKED ISSUE #2957: Requesting Bezier spline in psxy is defeated by default line resampling
For background, see [this case](https://forum.generic-mapping-tools.org/t/s-for-spline-doesnt-work/310/2).

The problem is _not_ that the **+s** spline modifier for the pen setting is not allowed for geographic projections.  The problem is that for sparse geographic coordinates (like that of the user) we first interpolate along great circles.  Hence, by the time the Bezier function is called we already have finely resampled the curve and the spline cannot do anything different than follow the points.  This can be defeated by using **-A** of course.

There are two solutions:

1. Recognize that **+s** means the user wants a Bezier spline fit through the original data points and automatically turn on **-A** under the hood.
2. Add documentation to **psxy** to raise awareness of this issue so that users can choose to use **-A** or not or **+s** or not.

For backwards compatibility reasons as well as for flexibility I think we opt for better documentation.  OK with @joa-quim, @seisman and @KristofKoch?

PaulWessel:
We want to turn off the automatic line resampling when the user selects Bezier splines.  Closes #2957.

### DIFF
--- a/src/psxy.c
+++ b/src/psxy.c
@@ -1798,6 +1798,9 @@ int GMT_psxy (void *V_API, int mode, void *args) {
 		struct GMT_DATASET_HIDDEN *DH = NULL;
 		struct GMT_DATASEGMENT_HIDDEN *SH = NULL;
 
+		if (Ctrl->W.pen.mode == PSL_BEZIER && GMT->current.map.path_mode != GMT_LEAVE_PATH)	/* Turn of path-resampling if Bezier */
+			GMT->current.map.path_mode = GMT_LEAVE_PATH;
+
 		gmt_map_clip_on (GMT, GMT->session.no_rgb, 3);
 
 		if (Ctrl->L.anchor == PSXY_POL_SYMM_DEV) n_cols = 3;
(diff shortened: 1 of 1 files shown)


src/psxy.c
```diff
@@ -1798,6 +1798,9 @@ int GMT_psxy (void *V_API, int mode, void *args) {
 		struct GMT_DATASET_HIDDEN *DH = NULL;
 		struct GMT_DATASEGMENT_HIDDEN *SH = NULL;
 
+		if (Ctrl->W.pen.mode == PSL_BEZIER && GMT->current.map.path_mode != GMT_LEAVE_PATH)	/* Turn of path-resampling if Bezier */
```
seisman:
```suggestion
		if (Ctrl->W.pen.mode == PSL_BEZIER && GMT->current.map.path_mode != GMT_LEAVE_PATH)	/* Turn off path-resampling if Bezier */
```

PR #2961, opened 2020-03-23
First cut to use the GEOS library to compute buffer zones around polylines

joa-quim:
This PR is a first cut in using the GEOS library. This library is a common but not mandatory dependency of GDAL. It's also a damn poorly documented lib, which makes it very hard to learn how to use.

It needs to be build with ``-DHAVE_GEOS`` and add this to ``ConfigUser.make``.
```
	set (GEOS_INCLUDE_DIR "/path/to/geos/include/")
	set (GEOS_LIBRARY     "path/to/geos/lib/libgeos_c.lib")
```

I have added a *tmp* **-Z<dist>** option to gmtspatial to compute buffer zones around polylines. It works nicely but we need to create an infrastructure to accommodate all the power the use of this lib can bring to GMT.

Example usage with julia:

```
buf = gmtspatial([0.0 0; 1 1], Z=1);
plot(buf, aspect=:equal)
plot!([0.0 0; 1 1], savefig="buf.png", show=true)
``` 
buf


PaulWessel:
Do you know if this is Cartesian only or does it work on spherical surfaces?

joa-quim:
Almost (100% - eps) sure it's Cartesian only.

PaulWessel:
Fake name GEO(s) then.

joa-quim:
Probably it stands for GEOmetries as that is the basic building block.

joa-quim:
Should we commit this? It can do more but the buffer line is already a quite useful thing.

PaulWessel:
If you have not already, please merge master into that branch first and make sure no conflicts and run all tests, etc.  Then we can continue discussing after the first intro session coming up.

PaulWessel:
May need to spell out the endif in Cmake:

CMake Warning (dev) in src/CMakeLists.txt:
```
  A logical block opening on the line

    /Users/pwessel/GMTdev/gmt-dev/src/CMakeLists.txt:158 (if)

  closes on the line

    /Users/pwessel/GMTdev/gmt-dev/src/CMakeLists.txt:162 (endif)

  with mis-matching arguments.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

PaulWessel:
A few more things - do I need to set anything to try this branch?

```
[155/254] Building C object src/CMakeFiles/gmtlib.dir/gmtspatial.c.o
../src/gmtspatial.c:2184:11: warning: implicit declaration of function 'geos_methods' is invalid in C99 [-Wimplicit-function-declaration]
                error = geos_methods(GMT, D, Ctrl->Out.file, Ctrl->S.width, "buffer");
                        ^
../src/gmtspatial.c:2185:3: warning: implicit declaration of function 'finishGEOS' is invalid in C99 [-Wimplicit-function-declaration]
                finishGEOS();
                ^
../src/gmtspatial.c:2190:11: warning: implicit declaration of function 'geos_methods' is invalid in C99 [-Wimplicit-function-declaration]
                error = geos_methods(GMT, D, Ctrl->Out.file, 0, "centroid");
                        ^
../src/gmtspatial.c:2191:3: warning: implicit declaration of function 'finishGEOS' is invalid in C99 [-Wimplicit-function-declaration]
                finishGEOS();
                ^
```
and

```
ndefined symbols for architecture x86_64:
  "_geos_methods", referenced from:
      _GMT_gmtspatial in gmtspatial.c.o
ld: symbol(s) not found for architecture x86_64
```

PaulWessel:
It seems like it found the library though:

`-- Found GEOS: /opt/local/lib/libgeos_c.dylib (3.8.1)`

joa-quim:
It needs to be build with ``-DHAVE_GEOS`` and I need to find again how my example worked (probably after dinner)

joa-quim:
Because we need an automatic way of setting ``HAVE_GEOS`` when the lib is found. Since it is not set, the call to ``finishGEOS();`` lead to errors.

PaulWessel:
Looks like you dont have anything about GEOS in gmt_confit.h.in?  E.g., something like this that gets configured

#cmakedefine HAVE_GDAL

joa-quim:
Don't get what is happening. The build is failing on Windows apparently because it finds ``geos.lib`` (which is the c++ lib) instead if ``geos_c.lib``. But works on *nix, although I can't find the logs (always so damn complicated to find the logs).
It works for me locally, but I set in my ``ConfigUser.cmake``
```
        set (GEOS_INCLUDE_DIR "/path/to/geos/include/")
	set (GEOS_LIBRARY     "path/to/geos/lib/libgeos_c.lib")
```

cmake/modules/FindGEOS.cmake
```diff
@@ -0,0 +1,181 @@
...
+    FIND_PATH(GEOS_INCLUDE_DIR geos_c.h /usr/local/include /usr/include c:/msys/local/include)
+    FIND_LIBRARY(GEOS_LIBRARY NAMES geos_c PATHS /usr/local/lib /usr/lib c:/msys/local/lib)
+  ENDIF (MINGW)
+
+  IF (MSVC)
+    FIND_PATH(GEOS_INCLUDE_DIR geos_c.h HINTS ${GEOS_DIR} ${GEOS_INCLUDE_DIR} $ENV{LIB_DIR}/include $ENV{INCLUDE})
+    FIND_LIBRARY(GEOS_LIBRARY NAMES geos geos_c_i geos_c  HINTS ${GEOS_DIR} PATHS
+      "$ENV{LIB_DIR}/lib"
+      $ENV{LIB}
+      )
```
seisman, line 30: As I understand it, find_library will try to find `geos` first on Windows. Perhaps this is why the CI on Win finds goes instead of geos_c?

joa-quim: Yes, looks very likely, though the same logic should apply to unix.

seisman: Not the same. I think on unix it runs `geos-config` to get the library path.

joa-quim: OK, so please try this order ``geos_c_i geos_c geos``. The *_i* is something that some use to mean *import_lib* which is what those libs are.

seisman:
Now the CI on Win passes.

joa-quim:
Good, so ready to merge.

Issue #2912, opened 2020-03-15
Transparent handling of 3-D grids in GMT


PaulWessel:
**Description of the desired feature**

While GMT deals with situations where a 3-D grid is created, read, or written, it does so very awkwardly (via arrays of 2-D grids and external layer arrays) as we have no 3-D grid representation.  This issue discusses the most transparent and non-intrusive way for GMT as a whole to support the notion of a 3-D grid without lots of rewriting.  My _key assumption_ is that all layers share the same number of nodes and horizontal coordinates.  Here is my list of modifications that would not change the API (but instead add three new API functions).

1. The GMT_GRID_HEADER structure has two parts: One dictated by legacy for what is written to binary native grids, and one that is accessible internally.  The first cannot change, but in the hidden area we can add new parameters as needed.  Here, we will add `uint32_t n_layers, double *layer, unsigned int l_type, l_units[GMT_GRID_UNIT_LEN80].`  These parameters can hold the number of layers, the layer values (we must allow for non-equidistant layers), the layer value type (float, time, etc.), and the name/unit of that dimension. These could be added now with no effect on GMT whatsoever.

Grids are either created via _GMT_Create_Data_, read via _GMT_Read_Data_, or written via _GMT_Write_Data_.  Here are changes required under the hood without changing these API prototypes.

2. _GMT_Create_Data_: Nothing special.  The `range, inc, dim` parameters are used to pass the dimension ranges, increments or lengths, so if _family_ is **GMT_IS_GRID** and the earlier parameters imply a cube then we set the header accordingly and allocate space (if requested).  If layer spacing is equidistant (implied by a positive `inc[GMT_Z]`, then the layer array can be computed. We do something very similar for a 3-D **GMT_MATRIX**.  However, if `inc[GMT_Z] == 0`, `range[ZHI] > range[ZLO]`, and  `dim[GMT_Z] > 0` then we have a non-equidistant layering scheme.  For this we need a new API function `int GMT_Put_Layers (void *API, struct GMT_GRID *G, unsigned int n_layers, double *layers)` to place this information in the grid header.
3. _GMT_Read_Data_: If the gmt_nc.c functions detect this is a data cube and no single layer has been specified (currently, the only way to read these grids is to specify a layer), then it calls _GMT_Create_Data_ as above and reads in the header and/or data.  The `Grid->data` array is simply of size _n_layers_ times the size of one slice (mx * my). Usually, `n_layers = 1`.
[...]

Issue #2013, opened 2019-11-09
Add more useful targets for gmt docs


seisman:
`gmt docs` can open a module's documentation using the default browser. Besides the module names, some special targets are also supported:

- gmt
- cookbook
- gallery
- settings
- api
- tutorial

@PaulWessel proposed to allow `gmt docs home` or `gmt docs website` to open the GMT homepage (https://www.generic-mapping-tools.org/).

Any suggestions for other useful targets?

PaulWessel:
Maybe data could open the Dataset page.

seisman:
I'm a little hesitating to add more special targets to "gmt docs". To me, the targets I need is "forum" (https://forum.generic-mapping-tools.org/), "homepage" (https://www.generic-mapping-tools.org/), and the "documentation landing page" (https://docs.generic-mapping-tools.org/).

The target to the "Dataset" page may be unnecessary. Users can always open the documentation page first (via "gmt docs doc" or something similar), then find Dataset page from the left sidebar.

PaulWessel:
I agree that dataset is not critical.  I find myself very often needing to open either our main site.  The forum I usually go to by responding to an email with an issues, and the top doc page I go to indirectly via a module then click on sidebar, so that would be useful.  I am not super-sure about what to call the targets.  I listed website and you listed homepage; both are kind of long.  What do you and @joa-quim think about these:

1. gmt docs _home_  [Go to the "home" of the docs - i.e., the top documentation site]
2. gmt docs _forum_  [Go to the user Discourse forum]
3. gmt docs _website_|_site_|_@_ [Goes to the GMT URL]

joa-quim:
If by *target* one mean the *keyword* we can add to ``gmt docs xxxx`` then I think *data* is a good one. Having several is not distracting anyway. I do blind trials when I don't remember what exists.

PaulWessel:
I tried gmt docs solve-my-problem and it failed miserably.

PaulWessel:
To resolve this issue, I think these are not too shabby:

- gmt docs forum: Go to forum
- gmt docs data: Open the dataset page

We need to more targets: the top documentation page and the website.  I feel the target home is ambiguous.  gmt docs home sounds like the home page of the documentation rather than the gmt website.  Is that how you see it too? If so, then maybe that is OK and we can use gmt docs website for the website?

seisman:
> To resolve this issue, I think these are not too shabby:
> 
[...]

PR #2966, opened 2020-03-24
WIP Implement support for 3-D grids

PaulWessel:
Addresses steps listed in #2912 to make GMT ready to transparently handle 3-D data grids, such as needed in grdinterpolate or produced by greens line.  There are several step, and we will do one at the time and test to make sure no bad side-effects sneak in.

This initial commute simply extends the GMT_GRID_HEADER with the required parameters.


joa-quim:
This going to break Julia

PaulWessel:
How can we do things without breaking Julia/GMT?  We have to be able to improve GMT and not be told no can do, right, so how?

joa-quim:
Can we have a different header for 3D?

PaulWessel:
No, since GMT_GRID contains the GMT_GRID_HEADER.  I think this is a version management thing.  You will need to require 6.0 of users of your current Julia release, not master.  So release a Julia4gmt version that requires 6.0 and done with it.  Then, master is always a work in progress and may need to be rebuilt when GMT changes.  Most users are probably installing 6.0 anyway.

joa-quim:
Don't know if this affects julia only or not. If yes, I can always add another ``if version`` but if it affects others than we should probably not put this in 6.1

seisman:
Any schedule for the 6.1.0 release?

PaulWessel:
The 3-D stuff branch can wait for after 6.1.  We should just pick a date since otherwise we keep on adding things and delaying.  How about April 15?

joa-quim:
What else it open that we may want to put in 6.1. If minor things we can do it earlier than A15

PaulWessel:
Not sure.  I think we have fixed a bunch of stuff and added new features worthy of a 6.1 release.
I will fix #2967 problems.  Does not look like the other issues are critical (?).

seisman:
I opened #2968 for tracking the upcoming 6.1.0 release. No critical issues but I expect some final decisions about PR #2013 before the release.

PR #2972, opened 2020-03-26
Fix seg fault when GMT_DATADIR environment variable is defined but invalid

dwcaress:
We've observed instances of gmt seg faulting on startup on some machines but not others. The problem turns out to be users with the GMT_DATADIR environment variable set to nonexistant directories, a case causing the current code to try parsing a null pointer value string. The change detects the case and avoids parsing the null pointer string.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -3129,8 +3129,8 @@ GMT_LOCAL int gmtinit_set_env (struct GMT_CTRL *GMT) {
 
 	/* Determine GMT_DATADIR (data directories) */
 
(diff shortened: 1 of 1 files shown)


PaulWessel:
I am not sure this patch will work when GMT_DATADIR is a list of separated paths?  The `access (this_c, R_OK) == 0` surely will not be true if this_c is "/path/dir1:/path/dir2", right?

dwcaress:
Yes, you're right.

PaulWessel:
Hi @dwcaress, if you don't mind I will work on a separate branch to fix this and check for all those paths.

dwcaress:
Absolutely. Thanks Paul.

Issue #2969, opened 2020-03-25
Slanted annotation requires increase in label offset


PaulWessel:
When the **+a**_angle_ slanted x-annotation is used we must compensate by moving any labels and titles further away from the x-axes.  This was discovered on [the forum](https://forum.generic-mapping-tools.org/t/custom-annotations-on-the-x-axis-with-gmt-6/320/5).  For instance,

`gmt basemap -R0/5000/0/20 -JX6i/3i -Bxaf+a45+l"MY VERY LONG LABEL" -Byaf -B+t"THIS IS MY TITLE" -png map`

produces the plot below, and once we move the label then the title needs to move up as well.

![map](https://user-images.githubusercontent.com/26473567/77583108-956e6580-6e84-11ea-9190-d3cf1dffb3b6.png)

PaulWessel:
An exact solution will require passing some parameters (the angle) into PostScript and make adjustments there.

PR #2975, opened 2020-03-26
Offset labels and titles if x-annotations are angled

PaulWessel:
An angled x-annotation means it will take more space in the y-direction. This has consequences for how far labels and titles need to be - yet this was not considered.  We did considered moving the annotations themselves further away depending on the angle, but noting else.

First step is just untangling PostScript calls for computing max string height from a negative scaling operation.

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -4854,7 +4854,7 @@ void gmt_xy_axis (struct GMT_CTRL *GMT, double x0, double y0, double length, dou
 		else
 			justify = (below) ? PSL_ML : PSL_MR;
 		cos_a = 0.5 * cosd (text_angle);	/* Half-height of text at an angle */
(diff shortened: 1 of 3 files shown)


seisman:
Add a test for the feature or modify a existing one?

PaulWessel:
Closes #2969.

PaulWessel:
The next step was simply to compute the width of the largest x-annotation, multiply it by sin (angle), and use that as an extra offset for labels and title.  We do this by always adding _PSL_slant_y_ to the position of x-label and the title, with _PSL_slant_y_ initialized to zero and possibly changed when the x-axis is plotted.  I tried this for all slant angles from -90 to +90 and made a movie:

rotate

PaulWessel:
Sorry, forgot to hit comment on this hours ago...

PR #2976, opened 2020-03-26
Reset expected columns to 1 before reading custom annotations file

PaulWessel:
The **pswiggle** problem reported in [the forum](https://forum.generic-mapping-tools.org/t/custom-annotations-on-the-x-axis-with-gmt-6/320/8) had to do with failure to reset an internal counter.  After reading the wiggle data that counter was 3 (columns) and thus when reading the 1 column plus trailing text we instead tried to read 3 columns and no text left.  The critical variable (_GMT->current.io.max_cols_to_read_) is now reset to 1 before reading the 1-column data file.


PaulWessel:
In my attempts to fix this I ended up renaming _gmt_disable_bhi_opts_ to _gmt_disable_bghi_opts_, etc., as **-g** should not affect the custom annotations reading, but turned out it was not the main culprit; this is why there are so many files that have changed.

PR #2985, opened 2020-03-29
Be consistant with use of rgb macros

PaulWessel:
We have macros _gmt_M_s255_ for multipying and _gmt_M_is255_ for dividing by 255.  This PR fixes a few places where we are not using them.


joa-quim:
... but I think the use of these macros make the code less readable. When I see ``gmt_M_is255()`` I instantly think it's a macro to ask if the arg **is** = 255 and never that it's dividing by 255.0

PR #2988, opened 2020-03-30
Make single-use functions internal and static

PaulWessel:
Moving a bunch of functions (~20-30) from external to internal since they are only used within a single file and thus should not be exported.

### DIFF
--- a/src/gmt_remote.c
+++ b/src/gmt_remote.c
@@ -274,7 +274,7 @@ GMT_LOCAL int gmthash_get_url (struct GMT_CTRL *GMT, char *url, char *file, char
 	return 0;
 }
 
-struct GMT_DATA_HASH * hash_load (struct GMT_CTRL *GMT, char *file, int *n) {
+GMT_LOCAL struct GMT_DATA_HASH * hash_load (struct GMT_CTRL *GMT, char *file, int *n) {
 	/* Read contents of the hash file into an array of structs */
 	int k;
 	FILE *fp = NULL;
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -5816,7 +5816,7 @@ static int chmatch (const char *target, const char *pat) {
    [[] - match '['.
    [][abc] match ], [, a, b or c
 */
-int matchwild (const char *str, const char *pattern) {
+GMT_LOCAL int matchwild (const char *str, const char *pattern) {
 	const char *target = str;
 	const char *pat = pattern;
 	int gobble;
@@ -7210,7 +7210,7 @@ int gmt_list_cpt (struct GMT_CTRL *GMT, char option) {
 }
 
 /*! . */
-void gmtlib_make_continuous_colorlist (struct GMT_CTRL *GMT, struct GMT_PALETTE *P) {
+GMT_LOCAL void support_make_continuous_colorlist (struct GMT_CTRL *GMT, struct GMT_PALETTE *P) {
 	/* Convert a (by default) discrete CPT made from a color list to a continuous CPT instead */
 	unsigned int k, i;
 	gmt_M_unused(GMT);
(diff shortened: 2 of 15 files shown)


seisman:
A warning if `USE_COMMON_LONG_OPTIONS` is not defined.
```
../src/gmt_init.c:488:16: warning: unused function 'gmtinit_handle_escape_text' [-Wunused-function]
GMT_LOCAL void gmtinit_handle_escape_text (char *text, char key, int way) {
```

seisman:
Another two warnings:
```
In file included from ../src/gmt_customio.c:1724:
../src/gmt_gdalwrite.c:177:17: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]
                for (k = 0; k < n_colors; k++) {
                            ~ ^ ~~~~~~~~
../src/gmt_gdalwrite.c:185:19: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]
                                for (k = 0; k < n_colors; k++)
                                            ~ ^ ~~~~~~~~
2 warnings generated.

```

joa-quim:
DO NOT MERGE yet. Removing ``EXTERN_MSC `` may break externals. I have no time today to test it but I'm afraid of Julia breaks.

PaulWessel:
We can wait for you to verity tomorrow.  But doubt you are calling these from Julia - you are just using the API calls.

joa-quim:
I'll have time latter today. But I'm not calling only API functions. I can call any function that is exported. Changing that is a kind of breakage.

PaulWessel:
It is, but none of the internals are guaranteed to be available to externals.  We can adjust of course but we don't want the externals to be so fragile that they depend on lower-level undocumented things.  This is what Caress did and suddenly we get complaints that some long-forgotten library function was used by him...

joa-quim:
Nothing is detectable broken on Julia side.

It's not the same as Dave's case that was (is) accessing lower level functions for grid reading when there are API functions to do that, I need, for example, to access to ``gmt_manage_workflow`` or ``gmt_getpen`` or structs like ``GMT_DATASEGMENT``. Without these there is no Julia wrapper.

src/gmt_remote.c
```diff
@@ -274,7 +274,7 @@ GMT_LOCAL int gmthash_get_url (struct GMT_CTRL *GMT, char *url, char *file, char
 	return 0;
 }
 
-struct GMT_DATA_HASH * hash_load (struct GMT_CTRL *GMT, char *file, int *n) {
+GMT_LOCAL struct GMT_DATA_HASH * hash_load (struct GMT_CTRL *GMT, char *file, int *n) {
```
joa-quim: Aren't all GMT_xx public?

src/gmt_support.c
```diff
@@ -7210,7 +7210,7 @@ int gmt_list_cpt (struct GMT_CTRL *GMT, char option) {
 }
 
 /*! . */
-void gmtlib_make_continuous_colorlist (struct GMT_CTRL *GMT, struct GMT_PALETTE *P) {
+GMT_LOCAL void support_make_continuous_colorlist (struct GMT_CTRL *GMT, struct GMT_PALETTE *P) {
```
joa-quim: Misses the gmt prefix? (gmt)support_make_continuous_colorlist

src/gmt_support.c
```diff
@@ -13858,7 +13858,7 @@ unsigned int gmtlib_pow_array (struct GMT_CTRL *GMT, double min, double max, dou
 }
 
 /*! . */
-uint64_t gmt_time_array (struct GMT_CTRL *GMT, double min, double max, double inc, char unit, bool interval, double **array) {
+GMT_LOCAL uint64_t support_time_array (struct GMT_CTRL *GMT, double min, double max, double inc, char unit, bool interval, double **array) {
```
joa-quim: Iden, misses gmt prefix

PaulWessel:
Right, because you wrap the C I think.  It would probably be a good idea for you/us to maintain a list of lower-level functions etc that you actually use.  It would potentially head off a problem.

PaulWessel:
I am going to commit this guy.  OK @joa-quim ?

src/gshhg_version.h
```diff
@@ -52,7 +52,6 @@ struct GSHHG_VERSION {
 };
 
 /* Prototypes */
-EXTERN_MSC int gshhg_get_version (const char* filename, struct GSHHG_VERSION *gshhg_version);
```
joa-quim: What if someone whats to know the GSHHG version?

joa-quim:
All functions and structs currently used in Julia are declared in

https://github.com/GenericMappingTools/GMT.jl/blob/master/src/libgmt.jl
https://github.com/GenericMappingTools/GMT.jl/blob/master/src/libgmt_h.jl

Except GMTAPI_DATA_OBJECT that is not actually used.

src/gshhg_version.h
PaulWessel: They can, this is a sub-function of the gshhg_version function.

src/gmt_remote.c
PaulWessel: See #2996 for this discussion - you are both right and wrong on these comments; I will resolve them for now.

Issue #2995, opened 2020-03-30
plot -Sf gap - unwanted size change


KristofKoch:
**Description of the problem**

When using `-Sf` the size of the gap between the symbols changes depending on supplied path. 
* Two identical fronts are used for illustration, boxes are 80p in size with a 160p spacing. 
* The right (blue) front is shifted 80p to create an alternating pattern. 
* The paths both run from bottom right to upper left. 
* In both examples the plot call is absolutely identical, only the supplied path changes.

**Full script that generated the error**

```
cat > path.txt << END
 1.3 -1.5
-1.8  1.2
END

cat > path2.txt << END
 1.3 -1.5
 0.2  0.4
-0.8  0.1
-1.8  1.2
END

gmt begin small_gap
  gmt basemap -R-2/2/-2/2 -JM6i -Bafg1
  gmt plot path.txt -W0p -Sf160p/80p+b+l+pred
  gmt plot path.txt -W0p -Sf160p/80p+b+r+pblue+o80p
  gmt plot path.txt -Skcrosshair/0.5c -Wthinnest,black
gmt end show

gmt begin large_gap
  gmt basemap -R-2/2/-2/2 -JM6i -Bafg1
  gmt plot path2.txt -W0p -Sf160p/80p+b+l+pred
  gmt plot path2.txt -W0p -Sf160p/80p+b+r+pblue+o80p
  gmt plot path2.txt -Skcrosshair/0.5c -Wthinnest,black
gmt end show
```
**Actual outcome**

![small_gap](https://user-images.githubusercontent.com/45796239/77954320-c3add400-72ce-11ea-9e0a-42ee34450286.jpg)

![large_gap](https://user-images.githubusercontent.com/45796239/77954358-ce686900-72ce-11ea-8ef7-3252d204d002.jpg)

The two different paths lead to two different gap sizes all other things being the same.

**Expected outcome**

I expected the gaps to be as wide as the boxes and not changing size depending on supplied path in an unpredictable manner.

**System information**

* Operating system: macOS
* GMT version: a6f3e31

KristofKoch:
Some further experimentation lead me to believe that this might actually be an intentional feature. By sightly nudging the gap size the line terminates on a symbol making for a nice line end. This is hardly noticeable at small symbol sizes but gets very obvious at larger sizes as shown above.

```
cat > path.txt << END
 1.3 -1.5
-1.8  1.2
END

gmt begin small_gap
[...]

PR #3006, opened 2020-04-01
Let positively defined front gaps be used exactly as given

PaulWessel:
If the front gap is specified with an explicit plus-symbol then we take that to mean the gap should be used as given.  The default remains to distribute front symbols exactly along the line and thus adjust the gap accordingly.  See #2995 for context.


PaulWessel:
Thanks, we can wait for @KristofKoch to give it a spin since he has an interest.

PR #3015, opened 2020-04-02
Disable unit scramble for virtual files

PaulWessel:
The angles determined for decorated symbols placed along the line got multiplied by 2.54 because we were trying to defeat the automatic scaling during reading from files.  But here we are reading from a _virtual_ file and no scaling should take place.  This addresses the [concern](https://forum.generic-mapping-tools.org/t/proper-usage-of-s/338/4) reported by @KristofKoch.  I also let the decorated symbols get drawn by the default pen if nether fill nor pen was chosen by user; otherwise nothing is plotted and that confuses the user.

### DIFF
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -479,5 +479,6 @@
                 Specifies how many (*x*,\ *y*) points will be used to estimate
                 symbol angles [Default is 10].
 
+        If neither **+g** nor **+p** are set we select the default pen outline (:term:`MAP_DEFAULT_PEN`.
         **Note**: By placing **-S~** options in the segment header you can change
         the decorated lines on a segment-by-segment basis.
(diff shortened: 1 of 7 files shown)


doc/rst/source/explain_symbols.rst_
```diff
@@ -479,5 +479,6 @@
                 Specifies how many (*x*,\ *y*) points will be used to estimate
                 symbol angles [Default is 10].
 
+        If neither **+g** nor **+p** are set we select the default pen outline (:term:`MAP_DEFAULT_PEN`.
```
seisman, line 482:
```suggestion
        If neither **+g** nor **+p** are set we select the default pen outline (:term:`MAP_DEFAULT_PEN`).
```

doc/rst/source/explain_symbols2.rst_
```diff
@@ -490,5 +490,6 @@
                 Specifies how many (*x*,\ *y*) points will be used to estimate
                 symbol angles [Default is 10].
 
+        If neither **+g** nor **+p** are set we select the default pen outline (:term:`MAP_DEFAULT_PEN`.
```
seisman, line 493:
```suggestion
        If neither **+g** nor **+p** are set we select the default pen outline (:term:`MAP_DEFAULT_PEN`).
```

PaulWessel:
Fixed after commit.

PR #3018, opened 2020-04-03
Decorated lines with custom symbol

PaulWessel:

The decorated lines symbol option (**-S~**) in plot has so far been limited to the fixed geometric symbols (circles, triangles, etc.) specified with the modifier **+s**`<symbol><size>`.  This PR allows the use of custom symbols via **+sk**_customsymbol_/_size_, just like in plot's **-Sk** option. The custom symbol must be a plain symbol that does not require any additional parameters other than symbol size.  We automatically modify the symbol to insert the rotation macro so that the symbol can be aligned with the path.  The rotation can be fixed via **+a**_angle_.  New test _psxy/custom_decorate.sh_ has been added.

### DIFF
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -472,8 +472,9 @@
                 optionally specify pen for outline [Default is width = 0.25p,
                 color = black, style = solid].
 
-            **+s**\ <symbol><size>
+            **+s**\ <symbol><size> or **+sk**\ *customsymbol*/*size*
                 Specifies the code and size of the decorative symbol.
+                Custom symbols need to be simple, i.e., not require data columns.
 
             **+w**
                 Specifies how many (*x*,\ *y*) points will be used to estimate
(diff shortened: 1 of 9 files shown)


KristofKoch:
I generally works but I found an issue with it:

When adding the `+i` option it fails with a rather unspecific error message and no line or symbol is plotted:

    plot [ERROR]: Option -S: Parsing failure

To reproduce: 


```
cat > path1.txt << END
-1.5 -1.5
 1.5  1.5
END

gmt begin symbol_line_i
  gmt plot path1.txt -S~d50p:+skvolcano/50p+i
gmt end show
```

Tested with 8121f5f

PaulWessel:
There is no **+i** in this branch. But the message should be improved.

PaulWessel:
Now says:

```
plot [ERROR]: Modifier +i not recognized!
plot [ERROR]: Option -S: Parsing failure
```

PaulWessel:
Better:

```
plot [ERROR]: Option -S~: Modifier +i not recognized!
plot [ERROR]: Option -S: Parsing failure

```

KristofKoch:
> There is no **+i** in this branch. 

Ooops, my mistake. Thank you for making the error message more meaningful.

Issue #3021, opened 2020-04-03
Polygon centroids


joa-quim:
``gmtspatial -Q`` computes polygon centroids, but the result is different from that of GEOS and Matlab (they both agree for the cases I tested).
The figs shows an example. Black circles are from ``gmtspatial -Q`` and red are fro GEOS obtained with ``gmtspatial -Sc`` from the ``buffer_GEOS`` branch.

![centroids](https://user-images.githubusercontent.com/537321/78411467-a8f99900-7607-11ea-937b-e282707d715c.png)

PaulWessel:
We are probably not doing it right - certainly not the same way.  I think I am returning the mean vector instead.

joa-quim:
The formula/example in this site gives agrees with the GEOS response
https://bell0bytes.eu/centroid-convex/

PaulWessel:
Thanks, found something similar - will fix.

PaulWessel:
Which data set is this and what command - if you have it handy?

joa-quim:
I did it al in julia. First example is from the page I linked above. The ``S=:c`` is the GEOS solution.

```
julia> gmtspatial([1.0 0; 2 1; 0 3; -1 2; -2 -1; 1 0], S=:c)
Array{GMT.GMTdataset,1} with 1 segments
First segment DATA
1×2 Array{Float64,2}:
 -0.0833333  0.916667


julia> Dc = gmtspatial([4.42998 5.81301; 1.86776 5.68526; 2.53231 7.42577; 3.24964 7.07831; 4.42998 5.81301; 4.42998 5.81301], S=:c)
Array{GMT.GMTdataset,1} with 1 segments
First segment DATA
1×2 Array{Float64,2}:
 2.99039  6.35543


julia> Dc = gmtspatial([4.42998 5.81301; 1.86776 5.68526; 2.53231 7.42577; 3.24964 7.07831; 4.42998 5.81301; 4.42998 5.81301], Q=true)
Array{GMT.GMTdataset,1} with 1 segments
First segment DATA
1×3 Array{Float64,2}:
 3.30193  6.36307  2.43609
```

PaulWessel:
I cannot see that data set in that link...
Also, things get more complicated on a sphere.   GIS seems to always worry about a backyard so Cartesian is good enough.

PaulWessel:
OK, I see, you giving me one polygon with those coordinates.  Fine.

PaulWessel:
[...]

PR #3022, opened 2020-04-04
Implement Cartesian centroid calculation

PaulWessel:
What we had was mean location.  Addresses #3021.  After this fix, I get the same solution as the GEOS solution reported by @joa-quim in #3021.
Will need more work later for a geographic (spherical) solution.


joa-quim:
[This ](https://math.stackexchange.com/questions/2333601/finding-centroid-of-spherical-triangle)looks useful for the spherical case. But it's only for triangles

PaulWessel:
I have implemented spherical but need this one committed first. Doing now.

PR #3023, opened 2020-04-04
Check if netcdf4 grid before calling nc_inq_var_deflate

PaulWessel:
netCDF 4.7.4 released April 4, 2020 on macports caused massive crashes when reading netcdf3 grids as the nc_inq_var_deflate now suddenly crashes with an error instead of previous behavior.  Isolating this call by first checking if netcdf4 before calling the problematic function.  As this change in netcdf is backwards incompatible I have posted an [issue](https://github.com/Unidata/netcdf-c/issues/1689) on their site.


PaulWessel:
Nothing like a detour dealing with shit like this when in the middle of 2-3 other branches.... Please approve @joa-quim or @seisman so I can continue...

PR #3024, opened 2020-04-04
Correct two issues in gmtvector

PaulWessel:

There were two problems discovered in gmtvector:

1. When **-Am** was used on geographic data, we failed to convert geodetic latitudes to geocentric (and back) when doing the vector averaging when **-E** was active.
2. When general 3-D Cartesian vectors were given (not geographic lon,lat -> x,y,z) we forgot and inverted for lon, lat even though these were not meant to be geographic unit vectors.

Added -E where appropriate in the tests and update the plot and results.


PaulWessel:
Need this one committed to finishe spherical areas and centroid, @joa-quim

PR #3025, opened 2020-04-05
Spherical area and centroid calculations in gmtspatial

PaulWessel:

GMT had several limitations when computing the area of large geographic polygons as well as their centroid.

- The centroid, to first order, suffered from the same issue as the Cartesian implementation (#3021 was fixed by #3022) in that it simply computed the mean vector.  This is incorrect of course since the density of points along a line should not affect the balancing point (centroid) of the polygon.  A correct spherical centroid calculation requires more work and involves breaking down the polygon to a set of _spherical triangles_.  The centroids of each of these triangles are the mean vectors, but the centroid of the polygon is then a weighted average of all triangle centroids, with triangle areas as weights.  A good overview is given [here](http://www.jennessent.com/downloads/Graphics_Shapes_Online.pdf).
- The area was simply performed by doing an azimuthal equal-area projection at the mean vector location and then do a Cartesian calculation on the projected coordinates (in meters).  This meant that any great circle was replaced with a straight line in the projection, thus the two measures of area would always differ, and more for larger areas.

Fortunately, the STRIPACK library of Renka, which GMT already includes for spherical triangulation, has a function to compute the area of a spherical triangle (already used in **triangulate**).  Unfortunately, it does not keep track if the triangle vertices are arranged clockwise of counter-clockwise, since counter-clockwise triangles should be assigned a negative area so that the proper weights are used in the centroid calculation as well as proper area cancellations in the polygon area sum.  With these complications and the area function, this PR contains these changes for how spherical polygon centroid and area are computed in **gmtspatial**:

1. We first compute the mean vector of the vertices (i.e., as in the old incorrect method for centroid).
2. This mean vector is assigned as a special central node and we then loop around the perimeter of the polygon and connect each side (2 points) with the central node, thus constructing a spherical triangle.
3. We compute its area with the Renka area code.
4. To determine its handedness (orientation) I compute the vector triple product of the three vectors in the triangle.  This product is positive for clockwise and negative for counter-clockwise arrangements, allowing us to obtained a signed area.
5. As the loop progresses we accumulate the necessary sums of weighted centroid components and the overall weight (i.e., area).
6. At the end of the loop we evalute the weighted centroid vector and convert back to lon,lat.
7. Because of spherical geometry, all data coordinates are converted to geocentric latitudes.

With these changes, some tests and examples fail, obviously.  Example 40 computes the area and centroid of Australia.  The old centroid was clearly wrong, being pulled too far north by the greater density of points along the irregular northern coastline.  The new area also differs a bit.

As an example of the differences involved, if you consider a 5x5 degree polygon with its lower left vertex at (0,0), a quick back-of-the-envelope calculation says the area is approximately `5x5*111.195*111.195*cos(2.5) = 308814 km^2`.  The old Cartesian/projection scheme gives 307249 km^2 while the spherical solution is 306858 km^2.  For a large polygon the differences can be much larger.  Below is two large polygons (mirrored across the Equator for testing they give the same area); this is a new test in _gmtspatial/sphareacentroid.sh_:

sphareacentroid

The black dots are the polygon vertices, the red dot is the mean vector location serving as the special central node, and the thin lines indicate the 8 spherical triangles.  As we march along this polygon from the lower-left point (the next point is the lower-right point, etc.) we get clock-wise positive contributions until we enter the notch on the left, where the three sides all yield a negative area.  It is basically the spherical equivalent of the trapezoid-sum rule for Cartesian coordinates. The yellow star is the final centroid and indicates how different it can be from the mean vector location.  The reported area (41,954, 219 km^2) is ~20% larger than what the old Cartesian/projection scheme computed.

Given that my 3-4 tests give reasonable and improved answers, and the test triangle [(0,0)-(90,0)-(90,90)] gives the correct 1/8th of the Earth's area, I think we should accept this PR.


src/gmt_sph.c
```diff
@@ -388,3 +388,80 @@ int gmt_ssrfpack_grid (struct GMT_CTRL *GMT, double *x, double *y, double *z, do
...
+	gmt_cross3v (GMT, A, B, X);
+	return (gmt_dot3v (GMT, X, C) < 0.0 ? -1 : +1);
+}
+
+/* Compute spherical triangle area */
+
+double gmtlib_geo_centroid_area (struct GMT_CTRL *GMT, double *lon, double *lat, uint64_t n, double *centroid) {
+	/* Based on ideas in http://www.jennessent.com/downloads/Graphics_Shapes_Online.pdf and Renka's area_ function.
+	 * Compute area of sphe
+	 int sgn;rical polygon and its centroid */
```
joa-quim, line 404: Fix text

PaulWessel: Done.

Issue #3016, opened 2020-04-02
Let GMT play nicer with MATLAB data files


PaulWessel:
**Description of the desired feature**

By default, GMT tables use the hashtag in the first position to indicate a data header or comment, e.g.,

`# This is a comment or header`

However, via **IO_HEADER_MARKER** any other single character can be used for input and/or output.

MATLAB uses % to indicate comments. While GMT can read such file by supplying `--IO_HEADER_MARKER=%`, MATLAB has no provision for accepting our # marker without some programming.

It would simplify life for users who (1) need to go between MATLAB and GMT and (2) that are not using the GMT/MEX tool if GMT could simply accept records starting with % as another comment or header without having to use `--IO_HEADER_MARKER=%` for **reading**.  For **writing** one needs to select this marker.  Internally, GMT maintains separate variables for the input and output markers.  My proposal for making this more flexible is to let the input marker variable be a **string** so that more than one marker would be recognized as flagging a header.  Thus, if the GMT defaults were like this:

`IO_HEADER_MARKER               = #%,#`

then GMT would seamlessly read MATLAB tables without any special arguments.  If there are other major softwares with another flag for header then that could be added too [I just don't know].

PaulWessel:
Do you see any negative effects of allowing this flexibility? Note, this is separate from multiple segment headers where Matlab uses NaNs to break up lines.  So it should only affect file headers.

joa-quim:
Reading ascii files already slow, also because every line is checked to be a comment line. I'm afraid that checking for strings instead of a single char slows it even more, and that for a very marginal use case.

PaulWessel:
So I made a test: I created a 10 million record three-column ASCII data table (343 Mb) and ran **gmt info** on it 5 times using the current single char check: Each record has its first character compared to the single marker character `(line[0] == marker[GMT_IN])`.  I then made some quick changes to the read function so that instead it used `strchr (list_of_markers, line[0])` to detect header records that started either with "#" or "%".  Recompiled and ran that 5 times too, timing it.  The two data sets of elapsed times are identical at the 99.84% level of confidence, using MATLAB's two-sample t-test.

It is indisputable that GMT ASCII i/o is on the slow side - and hopefully this can be addressed during the refactor.  But with no detectable change for an enhancement that addresses an actual issue, I think this is a reasonable feature to consider, no?

joa-quim:
Ok, than. But out of curiosity, how fast would it be if no testing for headers at all?

[...]

PR #3026, opened 2020-04-05
Let GMT recognize MATLAB headers/comments

PaulWessel:
By extending the **IO_HEADER_MARKER** setting to accept a string of possible input-header markers we can set the default to `#%'"` which means either GMT or MATLAB header records as well as quoted text records are automatically recognized.  For output only one marker can be selected `[#]`.

### DIFF
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -474,9 +474,10 @@ I/O Parameters
 
     **IO_HEADER_MARKER**
         Give a string from which any character will indicate a header record in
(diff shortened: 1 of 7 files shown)


joa-quim:
A low probability event, but **IO_SEGMENT_MARKER** should not be allowed to be any of the **IO_HEADER_MARKER**'s

PaulWessel:
Good point, I will add that test.

PaulWessel:
This PR is meant to close #3016.  Test described above has been added and works.

PaulWessel:
Please read and comment on the last entry on #3016 first.

Issue #2987, opened 2020-03-29
Need script to examine use of functions


PaulWessel:
**Description of the desired feature**

See #2986 for context.  I would like a script that examines all the *.c files and makes a listing of each exported (i.e., not static or GMT_LOCAL) function and a list of the files it are called in. E.g.,

```
name_of_function1 [declared in gmt_xxxx.c]
  blockmean.c
  gmt_support.c
  gmt_io.c

name_of_function2  [declared in gmt_xxxx.c]
  filter1d.c
  gmt_calclock.c

...
```

Output should be sorted so the functions called in the fewest files are listed first. This would make it easier to determine which functions might move from gmt_xxx.c to gmt_yyy.c and cease to be exported.

seisman:
Try this script. It's not exactly what you want.
```bash
#!/usr/bin/env bash

# Find all C source files
find src -name '*.c' |\
	# Run ctags to output exported functions in each C file
	# Options:
	#   --filter=yes=yes: let ctags read c source file name from stdin
	#	--file-scope=no:  exlude all "local" tags (i.e. local variables and static functions)
	# Output format:
	#	tag_name tag_type line_no source_filename the_source_line
	# Example output:
	#   shore_res_to_int function    386 src/gmt_shore.c  GMT_LOCAL int shore_res_to_int (char res) {
	ctags --filter=yes --file-scope=no -x |\
	# Further exclude some global variables and GMT_LOCAL (i.e. static) functions
	gawk '$2=="function" && $5 !~ "^GMT_LOCAL" {print $1, $4}' > /tmp/tmp.list

while read func file; do
	echo $func "[declared in $file]"
	find src -name '*.c' ! -path "$file" -exec grep $func {} \+ | awk -F':' '{print "\t", $1}' | sort -u
done < /tmp/tmp.list
```

PaulWessel:
Thanks, had to install ctags as Xcode also ships with a ctags that did not understand those options...

PR #3027, opened 2020-04-05
WIP Script to determine function usage

PaulWessel:
We want to minimize the number of functions that are exported in the library.  See #2987 for details. One way to do so is to keep functions static if they are only used in one file.  This script analyzes how functions are used. However, it is not yet finalized.  I want the output sorted such that the least frequently used functions are listed first.  This sorting and reassembly of the output is not yet working well.  If you run it, you will see it mostly works but then _GMT_Get_Index_ is listed as being used in src/gmt2kml.c yet its count is zero and no, it is not used in that module. Perhaps @seisman can see the probem...All intermediate results are for now left in /tmp/gmt.

### DIFF
--- a/admin/count_function_usage.sh
+++ b/admin/count_function_usage.sh
@@ -6,6 +6,10 @@
 # made static, perhaps by moving from where they are declared to where they are
 # used.
 
+if [ ! -d src ]; then
+        echo "count_function_usage.sh: Must be run from top-level gmt directory" >&2
+        exit 1
+fi
 # Use subdir in /tmp as working directory
 rm -rf /tmp/gmt
 mkdir -p /tmp/gmt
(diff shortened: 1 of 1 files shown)


PaulWessel:
To do list:

1. Exclude the FORTRAN bindings (i.e., ends with underscore, such as GMT_End_IO_)
2. Skip the kiss stuff since not our code
3. Flag functions not following naming convention

seisman:
> This sorting and reassembly of the output is not yet working well. If you run it, you will see it mostly works but then _GMT_Get_Index_ is listed as being used in src/gmt2kml.c yet its count is zero and no, it is not used in that module. 

There is a `gmt_get_index` function in src/gmt2kml.c. Obviously, macOS's file system is case-insensitive and the gmt_get_index information is written to `/tmp/gmt/GMT_Get_Index/t.lst`.

PaulWessel:
But this is C code doing strcmp etc.  Should not matter with case at all.

seisman:
The culprit is the redirection:
```
find src -name '*.c' ! -path "$file" -exec grep $func {} \+ | awk -F':' '{print "\t", $1}' | sort -u > /tmp/gmt/${func}/t.lis
```

`/tmp/gmt/GMT_Get_Index/t.lis` and `/tmp/gmt/gmt_get_index/t.list` are the same under macOS.

PaulWessel:
OK, but I think this is an abandoned branch (sorry).  I am doing so much more stuff in the C code in admin.

seisman:
OK, I see. Closing.

PR #3028, opened 2020-04-05
Address bad memory freeing related to polar cap

PaulWessel:
See [forum](https://forum.generic-mapping-tools.org/t/gmt-plot-malloc-error-message/348/4) background.  This PR fixes the issue by assigning _free_memory_ after _add_pole_ has been finalized.

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -3913,7 +3913,7 @@ GMT_LOCAL uint64_t plot_geo_polygon_segment (struct GMT_CTRL *GMT, struct GMT_DA
 	uint64_t n = S->n_rows, k;
 	double *plon = S->data[GMT_X], *plat = S->data[GMT_Y], t_lat;	/* Default is to plot incoming array as is via plon,plat pointers */
 	bool ap = at_pole (plat, n);	/* Is the first and last point exactly at the pole? */
-	bool free_memory = add_pole;
+	bool free_memory;
 	struct GMT_DATASEGMENT_HIDDEN *SH = gmt_get_DS_hidden (S);
 	if (ap) plon[n-1] = plon[0];	/* Just enforce the same longitude at the pole point */
 	GMT_Report (GMT->parent, GMT_MSG_DEBUG, "Polar cap: %d\n", (int)add_pole);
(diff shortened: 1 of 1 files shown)


PaulWessel:
One of you need to hit the review button and approve.

joa-quim:
Only that as a fix? But when
```
plon = gmt_M_memory (GMT, NULL, n, double);
```
it *forgets* about the memory that ``plon`` was pointing first in
```
double *plon = S->data[GMT_X]
```

PaulWessel:
Yes, but S is not going anywhere.  I think it is OK.  No more wrong frees, and no other effects across the tests.

joa-quim:
But if S->data was not freed, and it's not going anywhere, it leeks.

PaulWessel:
Those are all taken care of by the garbage man later.

PaulWessel:
Remember, we have a full system checking if things are freed properly, so even though it did not crash for me, it told me that if was trying.

joa-quim:
Later means at the end of the module. Why not free it right away? Clear to read and memory can be reused.

PaulWessel:
Not going to change the entire GMT memory management at this point.  The file has all been read into memory, there is no urgency to free up space as relatively little else is allocated (e.g., plotting, and those gets deallocated)

joa-quim:
OK, but it shouldn't even been doing this test at all. The seek polygon is in Cabo Verde. Far from any pole.

PaulWessel:
Well, no.  It is doing this test since the user gave an oblique region and needs to know if any part of that overlaps with this polar cap . The answer was no, so we move on - except for the failure to change that bool variable.  For straight w/e/s/n bounds we can determine that the segments w/e/s/n is outside the plot w/e/s/n.  I can look to see if we do that here or deeper down.  @KristofKoch could also use gmtselect to avoid plotting the entire world dataset for a small subregion.

KristofKoch:
>  @KristofKoch could also use gmtselect to avoid plotting the entire world dataset for a small subregion. 

You are absolutely correct, @PaulWessel. Thank you for the hint towards `select`. I didn't think about that.

PR #3030, opened 2020-04-06
Add a region arg to gmt_ogrread() to allow filtering the returned features

joa-quim:
The region filter is not exactly like -R. Instead it returns the features intersected by the rectangular *region*.

With this one can read Kristof *Cabo Verde* example in Julia with
```
r = [-25.14, -24.8, 16.75, 16.95];
D = GMT.ogr2GMTdataset(GMT.gmt_ogrread(API, "c:/v/land-polygons-complete-4326/land_polygons.shp", r));
```


joa-quim:
> Will gmt_ogrread ve called by modules or will it be doing indicectly via things in gmt_io.c?

This is an entirely open issue. So far it's only used by Julia via a direct call to the function.

PR #3033, opened 2020-04-06
Introduce admin tools to help us follow our code convention

### LINKED ISSUE #2987: Need script to examine use of functions

PaulWessel:
We have  declared that GMT functions should follow a certain naming convention depending of they are part of the external API, the developer library, internal calls, or statics.  This C code and script produce a listing of what we currently have. I will add more checks such as adding renaming recommendations etc.
Closes #2987.

### DIFF
--- a/admin/src_convention_check.c
+++ b/admin/src_convention_check.c
@@ -324,9 +324,13 @@ int main (int argc, char **argv) {
 		else if (F[f].dev && strncmp (F[f].name, "gmt_", 4U)) err = 1;
 		else if (F[f].lib && strncmp (F[f].name, "gmtlib_", 7U)) err = 2;
 		if (err == 3) {
-			p[L-2] = 0;
-			sprintf (message, "Name error, should be %s_*", p);
-			p[L-2] = '.';
+			if (F[f].n_files > 1)
+				strcpy (message, err_msg[err]);
+			else {
+				p[L-2] = 0;
+				sprintf (message, "Name error, should be %s_*", p);
+				p[L-2] = '.';
+			}
 		}
 		else
 			strcpy (message, err_msg[err]);
--- a/admin/src_convention_check.c
+++ b/admin/src_convention_check.c
@@ -14,6 +14,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <libgen.h>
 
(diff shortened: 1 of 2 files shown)


PaulWessel:
Ready for prime time testing.  It now reports if a functions is mis-named given the conventions listed above.  IT does not rename anything of course, it is just a reporting tool.  It lists how many times a function is called, in how many files, and which files, plus a status of S for static, D for developer, and L for library.  I am skipping the API functions and the GMT_blockmean module calls themselves since those are part of the API and well understood and correctly named.

PaulWessel:
Usage: Cd to your top-level gmt-dev folder and run

`admin/src_convention_check.sh > /tmp/shit`

for the full report.  Add options -f just to list the functions and -e to only do externals.

PaulWessel:
FYI, for now it will report that functions in the supplements should be called gmt_* or gmtlib_*.  I will fix that but we don't really have a convention for function names in the supplements.  I would think they would all be local to each supplement, but there are exceptions such as x2sys calling functions from mgd77.  I won't worry about this yet an only focus on the gmt core.

joa-quim:
> could we defeat that by having a file gmtlib.c that has #include "gmt_map.c", etc etc. and then they can all be static?

I fear that you are preparing to close to many functions under static. Remember that the Julia ability to access functions directly is not only a power user thing but it could (and should) be used to extend the tests beyond the plotting level.

PaulWessel:
Right, we can do that carefully.  I don't think you are using functions that are just supporting other library functions in the dev library, but who knows.
First priority is to clean up name messes, which is unlikely to cause you much problems since I bet you are mostly using gmt_* functions.

PR #3039, opened 2020-04-08
Internal renaming of static functions in gmt_api.c

PaulWessel:

That, plus removing a few EXERN_MSC entries from modules as these functions were either already in gmt_prototypes.h or they should have been there.


src/gmt_api.c
```diff
@@ -507,12 +507,12 @@ GMT_LOCAL unsigned int api_gmtry (unsigned int geometry) {
 
 GMT_LOCAL inline struct GMTAPI_CTRL * api_get_api_ptr     (struct GMTAPI_CTRL *ptr)  {return (ptr);}
```
seisman, line 508: Are you going to rename other functions e.g. `api_get_api_ptr` in future PRs?

PaulWessel: Yes, and there are a few that requires special checking such as the ones we build into gmt_core_module.c etc from the scripts.  This is just the bulk with more to come.

joa-quim:
I'm still  bit confused with this renaming rule. If allmost our source files are named gmt_xxxx what extra info does it give to add the *gmt* prefix to all those functions?

PaulWessel:
The idea is to make it easier for other people to navigate the source (e.g., new postdoc).  The name of a function will immediately tell you if it is an API, part of the dev library, an internal library function, or a local static one.  We have functions called gmt_* that are static and functions called api_* that are exported. It is a mess.

joa-quim:
Yes, I understand the idea but I don't see how prefixing the *same* prefix to all (all functions in gmt_xxx will gain it) of them will help in that regard.

PaulWessel:
Hm, not sure I follow.  For each gmt_*.c file there will be gmt_*, gmtlib_* and <gmtfilename>_* functions.  We have thousands of functions.  I am the one fixing most of the bugs since I wrote most of it.  Making things clearer for others is critical I think.  If something crashes in somefunction vs gmtinit_somefunction immediately tells you it is a static function gmt_init.c

PR #3044, opened 2020-04-08
Complete cleaning of gmt_api.c

PaulWessel:

Also, renamed badly named functions and removed some unused declarations.  Eliminated spurious extern declaration in some modules which duplicated entries in gmt_prototypes.h; same for externs listed in some gmt_*.c files that had entries in gmt_internals.h.  Also added missing entries for some functions to those two includes.


PaulWessel:
I think this on is good to go @seisman.  I had to double-check something and then another zoom meeting took place.

PR #3092, opened 2020-04-13
Allow new supplement testing

PaulWessel:
Specifically, allow a supplement to have some specific include and library requirements not part of the supplement (e..g, system libs and includes). These are the changes:

1. Allow the CMakeLists.txt file in a supplement directory to specify extra include dirs and libraries, e.g.,

```
set (SUPPL_EXTRA_INCLUDES /Users/pwessel/GMTdev/mbsystem/include)
set (SUPPL_EXTRA_LIBS /Users/pwessel/GMTdev/mbsystem/lib/libmbio.dylib)
```

1. Let the src/CmakeLists.txt file harvest all such settings then include them in the main setup:

```
	# supplement extra include files
	get_subdir_var (GMT_EXTRA_INCLUDES SUPPL_EXTRA_INCLUDES ${GMT_SUPPL_DIRS})
	include_directories (${GMT_EXTRA_INCLUDES})

	# supplement extra library files
	get_subdir_var (GMT_EXTRA_LIBRARIES SUPPL_EXTRA_LIBS ${GMT_SUPPL_DIRS})
	list (APPEND GMT_OPTIONAL_LIBRARIES ${GMT_EXTRA_LIBRARIES})
```

However, these are executed too late, I think, as the CMakeLists.txt file already have said

```
target_link_libraries (gmtlib
	${NETCDF_LIBRARIES}
	${GMT_OPTIONAL_LIBRARIES}
	pslib)
```

before we get down to the `if (BUILD_SUPPLEMENTS)` section.  However, it looks like the include files are found this way; the errors I get is during linking and the given library is not found.

Perhaps there is another way to have these settings be activated inside the if BUILD_SUPPLEMENTS section?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,6 @@ cmake/ConfigUser.cmake
 cmake/ConfigUserAdvanced.cmake
 share/coast
 share/dcw
+src/newsuppl*
 **/gmt.history
 .DS_Store
--- a/cmake/ConfigUserAdvancedTemplate.cmake
+++ b/cmake/ConfigUserAdvancedTemplate.cmake
@@ -194,7 +194,8 @@
 # List extra sub-dirs of 'src' with a CMakeList.txt to build non-module codes
 # that link against the full gmt libs (not just the API; for building codes
 # that only need the GMT API, see the gmt-custom project).
-#set (EXTRA_BUILD_DIRS newsuppl1 newsuppl2)
+# These supplemental modules end up in supplements.so like the GMT supplements.
+#set (EXTRA_BUILD_DIRS newsuppl1 newsuppl2 ...)
 
 # List extra new supplemental modules for testing without adding them to the module list
 #set (EXTRA_MODULES_SUPPL newsuppl1.c newsuppl2.c)
--- a/cmake/ConfigUserAdvancedTemplate.cmake
+++ b/cmake/ConfigUserAdvancedTemplate.cmake
@@ -194,7 +194,7 @@
 # List extra sub-dirs of 'src' with a CMakeList.txt to build non-module codes
 # that link against the full gmt libs (not just the API; for building codes
 # that only need the GMT API, see the gmt-custom project).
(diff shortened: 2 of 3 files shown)


seisman:
Why do you link the libmbio.dylib to the gmt core library? I think it's OK to link to the supplement library only, i.e. 
```
target_link_libraries (supplib
        gmtlib
        pslib
        ${GMT_EXTRA_LIBRARIES}
)
```

seisman:
BTW, newsupple1 and newsuppl2 will be built into the single supplement.so library. Is this what we want?

PaulWessel:
No need to link it to the core gmt - I am just confused.  I will try the supplib - that is what I was hinting at.
As for your other question: I think in general we would want these to be separate shared libs.  If we add a real gmt supplement we just add it to the current list of suppl dirs that go into supplements.so.  Perhaps we need a cmake parameter in those suppl dirs that indicates a separate .so is desired [Default is to use supplements.so]

PaulWessel:
This did not work:

```
target_link_libraries (supplib
        gmtlib
        pslib
        ${GMT_EXTRA_LIBRARIES}
)
```

PaulWessel:
Typo, take that back.  WOrks!

PaulWessel:
This limited change allows for one thing only: The ability to add one or more src subdirectories called new_suppl?, where ? is an integer.  These must have CMakeFiles like the other GMT supplements but these can also contain two special settings:

```
set (SUPPL_EXTRA_INCLUDES path-to-an-include-dir)
set (SUPPL_EXTRA_LIBS path-to-a-library)
```

The include files and library are then used in building supplements.so.  The special directory names means they are ignored by git.  They could also just be symbolic links to the actual directory elsewhere (that is how I tested things).

Note that all modules in new_suppl? end up in supplements.so - it is not a solution for building separate *.so files.  That capability we hope to add at some point but this PR will make it simpler to explore potentially new and approved GMT supplements.
See if you are OK with these changes, @seisman.

PR #3105, opened 2020-04-15
Disable GDAL-created auxilliary files

PaulWessel:
WHen GDAL creates rasters that cannot hold geographic metadata it defaults to writing *.xml files.  We don't want that in **grdimage**, hence this PR. @joa-quim should test it on Windows as I am assuming _putenv is available.


joa-quim:
Test is for tomorrow

PR #3114, opened 2020-04-16
Allow -Vd to debug modern usage

PaulWessel:
Commands like `gmt plot -? -Vd` would refuse to show the moderm mode syntax since the **-Vd** messed up the option count.  I now deal with this so that developers can run the synopsis cases with full debug.  However, I cannot get `gmt plot -Vd `to not complain, but this is not a likely command; developers can run `gmt plot -? -Vd` to test that case.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -642,6 +642,8 @@ GMT_LOCAL void gmtapi_check_for_modern_oneliner (struct GMTAPI_CTRL *API, const 
 	}
 
 	head = GMT_Create_Options (API, mode, args);	/* Get option list */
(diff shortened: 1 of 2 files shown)


seisman:
`gmt plot -+ -Vd` still doesn't work.

PaulWessel:
Works for me in Xcode debugger but not on command line (yet) because macOS seems to have a long memory of what is in a shared library (?).

seisman:
It doesn't work for me, too.

PaulWessel:
It has happened before that I make certain changes to the API and it works when running Xcode but still fails when running command line.  Not sure what to do but reboot.  I will do that as a test.

PaulWessel:
Try now.  I had a ~/.gmt/sessions/gmt6.1 directory so it already thought it was a modern mode session...

PR #3115, opened 2020-04-16
doc: Fix grdimage documentation

seisman:
The documentation says:
```
Append a +c<options> string where options is a list of 
one or more concatenated number of GDAL -co options.
```

What does the "GDAL -co options" mean? Typo?


PaulWessel:
No, I think many GDAL executables have an option called **-c**o that takes these types of args.

seisman:
Got it. I also make **-co** bold.

PR #3119, opened 2020-04-17
Change geographic grid to Cartesian

PaulWessel:
When a geographic grid has its region extended in grdcut and elsewhere, we must change it to Cartesian if it exceeds valid limits on longitude or latitude.
The immediate need is in grd2kml where the quadtree algorithm requires me to extend the global grid to -180/180/-180/180 for a square grid.  It then needs to loose its geographicity so that other modules dont go crazy and complain about latitudes out of range.  This happens via a call to GMT_grdcut, hence the change is in that module where -N can adjust the grid region and fill in NaNs.

### DIFF
--- a/src/gmt_grdio.c
+++ b/src/gmt_grdio.c
@@ -1891,6 +1891,22 @@ int gmt_decode_grd_h_info (struct GMT_CTRL *GMT, char *input, struct GMT_GRID_HE
 	return (int)uerr;
 }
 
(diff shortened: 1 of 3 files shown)


PaulWessel:
Anyone around to approve this one, @joa-quim or @seisman ?

PaulWessel:
Another one bites the dust...

PaulWessel:
yes, and beat you to it.

Issue #3090, opened 2020-04-12
Cannot build mbgmt anymore on Win


joa-quim:
Either #2929 or a subsequent change had broken it. Now I can only add the MB to the ``gmtsupplements``. I still manage to create a ``mbgmt_wXX.dll`` but it has nothing exported in it.

seisman:
Any instructions to build mbgmt on macOS?

joa-quim:
There is but it's not the same thing. My build on Windows is unique (in the sense that I'm the only one crazy enough to try to maintain it without any support). But it doesn't really matter if its MB or any other supplement that depend in other external libs that the GMT ones (and its dependencies).

PaulWessel:
I should try to build my extra supplement [GSFLM](http://www.soest.hawaii.edu/PT/GSFML/) for handling data by the same name.  I think the initial idea with the CMake changes was to make it simple for external supplements to "plug in" to the GMT build but produce separate file.so or file.dll that can be placed in the plugin, rather than go into supplements.so.

joa-quim:
Yes, but things turned out to be the other way around.

PaulWessel:
I will try to have a look.  I did get my GSFML cmakelist.txt file to build gsfml.so but it almost duplicated all of the gmt one.  I will give it a try but not sure when.  Too many shits to sort out first, including grd2kml.  At least grdvector is now on solid grounds.

seisman:
After #3092 and #3111, it should be straightforward to add a new supplement package to GMT. Anyone can build a src subdirectory into a supplement library, as long as there is a CMakeLists.txt in the subdirectory which contains some CMake settings.

@joa-quim For your mbgmt package, you may try a CMakeLists.txt like this (you may need to change some details since I know nothing about mbgmt). Note that you don't need to set EXTRA_MODULES_SUPPL and EXTRA_INCLUDE_NEWSUPPL anymore.
```
set (SUPPL_NAME mb)
set (SUPPL_LIB_NAME mb)
set (SUPPL_PROGS_SRCS mbcontour.c mbswath.c mbgrdtiff.c mbclean_j.c mbgrid_j.c mbimport.c
	mbgetdata.c mbflags.c mbset.c mbinfo.c mbareaclean.c gmtmbgrid.c mbprocess.c)
set (SUPPL_LIB_SRCS ${SUPPL_PROGS_SRCS} mbgmt_module.c)
[...]

Issue #2917, opened 2020-03-15
EXTRA_MODULES_SUPPL doesn't work as documented


seisman:
The `cmake/ConfigUserAdvanced.cmake` file documents that the variable contains a list of C codes.

https://github.com/GenericMappingTools/gmt/blob/758119cee25740c7f2cdc43437b7b9c7c697e189/cmake/ConfigUserAdvancedTemplate.cmake#L202-L203

But, in the `src/CMakeLists.txt`, the variable is used in the other way. The `include` directive in CMake expects to load and run a CMake script rather than a C code.

https://github.com/GenericMappingTools/gmt/blob/758119cee25740c7f2cdc43437b7b9c7c697e189/src/CMakeLists.txt#L489-L495

We have another three cmake variables:
- **EXTRA_MODULES**: list of C codes to be included into GMT's core modules. Works as documented.
- **EXTRA_BUILD_DIRS**: list of directories with a CMakeLists.txt file, just like seis. The codes in the directories will be added as GMT's supplemental modules.
- **EXTRA_INCLUDE_NEWSUPPL**: the variable is not documented.

PaulWessel:
We should revise these things.  For most new modules we usually do a branch anyway until it is ready to merge, so EXTRA_MODULES is not that useful.  What I think would be useful is this:

**EXTRA_BUILD_DIRS**: directories with CMakeLists.txt like other supplements but these should build a separate project.so file, one per subdirectory.  This would simplify how other developers can leverage GMT to build new supplements without having to manage their own build system.

Only our supplements should end up in supplements.so; the rest should be individual shared libraries.

seisman:
@joa-quim, I think some of the variables are used in your mbsystem plugin, right?

joa-quim:
The ones I need for MB are: ``EXTRA_MODULES_SUPPL`` and ``EXTRA_INCLUDE_NEWSUPPL``

PaulWessel:
So in your case, do the MB modules end up in supplements.so then?  I think they should go in mbsystem.so (or dll) like on the Unix side (the MB install script builds these).

joa-quim:
No, they end up in ``mbgmt_w64.dll``

PaulWessel:
Great, so do you place the mb dir inside gmt/src or do you link, or what? This is the capability I would like for other dirs.

joa-quim:
I put it (the installer does that) in ``bin/gmt_plugins``
My setting is something like this
```
[...]

PR #3125, opened 2020-04-17
cmake: Allow renaming non-GMT supplement DLLs on Windows

seisman:
Add the following line to `src/newsuppl/CMakeLists.txt` to rename DLLs on Windows:
```
set (SUPPL_DLL_RENAME "newsuppl_w64")
```

I aslo clean up two cmake variables EXTRA_MODULES_SUPPL and EXTRA_INCLUDE_NEWSUPPL,
since they are no longer used.

Address #2917 and #3090


seisman:
@joa-quim Please try it for your mbgmt.

joa-quim:
That wouldn't help much because would be the same as naming the supplement like that to start with.
But I made it dependent  dependent of a variable and it worked

```
set (SUPPL_DLL_RENAME ${MB_DLL_RENAME})
```

PR #3128, opened 2020-04-17
cmake: Rename CMake variable BUILD_EXTRA_DIRS to SUPPL_EXTRA_DIRS

### LINKED ISSUE #2917: EXTRA_MODULES_SUPPL doesn't work as documented

seisman:
Close #2917.


joa-quim:
Can we please stop renaming variables? I have next to spend hours debugging why my builds break.

seisman:
What you need to do is renaming BUILD_EXTRA_DIRS to SUPPL_EXTRA_DIRS in your own cmake/ConfigUser.cmake file.

joa-quim:
Yes, but before finding that I had to spend time and . trying to figure out why the MB was broken again, and all of that because of a variable rename that was not needed at all. 
It's not fun to find that older mex and julia tests that I don't run so often are broken because used files in the server had been renamed or deleted.

PaulWessel:
Well, that is a good point.  Perhaps when we make changes to things like Cmake files we should require approval from both of us before it is merged.

Issue #3130, opened 2020-04-18
Simplify or eliminate the C stubs needed for supplemental modules


PaulWessel:
**Description of the desired feature**

For plugins used in GMT, one must prepare files similar to gmt_core_modules.[ch], but for the custom supplement.  In GMT, we create this via the command `gmt_make_module_src.sh core`, which builds the necessary source files - these are also maintain in the repo.  This is not very new developer-friendly, prone to scripting mistakes (that script is pushing 500 lines), and seems very repetitive for different shared libraries.  I will try to place as much of these items in an abstract way into the regular gmt library and hope that a developer only has to specify a few things in their CMakeLists.txt file.  I will open a branch to explore this goal.  I would things to work this way:

Instead of separate function called gmtlib_core_module_list_all, gmtlib_supplements_module_list_all, gmtlib_gsfml_module_list_all, etc. there should only be a single function in the GMT library called gmtlib_module_list_all and it should accept the Gmt_moduleinfo struct pointer from each shared library.  Supplement maintainers would then only be responsible for a small include file with specific information for each of their modules.

PaulWessel:
This is close to being completed via #3133.  A few questions remain:

Truly external shared libraries, such as GSFML and MB-System, may have very different build systems.  However, if the supplements follow a few simple design rules they should all be able to tap into the GMT CMake build (?).

**Design Rules**

1. Each module.c file must set these preprocessor parameters (here example from psclip.c):

```
#define THIS_MODULE_CLASSIC_NAME    "psclip"
#define THIS_MODULE_MODERN_NAME     "clip"
#define THIS_MODULE_LIB             "core"
#define THIS_MODULE_PURPOSE         "Initialize or terminate polygonal clip paths"
#define THIS_MODULE_KEYS            "<D{,>X},C-("
```

Depending on integration with GMT, the KEYS can be set to blank "".  For most external supplements the CLASSIC and MODERN names will be identical.  The LIB names the external shared library group.

2. Each shared library needs to configure two files:

- gmt_suppl_moduleinfo.h: The structure contents generated from the above preprocessor parameters.
- gmt_suppl_glue.c: Configured from GMT's gmt_glue.c.in with SUPPL_NAME set to the name of the supplement (here suppl).

[...]

PR #3133, opened 2020-04-18
Simpler gmt_*_module.[ch] files

PaulWessel:

I have abstracted the key part of the various lookup functions and added them to the gmt_api.c.  Now, the core or supplement-specific look-up function simply are shells that pass in the relevant array of module information into the main gmtapi functions.  This reduced the code for gmt_module_*.[ch] by 22%.


PaulWessel:
I also think we could avoid the `#ifdef BUILD_SHARED_LIBS` branch in gmt_core_module.h by only keeping the one with the list of function names, since they are only used in a call if **BUILD_SHARED_LIBS** is defined, otherwise it does not matter, no? 
Builds for me and tests run, on macOS.  So this is 35% savings in lines of code.

PaulWessel:
I will simplify the shell script next, it ought to be even less code than 65%.

PaulWessel:
We need a test on Windows here of course since we are messing with the dark shared library side.

joa-quim:
I breaks my MB supplement again

PaulWessel:
Can you be more specific? It did not break my gsfml supplement even though I did not make any changes there, only to GMT.  My changes should not have any effect on other supplements since you are not calling things in gmt_core_*.c etc.

PaulWessel:
There are still parts of the *module*.h files that are created by the script but that probably could just be part of a static include file.  For instance, these prototypes:

```
/* Prototypes of all modules in the GMT core library */
EXTERN_MSC int GMT_psbasemap (void *API, int mode, void *args);
EXTERN_MSC int GMT_begin (void *API, int mode, void *args);
...
```

do not change, except when we very infrequently introduce a new module.  It is not a problem for us to maintain this include file.  What does change are things like the define statements at the top of each module: **THIS_MODULE_PURPOSE**, **THIS_MODULE_KEYS**, **THIS_MODULE_NEEDS**.  For that reason we need a script to create the module structure.  However, if we have the list of modules in that static file, the script would be much simplified since we can use the static file for the list of current modules and hence source files.
The question is could we make this a Cmake build thing: No gmt_core_module.c file in git but built on the fly.

joa-quim:
> since you are not calling things in gmt_core_*.c etc.

Yes, I am. I have a ``mbgmt_module.c|h``. I have more MB modules with keys that use in Mirone.

But I have restored them. Now it builds again ... till next

PaulWessel:
I know you have mbgmt_module.c|h but none of those are calling things in gmt_core_module.c, right?

PaulWessel:
I would like to break up the gmt_supplements_module.c|h stuff since it is an artificial collection of _some_ supplemental dirs.  To simplify life for other supplements (mg, gsfml) it would be better if each supplement had these things and that cmake helped.  We can still place stuff in a single supplements.so for the GMT approved supplements, but not make the code part special via gmt_supplements_*.c|h.

PaulWessel:
Take that partially back: I think the only good solution for supplemental dirs will result in individual *so/*.dll files.  If so then we are truly liberated from maintaining gmt_supplements_*.? files in the core directory.

joa-quim:
No, they don't. But things were not working and didn't know why. Perhaps it was only the BUILD_EXTRA_DIRS thing.

PaulWessel:
OK, I will use this branch to create modules.h files for core and for each supplement dir, then a small script that builds the glue.

PaulWessel:
Hi @seisman, each of the supplement CMakeLists.txt files have lines like this:

`set (SUPPL_NAME gsfml)`

I would like to add the same to the main src file with name "core", and then have gmt_shared.h.in files in each source directory containing statements like this:

`EXTERN_MSC void gmtlib_@SUPPLE_NAME@_module_show_all (void *API);`

in order to automatically create the functions via cmake.  Do you think the 

`configure_file (gmt_shared.h.in gmt_shared.h)`

in each directory (src, src/*) properly configure and build the correct sources?  I am trying to find out that answer before doing all the work and really finding out...

seisman:
Yes, `configure_file` can give exactly what you need.

PaulWessel:
Here is my plan for eliminating users and developers from dealing with things like gmt_make_module_src.sh and the hundreds of lines of code it produces:

1. Each src dir should have a fixed include file called gmt_shared.h.  It will list the prototypes for the modules in that directory, like for src:

```
EXTERN_MSC int GMT_psbasemap (void *API, int mode, void *args);
EXTERN_MSC int GMT_begin (void *API, int mode, void *args);
...
```

2. Each gmt_module.h file includes gmt_glue.h. It is auto-configured from gmt_glue.h.in and contains the four function prototypes needed to access the shared library; here is just one:

`EXTERN_MSC void gmtlib_@SUPPLE_NAME@_module_show_all (void *API);`

3. Each src dir also contains a file gmt_glue.c.  It is configured from gmt_glue.c.in and it builds the functions exported by gmt_glue.h.  Each gmt_glue.c file states `#include gmt_moduleinfo.h` which must be built from the parameters **THIS_MODULE_PURPOSE**, **THIS_MODULE_KEYS**, and **THIS_MODULE_NEEDS** and the list of modules in gmt_shared.h.

There are two steps that are nontrivial or unclear at this point.

- The building of gmt_moduleinfo.h per directory:  This is currently a part of what gmt_make_module_src.sh does, but can it be recast as Cmake module?  I can write this in bash first and then we can see how translatable it is.  Cmake has lots of functions and string tools that FLorian used when he translated all the old code that maintained gmtmath.h.in (now gone).
- RIght now there is some unholy dependence of GMT core on the supplements prototypes: gmt_supplements_module.h is included by gmt_dev.h.  But why? The MB or the GSFML prototypes are not included by gmt_dev.h, yet things are working fine.  We need to break this chain.

Do @joa-quim or @seisman see anything flawed in this scheme? I think if this can work we avoid having to worry about running things like gmt_make_module_src.sh manually, and can much more easily document what it takes to make a new supplement dir (just copy gmt_shared.h and gmt_glue.?.in from any supplement and edit the first file.

PaulWessel:
> Yes, `configure_file` can give exactly what you need.

Perhaps it is possible to do

`configure_file (../gmt_shared.h.in gmt_shared.h)`

for the subdir CMakeLists.txt files so that there is no need to place a gmt_shared.h.in files (and gmt_shared.c.in) in each directory, only in src?

seisman:
> > Yes, `configure_file` can give exactly what you need.
> 
> Perhaps it is possible to do
> 
> `configure_file (../gmt_shared.h.in gmt_shared.h)`
> 
> for the subdir CMakeLists.txt files so that there is no need to place a gmt_shared.h.in files (and gmt_shared.c.in) in each directory, only in src?

I'm a little confused. Are you talking about gmt_shared.h or gmt_glue.h?

PaulWessel:
Sorry (names are changing in my head), above I mean gmt_glue.?.in since gmt_shared.h is a static file we maintain manually.  The idea is to use the *.in files from src repeatedly in each subdir to make slightly different include files and C files.

PaulWessel:
BTW, probably for similar reasons, I see this having snuck into src/CMakeLists.txt:

`set (GMT_MBSYSTEM_SRCS gmt_mbsystem_module.c gmt_mbsystem_module.h)`

That should not be necessary either.

seisman:
> Sorry (names are changing in my head), above I mean gmt_glue.?.in since gmt_shared.h is a static file we maintain manually. The idea is to use the *.in files from src repeatedly in each subdir to make slightly different include files and C files.

Yes, it's doable.

PaulWessel:
OK, close to working but family insist on going swimming.  Back in the afternoon. I committed all my changes to this branch in case you want to try to see.  I still fail on a few externs but very close.  No longer using gmt_core_module.?, but gmt_modules.h (static), gmt_glue.?.

PaulWessel:
PS. I have not touched the supplements yet on this since the above scheme means breaking up supplements.so and the need to find those individual so files.  So only core testing for now.

seisman:
You didn't commit the gmt_shared.h file.

PaulWessel:
Sorry, gmt_shared is gmt_modules.h now.  Not sure why I get this though:

```
Undefined symbols for architecture x86_64:
  "_gmtlib_core_module_group", referenced from:
      _gmtapi_get_module_group in gmt_api.c.o
  "_gmtlib_core_module_keys", referenced from:
      _gmtapi_get_module_keys in gmt_api.c.o
ld: symbol(s) not found for architecture x86_64
```

Those are both being exported via gmt_glue.? and gmt_glue is included in gmt_api.c

PaulWessel:
I wonder if we should abandon the static build` #ifndef BUILD_SHARED_LIBS` since we are building shared libraries and at any rate it is not working anyway (#2436)?  Anyone out there care about the possibility of one day building static?

seisman:
> Sorry, gmt_shared is gmt_modules.h now. Not sure why I get this though:
> 
> ```
> Undefined symbols for architecture x86_64:
>   "_gmtlib_core_module_group", referenced from:
>       _gmtapi_get_module_group in gmt_api.c.o
>   "_gmtlib_core_module_keys", referenced from:
>       _gmtapi_get_module_keys in gmt_api.c.o
> ld: symbol(s) not found for architecture x86_64
> ```
> 
> Those are both being exported via gmt_glue.? and gmt_glue is included in gmt_api.c



> Sorry, gmt_shared is gmt_modules.h now. Not sure why I get this though:
> 
> ```
> Undefined symbols for architecture x86_64:
>   "_gmtlib_core_module_group", referenced from:
>       _gmtapi_get_module_group in gmt_api.c.o
>   "_gmtlib_core_module_keys", referenced from:
>       _gmtapi_get_module_keys in gmt_api.c.o
> ld: symbol(s) not found for architecture x86_64
> ```
> 
> Those are both being exported via gmt_glue.? and gmt_glue is included in gmt_api.c

These functions are defined in `gmt_core_module.c`. You need to add it to **GMT_LIB_SRCS**.

seisman:
> Sorry, gmt_shared is gmt_modules.h now. Not sure why I get this though:
> 
> ```
> Undefined symbols for architecture x86_64:
>   "_gmtlib_core_module_group", referenced from:
>       _gmtapi_get_module_group in gmt_api.c.o
>   "_gmtlib_core_module_keys", referenced from:
>       _gmtapi_get_module_keys in gmt_api.c.o
> ld: symbol(s) not found for architecture x86_64
> ```
> 
> Those are both being exported via gmt_glue.? and gmt_glue is included in gmt_api.c

gmt_glue.c is in the build directory. You need to change `gmt_glue.c` to `${CMAKE_CURRENT_BINARY_DIR}/gmt_glue.c`.

src/CMakeLists.txt
```diff
@@ -387,7 +388,7 @@ set (GMT_LIB_SRCS block_subs.h common_byteswap.h common_math.h
 	gmt_stat.c gmt_support.c gmt_regexp.c gmt_vector.c gmt_parse.c inset.c
 	gmt_error_codes.c gshhg_version.c gmtread.c gmtwrite.c common_math.c
 	common_runpath.c common_sighandler.c common_string.c compat/qsort.c
-	gmt_sharedlibs.c gmt_core_module.c gmt_init.c gmt_map.c gmt_plot.c
+	gmt_sharedlibs.c gmt_glue.c gmt_init.c gmt_map.c gmt_plot.c
```
seisman, line 391:
```suggestion
	gmt_sharedlibs.c ${CMAKE_CURRENT_BINARY_DIR}/gmt_glue.c gmt_init.c gmt_map.c gmt_plot.c
```

PaulWessel:
Rethinking the current mode of operations.  Because **KEYS**, **PURPOSE**, and **NEEDS** do change as we add new options or fix bugs, we want a system that is fool-proof, i.e., when these change then any functions that depend on these automatically gets updated.  I think this is a good way to prevent hard-to-trace bugs.  So because of that need, we do this:

1. Define several module-specific parameters via **THIS_MODULE_*** statements in each module.
2. Run kludgy bash scripts to extract this data and build a static structure.
3. Include this static structure and build functions that returns parts of its content for specified module.

While we do indeed need the ability to get to the parameters for a given structure, surely this way is not the only way.  It makes us dependent on a very ugly script which cannot be simplified greatly because we need to harvest several lines per module, then output a single line with all the stuff per module.  Here are some other thoughts - you may have more.

First, lets remember that most of the static strings are fed to this function that is the first to be called in any module after we get past any usage message request:

`if ((GMT = gmt_init_module (API, THIS_MODULE_LIB, THIS_MODULE_CLASSIC_NAME, THIS_MODULE_KEYS, THIS_MODULE_NEEDS, NULL, &options, &GMT_cpy)) == NULL) bailout (API->error); /* Save current state */`

Thus, as soon as the module is called, it knows its own keys, needs, name, etc.  This duplicates the same information in the moduleinfo structure built by the script.  The _gmt_init_module_ structure uses many of these arguments to make decisions.  So for basic GMT use there is no real need for the static structure.  However, the whole point of the structure and the collection of info relates to external usage of GMT.  In particular, it originated with the development of the GMT/MATLAB toolbox.  Here, we needed to figure out which Matlab arguments should be hooked up as input memory files to gmt modules, and that depended on knowing which module options accepted input and output, what the expected input or output was, and more.  So the API function _GMT_Encode_Options_ was born:

`struct GMT_RESOURCE * GMT_Encode_Options (void *API, const char *module_name, int n_in, struct GMT_OPTION **head, unsigned int *n);`

GMT/MATLAB calls it to translate a preliminary GMT command line into an actual one, inserting virtual files here and there given the above knowledge. I believe the Julia package uses the same scheme.  So what does _GMT_Encode_Options_ look like.  Well, it is ugly and does lots of sleuthing, but one of the first things it does is this:

```
/* 0. Get the keys for the module, possibly prepend "gmt" to module if required, or list modules and return NULL if unknown module */
if ((keys = gmtapi_retrieve_module_keys (V_API, module)) == NULL) {	/* Gave an unknown module */
	if (GMT_Call_Module (V_API, NULL, GMT_MODULE_PURPOSE, NULL))	/* List the available modules */
		return_null (NULL, GMT_NOT_A_VALID_MODULE);
	return_null (NULL, GMT_NOT_A_VALID_MODULE);	/* Unknown module code */
}
```

In other words, it retrieves the keys specific to the module in question. The gmtapi_retrieve_module_keys then gets the keys by trying the various core and shared libraries and pull out the entry from the static structure array.

Seems to me that if we just declare an external function in each of the module C files called

`char * get_<module>_keys () { return (strdup (THIS_MODULE_KEYS)); }`

then we could have the _gmtapi_retrieve_module_keys_ just do what _GMT_Call_Module_ does: build the function name from gmt_%s_keys and the module name and called that shared function.  Voila: No need for the static moduleinfo array and the ugly script.

I think this is the way to go so we can eliminate all the moduleinfo baggage.

PaulWessel:
Regarding static vs shared builds.  Allowing static is the only reason why we need the pointer to the functions, and they are only used in gmtlib_module_lookup:

	`return (M[module_id].p_func);`

Eliminate static and we can eliminate these function pointers.

PaulWessel:
Since this is an experimental branch, I think it is OK for me to delete option to build static since it never worked anyway and just makes things more complicated.  Any objection, @seisman ?

seisman:
> Since this is an experimental branch, I think it is OK for me to delete option to build static since it never worked anyway and just makes things more complicated. Any objection, @seisman ?

OK for me. I believe very few people want static builds.

PaulWessel:
Especially not a broken static build...

PaulWessel:
Hi @seisman, any clever way to do this:

for each module file, insert the single line

`EXTERN_MSC char * gmtlib_get_<module>_keys () { return (strdup (THIS_MODULE_KEYS)); }`

followed by a blank line right before the main function declaration

`int GMT_grdedit (void *V_API, int mode, void *args) {`

Maybe a script loop over the module names then

`sed -i .bak '/int GMT_$module/a EXTERN_MSC char * gmtlib_get_<module>_keys () { return (strdup (THIS_MODULE_KEYS)); }\n/' $module.c`

Does that looks right to you (asses module is blockmean psxy etc.  I will do the same in each of the subdirs.

PaulWessel:
Because of the single quotes I think I need to write that sed command to a file and then use -f.

seisman:
Try this script:
```
module=psxy
sed -i "" "/int GMT_${module}/i \\
	EXTERN_MSC char * gmtlib_get_psxy_keys () { return (strdup (THIS_MODULE_KEYS)); }\\
	\\
	" $module.c
```

PaulWessel:
Thanks, was just going to post that things worked for me after getting gsed.

PaulWessel:
So that worked fine.  However, now I realize that the

```
/* Pretty print all shared module names and their purposes for gmt --help */
void gmtlib_core_module_show_all (void *API) {
	gmtlib_module_show_all (API, modules, GMT_CORE_STRING);
}
```

functions are in fact fed the static structures so that we can loop over modules and print their purpose in gmt.c without have to call a shared library function for each module to report its purpose.  Aarghh!  Perhaps there is no good way to avoid building gmt_moduleinfo.h for each directory:

```
{"basemap", "psbasemap", "core", "Plot base maps and frames", ">X},>DA@AD)"},
{"begin", "begin", "core", "Initiate a new GMT modern mode session", ""},
{"blockmean", "blockmean", "core", "Block average (x,y,z) data tables by mean estimation", "<D{,>D},GG),A->"},
...
```

Maybe you can see a simple way to auto-create this type of file by cmake for each src directory, @seisman ?

seisman:
> Perhaps there is no good way to avoid building gmt_moduleinfo.h for each directory:
> 
> ```
> {"basemap", "psbasemap", "core", "Plot base maps and frames", ">X},>DA@AD)"},
> {"begin", "begin", "core", "Initiate a new GMT modern mode session", ""},
> {"blockmean", "blockmean", "core", "Block average (x,y,z) data tables by mean estimation", "<D{,>D},GG),A->"},
> ...
> ```
> 
> Maybe you can see a simple way to auto-create this type of file by cmake for each src directory, @seisman ?

It's possible to loop over **GMT_PROGS_SRCS**, read each C code (`file (READ)`, extract these values using regex (`string (MATCH)`), and then write it to a file `file (WRITE)`. I can try it if you think it's OK.

PaulWessel:
Give me a few minutes to run the tests.  I eliminated the option for static builds and removed the `&GMT_psxy `etc args from the static structure.  Will let you know if all pass.

PaulWessel:
OK, committed and no failures.  See if you can generate the same format as the current gmt_moduleinfo.h.  I guess we would then do this per src dir and get gmt_moduleinfo.h files in all the subdirs during the build.  This would require some more work to replace gmt_supplements_*.c with the glue files and these include files and possibly some more work in gmtapi.c but that is fine.

seisman:
@PaulWessel I need a regex to match **THIS_MODULE_KEYS**. What are the allowed characters in this macro?

PaulWessel:
Quite a few.  Certainly upper case and lower case letters. + and -, commas, open and close (, {, <, question mark, equal, @, and signle digits.

PaulWessel:
Hi @seisman, I have done the basics: 

1. Made gmt_suppl_glue.[ch].in after gmt_glue.[ch] and gmt_suppl_modules.h after gmt_modules.h and the contents of the old gmt_supplements_module.h.
2. Include gmt_suppl_modules.h from gmt_dev.h and gmt_suppl_glue.h from gmt_api.c.
3. Removed gmt_supplements_module.[ch] and the old gmt_core_module.c (forgot last time)
4. Updated the CMakefile to configure the two new files.

All compiles but currently the configured gmt_suppl_glue.[ch] file says "core" instead of "suppl" in the function names.

PaulWessel:
While we can have one supplements.so, we should be able to avoid the current way of including gmt_suppl_*.h in gmt_dev.h and gmt_api.c.  For instance, the truly external supplements like gsfml obviously has no include files in the gmt core.

src/CMakeLists.txt
```diff
@@ -430,12 +431,39 @@ add_custom_command (OUTPUT gmt_dimensions.h
...
 # Configure header files to pass some of the CMake settings to the source code
 configure_file (gmt_config.h.in gmt_config.h)
 configure_file (psl_config.h.in psl_config.h)
 configure_file (gmt_version.h.in gmt_version.h)
 configure_file (isogmt.in isogmt${GMT_INSTALL_NAME_SUFFIX} @ONLY)
 configure_file (gmt-config.in gmt-config${GMT_INSTALL_NAME_SUFFIX} @ONLY)
+configure_file (gmt_glue.h.in gmt_glue.h)
+configure_file (gmt_glue.c.in gmt_glue.c)
+configure_file (gmt_suppl_glue.h.in gmt_suppl_glue.h)
+configure_file (gmt_suppl_glue.c.in gmt_suppl_glue.c)
```
seisman, line 466:
```suggestion
set (SUPPL_NAME suppl)
configure_file (gmt_suppl_glue.h.in gmt_suppl_glue.h)
configure_file (gmt_suppl_glue.c.in gmt_suppl_glue.c)
```
I think we have to reset SUPPL_NAME.

PaulWessel:
Had to be supplements instead of suppl since we build the function calls from the name of the shared library (supplements).  Seems to work now.

PaulWessel:
I think there should not be a need for gmt_modules.h and gmt_suppl_modules.h.  We never offer a list of modules to choose from.  Users type the name of the module and we check to see if we can find it.  We never call GMT_grdimage from anywhere, except in GMT_Call_Module.  I think some of this dates back to the static days.

PaulWessel:
Hi @seisman , I am trying to get rid of gmt_modules.h. I have done this:

1. Comment out the gmt_modules.h and gmt_suppl_modules.h from gmt_dev.h
2. This lead to complaints about GMT_colorbar calling undefined GMT_psscale.  So I edited all those ps*.c files and moved the modern name wrapper to the end of the file.  Fixed.
3. I took out the GMT_CORE_STRING and made that configurable via cmake.  So CMakeLists.txt states SUPPL_PURPOSE and we pick that up in the *c.in files.  I set it again just before configuring the supplement file.

Compiles fine, but gmt --help does not list the supplement modules (yet).

seisman:
The Windows CI fails. It seems GMT can not access any modules now on Windows.

seisman:
Please let me know if you need help with the conflicting files.

PaulWessel:
Sure, please do - I am on a zoom with the Provost and other chairs...

seisman:
Done.

PaulWessel:
Update: I thought I fixed this yesterday but not committed perhaps.  Need 
`set (SUPPL_NAME supplements)`
not
`set (SUPPL_NAME suppl)`

since supplements is the name of the shared lib and we are making function calls based on the name of the shared lib so they have to match.

All works for me - and I commented out the use of gmt_modules.h and gmt_suppl_modules.h in CMakeLists.txt.  Please confirm all is well, @seisman and then we can ask JL to test.

seisman:
It fails on Windows. See the CI report.

PaulWessel:
Right now all the module files simply do things like this:

```
int GMT_pscoast (void *V_API, int mode, void *args) {
...
```

I wonder if we have all the module files do this, e.g., for pscoast:

```
EXTERN_MSC int GMT_pscoast (void *V_API, int mode, void *args) {
...
```

since otherwise these functions are not exported under Windows (and same for GMT_coast in that file).

PaulWessel:
Hi @joa-quim.  Is this usually enough to export a function:

EXTERN_MSC int GMT_pscoast (void *V_API, int mode, void *args) {
...

or must it be two separate steps:

...
EXTERN_MSC int GMT_pscoast (void *V_API, int mode, void *args);
 int GMT_pscoast (void *V_API, int mode, void *args) {
...

joa-quim:
Don't know. I never saw the first form.

PaulWessel:
@seisman, before we change 150 module files, could you please trye the MSC_EXTERN solution in pscoast.c only (two places) and see if the Win test at least pass the call the pcoast?

PaulWessel:
Or I guess it could just have those lines

EXTERN_MSC int GMT_pscoast (void *V_API, int mode, void *args);
EXTERN_MSC int GMT_coast (void *V_API, int mode, void *args);

added up front in pscoast.c

seisman:
```
EXTERN_MSC int GMT_pscoast (void *V_API, int mode, void *args) {
```
This one works.

PaulWessel:
If this commit works I can remove the gmt_modules.h and gmt_suppl_modules.h from src and CMakeLists.txt.  You are not counting on those files for anything like building the moduleinfo, @seisman ?

seisman:
> If this commit works I can remove the gmt_modules.h and gmt_suppl_modules.h from src and CMakeLists.txt. You are not counting on those files for anything like building the moduleinfo, @seisman ?

Yes, they're not used. 

BTW, it seems you removed the CMake codes for building gmt_suppl_moduelinfo.h. Is it intentional?

PaulWessel:
> BTW, it seems you removed the CMake codes for building gmt_suppl_moduelinfo.h. Is it intentional?

No! How did that happen? Can you fix that please?

seisman:
> > BTW, it seems you removed the CMake codes for building gmt_suppl_moduelinfo.h. Is it intentional?
> 
> No! How did that happen? Can you fix that please?

You removed them and also changed the SUPPL_NAME from supplements (correct) to suppl (wrong) in this commit https://github.com/GenericMappingTools/gmt/pull/3133/commits/dfd8b7d4b75dd7ce48889cab0cb9585eaaf3f356. Have no idea why it happened 😄

PaulWessel:
I am doing this with the new-api-lookout branch active and fully updated.  I dont understand.  As I mentioned above I in fact changed GMT_SUPPL to supplements to make it work!

seisman:
I just added the gmt_suppl_moduleinfo.h building back.

PaulWessel:
OK, thanks. I will remove the 2 fils and commented lines in gmt_dev.h and CMakeLists.txt.

PaulWessel:
Only coupling of core lib and supplement is this:

gmt_api.c expects supplements shared library to be called supplements.*
If found we sort list of all found shared libs so that supplements is first after core.

Thus, if 2 year from now we decide to change the name of the supplements shared library to something else, we will have to do the same in gmt_api.c.  One way around would be to add gmt_shared.h.in and have it configured to set a #define string with "@SUPPL_NAME@".
However, later down in CMake we have this:

```
# get supplement library name ["supplements" if undefined]
		get_subdir_var (_suppl_lib_name SUPPL_LIB_NAME ${_dir})
		if (NOT _suppl_lib_name)
			set (_suppl_lib_name "supplements")
```

Best if we only set the name "supplements" once, no?

seisman:
What about defining a CMake variable (e.g. SUPPL_LIB_NAME) in cmake/ConfigDefaults.cmake, so that the variable can be accessed in both gmt_sharedlibs.h.in and src/CMakeLists.txt.

PaulWessel:
Even better. You or me?

seisman:
Leave it to you.

cmake/ConfigDefault.cmake
```diff
@@ -187,10 +187,8 @@ if (NOT DEFINED CMAKE_VERBOSE_MAKEFILE)
 	set (CMAKE_VERBOSE_MAKEFILE false)
 endif (NOT DEFINED CMAKE_VERBOSE_MAKEFILE)
 
-# prefer shared libs over static
-if (NOT DEFINED BUILD_SHARED_LIBS)
-	set (BUILD_SHARED_LIBS true)
-endif (NOT DEFINED BUILD_SHARED_LIBS)
+# Only support shared libs
+set (BUILD_SHARED_LIBS true)
```
seisman, line 194: The old codes are better.
```suggestion
# Currently only support shared libs
if (NOT DEFINED BUILD_SHARED_LIBS)
	set (BUILD_SHARED_LIBS true)	
endif (NOT DEFINED BUILD_SHARED_LIBS)
```

PaulWessel:
Think I got it all.

CMakeLists.txt
```diff
@@ -153,8 +153,8 @@ set (PSL_LIB_NAME ${_prefix}${_name}${_lib_suffix})
 
 # Get name of shared supplemental library
 if (BUILD_SUPPLEMENTS)
-	get_target_property (_name supplements OUTPUT_NAME)
-	get_target_property (_prefix supplements PREFIX)
+	get_target_property (_name ${GMT_SUPPL_LIB_NAME} OUTPUT_NAME)
+	get_target_property (_prefix ${GMT_SUPPL_LIB_NAME} PREFIX)
 	set (GMT_SUPPL_LIB_NAME ${_prefix}${_name}${CMAKE_SHARED_MODULE_SUFFIX})
```
seisman, line 158:
```
set (GMT_SUPPL_LIB_NAME ${_prefix}${_name}${CMAKE_SHARED_MODULE_SUFFIX})
```
This lines changes `GMT_SUPPL_LIB_NAME` to `supplements.so` or `supplements.dll`. We may need two separate names. Why does it work?

PaulWessel: I dont know, my variable changes does not change that fact, no?

seisman: GMT_SUPPL_LIB_NAME with the new value (supplements.so or supplements.dll) is used only in `src/config.h.in`. Probably that's the reason why your changes don't break anything. But it would be better to use different variable names to avoid future confusions.

PaulWessel: Maybe you can fix this since I am not sure what two names.  Also, some simplifications are coming via #3159.

seisman: OK. Will do it in another PR.

PaulWessel: Great, thanks. Dinner for me now.

src/gmt_macros.h
```diff
@@ -34,6 +34,9 @@
  *			GMT MACROS DEFINITIONS
  *--------------------------------------------------------------------*/
 
+/* Used to build gmt module prototypes in gmt_modules.h files */
+#define gmt_M_module_prototype(name) EXTERN_MSC int GMT_ ## name (void *API, int mode, void *args)
+
```
seisman, line 39: This macro is no longer needed.

src/gmt_moduleinfo.h
```diff
@@ -0,0 +1,96 @@
+	{"basemap", "psbasemap", "core", "Plot base maps and frames", ">X},>DA@AD)"},
```
seisman, line 1: gmt_moduleinfo.h is no longer needed.

PaulWessel:
Perhaps @joa-quim could try to build this branch.  It seems to work for WIndows CI builds.

PaulWessel:
Hi @joa-quim, might you be able to test this branch? It passes now for Windows CI (I forgot to remove gmt_make_module_src.sh from CMakeLists.txt yesterday so it failed).  I think we have a short and slick solution to the shared library glue now, but need a Win run before removing the WIP.

PaulWessel:
Hi @seisman, might it be possible to add an if-test in this loop to distinguish the output file based on the vaule of ${SUPPL_NAME)?

```
# generate gmt_${SUPPL_NAME}_moduleinfo.h
get_subdir_var_files (_suppl_progs_srcs SUPPL_PROGS_SRCS ${_dir})
foreach (_prog_src ${_suppl_progs_srcs})
	file (READ ${_prog_src} _src_content)
	string (REGEX MATCH "#define THIS_MODULE_CLASSIC_NAME[ \t]*\"([^\n]*)\"\n" _this_module_classic_name ${_src_content})
	set (_this_module_classic_name ${CMAKE_MATCH_1})
	string (REGEX MATCH "#define THIS_MODULE_MODERN_NAME[ \t]*\"([^\n]*)\"\n" _this_module_modern_name ${_src_content})
	set (_this_module_modern_name ${CMAKE_MATCH_1})
	string (REGEX MATCH "#define THIS_MODULE_LIB[ \t]*\"([^\n]*)\"\n" _this_module_lib ${_src_content})
	set (_this_module_lib ${CMAKE_MATCH_1})
	string (REGEX MATCH "#define THIS_MODULE_PURPOSE[ \t]*\"([^\n]*)\"\n" _this_module_purpose ${_src_content})
			set (_this_module_purpose ${CMAKE_MATCH_1})
	string (REGEX MATCH "#define THIS_MODULE_KEYS[ \t]*\"([^\n]*)\"\n" _this_module_keys ${_src_content})
	set (_this_module_keys ${CMAKE_MATCH_1})
	list (APPEND _moduleinfo "\t{\"${_this_module_modern_name}\", \"${_this_module_classic_name}\", \"${_this_module_lib}\", \"${_this_module_purpose}\", \"${_this_module_keys}\"},")
endforeach (_prog_src ${GMT_PROGS_SRCS})
```

If SUPPL_NAME equals GMT_SUPPL_LIB_NAME then we append to the same moduleinfo.h file, otherwise we create one in the subdirectory instead?  That way, this CMakeLists.txt file when used with the gsfml and similar supplements would build src/gsfml/gmt_gsfml_moduleinfo.h for us. Separately, the src/gmt_gsfml_glue.c is build by the GSFML CMakeLists.txt file from src/gmt_glue.c.in?

joa-quim:
It builds fine but this Julia test fails with (works on master). Looks like a KEY issue.
```
julia> G2=grdfilter(G, filter="m600", distflag=4, inc=0.5, Vd=1);
        grdfilter  -I0.5 -D4 -Fm600
grdfilter [ERROR]: Option -F: filter fullwidth must be nonzero.
```

PaulWessel:
THe grdfilter bug was fixed a few days ago.  it was specifically for median filtering.

joa-quim:
Sorry, don't get that. If I check out again master and rebuild then I have no error.

PaulWessel:
I just merged master into the new-api-lookup branch - perhaps it had not been merged before... Surely you should not get the grdfilter issue now

joa-quim:
Ok, that one went away, but now have
```
pshistogram [ERROR]: Option -J: Only Cartesian scaling available in this module.
ERROR: LoadError: Something went wrong when calling the module. GMT error number = 78
Stacktrace:
```
must see what's the complain

PaulWessel:
OK, also recent.  Refuse to take anything but -Jx or -JX in histogram so perhaps you are calling it with something else?

joa-quim:
No, actually I'm calling with no -J and that used to work
```
julia> D = histogram(randn(1000), I=:o, W=0.1, Vd=1);
         -Io -W0.1
pshistogram [ERROR]: Option -J: Only Cartesian scaling available in this module.
```

PaulWessel:
Strange.  So if no -J is given the gmt_init_module should kick in and set the default. Perhaps it is setting -JQ by mistake and it gets caught.  Will check.

joa-quim:
**-Io** 
> Inquire about min/max x and y after binning. 

so having no -J makes sense.

PaulWessel:
No, that is not it.  I tried this
`gmt pshistogram -B  @v3206_06.txt -R-6000/0/0/1300 -Gorange -Z0 -W200 -pdf map`
and no problem.  So you must do something different.

PaulWessel:
I see - can reproduce:

gmt pshistogram @v3206_06.txt -R-6000/0/0/1300 -Io -W200
pshistogram [ERROR]: Option -J: Only Cartesian scaling available in this module.

Will fix this (in master).

joa-quim:
All julia tests pass again

PaulWessel:
We are good to merge into master... Why is the WIP check in progress, @seisman ?

src/CMakeLists.txt
```diff
@@ -19,6 +19,8 @@
 #
 # src/CMakeLists.txt
 #
+set (SUPPL_NAME core)
```
seisman, line 22: SUPPL_NAME can be "core" or "supplements". I think it's NOT a good name for this variable, since SUPPL_NAME is also used in src subdirectories (e.g. src/seis/CMakeLists.txt). It's better to rename SUPPL_NAME and SUPPL_PURPOSE to other names.

PaulWessel: OK.  Since nothing is really named "core" it is more of a flag. Maybe use SHARED_LIB_PURPOSE for the purpose strings and SHARED_LIB_NAME for the SUPPL_NAME part in src/CMakeLists.txt?

PaulWessel: The idea is that they are both shared libraries but only on is supplemental, so SUPPL is not a good name I agree.

seisman: These two names are good for me.

seisman:
> If SUPPL_NAME equals GMT_SUPPL_LIB_NAME then we append to the same moduleinfo.h file, otherwise we create one in the subdirectory instead? That way, this CMakeLists.txt file when used with the gsfml and similar supplements would build src/gsfml/gmt_gsfml_moduleinfo.h for us. Separately, the src/gmt_gsfml_glue.c is build by the GSFML CMakeLists.txt file from src/gmt_glue.c.in?

Do we still need this implemented?

PaulWessel:
Not sure, maybe.  There are three kinds of "supplements":

1. Official ones in GMT, like seis, spotter... These are built at the same time as GMT.
2. Unofficial ones, like gsfml, that take advantage of the GMT build machinery by setting a symbolic link newsuppl[?] to point to that supplements src dir.
3. Unofficial ones, like mbsystem, that has its own build system that expects GMT to be installed already.

See #3130 for some thoughts.  For type 2 above to work, the idea was to use configure_file with gmt_glue.c.in and make src/newsuppl/gmt_gsfml_glue.c which means there has to be a gmt_gsfml_moduleinfo.h available.  It would be slick if that was handled by GMT CMake, but if it adds to much complication... THen the 3rd option, and possibly the solution for type 3, is to let gmt --build-glue=_supplementname_ create those pieces.

PaulWessel:
I thought first of gmt-config, but it is a script that may not be ideal for all environments (e.g., Windows).

seisman:
Are gmt_glue.c and gmt_moduleinfo.h required for unofficial supplements? As I understand it, if a supplement package don't have gmt_glue.c and gmt_moduleinfo.h, users still can access all modules in the supplement package, except that these modules are not listed in gmt --help or gmt --show-modules. Am I right?

seisman:
The builds break after merging #3165 with following errors:
```
../../src/gmt_api.c:970:73: error: use of undeclared identifier 'GMT_SUPPL_LIB_NAME'
                                for (k = 0; list[k] && strncmp (list[k], GMT_SUPPL_LIB_NAME, strlen(GMT_SUPPL_LIB_NAME)); k++); /* Look for official supplements */
                                                                                                    ^
../../src/gmt_api.c:970:46: error: use of undeclared identifier 'GMT_SUPPL_LIB_NAME'
                                for (k = 0; list[k] && strncmp (list[k], GMT_SUPPL_LIB_NAME, strlen(GMT_SUPPL_LIB_NAME)); k++); /* Look for official supplements */
                                                                         ^
2 errors generated.
```

In commit https://github.com/GenericMappingTools/gmt/pull/3133/commits/d2016b474d0827f9a1553eff836bb199bcadba7e, you changed 
```
for (k = 0; list[k] && strncmp (list[k], "supplements", 11U); k++);
```
to 
```
for (k = 0; list[k] && strncmp (list[k], GMT_SUPPL_LIB_NAME, strlen(GMT_SUPPL_LIB_NAME)); k++);
```

PaulWessel:
Yes, you are right.  There are two things:

1. Visibility via gmt --help or gmt --show-modules.  Without the glue functions we cannot discover modules in a shared library.  You would need to know the module names and look up syntax online.
2. Building external interfaces. If GMT/MATLAB toolbox (or Julia I think) wanted to call some of these foreign modules and take advantage of matlab syntax, e.g.,

```
G = gmt ('surface', 'R0/30/0/30 -I0.1 .mbdata.txt');  %grid a large MB dataset
plot = gmt ('mbgrdtiff', 'some options', G);  % Make tiff file via mbsystem
```

then we need to be able to access the KEYS for those modules in the library.  Otherwise we cannot do the above since we would not know how to hook the G matrix into the mbgrdtiff command line options.

So that is why I want to encourage having the glue **and** making it super-painless for those developers to do this.

PaulWessel:
Anything else for this branch you are thinking of, @seisman ?

PR #3134, opened 2020-04-19
Auto-generate gmt_moduleinfo.h

seisman:
The auto-generated gmt_moduleinfo.h is located in the cmake build directory.

Known issues of the auto-generated file:

1. the entries are unsorted
2. some entries are missing as shown below.

To compare the auto-generated file with the src/gmt_moduleinfo.h file, you need to build this branch first, then run:

```
$ sort build/src/gmt_moduleinfo.h > jk
$ diff jk src/gmt_moduleinfo.h
1a2
> 	{"begin", "begin", "core", "Initiate a new GMT modern mode session", ""},
4a6
> 	{"clear", "clear", "core", "Delete current default settings, or the cache, data or sessions directories", ""},
10a13
> 	{"end", "end", "core", "Terminate GMT modern mode session and produce optional graphics", ""},
11a15
> 	{"figure", "figure", "core", "Set attributes for the current modern mode session figure", ""},
21a26
> 	{"gmtread", "gmtread", "core", "Read GMT objects into external API", "-T-,<?{,>?}"},
28a34
> 	{"gmtwrite", "gmtwrite", "core", "Write GMT objects from external API", "-T-,<?{,>?}"},
60a67
> 	{"inset", "inset", "core", "Manage figure inset setup and completion", ">X}"},
80a88
> 	{"subplot", "subplot", "core", "Manage modern mode figure subplot configuration and selection", ""},
```


seisman:
What to do with the missing entries?

PaulWessel:
Couple of issues we should address while doing this I think:

1. set (GMT_PROGS_SRCS  is the one that lists the current modules.  This is a duplication of gmt_modules.h or some other file with the official list.  So inset, begin, etc are simply not listed here.  Not sure what that does.  I see begin.c is listed under GMT_LIB_SRCS and not modules.  What is the actual effect of that?  Would anything actually change if we moved then to GMT_PROGS_SRCS?
2. set (GMT_LIB_SRCS is similar; easy to forget to add or subtract a file.  WOuld it not be simpler if we renamed the common_*.[ch] files to gmt_common_*.[ch] and simply used wildcard to collect gmt_*.[ch], postscriptlight.[ch] and the fixed items (kiss, triangle, etc).?

seisman:
> 1. set (GMT_PROGS_SRCS  is the one that lists the current modules.  This is a duplication of gmt_modules.h or some other file with the official list.  So inset, begin, etc are simply not listed here.  Not sure what that does.  I see begin.c is listed under GMT_LIB_SRCS and not modules.  What is the actual effect of that?  Would anything actually change if we moved then to GMT_PROGS_SRCS?


**GMT_PROGS_SRCS** is also used for creating the symlinks. That's why we don't list begin, inset etc in **GMT_PROGS_SRCS**.
```
string (REPLACE ".c" "" _gmt_modules "${GMT_PROGS_SRCS}")
install_module_symlink (${_gmt_modules} ${GMT_COMPAT_MODULES})
```

> 2. set (GMT_LIB_SRCS is similar; easy to forget to add or subtract a file.  WOuld it not be simpler if we renamed the common__.[ch] files to gmt_common__.[ch] and simply used wildcard to collect gmt_*.[ch], postscriptlight.[ch] and the fixed items (kiss, triangle, etc).?

Yes, it would be much shorter if the source codes all start with `gmt_`

PaulWessel:
I see, so we dont want to create links like begin since there never was a GMT4 begin call.  MIght it be simpler to have a GMT_CMD_SRCS with begin, end subplot etc and somehow strip those off from the full list, plus include in the LIB src?
I will work on renaming some files.

PaulWessel:
OK, I did the gmt_common thing and it is compiling.

PaulWessel:
Hi @seisman, what is our next step here?  I dont think sort-order is important.  Right now, --show-classic calls qsort anyway and we could do the same for modern.  There is no time wasted since this is just for these gmt --* calls.  Does your recipe above make a gmt_moduleinfo.h per src directory? I.e., we are going down the path of one shared lib per supplement dir, right?

seisman:
> Does your recipe above make a gmt_moduleinfo.h per src directory? I.e., we are going down the path of one shared lib per supplement dir, right?

No, currently it only makes the file for the top src directory.

seisman:
Now the newly generated file is the same as the original gmt_moduleinfo.h file. I think it's OK to merge. 

For src subdirectories, you probably need to revise gmt_supplements_module.* first.

PaulWessel:
Thanks, I will start a new branch related to supplements when I find the time,

PaulWessel:
Hi @joa-quim, please build for Win and try with Julia in case there aer any issues.  I suspect we will have more things that will be tricker once we split up supplements.dl into several *dll's

seisman:
Why do we need to split up the supplements.dll?

PaulWessel:
We don't unless we want to.  I was looking at how to let your moduleinfo creator be used by the supplements.  I know you have set things up so that each one can either be part of supplements.so or its own separate supplements.  For supplements.so we need to make a single moduleinfo file, which then would have to be in src like gmt_supplements_module.h currently is.  I was just wondering out loud if it is simpler to have a system where all the src/* dirs are decoupled and we make gmt_moduleinfo.h files for each one in build/src/*, and separate *.so, rather than coupling them all back into one include file and one shared lib.  I dont know the answer but seems like that is the first order question to discuss before moving on this.

PaulWessel:
Basically, this breaks any coupling between src and src/* I think.  Currently, the core knows about supplements.so in gmt_api.c lookups but not about others.  I think there may be benefits to decouple it entirely.

joa-quim:
We DONT't want to break the suppl in *.dlls We don't want the cross dll shit back again.

PaulWessel:
OK, so we need the building of gmt_moduleinfo.h content to either go into separate files or (as for supplements) go into a common gmt_suppl_moduleinfo.h file.  Is that possible, @seisman ? I dont think we can control the order of things so having several CMakeLists files appending to the same file is probably a no-go.

seisman:
Yes, it's possible to add all supplement modules (including custom ones) into one header file.

PaulWessel:
OK, so shall I start a branch where I remove the gmt_supplements_module.h file since it will be built and do the same I did for the core, and then you can add the cmake magic?

seisman:
OK for me.

joa-quim:
Wasn't this supposed to be tested by me before merge?

PaulWessel:
I don't think there are any actual changes in functionality, just simplifying how the list of module info (the structure) and the functions that calls them) are done.  Still one supplements.so. Is it actually breaking anything for you?

seisman:
This PR was merged into branch new-api-lookup. You can test that branch.

PR #3137, opened 2020-04-20
Fix new bug in filterwidth parsing for median filter

PaulWessel:
I recently introduced this bug, sorry. Now fixing it.


PaulWessel:
FYI, the bug is that if the median filter is selected then I forgot to add the above call to get teh width - hence it is always 0 and we get an error.

PR #3139, opened 2020-04-21
Ensure hoels of polygons have correct handedness

### LINKED ISSUE #3135: Polygon holes filling is tricky
It seems that adding **-Ph** to headers is not enough. It needs that the hole polygon has an orientation contrary to its parent. But that's not all, if origin is not at LowerLeft (for rectangles) wholes are painted too regardless if its orientation.

```
cat << EOF > quads.dat
> -Gred CCW
0  0
5  0
5  10
0  10
0  0
> -Ph  CW
2  1
2  3
4  3
4  1
2  1
> -Ph  CCW
2  4
4  4
4  6
2  6
2  4
> -Ph CCW origin TopLeft
2  9
2  7
4  7
4  9
2  9
> CW -Gblue
5  0
5  10
10 10
10 0
5  0
> -Ph  CW
7  1
7  3
9  3
9  1
7  1
> -Ph  CCW
7  4
9  4
9  6
7  6
7  4
EOF
gmt plot quads.dat -JX12c -Ba2f1WSen -R-0.5/10.5/-0.5/10.5 -Gred -W1p -png map
```
map

PaulWessel:
Do not trust the incoming data.  For polygons with holes that need to be filled, we ensure the holes have a different handedness than the parent, reversing the child if needed.  Closes #3135.

### DIFF
--- a/src/gmt_constants.h
+++ b/src/gmt_constants.h
@@ -257,6 +257,11 @@ enum GMT_enum_inside {
 	GMT_ONEDGE,
 	GMT_INSIDE};
 
+/*! Return codes from gmt_polygon_orientation */
+enum GMT_enum_polorient {
+	GMT_POL_IS_CCW = 0,
+	GMT_POL_IS_CW  = 1};
+
 /*! Codes for -q selections */
 enum GMT_enum_skiprows {
 	GMT_RANGE_ROW_IN   = 1,
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -8287,7 +8287,8 @@ void gmt_geo_polygons (struct GMT_CTRL *GMT, struct GMT_DATASEGMENT *S) {
 	struct GMT_DATASEGMENT *S2 = NULL;
 	struct GMT_DATASEGMENT_HIDDEN *SH = gmt_get_DS_hidden (S);
  	bool add_pole, separate;
-	int outline = 0;
+	int outline = 0, P_handedness = GMT_NOTSET, H_handedness;
+	int geo = gmt_M_is_geographic (GMT, GMT_IN);
 	uint64_t used = 0;
 	char *type[2] = {"Perimeter", "Polar cap perimeter"};
 	char *use[2] = {"fill only", "fill and outline"};
@@ -8323,7 +8324,17 @@ void gmt_geo_polygons (struct GMT_CTRL *GMT, struct GMT_DATASEGMENT *S) {
 	if (PSL->internal.comments) snprintf (comment, GMT_LEN64, "%s polygon for %s\n", type[add_pole], use[PSL->current.outline]);
 	used = gmtplot_geo_polygon_segment (GMT, S, add_pole, true, comment);	/* First lay down perimeter */
 	for (S2 = gmt_get_next_S (S); S2; S2 = gmt_get_next_S (S2)) {	/* Process all holes [none processed if there aren't any holes] */
+		if (P_handedness == GMT_NOTSET)
+			P_handedness = gmt_polygon_orientation (GMT, S->data[GMT_X], S->data[GMT_Y], S->n_rows, geo);
+		H_handedness = gmt_polygon_orientation (GMT, S2->data[GMT_X], S2->data[GMT_Y], S2->n_rows, geo);
+
 		if (PSL->internal.comments) snprintf (comment, GMT_LEN64, "Hole polygon for %s\n", use[PSL->current.outline]);
+		if (H_handedness == P_handedness) {
+			uint64_t row_f, row_l, col;
+			for (row_f = 0, row_l = S2->n_rows - 1; row_f < S2->n_rows/2; row_f++, row_l--) {
+				for (col = 0; col < S2->n_columns; col++) gmt_M_double_swap (S2->data[col][row_f], S2->data[col][row_l]);
+			}
+		}
 		used += gmtplot_geo_polygon_segment (GMT, S2, false, false, comment);	/* Add this hole to the path */
 	}
 	if (used) {
--- a/src/gmt_prototypes.h
+++ b/src/gmt_prototypes.h
@@ -351,6 +351,7 @@ EXTERN_MSC bool gmt_this_alloc_level (struct GMT_CTRL *GMT, unsigned int alloc_l
 
 /* gmt_support.c: */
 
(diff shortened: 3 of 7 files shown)


joa-quim:
OK, this works but there is some other bug from externals that probably is not looking for the **-Ph** option.
This paints the whole in blue
```
> -Gred
0  0
5  0
5  10
0  10
0  0
> -Ph -Gblue
2  1
2  3
4  3
4  1
2  1

D = gmtread("quad.dat");
plot(D, J="X12c", B="a2f1WSen", G=:red, W=1, show=1)
```
 but doe not male it transparent when I remove the **-Gblue** from second segment header.

PaulWessel:
Maybe make a mex version of the test?

joa-quim:
Use your Julia. Life is much easier with it
```
D=gmtmex('read quad.dat -Td');
gmtmex('psxy -JX12c -Ba2f1WSen -W1 -R-0.5/10.5/-0.5/10.5 > lixo.ps', D)
```

PaulWessel:
OK, fine.  I just installed/update julia to 1.3.1.  Followed your macOS hint of addint startup.jl.  Then tried ]add GMT and ERROR unexpected ].

PaulWessel:
Am I completely missing any instructions for actuallly installing GMT.jl?.

joa-quim:
Last version is Julia 1.4 but 1.3 is just as good.
You only have to do
``] add GMT``
When you type the ``]`` the prompt immediately turns into ``pkg``. What is your ` `ERROR unexpected``?

PaulWessel:
Seems like your instructions is missing some critical spaces between ] and add.  Works now.

joa-quim:
Ah, because you copy-pasted?

PaulWessel:
Yes, but also the README.md has no space.

PaulWessel:
Now here:
ia> D = gmtread("quad.dat");
ERROR: UndefVarError: gmtread not defined

PaulWessel:
FYI, the adding package seems to work?

```
v1.3) pkg>  add GMT
   Cloning default registries into `~/.julia`
   Cloning registry from "https://github.com/JuliaRegistries/General.git"
     Added registry `General` to `~/.julia/registries/General`
 Resolving package versions...
 Installed GMT ─ v0.17.0
  Updating `~/.julia/environments/v1.3/Project.toml`
  [5752ebe1] + GMT v0.17.0
  Updating `~/.julia/environments/v1.3/Manifest.toml`
  [5752ebe1] + GMT v0.17.0
  [de0858da] + Printf 
  [4ec0a83e] + Unicode 
```

joa-quim:
Ok, you should not need it but better to have the master version
``] add GMT#master``
And more thing, you have to do 
``D = gmtread("/path/to/file"); or do a ``cd("/path/to/file)`` first.
But you say ``gmtread`` is not found? Can't be. Try running the tests
``] test GMT``

joa-quim:
Ah sorry, I know. It's like Python. Before using any package you have to import it. So first command per session has to be
``using GMT``

PaulWessel:
D = gmtread("quad.dat");
ERROR: error compiling gmtread: error compiling #gmtread#60: error compiling gmt: could not load library "libgmt"

I put the startup.jp in config with path to my library dir and restarted.

julia
ERROR: LoadError: UndefVarError: Libdl not defined

joa-quim:
Ok the other user that passed through this had to do
``using Libdl`` before the ``push!(Libdl.DL_LOAD_PATH ...``
I have to update this instructions also

PaulWessel:
I added a call to Libdl in startup.jo before the LIB call.  Same error.
Admittedly I am a Julia rookie so this is a very good test - clearly your  README needs some work.  I would think the average rookie would give up by now.

PaulWessel:
OK, using Libdl works.
Plot generated.

joa-quim:
Good. I debugged to confirm that the header is well transmitted to GMT. And mex version confirms the same thing. No hole.

PaulWessel:
So when I debug MEX I have to load in gmtmex.c (compiled for Xcode) into Xcode, set a stop point, start matlab, connect Xcode to MATLAB processes, then run the matlab command.
What do I load into Xcode to set a stop point for Julia???

PaulWessel:
You may need to add a "how to debug" section.

PaulWessel:
Anyway, in the GMT io we do this:

`if (gmt_parse_segment_item (GMT, S->header, "-Ph", NULL)) SH->pol_mode = GMT_IS_HOLE;
`

so the hidden helper structure per segment sets a flag.  When gmt read returns to mablat it has the info:

```
ans = 

  struct with fields:

       data: [5x2 double]
       text: []
     header: '-Gred'
    comment: []
      proj4: []
        wkt: []

>> D(2)

ans = 

  struct with fields:

       data: [5x2 double]
       text: []
     header: '-Ph'
    comment: []
      proj4: []
        wkt: []

```

What does your Julia D data have?

joa-quim:
Ok, dinner time here (still preparing it).
To debug from Julia you should do exactly the same but load the Julia process in Xcode. There is no equivalent in C to gmtmex so you can't debug that part with Xcode (you need VScode).

In Julia you get the same thing as in Matlab . Have a function for display that only shows first segment. To see more you do

```
julia> D[1]
Header: -Gred
5×2 Array{Float64,2}:
 0.0   0.0
 5.0   0.0
 5.0  10.0
 0.0  10.0
 0.0   0.0


julia> D[2]
Header: -Ph -Gblue
5×2 Array{Float64,2}:
 2.0  1.0
 2.0  3.0
 4.0  3.0
 4.0  1.0
 2.0  1.0
```

PaulWessel:
As far as I can tell, when D is passed back into a matlab gmt call the header is extracted and copied into the GMT segment header so it should work - I just need to rebuild gmtmex here.

joa-quim:
I verified that it is indeed like that from Julia, but -Ph makes no effect (tried with both **-Ph -Gblue** and **-Ph**

joa-quim:
Don't want to try a debug from Julia
1. Load the Julia process in XCode
2. Open a GMT source code in XCode and set a breakpoint
3. run the julia command that triggers that breakpoint

Ofc never tried in Mac but that's what I do on Windows.

PaulWessel:
OK, it makes sense to set stop point in the GMT C in Xcode.  I will see - as you know multi-tasking on this end during the day...

joa-quim:
BTW, I updated the installing instructions so you can give a star to the project.

PR #3143, opened 2020-04-22
Fix wrong string comparison when converting berween GMT and GDAL ellipsoid names

joa-quim:


joa-quim:
This was causing 
> rdproject [WARNING]: Unknown conversion between GMT and GDAL ellipsoid names. Using a generic spherical body.

PR #3159, opened 2020-04-24
Eliminate glue include files

PaulWessel:
Since the glue functions are **not** called as is anywhere, their names are only assembled via strings and obtained via a pointer from _dlsym_, there is no need forgmt_glue.h and gmt_suppl_glue.h.  Like for the modules, I just added EXTERN_MSC to the functions in gmt_*glue.c instead.


PaulWessel:
Gotta avoid having both gmt_glue.c.in and gmt_suppl_glue.c.in.  If we only have the former, can I do this:

```
configure_file (gmt_glue.c.in gmt_${SUPPL_NAME}_glue.c)
set (SUPPL_NAME ${GMT_SUPPL_LIB_NAME})
configure_file (gmt_glue.c.in gmt_${SUPPL_NAME}_glue.c)
```

If so then we need to rename gmt_moduleinfo.h to gmt_core_moduleinfo.h and the other one to gmt_supplements_moduleinfo.h.  I can then delete gmt_suppl_glue.c.in as well. Seems about right?

PaulWessel:
OK, now we are talking.  Only gmt_glue.c.in remains in the repo.  From it we build both gmt_core_glue.c and gmt_supplements_glue.c, and these include your gmt_*_moduleinfo.h files.  Got rid of a few hundred lines of cryptic C code!

PR #3161, opened 2020-04-24
Don't complain about -J if -I is given in histogram

PaulWessel:
No projection is needed then.


PaulWessel:
This should fix you case, @joa-quim.

PR #3162, opened 2020-04-24
Clean up some GMT src shell scripts

PaulWessel:
I have moved _gmt_make_pattern_include.sh_ to the admin directory since it is not meant to ever be run again - it just is kept for reference.  The _gmt_clip_test.sh_ is very old and deprecated, so I deleted it.  The _gmt_make_enum_dicts.sh_ gave the wrong name for the API function to get an enum, so I fixed that part.


PaulWessel:
Where is that file?

seisman:
In the root directory https://github.com/GenericMappingTools/gmt/blob/master/.azure-pipelines.yml

PaulWessel:
Sorry, got it.  I was sniffing in the ci dir.

seisman:
@PaulWessel Since you changed `gmt_make_enum_dicts.sh`, you also need to run `gmt_make_enum_dicts.sh` manually to update the `gmt_enum_dict.h`.

PaulWessel:
Please merge when checks finishes - I will be out for a few hours.

PR #3164, opened 2020-04-25
Rename vars

PaulWessel:
Renamed SUPPL_* to SHARED_LIB_* in src/CMakeLists.txt and gmt_glue.c.in so the variable names are not confused with others that are specific to a supplement.


src/CMakeLists.txt
```diff
@@ -827,7 +827,7 @@ if (BUILD_SUPPLEMENTS)
 	# install more
 	foreach (_dir ${GMT_SUPPL_DIRS})
 		# get variables from subdirectory
-		get_subdir_var (_suppl_name SUPPL_NAME ${_dir})
+		get_subdir_var (_suppl_name SHARED_LIB_NAME ${_dir})
```
seisman, line 830: Here we need to get the value of SUPPL_NAME from src subdirectories.
```suggestion
		get_subdir_var (_suppl_name SUPPL_NAME ${_dir})
```

src/CMakeLists.txt
```diff
@@ -19,8 +19,8 @@
 #
 # src/CMakeLists.txt
 #
-set (SUPPL_NAME core)
-set (SUPPL_PURPOSE "GMT core: The main modules of the Generic Mapping Tools")
+set (SHARED_LIB_NAME core)
+set (SHARED_LIB_PURPOSE "GMT core: The main modules of the Generic Mapping Tools")
```
seisman, line 23: Could you move these two lines to Line 340?

PaulWessel: OK

PR #3170, opened 2020-04-25
Sort output of gmt --help

seisman:
This PR sorts the output of gmt --help, but it only works for core modules.

The order of supplement modules are controlled by the variable **SUPPL_PROGS_SRCS**
in each src subdirectory.


seisman:
After PR #3174, this PR no longer works.

PR #3176, opened 2020-04-26
cmake: get_subdir_var_files: Do not prepend dirname for absolute paths

seisman:
Check if a file is an absolute path before prepending the dirname.

Also update the indentations of the GET_SUBDIR_LIST macro.


seisman:
@PaulWessel Please try if this PR can solve the problem in #3130.

PaulWessel:
OK, do I change

	${CMAKE_CURRENT_BINARY_DIR}/gsfml_config.h)

and do I set

set (SUPPL_EXTRA_DIRS newsuppl)
or
set (SUPPL_EXTRA_DIRS full-path-to-gsfml-src)

seisman:
> OK, do I change
> 
> ```
> ${CMAKE_CURRENT_BINARY_DIR}/gsfml_config.h)
> ```

YES.

> and do I set
> 
> set (SUPPL_EXTRA_DIRS newsuppl)
> or
> set (SUPPL_EXTRA_DIRS full-path-to-gsfml-src)

I think both should work.

PaulWessel:
With newsuppl I get this: 

> file failed to open for reading (No such file or directory):
> 
>     /Users/pwessel/GMTdev/gmt-dev/src//Users/pwessel/GMTdev/gmt-dev/dbuild/src/newsuppl/gsfml_config.h

With full path I get this:

> CMake Error at src/CMakeLists.txt:698 (add_subdirectory):
>   add_subdirectory not given a binary directory but the given source
>   directory "/Users/pwessel/GMTdev/gsfml/TOOLS/src" is not a subdirectory of
>   "/Users/pwessel/GMTdev/gmt-dev/src".  When specifying an out-of-tree source
>   a binary directory must be explicitly specified.

So the latter seems allowable but requires a binary directory specified.  Do they mean the BUILD dir?

seisman:
I just realize that your SUPPL_PROGS_SRCS may be incorrect. You should only list the modules in SUPPL_PROGS_SRCS. For any header files, you should list them either in SUPPL_HEADERS or SUPPL_LIB_SRCS.

PaulWessel:
Yes, moving that correctly works (this is with newsuppl link).  

```
...
x2sys_merge      Merge an updated COEs table (smaller) into the main table (bigger)

Module name:     Purpose of gsfml module:
----------------------------------------------------------------
fzanalyzer       Analysis of fracture zones using crossing profiles
fzblender        Produce a smooth blended FZ trace
mlconverter      Convert chrons to ages using selected magnetic timescale
```

So this works for the relative path.

seisman:
> With full path I get this:
> 
> CMake Error at src/CMakeLists.txt:698 (add_subdirectory):
> add_subdirectory not given a binary directory but the given source
> directory "/Users/pwessel/GMTdev/gsfml/TOOLS/src" is not a subdirectory of
> "/Users/pwessel/GMTdev/gmt-dev/src". When specifying an out-of-tree source
> a binary directory must be explicitly specified.
> 
> So the latter seems allowable but requires a binary directory specified. Do they mean the BUILD dir?
> 

The full syntax of add_subdirectory is 
```
add_subdirectory(source_dir [binary_dir] [EXCLUDE_FROM_ALL])
```
So we can specify the binary directory explicitly. For the full path case, I think the binary directory is required.

PaulWessel:
So while gsfml.so is built, it is the supplements.so that has the functions and list of moduleinfo for that library.  It works, but ideally the glue should be in gsfml.so.  That means the moduleinfo file must skip adding the modules if SUPPL_NAME is set

PaulWessel:
> So we can specify the binary directory explicitly. For the full path case, I think the binary directory is required.

Yes I guess since it is all the same binary dir that might work.

PaulWessel:
Sorry, in the supplement dir CMakefile I now have this

```
set (SUPPL_NAME gsfml)
set (SHARED_LIB_NAME gsfml)
```

Delete the SUPPL_NAME, right? I added

`configure_file (../gmt_glue.c.in gmt_gsfml_glue.c)`

and it builds gmt_gsfhl_glue.c in the build/src/newsuppl dir.  However, there will have to be some changes to your macro to also separate out the gsfml modules into a gmt_gsfml_moduleinfo.h file in that directory.

seisman:
Ideally, it should be SUPPL_LIB_NAME, not SHARED_LIB_NAME.
```
set (SUPPL_NAME gsfml)
set (SUPPL_LIB_NAME gsfml)
```
and `configure_file (../gmt_glue.c.in gmt_gsfml_glue.c)` should be invoked in the src/CMakeLists.txt. I'll see how it works.

PaulWessel:
Couple of issues with that:

1. gmt_glue.c.in contains lines like this: `void gmt_@SHARED_LIB_NAME@_module_show_all` so we need to use the same variable name
2. src/CMakeLists.txt should not need to know anything about gsfml - it has to happen in its CMakeLists.txt file.

setting `set (SHARED_LIB_NAME gsfml) `works, otherwise it is not set and I get _supplements_ in that file.

PaulWessel:
We use SHARED_LIB_NAME to set it to supplements in the main CMakeLists.txt file so it seems reasonable to set it to something else if we want to build a separate shared library, no?

PaulWessel:
In case you want to play with it, here is the gsfml src directory (here called gsfml).
[gsfml.zip](https://github.com/GenericMappingTools/gmt/files/4534601/gsfml.zip)

seisman:
Your concerns above are addressed in #3177.

PR #3177, opened 2020-04-26
cmake: Build seperate gmt_glue.c and gmt_xxx_moduleinfo.h for non-official packages

seisman:
In the old implementation, the module information of both official and non-official supplement packages are built into gmt_supplements_moduleinfo.h, but the functions are built into seperate libraries (e.g. supplements.so and gsfml.so). It works but not ideal.

For non-official packages, they should have their own gmt_gsfml_glue.c and gmt_gsfml_moduleinfo.h. This PR implement such a feature.

To try this PR, you can add  `set (SUPPL_LIB_NAME seis)` to `src/seis/CMakeLists.txt` and build GMT, then you will see following files in the build directory
```
gmt_core_glue.c        gmt_seis_glue.c        gmt_supplements_glue.c
gmt_core_moduleinfo.h        gmt_seis_moduleinfo.h        gmt_supplements_moduleinfo.h
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -714,6 +714,9 @@ if (BUILD_SUPPLEMENTS)
 		get_subdir_var (_suppl_extra_libraries SUPPL_EXTRA_LIBS ${_dir})
 		list (APPEND SUPPL_${_suppl_lib_name}_EXTRA_LIBRARIES ${_suppl_extra_libraries})
 
+		# library purpose of non-official supplement packages
+		get_subdir_var (SUPPL_${_suppl_lib_name}_LIB_PURPOSE SUPPL_LIB_PURPOSE ${_dir})
+
 		# rename the supplement dll [Windows only]
 		get_subdir_var (SUPPL_${_suppl_lib_name}_DLL_RENAME SUPPL_DLL_RENAME ${_dir})
 	endforeach (_dir)
@@ -727,7 +730,11 @@ if (BUILD_SUPPLEMENTS)
 		if (_suppl_lib_name EQUAL GMT_SUPPL_LIBRARIES)
 			set (SHARED_LIB_PURPOSE "GMT ${GMT_SUPPL_LIBRARIES}: The official supplements to the Generic Mapping Tools")
 		else (_suppl_lib_name EQUAL GMT_SUPPL_LIBRARIES)
-			set (SHARED_LIB_PURPOSE "GMT ${_suppl_lib_name}: The non-official supplements to the Generic Mapping Tools")
+			if (SUPPL_${_suppl_lib_name}_LIB_PURPOSE)
+				set (SHARED_LIB_PURPOSE "${SUPPL_${_suppl_lib_name}_LIB_PURPOSE}")
+			else (SUPPL_${_suppl_lib_name}_LIB_PURPOSE)
+				set (SHARED_LIB_PURPOSE	"GMT ${_suppl_lib_name}: The non-official supplements to the Generic Mapping Tools")
+			endif (SUPPL_${_suppl_lib_name}_LIB_PURPOSE)
 		endif (_suppl_lib_name EQUAL GMT_SUPPL_LIBRARIES)
 
 		# generate gmt_${SHARED_LIB_NAME}_glue.c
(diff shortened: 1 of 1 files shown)


seisman:
@PaulWessel Please try as suggest above. 

If you want to try your gsfml package, you need 
```
set (SUPPL_NAME gsfml)
set (SUPPL_LIB_NAME gsfml)
```
and **remove**
```
configure_file (../gmt_glue.c.in gmt_gsfml_glue.c)
```
from src/gsfml/CMakeLists.txt.

PaulWessel:
OK, did that but got error below.  I have this in the file as well:

`set (SUPPL_LIB_SRCS ${SUPPL_PROGS_SRCS} ${CMAKE_CURRENT_BINARY_DIR}/gmt_gsfml_glue.c)`

but got 

```
*  GMT_MANDIR                 : /Users/pwessel/GMTdev/gmt-dev/dbuild/gmt6/share/man
-- Configuring done
CMake Error at src/CMakeLists.txt:756 (add_library):
  Cannot find source file:

    /Users/pwessel/GMTdev/gmt-dev/dbuild/src/newsuppl/gmt_gsfml_glue.c
```

I notice there is no gmt_supplements_moduleinfo.h in build/src.  But gmt_gsfml_glue.c is there, but no moduleinfo files anywhere.

seisman:
You don't need to add `${CMAKE_CURRENT_BINARY_DIR}/gmt_gsfml_glue.c` to SUPPL_LIB_SRCS.

seisman:
You can this gsfml [CMakeLists.txt](https://github.com/GenericMappingTools/gmt/files/4534618/CMakeLists.txt). It works well for me.

seisman:
I also updated **SHARED_LIB_PURPOSE**. Now it's

- GMT core: The main modules of the Generic Mapping Tools
- GMT supplements: The official supplements to the Generic Mapping Tools
- GMT gsfml: The non-official supplements to the Generic Mapping Tools

PaulWessel:
yes, now it works well, thanks.  But it is ignoring my

set (SHARED_LIB_PURPOSE "GMT gsfml: Tools for the GSFML project")

which I rather have than

GMT gsfml: The non-official supplements to the Generic Mapping Tools

PaulWessel:
Looking at the gsfml.so file with nm, I see this (I am not sure what all the output of nm -g means though):

```
nm -g gsfml.so | egrep '^0'
0000000000005810 T _GMT_fzanalyzer
000000000000c170 T _GMT_fzblender
0000000000010050 T _GMT_mlconverter
0000000000003c00 T _gmt_current_name
00000000000049b0 T _gmt_get_full_name
0000000000004d10 T _gmtlib_get_active_name
0000000000005780 T _gmtlib_gsfml_module_classic_all
00000000000057e0 T _gmtlib_gsfml_module_group
00000000000057b0 T _gmtlib_gsfml_module_keys
0000000000005750 T _gmtlib_gsfml_module_list_all
0000000000005720 T _gmtlib_gsfml_module_show_all
0000000000005220 T _gmtlib_is_modern_name
0000000000005660 T _gmtlib_set_KOP_strings
```

Looks to me that a few symbols are exported by this shared lib that is also in other shared libs (gmt_current_name, gmt_get_full_name, gmtlib_get_active_name, gmtlib_is_modern_name, gmtlib_is_modern_name, gmtlib_set_KOP_strings).  That is probably a (harmless) bug we should try to fix.  Also, maybe we should reconsider the naming convention for the glue files.  Looks odd that there are gmtlib_* files in gsfml.so.

PaulWessel:
Those I listed above are also in libgmt.dylib.  I think it is that gmt_modern.h inclusion I mentioned yesterday.  Will have to play with this a bit.

seisman:
> yes, now it works well, thanks. But it is ignoring my
> 
> set (SHARED_LIB_PURPOSE "GMT gsfml: Tools for the GSFML project")
> 
> which I rather have than
> 
> GMT gsfml: The non-official supplements to the Generic Mapping Tools

The issue is fixed in my latest commit. I use **SUPPL_LIB_NAME** instead of **SHARED_LIB_NAME**. i.e. you need the following line in src/newsuppl/CMakeLists.txt:
```
set (SUPPL_LIB_PURPOSE "GMT gsfml: Tools for the GSFML project")
```

If the variable is not defined, then the library purpose will be set to the default, i.e. 
```
GMT gsfml: The non-official supplements to the Generic Mapping Tools
```

PaulWessel:
Yep, working.  Great.  So I think this PR is good to go.

seisman:
> Looks to me that a few symbols are exported by this shared lib that is also in other shared libs (gmt_current_name, gmt_get_full_name, gmtlib_get_active_name, gmtlib_is_modern_name, gmtlib_is_modern_name, gmtlib_set_KOP_strings). That is probably a (harmless) bug we should try to fix.

It's because we include gmt_modern.c for all libraries. I'm not sure if it's necessary.

PR #3178, opened 2020-04-26
Rename glue

PaulWessel:

1. Strip from the glue functions names the leading **gmtlib_** prefix.
2. Remove _gmt_modern.c_ from being linked into supplemental shared libs


seisman:
Functions in gmt_mbsystem_glue.c also need renaming, right?

seisman:
Ping @PaulWessel. 

Functions in gmt_mbsystem_glue.c also need renaming, right?

PaulWessel:
Yes, see #3182 .

Issue #3180, opened 2020-04-26
Still custom supplements type 3


joa-quim:
I still not getting how build de MBGMT modules. I have this in a ``mggmt_module.c``
```
#include "gmt_dev.h"
#include "gmt_internals.h"
#include "mbgmt_module.h"

/* Sorted array with information for all mbgmt modules */

static struct GMT_MODULEINFO modules[] = {
	{"mbareaclean", "mbareaclean", "mbgmt", "Automatically flag bad beams in swath sonar bathymetry data", ""},
	{"mbcontour",   "mbcontour", "mbgmt",  "Plot swath bathymetry, amplitude, or backscatter",  "CC(,>X}"},
	{"mbclean_j",   "mbclean_j","mbgmt", "Identifies and flags artifacts in swath sonar bathymetry data", ""},
	{"mbflags",     "mbflags","mbgmt", "Update the flags .esf file from external file", ""},
	{"mbgetdata",   "mbgetdata","mbgmt", "Get swath bathymetry, amplitude, or backscatter data into Matlab", "<D{,ND),MD}"},
	{"mbgrdtiff",   "mbgrdtiff", "mbgmt",  "Project grids or images and plot them on maps",  "<G{+,CC(,IG("},
	{"mbgrid_j",    "mbgrid_j","mbgmt", "grid bathymetry, amplitude, or sidescan data of a set of swath sonar data files", ""},
	{"mbimport",    "mbimport","mbgmt", "Get swath bathymetry, amplitude, or backscatter Image into Matlab", "<D{,CC(,MI}"},
	{"mbinfo",      "mbinfo","mbgmt", "Read a swath sonar data file and outputs some basic statistics.", "<D{,>D},>DC,>DO"},
	{"mbset",       "mbset","mbgmt", "Set values in an mbprocess parameter file", ""},
	{"mbprocess",   "mbprocess","mbgmt", "Tool for processing swath sonar bathymetry data", "ID{,OD("},
	{"mbswath",     "mbswath", "mbgmt",  "Plot swath bathymetry, amplitude, or backscatter",  "CC(,NC(,>X}"},
	{"gmtmbgrid",   "gmtmbgrid","mbgmt", "Grid table data using adjustable tension continuous curvature splines", "<D{,DD(,LG(,GG}"},
	{NULL, NULL, NULL, NULL, NULL} /* last element == NULL detects end of array */
};
```
[...]

PR #3181, opened 2020-04-26
New glue builder

PaulWessel:

See #3180 and #3130 for content.  This PR addresses type 3 supplements that are built long after GMT has been built and installed.  To enable these supplements to have their modules listed in the `gmt --help` listing these maintainers only need to

1. Run `gmt --new-glue=name > gmt_name_suppl.c` in their source directory, with name the desired name of their shared library (e.g., mbsystem).
2. Include gmt_name_suppl.c when building the shared library.

Note: Unlike the automatic building of these glues via CMake configuration, the above command only creates a single source code where the *moduleinfo* part is inlined directly and not a separately configured file.  The modules are automatically sorted alphabetically.


PaulWessel:
PS. The only thing I don't like about this implementation is the fact that I am hardwiring in lots of C code based on gmt_glue.c.in.  We will have to remember to update two places.  I could not think of a simple way to read the gmt_glue.c.in at run time.  I could write a build_glue.sh script that creates an include snippet for gmt_support.c like we did in the old gmt_module_*.c|h days.  Comments welcome.

seisman:
A quick question: Can type 3 supplements access low-level APIs? Is this repo https://github.com/GenericMappingTools/custom-supplements type 3?

PaulWessel:
I will need to have a look at that given the recent cnanges.  This may be a type 4 supplement which is only relying on gmt.h.  if so they cannot make use of any glue (since depends on gmt_internals.h).

PaulWessel:
Finally, this call does not look for or handle a **SHARED_LIB_PURPOSE** setting.  I could look for an optional

`#define THIS_MODULE_LIB_PURPOSE "MB-System: GMT compatible modules"`

string in the modules and pick the first I find, else it just gets the name of the library.
Any better ideas?

seisman:
> This PR addresses type 3 supplements that are built long after GMT has been built and installed.

I'm confused why type 3 supplements like mbsystem can access low-level APIs, since GMT is already built and installed.

PaulWessel:
The mbsystem is a type 2 supplement because in has

`#include "gmt_dev.h"`

and that means it has access to the gmt_* and gmtlib_* functions of the GMT library.  Those supplements (I guess type 4) that only include gmt.h are limited to the GMT_* functions.

For many packages, there is a separate "-dev" install, e.g., gmt-dev, that would install the gmt_dev.h and other include files, while the main GMT package would only install gmt.h (and a few more called by gmt.h).

So may need to review the type numbering

1. Official GMT supplment in src/*
2. Wannabee supplement elsewhere but bilt with GMT via newsuppl symbolic link to external src
3. Outside supplement (e.g. mbsystem) that includes gmt_dev.h and thus can use glue to connet to gmt
4. External supplements only relying on the API (gmt.h).  No glue possible.

PaulWessel:
I think my suggestion above is not so bad for external type 3 supplements.  In at least one of the module C codes, include the line (e.g.)

`#define THIS_MODULE_LIB_PURPOSE "MB-System: GMT compatible modules"
`
to set the shared library purpose string.  Any better ideas?

PaulWessel:
And if no such thing is found I can print a warning to suggest they edit that line in the glue file.

seisman:
> I think my suggestion above is not so bad for external type 3 supplements. In at least one of the module C codes, include the line (e.g.)
> 
> `#define THIS_MODULE_LIB_PURPOSE "MB-System: GMT compatible modules" `
> to set the shared library purpose string. Any better ideas?

I don't have a better solution for this.

doc/rst/source/gmt.rst
```diff
@@ -79,6 +79,13 @@ If no module is given then several other options are available:
     scripting language among *bash*, *csh*, or *batch*.  Default is the main shell
     closest to your current shell (e.g., bash for zsh, csh for tcsh).
 
+**--new-glue**\ =\ *name*
+    Write the C code glue needed when building third-party supplements as shared
+    libraries.  The *name* is the name of the shared library. Run **gmt** in the directory
+    of the supplement and the glue code will be written to *stdout*.  Including this C code
+    when building the shared library means **gmt** can list available modules via the
+    **-show-modules**, **--help** options.  We recommend saving the code to gmt_*name*_glue.c.
```
seisman, line 87:
```suggestion
    **--show-modules**, **--help** options.  We recommend saving the code to gmt\_\ *name*\_glue.c.
```

src/gmt_support.c
```diff
@@ -16659,3 +16659,141 @@ char * gmt_add_options (struct GMT_CTRL *GMT, const char *list) {
...
+		error = GMT_RUNTIME_ERROR;
+		goto CROAK;
+	}
+
+	n++;
+	GMT_Report (API, GMT_MSG_INFORMATION, "%d module files found in current directory\n", n);
+
+	if (first_purpose) {
+		GMT_Report (API, GMT_MSG_NOTICE, "No #define THIS_MODULE_LIB_PURPOSE setting found in any module.  Please edit argument to gmtlib_module_show_all\n");
+		lib_purpose = library;
```
seisman, line 16740: Change the default library purpose to?
```
GMT xxx: The third-party supplements to the Generic Mapping Tools
```

PaulWessel: Done

PaulWessel:
testgmtio.c had a THIS_MODULE_OPTIONS that triggered my test...

PR #3185, opened 2020-04-27
Let GMT glue become accessible via API

PaulWessel:
See #3130 for context. Here, we have generalized the 6 gmtlib_show_* glue functions to two new GMT_* functions. That way, only gmt.h is needed for type 4 supplements to still supply module information to GMT.  The two functions are not documented in the API docs since they are only called from the auto-generated glue functions (either via CMake or `gmt --new-glue`).


seisman:
@PaulWessel Please run `src/gmt_make_enum_dicts.sh` and commit the changes.

PaulWessel:
Done. We need a better solution for this.  At least at release time we need a check.

seisman:
We already have a CI job checking this.

PaulWessel:
Great, forgot or did not know...

PR #3192, opened 2020-04-28
Handle font values with equal signs in them

### LINKED ISSUE #3190: Fonts with outline pen not handled by gmt set
The syntax for a font with an outline pen looks like this:

`32p,Helvetica-Bold,white=~1p,red`

However, when used in gmt set the embedded equal sign is seen as a PAR=value assignment and hell breaks loose.  Presumably the solution is to not do that when dealing with FONT settings.

PaulWessel:
From the olden days, we have been allowing the deprecated syntax **PAR** = _value_ and even **PAR**=_value_ given to gmt set, despite the syntax being **PAR** _value_.  This caused trouble for fonts with outline pens which contains an equal sign.  While being bakwards compatible, I removed the [=] from the docs to discourage the use of that old syntax.  We still allow it, but not **PAR**=_value_  if **PAR** contains the text **FONT_**.  Closes #3190.

### DIFF
--- a/doc/rst/source/gmtset.rst
+++ b/doc/rst/source/gmtset.rst
@@ -14,7 +14,7 @@ Synopsis
 
 **gmt set** [ |-C| \| |-D|\ [**s**\|\ **u**] \|
 |-G|\ *defaultsfile* ] [ **-**\ [**BJRXYp**]\ *value* ]
-PARAMETER1 [=] *value1* PARAMETER2 [=] *value2* PARAMETER3 [=] *value3*
+PARAMETER1 *value1* PARAMETER2 *value2* PARAMETER3  *value3*
 ...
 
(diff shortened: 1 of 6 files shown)


seisman:
Both commands don't complain:
```
gmt set FONT_TITLE=32p,Helvetica-Bold,white=~1p,red
gmt set FONT_TITLE 32p,Helvetica-Bold,white=~1p,red
```
But the gmt.conf file has (note the missing `~`):
```
FONT_TITLE                     = 32p,Helvetica-Bold,white=thick,red
```

PaulWessel:
Turns out we never updated gmt_putfont to deal with the special ~ flag.  So although getfont set the flag = 8, in putfont there was no knowledge of any 8. Now fixed.

Issue #3184, opened 2020-04-27
Bash script for building the GMT master branch


seisman:
**Description of the problem**

GMT already has many interfaces, GMT/Matlab, PyGMT, GMT.jl and 3rd-party supplements, e.g. gsfml, mbgmt and mbsystem. Ideally, these interfaces and supplements should run a nightly/weekly/monthly CI job against the GMT master branch, to make sure changes in the upcomming GMT releases don't break them. The CI job needs a bash script to download and build GMT. To avoid maintaining the same script in different repositories, it's better that the GMT repository can provide a bash script. 

The bash script contains the following steps:

- download the GMT master branch
- download GSHHG and DCW
- configure GMT
- build GMT
- install GMT

Assuming we have the script `ci/build-gmt-master.sh`, then other repository can easily install GMT by a command like
```
curl -sSL https://raw.githubusercontent.com/GenericMappingTools/gmt/master/ci/build-gmt-maste.sh | bash
```

The script should accept some environmental variables so that other users can control the installation locations and other details.

seisman:
Any thoughts? @PaulWessel @joa-quim @leouieda @weiji14

PaulWessel:
It is a good idea.  At the moment, there is no such checking for GMT/MEX (but I think JL and I are on top of that), GMT.jl (JL discovers trouble manually and then yells at me), MB-system (I am guessing they don't do CI), gsfml (me again, no CI for that small thing).  So probably only pyGMT is ready to take advantage of such a script in a CI.  It would be helpful if the MB folks could try someting like this since they use many low-level gmtdev library calls that are undocumented and not guaranteed to stay the same, and hence their source code has lots of #ifdefs for different GMT versions when these things have changed.  At least with the proposed script they would know right away if we broke something for them,

weiji14:
Yes that would be great. Ties in with what you mentioned at https://github.com/GenericMappingTools/pygmt/issues/345. My only two cents would be to 1) Extend it to Windows (make a .bat file?), and 2) have a way to track the version installed from master (would `gmt --version` give the git hash?) so that we can trace back to when things broke.

joa-quim:
GMT.jl already uses GMT master to run its tests.

seisman:
> 1) Extend it to Windows (make a .bat file?), and 

Having a bat file for Windows is a little challenging, because we need to build all GMT dependent libraries before building GMT on Windows.

[...]

PR #3193, opened 2020-04-28
CI: Add a script for building the latest GMT source codes

seisman:
Other projects can use the following command to install GMT:
```
curl https://raw.githubusercontent.com/GenericMappingTools/gmt/master/ci/build-gmt-master.sh | bash
```

See https://github.com/GenericMappingTools/custom-supplements/pull/3 for an example TravisCI configuration file.

Address #3184.

### DIFF
--- a/ci/build-gmt-master.sh
+++ b/ci/build-gmt-master.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+#
+# Build the lastest GMT source codes
+#
+
+# Following variables can be modified via environment variables
+GMT_INSTALL_DIR=${GMT_INSTALL_DIR:-${HOME}/gmt-install-dir}
+
+# General settings
+GSHHG_VERSION="2.3.7"
+DCW_VERSION="1.1.4"
+GSHHG="gshhg-gmt-${GSHHG_VERSION}"
+DCW="dcw-gmt-${DCW_VERSION}"
+EXT="tar.gz"
+
+cwd=${PWD}
+# 1. Create temporary directory for building
+rm -rf /tmp/build-gmt
+mkdir -p /tmp/build-gmt
+cd /tmp/build-gmt
+
+# 2. Download GMT, GSHHG and DCW from GitHub
+git clone --depth 10 https://github.com/GenericMappingTools/gmt.git gmt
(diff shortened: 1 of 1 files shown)


ci/build-gmt-master.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+#
+# Build the lastest GMT source codes
+#
```
leouieda, line 4: Would be good to list all the environment variables that can be set here so that it's easy to find.

ci/build-gmt-master.sh
```diff
@@ -0,0 +1,54 @@
...
+DCW_VERSION="1.1.4"
+GSHHG="gshhg-gmt-${GSHHG_VERSION}"
+DCW="dcw-gmt-${DCW_VERSION}"
+EXT="tar.gz"
+
+cwd=${PWD}
+# 1. Create temporary directory for building
+rm -rf /tmp/build-gmt
+mkdir -p /tmp/build-gmt
+cd /tmp/build-gmt
```
leouieda, line 22: There is a program called `mktemp` to do this and avoid deleting something that we shouldn't have. Not sure how common it is in Linux distros, though. It seems to be common since it's on the Ubuntu [GNU Core Utils package](https://launchpad.net/ubuntu/trusty/+package/coreutils). So it might be a safe bet that it will be present.

ci/build-gmt-master.sh
```diff
@@ -0,0 +1,62 @@
...
+DCW="dcw-gmt-${DCW_VERSION}"
+EXT="tar.gz"
+
+cwd=${PWD}
+# 1. Create temporary directory for building
+GMT_BUILD_TMPDIR=$(mktemp -d ${TMPDIR:-/tmp/}gmt.XXXXXX)
+cd ${GMT_BUILD_TMPDIR}
+
+# 2. Download GMT, GSHHG and DCW from GitHub
+git clone --depth 10 https://github.com/GenericMappingTools/gmt.git gmt
```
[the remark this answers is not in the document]
seisman: Of course.

seisman:
@joa-quim Any needs from GMT.jl side?

joa-quim:
I guess not, but honestly I'm short on time and my CI recipe works.

PR #3195, opened 2020-04-29
Clean up library functions, again

PaulWessel:
Rearrange internal functions that avoided the first going-through:

1. Eliminate single EXTERN_MSC prototypes in gmt_*.c files by moving those prototypes to gmt_internals.h.
2. Rename several gmt_* functions that are not used in modules to gmtlib_*.
3. Let gmt_map.c include gmt_proj.c to avoid having to export lots of functions just needed in gmt_map.c.
4. Eliminate gmt_compat.c by moving those wrapper functions to the modules they call.

Please test that this does not break anything for the external libs.

Two questions I need help with:

1. Now that gmt_map.c depends on gmt_proj.c, I removed gmt_proj.c from the list of library sources (since it should not be compiled separately).  However, I would like CMake to know that if gmt_proj.c changes then gmt_map.c needs to be recompiled.  How, @seisman ?
2. There is one function whose name fails the convention test: _GMT_blind_change_struct_.  I know it is only used by @joa-quim in Julia.  Since it probably should not be listed in the official API, it should not have capital GMT_ in the front (it is also not listed in any include file like gmt.h).  Would you be able to cope if it was renamed _gmtlib_blind_change_struct_?  Are you including gmt_dev.h anyway? The declaration of _GMT_Get_Ctrl_ makes me think you do.  If that one requires gmt_dev.h (since it returns a struct GMT_CTRL pointer) then it should  be called _gmtlib_get_ctrl_ anyway.

Comments? I WIP'ed this PR so let me know if all is well before I can change that.


seisman:
> 1. Now that gmt_map.c depends on gmt_proj.c, I removed gmt_proj.c from the list of library sources (since it should not be compiled separately).  However, I would like CMake to know that if gmt_proj.c changes then gmt_map.c needs to be recompiled.

I think CMake already knows that gmt_map.c depends on gmt_proj.c.

joa-quim:
Julia does not include any _.h_ or _.c_ it just access the shared lib directly. It does however have to mimic some C structs in Julia language. That is why when one of those mimicked types changes, the Julia wrapper breaks. 

I can cope with a _gmtlib_blind_change_struct_ (will have to deal with another breakage) as long as it keeps being _EXTERN_MSC'ed_

PaulWessel:
OK, thanks @joa-quim so that means I can also change _GMT_Get_Ctrl_ to _gmtlib_get_ctrl_ as long as it is exported.  I would like to remove that from gmt.h since it is not useful to anyone just using the API (since a GMT pointer is not used by the API).

PaulWessel:
OK, I have finished these changes and removed the WIP, @joa-quim and @seisman.

joa-quim:
I'm ok with this.

But there is an issue that is buzzing me. This all set of renamings is adding a lot of changes to history, which will make life way more difficult when we'll have to dig for a particular change. While we are at it, what do you think in adding an kind of EXTERN_MSC2(N) function qualifier that would be a build time parameter. By default it would do nothing, but when =YES it would export that function in the shared lib. Why? Because we could than access all those functions from Julia and write true unit tests at function level.

PaulWessel:
Yes, it is changing history, but I know of no other way to make the code more coherent.  having consistent naming is worth it I think.  I will merge this PR next.

Would you want EXTERN_DEBUG (better name I think) on everything that is not already exported, i.e., all the static functions as well?  Or, do we simply change the definition of GMT_LOCAL from static to EXTERN_MSC in that case? I.e.

```
#ifdef EXTERN_DEBUG
#define GMT_LOCAL EXTERN MSC
#else 
#define GMT_LOCAL static
#endif  
```
Finally, two separate issues for both of you:

1. I made exceptions to the conventions for a few functions: usage, parse, New_Ctrl and Free_Ctrl in all the modules.  These are all statics and clearly EXTERN_DEBUG could not be used given the same names.  Should we be consistent and rename these to grdimage_usage, grdimage_parse, etc? I think that would be the safest way.
2. I know @joa-quim has renamed some of the sub-structures to the local module CTRL structure, i.e., instead of being called struct A {}; you have struct GRDIMAGE_A {};  I think you explained that was needed for Visual studio debug as it gets confused about multiple structs with the same names even though they only have the scope inside a uniquely named main structure? Do we need to rename all the remaining structures the same way?

joa-quim:
1. EXTERN_DEBUG is not a good name. That's not about debugging but external access (i.e. testing or usage). I thought on the _(N)_ because we could than have more than one conditional exports qualifiers for _gmt_*_, _gmtlib_*_, etc.
Making GMT_LOCAL conditionally an export qualifier is a good idea too. But, please no, *usage* is perfect *grdimage_usage* in a grdimage.c source code is just redundant and make code harder to read. For those cases, (usage, parse, New_Ctrl and Free_Ctrl) let's just use static instead of GMT_LOCAL.

2. Definitively yes. The way we have them make it impossible to see the sub-structs contents in VS. I told you once that I was once told that VS is right since the way we have it is an undefined behavior in C, so compilers are free to interpret it as they want.

PaulWessel:
OK, please create two separate issues: One should explain the EXTERN stuff and the other is about renaming sub-structures.  We can then think of scripts to help with those two separate and unrelated tasks.

PR #3201, opened 2020-04-30
Rename module substructure to be unique

### LINKED ISSUE #3198: Renaming sub-structures
We have in many of our *mains* structures with sub-structres using naming conventions like this
```
struct GRDCLIP_CTRL {
	struct In {
		bool active;
		char *file;
	} In;
	struct G {	/* -G<output_grdfile> */
		bool active;
		char *file;
	} G;
	struct S {	/* -Sa<high/above>, -Sb<low/below>, -Si<low/high/between>, -Sr<old>/<new> */
		bool active; 
		...
	} S;
};
```
But Visual Studio debugger really doesn't like that and when debugging the contents of *opt->G* shows up whatever content a previous *struct G* from another module in the library may have. Apparently this naming is an undefined behavior in C.

The good practice is to give those sub-structs unique names, like
```
struct GRDCLIP_CTRL {
	struct GRDCLIP_In {
		bool active;
		char *file;
	} In;
	struct GRDCLIP_G {	/* -G<output_grdfile> */
		bool active;
		char *file;
	} G;
	struct GRDCLIP_S {	/* -Sa<high/above>, -Sb<low/below>, -Si<low/high/between>, -Sr<old>/<new> */
		bool active; 
		...
	} S;
};
```
This has been done in several modules but we need to apply it to all.

PaulWessel:
I.e., use struct PSXY_A instead of struct A.  That way it is unique across the modules.  Closes #3198.

It is possible a few more needs work since some where already renamed but did not use the full name, e.g. SRF_A instead of SURFACE_A.  If anyone wants to hunt down the ones I miss please do.

### DIFF
--- a/src/dimfilter.c
+++ b/src/dimfilter.c
@@ -79,47 +79,47 @@ struct DIMFILTER_INFO {
 };
 
 struct DIMFILTER_CTRL {
-	struct In {
+	struct DIMFILTER_In {
 		bool active;
 		char *file;
 	} In;
-	struct C {	/* -C */
+	struct DIMFILTER_C {	/* -C */
 		bool active;
 	} C;
-	struct D {	/* -D<distflag> */
+	struct DIMFILTER_D {	/* -D<distflag> */
 		bool active;
 		int mode;
 	} D;
-	struct E {	/* -E */
+	struct DIMFILTER_E {	/* -E */
 		bool active;
 	} E;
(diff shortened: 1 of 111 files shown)


seisman:
There are a few conflicts.

PaulWessel:
Should be better now, @seisman.

PR #3213, opened 2020-05-01
Fixing miscellaneous issues in gmtregress

PaulWessel:

Some issues have come to light via [this forum post](https://forum.generic-mapping-tools.org/t/issues-with-interpreting-the-correlation-r-and-coefficient-of-determination-r-in-gmt-regress/435) that required some changes:

1. Except for regular regression with misfit measured vertically (**-Ey**), the Pearsonian correlation coefficient _r_ and coefficient of determination _R_ are not defined (perhaps they are, but not by the same equations; my Google search could not find equivalent expressions for RMA and Ortho regressions - let me know if you can).  The output for the userʻs case gave wild _r_ and _R_.  I now only output these for **-Ey**.
2. The RLS algorithm of dividing the search angle range by 2 etc. could stop the search too early if the second iteration visited the same (crude) angle step and that happened to be the smallest misfit yet.  Then there would be no drop in best misfit and iterations would stop.  I know ensure that this cannot happen until the angle step in the search is at least < 0.05 degrees.
3. The latter change refined the fits in minor ways for three examples and hence their PS files have been updated.

### DIFF
--- a/doc/examples/ex43/ex43.ps
+++ b/doc/examples/ex43/ex43.ps
@@ -1,11 +1,11 @@
 %!PS-Adobe-3.0
 %%BoundingBox: 0 0 595 842
 %%HiResBoundingBox: 0 0 595.0000 842.0000
-%%Title: GMT v6.1.0_9ee9cff_2020.02.16 [64-bit] Document from psbasemap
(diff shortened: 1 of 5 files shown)


PaulWessel:
Actually, will let r be printed since it is entirely data driven.  But R is wrong (going negative).

Issue #3196, opened 2020-04-29
Possibly padding from externals is screwing at times


joa-quim:
This came out from a Julia [issue](https://github.com/GenericMappingTools/GMT.jl/issues/357)
It doesn't happen with linear projection, nor if the **-R** is that of the grid limits
```
G = gmt("grdmath -R70/130/-5/55 -I0.3015075376884 X Y MUL =");
grdimage(G, region=(70, 130, -5, 47), figsize=15, proj=:Mercator, name="stripes.png", show=true)
```

and I can reproduce it in Matlab as well

```
G = gmtmex('grdmath -R70/130/-5/55 -I0.3015075376884 X Y MUL =');
gmtmex('grdimage -R70/130/-5/47 -JM15 -Baf -BWSen -P > lixo.ps', G)
```

![stripes](https://user-images.githubusercontent.com/537321/80641621-0cd37e00-8a5d-11ea-9a87-8071a5c34b22.png)

PaulWessel:
Guessing it is related to the incompatible -R setting:

```
gmt grdcut -R70/130/-5/47 t.grd -Ga.grd
grdcut [WARNING]: (n - y_min) must equal (NY + eps) * y_inc), where NY is an integer and |eps| <= 0.0001.
grdcut [WARNING]: n reset from 47 to 46.8592964824
```

Will have to debug as to what damage this causes internally

PaulWessel:
Close to figuring this out.  When we read a subset from a memory grid we simply increase the pads since we do now have write-permission on the grid.  All that works well (north pad becomes 27 instead of 2).  But in grdimage, somewhere after projection, the north pad is reset to 2.  That cannot possibly be helpful as the G->data is of the same size as the original grid.

PaulWessel:
Fixed that issue so pad is not reset if already manipulated.  But same result in image....Now for yet another EXCOM zoom meeting...

PaulWessel:
Restarting the bug hunt on this after various side-shows.

PaulWessel:
To summarize the observations:

Actions that do **not** cause stripes:

1. Any subset with linear projection
2. Shrink east boundary inward
3. Shrink sound boundary inward

What **does** cause problems (with map projection):

1. Shrink west boundary inward
2. Shrink north boundary inward (case above)

Since no subset and map projection takes the memory grid through the resampling to Mercator without any trouble, the issues is clearly related to some bad interaction from

- Changing west or north
- Projection this subset

[...]

PR #3215, opened 2020-05-02
Do not use grid coordinates if a memory grid is a subset

PaulWessel:

See #3196 for background - this only affects external calls to GMT.  The problem was this: _gmt_grd_project_ does the conversion from a long-lat grid to the Cartesian rectangular projected grid we can plot. In that function we need the x and y coordinates for both original and projected grid.  These are usually available in the header->x and ->y.  However, when a memory grid is passed in we do not have permission to change the memory.  So when a subset is selected, and implemented by a pad change, the x and/or y coordinates are no longer correct.  This is why things screwed, and only if the beginning of the vector was skipped (west, north).  We now check if the pad has been reset, and if so we obtain fresh coordinate arrays.
Also, there was a forced pad reset to 2 when the projected grid was first initialized but that cannot happen when we have external subsets.

Works for me in Matlab.  Needs @joa-quim  to test in Julia.

### DIFF
--- a/src/gmt_map.c
+++ b/src/gmt_map.c
@@ -7731,6 +7731,7 @@ uint64_t gmt_compact_line (struct GMT_CTRL *GMT, double *x, double *y, uint64_t 
 
 /*! . */
 int gmt_project_init (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *header, double *inc, unsigned int n_columns, unsigned int n_rows, unsigned int dpi, unsigned int offset) {
+	struct GMT_GRID_HEADER_HIDDEN *HH = gmt_get_H_hidden (header);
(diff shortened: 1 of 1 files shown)


joa-quim:
You can test it in Julia. #3196 has the Julia commands too.

htyeim:
I think I have time to test it, but I don't know the workflow. I guest it involves compiling the source code [?]. Setting the environment must be complex.  Is there a way to build a Docker container (may be built upon [gcc](https://hub.docker.com/_/gcc), e.g., [ref](https://blobfolio.com/2018/12/using-docker-as-a-build-environment/)) for compile and test so I can do it?

PaulWessel:
Yes, this means building GMT from master branch.  If you want to test this from Julia then let's have @joa-quim chime in on that.

joa-quim:
Works fine in Julia too.

PR #3216, opened 2020-05-02
Let grd2xyz report variable grid spacing x,y,z

### LINKED ISSUE #3108: change the way grd2xyz writes out the x,y values?
If you have a netCDF grid that has irregularly spaced data, when using grd2xyz it writes out the x,y locations at locations corresponding to the average spacing, not the actual x or y locations specified in the x or y array in the netCDF file.

It would be nice if grd2xyz read the actual values from those arrays and wrote them to the xyz file, then the data could be regridded to a uniform lattice for proper plotting in GMT.

not sure if i explained it properly, but i think Paul and Joaquim know what I am talking about!

Thanks!

PaulWessel:
See #3108.  This PR lets **grd2xyz** report variable _x_ and _y_ coordinates if new option **-D** is set (and such coordinates are available in the netCDF grid).  Closes #3108.


seisman:
This is the old behavior. The warnings are very useful.
```
$ gmt grd2xyz NPac_Tohokuh.nc > jk
grd2xyz [WARNING]: The step size of coordinate (Grid Longitude [degrees_east]) in grid NPac_Tohokuh.nc is not constant.
grd2xyz [WARNING]: GMT will use a constant step size of 0.0666667; the original ranges from 0.0665894 to 0.0667114.
grd2xyz [WARNING]: The step size of coordinate (Grid Latitude [degrees_north]) in grid NPac_Tohokuh.nc is not constant.
grd2xyz [WARNING]: GMT will use a constant step size of 0.0529899; the original ranges from 0.0331001 to 0.0667001.
grd2xyz [ERROR]: (x_max-x_min) must equal (NX + eps) * x_inc), where NX is an integer and |eps| <= 0.0001.
grd2xyz (gmtlib_read_grd_info): Use grdedit -A on your grid file to make region and increments compatible [NPac_Tohokuh.nc]
```

This is the new behavior. I don't use **-D** option here, so I expect to see the same warnings. But here I have to use **-V** to see them.
```
$ gmt grd2xyz NPac_Tohokuh.nc -V > jk
grd2xyz [INFORMATION]: Processing input grid(s)
grd2xyz [INFORMATION]: Writing Data Table to Standard Output stream
grd2xyz [INFORMATION]: The step size of coordinate (Grid Longitude [degrees_east]) in grid NPac_Tohokuh.nc is not constant.
grd2xyz [INFORMATION]: GMT will use a constant step size of 0.0666667; the original ranges from 0.0665894 to 0.0667114.
grd2xyz [INFORMATION]: No range attribute, guessing registration to be gridline
grd2xyz [INFORMATION]: The step size of coordinate (Grid Latitude [degrees_north]) in grid NPac_Tohokuh.nc is not constant.
grd2xyz [INFORMATION]: GMT will use a constant step size of 0.0529899; the original ranges from 0.0331001 to 0.0667001.
grd2xyz [INFORMATION]: netCDF grid NPac_Tohokuh.nc information has zmin = zmax = NaN. Reset to 0/0.
grd2xyz [INFORMATION]: Round-off patrol changed geographic grid increment for longitude from 0.0666667247832666565 to 0.0666666666666666657
grd2xyz [INFORMATION]: Reading grid from file NPac_Tohokuh.nc
grd2xyz [INFORMATION]: Set boundary condition for all edges: natural
grd2xyz [INFORMATION]: Set boundary condition for left   edge: natural
grd2xyz [INFORMATION]: Set boundary condition for right  edge: natural
grd2xyz [INFORMATION]: Set boundary condition for bottom edge: natural
grd2xyz [INFORMATION]: Set boundary condition for top    edge: natural
grd2xyz [INFORMATION]: 1377600 values extracted
```

PaulWessel:
I certaintly have to turn of the error which otherwise exits the program.  I reasons that if you use -D then you do so because you want these non-equidistant coordinates and the warning is not that helpful - hence I put it as information.  But if you think it makes more sense to keep them as warnings then we can do that instead.  Then the only difference is the skipping of the error.

seisman:
> I reasons that if you use -D then you do so because you want these non-equidistant coordinates and the warning is not that helpful - hence I put it as information. 

I agree. In the example above, I don't use **-D** but it's still information.

joa-quim:
Why do we have to have an extra option to do the right thing? IMO, either constant or variable spacing should be dealt (if possible) silently.

PaulWessel:
I guess this is falling somewhere between things: Elsewhere in GMT we cannot accept variable x and y arrays (e.g., plotting grids), but we are making an exception for grd2xyz.  It is very limited (no similar xyzgrd possibility).  Since GMT is not set up to handle these grids in general, it seems reasonable to not be completely quiet when one such grid is used.  I can see users being totally confused that grd2xyz works fine (my data is there) but not working in grdimage.  So I think a warning is reasonable and it can explain what is happening.

PaulWessel:
So I am OK with removing -D and doing "the right thing" but it should come with a mild warning.

joa-quim:
OK, I'm ok with the "mild warning".

PaulWessel:
OK, now we get this:

```
gmt grd2xyz NPac_Tohokuh.nc > t.txt
grd2xyz [WARNING]: Grid NPac_Tohokuh.nc has non-equidistant x-coordinates (see grd2xyz docs for discussion)
grd2xyz [WARNING]: Grid NPac_Tohokuh.nc has non-equidistant y-coordinates (see grd2xyz docs for discussion)
```

and the man page has a new section discussion non-equidistant grid options.

PR #3218, opened 2020-05-02
Fix OpenMP in grdimage

PaulWessel:

The use of Open-MP in grdimage was commented out as it gave subtle errors.  Finally noticed that a shared variable (step) was missing in the pragma.  This PR seems OK to me for grdimage.  A few tests failed but they were all related to gpsgridder (which may need some looking into).


seisman:
Fails on Windows.
```
..\src\grdimage.c(1207): error C3016: 'row': index variable in OpenMP 'for' statement must have signed integral type
```

PR #3220, opened 2020-05-02
Let developers export all gmtlib functions

### LINKED ISSUE #3199: Exporting symbols in shared libs
Windows and Unix behave differently in what goes exported into shared libs. While Unix exports everything by default, Windows only exports the symbols explicitly marked to be exported. Because exporting everything makes the shared libs larger and is prone to name clashing we use the ``GMT_LOCAL`` qualifier to many functions that are meant to be local to the source file (for example the *usage* function in all modules). We also use a lot of ``EXTERN_MSC`` when exporting functions from one source code to be used in another. But there are still lots of functions that are not qualified at all and as such they are not accessible by external codes on Windows.

Now, being accessible from external would provide not only a potential larger usage ground but specially would allow writing unit tests in Julia.  But we still do not want to export everything on regular builds. An option is to create EXPORT qualifiers that would just do nothing on default builds but allow an export value with a building option. As an example
``EXPORT_GMTLIB gmtlib_xxx``
would export all *gmtlib_xxx* functions when a, say, -DEXPORT_GMTLIB=YES" cmake build options is used. And a similar one for ``EXPORT_GMT gmt_xxx``. Also the ``GMT_LOCAL`` could be conditionally turned into an export directive. If seen useful we can even create an ``EXPORT_ALL`` for the remaining functions. This one would set ``EXPORT_GMTLIB`` and ``EXPORT_GMT`` to YES. This way we can have different levels of functions exports, depending if only for testing or for real third party use.

PaulWessel:
See #3199.  This redefines **GMT_LOCAL** to mean **MSC_EXPORT** when **-DEXPORT_GMTLIB** is added to the CFLAGS. No duplicate symbols found (compiles and links fine) and all tests pass on macOS. Various structs and some remaining parse, usage functions set to static. Oh, and some minor function formatting cleaning.  Closes #3199.

### DIFF
--- a/src/gmt_dev.h
+++ b/src/gmt_dev.h
@@ -83,9 +83,6 @@ extern "C" {
 #	pragma warning( disable : 4244 )	/* conversion from 'uint64_t' to '::size_t', possible loss of data */
 #endif
 
-/* Used to restrict the scope of a function to the file it was declared in */
-#define GMT_LOCAL static
-
 /* CMake definitions: This must be first! */
 #include "gmt_config.h"
 
@@ -114,6 +111,14 @@ extern "C" {
 #include <errno.h>
 
 #include <time.h>
+
+#ifdef EXPORT_GMTLIB
+/* Used to export everything so external enviroments can do unit tests */
+#	define GMT_LOCAL EXTERN_MSC
+#	else
+/* Used to restrict the scope of a function to the file it was declared in */
+#	define GMT_LOCAL static
+#endif
 
 #include "gmt_common_math.h" /* Shared math functions */
 #include "gmt.h"             /* All GMT high-level API */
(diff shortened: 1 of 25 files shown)


src/gmt_dev.h
```diff
@@ -115,6 +112,14 @@ extern "C" {
 
 #include <time.h>
 
+#ifdef EXPORT_GMTLIB
+/* Used to export everything so external enviroments can do unit tests */
+#	define GMT_LOCAL EXTERN_MSC
+#	else
```
seisman, line 118:
```suggestion
#else
```

PaulWessel:
Agreed, @joa-quim please add the **-DEXPORT_GMTLIB** and see if you get any conflicts building GMT on Windows.

PaulWessel:
Ping @joa-quim to see if this passes compilation, linking on Windows.  I believe it will since the CI passed.

joa-quim:
I have some name clashing with my MBGMT but that's on y side to solve.

seisman:
Merge this PR?

PaulWessel:
Let's get a confirmation from @joa-quim first.  I know he said he had stuff on his mbgmt end but we dont have to rush this one unless it is OK.  So let us know @joa-quim .

joa-quim:
OK to merge 

Sent from my iDedo

PR #3226, opened 2020-05-02
Fix list of private and shared variables for OPM in gpsgridder

PaulWessel:
It had a woefully wrong list of variables, some missing, and some private listed in the shared.  With these changes the gpsgridder-based tests passes under OpenMP.


PaulWessel:
I believe with this all tests passes on macOS with gcc9 and OpenMP build.

PR #3227, opened 2020-05-02
Let the macOS bundle be built with OpenMP

PaulWessel:
As all tests pass with a MP build we should release a bundle built that way.  This PR improves the admin/build-release.sh in a few ways:

1. It loads in cached CMake variables to use gcc9 MP compiler
2. It checks that we remembered to set GMT_PUBLIC_RELEASE

However, I got this message right away:

`build-release.sh: Need Sphinx version 1 to build release`

Since mine is 3.0.3 I think it is reasonable to review why we need version 1.  For macports this is the py-sphinx package so I will need to temporarily switch to version 1, which it looks like I dont have, so need to install.  Please explain why we need 1 and what it might take to move to 3; I can then improve the documentation of this script (if we cannot move past 1).


admin/build-release.sh
```diff
@@ -54,6 +54,11 @@ if [ "X${GMT_DCW_SOURCE}" = "X" ]; then
 	exit 1
 fi
 
+if [ $(egrep -c '^set \(GMT_PUBLIC_RELEASE TRUE\)' cmake/ConfigDefault.cmake) -eq 0 ]; then
```
seisman, line 57: We set GMT_PUBLIC_RELEASE to TRUE in `admin/ConfigReleaseBuild.cmake`. Why do we need to check it here?

admin/cache-mp-gcc.cmake
```diff
@@ -0,0 +1,9 @@
+# Cache settings for building the macOS release with GCC and OpenMP
```
seisman, line 1: What about including these lines in the `admin/build-release.sh` script, so that we don't have to maintain too many files for releasing GMT.
```
cat > cache-mp-gcc.cmake << EOF
SET ( CMAKE_C_COMPILER "/opt/local/bin/gcc-mp-9" CACHE STRING "GNU MP C compiler" )
SET ( CMAKE_CXX_COMPILER "/opt/local/bin/g++-mp-9" CACHE STRING "GNU MP C++ compiler" )
SET ( CMAKE_C_FLAGS -flax-vector-conversions CACHE STRING "C FLAGS")
SET ( CMAKE_C_FLAGS_DEBUG -flax-vector-conversions CACHE STRING "C FLAGS DEBUG")
SET ( CMAKE_C_FLAGS_RELEASE -flax-vector-conversions CACHE STRING "C FLAGS RELEASE")
SET ( OpenMP_C_FLAGS -flax-vector-conversions CACHE STRING "C FLAGS OPENMP")
EOF
```

admin/build-release.sh
PaulWessel: Forgot. Just removed it

admin/cache-mp-gcc.cmake
PaulWessel: OK, done.

PaulWessel:
Looks very hard to try to install py-sphinx 1.8.5 under macports.

seisman:
The reason that we need Sphinx 1.x is because the sphinx_rtd_theme we're using is incompatible with Sphinx 2.x. Sphinx 2.x can still build the documentation without errors but the styling may look bad. We have made several patches to the official sphinx_rtd_theme, so I think the documentation is good now. 

I think you can try to build the documentation with Sphinx 3.x and see if everything works fine.

admin/build-release.sh
seisman: @PaulWessel I just realized that the check is still necessary. 

The build-release.sh script also builds the source tarballs. If we forget to set GMT_PUBLIC_RELEASE to TRUE, the macOS bundle is still correctly built, but the ConfigDefaults.cmake in the source tarballs have a FALSE GMT_PUBLIC_RELEASE.

PaulWessel: OK, will reinstate.

seisman:
I just tried to build GMT with GCC-9 from homebrew. It gives me a lot of warnings and errors. But I'm OK with the PR as long as it works for you.

seisman:
The documentation built by Sphinx 3.x looks good to me. I think we can use Sphinx 2.x and 3.x now.

PaulWessel:
Yes agree on sphinx - I've built the docs for a long time and not seen anything.  I will remove the test.

PaulWessel:
is there any way we can completely avoid having to make edits to the repo files in cmake? The build-release script got Ctrl-C'ed and my own cmake/ConfigUsers.cmake was gone...  This is a bit too fragile.

PaulWessel:
While my own gcc OMP build works, the build-release.sh fails.  I traced it to GLIB GTHREAD support.  That is disable when I build but it becomes enabled when build-release.sh is run because GMT_USE_THREADS is set in admin/ConfigReleaseBuild.cmake.  WIth it, this happens:

So I get stuck at grdfilter which is using glib.  I really want to rewrite grdfilter using OpenMP since it is so much simpler, but then there are others.  Anyway, here is the error (one of many):

```
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/_stdio.h:68,
                 from /opt/local/lib/gcc9/gcc/x86_64-apple-darwin19/9.3.0/include-fixed/stdio.h:78,
                 from ../src/postscriptlight.h:46,
                 from ../src/gmt_dev.h:101,
                 from ../src/grdfilter.c:48:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/pthread.h:249:31: error: expected ';', ',' or ')' before 'restrict'
  249 |   void * _Nullable * _Nonnull __restrict, size_t * __restrict);
      |                               ^~~~~~~~~~
Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/pthread.h:543:23: error: conflicting types for '_Nonnull'
  543 |   void * _Nullable (* _Nonnull)(void * _Nullable), void * _Nullable);
      |   ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/pthread.h:542:25: note: previous definition of '_Nonnull' was here
  542 |   pthread_t _Nullable * _Nonnull, const pthread_attr_t * _Nullable,
      |   ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
```

This Nullable shit we have encoutnered before (grep Nullable *.c) so I added this to grdfilter.c:

```
#ifdef	__APPLE__
	/* Apple Xcode expects _Nullable to be defined but it is not if gcc */
#ifndef _Nullable
#	define _Nullable
#	endif
#	endif
```

but made no difference.

PaulWessel:
The MP build works fine and bundle is built if I disable GLIB_THREADS.

PaulWessel:
If I enable GLIB_THREADS in my regular non-MP then the compile works since the includes are included and no complaints.  It is just some mismatch between GCC and Apple include files.

admin/build-release.sh
```diff
@@ -1,10 +1,15 @@
...
+# If run under macOS it also builds the macOS Bundle.
+#
 # Requirements in addition to libraries and tools to build GMT:
 #	1) pngquant for squeezing PNG files down in size
-#	2) GMT_GSHHG_SOURCE and GMT_DCW_SOURCE enviromental parameters set
+#	2) GMT_GSHHG_SOURCE and GMT_DCW_SOURCE environmental parameters set
 #	3) A ghostscript version we can include in the macOS bundle [MacPort]
+#   4) Version 1.* of build-sphinx
+#   5) grealpath (package coreutils)
+#   6) GNU tar (package gnutar)
```
seisman, line 12:
```suggestion
#	4) sphinx-build 
#	5) grealpath (package coreutils)
#	6) GNU tar (package gnutar)
```

PR #3237, opened 2020-05-03
New GMT defaults GMT_MAX_CORES

PaulWessel:

As GMT modules gain more multi-threaded capabilities, people will use the **-x** option more often.  However, the movie module is using all available cores when it spawns  shell scripts, hence the commands in those shells should not use multi-core since it will just lead to time-sharing and competition.  The only mechanism in those scripts to turn off **-x** is via a limit on usable cores set as  GMT defaults.  The default setting is 0 which means as many as you want.  However, the movie mainscript will be issued a `gmt set GMT_MAX_CORES 1` setting to avoid the problem.


seisman:
If I don't use the **-x** option, is grdimage multi-threaded or not?

PaulWessel:
Pretty sure the default is full multi-threaded if compiled that way - you can use -x to limit the number of cores

doc/rst/source/gmt.conf.rst
```diff
@@ -422,6 +422,10 @@ GMT Miscellaneous Parameters
         or ~/.gmt. **Note**: Some of these languages may require you to also
         change the :term:`PS_CHAR_ENCODING` setting.
 
+    **GMT_MAX_CORES**
+        Sets the upper limit on the number of cores a multi-threaded module might
+        use if **-x** is selected [0, i.e., as many as possible].
```
seisman, line 427: **GMT_MAX_CORES** only works if **-x** is selected?

How many cores are used for the following command?
```
gmt set GMT_MAX_CORES 2
gmt grdimage ...
```

PaulWessel: That is 1 unless you use -x and now you should get 2 (since you set that as a limit).

seisman: Sorry, I'm more confused. Let's assume I have a 4-core PC:

- **grdimage**: 4 cores
- **grdimage -x**: 4 cores
- **grdimage -x1**: single core

After setting GMT_MAX_CORES=1, how many cores are used in the following commands?

- **grdimage**: 4 cores or single core?
- **grdimage -x**: single core
- **grdimage -x2**: 2 cores or single core?

PaulWessel: Initially, your available cores is 4. GMT_MAX_CORES (-0) is therefore 4.  However, if you set GMT_MAX_CORES to 1 then you effectively cannot use more that 1 core per process.  So grdimage -x2 will get 1 core.  It has to be this way so that your grdimage call in a movie script does not do -x4 and then we have 4 of those jobs running at the same time all running grdimage -x4.

PaulWessel: So with GMT_MAX_CORES=1 you only have one core no matter what -x you do.

seisman: Then the description is misleading. When I read it, my understanding is that GMT_MAX_CORES has no effects if I don't use **-x**, thus **grdimage** without **-x** still uses 4 cores.
```
Sets the upper limit on the number of cores a multi-threaded module might
use if **-x** is selected [0, i.e., as many as possible].
```

PaulWessel: OK, will clarify.

PaulWessel:
> Looks good to me.
> 
> It's better to add one sentence in the movie -x documentation, so that users won't be surprised that their plotting scripts with grdimage are much slower when used for movie.

DOne.

PR #3238, opened 2020-05-03
Sub directory for full animations

PaulWessel:

Because of CI limitations and computational requirements, we do not have actually full-length high-resolution animation examples in GMT.  However, several of these have been made (e.g., the eLightning presentation at AGU and others) with ore to come.  I would like to keep these examples in the GMT repo under the proposed doc/examples/full_animations.  There are some conditions:

1. These are not included in any automatic runs, CI, or build sequences.
2. They should not contain any large datasets; if these are needed they must use remote files.
3. There needs to be a mechanism for developers to periodically run one or all of these.

For example the Pacific MOR flyover HD movie using @earth_relief_30m takes 6 hours on my 24-core MacPro to render.  It is a very good example of what GMT can do but we obviously do not want to run this script automatically.  The full movies will be displayed on our GMT YouTube channel.

This PR establishes the directory and the first animation script, with more to be added as I do the forensics on the AGU folder.


seisman:
What about sub-directories "movie01", "movie02" in the `doc/examples` directory? We will add it to the gallery anyway. Having names like movieXX making the reST much easier.

seisman:
Need to rename MOR_flyover.sh to movie01.sh.

PR #3242, opened 2020-05-04
Let movie delete input scripts if selected

### LINKED ISSUE #3241: moviexx.sh products

Currently, a moviexx.sh (only have movie01.sh today) will produce a PNG and a MP4 in the movie01 directory.  This is not ideal as a huge MP4 might be accidentally included in the repo.  Perhaps we should add

```
doc/examples/*.mp4
doc/examples/*.png
doc/examples/*.gif
```

to the .gitignore file? Is that the best solution?

PaulWessel:

Most/all of our GMT animations are scripts the simply build the required movie script parts on the fly, then feed these to the movie module, and then we have to manually delete all the parts.  This PR extends the **-Z** option with an optional s directive which will also delete the input scripts given (_mainscript_ plus any scripts given via **-E, -I, -S**).  All animation scripts have been updated with **-Zs** and removing the manual rm -f lines.
I also added 3 lines to .gitignore so that no MP4, PNG, GIF created in the movie directories are accidentally merged into the repo.  Closes #3241.

### DIFF
--- a/doc/examples/movie01/movie01.sh
+++ b/doc/examples/movie01/movie01.sh
@@ -18,6 +18,7 @@
 # https://www.youtube.com/watch?v=LTxlR5LuJ8g
 # The movie took ~6 hours to render on a 24-core MacPro 2013.
 
+gmt set PROJ_LENGTH_UNIT inch FONT_TAG 20p,Helvetica,white
 cat << EOF > pre.sh
 #!/bin/bash
 # Pre-script: Runs once to produce files needed for all frames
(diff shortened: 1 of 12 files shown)


doc/examples/movie01/movie01.sh
```diff
@@ -18,6 +18,7 @@
 # https://www.youtube.com/watch?v=LTxlR5LuJ8g
 # The movie took ~6 hours to render on a 24-core MacPro 2013.
 
+gmt set PROJ_LENGTH_UNIT inch FONT_TAG 20p,Helvetica,white
```
seisman, line 21: Move this line just before the "gmt movie" command?

PaulWessel: Yep done

PR #3247, opened 2020-05-04
doc: Separate gallery and animations

### LINKED ISSUE #3246: Separate pages for examples and animations
Now we have 50 examples, 8 animations, and 1 HD animation. They are all listed in the [Gallery page](https://docs.generic-mapping-tools.org/latest/gallery.html) and the animations are listed at the bottom.

As we expect to add more examples and movies in the future, I think it's better to separate them into two pages, gallery.html and animations.html. It also makes GMT's movie-making feature more visible to users.

seisman:
Close #3246.

This is what the sidebar looks like：

image


seisman:
Ping @PaulWessel

seisman:
Updated.

image

PaulWessel:
Hm, what do you think of the GMT Colors, GMT Modules, GMT Cookbook etc.  is the GMT really needed there? All it does to me is wonder if the others without GMT is about non-GMT things.  The whole site is about GMT so perhaps not use that in any headers below the very top.

seisman:
Done.

PR #3249, opened 2020-05-04
Let movie fading accept different fade colors

PaulWessel:

Apart from the bug in #3248, we had no way of selecting a different fade color.  For light movies, fading from/to white makes more sense than black.  Now, both the **-E** and **-K** options take **+g**_fill_ with default black. Also, the syntax of **-E** and **-K** have been made more similar, and now in sync with the documentation.


seisman:
Something wrong with the CI. Close and reopen to re-trigger the CI jobs.

PR #3251, opened 2020-05-04
Let movie build transparent PNG images

### LINKED ISSUE #3243: Transparent movies?

The **movie** module works by converting _PostScript_ plots to fixed-size png images and then these are converted to a movie (MP4 or GIF).  It was assumed that each frame replaces the previous frame in a movie and hence these are **opaque** pngs (format png).  Thus, a dumb movie that plots a point along a sine:

```
gmt begin
	y=\`gmt math -Q \${MOVIE_FRAME} SIND =\`
	echo \${MOVIE_FRAME} \$y | gmt plot -R0/360/-1.2/1.6 -JX3.5i/1.65i -Sc0.1i -Gred -BWSne -Bxa90g90f30+u@. -Bya0.5f0.1g1  -X0.35i -Y0.25i  --FONT_ANNOT_PRIMARY=9p
gmt end
```

and run it this way:

`gmt movie main.sh -C4ix2ix125 -T360 -Ntransp -D24 -A+l -Zs`

results in the GIF animation below:

opaque

Now, if we had an option to select **transparent** png images (format PNG; here I just changed the code) then the same movie looks like this:

transp

Here, the movie is the accumulated concatenation of all the transparent images up to a given frame.  This brings up a few questions:

1. It seems this effect can be useful for some movies so I would argue we need to control if gmt makes png or PNG.  Depending on where it is displayed (powerpoint, web), the movie itself may be transparent (depending on what is plotted [have not checked this yet]).
2. You can tell that the basemap frame has been stacked many times. It is curious to me that it does not start out like the top then get fatter, but perhaps if we played it at a very low frame rate we would see that the 2nd frame is fatter than the first.  Anyway, this begs the question if when transparent PNG is requested that perhaps any backgrounds layer (which might produce the basemap) shall only be included for frame 0 since otherwise we are just stacking them.
3. There may be other subtleties that will appear with more testing, but that is it for now.

Thoughts, @joa-quim and @seisman?

PaulWessel:

As an experimental feature, a new **+t** modifier to **-F** will change the format of the PNG images being created from opaque to transparent.  At the moment, few video codecs support transparency; VP9 in WebM being one exception. Also, animated GIF (**-A**) can also accept transparent images and here each additional frame is added to the previous, creating an artistic effect.  Users with commercial tools can use the transparent images to build a transparent video.  The movie documentation discusses the experimental nature, but the **-F**_webm_**+t** setting currently works.  Results may depend on having a very recent FFmpeg tool.
Closes #3243.

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -359,6 +359,15 @@ making a HD movie using the US unit dimensions then a background pink layer woul
 Note the canvas selection via :term:`PS_MEDIA`, the matching region and projection, and
 the zero location of the origin.
 
+Transparent images
+------------------
+
+By default, **movie** will build opaque PNG images which can then be assembled into an animation.
+Use **-F** modifier **+t** to instead make transparent PNG images.  Currently, few video codecs support
+transparency.  It is claimed both H.265 (HECV) and VP9 (Webm) offers this capability.
+Animated GIFs are able to build an animation from transparent PNGs.  Experts may create the transparent
+PNGs and take things from there.
+
 Technical Details
 -----------------
 
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -363,10 +363,11 @@ Transparent images
 ------------------
 
 By default, **movie** will build opaque PNG images which can then be assembled into an animation.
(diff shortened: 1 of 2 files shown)


seisman:
Warn "The feature is experimental" when -F+t is used?

PR #3253, opened 2020-05-05
Correct external function access to GMT libraries

### LINKED ISSUE #3233: Runing from externals is broken on Linux
https://travis-ci.org/github/GenericMappingTools/GMT.jl/builds/682647264

image

PaulWessel:

The handling of the shared libgmt library was initially expected to only be called from gmt.c.  However, with GMT/MATLAB, GMT.jl and pyGMT we need to make a few changes.  Because the different OS has different behaviors, we did not really notice a problem until we made a clean separation between supplements and core library and @joa-quim tested GMT.jl on Linux.  There it failed to find and open the core library.  Various forensics lead to the special handling in gmt_api of opening the core library by passing NULL as the path.  Clearly, this was only meant to be used for gmt.c.  This PR checks if we are creating sessions from external environments, and if we do we treat the libgmt library the same way as any other shared library: We open it with dlopen the first time it is needed.
Closes #3233.


joa-quim:
Now it breaks Windows
```
6.1.0_4e0d12f_2020.05.04
GMT [ERROR]: Unable to open GMT shared core library: Error <126>: The specified module could not be found.

ERROR: LoadError: GMT: No module by that name -- grdmath -- was found.
```

PaulWessel:
OK, faculty meeting in 10 minutes...

PaulWessel:
Maybe you can debug that and see.  I added creating the path to libgmt.

PaulWessel:
probably wrong extension etc.

joa-quim:
I don't understand. My dlls are called xxx_w64.dll or xxx_w32.dll. Why?
```
/* Name of core library */
#define GMT_CORE_LIB_NAME "gmt.dll"

/* Name of supplemental library */
#define GMT_SUPPL_LIBRARY "supplements.dll"

/* Name of PSL library */
#define PSL_LIB_NAME "postscriptlight.dll"
```

But it was calling gmt.dll before and it still worked.

seisman:
This branch works for PyGMT and GMT.jl, on both Linux and macOS.

joa-quim:
Almost for sure that the problem is this
```
#define GMT_CORE_LIB_NAME "gmt.dll"
```
instead of this
```
#define GMT_CORE_LIB_NAME "gmt_w64.dll"
```
because ``GMT_CORE_LIB_NAME`` was set before the coming of 
```
if (WIN32 AND GMT_DLL_RENAME)
	set_target_properties (gmtlib PROPERTIES RUNTIME_OUTPUT_NAME ${GMT_DLL_RENAME})
endif (WIN32 AND GMT_DLL_RENAME)

```

seisman:
These lines are in the src/CMakeLists.txt and are executed after `add_subdirectory (src)`. 
```
if (WIN32 AND GMT_DLL_RENAME)
	set_target_properties (gmtlib PROPERTIES RUNTIME_OUTPUT_NAME ${GMT_DLL_RENAME})
endif (WIN32 AND GMT_DLL_RENAME)
```

So, GMT_CORE_LIB_NAME is set after that. 
``` 
# Get name of GMT core library
get_target_property (_name gmtlib OUTPUT_NAME)
get_target_property (_prefix gmtlib PREFIX)
set (GMT_CORE_LIB_NAME ${_prefix}${_name}${_lib_suffix})
```

Here we get the OUTPUT_NAME property, not sure if we should change it to RUNTIME_OUTPUT_NAME for Windows.

seisman:
RUNTIME_OUTPUT_NAME should override OUTPUT_NAME.
```
Output name for RUNTIME target files.

This property specifies the base name for runtime target files. 
It overrides OUTPUT_NAME and OUTPUT_NAME_<CONFIG> properties.
```

joa-quim:
I can make it work if I make it
```
if (WIN32 AND GMT_DLL_RENAME)
	set (GMT_CORE_LIB_NAME ${GMT_DLL_RENAME}.dll)
else (WIN32 AND GMT_DLL_RENAME)
	set (GMT_CORE_LIB_NAME ${_prefix}${_name}${_lib_suffix})
endif (WIN32 AND GMT_DLL_RENAME)
```
and likewise for the other two. But what about custom supplements? They are not covered by this. And indeed a quick test
```
julia> I = gmt("mbimport -I C:/j/OT/PatchSurvezEM122PHINS/0000_20111219_154640_METEOR_EM122.mb59")
ERROR: Something went wrong when calling the module. GMT error number = 71
```
I really need this to keep working because it's the base of my use of MB in Mirone. Need to test what happens in Matlab but the above failure is worrying.

PaulWessel:
Just pointing out this is progress: Clean decoupling from supplements of all types. What we had was a mess (despite working). Yes, still need to figure out the windows naming schemes for supplements.  Seems that should be in the hand of the developer, so in their CMakeLists.txt file and not something on our side, no?

joa-quim:
Indeed it's broken in Matlab, so no merging of this while not solved
```
>> I = gmtmex('mbinfo -I C:/j/OT/PatchSurvezEM122PHINS/0000_20111219_154640_METEOR_EM122.mb59')
GMT: Module return with failure while executing the command
mbinfo -I C:/j/OT/PatchSurvezEM122PHINS/0000_20111219_154640_METEOR_EM122.mb59
Error using gmtmex
GMT: exiting
```

joa-quim:
Probably it was already broken before this PR.

PaulWessel:
Please run with -V so we can tell the name of the library.  I temporarilly moved this from DEBUG to INFORMATION:

`GMT_Report (API, GMT_MSG_INFORMATION "Shared Library # 0 (core). Path = %s\n", API->lib[0].path);`

PaulWessel:
> Probably it was already broken before this PR.

Yes it was.  But the initial problem and solution was not what we thought it was - it had to do with the _dlopen_special_ treatment and we "fixed" it by having the supplement include core external prototypes - not good, but that is why it "worked".

joa-quim:
There is no -V in mbinfo, so no difference.
"worked" is preferable to "not-good"

PaulWessel:
Of course, but we want correct and working.

joa-quim:
That's another discussion. We cannot tell people that we have this and that and then breaking it.

seisman:
@joa-quim When GMT_DLL_RENAME is set to "gmt_w64", we will have gmt.lib and gmt_w64.dll in the GMT's bin directory. Does it cause any problems?

joa-quim:
It doesn't cause problems per se but users that want to link against gmt.lib are expecting to find it under /lib and not /bin

But when I do ``make install`` it puts ``gmt.lib`` under ``lib`` not ``bin``. Why are you saying that?

seisman:
> It doesn't cause problems per se but users that want to link against gmt.lib are expecting to find it under /lib and not /bin
> 
> But when I do `make install` it puts `gmt.lib` under `lib` not `bin`. Why are you saying that?

You're right. It's lib/gmt.lib and bin/gmt_w64.dll.

PaulWessel:
> OK, I rebuilt my MBGMT supplement and it now works again. Don't understand why I had to rebuild it but this is the least (at least for me). So I think this PR is ready to merge, under two conditions:
> 
> 1. @PaulWessel  will never again doubt the Julia findings
> 2. @PaulWessel acknowledges that he knows enough GMT.jl to do testings (replace ' by " in gmtmex). He also has a working GMT.jl version since he added this to his `~/.julia/config/startup.jl` file
> 
> ```
> using Libdl
> push!(Libdl.DL_LOAD_PATH, "/Users/j/programs/gmt6/lib")
> ```

Hm, only if @joa-quim acknowledges that making all these changes was a good and necessary thing.

joa-quim:
> Hm, only if @joa-quim acknowledges that making all these changes was a good and necessary thing.

Ok, deal. But my Bank account of debugging hours must be credited with 1 k units

src/gmt_api.c
```diff
@@ -975,21 +975,33 @@ GMT_LOCAL int gmtapi_init_sharedlibs (struct GMTAPI_CTRL *API) {
...
+	else {	/* The handling of the core library is only special when gmt.c is used. */
+		API->lib[0].path = strdup (GMT_CORE_LIB_NAME);
+		GMT_Report (API, GMT_MSG_DEBUG, "Loading core GMT shared library: %s\n", API->lib[0].path);
+		if ((API->lib[0].handle = dlopen_special (API->lib[0].path)) == NULL) {
+			GMT_Report (API, GMT_MSG_ERROR, "Failure while loading core GMT shared library: %s\n", dlerror());
+			gmtapi_exit (API, GMT_RUNTIME_ERROR); return GMT_RUNTIME_ERROR;
+		}
+		dlerror (); /* Clear any existing error */
+	}
+	GMT_Report (API, GMT_MSG_INFORMATION, "Shared Library # 0 (core). Path = %s\n", API->lib[0].path);
```
seisman, line 1004: Change it back to DEBUG and merge?
```suggestion
	GMT_Report (API, GMT_MSG_DEBUG, "Shared Library # 0 (core). Path = %s\n", API->lib[0].path);
```

seisman:
Ping @PaulWessel, merge this PR?

PR #3254, opened 2020-05-05
Fix GMT_XXX_LIB_NAME when XXX_DLL_RENAME is set

seisman:
For the GMT core library, the properties OUTPUT_NAME and RUNTIME_OUTPUT_NAME are set to gmt by default (e.g. libgmt.dylib, libgmt.so and gmt.dll). 

For Windows, if GMT_DLL_RENAME is set, then RUNTIME_OUTPUT_NAME is changed. Thus, we can use the property RUNTIME_OUTPUT_TIME to build the variable GMT_CORE_LIB_NAME.


seisman:
@joa-quim I add a commit on top of @PaulWessel's branch. This should fix the problem for the gmt.dll library.

ci/config-gmt-windows.sh
```diff
@@ -12,6 +12,8 @@ set (DCW_ROOT "$ENV{COASTLINEDIR}/dcw")
 set (GMT_ENABLE_OPENMP TRUE)
 
 set (GMT_DLL_RENAME "gmt_w64")
+set (PSL_DLL_RENAME "postscriptlight_w64")
+set (SUPP_DLL_RENAME "supplements_w64")
```
joa-quim, line 16: I don't think this is correct. Those are not fixed names. They will be ``postscriptlight_w32`` for 32 bits builds

joa-quim: ``PSL_DLL_RENAME`` are variables set by me in my ConfigUser.cmake

seisman: They're only for the CI jobs. Currently, our CI jobs are building 64-bit only.

seisman:
With the changes in this PR, now we have:

Linux:
```
*  Build GMT core             : always [libgmt.so]
*  Build PSL library          : always [libpostscriptlight.so]
*  Build GMT supplements      : yes [supplements.so]
```

macOS
```
*  Build GMT core             : always [libgmt.dylib]
*  Build PSL library          : always [libpostscriptlight.dylib]
*  Build GMT supplements      : yes [supplements.so]
```

Windows (with **XXX_DLL_RENAME** set)
```
*  Build GMT core             : always [gmt_w64.dll]
*  Build PSL library          : always [postscriptlight_w64.dll]
*  Build GMT supplements      : yes [supplements_w64.dll]
```

joa-quim:
Then no need to use any of the ``XXX_RENAME_DLL`` variables and just stick to the default ``gmt.dll``, etc ...

seisman:
> Then no need to use any of the `XXX_RENAME_DLL` variables and just stick to the default `gmt.dll`, etc ...

Yes, just to test if XXX_DLL_RENAME work as we expect. I will remove them if this PR is going to be merged.

joa-quim:
OK.

seisman:
Now it's ready for reviewing and merging.

joa-quim:
OK, I tested this branch and see that ``config.h`` has
```
/* Name of core library */
#define GMT_CORE_LIB_NAME "gmt_w64.dll"

/* Name of supplemental library */
#define GMT_SUPPL_LIBRARY "supplements_w64.dll"

/* Name of PSL library */
#define PSL_LIB_NAME "postscriptlight_w64.dll"
```

So all seems fine.

PR #3258, opened 2020-05-05
Movie improvements and animation test scripts

PaulWessel:

This PR adds a new animation directory under test which is excluded from the automatic checks.  Its use is for collecting various test scripts that examines aspects of movie.  Because movies take time to generate and because we don't want to compare these movies with a set of reference movies or large sets of frames (at least not yet).
The PR also improves the syntax and parsing of **-K** and **-E** and the first test script (title.sh) generates 19 different short movies exploring various combination of how fading options affect a simple title plus animation sequence.  Documentation is also updated.


PaulWessel:
Would anyone know how to make csh shut up about changing directories?  in csh, it is apparently considered **a good idea** to print the directory you are in after cd.  This sounds like some interactive mode to me but we are running batch. I cannot find an option for csh to turn that off without very convoluted redirecting of stderr (which is not easy in csh to begin with).
I know, csh sucks but this is for important GMT users who still use csh at Scripps...

PR #3262, opened 2020-05-06
doc: Remove unused variable $ps

seisman:
the varialbe $ps is defined but never used.


seisman:
The ps checking is done by gmtest and the ps file name is assumed to be the same as the script name. 

animate.sh just copy all the animation scripts to the cmake binary directory and run them. ps is not needed there.

PR #3263, opened 2020-05-06
Shut up the csh interactive mode

PaulWessel:
Been so long since I used csh I forgot about **-f**.


seisman:
I just ran the following csh script and didn't see any printed directory in my terminal.
```
#!/usr/bin/env csh

mkdir build
cd build
```

This is what **csh -f** says
```
       -f  The  shell  does not load any resource or startup files, or perform
           any command hashing, and thus starts faster.
```
Perhaps you have some customized initialization files for your csh shell?

PaulWessel:
Shit, an ancient .tchsrc sitting around... OK, will close this PR.

PR #3275, opened 2020-05-07
Keep improving events docs and scripts

PaulWessel:
Clarify how labels are plotted.

### DIFF
--- a/doc/examples/anim09/anim09.sh
+++ b/doc/examples/anim09/anim09.sh
@@ -6,7 +6,7 @@
 # path to simulate a constant velocity at the given altitude, with synthetic
 # banking as we turn to follow the path.  We use the 30 arc second global relief
 # grid and overlay a few labels for named features.
-# Grid:   @earth_relief_30s.grd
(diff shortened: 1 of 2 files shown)


doc/examples/anim09/anim09.sh
```diff
@@ -6,7 +6,7 @@
 # path to simulate a constant velocity at the given altitude, with synthetic
 # banking as we turn to follow the path.  We use the 30 arc second global relief
 # grid and overlay a few labels for named features.
-# Grid:   @earth_relief_30s.grd
```
seisman, line 9: @PaulWessel I copied all the comments in the script to `doc/rst/source/animations/anim09.rst`. See https://docs.generic-mapping-tools.org/dev/animations/anim09.html for how it looks like. Perhaps we should remove the comments from the script and only maintain the one in the reST file.

PaulWessel: Yes I agree, and of course I just added a PR for 7 and 8 as well.

PR #3284, opened 2020-05-08
Warn if transparency and gs is 9.51052

PaulWessel:
Do that because that GS version is screwed.  Not tested though.

### DIFF
--- a/src/psconvert.c
+++ b/src/psconvert.c
@@ -2329,8 +2329,12 @@ EXTERN_MSC int GMT_psconvert (void *V_API, int mode, void *args) {
 		fclose (fpo);	fpo = NULL;
 		fclose (fp);	fp  = NULL;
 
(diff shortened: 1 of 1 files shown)


seisman:
@PaulWessel I just installed gs 9.52. It seems this PR doesn't work.

I tried 
```
gmt basemap -R0/10/0/10 -JX10c -Baf -B+gblue@90 -pdf map
gmt basemap -R0/10/0/10 -JX10c -Baf -B+gblue -t40 -pdf map
```

I don't see the warning. The variable **transparency** is false so the if block is not executed.

PaulWessel:
try png

seisman:
Yes, I can the warning for PNG and JPG.
```
psconvert [WARNING]: The file /Users/seisman/.gmt/sessions/gmt6.1084/gmt_0.ps- 
specifies transparency but your gs version 9.52 has a bug preventing it - 
please downgrade to 9.50
```
but the file name makes no sense to users.

PR #3291, opened 2020-05-09
Adding new grdmix module

PaulWessel:

Experimental module to blend two grids or images using weights from a third, or to just add transparency to an existing single image. The immediate goal is to get this to work:

`gmt grdmix @BlueMarble_06m.tif @BlackMarble_06m.tif -A@alpha.png -Gnewmap.png`

but it is still screwing.  At least 2 hours from scratch got me pretty far.  I added the dummy alpha.png to the cache for now - it will be removed once this all works.  I expect to add all the usual resolutions for those images though up to 30s.


seisman:
@PaulWessel I think it's a good opportunity to document what files should be added/updated when a new module is added.

PaulWessel:
Yes, with the new cmake glue there was no need to touch CMakeFiles anymore.  I have not touched the docs yet (other than adding grdmix.rst itself).  But for the C side it seems no need - all is auto. Docs still need to be added into the website.
Still finding lots of bugs though in img stuff.  I an reading a all-read 50x50 PNG which is stored as RGB 3 layers.  However, when GMT_Read_Data returns, it says all the right thing but only the first 834 triplets are 255/0/0.  SHould be 2500 triplets (50x50x3), so JL's GDAL reader is somehow dividing by 3 somewhere...  So hard to test grdmix which I think is otherwise correct.  I will implement import_ppm to go with export_ppm so I can test grdmix without relying on GDAL.

PaulWessel:
Working strictly with PPM files I can show that this works as expected:

gmt grdmix BlueMarble_06m.ppm @BlackMarble_06m.ppm -A@alpha.ppm -Gnewmap.ppm

Converting the latter to png yields this:

mix

PaulWessel:
After using Change_Layout everything runs fine but I get this image:

newmap

So supposedly that means bug in lines 12534-38.  Can you see anything obvious, @joa-quim ?

joa-quim:
Don't see any error. Could it be on writing?
Tomorrow I'll revisit the I->alpha case. I think I made it separate because of Matalab where the alpha band goes into a separate channel and like that we avoid making copies to extract the 4rth band, but for pixel interleaved it's silly to have separate. It becomes useless.

PaulWessel:
OK, will work a bit more on this and let you know where I left off.

PaulWessel:
This PR is now working as expected. Before removing WIP and merging there are some questions

1. I think the name **grdmix** is a bit to specific. Right now the module only does two things: Add a transparency alpha band (from a gray image, a grid, or a constant) to an image, or blend two images (or grids) using a third gray image or grid or constant.  Only the latter operation could be called a "mix".  As we are likely to add new operations to this module the name may become even less fitting.  Since the blending of grids only can be done in **grdmath** with the new **BLEND** operator, this module should probably focus more on things that pertain to images (fully or partly).
2. A module that works mostly on images probably should not be called grd*.  But it could, since it may still read and work on grids in addition to images.  I suggest grdrasterize, grdimager (too close to grdimage), grdprocess, grdimgtool, imagemath, imgoperator, ???.
3. If you have suggestions for operations that involves images or images and grids, let me know.  If not we can start with the current two operations.  It is easy to add simple things but many of them are already in gdal of course.  For instance, we could have a mode that would take an image and write separate GMT grids for red, green, blue (or gray) and alpha.  You can then do whatever with these in grdmath, and then perhaps we have a mode that takes 1 or 3 grid plus optional alpha grid and create an image again.  I think this could be two useful operations.  THoughts?

PaulWessel:
Regardless of the name, these commands seems useful to me:

```
gmt grdmix myimage.tiff -D -Glayers_%c.grd
gmt grdmix red.grd green.grd blue.grd -Atrans.grd -C -Gnewimage.png
```

The first command **D**econstructs an image into 1-4 grids (1 or 3 if gray or color, plus 1 if alpha present).  The second command takes 1 (gray) or 3 (r,g,b) grids, plus optional alpha to **C**onstruct an image.

As for name, I don't like the **img** prefix since we have **img2grd** which means something completely different.  So back to grd* again, maybe **grdimgtool**?

PaulWessel:
Another useful operation:

`gmt grdmix myimage.tiff -Iintensity -Gnewimage.png`

Brighten/darken the image accordinate to the intensity value or grid [0 = no change].

PaulWessel:
So I have added **-C -D -I** this morning.  Here is the result of

`gmt grdmix BlueMarble_06m.tif -Iintens.grd -Gfake.jpg`

where intens.grd was derived from earth_relief_06m:

fake

Sorry, a bit hard to see the shading as GitHub shrink the jpg, but it is working fine.

PaulWessel:
Likewise, to blend the two Marbles and add intensity from topo relief:

`gmt grdmix @BlackMarble_06m.tif @BlueMarble_06m.tif -Atrams.nc -Iintens.grd -Gnewmap.jpg`

newmap

This alpha-blends two separate images, and revises the final colors based on intensities.  OK with calling this grdimgtool, @joa-quim and @seisman ?

PaulWessel:
Off on mothers-day hike with Jill, back in late pm.

joa-quim:
Very nice images.
For the module's name, what about just ``grdimg``? The ``tool`` doesn't really add any info to the name.

src/grdmix.c
```diff
@@ -0,0 +1,526 @@
...
+					rgb[1] = rgb[2] = rgb[0];
+				gmt_illuminate (GMT, Gin[INTENS]->data[node], rgb);
+				for (band = 0, pix = node; band < h[0]->n_bands; band++, pix += h[0]->size)	/* March across the RGB values */
+					I->data[pix] = gmt_M_u255 (rgb[band]);
+			}
+		}
+	}
+	else {	/* Blend two grids or images using the weights given via -A */
+		if (Ctrl->In.type[0] == GMT_NOTSET) {	/* Two grids, just do the blend loop */
+			gmt_M_grd_loop (GMT, G, row, col, node) {
```
joa-quim, line 482: Good for OMP here

src/grdmix.c
```diff
@@ -0,0 +1,526 @@
...
+				Return (GMT_RUNTIME_ERROR);
+			}
+			for (node = 0; node < h[0]->size; node++)	/* Scale to 0-255 range */
+				I->alpha[node] = gmt_M_u255 (weights[node]);
+		}
+		if (Ctrl->I.active) {	/* Modify colors by the intensity */
+			gmt_M_grd_loop (GMT, I, row, col, node) {	/* The node is one per pixel in a band, so stride into additional bands */
+				for (band = 0; band < h[0]->n_bands; band++)	/* Get the normalized r,g,b triplet */
+					rgb[band] = gmt_M_is255 (I->data[node+band*h[0]->size]);
+				if (h[0]->n_bands == 1)	/* Duplicate grays so illuminate can work */
```
joa-quim, line 472: Maybe unroll this into two loops to avoid the ``if`` and OMP'them?

PaulWessel:
Thanks, will do the OPM on that.  grdimg is awfully close to grdimage.  I think that is likely to cause trouble.  Yes, I don't really like the 'tool" part, and in that sense grdmix is not so bad as we are mixing grids and images here.  Will keep it until a better candidate name is proposed or thought of.

PaulWessel:
Second thought.  Do we really save much time when the calculations we are doing are trivial.  It is not like we are doing a convolution like in filtering.  I worry that just setting up the OPM threads and duplicating private variables take longer than just running.  Let us just keep in this way until it stabilizes for now.

joa-quim:
For those loops what is real faster is the SIMD vectorization but for that there can be IF branches inside the loops.

And yes, *grdimg* is not good. Another possibility, since it deals with grids and images, *rasterblend*

seisman:
I got a crash:
```
$ gmt grdmix @BlueMarble_06m.tif -I0.5 -Gnewimage.png   
0   libgmt.6.dylib                      0x00000001087888b9 GMT_grdmix + 9417
1   ???                                 0x0000000000000008 0x0 + 8
Stack backtrace:
0   libgmt.6.dylib                      0x00000001083de3d2 sig_handler + 498
1   libsystem_platform.dylib            0x00007fff71b0d5fd _sigtramp + 29
2   ???                                 0x000000070bbb5a00 0x0 + 30261598720
3   libgmt.6.dylib                      0x000000010840ae81 GMT_Call_Module + 1073
4   gmt                                 0x00000001083cdecc main + 1228
5   libdyld.dylib                       0x00007fff71914cc9 start + 1
```

doc/rst/source/grdmix.rst
```diff
@@ -0,0 +1,141 @@
...
+.. include:: explain_-V.rst_
+
+.. |Add_-f| unicode:: 0x20 .. just an invisible code
+.. include:: explain_-f.rst_
+
+.. include:: explain_help.rst_
+
+.. include:: explain_grd_inout_short.rst_
+
+Examples
```
seisman, line 108: Why not just use the examples of BlueMarble grids?

seisman:
Am I doing something wrong below?
```
$ gmt grdimage @BlueMarble_06m.tif -JH120/15c -B0 -pdf map
$ gmt grdmix @BlueMarble_06m.tif -A0.5 -Gnewimage.tif
$ gmt grdimage newimage.tif -JH120/15c -B0 -pdf map
grdimage [WARNING]: Round-off patrol found north latitude outside valid range (1800)!
grdimage [ERROR]: Internal Failure = GMT_MAP_BAD_LAT_MIN
grdimage (grdimage.c:810(GMT_grdimage)): South is outside -90 to +90 degree range
```

PaulWessel:
Not really.  I need help from @joa-quim to make sure the geotiffs are written correctly.  RIght now they are just dump TIFFs with auto range of pixel dims.

PaulWessel:
The input TIFF says (per gdalinfo)

```
pwessel@macnut:~/test-> gdalinfo BlueMarble_06m.tif
Driver: GTiff/GeoTIFF
Files: BlueMarble_06m.tif
Size is 3600, 1800
Origin = (-180.000000000000000,90.000000000000000)
Pixel Size = (0.100000000000000,-0.100000000000000)
Metadata:
  TIFFTAG_DATETIME=2019:07:19 06:59:13
  TIFFTAG_RESOLUTIONUNIT=2 (pixels/inch)
  TIFFTAG_SOFTWARE=Adobe Photoshop CS6 (Macintosh)
  TIFFTAG_XRESOLUTION=72
  TIFFTAG_YRESOLUTION=72
Image Structure Metadata:
  COMPRESSION=DEFLATE
  INTERLEAVE=PIXEL
Corner Coordinates:
Upper Left  (-180.0000000,  90.0000000) 
Lower Left  (-180.0000000, -90.0000000) 
Upper Right ( 180.0000000,  90.0000000) 
Lower Right ( 180.0000000, -90.0000000) 
Center      (   0.0000000,   0.0000000) 
```

But the output TIFF says

```
pwessel@macnut:~/test-> gdalinfo newimage.tif 
Driver: GTiff/GeoTIFF
Files: newimage.tif
Size is 3600, 1800
Origin = (1.000000000000000,1800.000000000000000)
Pixel Size = (0.100000000000000,-0.100000000000000)
Image Structure Metadata:
  COMPRESSION=DEFLATE
  INTERLEAVE=PIXEL
Corner Coordinates:
Upper Left  (       1.000,    1800.000) 
Lower Left  (       1.000,    1620.000) 
Upper Right (     361.000,    1800.000) 
Lower Right (     361.000,    1620.000) 
Center      (     181.000,    1710.000) 

```

The image header structure is unchanged.  Need @joa-quim to advise on this.  The coordinate info is disturbing.

doc/rst/source/grdmix.rst
```diff
@@ -0,0 +1,147 @@
...
+*raster?*
+    If only one is given and **-C** is not set then *raster1* must be an image.
+    If two are given then *raster1* and *raster2* must both be either
+    images or grids.  If three are given then they must all be grids and
+    **-C** must be set.
+
+.. _-G:
+
+**-G**\ *outfile*
+    The name for the output raster.
```
seisman, line 56: Document the supported image formats.

joa-quim:
It's not only the info. It kills Mirone when I try to open it.
Regarding the coordinates, the ProjRefProj4 or ProjRefWKT info from the original geotif file was not carried on to output, but suspect this is not enough to explain the Mirone killing.

PaulWessel:
The PRojRefPRoj4 etc are all NULL after reading the geotiff.

joa-quim:
Hmm, that should not happen I think. Remember those annoying warnings we had from GDAL in tests when writing geotiff without CRS? The thing is that we don't specify a referencing system it writes a plain TIF file. Try to set the output PRojRefPRoj4 to *longlat*

doc/rst/source/grdmix.rst
PaulWessel: Thanks, I guess use the same stuff as for grdimage -A, @joa-quim ? Or for this (we are not writing grids) maybe just list the standard image formats geotiff, png, jpg, ???

joa-quim: It should be able to read all GDAL formats. But for the *standard* we do some guessings in grdimage to avoid having to state de driver name, I think.

PaulWessel:
Right, I will look at grdimage -A a bit more.  I did not do anything other than calling GMT_Write_Data.

PaulWessel:
I followed the steps in **grdproject** to add the proj4 longlat stuff and now the geotiff works as per @seisman's example.  However, I am relying on adding **-fg** to know it is a geographic image.  Surely the reading of the geotiff must set some flag somewhere that it is geographic so  I can use that instead?

seisman:
My example above gives me an error, although the tiff file seems correct and I can plot the tiff file.
```
gmt grdmix @BlueMarble_06m.tif -A0.5 -Gnewimage.tif -fg
grdmix [ERROR]: gmt_gdalwrite failed to convert the proj4 string
+proj=longlat +a=6378137.000000 +b6378137.000000 +no_defs
 to WKT
```

doc/rst/source/grdmix.rst
PaulWessel: OK, added one.

PaulWessel:
A few changes since earlier you should be aware of:

1. I move the blend weights to **-W** so that **-A** is only accepting alpha values.  That means we can do a weight blend **and** apply intensities **and** add alpha at the same time.
2. I added **-Q** for opaque: This removes any alpha layer in the final image.
3. I added **-R** to set a region for non-geographic regions.  Probably should add documentation that this only matters for geotiffs.
4. Need GMT_Read_Data to set the image header correctly if it detects a geotiff.
5. I will do the OMP stuff soon.  I can see it being useful for very large images.
6. Oh, added **-M** to make a monochrome image (as in grdimage).

How about grdmiximage as name?

PaulWessel:
> My example above gives me an error, although the tiff file seems correct and I can plot the tiff file.
> 
> ```
> gmt grdmix @BlueMarble_06m.tif -A0.5 -Gnewimage.tif -fg
> grdmix [ERROR]: gmt_gdalwrite failed to convert the proj4 string
> +proj=longlat +a=6378137.000000 +b6378137.000000 +no_defs
>  to WKT
> ```

gdal version? I have 3.0.4 and works fine here.

joa-quim:
It was a bug that I fixed directly on github. It was ``+b6378137.000000`` but should have been ``+b=6378137.000000``
However, the tiff file still kills Mirone but I can open it in QGis.

PaulWessel:
@joa-quim, how about calling gmt_set_geographic (GMT, GMT_IN); when you discoveres that you are reading a geotiff? I should not have to rely on a user adding -fg -Rd in grdmix, no?

joa-quim:
No, you must scan the input ProjRefProj4 for a "lonlat" or "latlong" string. If ProjRefProj4  does not exist (likely), then ProjRefWKT must be converted to Proj4 (we have bits of code who do that). If sill nothing we aere on our owns to guess. We can assume that a geotiff without RS is faulty and if BB is within [-80 360] [-9- 90] than its a geog file.
Maybe there other strategies to find if a geotiff is geog or not but I don't know them.

PaulWessel:
If you open teh BlueMargle_06m.tif in any tool, is there no information you see that tells it is a geotiff? I could pretty confidently use the w/e/s/n matching -Rd to say it is cleary geographic, but we may have smaller tiles and then do we always assume that any tif with non-zero west/south is geographic?

joa-quim:
That ``BlueMargle_06m.tif`` is faulty. It should always have a referencing system but it doesn't. It's for those cases that or we simply error and try to guess.

PaulWessel:
OK, so presumably we can run gdal_convert or similar and turn it into a proper geotiff before distributing?

PaulWessel:
Of course, the NASA site is extremely slow right now, trickle-download with time-outs, so unlikely I can download the 30s tiles anytime soon.

joa-quim:
Yes, gdal_translate can do it.

PaulWessel:
I have tested grdmix fairly extensively this morning, doing it option by option, and fixing issues as they arise. Some recent changes:

1. Better parsing of the **-A, -I -W** options and detect bad file names or missing files.
2. Avoid memory leaks upon errors and returns
3. If input image has no proj4 info, use its region to at least detect if it is a full global image (360/180) and then set geographic.  For other domains you can either use **-fg** (treat image region as geo) or add **-R** (give image geo region)
4. If product is geographic and output image is not TIF, I warn of loss of geo metadata.
5. Let **-N** take an optional factor [255], and let **-N** also control **-D** grid range.

I have slept on it, and I think **grdmix** is a pretty good name.  **grdmiximage** is slightly better but I know @joa-quim is not fond of having to type those extra 5 characters.  So unless there are objections I will stick with **grdmix** and ask for this PR to be accepted.  I will work on an example 52 for this as well, and another full-length HD movie.

seisman:
@PaulWessel It fails on Linux and Windows.

PaulWessel:
So for Windows I expected taht since my loop variables are unsigned and windows support for OPM is so old they insist signed ints.  I will fix that.
The Linux errors (I assume this is with OPenMP) are the same mysterious ones as before.

PaulWessel:
I have addressed the OpenMP issues and I have built and tested the MP enabled grdmix on the same text cases as before, and all work (faster).
I will add ex52 now before approvoal and merging.

seisman:
I'm getting more warnings:
```
../src/geodesy/gpsgridder.c:744:3: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]
                gmt_M_grd_loop (GMT, Out[GMT_X], row, col, k) if (gmt_M_is_fnan (Out[GMT_X]->data[k])) n_ok--;
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/gmt_grd.h:165:40: note: expanded from macro 'gmt_M_grd_loop'
#define gmt_M_grd_loop(C,G,row,col,ij) gmt_M_row_loop(C,G,row) gmt_M_col_loop(C,G,row,col,ij)
                                       ^~~~~~~~~~~~~~~~~~~~~~~
../src/gmt_grd.h:163:51: note: expanded from macro 'gmt_M_row_loop'
#define gmt_M_row_loop(C,G,row) for (row = 0; row < (int)G->header->n_rows; row++)
                                              ~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~
../src/geodesy/gpsgridder.c:744:3: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]
                gmt_M_grd_loop (GMT, Out[GMT_X], row, col, k) if (gmt_M_is_fnan (Out[GMT_X]->data[k])) n_ok--;
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/gmt_grd.h:165:64: note: expanded from macro 'gmt_M_grd_loop'
#define gmt_M_grd_loop(C,G,row,col,ij) gmt_M_row_loop(C,G,row) gmt_M_col_loop(C,G,row,col,ij)
                                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/gmt_grd.h:164:94: note: expanded from macro 'gmt_M_col_loop'
#define gmt_M_col_loop(C,G,row,col,ij) for (col = 0, ij = gmt_M_ijp (G->header, row, 0); col < (int)G->header->n_columns; col++, ij++)
```

PaulWessel:
Well, that was what the (int) casts were for, but Linux openMP don't like those "predicates" and Windows don't like unsigned counters.  The world is mad.  We cannot have no warnings with and without OMP it seems.

PaulWessel:
Hi @joa-quim and @seisman, do you have any clever thoughts on how to deal with this mess? I can complain about lame OpenMP implementations but that does not make these warnings go away.  I do not want to edit the entirety of GMT and remove unsigned integers.

doc/rst/source/grdmix.rst
```diff
@@ -0,0 +1,177 @@
...
+Description
+-----------
+
+**grdmix** will perform various operations involving images and grids.
+We either use a *alpha* grid, image, or constant to add a new alpha
+(transparency) layer to the image given as *raster1*, or we will blend
+the two *raster1* and *raster2* (grids or images) using the *weights* for
+*raster1* and the complementary *1 - weights* for *raster2* and save to
+*outfile*. Alternatively, we will deconstruct an image into its component
+grid layers or we construct an image from its normalized component grids.
```
seisman, line 42: The meaning of "component grid layers" is not clear. Perhaps adding "RGB" is better?
```suggestion
*outfile*. Alternatively, we will deconstruct an image into its RGB component
grid layers or we construct an image from its normalized RGB component grids.
```

doc/rst/source/grdmix.rst
```diff
@@ -0,0 +1,177 @@
...
+
+To blend the night and day views of the Earth using a weight image computed for
+a particular day/night terminus, try::
+
+    gmt grdmix @BlueMarble_06m.tif @BlackMarble_06m.tif -W@weight.png -Gnewmap.png
+
+Suppose map1.png and map2.png are overlapping maps of different quantities, but we wish
+to use the image visible.png to blend them into a single image.  We try::
+
+    gmt grdmix map1.png map2.png -Avisible.png -Gnewmap.png -V
```
seisman, line 152: I'm very confused by this example. How are the two images blended? How does the **-A** option control the blend?

I'm getting some segment faults with following commands:
```
gmt grdmix @BlueMarble_06m.tif @BlackMarble_06m.tif -Gabc.png
gmt grdmix @BlueMarble_06m.tif @BlackMarble_06m.tif -A@weight.png -Gabc.png
```

doc/rst/source/grdmix.rst
```diff
@@ -0,0 +1,177 @@
...
+
+To insert the values from the grid transparency.grd into the image gravity.tif as an alpha
+(transparency) layer, and write out a transparent PNG image, try::
+
+    gmt grdmix gravity.tif -Atransparency.grd -Gmap.png
+
+To break the color image layers.png into separate, normalized red, green blue grids (and possibly an alpha grid),
```
seisman, line 159:
```suggestion
To break the color image layers.png into separate, normalized red, green, and blue grids (and possibly an alpha grid),
```

doc/rst/source/grdmix.rst
```diff
@@ -0,0 +1,177 @@
...
+we run::
+
+    gmt grdmix layers.png -Glayer_%c.grd
```
seisman, line 162: **-D** option is missing?
```suggestion
    gmt grdmix layers.png -D -Glayer_%c.grd
```

doc/rst/source/grdmix.rst
```diff
@@ -0,0 +1,177 @@
...
+
+To recombine the three normalized grids red.grd, green.grd, and blue.grd into a TIFF file, but
+applying intensities from intens.grd and add transprency from transp.grd grids, try::
+
+    gmt grdmix red.grd green.grd and blue.grd -Glayer.png -Atransp.grd -Iintens.grd
```
seisman, line 167: The description says "into a TIFF file". Does the following command work? Perhaps need -R?
```suggestion
    gmt grdmix red.grd green.grd and blue.grd -Glayer.tif -Atransp.grd -Iintens.grd
```

doc/rst/source/modules.rst
```diff
@@ -79,6 +79,7 @@ All modules are requested via a call to the :doc:`gmt` program.
    grdlandmask
    grdmask
    grdmath
+   grdmix
```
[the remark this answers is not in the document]
PaulWessel: Done

doc/rst/source/grdmix.rst
PaulWessel: Done

doc/rst/source/grdmix.rst
PaulWessel: OK, changed to tiff.  Hypothetical example so we assume the grids are geographic.

doc/examples/ex52/ex52.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
PaulWessel: OK, will try.

doc/rst/source/grdmix.rst
PaulWessel: I have updated the docs to make it clear.  Will look at the SEGV now.

PaulWessel: OK, added more parsing checks to prevent your SEGV user error.

seisman: It's much clear now. Please also update the **-A** description, just like what **-W** says:
```
    When two input rasters are given, the weights are applied to *raster1* and
    (*1-weights*) are applied to *raster2*, then the products are summed.
```

PaulWessel: Note sure what you expect.  -A simply adds a layer that we add to the image.  There is no blending etc (my first grdmix version used -A for what -W is doing now, so that -A is just usd to add transparency).  I think it is clear?

doc/rst/source/grdmix.rst
```diff
@@ -0,0 +1,181 @@
...
+
+    gmt grdmix @BlueMarble_06m.tif @BlackMarble_06m.tif -W@weight.png -Gnewmap.png
+
+Suppose map1.png and map2.png are overlapping maps of different quantities, but we wish
+to use the image visible.png to blend them into a single image: Where visible.png has
+values close to 255 we will see predominantly the map1.png contents while for values
+closer to zero we will mostly see map2.png - values in between these extremes will
+lead to a weighted average.  We try::
+
+    gmt grdmix map1.png map2.png -Avisible.png -Gnewmap.png -V
```
seisman, line 156: Is this example still valid? Are you talking about **-W** or **-A**?

PaulWessel: Sorry, you are right, this is now -Q.  Fixed, plus a typo further down,

doc/examples/ex52/ex52.sh
```diff
@@ -0,0 +1,26 @@
...
+# GMT EXAMPLE 52
+#
+# Purpose:     Mixing images and overlaying on the Earth
+# GMT modules: grdgradient, grdimage, grdmath, grdmix, grdsample, solar
+
+gmt begin ex52 ps
```
seisman, line 7:
```suggestion
gmt begin ex52
```

doc/examples/ex52/ex52.sh
```diff
@@ -0,0 +1,26 @@
...
+
+  # Use the location of the Sun at noon on Jun 22, 2000, Hawaii Standard Time
+  # Make a global 6m grid with a smooth transition across day/night boundary.
+  # We take advantage of the ATAN function to relatively quickly ramp from 0 to 1
+  gmt grdmath -Rd -I20m -r --PROJ_ELLIPSOID=sphere $(gmt solar -C -o0:1 -I+d2000-06-22T24:00+z-10) SDIST KM2DEG 90 SUB NEG 2 DIV ATAN DUP UPPER DIV 1 ADD 2 DIV = w.grd
```
seisman, line 12: Define a variable `res=20m` and list the allowed values, so that users can easily try high-resolution data.

doc/examples/ex52/ex52.sh
```diff
@@ -0,0 +1,26 @@
...
+  gmt grdsample intens.grd -T -Gintens.grd
+
+  # Blend the Blue and Black Marble geotiffs using the weights, so that when w is 1 we get the Blue marble
+  # and adjust colors based on the intensity.
+
+  gmt grdmix @BlueMarble_20m.tif @BlackMarble_20m.tif -Ww.grd -Iintens.grd -Gnewmap.tif
+
+  # Plot this image on an Earth with view from over Mexico
+  gmt grdimage newmap.tif -JG80W/30N/18c -Bafg
+gmt end show
```
seisman, line 26: Clean up the intermediate files at the end.

doc/examples/ex52/ex52.sh
PaulWessel: Sorry, it snuck back in to make the PS orig.

doc/examples/ex52/ex52.sh
PaulWessel: Done

PaulWessel:
Will change grdmath call once #3319 is merged.

doc/examples/ex52/ex52.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+# GMT EXAMPLE 52
+#
+# Purpose:     Mixing images and overlaying them on the Earth
+# GMT modules: grdgradient, grdimage, grdmath, grdmix, grdsample, solar
+
+res=20m # Change to 15m, 10m, 06m, 05m, 04m, 03m, 02m, 01m, 30s for higher resolugions
```
seisman, line 7:
```suggestion
res=20m # Change to 15m, 10m, 06m, 05m, 04m, 03m, 02m, 01m, 30s for higher resolutions
```

doc/examples/ex52/ex52.sh
```diff
@@ -0,0 +1,30 @@
...
+# Purpose:     Mixing images and overlaying them on the Earth
+# GMT modules: grdgradient, grdimage, grdmath, grdmix, grdsample, solar
+
+res=20m # Change to 15m, 10m, 06m, 05m, 04m, 03m, 02m, 01m, 30s for higher resolutions
+
+gmt begin ex52
+
+  # Use the location of the Sun at noon on Jun 22, 2000, Hawaii Standard Time
+  # Make a global grid with a smooth 2-degree transition across day/night boundary.
+  # We take advantage of the ATAN function to relatively quickly ramp from 0 to 1
```
seisman, line 13: The ATAN function is used in the DAYNIGHT operator. No need to mention it here.
```suggestion
```

doc/examples/ex52/ex52.sh
```diff
@@ -0,0 +1,30 @@
...
+  gmt grdsample intens.grd -T -Gintens.grd
+
+  # Blend the Blue and Black Marble geotiffs using the weights, so that when w is 1
+  # we get the Blue marble, and then adjust colors based on the intensity.
+
+  gmt grdmix @BlueMarble_${res}.tif @BlackMarble_${res}.tif -Ww.grd -Iintens.grd -Gview.tif
+
+  # Plot this image on an Earth with view from over Mexico
+  gmt grdimage view.tif -JG80W/30N/18c -Bafg
+  # Clean up after us
```
seisman, line 28:
```suggestion
  # Clean up after use
```

seisman:
@PaulWessel These warnings are not directly related to OpenMP. I see a lot of warnings like this when OpenMP is disabled on macOS.
```
../src/geodesy/gpsgridder.c:744:3: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]
                gmt_M_grd_loop (GMT, Out[GMT_X], row, col, k) if (gmt_M_is_fnan (Out[GMT_X]->data[k])) n_ok--;
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/gmt_grd.h:165:40: note: expanded from macro 'gmt_M_grd_loop'
#define gmt_M_grd_loop(C,G,row,col,ij) gmt_M_row_loop(C,G,row) gmt_M_col_loop(C,G,row,col,ij)
                                       ^~~~~~~~~~~~~~~~~~~~~~~
../src/gmt_grd.h:163:51: note: expanded from macro 'gmt_M_row_loop'
#define gmt_M_row_loop(C,G,row) for (row = 0; row < (int)G->header->n_rows; row++)
                                              ~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~
../src/geodesy/gpsgridder.c:744:3: warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]
                gmt_M_grd_loop (GMT, Out[GMT_X], row, col, k) if (gmt_M_is_fnan (Out[GMT_X]->data[k])) n_ok--;
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/gmt_grd.h:165:64: note: expanded from macro 'gmt_M_grd_loop'
#define gmt_M_grd_loop(C,G,row,col,ij) gmt_M_row_loop(C,G,row) gmt_M_col_loop(C,G,row,col,ij)
                                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/gmt_grd.h:164:94: note: expanded from macro 'gmt_M_col_loop'
#define gmt_M_col_loop(C,G,row,col,ij) for (col = 0, ij = gmt_M_ijp (G->header, row, 0); col < (int)G->header->n_columns; col++, ij++)
```

PaulWessel:
They **are** related to OpenMP because Windows OpenMP **forces** us to switch to signed loop variables where we wish to use OpenMP.  Since our GMT_* struct  variables for positively definite quantities are unsigned (e.g., n_columns, n_rows), we get these stupid messages.  We cannot change n_columns to be signed since it is an API definition.  We suddenly cannot add casts (int) to those macros because Linux now suddenly has a hissy-fit.  So not sure what we can do but swallow the warnings...

joa-quim:
VS has #pragma macros to shut up specific warnings. Don't gcc & llvm have that?

PaulWessel:
Yes we can do -Wno-*** I think, but we then basically loose the ability of the compiler to warn us.  Can't you step on Bill Gate's balls and have him allow unsigned loop variables like the rest of the world?

PaulWessel:
I mean, we could do things like

```
#ifdef WIN32
  int64_t row, col;
#else 
  uint64_t row, col;
#endif 
```

and then we may only get those warnings on Windows, which seems fair since it is Bill's fault.

seisman:
> We suddenly cannot add casts (int) to those macros because Linux now suddenly has a hissy-fit.

It puzzles me a lot, since we are doing the (int) casts in the master branch and everything works fine.

PaulWessel:
yes, but the OpenMP pragmas creates new C code that is then compiled (all under the hood) and that is probably where the warnings are coming from.  Not sure though.

joa-quim:
Blame the OMPeers. Were them who imposed the signed ints at that version. And you know I never liked those *unsigned* that keep biting us with this.

src/grdmix.c
```diff
@@ -0,0 +1,707 @@
...
+		uint64_t dim[3] = {0,0,3};
+		GMT_Report (API, GMT_MSG_INFORMATION, "Construct image from %d component grid layers\n", Ctrl->In.n_in);
+		if ((I = GMT_Create_Data (API, GMT_IS_IMAGE, GMT_IS_SURFACE, GMT_CONTAINER_AND_DATA, dim, G_in[0]->header->wesn, G_in[0]->header->inc, GMT_GRID_PIXEL_REG, 0, NULL)) == NULL) {
+			GMT_Report (API, GMT_MSG_ERROR, "Unable to duplicate an image for output!\n");
+			Return (GMT_RUNTIME_ERROR);
+		}
+		H = I->header;
+		if (Ctrl->I.active && Ctrl->In.n_in == 3) {	/* Make the most work-intensive version under OpenMP */
+#ifdef _OPENMP
+#pragma omp parallel for private(row,col,node,band,rgb,pix) shared(GMT,I,G_in,H)
```
seisman, line 506: intens is also a shared variable, right?
```suggestion
#pragma omp parallel for private(row,col,node,band,rgb,pix) shared(GMT,I,G_in,H,intens)
```

PaulWessel: Yes, fixed.

PaulWessel:
I think this one is now ready to be merged, guys?

PR #3319, opened 2020-05-12
Add operator DAYNIGHT to grdmath

PaulWessel:

This operator takes a _longitude_ and _latitude_ position (assumed to be the sun) and a transition _width_ and computes a day/night mask grid that is one on the side facing the sun and zero on the side facing away.  If the transition width is zero then we get an exact 0/1 mask, but if the width is finite then we approximate the 0/1 Heaviside step function by a scaled _atan_ function and depending on the transition width the values will be > 0 and < 1.  A new test (daynight.sh) has been added showing the two types of results.  The Sun is specified at (30, 20) with transition widths of 0 and 5 degrees.

daynight

### DIFF
--- a/src/grdmath.c
+++ b/src/grdmath.c
@@ -1858,8 +1858,20 @@ GMT_LOCAL void grdmath_DAYNIGHT (struct GMT_CTRL *GMT, struct GRDMATH_INFO *info
 			}
 		}
 	}
-	else {
-		if (stack[last]->constant) iw = 1.0 / stack[last]->factor;	/* To avoid division */
+	else if (stack[last]->constant && !gmt_M_is_zero (stack[last]->factor)) {	/* Fixed nonzero transition width */
+		iw = 1.0 / stack[last]->factor;	/* To avoid division */
+#ifdef _OPENMP
+#pragma omp parallel for private(row,col,node,d) shared(info,stack,prev2,last,GMT,x0,y0,iw)
+#endif
+		for (row = 0; row < info->G->header->my; row++) {
+			node = row * info->G->header->mx;
+			for (col = 0; col < info->G->header->mx; col++, node++) {
+				d = gmt_distance (GMT, x0, y0, info->d_grd_x[col], info->d_grd_y[row]);	/* Distance in degrees from (A,B) */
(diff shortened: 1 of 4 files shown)


PaulWessel:
Upon merging of this PR I will revise ex52.sh and several other scripts using convoluted grdmath trigonometry to build fake shade grids.

src/grdmath.c
```diff
@@ -1828,6 +1829,51 @@ GMT_LOCAL void grdmath_D2R (struct GMT_CTRL *GMT, struct GRDMATH_INFO *info, str
...
+			for (col = 0; col < info->G->header->mx; col++, node++) {
+				d = gmt_distance (GMT, x0, y0, info->d_grd_x[col], info->d_grd_y[row]);	/* Distance in degrees from (A,B) */
+				stack[prev2]->G->data[node] = (d > 90.0) ? 0.0f : 1.0f;
+			}
+		}
+	}
+	else {
+		if (stack[last]->constant) iw = 1.0 / stack[last]->factor;	/* To avoid division */
+#ifdef _OPENMP
+#pragma omp parallel for private(row,col,node,d,w) shared(info,stack,prev2,last,GMT,x0,y0)
```
seisman, line 1864: Typo?
```suggestion
#pragma omp parallel for private(row,col,node,d,iw) shared(info,stack,prev2,last,GMT,x0,y0)
```

joa-quim:
I still would like to see this in *solar*. For example, what if one wants to make a realistic transition zone using the twilight terminators?

src/grdmath.c
PaulWessel: Thanks, fixed.

PaulWessel:
That is fine - I jus don't know how to do it.  Feel free to work on that if you think it should go in solar.

PaulWessel:
OK, now adjusted to work fine under OpenMP as well.

joa-quim:
In the middle of other things I'll keep thinking in a clever algorithm to do that. Or dumb algo if clever does not come.

PaulWessel:
> In the middle of other things I'll keep thinking in a clever algorithm to do that. Or dumb algo if clever does not come.

That is how we roll.

PaulWessel:
Like to get this approved so I can update many examples to make them much simpler, including the grdmix branch and example ex52.

Issue #3322, opened 2020-05-12
GMT module for assistance with batch jobs


PaulWessel:
**Description of the desired feature**

The building of an animation sequence is the classic case of a large set of batch jobs.  We may need to make hundreds or thousands of frames, each a single GMT plot converted to a PNG image.  This task has been greatly simplified by the introduction of the **movie** module which takes care of all the ugliness associated with running such batch jobs (using all cores, make sure things don't step on each other, preserving output, passing variables, etc.).

However, lots of time-consuming processes are not creating plots at all.  Think of all types of data processing operations, such as filtering, stacking, gridding, etc. ,etc.  If you need to filter 1000 grids, or you want to try 1000 different filter widths on a single grid, particular those non-convolution operations like mode and median, you are looking at writing scripts that loop over variables and issue commands.  Ah, but then you want to take advantage of all your cores, or perhaps even spread jobs across several computers.  Setting that up becomes quite involved.

As a solution to this need (and of course I have these needs) I propose to develop a _new module_ called gmt **batch**.  It takes lessons from gmt **movie** but rips out all the plot/rasterization/movie-building aspects of it, leaving just the batch machinery.  The synopsis would be similar to **movie** but much reduced:

**gmt batch** _mainscript_ **-T**_nframes_|_min_/_max_/_inc_[**+n**]|_timefile_[**+p**_width_][**+s**_first_][**+w**] [**-I**_includefile_] [**-Sb**_prescript_] [**-Sf**postscript] [**-V**[_level_]] [**-W**[_workdir_]] [**-Z**[**s**]] [**-f**[**i**|**o**]_info_] [**-x**[[-]_n_]] [**--PAR**=_value_]

The specifics:

1. _mainscript_: The bash, csh, or DOS batch script that is the core of the repeated processing job.  It gets fixed variables from the optional _includefile_, as well as all the parameters similar to the **MOVIE_*** ones but in this module they are called **BATCH_*** or similar.
2. The **-T** option either gives a data table with the required parameters, filenames, etc., or if things are simple enough just a number sequence (e.g., **-T**100 /1000/20 to try all those filter widths from 100 to 1000 in steps of 20).
[...]

PR #3326, opened 2020-05-13
Adding a batch processing module

PaulWessel:

Please see #3322 for context.  This PR implements that solution.  Here is an example that demonstrates the utility of the module: We want to filter topography for a wide range of filter widths, then stack all the results and determine the standard deviations at each point, and make a map that shows which areas are most sensitive to the choice of filter width.  Script:

```
#!/bin/bash
# Testing gmt batch by filtering data then examining the std of the stack
cat << EOF > pre.sh
gmt begin # Make list of filterwidths and extract a DEM subset
	gmt math -o0 -T10/200/10 T = widths.txt
	gmt grdcut -R-10/20/-10/20 @earth_relief_02m -Gdata.grd
gmt end
EOF
cat << EOF > main.sh
gmt begin # Highpass the grid using the given filter width and save the output
	gmt grdfilter data.grd -Fg\${BATCH_COL0}+h -G\${BATCH_NAME}.grd -D2
gmt end
EOF
cat << EOF > post.sh
# Stack the resulting grids, find the std, and make a plot
gmt begin \${BATCH_PREFIX} png
	gmt grdmath \${BATCH_PREFIX}_*.grd -S STD = \${BATCH_PREFIX}_std.grd
	gmt grdimage \${BATCH_PREFIX}_std.grd -B -B+t"STD of Gaussians residuals" -Chot
	gmt coast -Wthin,white
gmt end show
EOF
gmt batch main.sh -Sbpre.sh -Sfpost.sh -Twidths.txt -Nfilter -V -W -Z
```

After 20 jobs running in parallel, the postflight script completes the tasks and makes this plot:

filter


src/batch.c
```diff
@@ -0,0 +1,1154 @@
...
+	/* Waiting before checking if the completion file has been generated */
+#ifdef WIN32
+	Sleep ((uint32_t)microsec/1000);	/* msec are microseconds but Sleep wants milliseconds */
+	return 0;
+#else
+	return (usleep ((useconds_t)microsec));
+#endif
+}
+
+GMT_LOCAL void batch_set_value (struct GMT_CTRL *GMT, FILE *fp, int mode, int col, char *name, double value) {
```
seisman, line 159: These functions are the same as those in movie.c, right? Can we reuse these functions?

PaulWessel: Yes, I thought about that as I was doing my copy/cut.  Will check.

PaulWessel: In trying not to duplicate code (always a good thing), here are some options, given that I would like to follow our static function naming convention as well.  Let's look at the functions called

```
GMT_LOCAL int batch_gmt_sleep (unsigned int microsec)
GMT_LOCAL int movie_gmt_sleep (unsigned int microsec)

```
I would like one code base yielding two separate static functions with the above names.  I think these are viable options:

1. Make a gmt_script.c.in (or h.in?) which lists `GMT_LOCAL int @MODULE@_gmt_sleep (unsigned int microsec)` etc, and let CMake build both batch_sub.c and movie_sub.c from that.
2. Do some trickery with C preprocessors.  If I remember right, the #define statements is kind of like text replacement.  If so, might we set #define THIS movie in movie.c and have gmt_script.c be included and it will list G`MT_LOCAL int THIS_gmt_sleep (unsigned int microsec)`.  Will that work?
3. Just elevate these to gmtlib_* functions and place them in gmt_support.c and they can be used by both batch.c, movie.c, and possibly other future modules.

Perhaps there is no real savings and possibly confusion by **not** doing optino 3?

PaulWessel: Looks like the price to pay for keeping them static is too much, @seisman ?

seisman: Option 3 seems the easiest one to maintain.

PaulWessel: Great. Going with that.

PaulWessel: Done.

doc/rst/source/batch.rst
```diff
@@ -0,0 +1,273 @@
...
+Data Files
+----------
+
+The batch scripts will be able to find any files present in the starting directory when **batch** was initiated,
+as well as any new files produced by *mainscript* or the optional scripts set via **-S**.
+No path specification is needed to access these files.  Other files may
+require full paths unless their directories were already included in the :term:`DIR_DATA` setting.
+
+Plotting Temporal Changes
+-------------------------
```
seisman, line 177: Plotting? Should be changed to other words.

PaulWessel: Fixed. Cut/paste from movie.

src/CMakeLists.txt
```diff
@@ -350,7 +350,7 @@ set (GMT_PROGRAM ${GMT_SOURCE_DIR}/src/gmtprogram.c)
 set (GMT_COMPAT_MODULES minmax gmtstitch gmtdp grdreformat ps2raster originator PARENT_SCOPE)
 
 # gmt core modules
-set (GMT_PROGS_SRCS blockmean.c blockmedian.c blockmode.c docs.c dimfilter.c filter1d.c grd2kml.c
+set (GMT_PROGS_SRCS batch.c blockmean.c blockmedian.c blockmode.c docs.c dimfilter.c filter1d.c grd2kml.c
```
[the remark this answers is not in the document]
PaulWessel: Hm, mine is sorted?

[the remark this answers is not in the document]
seisman: I mean the rest C files in the list, for example:
```
gmtspatial.c gmtconnect.c grdgdal.c gmtregress.c gmtvector.c gmtwhich.c grd2cpt.c
```
are not sorted.

PR #3333, opened 2020-05-14
Simplify grdflexure, fix bugs and add transfer function output

PaulWessel:

Function renaming did some odd things here, plus K is not an argument that is used by the setup functions.  Adding a test mode under debug to just write the transfer functions to a file.

### DIFF
--- a/doc/rst/source/supplements/potential/grdflexure.rst
+++ b/doc/rst/source/supplements/potential/grdflexure.rst
@@ -244,7 +244,7 @@ Cathles, L. M., 1975, *The viscosity of the earth's mantle*, Princeton Universit
 
 Nakada, M., 1986, Holocene sea levels in oceanic islands: Implications for the rheological
 structure of the Earth's mantle, *Tectonophysics, 121*, 263–276,
-`http://dx.doi.org/10.1016/0040-1951(86)90047-8 <http://dx.doi.org/10.1016/0040-1951(86)90047-8>`_
+`http://dx.doi.org/10.1016/0040-1951(86)90047-8 <http://dx.doi.org/10.1016/0040-1951(86)90047-8>`_.
 
 Wessel. P., 2001, Global distribution of seamounts inferred from gridded Geosat/ERS-1 altimetry,
 J. Geophys. Res., 106(B9), 19,431-19,441,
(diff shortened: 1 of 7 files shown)


PaulWessel:
A bunch of things have been fixed or added in this branch:

1. The **-M** option for a viscoelastic solution had a bug where the Maxwell relaxation time given  was not passed to the rheology control structure during setup, so it was always zero.
2. The **-F**_nu_a/h_a/nu_m_ option for a two layer firmoviscous model correctly obtained the asthenospheric channel thickness but like **-M** it failed to copy that value into the rheology control structure. Being always zero, the two firmoviscous solutions were identical.
3. For very large wavenumbers, an expression given to _cosh_ and _sinh_ would yield infinity when computing tau, and this would propagate through and give a NaN on output.  It was better numerically to compute the reciprocal value for tau and check for NaNs, and then set the result to zero (which it should be for this case instead of trying to divide by infinity).

In debugging these problems I added a new firmoviscous2.sh script to test the two-layer rheology model, and I added a new test transfer.sh that does no flexure calculations but instead displays the transfer functions vs wavelength for 18 different cases.  This utility was added via the option **-G+d** which means to dump the transfer function selected for a range of elastic thicknesses and times.  In this mode, no input data are required (they are ignored, as is **-N**) , only the rheological options matter.

Finally, the two firmoviscous transfer functions are valid when the elastic thickness goes to zero, yielding viscous transfer functions for a half-space and a viscous layer over a half-space.  These two transfer functions are now accessible by using **-E** or **-E**0 with one of the two **-F** options.

Here is the 18 examples of various transfer functions (elastic, viscoelastic, firmoviscous1, firmoviscous2, viscous2, viscous2 from top to bottom row):

transfer

PaulWessel:
Sorry, I was initially tricked ty grdflexure.c to think **-Q** was taken and had to invent **-G+d**.  However, a clean **-Q** is better.

seisman:
> Sorry, I was initially tricked by grdflexure.c to think **-Q** was taken and had to invent **-G+d**. However, a clean **-Q** is better.

I still see **-G+d** in the "Transfer Functions" section. Should it be replaced with **-Q**?

PaulWessel:
Yep, pretty sure nothing got broken but several things got fixed, so a step forward.  Daniel who posted on the Forum is a student here and I am on his committee . Funny that he posts there instead of emailing me but that is fine.  So he will find any other problems anyway.

PR #3338, opened 2020-05-15
List countries limited to a continent

PaulWessel:
It is useful to be able to list all the countries in a continent. Now, **-E**[=_continent_]**+l+L** will do that [Default is all countries, as before].


PaulWessel:
Should work now. coast failed because of not checking for strings +l or +L but specific character positions.  `gmt coast -E=EU,AF+l `etc works for me now.

PR #3342, opened 2020-05-16
Clarify MOVIE_WIDTH and warn about MAP_FRAME_TYPE

### LINKED ISSUE #3341: Bug with -JG North/South Pole and -B0 and issues with MOVIE_WIDTH
I create an animation of the earth rotating from the south pole to the north pole following the example of the manpages of [Movie](https://docs.generic-mapping-tools.org/latest/movie.html) and I found: 

1) A bug with respect the figures of the South/North Pole. The border (only happened with -Bg) is slightly separated from the edge of the map (see example below).
2) An issue with movie. When I use %MOVIE_WIDTH% to set the width of the figure, the frame is cut in the sides and left out of the animation. The width is just for the map. Maybe this could clarify in the docs.

```
Movie.bat
gmt math -T-90/90/1 -o1 T = "angles.txt"
gmt movie "globe.bat" -Nglobe7 -T"angles.txt" -Agif -C15cx15cx100

Globe.bat
gmt pscoast -Rg -JG-65/%MOVIE_COL0%/%MOVIE_WIDTH% -Ggray -Sdodgerblue2 -Cgray -N1/0.2,- -Bg -X0 -Y0 -W1/
```

**Actual outcome**
globe7_000

globe7_018

**System information**

* Operating system: Win 10
* GMT version (`gmt --6.0.0`):

PaulWessel:
Addresses and closes #3341.

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -296,8 +296,8 @@ Parameters
 Several parameters are automatically assigned and can be used when composing *mainscript* and the optional
 *background* and *foreground* scripts. There are two sets of parameters: Those that are constants
 and those that change with the frame number.  The constants are accessible by all the scripts:
-**MOVIE_WIDTH**\ : The width of the canvas,
-**MOVIE_HEIGHT**\ : The height of the canvas,
+**MOVIE_WIDTH**\ : The width of the canvas (the full movie frame),
+**MOVIE_HEIGHT**\ : The height of the canvas (the full movie frame),
 **MOVIE_DPU**\ : The current dots-per-unit,
 **MOVIE_RATE**\ : The current number of frames per second,
 **MOVIE_NFRAMES**\ : The total number of frames.
@@ -365,7 +365,15 @@ making a HD movie using the US unit dimensions then a background pink layer woul
 Note the canvas selection via :term:`PS_MEDIA`, the matching region and projection, and
 the zero location of the origin.
 
-Transparent images
+Basemap Frames
+--------------
+
+Some map projections will by default draw a *fancy* map frame; this feature is under the control of :term:`MAP_FRAME_TYPE`.
+However, whether a *fancy* or *plain* frame is actually drawn also depends on the projection center *latitude*.
+Thus, if your movie varies the projection center latitude by changing the view, you should set the frame
+setting to *plain* as part of your setup.
+
+Transparent Images
 ------------------
 
 By default, **movie** will build *opaque* PNG images which can then be assembled into an animation.


Esteban82:
I don't know if I a was clear with the issue of MOVIE_WIDTH. 

What I meant is that if the parameter MOVIE_WIDTH is used to set the width of the map (as in the example of manpages copied below), then the frame (outside type) of the maps (and hence of the animation) would be left outside of the canvas and would not appear in the final animation.

_gmt coast -Rg -JG${MOVIE_FRAME}/20/${**MOVIE_WIDTH**} -Gmaroon -Sturquoise **-Bg** -X0 -Y0_

That is why in the figure #3341 the circle of the frame discontinous (cut in up/down/left/rigth sides). So to avoid this I have to set an slightly smaller width for the map. For example with the next scripts:

_gmt movie globe.bat -Nglobe -T360 -Agif -C**6i**x6ix100 -Lf
gmt coast -Rg -JG${MOVIE_FRAME}/20/**5.9i** -Gmaroon -Sturquoise -Bg -X0 -Y0_

PaulWessel:
Yes, but that is tied into the concept of a plot margin.  This could perhaps need some improvement. You are right that **MOVIE_WIDTH** is not suitable as argument to -J because annotations will be clipped.  There is also **-X -Y** that moves the origin and then using **MOVIE_WIDTH** really clips your map.  However, this is exactly the same thing users go through when maping a plot for an A4 paper, no? If you pick a too large width then clipped.  The only way I could offer a **MOVIE_WIDTH** that would work fine would be to compute it from the canvas width minus the margins on either side.  But we have no concept of margin dimensions in movie (we do in subplot for instance).  We do have a left/bottom margin via **-X -Y**, but no similar "margin" for right and top.

I have hesitated to introduce new concept for plotting in movie that are not used in other plotting.

Esteban82:
I wasn't asking to have a new concept for plotting should be add. It seems like to many work. 

But I think that a warning like: "using MOVIE_WIDTH may not be suitable as argument to -J because annotations will be clipped" could be add to the [manpage](https://docs.generic-mapping-tools.org/latest/movie.html?highlight=movie). 

And also I think it might be prudent to remove the -Bg from the examples of the manpage of movie to avoid any other user to have this issue.

PaulWessel:
Hi Federico, I understood what you meant. I was just thinking out load for a solution.  I can certainly add that warning the the man page.  We may be addressing the **-X -Y** situation via the new gmt themes we are working on (#3344), and I am thinking about a margin option...

PaulWessel:
I have added language on **MOVIE_WIDTH** etc to the man page.

PaulWessel:
Here are my thoughts on canvas composition.  First, the current system, which follows standard GMT layout given the size of your paper: Design your layout on the given page.  Instead of an actual A4 or letter paper, you have a slightly smaller paper, say 21.6 x 10.8 cm for a 16:9 movie.  As with any paper you want to move the plot origin away from the actual lower left corner.  This is done with **-X** and **-Y** and these have 72 points as default (~2.5 cm).  You then use a reasonable fraction of the remaining area to plot.  The **movie** module does not provide much help here: It inherits the standard GMT (72p, 72p) origin offset, and it produces two script variables available to users: **MOVIE_WIDTH** and **MOVIE_HEIGHT**. However, as pointed out by @Esteban82, these dimensions are not really suitable as projection dimensions since the plot will be clipped, severely so if you have annotations.  You have to decide on the green plot dimensions manually.The figure below indicates the situation:

canvass

A possible new way of doing things is illustrated below.  Here, we would rely on the _movie_ theme which changes the default origin to (0,0) (and probably set slightly smaller annotation fonts etc given the smaller paper). In addition, it would introduce a new **-O** option to **movie**, based on similar option in subplot:

**-O**_off_ | _xoff_/_yoff_ | _woff_/_eoff_/_soff_/_noff_

You can then specify a constant margin width, one that differs for x and y, or different margins for all four sides.  In addition, **MOVIE_WIDTH** and **MOVIE_HEIGHT** will be written out as the actual dimensions of the green rectangle, now suitable for plot dimensions:

canvas

The **-O** option, if not given, sets no margins, i.e., origin remains at (0,0). If **-O** is given with no arguments we could decide on some sensible defaults, perhaps 7.5%/2.5/5%/5% of the canvas dimensions.

**Pro**: This scheme would make **MOVIE_WIDTH** and **MOVIE_HEIGHT** usable parameters (they are currently not even used by our own animations), and it may be simpler to think about layout in terms of margins.  This is how word processors work as well (i.e., set margins).  It also hides the use of **-X -Y** for most simple layouts since the margins sets those implicitly.  Unlike plotting in modern mode in general, where there is no paper size at all, a movie actually has a strict paper size, hence a different approach is needed.
**Con**: This scheme treats movie plot layouts differently than for any other GMT plot.  Suddenly, we no longer talk  about **-X** and **-Y** but of _margins_! While classic mode people still think of media sizes, the modern crowd do not, hence margins may be a bit odd for them.

In conclusion, I am not sure if this is a good solution, but I think the above at least lays out the issue.

seisman:
@PaulWessel Perhaps you should open a new PR and repost it so that we won't forget it.

PaulWessel:
OK, now at #3357.

PR #3343, opened 2020-05-16
Help user avoid common movie/batch mistakes

PaulWessel:
All the **MOVIE_*** or **BATCH_*** script variables must have a leading $ or % to actually give a value.  It is pretty simple to forget to do so, and then you just have a random text in your script.  Now, both batch and movie looks for that. We also recogonize the deprecated ways to do a sub-shell (`...`) and recommend $(...) instead.
Finally, the common functions used by both batch and move were incorrectly named _gmtlib_.  They need to be _gmt__* per our convention.

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -16809,7 +16809,7 @@ CROAK:	/* We are done or premature return due to error */
  * input script types and to write shell commands in various syntax variants.
  */
 
-int gmtlib_gmt_sleep (unsigned int microsec) {
+int gmt_sleep (unsigned int microsec) {
 	/* Waiting before checking if the completion file has been generated */
 #ifdef WIN32
 	Sleep ((uint32_t)microsec/1000);	/* msec are microseconds but Sleep wants milliseconds */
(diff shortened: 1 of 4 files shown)


src/gmt_support.c
```diff
@@ -16975,13 +16975,46 @@ bool gmtlib_script_is_classic (struct GMT_CTRL *GMT, FILE *fp) {
 	return (!modern);
 }
 
+int gmt_token_check (struct GMT_CTRL *GMT, FILE *fp, char *prefix, unsigned int mode) {
```
seisman, line 16978: Some potential issues:
1. Tried with anim01, but it reports some errors like
```
Main script uses MOVIE_FRAME} but without the required leading $
```
2. Comment lines or inline comments should be skipped
3. csh doesn't support `$()` syntax

PaulWessel: Addressed in the update

seisman:
I'm not sure it's a good idea to check the MOVIE_* and BATCH_* variables. Sometimes users may want to write a string MOVIE_FRAME, e.g.,
```
echo "MOVIE_FRAME = ${MOVIE_FRAME}"
```

but this PR will report an error.

PaulWessel:
Hm, I see you point.  But why would anyone do an echo in those scripts? Progress options are available.  I think the benefit of being told of problems outweigh at least that case you are presenting.

seisman:
@PaulWessel This PR now fails, due to the renaming of some functions, e.g., gmt_set_comment.

PR #3345, opened 2020-05-18
Let grdmix write geotiff when it knows we have geographic data

PaulWessel:
For unclear reasons, _gmt_gdalread_ fails to detect that the @BlueMarble_02m.tif image is **-Rd**.  Hence, the header **-R** is 0/10800/0/5400 instead.  Surely, this should not happen, @joa-quim? In the meantime, since **grdmix** may be reading grids, and grids are more reliable indicators of the actual **-R** and whether the data are geographic, we consult any grids read in and acquire _wesn_ from them, as well as set geographic if they say so.

### DIFF
--- a/src/grdmix.c
+++ b/src/grdmix.c
@@ -315,17 +315,20 @@ GMT_LOCAL float *grdmix_get_array (struct GMT_CTRL *GMT, struct GRDMIX_AIW *X, i
 EXTERN_MSC int GMT_grdmix (void *V_API, int mode, void *args) {
 	char *type[N_ITEMS] = {NULL, NULL, NULL, "alpha", "blend", "intens"};
 
+	bool got_R = false;
+
 	int error = 0;
 	unsigned int img = 0, k, band;
 	int64_t row, col, node, pix;
 
 	float *weights = NULL, *intens = NULL, *alpha = NULL;
 
-	double rgb[4];
+	double rgb[4], wesn[4] = {0.0, 0.0, 0.0, 0.0};
 
 	struct GMT_GRID *G_in[N_ITEMS], *G = NULL;
 	struct GMT_IMAGE *I_in[N_ITEMS], *I = NULL;
 	struct GMT_GRID_HEADER *h[N_ITEMS], *H = NULL;
+	struct GMT_GRID_HEADER_HIDDEN *HH[N_ITEMS];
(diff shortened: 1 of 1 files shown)


joa-quim:
``gdalinfo`` says
```
Corner Coordinates:
Upper Left  (-180.0000000,  90.0000000)
Lower Left  (-180.0000000, -90.0000000)
Upper Right ( 180.0000000,  90.0000000)
Lower Right ( 180.0000000, -90.0000000)
Center      (   0.0000000,   0.0000000)
```
but nothing more informs these are geogs. We will have to rely on number guessing

PaulWessel:
OK, I guess if no grid input, only image, and image says not geographic then we would require -R to set things.
Since we are distributing these marbles, what gdal command should I do to make them truly geotiffs?

joa-quim:
I think
``gdal_translte -a_srs "+proj=longlat +datum=WGS84" ....``

PaulWessel:
Not quite.  I ended up with this (to also compress well):

```
gdal_translate -a_srs "+proj=longlat +datum=WGS84" -a_ullr -180 90 180 -90 -co "COMPRESS=LZW" -co "PREDICTOR=2" -co "TILED=YES" BlackMarble_10m.tif new.tif

gmt grdinfo new.tif
new.tif: Title: Grid imported via GDAL
new.tif: Command: 
new.tif: Remark: 
new.tif: Pixel node registration used [Cartesian grid]
new.tif: Grid file format: gd = Import/export through GDAL
new.tif: x_min: -180 x_max: 180 x_inc: 0.166666666667 name: x n_columns: 2160
new.tif: y_min: -90 y_max: 90 y_inc: 0.166666666667 name: y n_rows: 1080
new.tif: z_min: 0 z_max: 255 name: z
new.tif: scale_factor: 1 add_offset: 0
+proj=longlat +datum=WGS84 +no_defs
```

So the regions is good, and it reports the proj setting, but why is it not flagged as a geo? In gmt_gdalread, when you get that proj string from an image and it has longlat, do you call gmt_M_setgeographic (GMT, GMT_IN)?  You should.

PaulWessel:
I see no calls to gmt_set_geographic in gmt_gdalread.c.  If you detect the PROJ string, could you add that call for us please?

PR #3356, opened 2020-05-19
Split trailing text into words

PaulWessel:

Currently, movie and batch can split trailing text into individual words.  It always does this by using white spaces (tabs and spaces) as separators.  However, in many cases where text are involved, the individual "words" may have spaces, like "United Kingdom" and then we don't want to split on spaces, just on tabs only.  This PR adds a **+W** modifier to **-T** to go along with the **+w** modifier.  The latter is the current practice, while uppercase **+W** only splits on tabs.  This works in this PR.

However, (and why this is still marked as WIP for now), my test case (new example in the batch man page to make a map of each European nation) ran into trouble.  Specifically, this command

`gmt coast -E=EU+l > countries.txt`

creates a file like this:

```
  AD	Andorra
  AL	Albania
  AT	Austria
  AX	Åland Islands
  BA	Bosnia and Herzegovina
...

```

and it has two leading spaces.  Of course, one solution is to stop **coast** from doing this silliness.  The other is to anticipate that other files may have some leading whitespace as well.  Thus, do you think it is OK to strip off leaving spaces in trailing text? I think we always do when there is numerical data in front, so for most files it is not an issue.  But here it is _just_ text, and the spaces survive.  I think removing them is a safe procedure and should be implemented by the deep i/o machinery so this does not happen.

Thoughts?


PaulWessel:
To be clear, I am suggesting that I

1. Fix coast to not dump leading spaces anyway
2. Fix the i/o machinery to consistently skip leading spaces in trailing text.

PaulWessel:
I am (re)learning that gmt_DCW_list simply dumps stuff with printf so `gmt coast -E=EU+l` cannot possibly return a list to MEX, Julia, or Python.  I think I need to put in the work to make a DATASET and write it to stdout like a proper module... I will remove the leading spaces in the process.  Since the above I have learned that we currently do not string leading spaces if the file is entirely text, so we could (after fixing list) make this a user problem to prepare proper input.

PaulWessel:
Decisions: I will leave the inner GMT i/o related to text-only files so we dont break anything.  However, with #3358 we now have gained the ability to get country and state listings in the external environments since GMT_DCW_list now creates a proper dataset and writes it to stdout.
Thus, I took off the WIP as this one should be good to go.

PaulWessel:
Require eyeballs and approvals, @joa-quim and @seisman.

joa-quim:
Instead if space vs tabs that are very difficult to distinguish, why not use true true CSV separation? I.e
``United Kingdom,  Ireland, etc``

PaulWessel:
GMT can read those.  Here is what it looks like:

```
echo "1,2,3,United Kingdom,Shit" | gmt convert
1	2	3	United Kingdom,Shit
```

So commas are part of the trailing text, and yes, one could split on that as well.  We do not have a GMT defaults for this (we accept many things as column divider on input); it is only on output we have **IO_COL_SEPARATOR**.

An alternative to my implementation would be **+w**[_char_].  If no _char_ then any whitespace is used, else only char is used to separate.  How should we pass TAB here in a script? +w\t ?  Or better, make it a string [Deafult to " \t"] and I think then +w"\t" should work, or eveh +w"<tab>".

PaulWessel:
ANyway, I agree with your principle, but I want to control that char.  Might be | for some people.

joa-quim:
Well, the *C* is ``CSV`` can in fact be any char, not necessarily a comma.

PaulWessel:
OK, now uses +w[str].  Minor complication is that you may have to tive \t and then it is the txt \t.  So had to add a function that replaces "\t" with the TAB character.  Tested and works.

PaulWessel:
Hoping the wrap this PR up.  Here is the latest:

1. batch: Let the postflight script be any script, not necessarily a modern, or even a GMT script.
2. batch: Add example in documentation that makes a book of PDF maps of all European nations.
3. movie: Initialize sep to white space.
4. both: Allow replacement of "\t" literal with the TAB character under the hood.
5. both: Fix various typos in documentation and comments.

Might this pass muster now, @seisman ?

PaulWessel:
Forgot about this one @seisman I think it is good to merge, no?

PR #3366, opened 2020-05-21
Optional use of -Z in DCW header dumps

PaulWessel:

When gmt coast **-M -E** is used to write polygons to stdout, the segment headers now has the settings **-Z**_code_.  This is not always appropriate.  Hence, this PR adds the modifier **+z** for **-E** which is required to get the country codes written via **-Z** in the headers.

### DIFF
--- a/src/gmt_dcw.c
+++ b/src/gmt_dcw.c
@@ -43,10 +43,8 @@
 
 #define DCW_GET_COUNTRY			1	/* Extract countries only */
 #define DCW_GET_COUNTRY_AND_STATE	2	/* Extract countries and states */
-#define DCW_ADD_ZCOUNTRY		4	/* Report -Z<countrycode> in segment header if -M */
 #define DCW_DO_OUTLINE			1	/* Draw outline of polygons */
-#define DCW_DO_FILL			2	/* Fill the polygons */
-#define DCW_DO_FILL			2	/* Fill the polygons */
+#define DCW_DO_FILL				2	/* Fill the polygons */
 
 struct GMT_DCW_COUNTRY {	/* Information per country */
 	char continent[4];	/* 2-char continent code (EU, NA, SA, AF, AU, AN) */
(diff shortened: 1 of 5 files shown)


src/gmt_dcw.c
```diff
@@ -43,8 +43,10 @@
 
 #define DCW_GET_COUNTRY			1	/* Extract countries only */
 #define DCW_GET_COUNTRY_AND_STATE	2	/* Extract countries and states */
+#define DCW_ADD_ZCOUNTRY		4	/* Report -Z<countrycode> in segment header if -M */
 #define DCW_DO_OUTLINE			1	/* Draw outline of polygons */
 #define DCW_DO_FILL			2	/* Fill the polygons */
+#define DCW_DO_FILL			2	/* Fill the polygons */
```
joa-quim, line 49: Duplicated line?

PaulWessel: Sorry, copy paste and then abandonment.  Fixed

PaulWessel:
Please check that it works for you.  I am having the odd situation that happens from time to time when running it in Xcode shows that it works (and it is nothing magical here) but even though I rebuilt the command-line from scratch it does not work.  I know if I reboot it will - so some dark magic with macOS and shared library memories?

seisman:
It seems these two commands give me the same output.
```
gmt coast -EUS -M > jk1
gmt coast -EUS+z -M > jk2
```

joa-quim:
Yes, just confirmed that too.

PaulWessel:
Fascinating.  So when debugging in Xcode it works but running on command line it does not.  OK, back to fprintf....

joa-quim:
It's not only the -Z that is not written but the country name also
```
$ pscoast -EPT -M | head -1
> Segment 0
```

PaulWessel:
Found the problem, but still fascinating it ran fine in Xcode..

joa-quim:
Install [Visual Studio for Macs](https://docs.microsoft.com/en-us/visualstudio/mac/installation?view=vsmac-2019)

It now works, but it still doesn't write the country name when no **+z**, only the Segment number like I posted above
```
$ pscoast -EPT -M | head -1
> Segment 0
```

PaulWessel:
OK, that is by design.  So you want
`> Portugal segment 0`

or

`> -ZPT Portugal segment 0`

joa-quim:
> Portugal segment 0

This is what 6.0 does and I think it's useful to see the country name, much more than ``segment 0``.
Now with **+z** it adds **-ZPT**

PR #3368, opened 2020-05-21
Update Fedora install instructions

seisman:
Fedora is a continously-evolving distro and has a very short release cycle.
Fedora usually releases a new version every ~6 months, and maintains each version for ~13 months.

Currently, the supported Fedora versions are:

- Fedora 30
- Fedora 31
- Fedora 32
- Fedora Rawhide (testing/unstable)

Fedora >=31 are providing GMT 6.0.0, and are actively maintained, and
Fedora 30 will EOL at the end of this month.

Here we recommend Fedora users to use the GMT packages from the Fedora official repository.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -136,34 +136,15 @@ These three GMT versions cannot live side by side.
 
 ### Fedora
 
-The GMT binary packages provided by the Fedora official repositories are usually too old.
-We provide [the GMT official RPM repository](https://copr.fedorainfracloud.org/coprs/genericmappingtools/gmt)
-to allow Fedora users access the latest GMT releases in an easy way.
+GMT 6 packages are available for **Fedora 31 or newer**. Install it via:
 
-**NOTE: The RPM repository provides GMT packages for Fedora 30 or newer only!**
-
-Fedora users can add the GMT official RPM repository and install gmt by:
-
-	# enable the RPM repository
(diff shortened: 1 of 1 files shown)


seisman:
Ping @PaulWessel

INSTALL.md
```diff
@@ -136,34 +136,15 @@ These three GMT versions cannot live side by side.
...
-	dnf install gmt
+GMT 6 packages are available for **Fedora 31 or newer**. Install it via:
 
-	# Update to the latest version if available
-	dnf update gmt
+    dnf install GMT dcw-gmt gshhg-gmt-nc4 gshhg-gmt-nc4-full gshhg-gmt-nc4-high ghostscript
 
 You may also install other optional dependencies for more capabilities within GMT:
 
+    dnf install GraphicsMagick gdal
```
seisman, line 145:
```suggestion
```

seisman:
> Strange, I know I responded to this some time ago. My question was, does fedora build GMT with GDAL? It is a compile dependency, so I dont understand why you can use dnf to install gdal later since GMT is already built, no?

Yes, GMT is built with GDAL. Actually, both GDAL and GraphicsMagick are automatically installed by `dnf install gmt`, but `ghostscript` isn't. 

I've update the install instruction.

PR #3373, opened 2020-05-23
Update gmt_getremote.sh

seisman:

`sh gmt_getremote.sh data` only download the data files in the `/server` directory. The old `earth_relief` files in the root directory are not downloaded.

### DIFF
--- a/share/tools/gmt_getremote.sh
+++ b/share/tools/gmt_getremote.sh
@@ -6,6 +6,7 @@
 # This script downloads any missing cache or data sets from the remote server.
 #
 # Run this script on the command line with:
+#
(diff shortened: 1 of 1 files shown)


share/tools/gmt_getremote.sh
```diff
@@ -34,18 +35,41 @@ fi
 
 # Get current remote data server
 SERVER=$(gmt get GMT_DATA_SERVER)
+
+# The "gmt which" command
+gmtwhich="gmt which -Gu"
+#Uncomment the line below for a dry-run
+# gmtwhich="printf %s\n"
```
seisman, line 42: You can uncomment this line to check which files will be downloaded without downloading them.

PR #3377, opened 2020-05-23
Update PS files due to changes of earth relief data

seisman:
Some tests fail due to the recent updates of earth relief data. This PR updates the PS files using the new earth relief dataset.


seisman:
Ping @PaulWessel.

PR #3387, opened 2020-05-25
Also search in sharedir in addition to its given subdir

PaulWessel:
This PR will helps users who temporarily want to change where **GMT_SHAREDIR** points to: Because GMT only looks in _subdir_ under **GMT_SHAREDIR** it is more awkward to use when you just want to point to a temporary directory.  Now, we will also look in **GMT_SHAREDIR** itself  if **GMT_SHAREDIR**/_subdir_ fails to find the file.

Also, the two test that temporarily reset **GMT_USERDIR** have been changed to use the files directly without changing the environment as that negatively affects other jobs running in parallel.

### DIFF
--- a/test/psxy/line_geo.sh
+++ b/test/psxy/line_geo.sh
@@ -18,3 +18,4 @@ awk '{printf "@~a@~ = %s@.\n", $3}' q1.txt | gmt pstext -R -J -O -K -F+f10p+cBL 
 gmt psxy -R -J -Skgeo-lineation/5c -O -N -W2p,red -Gred -Bag0.5 -BWSne q2.txt -K -Y-10c >> $ps
 awk '{printf "@~a@~ = %s@.\n", $3}' q2.txt | gmt pstext -R -J -O -K -F+f10p+cBL -Gwhite -Dj0.2c >> $ps
 gmt psxy -R -J -O -T >> $ps
+rm -f geo-lineation.def
--- a/test/psxy/line_geo.sh
+++ b/test/psxy/line_geo.sh
@@ -3,7 +3,7 @@
 # This version is a Mercator plot and azimuths are given.
 ps=line_geo.ps
 # Must copy the specific symbol file from the documentation to here.
-DIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
+DIR=$(gmt --show-sharedir)/../doc/rst/source/users-contrib-symbols/geology
 cp -f ${DIR}/geo-lineation.def .
 echo "0 0  60 30" > q1.txt
 echo "0 0 150 30" > q2.txt
--- a/test/psxy/line_geo.sh
+++ b/test/psxy/line_geo.sh
@@ -2,8 +2,9 @@
 # Test strike/dip symbol from Jose for consistency across quadrants
 # This version is a Mercator plot and azimuths are given.
 ps=line_geo.ps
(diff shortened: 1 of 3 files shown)


test/psxy/line_geo.sh
```diff
@@ -2,8 +2,9 @@
 # Test strike/dip symbol from Jose for consistency across quadrants
 # This version is a Mercator plot and azimuths are given.
 ps=line_geo.ps
-# Must temporarily change GMT_USERDIR to the gallery documentation dir
-export GMT_USERDIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
+# Must copy the specific symbol file from the documentation to here.
+DIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
+cp -f ${DIR}/geo-lineation.def .
```
seisman, line 7: It would be better to remove the def file at the end of the script.

test/psxy/struct_geo.sh
```diff
@@ -2,9 +2,9 @@
 # Test all structural geology symbols from Jose
 ps=struct_geo.ps
 reg=-R0/10/0/10
-# Must temporarily change GMT_USERDIR to the gallery documentation dir
-export GMT_USERDIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
-
+# Must copy the specific symbol files from the documentation to here.
+DIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
+cp -f ${DIR}/*.def .
```
seisman, line 7: It would be better to remove the def files at the end of the script.

test/psxy/struct_geo.sh
```diff
@@ -2,9 +2,9 @@
 # Test all structural geology symbols from Jose
 ps=struct_geo.ps
 reg=-R0/10/0/10
-# Must temporarily change GMT_USERDIR to the gallery documentation dir
-export GMT_USERDIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
-
+# Must copy the specific symbol files from the documentation to here.
+DIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
```
seisman, line 6:
```suggestion
DIR=$(gmt --show-sharedir)/../doc/rst/source/users-contrib-symbols/geology
```

test/psxy/line_geo.sh
```diff
@@ -2,8 +2,9 @@
 # Test strike/dip symbol from Jose for consistency across quadrants
 # This version is a Mercator plot and azimuths are given.
 ps=line_geo.ps
-# Must temporarily change GMT_USERDIR to the gallery documentation dir
-export GMT_USERDIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
+# Must copy the specific symbol file from the documentation to here.
+DIR=`gmt --show-sharedir`/../doc/rst/source/users-contrib-symbols/geology
```
seisman, line 6:
```suggestion
DIR=$(gmt --show-sharedir)/../doc/rst/source/users-contrib-symbols/geology
```

test/psxy/line_geo.sh
PaulWessel: Done

test/psxy/struct_geo.sh
PaulWessel: Done

PR #3389, opened 2020-05-26
No exit calls, part 1

PaulWessel:

This PR removes numerous gratuitous _GMT_exit_ calls from any module and from a bunch of the gmt_*.c files.  There is more, but this is a start, hence "part 1".

All tests continue to pass.  Most of these exit calls were not possible to trigger, I think, unless a developer passed a wrong argument to a function internally, and those who could be triggered would probably just crash the module instead.  In some places I pass the error cope up, if possible, and call Return (error) when I can.

This remains now:

```
./gmt_api.c:        2
./gmt_init.c:       29
./gmt_io.c:       22
./gmt_map.c:       17
./gmt_plot.c:       11
./gmt_sph.c:       14
./gmt_support.c:       35
./gmt_types.h:        2
./mgd77/mgd77.c:       25
./spotter/spotter.c:       14
./x2sys/x2sys.c:       16

```

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -482,8 +482,8 @@ GMT_LOCAL inline GMT_putfunction gmtapi_select_put_function (struct GMTAPI_CTRL 
 		case GMT_UCHAR:		return (gmtapi_put_val_uchar);	break;
 		case GMT_CHAR:		return (gmtapi_put_val_char);	break;
 		default:
-			GMT_Report (API, GMT_MSG_ERROR, "Internal error in gmtapi_select_put_function: Passed bad type (%d), Must abort\n", type);
-			GMT_exit (API->GMT, GMT_RUNTIME_ERROR); return NULL;
+			GMT_Report (API, GMT_MSG_ERROR, "Internal error in gmtapi_select_put_function: Passed bad type (%d), Will probably crash\n", type);
+			return NULL;
 			break;
 	}
 }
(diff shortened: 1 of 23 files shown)


src/gmt_api.c
```diff
@@ -501,8 +501,8 @@ GMT_LOCAL inline GMT_getfunction gmtapi_select_get_function (struct GMTAPI_CTRL
 		case GMT_UCHAR:		return (gmtapi_get_val_uchar);	break;
 		case GMT_CHAR:		return (gmtapi_get_val_char);	break;
 		default:
-			GMT_Report (API, GMT_MSG_ERROR, "Internal error in gmtapi_select_get_function: Passed bad type (%d), Must abort\n", type);
-			GMT_exit (API->GMT, GMT_RUNTIME_ERROR); return NULL;
+			GMT_Report (API, GMT_MSG_ERROR, "Internal error in gmtapi_select_get_function: Passed bad type (%d), Must probably crash\n", type);
```
joa-quim, line 504: "Must probably crash"?

joa-quim:
And remember that we have ``PSL_exit()`` too

src/gmt_api.c
PaulWessel: OK,a dded lots more check on this pointer elsewhere

PaulWessel:
> And remember that we have `PSL_exit()` too, and from a distant recall it was more complicated to deal with.

Dealt with in #3390

PR #3393, opened 2020-05-26
Add vector translate operator to gmt vector

PaulWessel:

We had no place in GMT where it was possible to give coordinates and have them translated by a constant or variable amount in a constant or variable azimuth direction.  This has now been implemented in **gmtvector** as option **-Tt**[[_az_/_dist_][_unit_]].  If a fixed azimuth and distance is appended then that is what we use, else we expect to read azimuth and distance from the 3rd and 4th column.  The unit of distance (for geographic data) read from file is assumed to be meter unless another unit is give via **-T**tunit.
Two test scripts (one Cartesian, one geographic) have been added; here is the output of the geographic test:

translate

The data is a spiral and for the constant case we translate by 2 degrees in ax = 60, while for the variable we compute an azimuth 0-36o and a translation by 0-3.6 degrees.

### DIFF
--- a/doc/rst/source/gmtvector.rst
+++ b/doc/rst/source/gmtvector.rst
@@ -116,10 +116,10 @@ Optional Arguments
     angle for 2-D Cartesian data and *lon/lat/angle* for a 3-D rotation
     pole and angle), **R** will instead rotate the fixed secondary vector
     by the rotations implied by the input records, **t** to translate
-    the input point by *distance* in the *azimuth* direction (append *azimuth*/*distance*[*unit*]
+    the input point by *distance* in the *azimuth* direction (append *azimuth*/*distance*\ [*unit*]
     for the same translation for all input points, or just append
-    *unit* to read *azimuth* and *distance* (in specified unit)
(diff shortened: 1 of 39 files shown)


joa-quim:
This is spherical only, right?
The method is known as *[Dead Reckoning](https://en.wikipedia.org/wiki/Dead_reckoning)* so maybe name the *gmt_transalate_pos* function according to it.

I would rather this functionality in ``mapproject`` given all related (**-G**, **-L**) options. But ``mapproject`` is already so heavy loaded with things that are not *projection* related. Ideally we should create a new module to aggregate this type of operations. *gmtvector* doesn't seam the right place neither.

PaulWessel:
Spherical and Cartesian like all the other options.  Rotation and Translation are the two fundamental ways to move a point.  gmtvector now does both, so it fits it well I think.

joa-quim:
Still not convinced that gmtvector is the best place for a Dead Reckoning option but won't discuss more.

PaulWessel:
I thought of mapproject but it is completely packed.  And the name is not optimal, but too late for those changes.

PR #3395, opened 2020-05-27
Missing brackets around loop

PaulWessel:
We hae a loop in gmt_map.c that was missing {}  around two lines in six places.  Fixing this caused no changes in the tests so probably harmless.  Yet, why did it not crash? The loop sets i to GMT_GRID_UPPER and then it is passed into the next line?

### DIFF
--- a/src/gmt_map.c
+++ b/src/gmt_map.c
@@ -3108,9 +3108,10 @@ GMT_LOCAL bool gmtmap_init_stereo (struct GMT_CTRL *GMT) {
 		}
 		else {	/* Global view only */
 			/* No annotations or tickmarks in global mode */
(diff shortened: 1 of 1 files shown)


joa-quim:
Maybe ``GMT->current.map.frame.axis[GMT_Y].item`` is allocated > ``GMT_GRID_UPPER``

PaulWessel:
> Maybe `GMT->current.map.frame.axis[GMT_Y].item` is allocated > `GMT_GRID_UPPER`

Yes indeed, we have [8]. Wonder what items [6] and [7] were intended to do - or perhaps they do serve some useful function somewhere.  Anyway, that is why no crash.

PR #3401, opened 2020-05-28
Eliminate GMT_exit calls from gmt_map.c

PaulWessel:

We had 20 or so calls to exit in gmt_map.c, mostly when some map projection parameters was not in a valid range.  Many of these places were in void functions so no way to pass an error code up.  I have changed this to allow the error code to be returned so that the appropriate higher level function called in the module can call Return (error) and die a normal death.  Because many of these functions (e.g., _gmt_init_distaz_ ) are called in modules this PR affects numerous files.  I have slowly eliminated new bugs and fixed things until all tests again pass (for me).  
I also had to make a fix in pscoast for -JE since there are times we paint lakes with the ocean color when we cannot do the recursive painting.  However, we never considered -B+gfill setting in that scenario.  This accected the placement.sh test and tnow the new pscoast_JE4.sh test.  I also found a file in cache was still in a test dir, causing failure depending on which file was used.  
Please run the test to see if you agree that all is well before you approve.  Important that this is checked on WIn and Linux as well as macOS.


seisman:
> Important that this is checked on WIn and Linux as well as macOS.

Do you want to temporarily enable the whole tests in this branch?

PaulWessel:
yes please.  This PR got too big unfortunately, as I initially got lots of failed tests and had to expand it...  For the next exit round I will go small and frequently...

PaulWessel:
Thanks!

PR #3410, opened 2020-05-30
Add -Z option to grdconvert to enable scaling/translation services

PaulWessel:
We need to distinguish between what the **+s** and **+o** modifiers in a _filename_ do versus the ability to permanently change grid values (like what **grdmath** would do).  It is convenient to have a **-Z** option in **grdconvert** to avoid having to do separate **grdmath** calls for this simple task.

### DIFF
--- a/doc/rst/source/grdconvert.rst
+++ b/doc/rst/source/grdconvert.rst
@@ -16,7 +16,7 @@ Synopsis
 [ |-N| ]
 [ |SYN_OPT-R| ]
 [ |SYN_OPT-V| ]
-[ |-Z|\ [**+s**\ *factor*][**+o**\ *shift*] ]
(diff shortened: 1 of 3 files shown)


doc/rst/source/grdconvert.rst
```diff
@@ -16,6 +16,7 @@ Synopsis
 [ |-N| ]
 [ |SYN_OPT-R| ]
 [ |SYN_OPT-V| ]
+[ |-Z|\ [**+s**\ *factor*][**+o**\ *shift*] ]
```
seisman, line 19: Change shift to offset?

PaulWessel: Yep, done.

PR #3411, opened 2020-05-30
Improve docs and always write out the ffmpe command

PaulWessel:

Now we let the ffmpeg or gm call that assembles the movie always be displayed.  In addition, the documentation now addresses the case of rerunning the ffmpeg command without having to recreate all the frame images since you already have them,

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -537,7 +537,7 @@ Remaking Movie with Existing PNG Frames
 
 Perhaps you made your movie and then decided you want to change the frame rate or adjust something else in
 how the movie is put together from all the still images.  If you kept all the frame images then
-iyou do not have to rerun the whole render process.  Assuming you want a MP4 movie and that you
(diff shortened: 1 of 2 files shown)


doc/rst/source/movie.rst
```diff
@@ -527,11 +527,25 @@ Other Movie Formats
...
 At the end it will print the FFmpeg command used.  You can copy, paste, and modify this command to
 select other codecs, bit-rates, and arguments.  You can also use the PNG sequence as input to tools such
 as QuickTime Player, iMovie, MovieMaker, and other commercial programs to make a movie that way.
 
+Remaking Movie with Existing PNG Frames
+---------------------------------------
+
+Perhaps you made your movie and then decided you want to change the frame rate or adjust something else in
+how the movie is put together from all the still images.  If you kept all the frame images then
+iyou do not have to rerun the whole render process.  Assuming you want a MP4 movie and that you
```
seisman, line 540:
```suggestion
you do not have to rerun the whole render process.  Assuming you want a MP4 movie and that you
```

PR #3415, opened 2020-06-01
Set to NaN if input datetime strings are not in ISO datetime format

### LINKED ISSUE #3414: Better way to handle NaT (not-a-time)

Both [Python numpy](https://numpy.org/devdocs/reference/arrays.datetime.html) and [Matlab](https://www.mathworks.com/help/matlab/ref/nat.html) can have a datetime called NaT, i.e., not-a-time. It's similar to NaN but for time. Not sure if Julia also has something similar.

GMT can't handle NaT. It's OK for command line, but not for external interfaces like PyGMT. 

Currently, when a NaT string is passed to GMT, it reports an error and keeps running.
```
Unable to parse 1 datetime strings (ISO datetime format required)
```
We should at least change the error to warning.

What's worse is:
https://github.com/GenericMappingTools/gmt/blob/3003c5020e6615d38ceaac8bf4a8bdfa44cdfcaf/src/gmt_api.c#L12769-L12771
GMT converts the datetime strings as double values relative to the epoch time internally. In the codes above, if `dt[row]="NaT"`, the function `gmt_scanf()` returns `GMT_IS_NAN` and the variable `t_vector[row]` is untouched. The value of `t_vector[row]` is 0, which means 1970-01-01T12:00:00. The plot is incorrect if we're plotting a calendar plot around 1970-01-01.

Ping @PaulWessel, @joa-quim and @weiji14.

seisman:

Set the value to NaN and raise an warning if the input string can't be parsed as ISO datetime format.

Fixes #3414

**Reminders**

- [ ] Correct base branch selected? `master` for new features, `6.1` for bug fixes

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -3397,7 +3397,7 @@ GMT_LOCAL int gmtapi_write_matrix (struct GMT_CTRL *GMT, void *dest, unsigned in
 	GMT_2D_to_index = gmtapi_get_2d_to_index (GMT->parent, M->shape, GMT_GRID_IS_REAL);
 	if ((api_get_val = gmtapi_select_get_function (GMT->parent, M->type)) == NULL)
 		return (GMT_NOT_A_VALID_TYPE);
-	
+
 
 	/* Start writing Matrix to fp */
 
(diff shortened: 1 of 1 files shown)


weiji14:
Cool, thanks guys! I'll go and test it out after dinner, while recompiling `gmt` master from scratch :grinning:

PR #3416, opened 2020-06-01
Add --show-sharelib option to gmt.exe

joa-quim:
Very handy to tell Julia where to find the GMT shared lib in unix.


joa-quim:
It isn't. Shit

Issue #3424, opened 2020-06-05
Now GMT thinks all http files are OCEANIA


joa-quim:
```
grdinfo http://joa-quim.pt/mirone/misc/corvo.grd
grdinfo [NOTICE]: Remote data courtesy of GMT data server OCEANIA [https://oceania.generic-mapping-tools.org]

grdinfo [NOTICE]:   -> Download cache file: http://joa-quim.pt/mirone/misc/corvo
corvo.grd: Title:
corvo.grd: Command: triangulate -Gcorvo.grd -R-31.127/-31.079/39.664/39.725 -I.0002 -V
corvo.grd: Remark:
corvo.grd: Gridline node registration used [Cartesian grid]
corvo.grd: Grid file format: cf = GMT netCDF format (32-bit float, deprecated)
corvo.grd: x_min: -31.127 x_max: -31.079 x_inc: 0.0002 name: user_x_unit n_columns: 241
corvo.grd: y_min: 39.664 y_max: 39.725 y_inc: 0.0002 name: user_y_unit n_rows: 306
corvo.grd: z_min: 0 z_max: 710 name: user_z_unit
corvo.grd: scale_factor: 1 add_offset: 0
corvo.grd: format: classic
```

PaulWessel:
We are trying to take over the world, but I agree we should allow you to have your data in PT.  Will look.

seisman:
Just tried @joa-quim's command. Now it is:
```
gmt grdinfo http://joa-quim.pt/mirone/misc/corvo.grd
grdinfo [NOTICE]:   -> Download cache file: http://joa-quim.pt/mirone/misc/corvo
corvo.grd: Title:
corvo.grd: Command: triangulate -Gcorvo.grd -R-31.127/-31.079/39.664/39.725 -I.0002 -V
corvo.grd: Remark:
corvo.grd: Gridline node registration used [Cartesian grid]
corvo.grd: Grid file format: cf = GMT netCDF format (32-bit float, deprecated)
corvo.grd: x_min: -31.127 x_max: -31.079 x_inc: 0.0002 name: user_x_unit n_columns: 241
corvo.grd: y_min: 39.664 y_max: 39.725 y_inc: 0.0002 name: user_y_unit n_rows: 306
corvo.grd: z_min: 0 z_max: 710 name: user_z_unit
corvo.grd: scale_factor: 1 add_offset: 0
corvo.grd: format: classic
```
the grid suffix is missing in the notice message:
```
[...]

PR #3426, opened 2020-06-06
Fix mixup with GDAL-special URLs and regular URLs

### LINKED ISSUE #3425: accessing /vsicurl/http  files no longer works
```
 grdinfo /vsicurl/http://larryfire.files.wordpress.com/2009/07/untooned_jessicarabbit.jpg
grdinfo [ERROR]: File /vsicurl/http://larryfire.files.wordpress.com/2009/07/untooned_jessicarabbit.jpg was not found
grdinfo [ERROR]: Cannot find file /vsicurl/http://larryfire.files.wordpress.com/2009/07/untooned_jessicarabbit.jpg
grdinfo [ERROR]: Must specify one or more input files
```
 but gdalinfo works

```
 gdalinfo /vsicurl/http://larryfire.files.wordpress.com/2009/07/untooned_jessicarabbit.jpg
Driver: JPEG/JPEG JFIF
Files: /vsicurl/http://larryfire.files.wordpress.com/2009/07/untooned_jessicarabbit.jpg
Size is 1097, 1024
Metadata:
  EXIF_ColorSpace=65535
  EXIF_DateTime=2008:04:19 11:26:14
...
```

PaulWessel:
We need to route the /vsi*/ URLS to GDAL and the new gmt_remote.c changes failed to do that.  This PR closes #3425 and #3424.


src/gmt_grdio.c
```diff
@@ -3000,18 +3000,18 @@ void gmt_grd_flip_vertical (void *gridp, const unsigned n_cols32, const unsigned
...
 
-bool gmtlib_check_url_name (char *fname) {
+bool gmtlib_found_url_for_gdal (char *fname) {
 	/* File names starting as below should not be tested for existence or reading permissions as they
-	   are either meant to be accessed on the fly (http & ftp) or they are compressed. So, if any of
-	   the conditions holds true, returns true. All cases are read via GDAL support or other. */
-	if (gmt_M_file_is_url (fname) ||
+	   are either meant to be read via GDAL.  Regular URLs (e.g., htthp://, ftp:// are read via libcurl. */
+	if (!strncmp(fname,"WCS:", 4)  ||
+		!strncmp(fname,"WMS:", 4)  ||
```
joa-quim, line 3007: test only for /vsi

joa-quim: test also for
!strncmp(fname,"WCS:", 4)
!strncmp(fname,"WMS:",4)

src/gmt_macros.h
```diff
@@ -200,7 +200,7 @@
 
 /*! Determine if we have a special virtual, remote, or query files */
 #define gmt_M_file_is_cache(file) (file != NULL && file[0] == '@' && strncmp (file, "@GMTAPI@-", 9U))
-#define gmt_M_file_is_url(file) (file != NULL && (!strncmp (file, "http:", 5U) || !strncmp (file, "https:", 6U) || !strncmp (file, "ftp:", 4U)))
+#define gmt_M_file_is_url(file) (file != NULL && (!strncmp (file, "http:", 5U) || !strncmp (file, "https:", 6U) || !strncmp (file, "ftp:", 4U) || !strncmp (file, "/vsi", 4U)))
```
joa-quim, line 203: Test also for WCS & WMS

PaulWessel: We can continue this tomorrow. file_is_url are looking for regular URLs.  The /vsi and WCS are not regular URLs as I understand it - they need to go via GDAL

src/gmt_grdio.c
PaulWessel: ?? You mean just check for the 4 characters /vsi instead of the list?
And I thought I am checking wor WCS and WMS here.

src/gmt_macros.h
joa-quim: But if you also are testing for "/vsi" it's not regular URLs anymore.

src/gmt_grdio.c
joa-quim: Yes, the 4 /vsi chars. There are lots of vsi drivers now

PaulWessel: OK, now just checking for /vsi 4 chars

src/gmt_macros.h
PaulWessel: I am confused.  Are you saying that addresses that starts with WCS: or WMS: are regular URLs and we shall pass them to our libcurl download call?  I thought these were special things that must go via GDAL and pertaining to grids.  If I take those out from this function then should we be given that type of file it will end up in our libcurl call.  Please clarify.

joa-quim: No. I'm saying that if you test for /vsi (and you are), which are GDAL only, than you should test also for WCS/WMS that are GDAL only too.

PaulWessel: Great, that is what I thought.  The revised functions does all that:

```
bool gmtlib_found_url_for_gdal (char *fname) {
	/* File names starting as below should not be tested for existence or reading permissions as they
	   are either meant to be read via GDAL.  Regular URLs (e.g., htthp://, ftp:// are read via libcurl. */
	if (!strncmp(fname,"WCS:", 4)  ||
		!strncmp(fname,"WMS:", 4)  ||
	    !strncmp(fname,"/vsi", 4)) {
#ifdef WIN32
		/* On Windows libcurl does not care about the certificates file (see https://github.com/curl/curl/issues/1538)
		   and would fail, so no choice but prevent certificates verification.
		   However, a CURL_CA_BUNDLE=/path/to/curl-ca-bundle.crt will overcome this and will consult the crt file. */
		if (!getenv ("GDAL_HTTP_UNSAFESSL") && !getenv("CURL_CA_BUNDLE"))
			putenv ("GDAL_HTTP_UNSAFESSL=YES");
#endif
		return true;
	}
	else
		return false;
}
```

Good to go?

joa-quim: We must be talking different things. The ``gmtlib_found_url_for_gdal `` is fine like that but I'm referring to the ``gmt_M_file_is_url`` macro that you have

#define gmt_M_file_is_url(file) (file != NULL && (!strncmp (file, "http:", 5U) || !strncmp (file, "https:", 6U) || !strncmp (file, "ftp:", 4U) || !strncmp (file, "/vsi", 4U)))

PaulWessel:
OK, several changes implemented.  Many files affected mostly because of macro name changes:

1. _gmtlib_found_url_for_gdal_ only returns true if the file is one of the types that only GDAL can open (/vis.../, map server stuff) and not regular URLs.
2. New function _gmt_file_is_cache_ returns true if the file is one of our cache files (or pretends to be).  This could no longer be a macro since our datasets require a function to check.
3. New macro _gmt_M_file_is_remote_ is true if either a cache or a remote dataset is given.  We need both of these tests since there are places where either types are OK and other places (e.g., reading remote CPT) where a remote dataset would be wrong.

My tests passes, except gdal_nn.sh which always fails when run in parallel (ctest) and passes if run manually...

PaulWessel:
After a PS update, this PR seems to work fine for me.  What you think @joa-quim ?  Any gdal-urls are passed to the module as is and ends up with grd/img reads while any other type URLs survive into our libcurl jobs.  I've gone over all the instances where the old macros and now new functions are called to make sure we check for the right things.

PR #3438, opened 2020-06-07
Fix the duplication and freeing of GMT_IMAGE

PaulWessel:
Two main problems fixed, but the plotting is still not correct:

1. **GMT_IMAGE** members colormap and color_interp were not duplicated by _GMT_Duplicate_Data_.
2. **grdimage** projected I to Img_proj and deleted I, but then refered to I->colormap instead of Img_proj->colormap.

However, the colormap entries we do obtain in gmt_gdalread.c do not match the ones **gdalinfo** reports on the command-line and the image looks like shit.  See #3436 for more information.


PaulWessel:
I think this is now all working (at least for me on macOS locally).  Need to be tested on other OS as well.  Here are the changes:

1. Added new mode bit-flag GMT_IMAGE_NO_INDEX which can be passed into GMT_Read_Data when images are read.  If set and the image is an indexed image, then we expand to rgb using the currently selected memory layout.  This is needed because different modules work with different default layouts: grimage wants RGB[A]RGB[A]RGB... while grdmix expects RRRR...GGGG.BBBB.
2. Updated a few scripts and orig PS.  The images we use are slightly different from the original marbles, so the PS must change.  Also, some scripts did a forced registration changed that no longer is needed.
3. Implements the corrections needed in GMT_Duplicate_Data for images that we discussed yesterday.

This PR closes #3436.

PaulWessel:
I know @joa-quim said he may be off this evening but we probably want to wait to merge until he gets a test in.  Hopefully @seisman can give it a spin in the CI.

joa-quim:
It works but prints that annoying and false warning 3 three times and mixes ERROR and WARNING
```
gmt grdimage -JG0/0/3i @earth_day_01d -B -pdf map
grdimage [NOTICE]: Remote data courtesy of GMT data server OCEANIA [https://oceania.generic-mapping-tools.org]

grdimage [NOTICE]: Earth Daytime Image at 1x1 arc degrees [NASA image mosaic - Blue Marble].
grdimage [NOTICE]:   -> Download grid file [44K]: earth_day_01d_p.tif
grdimage [ERROR]: Illegal code [grdimage [ERROR]: Illegal code [grdimage [ERROR]: Illegal code [grdimage [WARNING]: The image memory layout (TRP ) is of a wrong type. It should be BRPa.
```

PaulWessel:
But I deal with that.... OK will check.

PaulWessel:
Try now.

PaulWessel:
I think grdimage.c can loose some lines.  If I am correct, @joa-quim, given that indexed images ahve been replaced earlier, we only either have gray or RGB[A] images in the loops, and we only need to copy the R[GB] into the rgb array for 1 or 3 values.  So the business with r_table, g_table, b_table is no longer needed I think

PR #3469, opened 2020-06-12
When images have 4 layers (with alpha), allocate an array with 4 and not 3 as it used to be.

joa-quim:
Fixes reading RGBA pix interleaved images.


joa-quim:
I did but we really do not have tests for this.

seisman:
We should add a test if possible.

PR #3472, opened 2020-06-14
Use list of keywords instead of bit-sum for MAP_ANNOT_OBLIQUE

PaulWessel:
Since not even I can remember what the hell **MAP_ANNOT_OBLIQUE**=34 might mean without reading the settings web page every single time, I finally broke down and replaced this bit value with a comma-separated list of understandable keywords.  Old integer values are of course accepted via backwards compatibility.  All scripts using **MAP_ANNOT_OBLIQUE** have been updated to the new syntax (e.g., --**MAP_ANNOT_OBLIQUE**=**separate**,**lat_horizontal**).  The list of keyword arguments are (with corresponding integer bit value in brackets):

- **separate** [0], let longitudes be annotated only on NS axes and latitudes on WE axes.
- **anywhere** [1], let longitudes and latitudes be annotated wherever an invisible gridline crosses the map boundary [Default].
- **lon_horizontal** [2], let all longitudes be typeset horizontally.
- **lat_horizontal** [4], let all latitudes be typeset horizontally
- **tick_extend** [8], let oblique tick lengths be extended so the .tip's distance from the frame equals the requested tick length.
- **tick_normal** [16] let ticks be plotted normal to frame regardless of gridline angle.
- **lat_parallel** [32] let latitudes be annotated parallel to the frame (this is not the same as **lat_horizontal**).

All tests still pass after the changes.

### DIFF
--- a/doc/examples/ex02/ex02.bat
+++ b/doc/examples/ex02/ex02.bat
@@ -4,7 +4,7 @@ REM Purpose:	Make two color images based gridded data
 REM GMT modules:	set, grd2cpt, grdimage, makecpt, colorbar, subplot
 REM
 gmt begin ex02
-	gmt set MAP_ANNOT_OBLIQUE 0
+	gmt set MAP_ANNOT_OBLIQUE separate
 	gmt subplot begin 2x1 -A+JTL -Fs16c/9c -M0 -R160/20/220/30+r -JOc190/25.5/292/69/16c -B10 -T"H@#awaiian@# T@#opo and @#G@#eoid@#"
 		gmt subplot set 0,0 -Ce3c
 		gmt grd2cpt @HI_topo_02.nc -Crelief -Z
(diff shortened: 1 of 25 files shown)


anbj:
It was confusing, but it was also a bit (pun intended) satisfying when you after 22 attempts got it right. You felt.. smart.

PaulWessel:
True, and it was a rite of passage for new users to know what 7 meant.  But, as my brain's buffer is shrinking and can only hold so many bytes for internal look-up tables, this is an area where we can help the old man.

However, you could still put a piece of duct-tape on the screen covering up the new explanations in gmt.conf.rst and go by rote memory from the old system and hence still feel smart.

anbj:
Gotta get me some duct-tape.

joa-quim:
Late but 💯 😃

PR #3475, opened 2020-06-14
Reset GMT->current.gdal_read|write_in|out stuff at end module

joa-quim:
Needed from externals


joa-quim:
> Does this fix the grdimage "failure"

Nope, not related. It's about *mem_layout* not being inherited between calls

PR #3490, opened 2020-06-18
Keep only one commit in the gh-pages branch

seisman:
The gh-pages branch contains the GMT documentations. We don't care about
the history of this branch, so we can keep only one commit in this
branch.


seisman:
After this PR was merged, a fresh clone of this repository has a size of 880Mb, compared to ~1.1 GB before.

PR #3493, opened 2020-06-18
gmtregress update to limit angles considered

PaulWessel:

Letting **-A** have power to limit which angles are used in determining the best regression.  This is helpful when you know from other information that a slope must be positive, yet outliers may trick it to find a negative solution that is not physically meaningful.  Using **-A+fp** forces a positive angle, for instance (**+fn** does the same for negative slopes), while a full -A30/66/1+f forces us to only use the given range.  This feature was requested by Dietmar Müller and I agree it is a useful addition.  So now **-A** has two different roles:

1. Scan all angles given and return misfit as function of angle
2. Limit angles used for robust determinations of slope

I added a new figure in doc that is now included in gmtregress.rst to show the purpose of the **-A** scanning feature. 

This PR also fixed a problem with the LS orthogonal solution when weights were either 0 or 1.  Now, if that is the case, I simply collect all the points with weight one and do a non-weighted orthogonal regression on those points instead.

### DIFF
--- a/src/gmtregress.c
+++ b/src/gmtregress.c
@@ -621,7 +621,8 @@ GMT_LOCAL double gmtregress_demeaning (struct GMT_CTRL *GMT, double *X, double *
 		if (beta && alpha) {	/* Compute beta (as alpha above) which is needed for weighted orthogonal regression */
 			for (i = 0; i < n; i++) {
 				if (w[GMT_Z]) corr_i = w[GMT_Z][i];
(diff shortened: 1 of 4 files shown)


src/gmtregress.c
```diff
@@ -70,8 +70,9 @@ struct GMTREGRESS_CTRL {
 		bool active;
 		char *file;
 	} Out;
-	struct GMTREGRESS_A {	/* 	-A<min>/<max>/<inc> */
+	struct GMTREGRESS_A {	/* 	-A[<min>/<max>/<inc>][+f] */
```
seisman, line 73:
```suggestion
	struct GMTREGRESS_A {	/* 	-A[<min>/<max>/<inc>][+f[n|p]] */
```

Issue #3498, opened 2020-06-19
Avoid environmental variables in movie


PaulWessel:
**Description of the desired feature**

The dark magic that handles movie labels and progress indicators relies on setting parameters in the environment (e.g., via export PAR=value) and then obtaining it from the environment (via _getenv_ system calls).  This issue discusses another mechanism that would avoid changing the environment completely.  To understand why it was done that way you will need to read the comments added via #3497. A motivation to avoid changing the environment is that there may be situations where it is not allowed, and I have had cases where it seems the settings have survived the ctest step and now pollutes my work terminal environment (ctest runs 24 movie frames at once and something odd happens at times).  Best to avoid this situation.

Every movie_frame.sh|csh|bat script created by movie starts with a call to **gmt figure**, e.g.,

`gmt figure ../${MOVIE_NAME} png,ps E${MOVIE_DPU},A+n+r${MOVIE_BACKGROUND}`

This line is created by **movie** and inserted before the users actual frame plot code.  It defines the plot name, format, resolution, any background color, etc.  My proposal is to add an undocumented option to **figure** called **-I**_frame_.  This option is only for the purpose of **movie** and will convert movie label and progress indicator information from comments in the parameter file to data files in the current session directory.  The label/indicator info would there simply be written as comments in the parameter file and **gmt figure** **-I** would read that file, strip out the relevant parameters, and write the expected files _gmt.movie_labels_ and _gmt.movie_progress_indicators_ to the session directory.  After that, GMT takes over using this information to plot the labels and indicators. In the example above, **movie** would instead write

`gmt figure ../${MOVIE_NAME} png,ps E${MOVIE_DPU},A+n+r${MOVIE_BACKGROUND} -I$1`

which when running for frame 67 of 500 would be -I067.  That is all **figure** needs to look at the file _movie_params_067.sh_ and extract the information (if any) [the $1 and .sh if for Bourne shells, other symbols are used for batch, for instance).

The **-I** would be undocumented because it is only needed in this particular situation.  Setting and obtaining environmental parameters is not always safe, while this method would be, and faster.

seisman:
I don't have good suggestions here.

If you're going to implement and use the undocumented **-I** option, please also add a comment after the line 
```
gmt figure ../${MOVIE_NAME} png,ps E${MOVIE_DPU},A+n+r${MOVIE_BACKGROUND} -I$1
```
emphasizing that **gmt figure -I** is undocumented and for internal use only.

[...]

PR #3499, opened 2020-06-19
Movie no env setting

PaulWessel:
This PR implements the ideas of #3498.  By avoiding using the environment to pass information between GMT modules we avoid potentials contamination of the environment and relying on _getenv_ to obtain parameters. All tests still work.


PaulWessel:
This PR does not change the code that @joa-quim is having issues with, but it avoids the entire setting and getting environmental values by using a safer read/write from/to files directly.  Unless you see something odd please approve - all test pass for me, including one long movie iwth lables.

PaulWessel:
And there it goes.

PR #3501, opened 2020-06-19
Movie fixes

PaulWessel:

Perhaps this is related: I am counting 15 vertical bars in that format but we only replace 13 of them with spaces.  So that does not seem right.  I also now keep the count of converted items so we can report it if failing.  Please see if this helps, @joa-quim .


PaulWessel:
Now without environmental interference.

PaulWessel:
Need your OK, @joa-quim

PR #3507, opened 2020-06-20
Just use a system call to wipe the entire sessions dir

### LINKED ISSUE #3476: gmt clear fails on Win if session dirs are not empty
```
gmt clear sessions
c:/j/.gmt/sessions: Directory not empty
```

PaulWessel:
No safety issues since this directory was created by GMT. See #3476 for discussion. It does not matter if the user has set their **USER_DIR** to any area, since we _only_ delete the subdirectory _sessions_ inside that directory.  GMT created it, so GMT can delete it.  Closes #3476.

### DIFF
--- a/src/clear.c
+++ b/src/clear.c
@@ -157,7 +157,7 @@ static int clear_sessions (struct GMTAPI_CTRL *API) {
 		GMT_Report (API, GMT_MSG_INFORMATION, "No directory named %s\n", API->session_dir);
 		return GMT_FILE_NOT_FOUND;
 	}
-#ifdef WIN32
+#ifdef _WIN32
 	sprintf (del_cmd, "rmdir /s /q %s", API->session_dir);
 #else
 	sprintf (del_cmd, "rm -rf %s", API->session_dir);
--- a/src/clear.c
+++ b/src/clear.c
@@ -151,35 +151,23 @@ static int clear_data (struct GMTAPI_CTRL *API, char *planet) {
 }
 
 static int clear_sessions (struct GMTAPI_CTRL *API) {
-	unsigned int n_dirs, k;
-	char **dirlist = NULL, *here = NULL;
+	int error;
+	char del_cmd[PATH_MAX] = {""};
 	if (access (API->session_dir, F_OK)) {
 		GMT_Report (API, GMT_MSG_INFORMATION, "No directory named %s\n", API->session_dir);
 		return GMT_FILE_NOT_FOUND;
 	}
-	if ((here = getcwd (NULL, 0)) == NULL) {	/* Get the current directory */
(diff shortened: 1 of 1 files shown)


src/clear.c
```diff
@@ -151,35 +151,23 @@ static int clear_data (struct GMTAPI_CTRL *API, char *planet) {
...
-	char **dirlist = NULL, *here = NULL;
+	int error;
+	char del_cmd[PATH_MAX] = {""};
 	if (access (API->session_dir, F_OK)) {
 		GMT_Report (API, GMT_MSG_INFORMATION, "No directory named %s\n", API->session_dir);
 		return GMT_FILE_NOT_FOUND;
 	}
-	if ((here = getcwd (NULL, 0)) == NULL) {	/* Get the current directory */
-		GMT_Report (API, GMT_MSG_ERROR, "Cannot determine current directory!\n");
+#ifdef WIN32
```
joa-quim, line 160: My only doubt is this bloody macro. Never clearly understood the difference for ``_WIN32`` but I think thing both MinGW and VS define it, whilst Cygwin has a ``WIN32``. For playing safe I would prefer it to be ``_WIN32``

PaulWessel:
@joa-quim may wish to check that we are consistently using WIN32 vs _WIN32.  I believe _WIN32 is set by the MicroSoft compiler only and hence it is pure Windows, while WIN32 may be set under both Windows and Cygwin.  So pure Windows API calls should be _WIN32.

joa-quim:
Not that obvious. MinGW defines ``_WIN32`` too. The big problem is that we never built/tested MinGW and Cygwin builds. When we want to restrict to VS we should is use ``_MSC_VER`` but then it also has versions.

PaulWessel:
This says things starting with underscore is for compilers providers: https://stackoverflow.com/questions/662084/whats-the-difference-between-the-win32-and-win32-defines-in-c

joa-quim:
I was reading that same post

PaulWessel:
Well, price we pay for supporting a non-Posix OS!
OK to merge this?

seisman:
> The big problem is that we never built/tested MinGW and Cygwin builds. 

Azure Pipepines already have MinGW installed. It seems straightforward to test GMT with MinGW.

joa-quim:
... still don't understand what's so f different between them.

joa-quim:
> Azure Pipepines already have MinGW installed. It seems straightforward to test GMT with MinGW.

That is right, but I really don't feel like start maintaining another build system.

seisman:
It seems there are very few people using Cygwin.

joa-quim:
Yes, and there even less (0 to my knowledge) BUILDING (try at least) in Cygwin.

PaulWessel:
still need to get go-ahead to merge though.

joa-quim:
You have two (but need none)

PR #3508, opened 2020-06-20
Forgot to update the theme in gmtinit_conf_modern_override

PaulWessel:
This function is called when modern theme is selected in classic or always in modern.  But it failed to actually set the **GMT_THEME** variable so in classic mode it reverted to classic before writing out the theme.


PaulWessel:
This PR makes it possible to use the modern theme in classic:

```
gmt set GMT_THEME modern
gmt psbasemap -R0/30/0/30 -JM4i -Baf -P > t.ps
```

or

`gmt psbasemap -R0/30/0/30 -JM4i -Baf -P --GMT_THEME=modern > t.ps`

PR #3510, opened 2020-06-20
Improve the two-step grid reading when source is a matrix

### LINKED ISSUE #3408: grdimage shading fails with external matrix
The issue was first reported by @liamtoney in https://github.com/GenericMappingTools/pygmt/issues/364. Using the PyGMT script in that issue, he got the following error:
```
```

After some debugging, it seems that GMT can correctly read the matrix (i.e., the earth relief data) and can plot it correctly if **shading=False** (i.e., no **-I+d** in the command line). However, if **shading=True** is used (i.e., grdimage -I+d), GMT reports the error above, possibly because the matrix in memory is modified/freed/re-initialized when calling `grdgradient`.

---

The issue can be partially reproduced using the following C codes. The C code below is modified from GMT's own test code `testapi_matrix_plot.c`. It reads a 2d matrix and plots the matrix using grdimage.
```C
#include "gmt.h"

int main () {
	void *API = NULL;                /* The API control structure */
	struct GMT_MATRIX *M = NULL;    /* Structure to hold input matrix */
	char input[GMT_VF_LEN] = {""};	/* String to hold virtual input filename */
	char args[128] = {""};         	/* String to hold module command arguments */
	/* Initialize the GMT session */
	M = GMT_Read_Data (API, GMT_IS_MATRIX, GMT_IS_FILE, GMT_IS_SURFACE, GMT_READ_NORMAL, NULL, "2d-matrix.txt", NULL);
	/* Associate our matrix with a virtual file */
	GMT_Open_VirtualFile (API, GMT_IS_GRID|GMT_VIA_MATRIX, GMT_IS_SURFACE, GMT_IN, M, input);

	/* Prepare the module arguments */
	sprintf (args, "%s -JX6i -P -Baf -Cgeo", input);
	/* Call the grdimage module */

	/* Close the virtual file */
	GMT_Close_VirtualFile (API, input);
	/* Destroy session */
	if (GMT_Destroy_Session (API)) return EXIT_FAILURE;
};
```

The data `2d-matrix.txt` is the 60m earth relief data, converted from netCDF to matrix format using following Python code. The data is attached [2d-matrix.txt](https://github.com/GenericMappingTools/gmt/files/4703975/2d-matrix.txt).

```python
import numpy as np
import pygmt
er = pygmt.datasets.load_earth_relief("60m")
np.savetxt("2d-matrix.txt", np.flip(er.data, axis=0), fmt='%.1f')
```

The C code is compiled using
```
gcc testapi_matrix_plot.c $(gmt-config --cflags --libs)
```

The output image looks correct. The coordinates are incorrect, simply because I didn't pass the geographic information.

image

Changing the grdimage arguments to:
```
sprintf (args, "%s -JX6i -P -Baf -Cgeo -I+d", input);
```
to enable shading, then the output image is:

image

This is what I expect:

image

PaulWessel:

See #3408 for the initial problem.  Here is what goes wrong:

1. We first read the container header from a file.  The file is a memory reference to a matrix.  This returns a grid structure with a header.
2. Then grdgradient is called with the same file to make an intensity grid.  This all works.
3. Later, we wish to read the data from the matrix source.  However, despite me implementing more meaningful memory file names like @GMTAPI@-S-I-G-M-G-N-000000, the old code has not been updated: It loops over the registered objects and finds the first one that matches the specs (image, input, etc). This turns out to be the grid header from (1) above instead of item 0.  So we go down the rabbit hole and do strange things.

Time to implement a new scheme that extracts the information directly from the filename.

Closes #3408.


PaulWessel:
I have made a few changes.  First, I am parsing the full name and get the various items.  For now I am only checking direction and return the object ID.  I also made a small change that when we check for the validity of a given object I ignore if it has been used before (it has).  I added @seisman example C code as well and two tests (with and without shading) which both passes.  All other tests pass as well.  There are more simplifications that can be done in gmt_api.c but I want to do that very carefully...

PaulWessel:
Closes #3408.

PR #3514, opened 2020-06-21
Fix the grdinfo crash for external grids

seisman:
@PaulWessel 

I changed the C code `testapi_matrix_as_grid.c` to pass an external grid to `grdinfo`. Now it crashes for me. 

Is it another API bug?


PaulWessel:
See if this does it for you, @seisman.

PaulWessel:
While at it, see if your vector strings pstext still fails after this fix.

seisman:
Your commit fixes the grdinfo crash, but doesn't work for the vector-string case.

seisman:
@PaulWessel I added two grdinfo calls in testapi_matrix_as_grid.c to make sure that it won't crash in the future. 

I think we can merge the PR into master, so that you can work on the pstext issue.

PR #3516, opened 2020-06-21
Add a API test for checking passing vectors and strings to pstext

seisman:
Here is a C code which passes *x, y, angle, justify, text* to pstext. It should be equivalent to the following bash script:
```
gmt text -JX10c -R0/10/0/10 -Baf -F+a+j -Vd > testapi_vector_strings.ps << EOF
5.0 3.0 30.0 ML First Label
5.0 8.0 60.0 MR Second Label
EOF
```

However, it seems the *angle* is not correctly passed. 

**Output of the C code**

image

**Output of the bash script:**

image

Anything wrong?

### DIFF
--- a/src/testapi_vector_strings.c
+++ b/src/testapi_vector_strings.c
@@ -1,34 +1,41 @@
 #include "gmt.h"
 /*
- * Testing a specific case where we wish to pass and receive a GMT_VECTOR or GMT_MATRIX
- * to/from a module that expect to read/write GMT_DATASETs.
+ * Testing the use of user data provided via a GMT_VECTOR to pstext,
+ * passing both numerical vectors and a string array.
  */
+
+/* Dimensions of the test dataset */
+#define NCOLS 3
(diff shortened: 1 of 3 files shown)


PaulWessel:
Problem was your dim[0] = 2 but you have three columns.  And _GMT_Put_Vector_ returns an error but not checked by main.  It does not print a message.  We have lots of places where we don't print a message but return error code.  Maybe we should always print an error?

seisman:
> Problem was your dim[0] = 2 but you have three columns.

Yes, changing it to 3 works!

> We have lots of places where we don't print a message but return error code. Maybe we should always print an error?

Not sure whether it's necessary, but at least PyGMT checks the return error code carefully.

PaulWessel:
OK, maybe commit your dim edit and I can approve this merge?

seisman:
Will do it later and also add a test.

seisman:
@PaulWessel Done.

PR #3519, opened 2020-06-22
Add testapi wrapper so that gmtest can find them on Windows

seisman:
Following tests fail on Windows:
```
	208 - api/apigrdpix.sh (Failed)
	211 - api/apimatimageplain.sh (Failed)
	212 - api/apimatimageshading.sh (Failed)
	216 - api/apiuserstrings.sh (Failed)
```
These tests can't find the testapi_* commands, because these new
testapi_* commands are not wrapped.

### DIFF
--- a/test/gmtest.in
+++ b/test/gmtest.in
@@ -82,7 +82,10 @@ for apiprog in \
     testapi_uservectors \
     testapi_vector \
     testapiconv \
-    testgrdio
+    testgrdio \
+    testapi_imageshading \
+    testapi_matrix_as_grid \
+    testapi_vector_strings
 do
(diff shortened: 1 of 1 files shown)


seisman:
> How can they work on Win if they depend on a executable built with gcc?

I don't understand it. These executables are built with MSVC in the Windows CI. Do you mean you're building them with gcc on your laptop?

joa-quim:
No, ofc not. And yes I have many ``testapi_*.exe`` but I was confused with the failures of 
```
testapi_uservectors
testapi_userdataset
```
they both crash. I have other crashes wen running the tests but simply have no time to chase them all. Those crashes occur only in debug builds and past has shown that they are due to *little* memory corruptions that either gcc or c-lang let go but not VS.

seisman:
Yes, these two also crash in our Windows CI jobs for a long time.

Issue #3520, opened 2020-06-22
-R auto in subplot is not per pannel. 


joa-quim:
Running this shows two issues.
1- In first panel -R is auto computed but that same -R is applied to second panel
2- minor issue, the -R estimate is an exact computation, *tight* in Matlab parlance but that is not very nice. Compare with what is obtained with the Julia wrapper
```
gmt begin tales png
	gmt subplot begin 2x1 -Fs8c/4c
		gmt plot ds1.dat -W0.4 -Ba -c0,0
		gmt plot ds2.dat -W0.5 -Ba -c1,0
	gmt subplot end
gmt end show
```

[data.zip](https://github.com/GenericMappingTools/gmt/files/4814978/data.zip)
![tales](https://user-images.githubusercontent.com/537321/85323104-57063800-b4bf-11ea-8eae-4d2e9b0524d8.png)

the julia version (more things but the essential are the limits)
![GMTplot](https://user-images.githubusercontent.com/537321/85323117-5bcaec00-b4bf-11ea-92ef-1993b361796d.png)

PaulWessel:
So what should our rule be here: Clearly, in panels a data-derived -R should only apply to that panel.  So I guess the same should be true for figures?  If you start off with a plot call with no -R you get a -R from the data.  Subsequent overlays must inherit that one of course.  But a new figure should not.  So I think that is a clear rule, yes?
We have both exact (your tight) and approx options for this.  Not sure what the default is (your example suggests exact), but we could reconsider that.

joa-quim:
Yes, each figure should get its own -R.
Approx -R looks nicer then tight.

PaulWessel:
Seems like the approx R should go with datasets and the exact R goes with grids , unless the user uses -R

joa-quim:
Agreed.

PaulWessel:
Hi @joa-quim and @seisman: To fix this problem I am implementing the same hierarchy of gmt.history files as we do for cpts.  So a panel will create its own gmt.history file, etc. but will read from a higher hierarchical level if not found.  When `subplot end `is called I wipe all the subplot/panel histories.  Nice and clean.

Then I ran the tests and had a a dozen or so failures.  Have not looked at them all but looks like they are similar to this one (panels.sh) that fails:

```
gmt begin panels ps
  gmt subplot begin 2x2 -Fs3i -M5p -A -SCb -SRl -Bwstr
    gmt basemap -R0/80/0/10
    gmt basemap -c
[...]

PR #3344, opened 2020-05-17
Implement GMT themes (sets of GMT defaults)

PaulWessel:
**Description of proposed changes**

This PR follows up on #2290 and adds the new GMT defaults parameter **GMT_THEME**.  A theme is simply a _collection_ of GMT default parameter settings that make sense together and have been given a suitable name.  GMT comes with a few built-in themes:

- _classic_: The basic GMT defaults as distributed
- _modern_: Currently same as classic (but see #2290 for ideas)
- _inverse_: White pen/fonts on black background
- _darkmode_: Light-gray pens/fonts on very dark-blue background

The last two are of course experimental, and to implement _modern_ (#2290) requires new coding as well.  Users can add their own themes as theme.conf files in ~/.gmt.

The theme machinery has been implemented and working.  For instance,

```
gmt coast -W -RFR -B -pdf map1 --GMT_THEME=inverse
gmt coast -W -RIS -B -Gturquoise -pdf map1 --GMT_THEME=darkmode
```

Remaking tasks:

1. Better define a cool _darkmode_ set of parameters
2. Implement the font and dimensions scalings discussed in #2290.

PR #3523, opened 2020-06-23
Let gmt.history and gmt.conf be hierarchical and separate for panels and figures

PaulWessel:
We do not want an automatic region determined for one panel to apply to another panel.
This addresses #3520 but still needs some more work.  It also extends this idea to gmt.conf which now is also hierarchical.  Hence, current cpt, settings and history are now all hierachical.  This leaves us with ~12 failures due to syntax that is now incorrect.


PaulWessel:
Still to do:

1. Update the 11 failing scripts to follow the new hierarchical system
2. Update the documentation to extend the hierarchical explanation to settings and history as well as cpt.

I suggest @joa-quim and @seisman try this branch,  I have tested it on things like this and the levels work. If you run one command at the time and monitor what is happening in the .gmt/sessions/gmt_session.* directory you will see many gmt.conf and gmt.history files now come and go at the right times.  The example below sets a figure-level title font (not used), then a subplot-level font (used for all panels), except overridden for one panel.  One panel also overrides the interpolant to be cubic but it is back to being Akima at a later panel:

#!/usr/bin/env bash
cat << EOF > t.txt
0 0
5 7
10  5
40  3
EOF
```
gmt begin
gmt figure map
  gmt set FONT_TITLE Helvetica-Bold
  gmt subplot begin 2x2 -Fs3i -M5p -A -SCb -SRl -Bwstr -R0/80/0/10
    gmt set FONT_ANNOT_PRIMARY=Times-Roman
    gmt basemap
    gmt basemap -c
    gmt set GMT_INTERPOLANT cubic
    gmt sample1d -I1 t.txt | gmt plot -W1p 
    gmt basemap -c
    gmt basemap -c --FONT_ANNOT_PRIMARY=Helvetica-Bold
   gmt subplot end
   gmt sample1d -I1 t.txt | gmt plot -W1p 
gmt end show

```

joa-quim:
Hm, this is what I get (no one-one command checking sorry). 
And, the Times-Roman looks so much nicer in annotations then the AvantGard

map

PaulWessel:
I can tell we have not converged on a modern font selection.  And @KristofKoch has gone quiet and not here to defend it.  I will see what I need to do to strengthen the implementation but it seems straightforward.  Oh, I cut/paste the last plot call **after** subplot end!  I will edit the script next.

PaulWessel:
Script updated.  I will work on documentation for all of this and the 11 failing tests due to "wrong" syntax.

joa-quim:
Still not filling the entire panel.
map

PaulWessel:
Data stops at 40, no?

joa-quim:
Yes, so it seems. Had not looked at it since first run.

KristofKoch:
> And, the Times-Roman looks so much nicer in annotations then the AvantGard

@joa-quim beauty is in the eye of the beholder. What makes you prefer the Times-Roman font? I wrote [a lengthy comment over at 3344](https://github.com/GenericMappingTools/gmt/pull/3344#issuecomment-629817436) why I prefer a sans-serif font and backed it with the NASA Technical Report ["On the typography of flight-deck documentation"](https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19930010781.pdf); especially chapter 3.2 "Typeface (Fonts)". 

> Several researchers have reported that when other typographical factors are controlled, sans-serif fonts are more legible than roman.

Source: chapter 3.2 "Typeface (Fonts)", [NASA-CR-177605](https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19930010781.pdf)

As @PaulWessel was unhappy with Helvetica as the default GMT font (I still prefer it) the only other sans-serif option is the Avant Garde font family. This is if we stick to the default fonts. Maybe there is the possibility to include an open source font with GMT which appeals to all/most of us?

 Paul: I'm still alive - but the current COVID19 situation ties up almost all of my resources.

PaulWessel:
I am down to two failures: ex16 and ex18.  What happens here is this sequence of events:

1. In both cases a subplot is defined.  Both subplot begin calls sets the plot region.
2. As part of panel preparations, we run commands like surface or grdmath.  Neither specifies the region.
3. While plotters can use data regions if no plot region is found, we decided that producers **cannot** use plot regions if no data region is given.
4. So surface, grdmath complain they have no region.

So, as usual, there are options to choose from to fix this:

1. We can backtrack on the requirement that producers get a data **-R** and use a plot region if no data region is set.
2. We can add specific region setting to these and even move the calls outside of subplot so that they can share the **-R** setting (even if I added a full **-R** to one surface call in a panel, the next panel would also need a **-R** setting since they are isolated now.

This is not entirely black and white to me.   In those examples, we do have plot region = data region so we would want a shared region.  However, I can imagine entirely different cases where a surface call inadvertently picks up a **-Rg** plot region and spends forever.  Here is an example from the (still pending) NASA proposal: We plan to let plotters be allowed to take input files like @earth_relief (i.e., no resolution specified) and we let GMT figure out what is the best given plot size.  But we will not let producers have that capability.  What the hell would grdfft @earth_relief be expected to do, for instance? So those modules need resolution.  Same here I think:  Leaving the data region undefined is not good, and all it takes is to specify it outside the subplot and now both plot and data regions will be set.

 Agree?

joa-quim:
I'm lost. The (one) problem in ``ex18`` is the
```
gmt grdmath pratt.txt POINT SDIST = mask.nc -fg
```
line? What is the problem of it inherits the the **-R** from ``-R@AK_gulf_grav.nc`` in ``subplot begin``?
Moving this lines to before the ``subplot begin`` is one solution but it makes the script less readable since those lines move away from their context.

PaulWessel:
You are not lost - that is the line in ex18.  The problem is one of coding - we currently do not allow a RP history to serve as RG history.  However, one possibility is when a plot region is given via a data file (-R@AK_gulf_grav.nc, -Rmygrid.grd).  Since those are data domains perhaps that is a reasonable rule to add? It means determining if the -R argument in RP history is a file and not w/e/s/n.

joa-quim:
> @joa-quim beauty is in the eye of the beholder. What makes you prefer the Times-Roman font?

Well, @KristofKoch you are 100% right. It's in my eyes. I guess that I don't like ``ones`` being represented as ``T's`` with small arms and in which one of those arms is broken. Or ``zeros`` as squares with rounded corners.

PaulWessel:
Before I forget: We are please that @KristofKoch is still alive! And yes we understand very well what you are up against.

joa-quim:
>  It means determining if the -R argument in RP history is a file and not w/e/s/n.

You mean ``-R@AK_gulf_grav.nc`` would please ``grdmath`` but ``-R0/1/0/1`` not? That would be very confusing.

PaulWessel:
Yes it would.  Maybe this is such a frequently occurring situation that we should accept any **-R** given to subplot to serve both as plot region and data region unless overridden below.  I think this is the only situation where this could make sense.

PaulWessel:
Thinking some more about this, I think letting subplot -R set both data region and plot region makes sense. If you have another data region in mind in that context you can specify that separately on the modules before or after subplot begin.  What say you @seisman? I think that would add plenty of flexibility and avoid us changing those examples as well.

PaulWessel:
Please comment on this @joa-quim before it is bedtime.

joa-quim:
Yes, at this time I agree that subplot -R should set both data and plot regions ... but it has been a very intense pair of days so mind changes are allowable.

PaulWessel:
So that handles ex18, but not ex16: In that example, we are sharing the **-R** (now) for gridders and plotters, but in the first panel we run `gmt surface @Table_5_11.txt -I0.2 -Graws0.nc`, while in another panel we run `gmt surface @Table_5_11.txt -Graws5.nc -T0.5`.  The problem is that no **-I** is given to surface since the `-I0.2` is only visible to a panel.  So I think there is no way around this problem but to add the **-I** to those commands (-I is not really a common option anyway) or move the gridding to outside the start of the panels.  it would be the sampe problem with any **-r**.

PaulWessel:
Just moving the first surface call to before the first panel plot fixes the problem since now all grid settings (R, I, r) are shared at the subplot level.  I think this is good.  I will update subplot man page to explain how this works.

PaulWessel:
I have completed this upgrade.  After minor changes to some rouge scripts, all test pass as before minus the 2 known.  Implementing the hierarchical scheme for history and conf was useful and found a few problems in the C code as well as bad scripting.  Please check and approve, @joa-quim and @seisman .

seisman:
Just tried a little with this branch. I would say it works as I expect. 

Here are some comments:

1. Although we have ~800 tests, there are no tests which is designed for testing this feature. Perhaps you can add your example above as a test?
2. It's not clear to me what history is remembered. The [documentation](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#command-line-history) says only common options are remembered, but from your discussions above, it seems **-I** is also remembered.

PaulWessel:
Only common options are remembered, but we learned the hard way that for grids we kind of need not only region and registration, but also increment.  Since -I was never a common option (think -I in grdimage), I made grid increments behave that way in the history.
I will add that script as a test shortly.

PaulWessel:
Nice test of splines as well:

panels_history

test/modern/panels_history.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+# Test a 2x2 basemap matrix with different history settings
+# at different hierarchical levels, showing of splines
+#!/usr/bin/env bash
```
seisman, line 4:
```suggestion
```

PR #3524, opened 2020-06-23
Use GMT_SETTINGS_FILE parameter instead of hirdwire gmt.conf

PaulWessel:
We may feel the need to change the name of the settings file at some point - this makes it a bit simpler and matches the syntax for the history file and session names.


PaulWessel:
Currently gmt begin always checks for a gmt.conf in the search path (home, .gmt,cwd).  That makes it possible that a modern mode script will give different results depending on the presence of such files.  I think having a way for a clean start would be good.  Options:

1. Accept a gmt begin option to not look for outside gmt.conf files, e.g., **-C** for clean.
2. Let the syntax be gmt begin [**clean**] [_prefix_] [_format(s)_] [_psconvertoptions_] [-V[<level>]]
3. Have another begin-like module called begin-clean that does the same.
4. Do a better documentation to let people know that` gmt clear settings` as first command will do this.
5. Same, for `gmt set -Du|s`.
6. ???

seisman:
Options 1, 4, and 5 look good to me.

PaulWessel:
Not a super high priority for me, but what does @joa-quim think?

seisman:
Just opened the issue #3525 so that we won't forget it.

PR #3537, opened 2020-06-24
Enforce that no grid header inquiry passes a subset

PaulWessel:
When _GMT_Read_Data_ uses **GMT_CONTAINER_ONLY** as mode, we return the full header of the grid.  Hence, it makes no sense to pass a subset wesn argument other than NULL.  To discourage such confusing behavior we now issue an error if that happens.  In GMT, only one instance (in **grdmath**) called _GMT_Read_Data_ in that way, now fixed.  Al tests pass. Any externals will need to fix their calls if they ever abused the function.


joa-quim:
What if it breaks Julia? I had no time to test it.

PaulWessel:
Please check, I can always comment that out if it poses problems.  But presumably you are not doing the bad thing.  There is nothing like this in gmtmex.

joa-quim:
No side effects detected but most of my Julia tests test the syntax generation.

PaulWessel:
Good.  Do you do any GMT_Read_Data on any items from Julia?

joa-quim:
I have a wrapper to that function, but do not use it in any place (yet)

PR #3546, opened 2020-06-27
Add -Q to gmtget to simply list aviailble datasets

PaulWessel:
With **-Q**, no downloading takes place when **-D**  is given, and **-D** and **-I** can be used to limit what is shown.  We return one record with information about each dataset that passes the selection criteria.  Advanced users can use this option to script particular download requests via gmt which or gmt get.

### DIFF
--- a/doc/rst/source/gmtget.rst
+++ b/doc/rst/source/gmtget.rst
@@ -85,7 +85,7 @@ PARAMETER
     Can be used in conjunction with **-D** (and **-I**) to provide a listing of
     available datasets (no downloading takes place)  The output is one record per
     dataset giving the information as *planet group dataset size remark*.  For datasets
-    that are tiled, the *size* is set to N/A (tile sizes various but are usually just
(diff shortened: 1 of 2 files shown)


PaulWessel:
Would any other columns be useful, e.g., number of tiles [1]?

doc/rst/source/gmtget.rst
```diff
@@ -78,6 +79,15 @@ PARAMETER
     This speeds up the total data request and defers the conversion to when the
     tile is requested by a module.
 
+.. _-Q:
+
+**-Q**
+    Can be used in conjunction with **-D** (and **-I**) to provide a listing of
+    available datasets (no downloading takes place)  The output is one record per
```
seisman, line 86:
```suggestion
    available datasets (no downloading takes place). The output is one record per
```

seisman:
> Would any other columns be useful, e.g., number of tiles [1]?

Yes, sounds useful.

PaulWessel:
Dinner first...

PR #3550, opened 2020-06-28
Use GMT_IS_REFERENCE for outputs in general

PaulWessel:
This is the other shoe dropping.  Earlier, after fixing the virtual file system and dealing properly with **GMT_IS_REFERENCE** and **GMT_IS_DUPLICATE**, we realized most/all places inside GMT should use **GMT_IN**|**GMT_IS_REFERENCE** when passing data to other modules.   We can do this since we control the information.  Of course, there are plenty of places where we expect to return data from those modules, and prior to the fixes the **GMT_OUT** simply gave us **GMT_IS_REFERENCE** by default.  Now it takes **GMT_OUT**|**GMT_IS_REFERENCE** to simply pass out the internally allocated item.

Looks like all tests still pass. May need to check if all is will in pyGMT for this as well, @seisman.

### DIFF
--- a/doc/rst/source/api.rst
+++ b/doc/rst/source/api.rst
@@ -1688,7 +1688,8 @@ compatible with the data type used by the modules then we could use your array d
 This *may* have the side-effect that your input array is modified by the module, especially
 if the module writes the results to a netCDF grid file.
 If that is a price you are willing to pay then you can add GMT_IS_REFERENCE to the ``direction``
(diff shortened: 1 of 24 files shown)


seisman:
All are good for PyGMT, but PyGMT doesn't read outputs via GMT API.

PaulWessel:
All the testapi*.c functions are doing this successfully, so that means it works.  Again, this is (internally) how it used to be anyway: output via reference.

PR #3552, opened 2020-06-28
Make sure we only try to download input files

PaulWessel:
In _gmtapi_next_io_source_ we call _gmt_download_file_if_not_found_ wihtout checking if it is an input file first.  Not sure if this is causing any problems but now isolated inside an if-test.  Addresses the issue in #3532.


PaulWessel:
Did this prevent the problem you had with output.txt?

PR #3563, opened 2020-06-30
Add MVT (Mapbox Vector Tile) as another GDAL URL

joa-quim:


PaulWessel:
FYI, we are probably starting a freeze and building 6.1.0 release later today.

PR #3567, opened 2020-06-30
Fix type and change 6.0 to 6.x

PaulWessel:
May need to look at some of the Linux instructions in case they have changed, @seisman ?


seisman:
The other instructions look good to me.

PR #3570, opened 2020-07-01
Fix grdveiw's -I explanations

PaulWessel:
While **grdview** **-I** has the same syntax as **grdimage -I,** the documentation had not been changed and mentioned **-I+**. The parsing of arguments is the same in both modules, so this is just a documentation fix.  Closes #3570.


seisman:
> The parsing of arguments is the same in both modules, so this is just a documentation fix

Is it true?

grdimage uses the following code to detect **+d**,

https://github.com/GenericMappingTools/gmt/blob/c102aee4660acedfcf105c88a3a44a1acfa4acdd/src/grdimage.c#L353-L356

but I don't see the same code in grdview.

PaulWessel:
Should be the same treatment now?

PR #3572, opened 2020-07-02
Set backwards nx,ny to n_cols, n_rows in 6.1

PaulWessel:
Because of the ubuntus of the world, people are failing to build GMT and GMTSAR since we removed the backward nx,ny settings.  This reinstate those settings just for GRIDs since that is the only thing GMTSAR uses.  I figure we can eventually remove that when ubuntu 16 dies or the people using it dies or stop using GMTSAR.  This will help in the short term and will be in 6.1

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -5450,6 +5450,11 @@ GMT_LOCAL struct GMT_GRID * gmtapi_import_grid (struct GMTAPI_CTRL *API, int obj
 	if (done) S_obj->status = GMT_IS_USED;	/* Mark as read (unless we just got the header) */
 
 	return (G_obj);	/* Pass back out what we have so far */
+
+#ifdef GMT_BACKWARDS_API
+	G_obj->header->nx = G_obj->header->n_columns;
+	G_obj->header->ny = G_obj->header->n_rows;
+#endif
 }
 
 /*! Writes out a single grid to destination */
--- a/src/gmt_grdio.c
+++ b/src/gmt_grdio.c
@@ -1412,6 +1412,10 @@ int gmtlib_read_grd_info (struct GMT_CTRL *GMT, char *file, struct GMT_GRID_HEAD
 	/* unpack z-range: */
 	header->z_min = header->z_min * header->z_scale_factor + header->z_add_offset;
 	header->z_max = header->z_max * header->z_scale_factor + header->z_add_offset;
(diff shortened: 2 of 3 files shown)


joa-quim:
This will break Julia

PaulWessel:
Explain. it was always there until a few weeks ago?

joa-quim:
Not in ``GMT_GRID_HEADER`` that is not changed in 3 years. Changing it that's what would break Julia

PaulWessel:
I only put back a subset of what as in GMT until a few weeks/months ago, and only the grid stuff. I did not do the CPT and DATASET parts. The nx,ny was in the grid header until very recently.  So how did Julia work 2-3 months ago that is different from now w.r.t gmt_resources.h?

PaulWessel:
I can understand if you made changes to match our changes in the last 1-2 months on this, but otherwise I dont see why it would matter to Julia.

joa-quim:
No I didn't and I can't be doing it all the time. Maybe it worked because those 2 were at the end of the members list and hence *forgotten* by the julia struct
```
#ifdef GMT_BACKWARDS_API
	uint32_t nx;
	uint32_t ny;
#endif
```
The ``#ifdef`` doesn't help here because ``GMT_BACKWARDS_API`` is defined unconditionally . The corresponding Julia struct is [here](https://github.com/GenericMappingTools/GMT.jl/blob/master/src/libgmt_h.jl#L214)

PaulWessel:
If it continuous to work for Julia (as before), then presumably not a problem for you?

joa-quim:
Now that I see it was not conditionally include the situation is of a clear breaking, waiting to manifest mysteriously someday when something try to access those 8 bytes left in the ether.

PaulWessel:
Well, I would rather have it break for someone using 5.2 with GMTSAR than for Julia to break with 6.1.  So we will delete that PR and move on.

joa-quim:
But those have no problems with this. For what I understand is really GMTSAR that must update if it wants to use 6.1, and it's nothing complicated to do.

PaulWessel:
GMTSAR _has_ updated, but if you download GMTSAR now and want to build with GMT 5.2... no more nx, ny.

PaulWessel:
I think I have exceeded my number of branches in two weeks.  This is almost funny.  OK, will un-restore that branch once again.

PR #3574, opened 2020-07-02
Adding a smoothing spline to sample1d

PaulWessel:

There are times when having a smoothing spline is useful.  I wrote one in Matlab for a class but never put it into GMT.  Well, my students are in need of it so I decided yesterday to add it.  A bit more tedious to do matrix algebra in GMT but once in it is better.  I apologize for adding this so close to 6.1 release but the theory and coding in MATLAB is robust and after translating to C I get the same results.  As for the other splines I am also supporting first and second derivative output.  Here are some of the changes:

1. Add **-Fs**_p_ as smoothing spline with fit value _p_.
2. Add **-W** to allow for a weighted fit (only for **-Fs**)
3. Replace **+1** and **+2** with **+d**1 and **+d**2 to avo.id someone's positive number being confused for a modifier (e.g., -Fs+1.3 would be picked as a derivative request).  Backwards compatible.
4. Added updates to the man page, including equations for what is minimized.
5. Added two test scripts to demonstrate that it works, including derivatives.
6. Added data set to cache so that the example in the docs can be cut/paste.

Here is the result of the new script sample1d/smooth.sh. To compare the weighted and unweighted solutions, consider the data points at around 345 (a local low with large errors) and 410 (a local hight with very small error) and see how the smoothed fits are affected by these low and hight weights, respectively.

smooth

Here is the result of the new script for the derivatives (smooth_deriv.sh):

smooth_deriv


joa-quim:
Your examples show that the auto-legend lines are too short.

PaulWessel:
Yes, agree. Perhaps when symbol is line we should set a minimum size if no size is given, as here.

PR #3588, opened 2020-07-03
CPT with L|U|B ;label failed to be read

PaulWessel:

Read error on records line this:

`10.900	245/45/0	20.10	245/45/0	L	;5o`

Hope to fix inability to select the custom label or the custom numerical value.  The **-B** only gives equidistant values unless we make a second file with custom labels for **-B**.  That is too much trouble since the information is in a CPT.  Will allow the switch via **-S**.


PaulWessel:
Tested this PR and

1. Fixes the crash reported earlier
2. Leaves other tests working
3. Allows for turning on **-S+c** to use alternate labels.

PR #3595, opened 2020-07-04
Finalize the ChangeLog

PaulWessel:
To include the latest, plus some spellings.  @seisman, did you want to expand on teh seis entry or are you OK with what I put?


seisman:
> did you want to expand on teh seis entry or are you OK with what I put?

I think it's OK. Most changes in seis are just bug fixes.

PR #3596, opened 2020-07-05
Let gdal_open know about and apply cache dir

PaulWessel:
Since some files passed to GDAL has a leading driver: we need to insert the cach dir in-between the driver and the filename.  This is unique to files that must be read via GDAL this way. This allows a remote file to be used in test WL_example_3.sh.

WIP for now since only being merged after tag and creation of a 6.1 branch.


seisman:
We probably need to create the 6.1 branch and merge this PR into 6.1 since it's a bug-fix PR.

PaulWessel:
Yes, I agree.  I was going to wait until we wrap up and set the 6.1.0 tag first.

PR #3599, opened 2020-07-05
Update gmt_mb.h

PaulWessel:
The gmt function _gmt_get_cpt_ dropped its 7th argument for 6.1 and that affects the _gmt_get_cpt_ macro used in MB-System.

Oddly, the previous version of this include file had #ifdef checks on the GMT version, but that is pointless since the GMT version is what it is.  Any check would need to be for MB version.


PaulWessel:
Should have said, tested with MB-System master and MB-System-5.7.5.

PR #3605, opened 2020-07-06
Honor the colon switch for external data

### LINKED ISSUE #3289: gmtinfo -C -: gives wrong result from externals
Without -C it behaves well

```
julia> gmtinfo("", [1 10;2 20])
 "<matrix memory>: N = 2\t<1/2>\t<10/20>"


julia> gmtinfo("-:", [1 10;2 20])
 "<matrix memory>: N = 2\t<10/20>\t<1/2>"
```
but with it, no

```
julia> gmtinfo("-C", [1 10;2 20])
 1.0  2.0  10.0  20.0


julia> gmtinfo("-C -:", [1 10;2 20])
 20.0  10.0  1.0  2.0                           # Should be 10.0 20.0 1.0 2.0
```

PaulWessel:

When we use **-:** we basically set both **-:i** and **-:o**.  With file i/o this works fine.  With externals there were several issues (#3289 ):

1. When we read data from externals we either get a dataset, a matrix, or vectors.  We then build a dataset.  For these paths, there were no check on **-:i**.  I have added that.
2. When we write to externals from an internal dataset we either write a dataset, a matrix, or a vector.  Each of those cases needed a check for **-:o**.
3. Some modules do rec-by-rec i/o and go via a different path entirely.  I have similarly added **-:** checks on i/o for those pathways that did not check.

I think that corrects all these issues and closes #3289.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -3829,6 +3829,21 @@ GMT_LOCAL void gmtapi_increment_d (struct GMT_DATASET *D_obj, uint64_t n_rows, u
 	D_obj->n_tables++;	/* Since we just read one table */
 }
 
+GMT_LOCAL void gmtapi_switch_cols (struct GMT_CTRL *GMT, struct GMT_DATASET *D, unsigned int direction) {
+	uint64_t tbl, seg;
+	struct GMT_DATASEGMENT *S = NULL;
+
+	/* Implements the effect of -: when we are not writing to file */
+
+	if (D->n_columns < 2 || !GMT->current.setting.io_lonlat_toggle[direction]) return;	/* Nothing to do */
+	for (tbl = 0; tbl < D->n_tables; tbl++) {
+		for (seg = 0; seg < D->table[tbl]->n_segments; seg++) {
+			S = D->table[tbl]->segment[seg];
+			gmt_M_doublep_swap (S->data[GMT_X], S->data[GMT_Y]);
+		}
+	}
+}
+
 /*! . */
 GMT_LOCAL struct GMT_DATASET * gmtapi_import_dataset (struct GMTAPI_CTRL *API, int object_ID, unsigned int mode) {
 	/* Does the actual work of loading in the entire virtual data set (possibly via many sources)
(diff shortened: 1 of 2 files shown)


PaulWessel:
Just to document the fix and possibly have @seisman approve and/or test from PyGMT; this is Matlab:

```
>> D = gmt ('read -Td t.txt');
>> D.data

ans =

     1    10
     2    20
```

First we do **-:** which is short for both **-:i** and **-:o** so we read, toggle, processes, then toggle, write.  Except for **gmtinfo** **-C** where toggle on output gets deactivated when **-C** is used (#3604).  Below I test file, dataset (D), and matrix (D.data).  Note gmt info uses rec-by-rec while convert uses _GMT_Read_Data_, so exercises different parts of the API.

```
>> X = gmt ('info -: -C t.txt');
>> X.data

ans =

    10    20     1     2

>> X = gmt ('info -: -C', D);   
>> X.data                    

ans =

    10    20     1     2

>> X = gmt ('info -: -C', D.data);
>> X.data                         

ans =

    10    20     1     2
```

Now do convert.  The output should get double-toggled and give the same as input.

```
>> X = gmt ('convert', '-: t.txt');
>> X.data

ans =

     1    10
     2    20

>> X = gmt ('convert', '-:', D);   
>> X.data                       

ans =

     1    10
     2    20

>> X = gmt ('convert', '-:', D.data);
>> X.data                            

ans =

     1    10
     2    20
```

Then just toggle on input, so output remains toggled.

```
>> X = gmt ('convert', '-:i t.txt'); 
>> X.data                           

ans =

    10     1
    20     2

>> X = gmt ('convert', '-:i', D);    
>> X.data                        

ans =

    10     1
    20     2

>> X = gmt ('convert', '-:i', D.data);
>> X.data                             

ans =

    10     1
    20     2

```

Now for output:

```
>> X = gmt ('convert', '-:o t.txt');
>> X.data

ans =

    10     1
    20     2

>> X = gmt ('convert', '-:o', D);   
>> X.data                        

ans =

    10     1
    20     2

>> X = gmt ('convert', '-:o', D.data);
>> X.data                             

ans =

    10     1
    20     2
```

PR #3606, opened 2020-07-06
Let build-release always run the gs_check.sh script

PaulWessel:
Right now, a human error could let us forget to check if gs works well and be included in the macOS bundle.  This PR always runs the test and quits if there is a failure.


seisman:
Merge in 6.1 branch?

Issue #3609, opened 2020-07-06
GMT should not download the gmt_data_server.txt file if no remote files are used


seisman:
Every gmt command checks if `gmt_data_server.txt` and `gmt_hash_server.txt` exist locally. If not, GMT will download these two files from the GMT data server. GMT command also checks if these two files are older than 24 hours and try to refresh them if they are. It works well for users with good internet connections and users without internet connections.

However, it's a terrible experience for users who have very slow internet connections or work with intranet (intranet has network connections but can't visit external websites). Every time they run a gmt command, they need to wait for the timeout of the GMT server (~30 s). They won't understand why GMT is so slow, since the timeout message is not shown unless **-Vi** is used. Even **-Vi** is used, they may still be confused why GMT tries to download the two files even though they don't use any remote files like `@earth_relief_01d`.

IMHO, it's not a good design to let gmt check and download the two files every time. A better behavior is only try to download the two files if remote files are needed.

PaulWessel:
Agreed.  I will see where to implement a just-in-time solution.

PaulWessel:
Right now we call _gmtlib_refresh_server_ during the start of every session (so for most uses, this means every time a module is run, regardless of what the module actually does or if remote files are even requested).  I think this might work well:

1. Move that call to the top of _gmt_download_file_if_not_found_.  That is the function that determines if a requested file will need to be downloaded so we could do a just-in-time call to _gmtlib_refresh_server_.  There may be several calls to _gmt_download_file_if_not_found_ in that one session but it does not matter since once we do it once we set a bool flag so the function returns immediately if called again.
2. We would also need to call it explicitly when running **gmtget** **-D**.

Think this will cover all bases, @seisman?

seisman:
1. As I understand it, *gmt_download_file_if_not_found* function also deals with other URLs (not GMT remote files). Ideally, GMT should call *gmtlib_refresh_server* only if a remote file `@xxx` is requested, but I'm OK with it as downloading any files (GMT remote files or other URLs) needs good internet connections.

2. I'm not sure if my understanding of *gmt_download_file_if_not_found* is correct. If a file already exists locally, this function will do nothing, right? Does it mean that if I already have earth_relief_01d_p.grd locally, it will never refresh the server files, so the local earth_relief_01d_p.grd won't be updated?

PaulWessel:
[...]

PR #3616, opened 2020-07-06
Try to refresh only when remote files are requested

PaulWessel:

See #3609 for background.  This PR examines the option arguments in gmt_init_module, before any parsing happens, and if it finds an argument starting with @ or an argument containing a @ that is not inside quotes, it errs on the side of caution and refreshes the server (once).

All tests pass for me.  We consider this a but (excessive checking) so goes into 6.1.


seisman:
Looks good, but only have time testing it in next few days.

Issue #3615, opened 2020-07-06
psrose (rose) cannot make plots without frames


tianzeliu:
I just found that `psrose (rose)` cannot make plots with frames. Even when I use the option -B+n it still plots a circular frame.

PaulWessel:
It seems reasonable that **rose** should follow the same defaults as other maps.  Historically, it is doing its own frame plotting so that is why this issue appeared.  I think we want this behavior:

1. **-B+n**.  No circular frame, gridlines, or WESN labels.  Just the sector plotting should show.
2. **-B**0.  This should just draw the outline of the frame (circle or half-circle)
3. Default (no **-B** given):  Should be the same as **-B+n** I think.  Nobody expects a basemap in grdcontour if no -B is given either.

Current, no **-B** gives this interesting plot:

![map](https://user-images.githubusercontent.com/26473567/86668334-3e515400-bf8e-11ea-829a-bc6e7a053302.png)

Any objections to this plan, @GenericMappingTools/core ?

PaulWessel:
The code clearly has an automatic mode where it turns on part of the machinery but not the circle or radial grids, for instance, only azimuthal.

PaulWessel:
Apart from the clear goals above, I am finding other problems in **rose**.  In our _gallery.sh_ test script we have a typo regarding gridline intervals in azimuth: `-Byg190`, all the other places we use -`Byg90`, so a "1" slipped in.  But it does not matter since **psrose** simply _overrides_ that with a 30 degree selection if you did not specify an _annotation_ interval (never mind we do not annotate anyway except for the W,E,S,N labels).  So this is another bug: Ignoring the users gridline specification.  I know **rose** is old and has some quirks, but no other program automatically draws gridlines, so I think we must declare this a bug and fix it.  The goal must be for **-B** to work the same across the suite.  I assume @seisman agrees with this stand, yes?

seisman:
Yes, sounds reasonable to me.

PaulWessel:
Fixed by #3618.

PR #3618, opened 2020-07-07
Numerous bugs fixed in psrose

PaulWessel:

This PR started out by addressing #3615 which was an easy fix.  However, it became clear that **psrose -B** was violating most rules about base frames:

1. With no **-B** there should be no frame
2. With **-Bafg** or any of them, we should automatically estimate an interval.
3. The w/e/s/n labels should become -90/90 or 0/180 if half-circle.

Now, giving no **-B**, or **-B+n**, results in no frame, like all other modules with **-B**.  However, the -Bauto was interesting.  We had a hard-wired setting in **rose** that make this setting 30 degrees for all plot, no matter what size. This happen even if you actually specified a separate grid-interval like **-Bg**90.  Several tests used **-Bg**90 but were overruled to basically do **-Bg**30.  Once this hard-wired case was removed it became clear that the auto-interval was garbage. This is because internally, the polar projection switches to a square Cartesian one and hence the 0/360 angles are gone, yielding silly auto-intervals for azimuth.  OK, so we clearly had to overrule those in the module if auto-intervals were set. After that, I tried **-R**-90/90 with annotations and that actually crashed with SEGV, because we were trying to printf to Ctrl-L.w which once upon a time was a string but is now a pointer, and it was NULL.  So fixing that, and clarifying the man pages in several pages.  Some of the test scripts needed updates as there were typos in **-By** that not cause problems due to the other bugs.  A handful of PS originals had to be updated.  However, now **psrose -B** should behave like we expect.

### DIFF
--- a/doc/rst/source/psrose.rst
+++ b/doc/rst/source/psrose.rst
@@ -24,7 +24,7 @@ Synopsis
 [ |-M|\ *parameters* ]
 [ |-O| ] [ |-P| ]
 [ |-Q|\ *alpha* ]
-[ |-R|\ *r0*/*r1*/*az_0*/*az_1* ]
+[ |-R|\ *r0*/*r1*/*az0*/*az1* ]
 [ |-S| ]
 [ |-T| ]
 [ |SYN_OPT-U| ]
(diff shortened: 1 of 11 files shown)


PaulWessel:
Closes #3615.

PaulWessel:
Hi @dongdong, this one has been stuck all night in tests I think.  Shall I squash and merge?

seisman:
Yes, OK to merge.

PR #3622, opened 2020-07-07
Reset GMT->parent->cache otherwise julia would breake when reading images

joa-quim:
To prevent the side effects of #3596 that caused a [regression ](https://github.com/GenericMappingTools/GMT.jl/issues/389)in Julia


seisman:
@joa-quim You're merging master commits into 6.1 branch.

PaulWessel:
Same here with @joa-quim.  it is a tiny change so you can delete this PR and add that line to a 6.1 branch.

PR #3626, opened 2020-07-08
Delete local versions of remote files and tile-directories older than creation date

PaulWessel:
Since we anticipate that we will update some of the data sets on the server as new versions of SRTM15+ or age grids arrive, we check if the _date_ in the server data listing is more recent than the creation date of any copy found in the user's server directory.  Any files older that the server date will unceremoniously be removed, so that the next time the user accesses that file it will be downloaded from the server.
This check is only run if the _gmt_data_server.txt_ file has just been updated (by the current module) which only happens once a day, assuming the user runs GMT that often.  I am using the same scheme as **gmt clear** for (possibly populated) directories: Let the operating system delete the directory for us.  Single files we delete via C functions.

### DIFF
--- a/src/gmt_remote.c
+++ b/src/gmt_remote.c
@@ -158,15 +158,13 @@ GMT_LOCAL int gmtremote_parse_version (char *line) {
 
 GMT_LOCAL int gmtremote_remove_item (struct GMTAPI_CTRL *API, char *path, bool directory) {
 	int error = GMT_NOERROR;
-	fprintf (stderr, "Would delete: %s\n", path);	/* STOP HERE FOR NOW AS TESTING */
-	return GMT_NOERROR;
-	if (directory) {	/* Delete populated directories via the operating system */
+	if (directory) {	/* Delete populated directories via an operating system remove call */
 		char del_cmd[PATH_MAX] = {""};
 #ifdef _WIN32
-		char *t = gmt_strrep (path, "/", "\\");	/* rmdir needs paths with back-slashes */
+		char *t = gmt_strrep (path, "/", "\\");	/* DOS rmdir needs paths with back-slashes */
 		strcpy (del_cmd, "rmdir /s /q ");
 		strcat (del_cmd, t);
-		gmt_M_str_free(t);
+		gmt_M_str_free (t);
 #else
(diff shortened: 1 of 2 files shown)


src/gmt_remote.c
```diff
@@ -659,6 +729,8 @@ GMT_LOCAL int gmtremote_refresh (struct GMT_CTRL *GMT, unsigned int index) {
 			if (!access (old_indexpath, F_OK))
 				remove (old_indexpath);	/* Remove old index file if it exists */
 		}
+		else 
+			GMT->current.io.new_data_list = true;	/* Flag that we wish to delete datasets older than entries in this file */
```
joa-quim, line 733: Should ``gmt_end_module()`` reset this to false so that externals (MEX/Julia) do not inherit this value in posterior module runs?

PaulWessel: IT is set to falske nanoseconds later

PaulWessel:
FYI, I tested this PR by making my gmt_data_server.txt older using touch -t, then ran in debug.  Once it detected the file needed refreshing and after getting the updated file, I then edited the new file and changed the dates for two entries: one file and one tile directory.  Then I continued the debug and it correctly got the creation date of the file and directory on my local system, the date comparision showed they were older, and the system call removed the directory and the gmt_remove function removed the data set.  So, I think this should work well, including Windows since we are using the same scheme that @joa-quim used in gmt clear.  If you like, we could do another check by updating the dates on a few entries in the gmt_data_server.txt file on the server.

@seisman I guess we are still in purgatory w.r.t. the checks?

seisman:
Yes. You can still merge it if you already tested it locally.

PaulWessel:
Yep, I am pretty confident it is working as intended.

PaulWessel:
So I merged but fucked up and forgot to set it for 6.1.  Super-irritating that it selects master right now. Help, @seisman .

joa-quim:
DELETE 6.1

seisman:
> So I merged but fucked up and forgot to set it for 6.1. Super-irritating that it selects master right now. Help, @seisman .

Done. I cherry-picked the commit into 6.1 branch, and then merge 6.1 into master.

Issue #3612, opened 2020-07-06
Rethink our branching model and workflow


seisman:
**Problem**

Currently, the GMT repository has two long-lived branches, *master* and *6.1*. The *master* branch is unstable and developing, which hosts the codes for the next minor release (e.g., 6.2.0), and the *6.1* branch is for maintaining the GMT 6.1.x. All bug-fixes commits should be merged into 6.1 branch first, and then merge the 6.1 branch into the master branch. All new features should be merged into the master branch only.

It usually causes problems for new contributors and even core developers, since master is the default branch. Contributors may create a bug-fix PR based on the master branch, but the bug-fix PR should be merged into 6.1 branch instead.

Below are some discussions in #3610.

---

Due to the design of our branching model (master and 6.1), I can see that other new contributors are also likely choose the wrong base branch, since master is always the default one.

_Originally posted by @seisman in https://github.com/GenericMappingTools/gmt/pull/3610#issuecomment-654384419_

---

Others and me. It was happening all the time when we had a 6.0 branch

_Originally posted by @joa-quim in https://github.com/GenericMappingTools/gmt/pull/3610#issuecomment-654387710_

---

We may make less mistakes if we develop GMT 6.1.x in the master branch and add new features to **6.2** branch.

_Originally posted by @seisman in https://github.com/GenericMappingTools/gmt/pull/3610#issuecomment-654393616_

---

We could probably determine the statistics on this, but from gut feeling the vast majority of commits are in response to fixing a problem.  So currently, that means every time we have to remember to switch the branch to 6.1.  Since I have 6.1 checked out when I make a PR, it is always surprising that GitHub opens up a PR to be merged with master instead of the branch I am in.  So while we will screw up from time to time with any of these schemes, perhaps less screwups if we made a 6.2 for new things and continued our habit of doing most stuff to master.  That matches the pattern we create a new branch to test new features anyway.

_Originally posted by @PaulWessel in https://github.com/GenericMappingTools/gmt/pull/3610#issuecomment-654444077_

---

> Since I have 6.1 checked out when I make a PR, it is always surprising that GitHub opens up a PR to be merged with master instead of the branch I am in. 

It's possible to set the default branch to "6.1", but we have to remember to update the default branch when we release GMT 6.2.0 and create the 6.2 branch.

![image](https://user-images.githubusercontent.com/3974108/86637890-57fd9600-bfa4-11ea-9c1e-786dd05d74ac.png)

[...]

PR #3627, opened 2020-07-08
gmt.conf.rst: make table for listing terms

anbj:
gmt.conf.rst: make table for listing terms

### DIFF
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -243,16 +243,27 @@ FORMAT Parameters
         By default, longitudes will be reported in the range [-180,180]. The
         various terms have the following purpose:
 
-        - **D**: Use :term:`FORMAT_FLOAT_OUT` for floating point degrees [default].
-        - **+D**: Output longitude in the range [0,360]
-        - **-D**: Output longitude in the range [-360,0]
-        - **ddd**: Fixed format integer degrees
-        - **:**: Delimiter used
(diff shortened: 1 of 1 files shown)


seisman:
@anbj Thanks for making it look better.

You're using a "grid table" syntax to list the terms. It's OK, but we (at least myself) usually find it difficult to maintain/update/change a grid table. Thus we (again, at least myself) prefer to use a much simpler table syntax.

Please see https://docs.generic-mapping-tools.org/latest/rst-cheatsheet.html#simple-table for an example of the "simple table" syntax (right column for the raw code, left column for preview).

anbj:
@seisman Great, thanks for the tip! I’ll do it that way (if not anyone beats me to me) when I get back from vacation.

seisman:
Hi @anbj, we know that it's frustrating to switch between master and 6.1 branches when submitting a PR. After [some discussion](https://github.com/GenericMappingTools/gmt/issues/3612), we have improved our workflow to make it more friendly to contributors (and also to us). 

All new branches should be created from the master branch, and PRs should be merged into master branch, too. If the changes are critical for GMT 6.1.x release, then we will backport the changes into 6.1 branch, and the fix will be available in the next patch release.

So I've changed the base branch of this PR to master branch. It doesn't affect your branch and this PR.

anbj:
Great @seisman, thanks for the info. I’ll close this PR and make a new one with the simple table syntax when I have the opportunity.

Issue #3608, opened 2020-07-06
GMT_AUTO_DOWNLOAD may not work as expected.


seisman:
**Description of the problem**

Please check if you can reproduce the issue by following the steps:

1. Turn off your internet connection
2. Run `gmt set GMT_AUTO_DOWNLOAD off -Vd`. Since the internet connection is off, the debug messages say `Libcurl Error: Couldn't resolve host name`
3. Run `gmt defaults -Vd`. The command still tries to download the gmt_data_server.txt file and reports the the same error `Libcurl Error: Couldn't resolve host name`

PaulWessel:
Hi @seisman, is this issue still relevant after merging #3616 ?

seisman:
Yes, the issue is still here. To reproduce the issue, run:

```
gmt set GMT_AUTO_DOWNLOAD off
gmt grdimage @earth_relief_06m -pdf map -Vd
```

PR #3630, opened 2020-07-08
Remote update frequency

PaulWessel:

Adds new defaults parameter **GMT_DATA_SERVER_UPDATE** [1d] to control frequency of catalog refresh on local computer.  Also rearranges the description of global dataset with a leading chapter on how this is done, followed by sections per dataset.   I also fixed a memory leak when a module called a module and the server catalog ended up being allocated twice.

I need help with arranging this correctly so that

1. The sidebar listing is sensible once we open Dataset
2. The size of headers is correct [===, ---, and ~~~ do not seem to work the way I thought]


doc/rst/source/datasets/remote-data.rst
```diff
@@ -0,0 +1,135 @@
...
+becomes obsolete), and GMT will take actions accordingly.  It is our policy to only supply the *latest*
+version of any dataset that undergoes revisions.  If you require previous versions for your work you
+will need to get those data from the data provider separately.  Unless you deactivate the remote data service,
+GMT will do the following when you request a remote file in a GMT command:
+
+#. We check if the locally cached catalog with information about the data available from the server
+   is up-to-date or if it needs to be refreshed.  If the file is older that the :term:`GMT_DATA_SERVER_UPDATE`
+   limit then we refresh the catalog.
+#. When the catalog is refreshed, we determine the publication date for each dataset on the server,
+   and if any local copies you may have are now obsolete will will remove them to force a re-download from the server.
```
seisman, line 23:
```suggestion
   and if any local copies you may have are now obsolete we will remove them to force a re-download from the server.
```

doc/rst/source/datasets/remote-data.rst
```diff
@@ -0,0 +1,135 @@
...
+Many of the remote datasets have a preferred, default color table that will be used unless you
+override that default by giving your desired CPT information.
+
+Controlling the Process
+~~~~~~~~~~~~~~~~~~~~~~~
+
+There are several ways you can control the remote data process and the amount of space taken up by your
+own server directory:
+
+#. You can select the GMT data server closest to you to minimize download time [:term:`GMT_DATA_SERVER`].
```
seisman, line 53: Should also mention `GMT_AUTO_DOWNLOAD` here, but it may still have a bug (https://github.com/GenericMappingTools/gmt/issues/3608).

doc/rst/source/datasets/remote-data.rst
```diff
@@ -0,0 +1,135 @@
...
+Remote cache files are our collection of miscellaneous files that are used throughout the GMT examples,
+man pages, and test suite.  There is no system nor catalog and files come and go as we need them. The cache
+files are subject to similar rules as the remote data set when it comes to refreshing or deleting them.
+If any of these files is precious to you we suggest you make a copy somewhere.
+
+Getting a single grid
+~~~~~~~~~~~~~~~~~~~~~
+
+Should you need a single grid from any of our tiled dataset, e.g., to feed into other programs that do
+not depend on GMT, you can create that via :doc:`/grdcut.  For instance, to make a global grid from the
```
seisman, line 122:
```suggestion
not depend on GMT, you can create that via :doc:`/grdcut`.  For instance, to make a global grid from the
```

doc/rst/source/datasets/remote-data.rst
```diff
@@ -0,0 +1,134 @@
...
+becomes obsolete), and GMT will take actions accordingly.  It is our policy to only supply the *latest*
+version of any dataset that undergoes revisions.  If you require previous versions for your work you
+will need to get those data from the data provider separately.  Unless you deactivate the remote data service,
+GMT will do the following when you request a remote file in a GMT command:
+
+#. We check if the locally cached catalog with information about the data available from the server
+   is up-to-date or if it needs to be refreshed.  If the file is older that the :term:`GMT_DATA_SERVER_UPDATE`
+   limit then we refresh the catalog.
+#. When the catalog is refreshed, we determine the publication date for each dataset on the server,
+   and if any local copies you may have are now obsolete we will remove them to force a re-download from the server.
```
seisman, line 23: It would be useful if we can list all the available dataset here, and then link to the section titles below.

PaulWessel: You mean just a very short list of the types, like Earth DEM [link], Images [link], crustal age [link], etc?

doc/rst/source/datasets/remote-data.rst
PaulWessel: OK, will add that

PaulWessel: Updated.

PaulWessel:
Wonder if GMT_DATA_SERVER_UPDATE is the best word.  We want to convey interval, frequency, etc.  I can go very long with

GMT_DATA_SERVER_UPDATE_INTERVAL

but it is a bit long.  remove _SERVER_ and go with GMT_DATA_UPDATE_INTERVAL?

seisman:
> Wonder if GMT_DATA_SERVER_UPDATE is the best word. We want to convey interval, frequency, etc. I can go very long with
> 
> GMT_DATA_SERVER_UPDATE_INTERVAL
> 
> but it is a bit long. remove _SERVER_ and go with GMT_DATA_UPDATE_INTERVAL?

*GMT_DATA_UPDATE_INTERVAL* is good to me.

doc/rst/source/datasets/remote-data.rst
seisman: This is what I have done:
image

PaulWessel: OK, that is great, can you commit it? I just pushed the GMT_DATA_UPDATE_INTERVAL change

seisman: Done.

PaulWessel:
looks pretty nice. If we are OK with merging this then I will wait for it to update on the web page and ask Sandwell and Müller if they would like any changes.  Actually, before I do I think we should probably place their logos to give them more visual credit.  I will do that later today - have another zoom then must drive all the way in to UH to pick up some books and a RAID system that needs work...

PaulWessel:
I added the IGPP and EarthByte logos, at the end of each section, but looks bad.  Need to move to the header for each section, and I think we could use a horizontal line between each data set section.  Will head out now, if you have ideas to make this look good please do, else I will hack later.

seisman:
I've added horizontal lines between each data sets.

PR #3631, opened 2020-07-09
Only make tiny adjustments to dx and dy if both need it

PaulWessel:
We have a roundoff-patrol checker that looks to see if an geographic grid increment is close to a multiple of integer seconds or minutes.  If it is close enough then we set it to be exact.  However, the [case](https://forum.generic-mapping-tools.org/t/grdedit-a-seems-not-to-work/636/13) that triggered this discussion was close enough to 3s in x but not in y.  I think we should at least require **both** increments to be adjusted or neither, and possibly tighten the decision when to do this.  This PR requires both increment to be flagged to be adjusted instead of just one, so it fixes the immediate problem, but we can discuss the check.  To be flagged, the increment and its nearest integer has to differ by less than 10^-4 (GMT_CONV4_LIMIT).  Unfortunately, all this code was of course added for the same reason: Someones grid made elsewhere with ugly increments and boundaries, so it is there for a purpose.  We could tighten it to 10-5 or -6 but it could have other consequences.  I also turn off the roundoff-patrol when **grdedit** is used since it interferes with options like **-A**.

So summary of changes:

1. Only adjust increment to nearest integer minute or second if **both** increments are flagged this way
2. Do not do the adjustment when **grdedit** is used

### DIFF
--- a/src/gmt_grdio.c
+++ b/src/gmt_grdio.c
@@ -1293,22 +1293,23 @@ int gmtlib_get_grdtype (struct GMT_CTRL *GMT, unsigned int direction, struct GMT
 
 GMT_LOCAL void gmtgrdio_doctor_geo_increments (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *header) {
 	/* Check for sloppy arc min/sec increments due to divisions by 60 or 3600 */
-	double round_inc, scale, inc, slop;
-	unsigned int side;
+	double round_inc[2], scale[2], inc, slop;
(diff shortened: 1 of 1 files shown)


seisman:
@PaulWessel Let me know if you want to backport any PRs or commits into the 6.1 branch. I have to do it manually before we can set up a backport bot.

PaulWessel:
Yes, I think this is bad enough to warrent a 6.1 backport.  I will try hard to keep my PR small and focused and not mix too many things...

PR #3634, opened 2020-07-09
Fix sign-error on slope for RMA regression

PaulWessel:
Somehow, we never checked the correlation between x and y to know which sign to assign to the absolute value of the slope.  This PR fixes that and needs back-porting to 6.1.


seisman:
Blindly approve, again.

PaulWessel:
> Blindly approve, again.

I think that is OK.  If data are negatively correlated we have a negative slope, but the eq for the slope is based on the ratio of two positive numbers (stdev).  So using r to change the sign.

seisman:
Backport to 6.1 is done.

PaulWessel:
Great thanks.

PR #3644, opened 2020-07-12
WIP Let gmt download DCW from server if not found locallly

PaulWessel:
For now we look in all the usual pre-installed places, but if that fails we also look in ./gmt/geography/dcw and if no such dir or no files we download from server and use this path.
I have tested this in debug and command line and it works fine.  However, I think our cmake setup expects DCW_ROOT to be set correctly.  We would need to let this be an optional setting and if not then we do this all just-in-time.  I may need help from @seisman on the cmake side.  I dont think we (yet) want to remove the pre-install mechanism but just add the server download as a fallback if we cannot find the files where we expect them.

### DIFF
--- a/src/gmt_dcw.c
+++ b/src/gmt_dcw.c
@@ -69,32 +69,48 @@ static char *GMT_DCW_continents[GMT_DCW_N_CONTINENTS] = {"Africa", "Antarctica",
 /* Local functions only visible inside this file */
 
 GMT_LOCAL bool gmtdcw_get_path (struct GMT_CTRL *GMT, char *name, char *suffix, char *path) {
-	bool found = false;
-
 	/* This is the order of checking:
 	 * 1. Check in GMT->session.DCWDIR, if set
 	 * 2. Look via gmt_getsharepath.
+	 * 3. Look in userdir/geography/dcw
+	 * 4. Try to download from server into (3)
+	 * 5. Give up.
 	 */
 
 	if (GMT->session.DCWDIR) {	/* 1. Check in GMT->session.DCWDIR */
 		sprintf (path, "%s/%s%s", GMT->session.DCWDIR, name, suffix);
-		if ( access (path, R_OK) == 0)
-			found = true;
-		else {
-			/* remove reference to invalid GMT->session.DCWDIR but don't free
-			 * the pointer. this is no leak because the reference still exists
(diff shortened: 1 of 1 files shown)


seisman:
> I think our cmake setup expects DCW_ROOT to be set correctly.

Perhaps not. I can still build and install GMT without setting GSHHG_ROOT and DCW_ROOT.

cmake says:
```
*  Found GSHHG database       : GSHHG_PATH-NOTFOUND ()
*  Found DCW-GMT database     : DCW_PATH-NOTFOUND ()
```

Running `gmt pscoast` gives me errors like:
```
coast [WARNING]: GSHHG version 2.2.0 or newer is needed to use coastlines with GMT.
	Get and install GSHHG from ftp://ftp.soest.hawaii.edu/gshhg/.
coast [ERROR]: Could not find file [GSHHG intermediate resolution shorelines]
coast [ERROR]: No GSHHG databases available - must abort
```

PaulWessel:
OK, yes I guess that is true.  I got some build errors because I renamed my path to gshhg during testing after cmake had done its config.  Would we want the above message to say something different, e.g.

```
*  Found GSHHG database       : GMT will download on demand
*  Found DCW-GMT database     : GMT will ownload on demand
```

so a user wont freak out?

seisman:
How about this?
```
*  Found GSHHG database       : GSHHG_PATH-NOTFOUND (GMT will download on demand)
*  Found DCW-GMT database     : DCW_PATH-NOTFOUND (GMT will download on demand)
```

PaulWessel:
Sure, if that is simpler that is good enough.

PaulWessel:
Same here, are you OK with me removing WIP and that means that for GMT 6.2 there will be no discussion of how to install DCW as part of the installation.  We can discuss later if the installers should continue to contain DCW and GSHHG or if they should be lighter and data appear during use.

PaulWessel:
The way I am seeing this PR is that it will greatly simplify life for GMT users as they do not have to add one extra prerequisite that they may need to do manually via our gshhg tarball.  However, we would not simplify it if we continue to distribute the dcw tar balls and talk about it on the installation page.  The biggest issue is the lag it takes before the ubuntus of the world will reach and distribute 6.2.  Only then will people no longer look for the dcw tarball.

PR #3645, opened 2020-07-12
WIP if GSHHG cannot we be found we download from server

PaulWessel:
Instead of giving up if GSHHG files are not found, we simply download them from the server and read from the local ./gmt/geography/gshhg directory.

Tested and works. Like DCW, probably need to change the cmake setup so that not having GSHHG_ROOT set is not a fatal problem, @seisman.

### DIFF
--- a/src/gmt_shore.c
+++ b/src/gmt_shore.c
@@ -348,9 +348,43 @@ L1:
 			GMT_Report (GMT->parent, GMT_MSG_ERROR, "3. GSHHG: Found %s but cannot read it due to wrong permissions\n", path);
 	}
 
+	/* 4. Try the just-in-time download option from the server */
+
+	if (GMT->session.USERDIR) {	/* Check user dir via remote download */
+		char remote_path[PATH_MAX] = {""};
+		GMT_Report (GMT->parent, GMT_MSG_DEBUG, "4. GSHHG: Trying via remote download\n");
+		sprintf (path, "%s/geography/gshhg/%s.nc", GMT->session.USERDIR, stem);
+		if (access (path, R_OK) == 0) {	/* Found it here */
+			if ( gshhg_require_min_version (path, version) ) {
+				GMT_Report (GMT->parent, GMT_MSG_DEBUG, "4. GSHHG: OK, could access %s\n", path);
+				return (path);
+			}
+		}
+		/* Must download it the first time */
+		if (GMT->current.setting.auto_download == GMT_NO_DOWNLOAD) {
+			GMT_Report (GMT->parent, GMT_MSG_ERROR, "Unable to download the GSHHG for GMT since GMT_AUTO_DOWNLOAD is off\n");
+			return NULL;
+		}
+		sprintf (path, "%s/geography/gshhg", GMT->session.USERDIR);	/* Local directory destination */
+		if (access (path, R_OK) && gmt_mkdir (path)) {	/* Must first create the directory */
+			GMT_Report (GMT->parent, GMT_MSG_ERROR, "Unable to create GMT directory : %s\n", path);
+			return NULL;
+		}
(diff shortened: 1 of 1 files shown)


joa-quim:
I would still prefer that it first tried to put the files under ``share/coast`` (its natural location) and only resort to ``.gmt`` if permission don't allow it. Storing in ``.gmt`` means that a ``gmt clear`` will wipe them out.

seisman:
I've improved the GSHHG message in bd20180b0d33453d437e965a8e1f1e95327d2564.

If GSHHG is found, you'll see:
```
*  Found GSHHG database       : /path/to/coast (2.3.7)
```
otherwise:
```
*  Found GSHHG database       : GSHHG_PATH-NOTFOUND (GMT will download on demand)
```

seisman:
If GSHHG is not found, CMake still reports a fatal error when tests or examples are enabled. I think we can remove these message since GMT will download the coastline data when requested.

https://github.com/GenericMappingTools/gmt/blob/308386edd70fba6a5a8e0233f9bdbaa608e5f12f/src/CMakeLists.txt#L306-L311

PaulWessel:
Thanks @seisman .  Hi @joa-quim, the problem with your suggestion is that it will go against the policy of making things simple for the user.  Like our remote data service, we want to simply update grids and coastlines on the server and the users automatically get them when needed.  Thus, trying to place these files in an install directory that presumably may go away in 6.2 is not good.  Yes you can run gmt clear and delete it if you want to, but they get downloaded again when you ask for them.   The 'natural location' will become ./gmt/geography in the users directory and the whole business of tarballs etc will go away.

PaulWessel:
> If GSHHG is not found, CMake still reports a fatal error when tests or examples are enabled. I think we can remove these message since GMT will download the coastline data when requested.

yes, I saw that too and thought the same.  It is no longer relevant.  However, I think for our CI testing we would do the same we do for the remote data and pre-download to avoid any race condition of multiple downloads of the same file, etc.

joa-quim:
Well, that's a big novelty. You mean that you intend to get rid of the tarballs altogether and remove the GSHHG-DCW from cmake?

But  the files will still be searched in ``share/coast``, right? I have my modified version of the full resolution with high detail Portuguese coastlines and certainly want to continue using it

PaulWessel:
I am in no hurry to remove the search path system we have since we know ubuntu will have these until 2028.  So this will let you continue as before.  By 2028 we either have an openstreetmap solution or I will still be tinkering with this in retirement.

seisman:
> I have my modified version of the full resolution with high detail Portuguese coastlines and certainly want to continue using it.

Wow, does it take a lot of efforts? The official GSHHG data isn't good for Chinese users due to the [China-India border dispute](https://en.wikipedia.org/wiki/Sino-Indian_border_dispute). I'm also planning to modify GSHHG for Chinese users.

PaulWessel:
It is a bit rabbit hole to creep into.  Better to wait for OpenStreetMap and our postdoc.  News: We have applicants (4 or more I think).  We will discuss this at next GMT zoom.  We may want to have that zoom this Wednesday or Thursday due to the UNAVCO course as well.

PaulWessel:
Family  zoom will take up next 2 hours.

joa-quim:
I wrote tools in Mirone to help with the process but the biggest problem is that to build the nc files one needs tools that link with GMT4 and I suspect I wont be able to build GMT4 again, at least without substantial pain.

PaulWessel:
Are we happy with this solution, in which we will still look in the old places first (which lets @joa-quim have a custom file somewhere) but going forward we should

1. Stop requiring gshhg tar install as a prerequisite [there are non-GMT users of GSHHG so I wont remove the files from SOEST]
2. Deliver coastline in whatever format we end up with down the road via the data server only

If so then I can remove the WIP and we can let it rip.

PR #3668, opened 2020-07-16
Update grdcyclic.ps

seisman:
Need to update the original PS file due to changes in #3665.


seisman:
grdcyclic.sh still fails in the CI but works for me locally. Here is the diff image:

image

It looks something wrong with the earth relief data or the CPT. Does it pass for you? @PaulWessel

PaulWessel:
Fails for me locally as well.  Very subtle differences and I cannot connect it to any particular change yet.

PR #3671, opened 2020-07-18
Add new bitflag GMT_SESSION_NOHISTORY to turn off shorthand and history

PaulWessel:
Developers may wish to _not_ rely on the file-based communication between modules via the gmt.history file and the **-R -J** shorthands.  To do that they need to pass the new **GMT_SESSION_NOHISTORY** bit-flag to _GMT_Create_Sessions_. 

Note: gmt.c does not pass this value since we rely on gmt.history. It is possible there are unwanted side-effects to turning off history, such as for clipping levels, so developers need to use with caution.

This PR also removes unused enums from gmt_resources, such as **GMT_SESSION_BEGIN**, **GMT_SESSION_END**, **GMT_SESSION_CLEAR** and **GMT_SESSION_FIGURE**.


seisman:
GMT_SESSION_BEGIN et. al. are defined in GMT.jl, but not sure if they're used.

PaulWessel:
Yes of course, they are mirrored into his C wrapper, so since there is no version control then removing these causes trouble for him.  So I will need to reinstate them for now.  But both Julia and PyGMT needs to have some sort of system for version of GMT I think.

PaulWessel:
Let's wait for @joa to resurface.  It is Friday and trip to Lisbon and parents I think.

PaulWessel:
Hi @joa-quim, since there are just enums/constants I do not think their removal will affect you, unlike members of a structure, for instance.

joa-quim:
They are not used in Julia so save to remove.

I don't see a way of having an automatic system to control this. The file with the structure and constants definition was obtained initially with a Julia program that scanned the GMT C headers but after it I did a lot of manual cleaning & modifications. This is a process that I don't want to repeat.

PaulWessel:
OK, need a review for this one then.

PR #3672, opened 2020-07-18
Let psxy -T without -B skip the check for -R -J

PaulWessel:
Since no projection is needed to in this case (presumably to set **-X -Y** only) then we dont need to complain if **-R -J** are not set.  This means a **gmt psxy -T -X -Y** call can be the first in a plot sequence without running afoul of no **-R -J** history.
This PR also adds documentation of -T to the classic section of psxy.


seisman:
This PR doesn't work. `gmt plot -T -X1c -Y1c` is waiting for stdin input.

PaulWessel:
Yep, I only tested classic... SO in modern, we are trying to auto-determine the region from the data file... I will need to intervene in gmt_init_module based on -T.  Stay tuned.

PaulWessel:
This works for me now:

```
gmt begin map
  gmt basemap -R0/4/0/4 -JX4i -B1
  gmt plot -T -X2i
  gmt basemap -R0/4/0/4 -JX4i -B1
 gmt end show
```

PR #3675, opened 2020-07-18
if -R -J is given with -T then we must deal with them

PaulWessel:
The initial PR #3672 only called _gmt_map_setup_ if **-R -J** was _required_ but in fact we must of course call _gmt_map_setup_ if **-R -J** is _given_.


PaulWessel:
At issue are uses like this:

`gmt psxy -R0/360/-60/60 -JM5i -T -P -K -Xc`

which implicitly requires map width (and map height for -Yc).  So I am adding a check for that first.

PaulWessel:
Tests pass for me locally with this change.  Let's make sure it works for you or CI before we merge.

PaulWessel:
I think this is still good go go, @seisman  despite the late edit.  Now, -Xc or -Yc will require -R -J.

PR #3679, opened 2020-07-18
Turn outline on when a tile pen is found

PaulWessel:
We parsed the pen but forgot to set the bool flag.


joa-quim:
But will be able to test only later.

PR #3693, opened 2020-07-20
grd2cpt now has option to use data cumulative (weighted) distribution function

PaulWessel:

I have revised my plans on this after feedback from @WalterHFSmith.  The **grd2cpt** module default behavior remains the same as before for backwards compatibility. However, I have added two _new_ modifiers to option **-E**:

1. **+c** will compute the empirical cumulative density function and use that to design an equal-area CDF for the grid.
2. **+f**_file_ will write that CDF to a datafile (only if **+c** is used).  The format can be further controlled with **+b** and **+o**.

I have added a test for this, with output shown below.  Note that for a global grid, the optional CDF output is basically the cumulative hypsometric curve since pixel area is considered.  All tests pass as before.

equalarea

Top panel is regular linear CPT while bottom plot shows the CDF-derived CPT.  Below is the optional cdf output for this data set:

cdf


PaulWessel:
Since this has sat for a while, just pinging @seisman, @joa-quim and @WalterHFSmith that thise PR is now ready to be assessed.

PaulWessel:
Hi @WalterHFSmith, any comments on this or is it good to go? It preserves past behavior.

PaulWessel:
@seisman, Walter is OK with this - he cannot find his GitHug login to approve.

PR #3696, opened 2020-07-21
Let plotting of DCW in coast be done by plot

PaulWessel:
Because of the few polygons that have holes, we need to assemble a dataset and pass that to **psxy** in order for the holes to be properly handed.

It is WIP for now because I need to track down some memory leaks related to the dataset.  That part of the code had many bugs (it was never exercised) so this is a good test.  However, with DCW 1.1.5 it correctly renders the holes and works as before (ignoring holes) with DCW 1.1.4.  However, gmt 6.1.0 and earlier cannot use DCW 1.1.5 since the NaN coordinates have changed to pass the hole flag.  So we need to prevent users from installing 1.1.5 unless they have GMT 6.1.2 or higher.

### DIFF
--- a/src/gmt_dcw.c
+++ b/src/gmt_dcw.c
@@ -220,7 +220,7 @@ GMT_LOCAL bool gmtdcw_country_has_states (char *code, struct GMT_DCW_COUNTRY_STA
 struct GMT_DATASET * gmt_DCW_operation (struct GMT_CTRL *GMT, struct GMT_DCW_SELECT *F, double wesn[], unsigned int mode) {
 	/* Given comma-separated names, read the corresponding netCDF variables.
  	 * mode = GMT_DCW_REGION	: Return the joint w/e/s/n limits
-	 * mode = GMT_DCW_PLOT		: Plot the polygons
+	 * mode = GMT_DCW_PLOT		: Plot the polygons [This is actually same as GMT_DCW_EXTRACT internally but plots instead of returning]
 	 * mode = GMT_DCW_DUMP		: Dump the polygons
 	 * mode = GMT_DCW_EXTRACT	: Return a dataset structure
 	 */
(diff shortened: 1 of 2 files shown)


PaulWessel:
So GMT master (and hence 6.1.1 and beyond) can read dcw-gmt-1.1.x (x up to 5 which is not yet released), while GMT 6.1.0 an earlier can only use dcw-gmt-1.1.x up to 4.  While future code could add a check on the DCW version, released code cannot.  So what are our options?

1. Not release 1.1.5 until we release 6.1.1 presumably soon (August)?
2. Release 1.1.5 on my DCW site but state it requires 6.1.1 or later?

I think most distros grab the tarballs from us, so we will leave at 1.1.4 as far as tarball goes until 6.1.1 is out.  BUt once it is out, I guess we will continue to leave 1.1.4 there until the cows come home anyway.

Thoughts? I have not removed the WIP although I fixed my memory leak since ex34 (involving the IT polygon) is causing some issues I need to look at.

seisman:
Although there are only small changes in DCW, but it's not fully backward-compatible. Perhaps we should release DCW 1.2.0, rather than 1.1.5?

I think we should make the official DCW 1.2.0 (or 1.1.5) release after releasing 6.1.1, but before that, it would be better to have the tarballs uploaded to your own FTP, so that at least @joa-quim and me can test the new DCW data for a while before the official GMT 6.1.1 and DCW 1.2.0 release.

Is it possible to add one more attribute (the minimum required GMT version) to the DCW netCDF file, so that gmt_dcw.c can check if the current DCW data is compatible with the current GMT version? It won't help old releases <=6.1.0, but it will avoid similar troubles when we make similar incompatible changes in the future.

PaulWessel:
Yes, those are good ideas.  I will implement the checks and make it 1.2.0.

PaulWessel:
The PR #3703 will help in future releases.  This PR has no effect on 1.1.4 and is compatible with it and the future 1.2.0.  I think this PR can be merged now.  Tested and San Marino shows up as a hole in Italy, as does Lesotho in South Africa.  I can even see the tiny hole of the Vatican.

PaulWessel:
Need to merge this in so we can plot DCW, @joa-quim and @seisman .

PaulWessel:
FYI, this PR is DCW-version neutral.  It just switches plotting from inside gmt_dcw.c to a call via psxy.

PaulWessel:
You can run this branch with DCW 2.0.0 and it works - the checking of version is the other branch...

joa-quim:
Needs Label

PR #3697, opened 2020-07-21
Add missing KEY for -N option

joa-quim:


joa-quim:
But this raise a new case now, Since both -C and -N are allowed to pass a cpt I now get a ``Expected a CPT structure for input but got a Nothing``. Have to deal with it in Julia code (probably the same in Matlab)

PaulWessel:
Hm, things may be different since we are just passing the args to other modules.  Will have to look at this on a calmer day.

joa-quim:
Yes, I was afraid it would not work because of that, but it did.

PR #3711, opened 2020-07-23
Address bug in grdseamount -M option

PaulWessel:
The writing of the filenames etc to the list file with **-M** crashed because we did not pass the **GMT_WITH_STRINGS** mode when creating the dataset.  I also added time as the first numerical column since the initial version of this went back to text output days (it was all a string record) but it makes sense to write the time as a floating point, followed by the filename and the formatted time label.  I assume nobody has actually used this option (even me) since we have not been told of the crash, so adding a numerical column does not break anyone's workflow.


seisman:
@PaulWessel Do you want to backport this PR?

PaulWessel:
Yes I did, sorry.  Please do you magic...

PR #3728, opened 2020-07-25
Let table-reading module us the same language regarding the input name

PaulWessel:
Most modules use table, but a few said files, modelfile, intable, etc.  Now we use table throughout.


doc/rst/source/supplements/spotter/originater.rst
```diff
@@ -89,8 +89,8 @@ Required Arguments
 Optional Arguments
 ------------------
 
-.. |Add_intables| unicode:: 0x20 .. just an invisible code
-.. include:: ../../explain_intables.rst_
+.. |Add_intable| unicode:: 0x20 .. just an invisible code
+.. include:: ../../explain_intable.rst_
```
seisman, line 93: This file `explain_intable.rst_` doesn't exist.

doc/rst/source/supplements/spotter/originater.rst
```diff
@@ -89,7 +89,7 @@ Required Arguments
 Optional Arguments
 ------------------
 
-.. |Add_intables| unicode:: 0x20 .. just an invisible code
+.. |Add_intable| unicode:: 0x20 .. just an invisible code
```
seisman, line 92: This one also need to be reverted.
```suggestion
.. |Add_intables| unicode:: 0x20 .. just an invisible code
```

Issue #3723, opened 2020-07-24
Update and processing of gmt_data_server.txt is not robust


PaulWessel:
I've seen these when running the tests in parallel at a time that happens to coincide with the refresh of the data server list.  One gets messages like this:

`ERROR]: Read error first record in file /home/gia/.gmt/server/gmt_data_server.txt
`

apparently because that file got clobbered by another process and now it is zero bytes.  Because of the short course I have not had time to investigate more but will be looking at flock as we do for gmt.history.  Also, when the above situation arises we could take some other action than complain, such as delete and redownload.

PaulWessel:
The current plan is to use flock on a dummy file since I cannot set flock on the file being downloaded.  So will try this scheme when we wish to download file A:

1. Create /tmp/A.download
2. Get exclusive flock on that file, meaning other files trying to do the same must wait
3. Do the libcurl download of A
4. Release the lock and delete the file A.download.

Will make a branch and simulate this via the schemes listed on[ this page](https://gavv.github.io/articles/file-locks/).

seisman:
I believe the issue was fixed in #3735.

PR #3735, opened 2020-07-26
Let libcurl downloads be managed via lockfiles

PaulWessel:

When the time is up to refresh the gmt_data_server.txt file, and you run several commands, possibly simultaneously, that all need to access remote files (not necessarily the same file), they all realize that the local copy of gmt_data_server.txt is too old and must be refreshed.  So then the mad dash of being the first to do this starts among competing processes. What seems to happen is that they are clobbering each other and sometimes we are left with a zero-size gmt_data_server.txt file.  This file now has a fresh date, so it is no longer downloaded, but it has no contents so any remote grids will fail to be aquired.  This problems seems particulary bad when I run ctest with 20+ cores and that file is due for a refresh.

This PR tries to use the lockfile mechanism we have used for years for the gmt.history file (to protect them from being clobbered with your run process1 | process2, for instance.  Except here, the file is acquired inside libcurl.  The solution explored in this PR is to start the download sequence as outlined in #3723.  So far so good.  I did two experiments:

1. I removed a cache file and then ran gmt which -G in Xcode debugger and placed a break point right after setting the lock.  I then tried to run the same command in the terminal.  Because of the filelock the terminal command was blocked and it just sat there until I hit continue in the debugger.  Once the file was downloaded and the lock released and deleted, the terminal job immediately finished its own copying of the file.
2. I deleted my gmt_server_data.txt file and ran make check with 20 cores.  I know many of these cores would be wanting to refresh that file at the same time and normally leave an emtpy file, but this time it worked just fine.

I hope @joa-quim and @seisman can give this branch a spin and see if they discovered any issues.  It already seems to me it works better than the no-lock release we have.  I added WIP so that we dont merge until we test a bit more.  Comments welcome.


PaulWessel:
The latest commit (delete gmt_server_data.txt if faulty) should also help as it will allow a refresh teh next time accessed.  Note that the gmtserver apparently has powered down due to the hurricane so you cannot really test on oceania until we resurface.

PaulWessel:
THis works well for me and I have not had any gmt_data_server.txt errors since adding the flock and remove.

PR #3738, opened 2020-07-26
gmt_make_enum_dicts.sh needed some TLC

PaulWessel:
It had an aggressive grep -v to avoid lines that start with struct etc but it would also trigger if the word struct appeard in a comment.  Added ^ to indicate start of line.  This picks up the missing ones.  GMT_GRID_ALL is a backwards compat for < 5.3.3 so wont be included (it is not an enum, just a #define).


seisman:
The automatic backport fails due to some conflicts. I'll do it manually.

PR #3742, opened 2020-07-27
Put quotes when writing DIR_DATA to scripts

### LINKED ISSUE #3698: movie requires that the path has no spaces in it
e.g. if I have animations/ in my Google Drive and run the command, 

(base) see2-236:animation chris$ gmt movie earth.sh -C360p -T91 -Nearth -M55,pdf -Fnone
gmtset [ERROR]: Unrecognized keyword Drive/animation,/Users/chris/Google.
gmtset [ERROR]:  1 GMT Defaults conversion errors from command options
usage: mv [-f | -i | -n] [-v] source target
       mv [-f | -i | -n] [-v] source ... directory

PaulWessel:
Since gmt set **DIR_DATA** _path_ needs to be one word, we simply place it under quotes, always, when movie.c or batch.c write this to a script.
I tested this solution by making a dir called "dumb dir" in my home dir and tried to run the globe.sh example from the movie documentation before recompiling.  It failed badly due to the space.  After this PR it worked fine.  It would be great if @joa-quim or @seisman could test under Windows just in case.

Closes #3698.

### DIFF
--- a/src/batch.c
+++ b/src/batch.c
@@ -570,7 +570,7 @@ EXTERN_MSC int GMT_batch (void *V_API, int mode, void *args) {
 		while (gmt_fgets (GMT, line, PATH_MAX, Ctrl->S[BATCH_PREFLIGHT].fp)) {	/* Read the preflight script and copy to the temporary preflight script with some exceptions */
 			if (gmt_is_gmtmodule (line, "begin")) {	/* Need to insert the DIR_DATA statement */
 				fprintf (fp, "%s", line);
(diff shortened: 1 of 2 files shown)


PaulWessel:
I think this one is fine since it failed before and now passes if you run any movie in a directory with spaces.  It would be gerat if you could just copy/paste the globe.sh movie form the movie man page, create a badly named directory, and run the movie there using this branch, @joa-quim and @seisman

PR #3745, opened 2020-07-27
Do not consider memory references remote files

PaulWessel:
The gmt_init_module check for remote files foolishly considered @GMTAPI@-* files as remote since it only looked for the leading @.


PaulWessel:
Must be CI issues and not the PR, right?

seisman:
The CI fails to download the earth relief data from the server. It should be good to merge.

PR #3751, opened 2020-07-27
Make sure -L pen does not replace frame pen

### LINKED ISSUE #3750: pscontour -L may paint the axes
If in the hilow.sh test one add ``-L0.25p,red,- `` to the colored part, we get
hilow

PaulWessel:
Placing the mesh drawing under gsave/grestore does the trick and closes #3750.

### DIFF
--- a/src/pscontour.c
+++ b/src/pscontour.c
@@ -1255,6 +1255,7 @@ EXTERN_MSC int GMT_pscontour (void *V_API, int mode, void *args) {
 
 	if (Ctrl->L.active) {	/* Draw triangular mesh */
 
+		PSL_command (GMT->PSL, "V\n");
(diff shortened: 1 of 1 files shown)


PaulWessel:
Since CI tests (and my running the tests locally) will fail due to gmt server, just want to know that you built and made sure it works for you.  If so I can merge.

seisman:
I believe the GMT server is up.

joa-quim:
No I did not build this branch, and the server being up wont help if the script is not changed as this combination of **-L -I** was not tested before.

PaulWessel:
I think the fix is fine, but not sure if -L with -I does anything.  I could not tell from the plot.
I did reproduce your example and problem went away after this simple fix, so OK I think.  But server is now up I see.

PR #3760, opened 2020-07-28
Better key verification

PaulWessel:
Only return a remote file ID when it truly is a remote file (and not files like "earth").


PaulWessel:
This works for me now, @seisman, let me know if still pyGMT issues.

seisman:
Backport to 6.1?

PaulWessel:
Yes, sorry. please do.

PR #3771, opened 2020-07-29
Do not let the API object for the assembled grid live on

PaulWessel:
_GMT_Read_Data_ assigns a new object number to what is read, but if input file is a set of tiles it calls _gmt_assemble_grid_ which also creates an object to pass from **grdblend**.  However, once we get it back from **grdblend** we only want to keep the grid and discard the container since _GMT_Read_Data_ already has a container for it.

gmtmex('grdcut @earth_relief_01m -R-37.0/-35.0/-2.3/-1.0')

works when built with this PR.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -7564,6 +7564,27 @@ int GMT_Close_VirtualFile_ (unsigned int *family, char *string, int len) {
 }
 #endif
 
+int gmtlib_delete_virtualfile (void *V_API, const char *string) {
+	/* Given a VirtualFile name, delete its record bu NULLing it */
+	int object_ID, item;
+	struct GMTAPI_DATA_OBJECT *S_obj = NULL;
+	struct GMTAPI_CTRL *API = NULL;
+	if (V_API == NULL) return_error (V_API, GMT_NOT_A_SESSION);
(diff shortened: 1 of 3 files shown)


PaulWessel:
Pinging @seisman

seisman:
OK, will see what happened later.

seisman:
So there is a conflict in src/gmt_internal.h
```diff
++<<<<<<< HEAD
++=======
+ EXTERN_MSC int gmtlib_delete_virtualfile  (void *API, const char *string);
+ EXTERN_MSC bool gmtlib_file_lock (struct GMT_CTRL *GMT, int fd);
+ EXTERN_MSC bool gmtlib_file_unlock (struct GMT_CTRL *GMT, int fd);
++>>>>>>> 46ff22613b... Do not let the API object for the assembled grid live one (#3771)
```

These two functions were added in PR #3735, but not backported to 6.1 branch. So they shouldn't be added to 6.1.
```
+ EXTERN_MSC bool gmtlib_file_lock (struct GMT_CTRL *GMT, int fd);
+ EXTERN_MSC bool gmtlib_file_unlock (struct GMT_CTRL *GMT, int fd);
```

PaulWessel:
yes, correct.

Issue #3694, opened 2020-07-20
image shading problem for tiled earth relief data?


seisman:
Plotting a global earth relief map using the following command:
```
gmt grdimage -JN120/20c -Rg -Baf @earth_relief_05m -I+d -pdf map
```
It looks good at first glance, but after zooming in, I saw a lot of weird curves:

![image](https://user-images.githubusercontent.com/3974108/87905332-531df500-ca2e-11ea-98b9-1fd2ddb8227a.png)

I don't see these curves for `@earth_relief_06m` or `earth_relief_05m` without shading.

PaulWessel:
Wow, strange.  Well ,good that it is not in the data files I guess.  Running grdgradient manually and passing the file works.  So something with passing @file to grdgrading via GMT_Call_Module.

PaulWessel:
Just an update.  This only applies to tiled data sets.  The 06m and larger do not have this problem.

PaulWessel:
And of course has nothing to do with modern mode.  A plain 

`gmt grdimage -JN120/20c -Rg -Baf @earth_relief_05m -I+d > t.ps`

yields the same problem.

PaulWessel:
Some more experimentation.:

1. I tried a non-curvilinear projection -JQ.  No problems, so this involves the projection machinery in some way.
2.  I added a _gmt_M_memset_ call to first set the entire intensity grid to zero after projection.  The resulting map had the stripes, so they are in the data grid.
3. I then flipped that to instead set the dataset to zero and plotted.  The resulting map had intensity (shading) only and it was fine with no stripes.
4. I made a copy of the data grid before passing it to **grdgradient**, then restored it.  Made no difference.

Conclusions:

1. Since the intensity grid is derived from the topo grid via an internal call to **grdgradient**, and since the intensity grid is fine, then any problem with the data grid must happen during projection.
2. The argument against happening afterwards is that the data and intensity grid take exactly the same route through the projection steps, so it is unclear how the projection could mess up the data grid but not the intensity grid.

Yet, will need to see if there is any difference in the two calls to _gmt_grd_project_.

PaulWessel:
Slowly making progress on understanding this but not out of the woods.  Can @seisman and @joa-quim try the same command but turning off **anti-aliasing**?  That gives a good result for me but need to verify before I dig into the why (since it works find with a single file or without shading...)

[...]

PR #3799, opened 2020-08-01
Special attention when tile grids are passed to grdimage

PaulWessel:
This addresses issue #3694.  The problem turned out to be this:

**grdimage** usually reads in a grid in two steps:

1. Read in the grid header only to determine the grid region
2. After calling _gmt_map_setup_, which may adjust **-R** for global grids, we read the grid, this time passing the adjusted wesn for the map.

So, in the case that caused this issue, we may use -Rg and -JN120/20c, but after _gmt_map_setup_, the R.wesn is no longer 0/30 but -60/300.  So the grid is read in correctly relative to the desired region.

However, when we read a global tiled region there is a difference: Because the grid is assembled by **grdblend**, the first call (to get the header) needs to do the whole thing since we are not going to assemble the grid twice just to be the region.  Thus, the tiled region in this case remains 0/360 since there is no second read after projection is set.  Nevertheless, none of this caused a problem until we also selected auto-shading.  Now, the 0/360 grid is passed to **grdgradient** with the actual region and two grids are out of sync.

There may be more than one solution to this.  Since this case is unique to **grdimage** the following solution was implemented there: Since we know we read a tiled_grid up front in this specific case (tiled grid + auto-shading) we simply pass the 0/360 region to **grdgradient** rather than the adjusted region.  With that change, the case works and all tests work fine.

### DIFF
--- a/src/gmt_map.c
+++ b/src/gmt_map.c
@@ -8029,8 +8029,9 @@ int gmt_grd_project (struct GMT_CTRL *GMT, struct GMT_GRID *I, struct GMT_GRID *
 
 			if (!GMT->common.n.antialias || nz[ij_out] < 2)	/* Just use the interpolated value */
 				O->data[ij_out] = (gmt_grdfloat)z_int;
-			else if (gmt_M_is_dnan (z_int))		/* Take the average of what we accumulated */
-				O->data[ij_out] /= nz[ij_out];	/* Plain average */
+			else if (gmt_M_is_dnan (z_int)) {		/* Take the average of what we accumulated */
+				if (nz[ij_out]) O->data[ij_out] /= nz[ij_out];	/* Plain average */
+			}
 			else {					/* Weighted average between blockmean'ed and interpolated values */
 				inv_nz = 1.0 / nz[ij_out];
 				O->data[ij_out] = (gmt_grdfloat) ((O->data[ij_out] + z_int * inv_nz) / (nz[ij_out] + inv_nz));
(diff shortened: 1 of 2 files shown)


PaulWessel:
Maybe have a look at this tomorrow, @seisman

seisman:
There is a conflict in grdimage.c when cherry-picking this PR:
```diff
++<<<<<<< HEAD
 +              /* Prepare the grdgradient arguments using selected -A -N and the data region in effect */
 +              sprintf (cmd, "-G%s -A%s -N%s+a%s -R%.16g/%.16g/%.16g/%.16g --GMT_HISTORY=false ",
 +                      int_grd, Ctrl->I.azimuth, Ctrl->I.method, Ctrl->I.ambient, wesn[XLO], wesn[XHI], wesn[YLO], wesn[YHI]);
 +              if (got_data_grid)      /* Use the virtual file we made earlier */
++=======
+               /* Prepare the grdgradient arguments using selected -A -N and the data region in effect.  If we read in a tiled grid
+                * then it was made with 0/360 so we must use that region in grdgradient.  For non-tiles, we read the actual grid
+                * AFTER calling gmt_mapsetup which changes the region.  The tiled region remains unchanged because it is read in
+                * all at once as it is being assembled.  This fixes https://github.com/GenericMappingTools/gmt/issues/3694 */
+               sprintf (cmd, "-G%s -A%s -N%s+a%s -R%.16g/%.16g/%.16g/%.16g --GMT_HISTORY=readonly ",
+                       int_grd, Ctrl->I.azimuth, Ctrl->I.method, Ctrl->I.ambient, region[XLO], region[XHI], region[YLO], region[YHI]);
+               if (got_data_tiles)     /* Use the virtual file we made earlier */
++>>>>>>> a336215a12... Special attention when tile grids are passed to grdimage (#3799)
``` 
I believe I should accept the changes in the 2nd block, but change `--GMT_HISTORY=readonly` to `--GMT_HISTORY=false` (we didn't backport the GMT_HISTORY changes of https://github.com/GenericMappingTools/gmt/pull/3687 to 6.1 branch).

PR #3804, opened 2020-08-01
Must close file before delete it

joa-quim:
Fix broken situation on Windows with the lock files where files cannot be deleted if they were not closed,

PR #3780, opened 2020-07-30
Checked wrong string for URL

PaulWessel:
Hopefully fixes #3765.
I think backport also, yes?

PR #3768, opened 2020-07-29
Exempt URL queries from having lock files

PaulWessel:
Since the URL of a quiery is not a good file name, we do not create an advisory lock file for such downloads.  Addresses #3765 hopefully.

PR #3808, opened 2020-08-01
[Backport 6.1] File lock on CURL downloads

seisman:
Backport #3735, #3768, #3780, #3785, and #3804.


PaulWessel:
I should be more careful adding the label BUG when it is one - like clobbering teh gmt_data_server.txt really isthe

PR #3811, opened 2020-08-01
Adding API test for plotting a global matrix/grid for different central meridians

PaulWessel:

See https://github.com/GenericMappingTools/pygmt/issues/515#issue-655281714 for the backstory.  This PR adds a new testapi_matrix_360.c test program in C that recreates the same type of problem.  The plot with a zero central meridian looks fine but the one for 180 is not rotated to show the Pacific and has a nasty notch at Greenwhich instead.  I will try to make this case work once I better understand the reason.  Initial attempts to rotate the grid failed, so there is more to it.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -5482,6 +5482,10 @@ GMT_LOCAL struct GMT_GRID * gmtapi_import_grid (struct GMTAPI_CTRL *API, int obj
 					}
 				}
 			}
+			if (gmt_M_is_geographic (GMT, GMT_IN) && gmt_M_360_range (M_obj->range[XLO], M_obj->range[XHI])) {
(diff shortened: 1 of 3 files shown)


PaulWessel:
Just for reference, here is the current output of testapi_matrix_360.c.  The top plot should show the Pacific centered on 180.

t

PaulWessel:
Hi @seisman, I'd like to merge this to master even though there is no fix yet (so test will fail since no orig PS).  Problem I have is it takes Xcode so long to completely cmake up a new branch, and since this branch adds a new C file I need to wipe and start fresh since otherwise it complains about missing C module (if I switch back/forth with master).  So my plan is to put this into master now and then when a fix is found we can merge and backport as well.  If there is no problems with this plan then please approve.

PR #3817, opened 2020-08-02
Use the currrent default export type, not GMT_DOUBLE

PaulWessel:
The GMT defaults setting **GMT_EXPORT_TYPE** controls the default data types for matrices and vectors.  It defaults to double, but we had hardwired the use of **GMT_DOUBLE** in some places where we should use `GMT->current.setting.export_type`.  This prevented us from controlling what _GMT_Read_Data_ would return when **GMT_MATRIX** was selected as the family.


PaulWessel:
Sorry, fixed.

PR #3825, opened 2020-08-03
gmt.conf.rst: make table of FORMAT_GEO_OUT terms

anbj:
gmt.conf.rst: make table of FORMAT_GEO_OUT terms


PaulWessel:
But this is not really a bug fix, just enhancement, so not really needed as backport, no?

PR #3829, opened 2020-08-03
Handle matrix via reference for grid input

PaulWessel:
The passing of a matrix masquerading as a grid via **GMT_IS_DUPLICATE** [the default mode] now seems to work.  However passing such grids by reference is also allowed, but I suspect there are cases where it cannot be done (since we would need to do something special to the matrix).  Testing is done via testapi_matrix_360_ref.c. Seems to work (please verify if this works from PyGMT, @seisman) but raises an issue related to memory freeing.  Since the matrix data is only pointed to by the grid, I set the grid alloc_mode to external so both of them do not try to free the memory.  However, that also means that the x and y arrays that were created when creating the grid is left at the end.  We could just ignore this rare memory leak or do something about it.  One way is to add info to the struct **GMT_GRID_HEADER_HIDDEN** so that we can free those arrays even those the matrix is left untouched.

I don't know if @joa-quim is wrapping the gmt_hidden.h file in Julia and so adding a new variable will cause trouble.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -5210,7 +5210,7 @@ GMT_LOCAL struct GMT_GRID * gmtapi_import_grid (struct GMTAPI_CTRL *API, int obj
  	uint64_t row, col, kol, row_out, i0, i1, j0, j1, ij, ij_orig;
 	size_t size;
 	unsigned int both_set = (GMT_CONTAINER_ONLY | GMT_DATA_ONLY);
-	unsigned int method;
+	unsigned int method, start_over_method = 0;
 	double dx, dy, d;
 	p_func_uint64_t GMT_2D_to_index = NULL;
 	struct GMT_GRID *G_obj = NULL, *G_orig = NULL;
@@ -5243,6 +5243,9 @@ GMT_LOCAL struct GMT_GRID * gmtapi_import_grid (struct GMTAPI_CTRL *API, int obj
 		if (grid->header->wesn[XLO] == S_obj->wesn[XLO] && grid->header->wesn[XHI] == S_obj->wesn[XHI] && grid->header->wesn[YLO] == S_obj->wesn[YLO] && grid->header->wesn[YHI] == S_obj->wesn[YHI]) S_obj->region = false;
 	}
 	method = gmtapi_set_method (S_obj);	/* Get the actual method to use since may be MATRIX or VECTOR masquerading as GRID */
+
+start_over_import_grid:		/* We may get here if we cannot honor a GMT_IS_REFERENCE from below */
+
 	switch (method) {
 		/* Status: This case is fully tested and operational */
 		case GMT_IS_FILE:	/* Name of a grid file on disk */
@@ -5491,13 +5494,25 @@ GMT_LOCAL struct GMT_GRID * gmtapi_import_grid (struct GMTAPI_CTRL *API, int obj
 				return_null (API, GMT_GRID_BC_ERROR);	/* Set boundary conditions */
 			API->object[new_item]->status = GMT_IS_USED;	/* Mark as read */
 			API->object[new_item]->actual_family = GMT_IS_GRID;	/* Done reading from matrix */
+			if (start_over_method) API->object[new_item]->method = start_over_method;	/* We changed our mind from reference to duplicate due to region */
 			GH->alloc_level = API->object[new_item]->alloc_level;	/* Since allocated here */
 			break;
 
 	 	case GMT_IS_REFERENCE|GMT_VIA_MATRIX:	/* The user's 2-D grid array of some sort, + info in the args [NOT YET FULLY TESTED] */
 			/* Getting a matrix info S_obj->resource. Create grid header and then pass the grid pointer via the matrix pointer */
 			if ((M_obj = S_obj->resource) == NULL) return_null (API, GMT_PTR_IS_NULL);
-			//if (S_obj->region) return_null (API, GMT_SUBSET_NOT_ALLOWED);
+			/* Determine if it is possible to use the matrix given the region selected and the fact we chose GMT_IS_REFERENCE. This test will
+			 * only kick in after we allocate the G_obj and come back the second time (after getting header) since otherwise S_obj->wesn is not set yet */
+			if (!(!S_obj->region ||
+				(S_obj->wesn[XLO] >= M_obj->range[XLO] && S_obj->wesn[XHI] <= M_obj->range[XHI] && S_obj->wesn[YLO] >= M_obj->range[YLO] && S_obj->wesn[YHI] <= M_obj->range[YHI]) ||
+				(gmt_M_is_geographic (GMT, GMT_IN) && gmt_M_360_range (M_obj->range[XLO], M_obj->range[XHI]) && gmt_M_360_range (S_obj->wesn[XLO], S_obj->wesn[XHI])))) {	/* Cannot do this by reference, switch to duplication */
+				method -= GMT_IS_REFERENCE;
+				method += GMT_IS_DUPLICATE;
+				start_over_method = GMT_IS_DUPLICATE;
+				GMT_Report (API, GMT_MSG_DEBUG, "Subset selection requires GMT_IS_DUPLICATION instead of GMT_IS_REFERENCE - method has been switched\n");
+				goto start_over_import_grid;
+			}
+
 			/* This method requires the input data to be a GMT_GRD_FORMAT matrix - otherwise we should be DUPLICATING */
 			MH = gmt_get_M_hidden (M_obj);
 			if (!(M_obj->shape == GMT_IS_ROW_FORMAT && M_obj->type == GMT_GRDFLOAT && (mode & GMT_GRID_IS_COMPLEX_MASK) == 0))
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -5545,17 +5545,25 @@ GMT_LOCAL struct GMT_GRID * gmtapi_import_grid (struct GMTAPI_CTRL *API, int obj
 #endif
 			GH = gmt_get_G_hidden (G_obj);
 			S_obj->alloc_mode = MH->alloc_mode;	/* Pass on alloc_mode of matrix */
-			GH->alloc_mode = MH->alloc_mode;
+			//GH->alloc_mode = MH->alloc_mode;
+			GH->alloc_mode = GMT_ALLOC_EXTERNALLY;	/* Since we cannot have both M and G try to free */
 			API->object[new_item]->resource = G_obj;
 			API->object[new_item]->status = GMT_IS_USED;	/* Mark as read */
 			GH->alloc_level = API->object[new_item]->alloc_level;	/* Since allocated here */
-			if (S_obj->region) {	/* Possibly adjust the pad so inner region matches wesn */
+			if (gmt_M_is_geographic (GMT, GMT_IN) && gmt_M_360_range (M_obj->range[XLO], M_obj->range[XHI]) && gmt_M_360_range (G_obj->header->wesn[XLO], G_obj->header->wesn[XHI])) {
+				/* Global grids passed via matrix are not rotated to fit the desired global region, so we need to correct the wesn for this grid to match the matrix */
+				gmt_M_memcpy (G_obj->header->wesn, M_obj->range, 4U, double);
+			}
+			else if (S_obj->region && S_obj->orig_wesn[XLO] >= M_obj->range[XLO] && S_obj->orig_wesn[XHI] <= M_obj->range[XHI] && S_obj->orig_wesn[YLO] >= M_obj->range[YLO] && S_obj->orig_wesn[YHI] <= M_obj->range[YHI]) {	/* Possibly adjust the pad so inner region matches wesn */
 				if (S_obj->reset_pad) {	/* First undo a prior sub-region used with this memory grid */
 					gmtlib_contract_headerpad (GMT, G_obj->header, S_obj->orig_pad, S_obj->orig_wesn);
 					S_obj->reset_pad = 0;
 				}
 				if (gmtlib_expand_headerpad (GMT, G_obj->header, S_obj->wesn, S_obj->orig_pad, S_obj->orig_wesn))
 					S_obj->reset_pad = 1;
+			}
+			else {
+				GMT_Report (API, GMT_MSG_ERROR, "Unable to select a subset from this GMT_IS_REFERENCE matrix\n");
 			}
 			break;
 
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -5479,7 +5479,7 @@ GMT_LOCAL struct GMT_GRID * gmtapi_import_grid (struct GMTAPI_CTRL *API, int obj
 					}
 				}
 			}
-			if (gmt_M_is_geographic (GMT, GMT_IN) && gmt_M_360_range (M_obj->range[XLO], M_obj->range[XHI]) && gmt_M_360_range (G_obj->header->wesn[XLO], G_obj->header->wesn[XHI])) {
+			if (gmt_M_is_geographic (GMT, GMT_IN) && gmt_M_360_range (M_obj->range[XLO], M_obj->range[XHI]) && gmt_M_360_range (S_obj->wesn[XLO], S_obj->wesn[XHI])) {
 				/* Global grids passed via matrix are not rotated to fit the desired global region, so we need to correct the wesn for this grid to match the matrix */
 				gmt_M_memcpy (G_obj->header->wesn, M_obj->range, 4U, double);
 			}
@@ -5549,7 +5549,7 @@ GMT_LOCAL struct GMT_GRID * gmtapi_import_grid (struct GMTAPI_CTRL *API, int obj
 			API->object[new_item]->resource = G_obj;
 			API->object[new_item]->status = GMT_IS_USED;	/* Mark as read */
 			GH->alloc_level = API->object[new_item]->alloc_level;	/* Since allocated here */
-			if (gmt_M_is_geographic (GMT, GMT_IN) && gmt_M_360_range (M_obj->range[XLO], M_obj->range[XHI]) && gmt_M_360_range (G_obj->header->wesn[XLO], G_obj->header->wesn[XHI])) {
+			if (gmt_M_is_geographic (GMT, GMT_IN) && gmt_M_360_range (M_obj->range[XLO], M_obj->range[XHI]) && gmt_M_360_range (S_obj->wesn[XLO], S_obj->wesn[XHI])) {
 				/* Global grids passed via matrix are not rotated to fit the desired global region, so we need to correct the wesn for this grid to match the matrix */
 				gmt_M_memcpy (G_obj->header->wesn, M_obj->range, 4U, double);
 			}
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -5193,12 +5193,13 @@ GMT_LOCAL int gmtapi_export_image (struct GMTAPI_CTRL *API, int object_ID, unsig
 	return (GMT_NOERROR);
 }
 
-GMT_LOCAL bool gmtapi_whole_earth (struct GMT_CTRL *GMT, double we_in[], double we_out[]) {
+unsigned int gmt_whole_earth (struct GMT_CTRL *GMT, double we_in[], double we_out[]) {
 	/* Determines if this is a global geographic grid and we want the whole world, regardless of central longitude */
-	if (!gmt_M_is_geographic (GMT, GMT_IN)) return false;
-	if (!gmt_M_360_range (we_in[XLO],  we_in[XHI])) return false;
-	if (!gmt_M_360_range (we_out[XLO], we_out[XHI])) return false;
-	return true;
+	if (!gmt_M_is_geographic (GMT, GMT_IN)) return 0;
+	if (!gmt_M_360_range (we_in[XLO],  we_in[XHI])) return 0;
+	if (!gmt_M_360_range (we_out[XLO], we_out[XHI])) return 0;
+	if (doubleAlmostEqualZero (we_in[XLO], we_out[XLO])) return 2;	/* Both regions are the same */
+	return 1;	/* Different central meridians */
 }
 
 /*! . */
@@ -5462,7 +5463,7 @@ start_over_import_grid:		/* We may get here if we cannot honor a GMT_IS_REFERENC
 
 			/* Get start/stop row/cols for subset (or the entire domain) */
 			/* dx,dy are needed when the grid is pixel-registered as the w/e/s/n bounds are off by 0.5 {dx,dy} relative to node coordinates */
-			if (!S_obj->region || gmtapi_whole_earth (GMT, M_obj->range, S_obj->wesn)) {	/* Easy, get the whole enchilada */
+			if (!S_obj->region || gmt_whole_earth (GMT, M_obj->range, S_obj->wesn)) {	/* Easy, get the whole enchilada */
 				j0 = i0 = 0;
 				j1 = G_obj->header->n_rows    - 1;	/* Minus 1 since we loop up to and including below */
 				i1 = G_obj->header->n_columns - 1;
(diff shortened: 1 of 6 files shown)


seisman:
I can confirm that https://github.com/GenericMappingTools/pygmt/issues/515 now works for `GMT_IN`, `GMT_IN|GMT_IS_REFERENCE`, and `GMT_IN|GMT_IS_DUPLICATE`. But this PR breaks two PyGMT tests. The two PyGMT tests call grdcut to extract a subregion from a matrix. 

BTW, there is another grdimage bug from PyGMT (https://github.com/GenericMappingTools/pygmt/issues/390). I thought it's caused by the same reason as https://github.com/GenericMappingTools/pygmt/issues/515. But, this PR works for https://github.com/GenericMappingTools/pygmt/issues/515, not for https://github.com/GenericMappingTools/pygmt/issues/390.

PaulWessel:
Could you tell me the grdcut arguments? I assume input is a global grid and the subset is not global?  If so then I am not sure why it would break since it would enter the S_obj->region if-test which was there all along.
Less clear why your Py issue 390 would fail after this PR.  it seems exactly like the case herein, but there must be a difference.

seisman:
> Could you tell me the grdcut arguments? I assume input is a global grid and the subset is not global? If so then I am not sure why it would break since it would enter the S_obj->region if-test which was there all along.
> Less clear why your Py issue 390 would fail after this PR. it seems exactly like the case herein, but there must be a difference.


Here is the core PyGMT code. It reads the 01d pixel-registered earth relief global grid into an xarray, and extract the subregion of the grid.
```python
grid = load_earth_relief(registration="pixel")
grdcut(grid, outgrid=tmpfile.name, region="0/180/0/90")
```

seisman:
> Less clear why your Py issue 390 would fail after this PR. it seems exactly like the case herein, but there must be a difference.

PyGMT issue 390 fails since GMT 6.0, and is worse (crashes or black images) in GMT 6.1, and this PR doesn't fix it.

PaulWessel:
I see the report for 390 about zmax < zmin.  This is puzzling since when we just request header info in grdimage, for a matrix, it scans the entire matrix for min/max to be sure it is set.
I don't think we can keep writing C clones to test every possible case in PyGMT.  We need to learn how to run these python scripts and step into the C code in Xcode or whatever debugger.

PaulWessel:
And the grdcut case uses reference or duplicate?

seisman:
> I see the report for 390 about zmax < zmin. This is puzzling since when we just request header info in grdimage, for a matrix, it scans the entire matrix for min/max to be sure it is set.

It usually means zmin=zmax=0.0.

> I don't think we can keep writing C clones to test every possible case in PyGMT. We need to learn how to run these python scripts and step into the C code in Xcode or whatever debugger.

I already know how to debug PyGMT and C code in Xcode. It's just like what you do with Matlab: run a python console, attach the process id, and run PyGMT codes in the Python console and xcode will stop at the breakpoint. But I don't know where to debug, since the C API is too complicated to understand.

seisman:
> And the grdcut case uses reference or duplicate?

Currently, PyGMT always uses `GMT_IN|GMT_IS_REFERENCE`. 

Both `GMT_IN` and `GMT_IN|GMT_IS_DUPLICATE` work, but sometimes they also crash.

PaulWessel:
OK, let me give it a try then.  I can check out the pyGMT master on macOS.  Can you give me a quick guide for how to install and select my particular GMT build?

seisman:
First of all, you need to install PyGMT dependencies and PyGMT itself. PyGMT will tries to find the system GMT library from the LD_LIBRARY_PATH. 

For debugging purpose, you can just set the environmental variable `GMT_LIBRARY_PATH`, like
```
export GMT_LIBRARY_PATH=~/Gits/gmt/gmt/build/xcode/src/Debug
```
then PyGMT will find the GMT library in this path first.

In the Python console, you can run following commands to verify your GMT library:

```
>>> import pygmt
>>> pygmt.show_versions()
PyGMT information:
  version: v0.1.2+30.gbe75223e
System information:
  python: 3.7.6 (default, Jan  8 2020, 13:42:34)  [Clang 4.0.1 (tags/RELEASE_401/final)]
  executable: /Users/seisman/.anaconda/bin/python
  machine: Darwin-19.6.0-x86_64-i386-64bit
Dependency information:
  numpy: 1.18.1
  pandas: 1.0.1
  xarray: 0.15.1
  netCDF4: 1.5.3
  packaging: 20.1
  ghostscript: 9.50
  gmt: 6.2.0_27ceb6e_2020.08.03
GMT library information:
  binary dir: /Users/seisman/.anaconda/bin
  cores: 8
  grid layout: rows
  library path: /Users/seisman/local/GMT/lib/libgmt.dylib
  padding: 2
  plugin dir: /Users/seisman/local/GMT/lib/gmt/plugins
  share dir: /Users/seisman/local/GMT/share
  version: 6.2.0
```

PaulWessel:
I am looking for even more basic info.  I dont see anything on the pyGMT site of installing anything.  I think I need to install conda which I have not used.  Is there an INSTALL page or similar I can just follow?  Sorry to be such a newbie.

seisman:
PyGMT has an [installation page](https://www.pygmt.org/latest/install.html). I think you can follow the instruction, but when installing dependencies:
```
conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging gmt
```
You can leave the "gmt", because you don't need the gmt package from conda.

PaulWessel:
OK, thanks. Do I first install the full Anaconda or is miniconda OK?

seisman:
> OK, thanks. Do I first install the full Anaconda or is miniconda OK?

miniconda is OK.

PaulWessel:
OK, microconda installed, dependencies installed.  How does python find pygmt on import? Must be some path setting since I get errors if i just run python and enter the import command.

seisman:
> OK, microconda installed, dependencies installed. How does python find pygmt on import? Must be some path setting since I get errors if i just run python and enter the import command.

Run
```
cd pygmt
make install
```

PaulWessel:
```
(base) pwessel@macnut:~/GMTdev/pygmt-> make install
pip install --no-deps -e .
Obtaining file:///Users/pwessel/GMTdev/pygmt
Installing collected packages: pygmt
  Running setup.py develop for pygmt
Successfully installed pygmt
(base) pwessel@macnut:~/GMTdev/pygmt-> python
Python 3.8.3 (default, May 19 2020, 13:54:14) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygmt
  File "<stdin>", line 1, in <module>
  File "/Users/pwessel/GMTdev/pygmt/pygmt/__init__.py", line 15, in <module>
    from .session_management import begin as _begin, end as _end
  File "/Users/pwessel/GMTdev/pygmt/pygmt/session_management.py", line 4, in <module>
    from .clib import Session
  File "/Users/pwessel/GMTdev/pygmt/pygmt/clib/__init__.py", line 8, in <module>
    from .session import Session
  File "/Users/pwessel/GMTdev/pygmt/pygmt/clib/session.py", line 10, in <module>
    from packaging.version import Version
ModuleNotFoundError: No module named 'packaging'

```

seisman:
> (base) pwessel@macnut:~/GMTdev/pygmt-> make install

It seems you're using the base (the default) conda environment. If you follow the PyGMT install instruction, you should have created a new environment called "pygmt".

Just run `conda activate pygmt` to switch to the newly created enviroment and run `make install` again.

PaulWessel:
Sorry, got side-tracked.  OK, verified it found the xcode-built lib.  Trying to run a test by copy paste from test_grdcut.py.  COUld import os, numby but not pytest which is not found.

seisman:
Those tests are designed to be run by pytest. You can use this one instead:
```python
import pygmt

from pygmt.datasets import load_earth_relief

grid = load_earth_relief(registration="pixel")
pygmt.grdcut(grid, outgrid="out.grd", region="0/180/0/90")
```

PaulWessel:
Thanks for your help, up and running in Xcode now.

joa-quim:
> grid = load_earth_relief(registration="pixel")
pygmt.grdcut(grid, outgrid="out.grd", region="0/180/0/90")

Does this mean one have to load the full grid first and only than extract a sub-region?

Tried to reproduce this on Win but it looks my pygmt doesn't work anymore
```
>>> pygmt.show_versions()
  File "<stdin>", line 1, in <module>
AttributeError: module 'pygmt' has no attribute 'show_versions
```
Note, I refuse to install condas to make py things work (and it worked already)

seisman:
> > grid = load_earth_relief(registration="pixel")
> > pygmt.grdcut(grid, outgrid="out.grd", region="0/180/0/90")
> 
> Does this mean one have to load the full grid first and only than extract a sub-region?

For this example, it's yes. But `load_earth_relief` can also accept the *region* (i.e., **-R**) argument, which calls `gmt grdcut @earth_relief_xxy -Rw/e/s/n` and returns a subregion. 

> Tried to reproduce this on Win but it looks my pygmt doesn't work anymore
> 
> ```
> >>> pygmt.show_versions()
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> AttributeError: module 'pygmt' has no attribute 'show_versions
> ```
> 
> Note, I refuse to install condas to make py things work (and it worked already)

Perhaps you're using an old pygmt? `pygmt.show_versions()` was added in PyGMT 0.1.2, which was released a month ago.

joa-quim:
> Perhaps you're using an old pygmt?

Nope, I have only the master source code.

PaulWessel:
@seisman I made a few changes in gmt_api.c and the grdcut test writes a correct out.grd.  Perhaps you can check as well.

seisman:
> > Perhaps you're using an old pygmt?
> 
> Nope, I have only the master source code.

Please try `pygmt.print_clib_info()` first and see the PyGMT version.

joa-quim:
Hmm, nothing about pygmt version
```
>>> pygmt.print_clib_info()
GMT library information:
  binary dir: C:/programs/WPy64-3.7.2/python-3.7.2.amd64
  cores: 12
  grid layout: rows
  library path: C:/programs/GMTdir/gmt_w64.dll
  padding: 2
  plugin dir: C:/programs/GMTdir/gmt_plugins
  share dir: C:/progs_cygw/GMTdev/gmt5/compileds/gmt6/VC14_64/share
  version: 6.2.0
```

seisman:
> Hmm, nothing about pygmt version
> 
> ```
> >>> pygmt.print_clib_info()
> GMT library information:
>   binary dir: C:/programs/WPy64-3.7.2/python-3.7.2.amd64
>   cores: 12
>   grid layout: rows
>   library path: C:/programs/GMTdir/gmt_w64.dll
>   padding: 2
>   plugin dir: C:/programs/GMTdir/gmt_plugins
>   share dir: C:/progs_cygw/GMTdev/gmt5/compileds/gmt6/VC14_64/share
>   version: 6.2.0
> ```

Sorry, run `pygmt.__version__` instead.

joa-quim:
```
>>> pygmt.__version__
'v0.1.2+30.gbe75223'
```

seisman:
> ```
> >>> pygmt.__version__
> 'v0.1.2+30.gbe75223'
> ```

It's weird. You're using the latest version and should be able to run `pygmt.show_versions()`.

seisman:
@PaulWessel I think your changes work for `GMT_IN|GMT_IS_REFERENCE`, but for the other two, sometimes the test script crashes. Here are the debug messages:
```
grdcut [DEBUG]: gmt_get_filename: In: 0/180/0/90 Out: 0/180/0/90
grdcut [DEBUG]: Look for file 0/180/0/90 in /Users/seisman/.gmt
grdcut [DEBUG]: Look for file 0/180/0/90 in /Users/seisman/.gmt/cache
grdcut [DEBUG]: Look for file 0/180/0/90 in /Users/seisman/.gmt/server
grdcut [DEBUG]: Got regular w/e/s/n for region (0/180/0/90)
grdcut [INFORMATION]: Processing input grid
grdcut [DEBUG]: gmtapi_begin_io: Input resource access is now enabled [container]
grdcut [DEBUG]: gmtapi_import_grid: Passed ID = 0 and mode = 1
grdcut [DEBUG]: Grid/Image dimensions imply w/e/s/n = -180/180/-90/90, inc = 1/1, pixel registration, n_layers = 1
grdcut [INFORMATION]: Cartesian input grid, yet x spans exactly 360 and -90 <= y <= 90.
grdcut [INFORMATION]:      To make sure the grid is recognized as geographical and global, use the -fg option
grdcut [DEBUG]: Chosen boundary condition for all edges: geographic
grdcut [DEBUG]: Geographic input grid, longitudes span exactly 360
grdcut [DEBUG]: Object ID 1 : Registered Grid Memory Reference 7f8e9fe73610 as an Input resource with geometry Surface [n_objects = 2]
grdcut [DEBUG]: Successfully created a new Grid container
grdcut [DEBUG]: GMT_End_IO: Input resource access is now disabled
grdcut [DEBUG]: gmtapi_begin_io: Input resource access is now enabled [container]
grdcut [DEBUG]: gmtapi_import_grid: Passed ID = 0 and mode = 2
grdcut [INFORMATION]: Importing grid data from user matrix memory location
grdcut [DEBUG]: Grid is considered to have a 360-degree longitude range.
grdcut [DEBUG]: Chosen boundary condition for left and right edges: geographic
grdcut [DEBUG]: Chosen boundary condition for bottom edge: natural
grdcut [DEBUG]: Chosen boundary condition for top    edge: geographic
grdcut [INFORMATION]: Set boundary condition for top    edge: geographic
[1]    4828 segmentation fault  python test.py
```

joa-quim:
Ah, it's working now when I launched the python console from within ``C:\progs_cygw\GMTdev\gmt5\pygmt``
Before, I was *somewhere else*

PaulWessel:
Sometimes crashes? So if I keep running that grdcut command it will eventually crash? Or do you mean something else?

seisman:
> Ah, it's working now when I launched the python console from within `C:\progs_cygw\GMTdev\gmt5\pygmt`
> Before, I was _somewhere else_

I believe it means that you already have pygmt installed, but it's not in the editable/developer mode. So even you have the latest pygmt codes in your pygmt directory, the installed version is the still old one.

To install pygmt in editable mode, you need to cd into the pygmt directory, and run `pip install --no-deps -e .` to install pygmt (here `-e` means editable mode).

seisman:
> Sometimes crashes? So if I keep running that grdcut command it will eventually crash? Or do you mean something else?

At line 1274 of `pygmt/clib/session.py`, I changed `GMT_IN|GMT_IS_REFERENCE` to `GMT_IN` or `GMT_IN|GMT_IS_DUPLICATE`, and ran the test script above. Sometimes the script runs fine and the output grid is also correct, but sometimes the same script crashes (see the segment fault error above).

PaulWessel:
OK, I changed to duplicate and now no longer the memory leak warnings but no crash after dozens of repeated calls.

PaulWessel:
Part of the problems inthis section has to do with GMT doing file-based i/o for decades and then with the API we added other ways.  Initially (pre-mex) we had no way to test any of it.  Mex found tons of basic bugs once we started to pass grids.  Now we are looking more at passing matrix.  We set it up to handle by reference or duplicate.  Duplicate should work since we can copy values into a new grid. However, there are some limitations on how reference can work since we are treating the matrix as readonly.  Let's review those:

1. The matrix is a global grid and we want the whole grid but maybe with a different central meridian.  This was the reason we added testapi_matrix_360.  So this is a possible use of reference since the projection is a separate step.
2. The matrix is used just as is (no region subset).  This should work fine.
3. A subset of the matrix is used.  This also should work fine.  We simply move the boundary inwards by increasing the pad and changing the wesn. This is reversed when done.

However, these two cases cannot work with a readonly matrix:

1. Want a new region that partly overlaps the matrix and party is outside.  We have no way of moving the pad outwards.
2. You want a subset from a global grid and it crosses the periodic boundary,  E.g., reliefs are -180/180 but if I want a cut from 170 to 190 there is no way to achieve that with pad changes.

The solution to this I think is to recognize those two last cases as early as possible inside gmtapi_import_grid and change the method to duplicate and throw in a goto label to get there.

Am I leaving out any other scenarios, @joa-quim and @seisman ?

seisman:
> OK, I changed to duplicate and now no longer the memory leak warnings but no crash after dozens of repeated calls.

After several successful runs, I finally catch a crash using `GMT_IN`. Here is the screenshot showing where it crashes. Perhaps you can see something?

image

seisman:
Here are the values of the related variables when it crashes:

image

PaulWessel:
OK, this is in the boundary settings.  With duplicate, the grid should have a pad I think.  Just confirming we are still on the grdcut call and this happens inside the call to gmt_BC_init at around line 5486?

seisman:
> OK, this is in the boundary settings. With duplicate, the grid should have a pad I think. Just confirming we are still on the grdcut call and this happens inside the call to gmt_BC_init at around line 5486?

Yes, still using the grdcut example in https://github.com/GenericMappingTools/gmt/pull/3829#issuecomment-668284539.

But it crashes at line 12306, in the `gmt_grd_BC_set` function.

PaulWessel:
The latest commit takes care of the memory leak and testapi_matrix_360_ref runs without leak complaints now.  However, this does not address the grdcut and other issues.

PaulWessel:
Patched a bad bug in the **MATRIX/DUPLICATE** case.  The ij used for the output grid were crazily wrong so your crash is probalby a delayed crash due to stepping on some memory elsewhere.  We now start with output row 0 and not whatever the input row is, and we start with output column 0, not what the input column is.  Could you check if this helps before I return to the **GMT_REFERENCE** case, @seisman ?

seisman:
It still crashes sometimes.

PaulWessel:
I forgot to check this before the latest commit.  It would be good do know if you could try your current version (without the...f6b commit) and run the grdcut with the region="170/190/0/30" and GMT_IS_REFERENCE.  It should not work - may not crash, but look at the grid and see if a grdimage map of it looks good.
The latest commit deals with the cases I discussed earlier.  If we cannot use reference then we switch to duplicate and jump back to the top and try again as duplicate, which will work.

seisman:
> I forgot to check this before the latest commit. It would be good do know if you could try your current version (without the...f6b commit) and run the grdcut with the region="170/190/0/30" and GMT_IS_REFERENCE. It should not work - may not crash, but look at the grid and see if a grdimage map of it looks good.

GMT version: 6.2.0_6e69b36_2020.08.03 + region: "170/190/0/30" + `GMT_IN|GMT_IS_REFERENCE`. The information from grdinfo looks good, but the image (using `gmt grdimage out.grd -Baf -Cgeo -png map`) looks incorrect:

image

> The latest commit deals with the cases I discussed earlier. If we cannot use reference then we switch to duplicate and jump back to the top and try again as duplicate, which will work.

Using the latest commit gives the correct image (compared with `gmt grdimage @earth_relief_01d -R170/190/0/30`:
image

PaulWessel:
For matrix and duplication, we have this section that is meant to determine the min/max of the matrix:

```
if (! (mode & GMT_DATA_ONLY)) {	/* Must first init header and copy the header information from the matrix header */
	gmtapi_matrixinfo_to_grdheader (GMT, G_obj->header, M_obj);	/* Populate a GRD header structure */
	if (mode & GMT_CONTAINER_ONLY) {	/* Just needed the header */
		/* Must get the full zmin/max range since not provided by the matrix header */
		gmt_M_grd_loop (GMT, G_obj, row, col, ij) {
			ij_orig = GMT_2D_to_index (row, col, M_obj->dim);
			api_get_val (&(M_obj->data), ij_orig, &d);
			if (gmt_M_is_dnan (d))
				HH->has_NaNs = GMT_GRID_HAS_NANS;
			else {
				G_obj->header->z_min = MIN (G_obj->header->z_min, (gmt_grdfloat)d);
				G_obj->header->z_max = MAX (G_obj->header->z_max, (gmt_grdfloat)d);
			}
		}
		break;	/* Done for now */
	}
}
```

But does not look right to me: If I call GMT_Read_Data with mode GMT_CONTAINER_AND_DATA then I want to get the header set, including the min/max, as well as reading the data. I think in that case, the current code would skip the min/ax step since the (mode & GMT_CONTAINER_ONLY) would be false.  THe only time we do not want to find min/max is when we just want the data, which means we already asked for the header and min/max got set.  I think the (mode & GMT_CONTAINER_ONLY) test should only determine if we call break:

```
if (! (mode & GMT_DATA_ONLY)) {	/* Must first init header and copy the header information from the matrix header */
	gmtapi_matrixinfo_to_grdheader (GMT, G_obj->header, M_obj);	/* Populate a GRD header structure */
	/* Must get the full zmin/max range since not provided by the matrix header */
	gmt_M_grd_loop (GMT, G_obj, row, col, ij) {
		ij_orig = GMT_2D_to_index (row, col, M_obj->dim);
		api_get_val (&(M_obj->data), ij_orig, &d);
		if (gmt_M_is_dnan (d))
			HH->has_NaNs = GMT_GRID_HAS_NANS;
		else {
			G_obj->header->z_min = MIN (G_obj->header->z_min, (gmt_grdfloat)d);
			G_obj->header->z_max = MAX (G_obj->header->z_max, (gmt_grdfloat)d);
		}
	}
	if (mode & GMT_CONTAINER_ONLY)	/* Just needed the header */
		break;	/* Done for now */
}

```
We should enter that loop if we either just want the header or if we want both header and data.  Does that make sense to you @seisman ?

PaulWessel:
Thanks for checking the 170/190.  Now that other one.  We are getting closer to the end.

PaulWessel:
Also the lines

```
		G_obj->header->z_min = +DBL_MAX;
		G_obj->header->z_max = -DBL_MAX;
```
	
needs to be inside that main if-test, otherwise the min/max get reset when we come to get the data...

seisman:
I think you're right with https://github.com/GenericMappingTools/gmt/pull/3829#issuecomment-668365091.

PaulWessel:
I've made some additional revisions to this part of the code, especially when we grab the entire matrix.  I also applied what we have learned for the matrix duplicate case to the grid duplicate case which I dont think we are using but it was clearly wrong, just like the matrix initially was clearly wrong.  Let me know if there is any changes from this on the python side.  The tests pass on command line.

seisman:
I think your changes have fixed most issues with matrix.

Here is one which doesn't make sense to me:
```
import pygmt
from pygmt.datasets import load_earth_relief

grid = load_earth_relief()
fig = pygmt.Figure()
fig.grdimage(grid, region='g', projection='Q0/0/6i', cmap='geo', frame='a90f30g30')
fig.grdimage(grid, region='g', projection='Q180/0/6i', cmap='geo', X='7i', frame='a90f30g30')
fig.savefig("earth.png")
```
The figures are the same, but the longitudes are wrong:
image

PaulWessel:
Will have a look.  I just added another update but unlikely to affect the case above.

PaulWessel:
And this is with REFERENCE in session,py?

seisman:
REFERENCE and DUPLICATE give the same wrong results

PaulWessel:
Soemthing else I need to set?

>>> fig.grdimage(grid, region='g', projection='Q0/0/6i', cmap='geo', frame='a90f30g30')
grdimage [ERROR]: File /Users/pwessel/opt/miniconda3/envs/pygmt/share/geo.cpt was not found

seisman:
Yes, when debugging, it seems GMT incorrectly determines the GMT_SHAREDIR, so you need to change GMT_SHAREDIR to the correct GMT/share directory.

PaulWessel:
The 180 plot does not work because in this case there is no need for a projection but I cannot rotate the grid without projecting.  So need to find a way to set need_to_project to true when the grid is by reference.

PaulWessel:
Your example works for now now.   Let me know if this branch can be merged and I will remove the WIP.

seisman:
Find one more crash with `GMT_IN|GMT_IS_REFERENCE`, but it works for `GMT_IN` or `GMT_IN|GMT_IS_DUPLICATE`.

The script reads the file `@tut_relief.nc` into xarray and pass it to grdimage:
```python 
import pygmt
import xarray as xr

pygmt.show_versions()

fig = pygmt.Figure()

fname = pygmt.which("@tut_relief.nc", download='a')
dem = xr.open_dataarray(fname)

fig.grdimage(dem, frame='af', projection='M6i', cmap='geo', shading=True, V='d')

fig.savefig("shading.png")
```

PaulWessel:
Unable to reproduce any crash - plot looks good.  Intermittent?

seisman:
It always gives me a crash. 

Anyway, I think this PR already fixes the PyGMT issues 390 and 515, and this PR is already very large. 

I think we can merge this PR first. Later I can provide a detailed bug report for the crash above.

PaulWessel:
Yes, I was going to suggest something like that as well. Better to focus on a specific - this PR solves many things but possibly not all.

PR #3841, opened 2020-08-05
Enable the GMT build script to build any branches or tags

### LINKED ISSUE #3837: Rename and improve the build-gmt-master.sh script

Currently, we provide the [ci/build-gmt-master.sh](https://github.com/GenericMappingTools/gmt/blob/master/ci/build-gmt-master.sh) script for building GMT master branch. It's not used by GMT, but mostly useful for external projects. AFAIK, the script is already used in https://github.com/GenericMappingTools/pygmt and https://github.com/GenericMappingTools/custom-supplements. 

The biggest limitation of the script is that it can only build GMT master branch. It's really easy to make it support building a specific branch (master, 6.1 or any others), a specific release (6.1.0 or 6.0.0) or even a specific commit (the full commit hash is needed). 

I think we can:

1. Allow the script to build any branch, release or commit. It still builds the master branch by default
2. Rename the script to `ci/build-gmt.sh`

seisman:

Add one more environmental variable `GMT_GIT_REF` to the building script, so that it can build any branches or tags. Some example values of `GMT_GIT_REF` are:

- master [the default value]
- 6.1 (build the 6.1 branch)
- 6.1.0  (build the 6.1.0 release)
- et. al.

~The only issue is that, the build script uses the tarballs provided by GitHub, which don't have any git history information. So the GMT version string doesn't contain commit hash and commit date. For example, building the master branch gives you `6.2.0` instead of `6.2_b234d76_2020.08.04`.~

The script is also renamed from `build-gmt-master.sh` to `build-gmt.sh`, so https://github.com/GenericMappingTools/custom-supplements and https://github.com/GenericMappingTools/pygmt should make changes after this PR is merged.

Fixes #3837

### DIFF
--- a/ci/build-gmt.sh
+++ b/ci/build-gmt.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Build the GMT source codes of a specific branch, version or commit
+# Build the GMT source codes of a specific branch or tag/release
 #
 # Usage:
 #
@@ -12,7 +12,7 @@
 # Environmental variables that affect the building process:
 #
 # - GMT_INSTALL_DIR : GMT installation location [$HOME/gmt-install-dir]
-# - GMT_GIT_REF     : branch name, version, or commit full sha [master]
+# - GMT_GIT_REF     : branch name or tag/release [master]
 #
 
 set -x -e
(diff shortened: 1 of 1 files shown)


seisman:
@weiji14 Since the script is used by PyGMT. Please see if you have any comments here.

ci/build-gmt.sh
```diff
@@ -30,18 +34,19 @@ GMT_BUILD_TMPDIR=$(mktemp -d ${TMPDIR:-/tmp/}gmt.XXXXXX)
 cd ${GMT_BUILD_TMPDIR}
 
 # 2. Download GMT, GSHHG and DCW from GitHub
-git clone --depth 1 https://github.com/GenericMappingTools/gmt.git gmt
+curl -SLO https://github.com/GenericMappingTools/gmt/archive/${GMT_GIT_REF}.${EXT}
```
weiji14, line 37: > The only issue is that, the build script uses the tarballs provided by GitHub, which don't have any git history information. So the GMT version string doesn't contain commit hash and commit date. For example, building the master branch gives you 6.2.0 instead of 6.2_b234d76_2020.08.04

It would be better to have the commit hash. I haven't tried it before, but maybe try `git clone --single-branch --branch <branchname> <remote-repo>` to clone a single branch? See https://stackoverflow.com/questions/1911109/how-do-i-clone-a-specific-git-branch.

seisman: OK, will see if a shallow clone is possible for a single branch or tag.

seisman: Done. Now using:
```
git clone --depth=1 --single-branch --branch ${GMT_GIT_REF} https://github.com/GenericMappingTools/gmt
```
GMT_GIT_REF can be a branch name or a tag, NOT a commit hash (but most of the time we don't need to build a specific commit).

PaulWessel:
How about that option for using $1?

seisman:
> How about that option for using $1?

Yes, I will add it, too. So the branch/tag name can be specified either using the command line argument or by the environmental variable.

PR #3848, opened 2020-08-05
Reading/writing a matrix from/to a grid file

PaulWessel:

This PR started out with trying to generalize _GMT_Read_Data_ for **GMT_IS_MATRIX**.  Now, if you pass the geometry **GMT_IS_SURFACE** then we read the file as a grid but return a matrix.  The conversion between grid to matrix is simple (we do this in the general import_grid anyway) but it pointed out some issues when testing via a short C code starting a session: Since no modules were involved, the remote files were unknown.  So now we added a check to refresh the server if the list is NULL.  The test codes now pass remote file names and **GMT_IS_SURFACE** and this works fine.
Then the same was applied to _GMT_Write_Data_, i.e., ability to write a **MATRIX** (camouflaging as a grid) to a grid file.  This exposed another problem in gmt_nc.c in that the file-specific chunking settings were not reset, causing a second _GMT_Write_Data_ to fail.
I have updated a few of the testapi*.c codes to indicate the geometry (**GMT_IS_POINT** if just a dumb text table and **GMT_IS_SURFACE** if a real grid), and updated some of the PS originals that changed due to the roundoff involved in the text version.
A new test program testapi_grid2matrix.c was added to test the reading of a grid into a matrix and then writing the matrix out as a grid, and the new test apimat2grd.sh runs this and makes a plot.

Because there are several bug fixes involved in this PR I have labeled this as a bug fix which requires backport.


seisman:
This PR breaks a PyGMT test, but maybe we need to update PyGMT.

The PyGMT does following things:

1. Create a grid by using 
```
family="GMT_IS_GRID|GMT_VIA_MATRIX", 
geometry="GMT_IS_SURFACE", 
mode="GMT_CONTAINER_ONLY"
```
2. Attach a matrix to the dataset by calling GMT_Put_Matrix
3. Call GMT_Write_Data with parameters
```
family="GMT_IS_MATRIX",
geometry="GMT_IS_SURFACE",
mode="GMT_CONTAINER_AND_DATA",
```

Before this PR, it writes a table/matrix; after this PR, it writes a netCDF. Is this what we want?

With this PR, to write a table data, I should use `GMT_IS_POINT`, right?

PaulWessel:
Yes, passing geometry as GMT_IS_POINT will case it to read or write an ascii table, while GMT_IS_SURFACE ends up calling import|export_grid.  So this is backwards incompatible to some extent.

seisman:
This is a good backwards incompatible to PyGMT, but I'm not sure if GMT.jl likes it.

PaulWessel:
We need to check with @joa-quim to see if GMT.jl uses _GMT_Read_Data_ directly on a text file to read in a **GMT_IS_MATRIX**.  I don't think he is, but we will wait for confirmation. Depending on how long he is banned from checking email, this could be 10 days.  Meanwhile, I will update the API docs.

joa-quim:
Leave only tomorrow.

I use 
```
M = GMT_Create_Data(API, GMT_IS_MATRIX|GMT_VIA_MATRIX, GMT_IS_PLP, 0, dim, C_NULL, C_NULL, 0, 0, C_NULL)
actual_family[1] = actual_family[1] | GMT_VIA_MATRIX
```
but not ``GMT_Read_Data``

PaulWessel:
OK, so it will not affect GMT.jl since GMT modules (other than a few testapi_*.c programs) never called _GMT_Read_Data_ with intent of returning a matrix or sorts.  So if @seisman is OK with this being merged then it might as well be approved now?

Issue #3844, opened 2020-08-05
Matrix as grid with changing central meridian doesn't work well for gridline grids


seisman:
**Full script that generated the error**

The script generates 4 maps, for pixel or gridline grids passing as grid or matrix:
```python
import pygmt
from pygmt.datasets import load_earth_relief
import xarray as xr

# pixel grid
fig = pygmt.Figure()
fig.grdimage("@earth_relief_01d_p", projection="W120/15c", cmap="geo")
fig.savefig("grid_p.png")

# pixel matrix
grid = load_earth_relief(registration="pixel")
fig = pygmt.Figure()
fig.grdimage(grid, projection="W120/15c", cmap="geo")
fig.savefig("matrix_p.png")


# gridline grid
fig = pygmt.Figure()
fig.grdimage("@earth_relief_01d_g", projection="W120/15c", cmap="geo")
fig.savefig("grid_g.png")

# gridline matrix
grid = load_earth_relief(registration="gridline")
fig = pygmt.Figure()
fig.grdimage(grid, projection="W120/15c", cmap="geo")
fig.savefig("matrix_g.png")
```

Use the command below to compare the two maps of pixel-registered grid/matrix. The RMS is zero. So passing grid or matrix give the same results:
```
gm compare -density 200 -highlight-color magenta -highlight-style assign -metric rmse -file diff_p.png matrix_p.png grid_p.png
```

Use the command below to compare the two maps of pixel-registered grid/matrix.
```
gm compare -density 200 -highlight-color magenta -highlight-style assign -metric rmse -file diff_g.png matrix_g.png grid_g.png
```
The RMS is non-zero. Here is the diff image. There are tiny difference along the x=180 line. Perhaps a 0.5 grid offset?

![image](https://user-images.githubusercontent.com/3974108/89378026-f9215e80-d6c0-11ea-8540-5c06bc03c02a.png)

PaulWessel:
Will look tomorrow.  I know the issue: When we rotate the central longitude the repeated columns at -180 and 180 need to collapse to one.

PaulWessel:
So 1-2 days go by, an automatic OS update happened over night, so GMT_LIBRARY_PATH was unset.  However, I am missing another step before I can sy import pygmt, something about switching from base to pygmt or a setup step?

seisman:
Yes, you need to activate your conda environment by `conda activate pygmt`.

PaulWessel:
??? My outputs look like this (with master):

[...]

PR #3849, opened 2020-08-05
Add special check for non-rotated global grids

PaulWessel:
We usually read global grids from file and they get rotated depending on the wesn selection.  However, when a global grid is passed from another environment, as **GMT_IS_REFERENCE**, we may be projecting that image in _gmt_grd_project_ and if the repeated meridians fall inside the map (since no rotation has happened) they are counted twice, yielding a slightly different result.  Only applies to gridline-registrered grids.  This fix detects this special situation and skips one of the repeated meridians. Addresses the initial issue raised in #3844.
Note sure about backport.  Was the gmt_whole_earth backported?

### DIFF
--- a/src/gmt_map.c
+++ b/src/gmt_map.c
@@ -7966,7 +7966,7 @@ int gmt_grd_project (struct GMT_CTRL *GMT, struct GMT_GRID *I, struct GMT_GRID *
 			 * This is only likely to happen when external global grids are passed in via GMT_IS_REFERENCE. */
 			skip_repeat = true;	/* Since both -180 and +180 fall inside the image, we only want to use one of then (-180) */
 			duplicate_east = gmt_M_is_periodic (GMT);	/* Since the meridian corresponding to map west only appears once we may need to duplicate on east */
-			if (duplicate_east)
+			if (duplicate_east)	/* Find the column in I->data that corresponds to the longitude of the left boundary of the map */
 				duplicate_col = gmt_M_grd_x_to_col (GMT, GMT->common.R.wesn[XLO], I->header);
 		}
 
@@ -8004,13 +8004,14 @@ int gmt_grd_project (struct GMT_CTRL *GMT, struct GMT_GRID *I, struct GMT_GRID *
 			}
 			if (duplicate_east) {
 				ij_in = ij_in - I->header->n_columns + duplicate_col;	/* Rewind to the col to be duplicated */
+				col_in = duplicate_col;
 				if (gmt_M_is_rect_graticule (GMT))
 					x_proj = GMT->current.proj.rect[XHI];
 				else if (inverse)
 					gmt_xy_to_geo (GMT, &x_proj, &y_proj, x_in[col_in], y_in[row_in]);
 				else {
 					if (GMT->current.map.outside (GMT, x_in[col_in], y_in[row_in])) continue;	/* Quite possible we are beyond the horizon */
-					gmt_geo_to_xy (GMT, GMT->common.R.wesn[XLO], y_in[row_in], &x_proj, &y_proj);
+					gmt_geo_to_xy (GMT, GMT->common.R.wesn[XHI], y_in[row_in], &x_proj, &y_proj);
 				}
 
 				/* Here, (x_proj, y_proj) is the projected grid point.  Now find nearest node on the output grid */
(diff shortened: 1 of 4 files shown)


seisman:
This PR should be backported.

Please see if you could still improve this PR:

Here are diff images generated by matplotlib image comparison functions. I find then more readable than the gm diff images.

This is from GMT master branch:
image

This is from this PR:
image

The central meridian is fixed by this PR, but please see if you can reduce the right side differences.

PaulWessel:
Not sure.  It was relatively clear what to do for when the repeated meridian lands inside.  While I am not sure why there is  nothing on the west border, I would suspect this has to do with BCs.  When the grid is read in and we know it is global, we use actual data values from the other side to set the two extra rows/cols in the grid data.  THe matrix has no padding, so when it is passed into _gmt_grd_project_, this happens:

```
/* Only input grid MUST have at least 2 rows/cols padding - otherwise we must allocate a temp grid */
if (I->header->pad[XLO] < 2 || I->header->pad[XHI] < 2 || I->header->pad[YLO] < 2 || I->header->pad[YHI] < 2) {
	unsigned int pad2[4] = {2, 2, 2, 2};
	GMT_Report (GMT->parent, GMT_MSG_INFORMATION, "gmt_grd_project: Input grid has insufficient padding - create and work on a duplicate with r row/col pad\n");
	if ((I2 = GMT_Duplicate_Data (GMT->parent, GMT_IS_GRID, GMT_DUPLICATE_DATA, I)) == NULL) {
		GMT_Report (GMT->parent, GMT_MSG_ERROR, "gmt_grd_project: Unable to duplicate grid\n");
		return GMT_RUNTIME_ERROR;
	}	
	gmt_grd_pad_on (GMT, I2, pad2);	/* Add pad */
	gmt_BC_init (GMT, I2->header);	/* Initialize grid interpolation and boundary condition parameters */
	gmt_grd_BC_set (GMT, I2, GMT_IN);	/* Set boundary conditions */
	I = I2;	/* Use this input grid instead */
}
```

So we end up duplicating the non-padded matrix so we can project.  So assuming the grid/matrix is flagged as geographic, it should end up applying periodic boundary conditions at west and east and it should be the same for the grid as well as the matrix.  But clearly the east boundary differs.  Will see if I can find any inconsistencies.

PaulWessel:
Think I have the answer: It is the opposite of the fix I added.  For this projection and a gridline-registered grid, the longitude that corresponds to the map west and east will be plotted twice.  Since  the rotated grid (read from file) has a min/max that equal east/west, it is plotted twice and all is well.
For the **GMT_IS_REFERENCE** grid that is not rotated, the repeated -180 and +180 meridian both project to the the interior of the image (unless we use **-Rd**).  So the earlier fix was to only project one of them and that fixed the reported problem.  However, in the above case, w=-60 and e=300 but the duplicated longitude is still -180/180 for a referenced grid. That means that when we plot the single meridional line for west=-60 we must actually add 360 to its longitude and do it again - otherwise there is nothing projected at the east boundary.  Will try to add such a fix.

PaulWessel:
So I would like to debug my new algorithm but now I cannot get the python script to pass control to Xcode when it calls `fig.grdimage(grid, projection="W120/15c", cmap="geo")`.  Despite making sure I have GMT_LIBRARY_PATH=/Users/pwessel/GMTdev/gmt-dev/xbuild/src/Debug as well as GMT_SHAREDIR=/Users/pwessel/GMTdev/gmt-dev/dbuild/gmt6/share, I run conda activate pygmt, then open up the python terminal. attach Xcode to the process (Python, the only one) and set stop points in Xcode in gmt.c, it just finishes.  I have rebooted, rebuilt, etc.  No go.  This worked yesterday, so I must be missing something.  My steps after reboot:

1. export GMT_LIBRARY_PATH=/Users/pwessel/GMTdev/gmt-dev/xbuild/src/Debug
2. export GMT_SHAREDIR=/Users/pwessel/GMTdev/gmt-dev/dbuild/gmt6/share
3. conda activate pygmt
4. python
5. Run the first few commands (import pygmt, etc)
6. Attach Xcode to process python.
6. Run fig.grdimage(grid, projection="W120/15c", cmap="geo")

but it just completes.  Plot is made, but I am unable to debug.  Any ideas?

seisman:
> set stop points in Xcode in gmt.c

Do you mean "gmt_api.c"?

PaulWessel:
Hm, perhaps that is the problem. I am still in my CLI module being called by gmt but I need to think GMT_Call_Module and put a stop point inside Call_Module.

PaulWessel:
All is well.  Another lesson learned on debugging.  99% of my debugging has been via gmt calling GMT_Call_Modules so that has been my entry point into the library...

PaulWessel:
Please give this another shot.  My gm compare show zero difference between the two gridline plots and all tests pass (well minus the gdalnn and the imagepluscpt).

PR #3880, opened 2020-08-08
Better documentation of grid-registration switching

PaulWessel:
Move the figure from grdsample to the option -r section, and add cross-links to/from grdsample and grdedit, and explain the difference between destructive and non-destructive changes.


PaulWessel:
Thanks, I did and the links work so we will merge.

PR #3896, opened 2020-08-08
Improve the plot man page for symbols

PaulWessel:
Improve the organization by breaking the symbols into different groups and showing simple plots of what they look like.  Starting with the simplest and ending with the more complicated.

### DIFF
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -524,6 +524,14 @@
    between a *front* and *quoted* lines in that it offers symbols similar to a front
    but the placement can be specified in ways similar to the quoted line:
 
+.. figure:: /_images/GMT_base_symbols9.*
+   :width: 600 px
+   :align: center
+
+   Decorated lines with eleven squares evenly spaced along the line.  By default,
+   the symbol is aligned with the line trend, but numerous modifiers allow you to
+   customize the appearance.
+   
 .. _-basic_symbols9:
 
     **-S~**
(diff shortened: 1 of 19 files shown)


PaulWessel:
Think this is good for a review, @seisman; will probably have to build the docs given the new figures.

seisman:
@PaulWessel I just made some changes in commit 60555881fadbbde9e49a8f59d580dbe7c22c7971 to fix the issue you mentioned in the email.

This PR looks great to me, but there are some issues with other related files.

`explain_symbols2.rst_` and `explain_symbols_only.rst_` are very similar to `explain_symbols.rst_`. `explain_symbols2.rst_` is inlcude in plot3d, `explain_symbols_only.rst_` is used in ternary. These two files are not well maintained. Can we keep only one file, and add cross links to the **plot -S** option when necessary and explain the differences?

PaulWessel:
Yes, I think that is a good idea.  Otherwise they start to diverge as we fail to update similar pages. WOUld it work to have more include files, i.e., consolidate the basic symbols in a basic_symbol.rst_ file that is then included where needed (plot, plot3, ternary) so it is only written once.  I think having ternary use the full page with pages and pages of quoted lines etc is too much and just confusing.  So the file included in ternary is also included in plot and plot3, bot those have more included files doing the other types of symbols?

seisman:
> WOUld it work to have more include files, i.e., consolidate the basic symbols in a basic_symbol.rst_ file that is then included where needed (plot, plot3, ternary) so it is only written once

Yes, nested `include` is supported.

PaulWessel:
OK.

PR #3911, opened 2020-08-10
Accept deprecated CN.## numbers and look up corresponding codes

PaulWessel:
Since DCW 2.0.0 will use updated Chinese province codes instead of numbers, it is only reasonable that we handle this as a backwards support case now.  This PR should do that - but needs testing.  I have updated the 2.0.0 tar/zip packages on my release ftp site.  Please see if it works or report problems.  This gives me some confidence in the fix:

pw```
essel@macnut:~-> gmt coast -RCN -JM6i -B -Glightgray -ECN.41+p1p -pdf map
coast [NOTICE]: FYI, Chinese province code 41 is deprecated. Use HA instead

```
and the plot looks fine.

### DIFF
--- a/src/gmt_dcw.c
+++ b/src/gmt_dcw.c
@@ -204,7 +204,7 @@ GMT_LOCAL int gmtdcw_find_country (char *code, struct GMT_DCW_COUNTRY *list, int
 	return (low);
 }
 
-GMT_LOCAL int gmtdcw_find_state (struct GMT_CTRL *GMT, char *scode, char *ccode, struct GMT_DCW_STATE *slist, int ns) {
+GMT_LOCAL int gmtdcw_find_state (struct GMT_CTRL *GMT, char *scode, char *ccode, struct GMT_DCW_STATE *slist, int ns, bool check) {
 	/* Return state id given country and state codes using a linear search */
 	int i;
 	static struct GMT_DCW_CHINA_CODES gmtdcw_CN_codes[DCW_N_CHINA_PROVINCES] = {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by PaulWessel** [comment]
I fixed a few things and also improved the message to avoid saying

coast [WARNING]: Country CN does not have a state named 10 (skipped)

Now we get

coast [WARNING]: Country CN does not have a state named 10 (skipped)

File: src/gmt_dcw.c [hunk 1]
```diff
@@ -197,10 +204,24 @@ GMT_LOCAL int gmtdcw_find_country (char *code, struct GMT_DCW_COUNTRY *list, int
 	return (low);
 }
 
-GMT_LOCAL int gmtdcw_find_state (char *scode, char *ccode, struct GMT_DCW_STATE *slist, int ns) {
+GMT_LOCAL int gmtdcw_find_state (struct GMT_CTRL *GMT, char *scode, char *ccode, struct GMT_DCW_STATE *slist, int ns) {
 	/* Return state id given country and state codes using a linear search */
 	int i;
-
+	static struct GMT_DCW_CHINA_CODES gmtdcw_CN_codes[DCW_N_CHINA_PROVINCES] = {
```
seisman, line 210: Need to check DCW version here. For GMT>=6.1.1 + DCW<=1.1.4, `CN.11` is converted to `CN.BJ`, but `CN.BJ` is not available in DCW<=1.1.4

PaulWessel: That is a good point!.  See new commit.

PR #3939, opened 2020-08-12
Add option to specify the debug pen in helping with contour crossing lines

PaulWessel:
The **+d** for debug crossing lines for contours and decorated lines can now take an optional pen argument.  This was needed because in the periodic.sh test I could not tell the helper line from a gridline since they used similar pens.

The periodic.sh test script still has issues but I updated the comments.

Not sure if it is worth backporting @seisman so I did not label it.  Won't be fixed before 6.2 anyway.

### DIFF
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -402,7 +402,8 @@
 
             **+d**
                 Turns on debug which will draw helper points and lines to illustrate
-                the workings of the quoted line setup.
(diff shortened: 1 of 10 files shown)


doc/rst/source/explain_symbols.rst_
```diff
@@ -402,7 +402,8 @@
 
             **+d**
```
seisman, line 403: `+d` or `+d[pen]`?

Issue #3942, opened 2020-08-12
A simpler way to count values inside a circle


PaulWessel:
**Description of the desired feature**

Requests such as [this](https://forum.generic-mapping-tools.org/t/number-of-datapoints-area/765/2) come up from time to time.  It is a reasonable request: Given a table of random x,y[,z] values, make a grid that counts how many point falls inside a circle of some _radius_ centered on each grid node, or what the sum is.  I do not think we have a simple answer.  The one I gave is basically this recipe:

1. Decide on your grid specifications and make a dummy empty grid with **grdmath**
2. Dump the grid with **grd2xyz** to ascii
3. Write a script that loops over the records in this list of nodes
4. For each record, dump it to a point.txt file
5. Run **gmtselect** **-C**point.txt **+d**_radius_ on the whole dataset, then count the number of records found
6. Add the node location and the count to a growing output table
7. Run **xyz2grd** on that table to make the final grid

Clearly, this is only easy for gurus.  Am I missing a simpler approach?  If not, then I think such a list of steps require a C solution.  Various options:

- Existing modules **nearnneighbor** and **grdmask** have much of the machinery needed and could do the job given a new option.
- Alternatively, a new module, maybe **gmtcount**, could implement the steps.

I imagine the usual modifications: You may want the count, the sum, and there may be a 4th column with weights, or perhaps you want the mean, median or mode, or some other statistic.  Maybe you want to normalize the result by the area of the circle.  The "get points inside geographic circle" step gets used in many modules, e.g., **grdfilter**, but it reads a grid.  If we wanted to handle all the potential desired results then I think a new option to an existing tool will be very cumbersome.  So, maybe **gmtcount** is the best suggestion?

gmt count table **-R**_region_ **-I**_inc_ [**-r**] **-C**a|m|p|l|L|u|U|n|s [**-W**] **-G**_gridfile_ **-N** **-S**_radius_[**u**]

PaulWessel:
As you might have guessed, I am about to finish the work on this module, @dongdong.  I have not yet settled on a name.  The process can compute a whole long list of statistics from the data inside the circle, from sum, n, mean, median, etc to std, MAD, inter-quartile range, etc.  I think this will be a very useful module.  It cannot be part of **gmtspatial** since it is completely full of options, but it is a spatial tool that does statistics.  **gmtcount** is a bit too specific and count is only one of 15 types of outputs.  Maybe

- gmtstatistic or gmtstats
- gmtcount
[...]

PR #3950, opened 2020-08-13
New module gmtbinstats for statistical summaries inside circles, hexagons, rectangles on a grid

PaulWessel:

This new module **gmtbinstats** follows the suggestion in #3942 by implementing those ideas.  It offers 17 different operators and most of them can be used with or without weights.  The results can be normalized by the search circle area for density values.  I have documented it and added two test scripts that run most of the operations.  Attached is the result of the unweighted result by gridding the populations of the World's capitals:

nonweighted

The tags referes to the -Ccode used, with (**a** = mean, **d** = MAD, **s** = stdev, **i** = interquartile range, **l** = minimum, **m** -= median, **n** = count, **o** = LMS scale, **p** = mode, **q75** the 75% quartile (any quantile is allowed), **g** = full range (max-min), **z** = sum.

It was quick to put together since it inherited most of the i/o machinery from **nearneighbor** and most of the algorithms are inherited from **gmtmath** and **grdmath**, so it is mostly tested indirectly.  Looking at the results the relationship between the various maps seems sensible.


PaulWessel:
Some users already giving the new module a [spin](https://forum.generic-mapping-tools.org/t/number-of-datapoints-area/765/11).

PaulWessel:
Added support for hexagonal binning via **-T.**  This requires **-I** to set the y-increment and then dx is set to sqrt(3) * dy as required by the geometry.  We may adjust xmax for the region to be inclusive.  Below is what it may look like, and the yellow data points are properly assigned to the correct hexagons.  Two new tests were added to demonstrate the hexagonal tiling.

hexgrid

src/gmtcount.c
```diff
@@ -0,0 +1,935 @@
...
+		case GMTCOUNT_SUM: case GMTCOUNT_COUNTW:
+			G->data[ij] += in[GMT_Z];
+			break;
+		default:	/* count */
+			G->data[ij] += 1.0;
+			break;
+	}
+}
+
+bool outside_hexagon (struct GMT_CTRL *GMT, double x0, double y0, double x, double y, double distance, double radius_inner) {
```
joa-quim, line 381: This is a bool function returning an int. I had recently hard to find bug in a similar situation and VS2019.

The binning could perhaps be speed up if points are first checked to fall inside the circumscribed rectangle. Also remember that each **GMT_Report()** is an *if* branch, and there seems to be one for each input data point.

PaulWessel: OK, fixing the bool.  The GMT_Report was mostly there for me during debugging.  I can place them under #ifdef DEBUG since just there for keeping.
I will look at **gmtselect** for the pre-circle rectangular skip.  But implementing -Tr for rectangles anyway.  For geographic we can do a check on latitude, but longitude by itself cannot work with -S.

PaulWessel: latitude checks for rectangle is also tricky across poles.

PaulWessel:
Hi @joa-quim and @seisman.  I think this module is ready to be introduced.  Ideally, you'd try it on some other data to try to break it.

doc/rst/source/gmtbinstats.rst
```diff
@@ -0,0 +1,201 @@
...
+**-S**\ *search_radius*
+    Sets the *search_radius* that determines which data points are
+    considered close to a node. Append the distance unit (see `Units`_).
+    Not compatible with **-T**.
+
+.. _-T:
+
+**-T**\ [**h**\|\ **r**]
+    Instead of circular, possibly overlapping areas, select non-overlapping tiling.  Choose between
+    **r**\ ectangularor **h**\ exagonal binning. For **-Tr**, set bin sizes via **-I** and we write
```
weiji14, line 113:
```suggestion
    **r**\ ectangular **h**\ exagonal binning. For **-Tr**, set bin sizes via **-I** and we write
```

doc/rst/source/supplements/module_supplements_purpose.rst_
```diff
@@ -30,6 +30,12 @@
 
 .. |mgd77track_purpose| replace:: Plot track-lines of MGD77 cruises
 
+.. |fzanalyzer_purpose| replace:: Analysis of fracture zones using crossing profiles
+
+.. |fzblender_purpose| replace:: Produce a smooth blended FZ trace
+
+.. |mlconverter_purpose| replace:: Convert chrons to ages using selected magnetic timescale
+
```
weiji14, line 38: These seem unrelated?

doc/rst/source/gmtbinstats.rst
```diff
@@ -0,0 +1,201 @@
...
+-----------
+
+**gmtbinstats** reads arbitrarily located (x,y[,z][,w]) points
+(2-4 columns) from standard input [or *table*] and for each
+node in the specified grid layout determines which points are
+within the given radius.  These point are then used in the
+calculation of the specified statistic. The results may be
+presented as is or may be normalized by the circle area to
+perhaps give density estimates.  Alternatively, select
+hexagonal tiling instead of a rectangular grid layout.
```
weiji14, line 50:
```suggestion
presented as is or may be normalized by the circle area to
perhaps give density estimates.  Alternatively, select
hexagonal tiling instead or a rectangular grid layout.
```

Not sure if I understand this sentence correctly. Is the circle area based on a rectangular grid, and there is an alternative hexagonal tiling scheme? Or are there circular 'tiles', hexagonal tiles, and rectangular grid tiles.

weiji14 (commented):
Exciting stuff! I haven't got time to test this out now, but look forward to trying it once I have time! Just spotted a few typos to fix.

doc/rst/source/supplements/module_supplements_purpose.rst_
PaulWessel: yes, a personal supplement - will need to check why but removed for now.  Thanks for the fixes!

doc/rst/source/modules.rst
```diff
@@ -47,6 +47,7 @@ All modules are requested via a call to the :doc:`gmt` program.
    gmt2kml
    gmtconnect
    gmtconvert
+   gmtbinstats
```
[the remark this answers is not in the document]
PaulWessel: Done

PR #3967, opened 2020-08-15
Specify x-axis in sript

PaulWessel:

Only apply to the axis in question.


seisman:
Other GMT_-B_time*.sh scripts also need change.

PR #3979, opened 2020-08-16
Update ex22.rst

Esteban82:
The seismicity is for the last month, not the las 7 days. 

Also, now GMT can download the data directly from the URL, so I think that maybe this could be clarify and included in the scripts.

### DIFF
--- a/doc/rst/source/gallery/ex22.rst
+++ b/doc/rst/source/gallery/ex22.rst
@@ -1,6 +1,6 @@
 .. _example_22:
 
-(22) World-wide seismicity the last 7 days
+(22) World-wide seismicity the last month
 ------------------------------------------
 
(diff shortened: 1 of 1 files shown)


seisman:
> Also, now GMT can download the data directly from the URL, so I think that maybe this could be clarify and included in the scripts.

Yes, feel free to open another PR for it.

seisman:
@Esteban82 Just realize that you chose the 6.1 branch as the base branch. Usually that's not what we want. PRs should be merged into master branch first, and then backported to 6.1 branch if necessary. For future PRs, please work on the master branch instead, unless the bug you're trying to fix is only on 6.1 branch.

As for this PR, I'll cherry-pick it to master branch manually.

Esteban82:
Ok, I didn't realized that. I will check the branch the next time.

Issue #3983, opened 2020-08-16
A couple of halph bugs in plot3d


joa-quim:
1- The fill should extend to the vertical walls
2- The *x,y* gridlines as well
3- The cube should be open (new option to not show the facing two faces)
4- Z annot labels should be horizontal by default (probably only in a new theme to not break compat)

```
gmt basemap -JX12c/0 -Bafg -Bzag -BWSenZ+b+ggray -p210/30 -JZ10 -R-2/2/-1/3/0/4 -png map
```

![map](https://user-images.githubusercontent.com/537321/90344564-be012280-e012-11ea-9fb9-74c00c97161b.png)

PaulWessel:
> 1- The fill should extend to the vertical walls

Always, or as options
> 2- The _x,y_ gridlines as well

That should be doable
> 3- The cube should be open (new option to not show the facing two faces)

I think the **+b** modifier could do more here. It is the box modier and could take args to allow some of the changes you want.
> 4- Z annot labels should be horizontal by default (probably only in a new theme to not break compat)

Yes I agree but perhaps by modifier?
> 
> ```
> gmt basemap -JX12c/0 -Bafg -Bzag -BWSenZ+b+ggray -p210/30 -JZ10 -R-2/2/-1/3/0/4 -png map
> ```
> 
> ![map](https://user-images.githubusercontent.com/537321/90344564-be012280-e012-11ea-9fb9-74c00c97161b.png)

Esteban82:
I think it is related with3). In [3D view of Parana Basin](https://forum.generic-mapping-tools.org/t/3d-view-of-parana-basin/519) I used the plot3d twice. I add +b the first time I used it (to plot the grayscale surface). So, then I used it again (with the color one) and this surface overlap the outlne of the cube in the front of the figure (near the river mouth).

joa-quim:
>    1- The fill should extend to the vertical walls
> Always, or as options

I think always. That's what Matlab does. If user wants only the base he can do it with two calls.

- 3 Thinking better, the cube should be always open. Why would one want to draw 3 lines ON TOP of the image?

PaulWessel:
I started last night to add new frame options +x[fill], +y[fill] so they are optional, can take same fill as the ground, or be separate.  But struggling with a few issues so on hold during my meetings...

PaulWessel:
Branch wall-paint will let you run

[...]

PR #3985, opened 2020-08-17
Let z-axis annotations be normal to axis as an option

PaulWessel:
The y-axis by default plots annotations normal to the axis, but accepts **+ap** to plot parallel to axis.  The z-axis only plotted parallel to the axis and did not accept **+a**.  This PR allows **+a** but keeps the default **+ap** for backwards compatibility.  Use **+an** to change it, or use  **--MAP_ANNOT_ORTHO**=wez.  **Note:** I am willing to entertain suggestions that we should declare this a bug and let **+an** be the default for both the y- and z-axes.  That decision decides if this is a bug+6.1 backport of a new feature for 6.2.

This PR addresses one of the issues raised in #3983, and the result of

`gmt basemap -JX12c/0 -Bafg -Bzag+an -BWSenZ+b+ggray -p210/30 -JZ10 -R-2/2/-1/3/0/4 -png map`

Looks like this now (note the +an for the z-axis):

map

### DIFF
--- a/doc/rst/source/explain_-B_full.rst_
+++ b/doc/rst/source/explain_-B_full.rst_
@@ -62,9 +62,8 @@
     will plot slanted annotations; *angle* is measured with respect to the horizontal
     and must be in the -90 <= *angle* <= 90 range only.  Also, **+an** is a shorthand
     for normal (i.e., **+a**\ 90) and **+ap** for parallel (i.e., **+a**\ 0) annotations
-    [Default].  For the y-axis, arbitrary angles are not allowed but **+an** and **+ap**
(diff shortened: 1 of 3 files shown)


joa-quim:
Orthogonal on *z* is clearly nicer and is consistent with the *y* behavior. I think we should consider this a bug and make it the default behavior.

PaulWessel:
I agree.  @dongdong?

seisman:
I agree with declaring it as a bug.

PaulWessel:
OK, should be the new default now.  Please test and approve.  Heading on the road for a while.

PR #3993, opened 2020-08-18
Expand 3-D basemap properties with back-walls and gridlines

seisman:
Until now, the **+b** did two things: Turned on 3-D back-walls if gridlines were requested and draw the 3-D box, sometimes overprinting the main illustration.  This PR adds **+B** as an alternative.  Same as **+b** but does not draw the 3-D box outline.  This addresses one of the many points in #3983.

I left this as WIP because I would have preferred **+b** to do what **+B** does now, and let **+B** be the one that adds the corner lines, but that would break backwards compatibility.  
What say you @joa-quim who started this?


joa-quim:
I find it a bit awkward to have a **-B...+B...** syntax. Can't we just make it **+w** for *walls*?

PaulWessel:
Yes, I guess I did not spend too much time thinking about this.  We have +b which does both walls, gridlines (if selected), and the corner box.  I am OK with using +w to turn on walls.  So then +b is +w plus the corner box.
I'm having a zoom with Remko tonight regarding the plane proejction and drawing on those walls.

joa-quim:
But **+b++ is required for *z* grid lines so **+w** should do that grid lines activation as well.

PaulWessel:
Yes, only difference between +b and +w is that +b draws the corner lines you dont like on top of things.

PaulWessel:
Hope you like the **+w**, @joa-quim.

PaulWessel:
Update: Given that I will be working on a PR that will allow painting of the backwall, I am not sure **+w** is needed at all.  I think we want this behavior for 3-D plots:

1. If **-B** specifies the z-axis then the vertical z-axis is plotted and annotated as requested [current behavior]
2. If **-B** specifies gridlines via -**Bz** then we draw the backwall gridlines, no **+b** required for this [currently requires **+b**]
3. If **-B** specifies painting of the backwall (via **+x**[_fill_} and **+y**[_fill_] modifiers) then we do that, no **+b** required.

This means **+b** is simply a modifier that adds the 3-D box and nothing more.  This means it is not the default behavior and thus should meet with @seisman approval.  The rest are improvements, @seisman.

Please give me the go-ahead to finalized these changes via a comment here, and I will finish this PR to just deal with point 2; the coloring will be another PR.

seisman:
> 2\. If **-B** specifies gridlines via -**Bz** then we draw the backwall gridlines, no **+b** required for this [currently requires **+b**]

It makes sense to me. One more thing, as @joa-quim  mentioned in #3983, should we also draw the X and Y gridlines in the backwalls?

PaulWessel:
Yes, I am/will be working on the xy gridlines as well - had a zoom with Remko last night and ready to go.  But trying to break this into smaller PRs so it will come as an enhancement to the gridlines.  I think we only draw gridlines on the back if -Bz gridlines are set.  If you just give -Bafg -Bzaf then you get no gridlines on the backwalls since you did not request z gridlines.

seisman:
How to plot a figure like this one, but without any gridlines on the backwall?
image

PaulWessel:
`gmt basemap -JX12c/0 -Bafg -Bzaf -BWSenZ+b+ggray -p210/30 -JZ10 -R-2/2/-1/3/0/4 -png map`

Since no **-Bz** the **+b** just draws the cube @joa-quim hates. BTW, for modern mode I think **-Bz** with no arguments, like **-B**, should imply **-Bzaf** [currently gives an error].

seisman:
Your command gives me this figure. 
image

But what I'm expecting is:
image

seisman:
> BTW, for modern mode I think **-Bz** with no arguments, like **-B**, should imply **-Bzaf** [currently gives an error].

Yes, I agree.

PaulWessel:
I see. Well, we have no scheme for that.  Let's see the many ways we want to backwalls to be

- No back walls, just the cube outlines [my example using +b]
- Transparent back walls, just the wall outlines [your example - not possible yet]
- Transparent back walls with gridlines [currently only z-gridlines but xy coming]
- Opaque back walls, to be implemented with +x +y.  If no fill given then same as xy fill (-B+gfill)
- Opaque back walls, with gridlines (will need +x, +y and -Bz...g)

Possible ways to include your case:

1. Let that be the default (backwards incompatible but we can always claim it as a bug) so no new modifier needed
2. Re-introduce a **+w** for just the back wall outline.  I think we want to be able to show or not show the wall, so taht will need a modifier, hence +w.  The +w can mean "Draw the back wall outlines".  That way, you can separately paint via +x +y without drawing those lines.

**Unresolved or hidden issues:**

Pen: I think the box pen uses the gridline pen.  This seems certainly appropriate for the 3-D box part, but the outlines of the back wall should possibly be using the **FRAME_PEN**? I am not sure if that will look too bold, but since it is a 3-D plot it would seem reasonable to use that default for this and not have a separate **+w**_pen_ modifier, no?
Back wall with gridlines: There is no fill of this wall, and I think that is OK unless fill is requested, just as for the xy plane.  I.e., there is no default fill, just open.

PaulWessel:
> > BTW, for modern mode I think **-Bz** with no arguments, like **-B**, should imply **-Bzaf** [currently gives an error].
> 
> Yes, I agree.

Hm, the code seems to do this already:

```
	else if (code[0] == 'z' && code[1] == '\0') {	/* If indicating z we do -Bzaf */
		strcpy (args, "zaf");	if (c) strcat (args, c);
	}

```
but 

```
gmt basemap -JX12c/0 -Bafg -Bz -BWSenZ+b+ggray -p210/30 -JZ10 -R-2/2/-1/3/0/4 -png map
basemap [ERROR]: Option -B parsing failure. Correct syntax:
	-B[p|s][x|y|z]<intervals>[+a<angle>|n|p][+l|L<label>][+p<prefix>][+u<unit>] -B[<axes>][+b][+g<fill>][+o<lon>/<lat>][+t<title>] OR
	-B[p|s][x|y|z][a|f|g]<tick>[m][l|p] -B[p|s][x|y|z][+l<label>][+p<prefix>][+u<unit>] -B[<axes>][+b][+g<fill>][+o<lon>/<lat>][+t<title>]
basemap [ERROR]: Offending option -BWSenZ+b+ggray

```

So will debug that case.

PaulWessel:
But switching the order to -Bz -Bafg works, since the other way sets a bool so that we dont go into this section of code again...  So that needs some work.

joa-quim:
The PENs. See [this](https://discourse.julialang.org/t/plotting-in-julia-like-matlab-plots/44997). I think it looks better the way they have it. Just grid line pens even at the base of the walls. At least when the walls are painted.

PaulWessel:
I disagree that it looks better by not clearly drawing the xy plane bounds as we do.  That figure reeks of ugly, especially the crooked flat annotations. Anyway, sounds like this conundrum can only be solved by **+w**[_pen_] then since taste cannot be dictated.

PaulWessel:
Hi @joa-quim and @seisman, I have finished the work on the 3-D stuff.  It is all in this branch:

1. Now does xz and yz gridlines on back walls if gridlines are requested
2. Back-walls are transparent but can now be painted with **+x** and **+y**
3. The outlines of the backwalls can be drawn via **+w** and using the optional pen [grid pen]
4. Documentation is updated to reflect all the changes.

Closes #3983.

joa-quim:
It looks like after the **+w** pen is drawn the black pens come again and plot on top.
```
gmt basemap -JX12c/0 -Bzafg -Bafg -BWSenZ+wthin,green -p250/30 -JZ10 -R-12/2/-51/-37/0/4 -pdf map
```
This is a screencapture of the NE bottom corner.
Also, ``gmrdefaults`` mention twice a line thickness of *default*. *e.g.*

> MAP_GRID_PEN_PRIMARY
    Pen attributes used to draw primary grid lines in dpi units or points (append p) [default,black]. 

but nowhere the *default* size is explained.

image

joa-quim:
and the **-Bzafg** does not plot the transparent vertical walls with the grid lines. One need to add **+w**.
```
gmt basemap -JX12c/0 -Bzafg -Bafg -BWSenZ -p250/30 -JZ10 -R-12/2/-51/-37/0/4 -pdf map
```

joa-quim:
I thought the idea was that **+g**_fill_ alone would paint the vertical walls too, but it doesn't. It should and the **+x,y** options could accept no argument to leave one or both of them transparent.

PaulWessel:
Thanks, I will check those problems.  I did not have much time to test last night.
I think it is odd that one would have to use +x+y to turn _off_ back fill when we never had backfill.  We only every painted the xy plane, and now you can paint the backwalls if you want to, with +x and +y.  Your suggestion would make +x either turn on or off something, unlike +g which only turns on.

joa-quim:
Not that I particularly appreciate painting the walls but it may be useful sometimes when we have scatter points or lines. Then having **+g** setting it all is handy. The **+z** complete the **+y,+y** is a good solution.

One can even imagine that one day the *fill* can accept an image or a projection of the 3D curves/points in that plane.

PaulWessel:
Regarding your black on green line issue. The wall is drawn as a green skewed rectangle.  Then gridlines are drawn.  If you use 0/4.2 for the z-range then you will see that.  So the question is, do we skip drawing the gridline if it coincide with the xy-level and if the top gridline coincides with zmax?  For reference, when we draw gridlines on the xy plane it looks like we are drawing them on top of any frame.  E.g., Mercator -R0/10/0/10 with -Bafg2 will draw gridlines on top of w,s,n,e frames.

PaulWessel:
Actually, for the x-y plan we draw all the gridlines, but then plot the map frame on top.  Hence, if you use -BWS you will see a grid line at n and east if they are a multiple of gridspacing.  Perhaps I can have the gren frame drawn after the z-gridlines as well so it will be the same.

PaulWessel:
Give your green line another try and see if this is what you expect.

PaulWessel:
> and the **-Bzafg** does not plot the transparent vertical walls with the grid lines. One need to add **+w**.
> 
> ```
> gmt basemap -JX12c/0 -Bzafg -Bafg -BWSenZ -p250/30 -JZ10 -R-12/2/-51/-37/0/4 -pdf map
> ```

What is the solution here? Detect that gridlines involving -Bz has been set and turn on what +w turn on implicitly, going with the default pen? I guess that seems reasonable, no?

joa-quim:
Not sure what I expect. But it's strange to get this.
image

since common edges cannot satisfy contradictory requests (black or green?), perhaps solve it by letting the *x,y* setting dominate. I.e, no green at the bottom.

joa-quim:
> What is the solution here? Detect that gridlines involving -Bz has been set and turn on what +w turn on implicitly, going with the default pen? I guess that seems reasonable, no?

Yes, if **-Bzg** and no **+w**implicitly set it.

PaulWessel:
Hm, regarding the green line: The horizontal part you dong like is of course plotted in the yz plane while the black line on the bottom is perspective in the xz plane.  Maybe we simply cannot draw the base line since the two are completely different lines.
Meanwhile, try the auto +w.

PaulWessel:
Seems I can answer my own question.  The outline (+w) for the backwalls should not draw along the base since that would clash with the lines that are plotted in the x-y perspective.  This has largely gone unnoticed due to hardwiring with gridline pen.  So it will be like a door frame: Thick flat pieces of wood on three sides but not on the floor.

PaulWessel:
Hi @joa-quim try green lines again.  Now they draw projected portals.

joa-quim:
Yes, all looks good now.
Just that little thing a mentioned at the beginning. What is *default* size referred in ``gmtdefaults``?

joa-quim:
Yes, all looks good now.
Just that little thing a mentioned at the beginning. What is *default* size referred in ``gmtdefaults``?

PaulWessel:
Default is the default pen.  I believe it is listed in the table with named pens, e.g., thin, thick, default, fat.

PaulWessel:
Now +g sets all, while we have +x,y,z for individual control (and overriding +g).

PaulWessel:
I will change **+x**[_fill_] to **+x**_fill_ (and same for **+y, +z**) since with **+g** I do not see a need to just use **+x** with no fill argument anymore.  OK?

PaulWessel:
Note that transparency on the walls allow you to see the outside ticks on the back side!

joa-quim:
OK for **x**_fill_

But the *default* story (I remember to see it defined somewhere), it's not nice to see this

> Pen attributes used to draw primary grid lines in dpi units or points (append p) [default,black].

and not knowing what *default* is.

joa-quim (approved):

seisman:
Some issues I found:

The z gridlines is too dense:
```
gmt basemap -JX12c/0 -Bafg -Bzg -BWSenZ -p250/30 -JZ10 -R-12/2/-51/-37/0/4 -pdf map
```

**+b** doesn't draw the box?
```
gmt basemap -JX12c/0 -p250/30 -JZ10 -R-12/2/-51/-37/0/4 -pdf map -Bafg -BWSenZ+b -Bzafg
```

PaulWessel:
Will work on those after a few zooms...  I think the auto-Bzg does not account from the shortening due to the perspective.  I am sure that is the same for x/y as well in 3D, but your case suggests it is mostly an z issue.

seisman:
> Will work on those after a few zooms... I think the auto-Bzg does not account from the shortening due to the perspective. I am sure that is the same for x/y as well in 3D, but your case suggests it is mostly an z issue.

It affects all axes:
```
gmt basemap -JX12c/0 -Bg -Bzg -BWSenZ -p250/30 -JZ10 -R-12/2/-51/-37/0/4 -pdf map
```

PaulWessel:
OK, that makes more sense.  So we need to do a shrinking due to vantage point - at least try.

PaulWessel:
Dealt with the **+b** for the wall.  For shrinking the number of gridlines.  There is no exact answer here.  For some -p values you will be seeing a very large x-y plane and then just a very oblique view of a side wall.  Almost any number of gridlines there will be too dense, but perfect for the x-y plane.  It seems not ideal to change the gridspacing between planes sharing the same lines.  Perhaps the best that can be done without major work is to scale the "size" that the auto-algorithm sees by 50%, then let users fine-tune if they want more or less line densities.

PaulWessel:
@seisman, your case had no annotation interval so in that case the algorithm gets a much smaller interval.  if you add a it uses that computed value instead for gridlines.  So the auto computes an interval suitable for annotations (d) which seems good, and then a smaller one (f).  It uses d for annot interval, if requested, f for tick interval, and for gridline interval it uses d if annot was requested, else f.  So we could decide that this is wrong and always use d for gridline interval (it is always computed).  I think the denser ticks is not a big deal but the denser gridlines are.

PaulWessel:
If I instead use d for auto-spacing of gridlines, your example yields this plot:

map

I think that looks better.  If someone just wants gridlines it does not mean they want a super-dense cloud of them.  I also made a first-order adjustment to the "width" used in these calculations by using these scales:

```
	sxy = 1.0 - cosd (GMT->current.proj.z_project.view_elevation);
	sz  = 1.0 - fabs (sind (GMT->current.proj.z_project.view_elevation));
```

The idea is that if you are looking at a very low elevation then the size of the z-y plane shrinks (sxy) while for a very large eleation angle the size of the back-walls shrinks (sz).

PaulWessel:
Hi @seisman, some updates:

1. We have lots of illustrations (think map projections in the cookbook) that display just gridlines.  If I use the earlier idea of selecting the annotation interval for gridlines then it makes a major difference (gridlines every 90 or 180 degrees longitude looks very barren)
2. However, using the first-order adjustment for perspective views seems better.  This means most of the scripts that do **-p** end up failing right now since the computed interval has changed.  This mostly affected things in the x-y plane as that is where most of the reduction in size takes place.  The z-axes seem largely unchanged.
3. It seems fixing your example of just gridlines is very difficult to do without changing things everywhere.  I dont think we will go there.  A figure without annotations is also not very scientific of course.  If the user don't like the result they can use -Bz with an interval specified.

If you have time, please build and run this branch on all the examples and have a look at the **-p** failures to see if you agree that this is progress or not.

seisman:
> Ouch, how to undo? Can you do it?

Done.

PaulWessel:
Thanks, so I may approve this one now?

seisman:
OK to me.

seisman:
Master branch CI jobs fail with following errors:
```
	176 - ex08/ex08.sh (Failed)
	178 - ex10/ex10.sh (Failed)
	487 - grdview/cuerpo.sh (Failed)
	488 - grdview/denver.sh (Failed)
	499 - greenspline/gspline_4.sh (Failed)
	843 - psxyz/3dbars.sh (Failed)
	848 - psxyz/geovector.sh (Failed)
	849 - psxyz/geovector2.sh (Failed)
	859 - psxyz/wedges.sh (Failed)
```
Did you forget to update the original PS files?

PR #3996, opened 2020-08-18
Be more careful cleaning up after failed curl

### LINKED ISSUE #3987: Baddly breaking in externals
Something in #3848 broke badly the externals.
Either this in Matlab
```
G = gmtmex('grdmath -R0/10/0/10 -I0.1 X Y MUL =');
gmtmex('write lixo.grd -Tg', G)
gmtmex('write lixo.grd -Tg', G) 
```
or this in Julia
```
julia> G = GMT.peaks();
julia> gmtwrite("lixo.grd", G)
julia> gmtwrite("lixo.grd", G)
```
work for the first writing but hangs completely on second and one has to kill the processes.

PaulWessel:
If there are curl failures, we need to remove the lock file etc, not just return.  This PR captures all failures and forces via a goto statement to always deal with cleanup.  Closes #3987 (I hope).


joa-quim (approved):
It doesn't completely solve #3987 The remaining issue is that ``gmtremote_get_url()`` should not have been called in first place because it was reading from a MEM file.

PaulWessel:
OK, let's follow up on that in the #3987 but approve this for now.

PR #3998, opened 2020-08-19
Simplify Curl operations

PaulWessel:

We have two separate functions for Curl download.  One is specific to the hash and index files (and has some special checks) while the other is for data download.  There was a lot of repeated code.  We still have two functions, but this PR uses sub-functions to greatly simplify them and avoid the duplications.

### DIFF
--- a/src/gmt_remote.c
+++ b/src/gmt_remote.c
@@ -520,45 +520,46 @@ GMT_LOCAL size_t gmtremote_skip_large_files (struct GMT_CTRL *GMT, char * URL, s
 	return action;
 }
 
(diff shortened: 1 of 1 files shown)


PaulWessel:
Pinging @joa-quim and @seisman on this one.

src/gmt_remote.c
```diff
@@ -73,6 +73,11 @@ struct sigaction cleanup_action, default_action;
 char *file_to_delete_if_ctrl_C;
 #endif
 
+struct LOCFILE_FP {
+	char *file;	/* Pointer to file name */
+	FILE *fp;	/* Open file p[pointer */
```
seisman, line 78:
```suggestion
	FILE *fp;	/* Open file pointer */
```

PR #4001, opened 2020-08-19
Improve the auto -B settings under modern mode

PaulWessel:
We already had **-B** expanding to **-Baf** under modern mode, but several cases were left on the table:

1. **-Bz** after **-B** was not processed and gave an error
2. **-Bp** or **-Bs** were not expanded to **-Bpaf** and **-Bsaf**
3. Same with **-Bpz** and **-Bsz** since the same issue as 1 above.

These cases are now considered.


seisman:
**-B** is expanded to **-Baf**, and **-Baf** means **-Bxaf** and **-Byaf**? Am I understanding it correctly?

**-B -Bz** works but **-Bz -B** still gives an error.

PaulWessel:
Yes you are right.  Try this update.

PR #4025, opened 2020-08-22
Make sure -Q is parsed before -T is parsed if logarithmic

### LINKED ISSUE #4016: makecpt -Q doesn't work for continuous CPT?

The **makecpt -Q** documentation says:
> -Q
> For logarithmic interpolation scheme with input given as logarithms. Expects input z-values provided via -T to be log10(z), assigns colors, and writes out z.

My data has a range from 10 to 10000, so I should use `-T1/4`:

- `gmt makecpt -Cturbo -T1/4/1 -Q` writes a discrete CPT with z from 10 to 10000, as expected.
- `gmt makecpt -Cturbo -T1/4/1 -Q -Z` writes a continuous CPT with z from 10 to 10000, as expected.
- `gmt makecpt -Cturbo -T1/4 -Q` writes a continuous CPT with z from 1 to 4. It seems a bug.

PaulWessel:
If **-Q** is set the **-T** optionmust be parsed for a log10 setup.  However, no **+l** was passed so errors could creep in, like here.  Closes #4016.

### DIFF
--- a/src/gmt_prototypes.h
+++ b/src/gmt_prototypes.h
@@ -480,6 +480,7 @@ EXTERN_MSC void gmt_scale_cpt (struct GMT_CTRL *GMT, struct GMT_PALETTE *P, doub
 EXTERN_MSC void gmt_stretch_cpt (struct GMT_CTRL *GMT, struct GMT_PALETTE *P, double z_low, double z_high);
 EXTERN_MSC struct GMT_PALETTE * gmt_sample_cpt (struct GMT_CTRL *GMT, struct GMT_PALETTE *Pin, double z[], int nz, bool continuous, bool reverse, bool log_mode, bool no_inter);
 EXTERN_MSC void gmt_invert_cpt (struct GMT_CTRL *GMT, struct GMT_PALETTE *P);
+EXTERN_MSC void gmt_undo_log10 (struct GMT_CTRL *GMT, struct GMT_PALETTE *P);
 EXTERN_MSC void gmt_cpt_transparency (struct GMT_CTRL *GMT, struct GMT_PALETTE *P, double transparency, unsigned int mode);
 EXTERN_MSC int gmt_contlabel_info (struct GMT_CTRL *GMT, char flag, char *txt, struct GMT_CONTOUR *G);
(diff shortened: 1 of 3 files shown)


PaulWessel:
I agree with things not being intuitive for the log stuff. In regards to your expectation above, is taht turbo sampled linearly at 1,2,3,4 and then we report 10^z, or is it linearly sampled at 10-10000 directly.  Those two would give different colors - just checking what you did and expect.

seisman:
It seems `gmt makecpt -Cturbo -T1/4/0.5 -Q -Z` do the linear interpolation from 1 to 4, then report 10^z:
```
gmt begin colorbar png
	gmt makecpt -Cturbo -T10/10000
	gmt colorbar -Dx0/0+w10c -Baf

	gmt makecpt -Cturbo -T1/4
	gmt colorbar -Dx0/0+w10c -Baf -X3c

	gmt makecpt -Cturbo -T1/4/0.5 -Q
	gmt colorbar -Dx0/0+w10c -Baf -Q -X3c

	gmt makecpt -Cturbo -T1/4/0.5 -Q -Z
	gmt colorbar -Dx0/0+w10c -Baf -X3c -Q
gmt end show
```

image

PaulWessel:
Yes that makes sense of course, otherwise most colors would come out red...

PaulWessel:
So notice the difference between -Tmin/max and -Tmin/max/inc.  The first flavor simply takes the CPT and stretches the range from 0/1 or -1/1 to min/max.  There is no _interpolation_ of the CPT, just a stretching.  Because the turbo CPT has 256 colors you are getting 256 colors back out, stretched from 1 to 4.  So, the problem here is that **-Q** did nothing.
To sample (i.e., interpolate) a CPT you actually specify _where_ to sample, and that requires an array, and hence min/max/inc.  Now you are sampling the CPT  at 1,2,3,4 and then report 10^z instead. 
There is no way `gmt makecpt -Cturbo -T1/4 -Q` could produce 3 slices since it simply stretches the CPT to that range, and the CPT has 255 slices (for 256 colors).  The only bug I see here is that -Q is not turning those z values to 10^z.  I think this will require a special check.

PaulWessel:
I have implemented this so that `gmt makecpt -Cturbo -T1/4 -Q `gives the only output it can give.  All tests pass, @seisman.

PR #4045, opened 2020-08-24
Update the 3-D plot after box/wall improvements

PaulWessel:
Leaving cuerpo off for now since need some investigation, following #3993


seisman:
No backporting.

PR #4056, opened 2020-08-25
No proj

joa-quim:
And no warning (pass to INFO) when plotting things like this.
```
gmt pscoast -Rd -J+proj=fouc+width=12c -Bg -Gred -png map
```

map


PaulWessel:
Sorry, add backport as bug to .1.1?

PR #4060, opened 2020-08-26
Fixed strsignal(3) usage

uleysky:
Fixed compilation on a systems with strsignal(3) function

### DIFF
--- a/src/gmt_common_sighandler.c
+++ b/src/gmt_common_sighandler.c
@@ -195,7 +195,7 @@ void sig_handler(int sig_num, siginfo_t *info, void *ucontext) {
 	}
 	else {
 #ifdef HAVE_STRSIGNAL
-		fprintf (stderr, "ERROR: Caught signal number %d (%s) at\n", sig_num, strsignal(sig_num);
+		fprintf (stderr, "ERROR: Caught signal number %d (%s) at\n", sig_num, strsignal(sig_num));
 #else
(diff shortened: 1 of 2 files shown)


PaulWessel:
Thanks for checking and fixing!

seisman:
We tests GMT on different Ubuntu and Debian versions, but the latest glibc version on these Linux distros are 2.31. That's why we don't catch the bug in our CI.

BTW, this PR should be backported to 6.1 branch. I will do it manually after merging.

seisman:
Merging this PR gives following warnings when compilation:
```
../src/gmt_common_sighandler.c:198:73: warning: implicit declaration of function ‘strsignal’; did you mean ‘ssignal’? [-Wimplicit-function-declaration]
   fprintf (stderr, "ERROR: Caught signal number %d (%s) at\n", sig_num, strsignal(sig_num));
                                                                         ^~~~~~~~~
                                                                         ssignal
../src/gmt_common_sighandler.c:198:54: warning: format ‘%s’ expects argument of type ‘char *’, but argument 4 has type ‘int’ [-Wformat=]
   fprintf (stderr, "ERROR: Caught signal number %d (%s) at\n", sig_num, strsignal(sig_num));
```
Need `string.h`?

PaulWessel:
Yes, I think so.  It is in signal.h on macos but googling suggest string.h

PR #4072, opened 2020-08-27
INitialize lon-lat nodes for non-GMT projections

### LINKED ISSUE #4055: gmt_geo_to_xy_line crash when GMT->current.map.dlat = Inf
```
gmt begin
gmt subplot begin 3x3  -Blrbt -M0 -F5/0+w1,dashed
plot [ERROR]: Option -R parsing failure. Correct syntax:
        -R<xmin>/<xmax>/<ymin>/<ymax>[/<zmin>/<zmax>]
          Append +r if giving lower left and upper right coordinates
        -Rg or -Rd for global domain
        -R<grdfile> to take the domain from a grid file
plot [ERROR]: Offending option -R0/1.9685/0/0

gmt coast  -Rd -JG30/30/? -Bg -c0,0 -Gred -Da
```

Last command crashes inside ``gmtplot_map_lonline()``, on the 
``GMT->current.plot.n = gmt_geo_to_xy_line (GMT, llon, llat, nn);``
because ``llon, llat`` are NULL. That happens because ``GMT->current.map.dlat = Inf`` in the ``gmtlib_lonpath()`` call and the arrays are not allocated.

Not obvious where to do the right check for this and reurn an error code.

PaulWessel:
If not a GMT projection then we set these to 360,180 to be safe.  Closes #4055

### DIFF
--- a/src/gmt_map.c
+++ b/src/gmt_map.c
@@ -4977,6 +4977,7 @@ GMT_LOCAL int gmtmap_init_polyconic (struct GMT_CTRL *GMT, bool *search) {
 			GMT->current.proj.fwd = &gmt_proj4_fwd;
 			GMT->current.proj.inv = &gmt_proj4_inv;
 			GMT->current.proj.scale[GMT_X] = GMT->current.proj.scale[GMT_Y] = GMT->current.proj.proj4_scl;
+			GMT->current.map.n_lon_nodes = 360;	GMT->current.map.n_lat_nodes = 180;
 			if (GMT->common.R.oblique) {
 				gmt_proj4_fwd (GMT, GMT->common.R.wesn[XLO], GMT->common.R.wesn[YLO], &xmin, &ymin);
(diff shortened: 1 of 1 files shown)


joa-quim (approved):
OK, this solve the current crash but I still feel that's not right that ``gmtlib_latpath ()`` has a test 
```
if ((n_alloc = lrint (ceil (fabs (lon2 - lon1) / GMT->current.map.dlon))) == 0) return (0);
```
and next if it returns 0 we still let it happily go to the bomb in 
```GMT->current.plot.n = gmt_geo_to_xy_line (GMT, llon, llat, nn);```

PaulWessel:
I think gmt_geo_to_xy_line needs to return right away (0) if given 0, since otherwise it tries to use lon[0].

joa-quim:
I think I tried that but it only postpones the crash to a later place.

PaulWessel:
Still the right thing to do though. PR coming.

PR #4083, opened 2020-08-28
Better message when the Registry registerd ghostscript is not found.

joa-quim:


seisman:
Backporting?

joa-quim:
It was found in Registry, but not in the place where it should be according to the Registry value.

PR #4095, opened 2020-08-28
Not let API call exit

PaulWessel:
No longer call exit in the API.


joa-quim:
GMT_gmtlogo(), sig_handler(), grdfft() (though #ifdefed), psconvert (on *nix only) and testapi*() still call exit()

PaulWessel:
Yep, doing this slowly since I had a full branch I had to remove because too many failures after removing all exits...
mgd77.c is full of them.

PaulWessel:
test*.c are not part of the GMT distribution so they can call exit as much as they want.  The others have pending PRs.

Issue #2287, opened 2019-12-16
-JM standard parallel ignored


viktor76525:
Setting projected area using Mercator
-Jm[lon0/[lat0/]]scale or -JM[lon0/[lat0/]]width (Mercator [C])
`psbasemap -JM173/-42/7i -R-200/200/-200/200+uk -Ba1 > test.ps && psconvert -A test.ps`
![test](https://user-images.githubusercontent.com/7464462/70872651-92a2df80-200e-11ea-932c-936677c3f1b2.jpg)

In the attached output the centre is at 0 latitude, I was expecting it to be at lat0 (-42).

**System information**
* Operating system: Linux 64 bit
* GMT version (`gmt --version`): 6.1.0_e60fa2f_2019.12.14

joa-quim:
I see your point but notice that *lat0* is the latitude where the scale in preserved, not the mid-height of the map.

PaulWessel:
Seems like this should be possible though. Maybe when -R is given in local coordinates relative to origin we should make that correction.

PR #4120, opened 2020-08-30
Fix issues in Mercator projection

PaulWessel:

There were two problems in the Mercator projection in 6.1.0 that this PR will fix:

1. The handling of relative distances around a projection origin in setting up a Mercator map (issue #2287) leaked into coordinate projection from **mapproject** and affected the output of projected values, as reported on the [forum](https://forum.generic-mapping-tools.org/t/mapproject-grdproject-issues-and-changes-between-versions/823).
2. The conformal rather than geodetic standard latitude was used in computing the true scale at that latitude.

With the corrections we are again compatible with 6.0.0 output while handling the relative coordinates for projection setup.  For flexibility, we also added a **+m** modifier to **-C** in **mapproject** to allow moving the origin for projected Mercator y-values to the standard latitude [Default remains the Equator].

The recently added merc_origin.ps test changed slightly due to point 2 above.

### DIFF
--- a/doc/rst/source/mapproject.rst
+++ b/doc/rst/source/mapproject.rst
@@ -15,7 +15,7 @@ Synopsis
 **gmt mapproject** [ *tables* ] |-J|\ *parameters*
 |SYN_OPT-R|
 [ |-A|\ **b**\|\ **B**\|\ **f**\|\ **F**\|\ **o**\|\ **O**\ [*lon0*/*lat0*][**+v**] ]
-[ |-C|\ [*dx*/*dy*] ]
+[ |-C|\ [*dx*/*dy*][**+m**] ]
 [ |-D|\ **c**\|\ **i**\|\ **p** ]
 [ |-E|\ [*datum*] ] [ |-F|\ [*unit*] ]
 [ |-G|\ [*lon0*/*lat0*][**+a**][**+i**][**+u**\ *unit*][**+v**] ]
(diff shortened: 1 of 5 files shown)


joa-quim (approved):

PaulWessel:
Hope @seisman can do this to be safe.

seisman:
A small conflict when manually backporting this PR to 6.1:
```diff
++<<<<<<< HEAD
 +      snprintf (buffer, GMT_LEN256, "-R%g/%g/%g/%g -J%s -I -F%c -C -bi2d -bo2d -<%s ->%s --GMT_HISTORY=false",
 +              wesn[XLO], wesn[XHI], wesn[YLO], wesn[YHI], Jstring, unit, in_string, out_string);
++=======
+       snprintf (buffer, GMT_LEN256, "-R%g/%g/%g/%g -J%s -I -F%c -C%s -bi2d -bo2d -<%s ->%s --GMT_HISTORY=readonly",
+               wesn[XLO], wesn[XHI], wesn[YLO], wesn[YHI], Jstring, unit, origin_flag, in_string, out_string);
++>>>>>>> ca699632c1... Fix issues in Mercator projection (#4120)
```

We changed some GMT_HISTORY from `false` to `readonly` in PR #3687, but it wasn't backported to 6.1.

I think we should accept changes in this PR but still let GMT_HISTORY=false, right?

PaulWessel:
Yes, false should be fine for this - that is what it was

PR #4123, opened 2020-08-30
Implementing 1-D variable filter-widths in filter1d

PaulWessel:
Instead of giving a constant width, we can pass a time-series with filter widths.  Seems to work (one test added) but I will play with this a bit more before asking for approval.  Complements the variable 2-D filter width option in **grdfilter**.


PaulWessel:
I've ensured it works with **-E** as well so by default it skips half the filter-width of data at either end.  Since for this variable filter the width changes with each output point so we must reset the checks each time.  I also added a panel for a non-convolution filter (median) so that both convolutions and spatial filters are tested. I attach the revised test example:

varfilt

This completes the filtering revision as now both 1d and 2d filtering can deal with variable filter widths.

PaulWessel:
Ping @joa-quim and @seisman for a review.

joa-quim (approved):

Issue #3949, opened 2020-08-13
Improve the gmtest script


seisman:
GMT's more than 900 test scripts are stored in three separated directories, `test`, `doc/examples`, and `doc/scripts`. Each  directory has its own gmtest script, and the three gmtest scripts are almost the same, except some difference on full path to test scripts. It seems possible to have only one gmtest script in the test directory and still be able to run the tests in other  directories. Thus, we don't have to maintain gmtest scripts at three separated places.

PaulWessel:
yes, good ideal.  Those script are neither short nor clear, so having just one with a few more comments is probably a good plan.

Issue #3206, opened 2020-04-30
Request for earth_relief_01s in area of no SRTM yields incorrect 15s resampled grid


PaulWessel:
Requests for @earth_relief_01s will be filled in with @earth_releif_15s in the oceans since there are no SRTM tiles there.  However, in the case below there are no land area so we end up with earth_relief_15s being resampled at 1s.  Because the chosen region is not a multiple of 15s it gets truncated inward, not outward, and we end up with a strange strip near the north border.  A plain map using earth_relief_15s directly has no such issues.

Script to produce the problem:

```
#!/usr/bin/env bash
RR2="-R-108.444/-108.3724/23.53499/23.6057"
gmt begin two png
  gmt subplot begin 2x1 -Fs10c -B -BwsNE $RR2
    gmt grdimage @earth_relief_15s.grd -V -JM? -c
    gmt grdimage @earth_relief_01s.grd -V -c
  gmt subplot end
gmt end show
```

Result:

![two](https://user-images.githubusercontent.com/26473567/80758694-c2401900-8ad1-11ea-9a68-ad8931564ba5.png)

PaulWessel:
The issue seems to be this: The users desired _w/e/s/n_ are not exact multiples of the grid increment (here 1s).  That should be fine: this sets the **plot** region, not a particular **grid** domain.  However, as we assemble the required grid from srtm tiles and earth_relief_15s via grdblend, the grdblend module creates the final composite grid that is returned to grdimage. Ah, but the _GMT_Create_Data_ call to make that grid uses the **plot** region w/e/s/n and the chosen grid increments (here 1s) to make that grid.  Yikes, this _w/e/s/n_ cannot be used as is for making the grid, since our default procedure for creating a grid from w/e/s/n and dx/dy is to adjust dx/dy if they do not fit evenly into the range.  So we end up with a modified dx/dy that no longer is 1s.  There may be other issues downstream from this but I will start here.
Solution: grdblend needs a way to adjust the w/e/s/n for building the grid, since the given w/e/s/n is a **plot** domain.  It then internally knows to make the w/e/s/n exact integer multiples of dx/dy and create the grid thus.  This grid with a slightly different region is returned by to grdimage, which sets up plotting with the original plot region (which will possibly clip fractions of pixels)

I think the solution is to recognize that when these srtm blend requests are given by a plot module we must make the w/e/sn adjustment.  Someone running grdblend on the command line to make a grid should get the default treatment (with warnings).

PR #4124, opened 2020-08-31
Region for 3s tiles fails if not multiple of 3s

PaulWessel:
See [form post](https://forum.generic-mapping-tools.org/t/error-pygmt-gmtcliberror-module-grdimage-failed-with-status-code-78/829) for details. This PR simply adds a failing test.


PaulWessel:
Note: I thought I had fixed that since it was also reported [earlier](https://forum.generic-mapping-tools.org/t/gradient-problem-using-earth-relief-03s/787) but I cannot find the pull request of issue.  Can you, @seisman?

PaulWessel:
Note it works without -I but fails with -I.

seisman:
This one? https://github.com/GenericMappingTools/gmt/issues/3206

PaulWessel:
Thanks yes.  OK, so this current case does not involve 15s so perhaps takes a different path.  I will debug.

PaulWessel:
Fixed by #4130.

seisman:
But do we want to add the test for it?

PaulWessel:
Ooops, yes, this PR has the test script and now the correct PS.

seisman:
One concern about this PR is, this PR uses earth_relief_03. It means we have to cache the earth_relief_03s grid (at least some tiles).

PaulWessel:
Might we have a flag to prevent a test from being run in the CI, only locally?

PaulWessel:
Likewise, perhaps remove backport?

PaulWessel:
Before adding another single flag, perhaps thing about deeper about test partitioning.  Ways to consider subsets of tests:

```
Ability to skip or select tests that are CPU/time intensive
Ability to skip or select tests that require large data sets
Ability to skip or select tests only for the API (the api dir)
Ability to select tests only for one module (i.e., only the test/module folder)
```

Might be others.  Hopefully the flag (maybe this is a bitflag) can help the CI know what to pre-download? A key issue for us is often the time it takes.  I made a histogram of run times for the tests once.  It is long-tailed, with mostly quick tests and a few slow ones.  A linear scheme does not seem sensible, maybe a flag 1, 2 , 3 that maps to log10 of some scaled time?

seisman:
> Ability to skip or select tests only for the API (the api dir)
> Ability to select tests only for one module (i.e., only the test/module folder)

It's already supported. `ctest -R api` to run api tests only, and `ctest -R psxy` to run psxy-specific tests.

Currently, tests have inconsistent names (e.g., `GMT_hammer.sh` in `doc/scripts` directory, `ex01/ex01.sh` in `doc/examples` directory, and `pshistogram/flip.sh` in `test` directory). In #3949, I plan to let all tests use their relative path as the test name. For example, the test name should be `doc/scripts/GMT_hammer.sh` instead of `GMT_hammer.sh`. After #3949 is done, it's possible to split all tests into three groups: examples, scripts in the documentation, and the real tests.

test/grdimage/rounding.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+# Failing script while fixing https://forum.generic-mapping-tools.org/t/error-pygmt-gmtcliberror-module-grdimage-failed-with-status-code-78/829
+# Now reported as issue ###
```
seisman, line 3: Please fix/improve the comments.

seisman:
For this test, I think we can run `gmt grdcut -R3:57/4:18/44:00/44:15 @earth_relief_03s -G/dev/null` to cache the two tiles (~3 MB) in the CI.

PaulWessel:
Like this?

test/grdimage/rounding.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# Failing script while fixing https://forum.generic-mapping-tools.org/t/error-pygmt-gmtcliberror-module-grdimage-failed-with-status-code-78/829
+# Now reported as issue https://github.com/GenericMappingTools/gmt/pull/4130
+
+# Ensure the two tiles are downloaded first:
+gmt grdcut -R3:57/4:18/44:00/44:15 @earth_relief_03s -G/dev/null
```
seisman, line 6: No. I didn't mean this. It should be done in the CI script:

https://github.com/GenericMappingTools/gmt/blob/66f5bdce05d4e34ecc371633da95d358f517dd57/ci/azure-pipelines-mac.yml#L80-L90

PaulWessel: OK, I agree. Should I or you move that code to the .yml file then?

seisman: I'll do that.

PR #4130, opened 2020-08-31
Make sure intensity subset matches tiled region if given

PaulWessel:
If an intensity grid is given, and a remote earth_relief grid is given, then when the intensity grid is read in and we specify the subregion wesn, make sure this region matches the possibly adjusted tile_wesn due to rounding in gmt_remote.c  See #4124 for background.  The example that showed the problem where these commands:

```
gmt grdgradient @earth_relief_03s -R3.970/4.270/44.000/44.250 -A45 -Nt1 -Gint.nc
gmt grdimage  @earth_relief_03s -Iint.nc -Cgeo -R3.970/4.270/44.000/44.250  -B -png map
```


PaulWessel:
Note: This second commit fixes the original problem reported in #4124.

PR #4159, opened 2020-09-03
Allow auto-legend option -l in histogram

### LINKED ISSUE #4143: Auto legends for histograms

Auto-legend (**-l**) would be very useful when plotting multiple histograms in one plot. For example (example figure from https://www.r-graph-gallery.com/2-two-histograms-with-melt-colors.html):

image

PaulWessel:
It will place a square or rectangle in the legend queue.  Closes #4143 once ready.  RIght now I just allowed -l as an option and will work on size settings.
Maybe @seisman or @joa could contribute a test script for me (just assume -l is implemented for now) so I can test and debug?


PaulWessel:
Some progress using the default sizing:

`gmt histogram t.txt -T0.1 -B -Gpink -W0.5p -l"Grapefruit" -png map`

map

Various issues encountered:

1. Instead of t.txt (1-col file) I used data.txt -i1 to get second column.  Well, the -i setting endured deep into the call to plot that single square and at that time plot would only assign one col of input so both y and size got lost.  Unresolved for now.
2. Also, the -l active setting survives and again when plot i called it refused because we cannot do auto-legend with variable symbol types.  Of course we are not, but taht is what plot things.  That I fixed in process_figures.

The unresolved issue is basically this: When top-level module (here histogram) uses -i, -o, -l and probably others, surely these should not be active when the session calls other modules?

PaulWessel:
Progress on rectangles needed some changes in legend and elsewhere to anticipate width.height strings as size.  Also, this is symbol height set equal to font size. I wonder is a scale might be helpful here, say 90% of font size, @seisman?  See the size of the K vs rect heights.

two

seisman:
What's the definition of *font-size*? It seems that the rectangle height is slightly larger than the character **K**.

PaulWessel:
It depends on the font, among other things.  But typically a fraction.  We do this in GMT:

`gmt_constants.h:#define GMT_LET_HEIGHT	0.728	/* Height of an "N" compared to point size */`

There is also the line-spacing between entries which is 1.1 times fontsize by default   I will test GMT_LET_HEIGHT first and we can see.

PaulWessel:
I think that works well:

two

seisman:
> I think that works well:
> 
> two

Yes, this one looks much betters. Is it 90%?

PaulWessel:
No, GMT_LET_HEIGHT (0.723).

PaulWessel:
I think this one is good to go.  The default settings are working well, and I added a test.

PaulWessel:
histlegend

seisman:
Just tried the following commands:
```
gmt histogram -R2/7/0/10 hist.txt -T0.2 -B -BWSrt -Gred@50 -W0.5p -l"Grapefruit" -pdf map
gmt histogram -R2/7/0/10 hist.txt -T0.2 -B -BWSrt -Gred@50 -W0.5p -l"Grapefruit"+s1.0 -pdf map
```
The two commands both plot squares, but give different sizes.

PaulWessel:
Without looking, I think this is because one is plotted as a square with basically -Ss and that will use the size as the circumference of the circumscribed square, while the other forces a rectangle which as no such circle.  I am not sure what a good solution is here other than ignore.  You are not going to use +s1 I think.

PaulWessel:
Since we only allow square or rect, I can fix this by scaling size by sqrt(2) internally.

seisman:
The first one for **-l** without **+s**:
image

The second one for **-l+s1.0**.
image

seisman:
Square is just a special rectangle. Can we always plot rectangles? If **+s** is given, then width and height of the rectangle are equal.

PaulWessel:
Fixed already.

seisman:
Almost works, but still tiny differences:

without **+s**:
image

with **+s1.0**:
image

PaulWessel:
Because the gap on either side of the symbol is auto-calculated in legend, it is based on the symbol size. So, since I scaled it up via sqrt(2) we get that much wider gaps.  Perhaps your idea of always doing rectangles is better to avoid it.  Checking.

PaulWessel:
Try now.

seisman (approved):

seisman:
Wait. I may find some other issues. Will report later.

test/pshistogram/histlegend.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Test auto-legend for histogram and add in a single plot call as well
+gmt begin histlegend ps
+	gmt convert hist.txt -i0+o-1.5 > shift.txt
+	gmt histogram -R2/7/0/10 hist.txt -T0.2 -B -BWSrt -Gred@50 -W0.5p -l"Grapefruit"+s1.5
+	gmt histogram shift.txt -T0.2 -B -Ggreen@50 -W0.5p -l"Kiwifruit"+s1.5
```
seisman, line 6: The **-l** documentation says:

> Note that +H, +g, +j, +o, +p, +w, and +s will only take effect if appended to the very first -l option for a plot.

But here you use **+s1.5** for both histograms.

Removing the second **+s1.5** gives:
image

PaulWessel:
Maybe we should pick a default like +s1.5 unless the user messes with it.  The user can also mess with +Ssize and affect the plot.  I do like the rectangle better that square which is something we use for symbols, so having the default 1.5 is probably prettier.  What do you think?

seisman:
> Maybe we should pick a default like +s1.5 unless the user messes with it. The user can also mess with +Ssize and affect the plot. I do like the rectangle better that square which is something we use for symbols, so having the default 1.5 is probably prettier. What do you think?

So **+S** controls the size (i.e., the height) of the rectangle, and **+s** controls the width-to-height ratio. I'm OK with a default +s1.5, but I don't think it's related to the issue I just mentioned.

PaulWessel:
No, not really, and the way +s was first introduced was as an overall scale factor.  But here we use it as something else (the ratio).  Once used, I set it to zero so it has no further effect.  I will set the default to do 1.5.

seisman:
> the way +s was first introduced was as an overall scale factor. But here we use it as something else (the ratio). Once used, I set it to zero so it has no further effect.

But should it be the overall ratio for all histograms? Is it technically impossible?

PaulWessel:
No, not really.  The Initial +s scale is written to the legend file, but when the 2nd histogram comes along it just appends to that file and has no idea what +s was set the first time.  That is why the docs says it is ignored.  I think the best for us would be to select the 3:2 rectangle, remove the discussion of using +1 to change the ratio, and always do this shape.  You can change teh size wiht +S but not the aspect ratio.

PaulWessel:
Just to summarize the various options we have on this:

1. We _fix_ the histogram symbol to be a rectangle with ratio 3:2.  This would not not user-changeable, but the size is via **+S**
2. We want the ability to _change_ the ratio. We can do this via **+s** now but there is no memory across modules, so you cannot change it as a default - it is per-process only. It also violates what **+s** is used for and the documentation.
3. As above, but we introduce a GMT defaults for, e.g.,  **GMT_RECTSYMB_RATIO** 1.5 and then you can change that and you are in effect changing it everywhere.
4. We allow **+S**_size_ to accept **+S**_sizex_/_sizey_ and that is how you sneak in a different ratio.  However, you have no way of specifying the ratio, only these dimensions.
5. We introduce **+r**_ratio_ as yet another legend modifier.  That field is already crowded, and seems too special to just have its own variable.

Since the **+s** means scale not ratio, it is bad form of us to use it for this purpose, hence I think 1 is the best option, followed by 4 and 3.

seisman:
Yes, let's do option 1.

PaulWessel:
OK, working on it. One question is +Ssize.  It is always used to set the symbol size, but for a rectangle should it mean width or height? If not given we compute height (from font size) and then scale it by 1.5 to get width.  I think if anyone is doing +Ssize for a histogram and next for a circle, they probably expect the width of the two to be the same, no? So perhaps +Ssize should set width and I divide by 1.5 to get height.  Agree with this scheme?  It avoids talking about setting height.

PaulWessel:
Testing the above of setting the width.  Here are three symbols all given the same 0.5c width.  Looks OK I think (before it would be too large relative to the triangle):

histlegend

seisman:
> So perhaps +Ssize should set width and I divide by 1.5 to get height.

Make sense to me.

One more question, what **+S** does when plotting rectangles (not histograms)? 

Here is a command. It gives a warning but plots a legend item. It seems **+S2.0** means 2c width for the rectangle, right?
```
echo 5 5 2 3 | gmt plot -R0/10/0/10 -JX10c -Baf -Sr -lAPPLE+S2.0 -pdf map
plot [WARNING]: Cannot use auto-legend -l for variable symbol types. Option -l ignored.
```

image

PaulWessel:
I am not sure rectangles in general are supported (I just did that in pshistogram and set both dimensions).  Maybe create another issue for a general rectangle.  I think +Swidth/height is the way to go since that is how things are written and passed to pslegend.

seisman:
OK. Your new scheme looks good to me. You still need to update the test.

PaulWessel:
Why would the test need to change? It is doing the default and passes?

seisman:
> I am not sure rectangles in general are supported (I just did that in pshistogram and set both dimensions). Maybe create another issue for a general rectangle. I think +Swidth/height is the way to go since that is how things are written and passed to pslegend.

If we decide to support **+Swidth/height** for rectangles, we can also set width and height separately for histogram legend items. 

Perhaps we should merge this PR after **+Swidth/height** is implemented in another PR?



> Why would the test need to change? It is doing the default and passes?

Right. I didn't realize that you already removed the **+s**.

PaulWessel:
In plot, rectangles are a bit different from the basic geometric symbols.  Those can get a width via -S, e.g., -Sc1c, but -Sr is always -Sr and we read cols 2 + 3 to get the two dims.  There is -Sr+s to say we instead are getting lon0 lat0 lon1 lat1 for the two diagonals.  However, there is no **-Sr**_width_/_height_ which is kind of peculiar.  I think whe I implemented -Sr eons ago I was doing things where it made sense to read those dims from file, but it seems we should have a way to set them on the commandline as well.  Even -SR for rounded rectangles should be able to read width/height/radius on command line.

If we implemented the above, then we would maybe have a default size for -l, but +S could override.  So maybe an -Sr update should happen first...

PaulWessel:
Implemented +Swidth[/height] and seems to work OK.  Let me know if you find any issues.

seisman (approved):

PR #4170, opened 2020-09-05
grdimage with input image and flipped y-axis (-JX6c/-6c) failed

PaulWessel:
We only set _node_RGBA = kk_ (start index) and then increment it for each row, but if the image needs to be flipped up/down then we need to reset kk at each row since we are going from bottom to top (so node_RGBA will get smaller and smaller per row).

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -1243,7 +1243,7 @@ EXTERN_MSC int GMT_grdimage (void *V_API, int mode, void *args) {
 		else {	/* Dealing with images, three grids, and/or PostScript colormasking */
 			for (row = 0, byte = colormask_offset; row < n_rows; row++) {	/* March along scanlines */
 				kk = gmt_M_ijpgi (header_work, actual_row[row], 0);	/* Start pixel of this row */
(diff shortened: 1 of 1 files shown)


seisman (approved):

seisman:
Two warnings after merging this PR into master:
```
../../src/grdimage.c:1246:24: warning: '&&' within '||' [-Wlogical-op-parentheses]
                                if (Ctrl->D.active && row == 0 || !normal_y) node_RGBA = kk;            /* First time per row equals 'node', afterwards it grows alone */
                                    ~~~~~~~~~~~~~~~^~~~~~~~~~~ ~~
../../src/grdimage.c:1246:24: note: place parentheses around the '&&' expression to silence this warning
                                if (Ctrl->D.active && row == 0 || !normal_y) node_RGBA = kk;            /* First time per row equals 'node', afterwards it grows alone */
                                                   ^
                                    (                         )
1 warning generated.
```

PaulWessel:
Yep fixing that, but we have a similar but when -JX-6c/6c is used (flip left/right) so working on that as well..

PR #4172, opened 2020-09-05
Use one gmtest script to run all different kinds of tests

### LINKED ISSUE #3949: Improve the gmtest script

seisman:

As mentioned in #3949, there are three gmtest scripts in three different directories. This PR makes some changes to test/gmtest.in so that it can be used to run all tests. Thus, we don't have to maintain the gmtest scripts in the `doc/examples` and `doc/scripts` directories.

Some important changes after this PR:

1. We now use the path relative to the GMT root directory as the test names. For example, GMT_autolegend.sh` becomes `doc/scripts/GMT_autolegend.sh`; `ex01/ex01.sh` becomes `doc/examples/ex01/ex01.sh`, and `blockmean/etest.sh` becomes `test/blockmean/etest.sh`.
2. There were three `fail_count.txt` files. Now there is only one, in the `build/test` directory.

With the more consistent test naming, now it's possible to run three groups of tests separately:

- `ctest -R doc/example` to only run the examples
- `ctest -R doc/scripts` to run the scripts used in the documentation
- `ctest -R test/` to run all the "real" tests
- `ctest -R test/api` to run api tests only

Fixes #3949.


PaulWessel (approved):
Nice simplification!

PR #4173, opened 2020-09-05
Fix error when x-dir is flipped for an image

PaulWessel:
For various signs in -JX±width/±height the plotting of an image did not work for negative _width_.  The logic is a bit messier now but works.  In the revised branch (#4163) we will simplify this further.  This completes the fix for both x-dir and -ydir changing directions.


PaulWessel:
I think doing #4163 has been helpful since so far it has revealed (and let us fix( bugs in both **grdmix** and **grdimage**.  Hopefully, someone can approve this one so I can merge it back into that branch.

joa-quim (approved):

PR #4176, opened 2020-09-05
Better symbol formatting of text

PaulWessel:
Clean up the text of the explain_symbol rst files. Mostly grammar, _italicizing_ variables, and internal links.

### DIFF
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -82,10 +82,10 @@
 
     **Note**: The uppercase symbols **A**, **C**, **D**, **G**, **H**, **I**, **N**,
     **S**, **T** are normalized to have the same *area* as a circle with
-    diameter *size*, while the size of the corresponding lowercase symbols
+    diameter *size*, while the *size* of the corresponding lowercase symbols
     refers to the diameter of a circumscribed circle.
 
-    The next collection of the five symbols require two or more size parameters, and
+    The next collection shows five symbols that require two or more size parameters, and
     some have optional modifiers to extend the symbol appearance.
 
     .. figure:: /_images/GMT_base_symbols2.*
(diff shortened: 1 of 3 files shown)


doc/rst/source/explain_symbols.rst_
```diff
@@ -95,61 +95,61 @@
...
+        For degenerate ellipses (circles) requiring just the *diameter*, use **-SE-**.
+        The *diameter* is excepted to be given in column 3.  Alternatively, append
+        the desired *diameter* to **-SE-** and this fixed *diameter* is used instead.
         For allowable geographical units, see `Units`_ [Default is **k** for km]. The
         shape of the ellipse will be affected by the properties of the map projection.
 
     **-Sj**
-        Rotated rectangle. Direction (in degrees counter-clockwise from
-        horizontal), x-dimension, and y-dimension must be found in columns 3, 4, and 5.
+        Rotated rectangle. The *directiond (in degrees counter-clockwise from
```
seisman, line 119:
```suggestion
        Rotated rectangle. The *direction* (in degrees counter-clockwise from
```

seisman (approved):

seisman:
Perhaps the documentation is more clear if we give the full syntax for each symbol, for example **-Sa***size*, **-Sr***width*[/*height*].

PaulWessel:
Yes, but should it be **-S**[**a**][_size_] since symbol code is optional (read from file if missing) and size is optional (read from file if missing)?

seisman:
> should it be **-S**[**a**][_size_] since symbol code is optional

Yes, both symbols and sizes are optional, but the above syntax may make more confusion.

For example, when I read the description of **-Sr**. What I expect to know is how to give *width* and *height*. 

Changing it to **-Sr**_width_[/*height*][**+s**] makes the syntax consistent with the description. 

image

The optional *symbol* and *size* are already well explained at the beginning.


> The general input expectations are:
> 
> x y [ z ] [ size ] [ symbol-parameters ] [ symbol ]
> 
> where the optional z is required when -C is used, the optional size is required when no symbol size is specified, and the trailing text with the symbol code is required when symbol code is not specified on the command line. 
> 
>

PaulWessel:
OK.  There are still some symbols like **j,** **w**, maybe others, that could accept arguments on command line, just like now **r** and **R** do.  It seems a bit arbitrary that some take width/height/radius while others cannot take start/stop/radius, etc. RIght now you could have your fuller -Sr explanation but -Sj would have no args, for instance.  Maybe I should enhance that first.

seisman:
Yes to me.

PaulWessel:
Working on e, j, E, J.  So our degenerate ellipses and rectangles uses -SE- and -SJ-, but as I implement this like -Sr etc, we do not need that - sign to tell we only have one arg.  So I will leave this for backwards compatibility but we will ahve to remove those -SE- -SJ- from docs and scripts and examples.

PaulWessel:
Oops, too fast.  For no arguments we still need -SE- so we know to read 1 or 3 columns.

PaulWessel:
So for **-Sw**|**W** there is the issue that the current syntax always takes the diameter on the command line, while the two angles are from the file.  This is a bit strange and different from **-SE -SJ -Se -S**j.  Furthermore, you can append /_inner_ to the diameter to set the inner diameter [0].  Since that is an option, it cannot be read from file I think.  In hindsight, we should have done this:

**-SW**[_diameter_[/_angle1_/_angle2_]][**+i**[_inner_]][**+a**[_darc_]][**+r**[_dr_]][**+p**_pen_]

so that it is possible to give a **+i** with no args meaning read it after the other items from file. **-SW+i** would expect _outerdiameter angle1 angle2 innerdiameter_ from file while -SW would expect just _diameter angle1 angle2_.  So I think we can do this:

1. If **+i** is found then it is new syntax and that diameter is either given or read from file as the last of the 0-3 args.
2. If no **+i** then inner is never read, but may be found via /_inner_.
3. If /_inner_ is set it is old syntax and both _angle1_ and _angle2_ must be given via the file.
4. If no /_inner_ is set then (regardless of any **+i**) we can handle from 0-4 args from file.
5. If diameter/angle1/angle2[+mods] is given then we don't read anything unless +i is empty.

We would simply deprecate the /_inner_ from the docs.  Does this seem clear and backwards compatible to you, @seisman ?

seisman:
Sounds reasonable to me.

PR #4185, opened 2020-09-09
Improve symbol documentation again

PaulWessel:
Now we show the syntax for each symbol, and clean up the _symbols_only.rst__ file which is only used by **ternary**.

This we could still do better:

1. There is tremendous overlap between the three explain_symbol*.rst_ files.  E.g., the lines explaining the syntax to draw a circle is repeated verbatim across three files.
2. The _explain_symbols2.rst__ (for plot3d) and _explain_symbols.rst__ (for plot) are almost identical except the plot3d version has two 3-D symbols (cube and column) and uses different column numbers for start of optional arguments (off by one due to z).
3. The _explain_symbols_only.rst__ is only used by ternary and has lots of overlap (for the siple symbols) with that for plot.

What are the best options here? Given _explain_symbols.rst__ was completely revamped with illustrations, it should be our source.  Are any of these ways possible, @dongdong and @joa-quim ?

- Keep the simplest symbols in a new explain_simple_symbols.rst_ and it can be included by both explain_symbol*.rst_ files?
- Reword to avoid specific references to columns, maybe "If -C is used then the symbol data starts one columns after the coordinates".
- Other ideas?


PaulWessel:
If we could avoid specific column numbers then the **plot3d** version could include the **plot** version plus add a final section on 3-D symbols (currently cube and column only, but boy would we like sphere and ellipsoid...).  I think avoiding specific column numbers and instead use relative (i.e., not 3,4,5 or 4,5,6 but the three next columns).

PaulWessel:
I followed the above plan so now we only have a shared _explain_symbols.rst_ for **plot** and **plot3d** (the latter adds just the two 3-D symbols at that end), and we have a shorter one _explain_symbols_ternary.rst__ just for those that can be plotted by **ternary**.  Fixed some typos as well.  I think this is now pretty clear and can be approved. Eliminated _explain_symbols2.rst__  and all that duplication.

doc/rst/source/explain_symbols.rst_
```diff
@@ -1,29 +1,28 @@
...
-    used for the symbol size or append the units to the sizes in the file.
-    If symbol sizes are expected via the third data column then you may convert
+    input file you *must* use :term:`PROJ_LENGTH_UNIT` to indicate the unit
+    used for the symbol size or append the units to the sizes listed in the file.
+    If symbol sizes are expected via the a data column then you may convert
     those values to suitable symbol sizes via the **-i** mechanism. The general
     input expectations are:
 
-    *x y* [ *z* ] [ *size* ] [ *symbol-parameters* ] [ *symbol* ]
+    *coordinates* [ *value* ] [ *size* ] [ *symbol-parameters* ] [ *symbol* ]
```
seisman, line 16: Perhaps explain that *coordinates* is *x y* for plot, and *x y z* for plot3d.

doc/rst/source/explain_symbols.rst_
```diff
@@ -1,29 +1,28 @@
...
 
-    where the optional *z* is required when **-C** is used, the optional
+    where the optional *value* is required when **-C** is used, the optional
     *size* is required when no symbol size is specified, and the trailing
```
seisman, line 19: It's still unclear how to specify *size* in the input file for symbols with more than one size parameters. For example, for a rectangle, users may think that the *size* should be `width/height` or `width height`.

doc/rst/source/explain_symbols.rst_
```diff
@@ -97,52 +96,52 @@
         Upper-case versions **E**, **J**, and **W** are similar to **e**, **j** and **w**
         but expect geographic azimuths and distances.
 
-    **-Se**
+    **-Se**\ *direction*/*major_axis*/*minor_axis*
         **e**\ llipse. If not given, then *direction* (in degrees counter-clockwise from horizontal),
-        *major_axis*, and *minor_axis* must be found in columns 3, 4, and 5. This option yields
-        a Cartesian ellipse whose shape is unaffected by the map projection.  If only a *size* is
-        given then we plot a degenerate ellipse (circle) with given diameter.
+        *major_axis*, and *minor_axis* must be found after the location [and *value*] columns. This option yields
```
seisman, line 101:
```suggestion
        *major_axis*, and *minor_axis* must be found after the *coordinates* [and *value*] columns. This option yields
```

doc/rst/source/explain_symbols.rst_
```diff
@@ -97,52 +96,52 @@
...
+        a Cartesian ellipse whose shape is unaffected by the map projection.  If only a single
+        *size* is given then we plot a degenerate ellipse (circle) with given diameter.
```
seisman, line 103: > If only a single *size* is given then we plot a degenerate ellipse (circle) with given diameter.

I don't test it, but does it mean I can give only one *size* even in an input file?

doc/rst/source/explain_3D_symbols.rst_
```diff
@@ -0,0 +1,31 @@
+    All the previous symbols are 2-D and are plotted in the *x*-*y* plane only.
```
seisman, line 1: This file is not included in plot3d.

doc/rst/source/explain_symbols.rst_
PaulWessel: We should check.   I think **-Se-** or **-SE-** means read a single item, while **-Se**_diam_ or **-SE**_diam_ do no longer need that extra minus sign.

doc/rst/source/explain_symbols.rst_
PaulWessel: OK, I did something like that without mentioning modules.

doc/rst/source/explain_symbols.rst_
PaulWessel: Will look.

doc/rst/source/explain_3D_symbols.rst_
PaulWessel: In plot3d_common.rst_
```
.. _-S:

.. include:: explain_symbols.rst_

.. include:: explain_3D_symbols.rst_

```

doc/rst/source/explain_symbols.rst_
PaulWessel: That is what [ *size* ] [ *symbol-parameters* ] means, no? If there is more than one size argument then that means you have symbol-parameters.

seisman: It's a little confusing for me, especially for some symbols like -Se (-Sedirection/major_axis/minor_axis, what does *size* means?)

PaulWessel: Maybe part of the problem is the listing of two things: [ *size* ] [ *symbol-parameters* ].  What if it was just *size* and we say the one or more columns needed to supply the size parameter(s)?

seisman: For **-Sb**, *base* can also be given in the input file, is *base* part of *size* or *symbol-parameters*?
```
-Sb[size[c|i|p|u]][+b|B[base]]
```

doc/rst/source/explain_3D_symbols.rst_
seisman: OK, I must have missed it in my previous build.

doc/rst/source/explain_symbols.rst_
PaulWessel: These all work for degenerate geo ellipse

echo 0 0 300 | gmt plot -R-3/3/-3/3 -Jm1i -SE- -Gred -B -pdf t
echo 0 0 | gmt plot -R-3/3/-3/3 -Jm1i -SE-300 -Gred -B -pdf t
echo 0 0 | gmt plot -R-3/3/-3/3 -Jm1i -SE300 -Gred -B -pdf t

doc/rst/source/explain_symbols.rst_
PaulWessel: Hm, no, I would say that is part of _symbol-parameters_.  Maybe we need to say that *size* can be more than one column depending on the symbol, (e.g., _width_ and _height_ for a rectangle requires two columns) while some symbol has additional parameters (such as _azimuth_ for an oriented symbol).  Would that be clearer?  Actually, I think that fails too because the ellipse starts with teh angle, then the two dimensions, so listing size first fails.  I think the bets would be a single entry called symbol-parameters which may be one or more parameters needed to define the symbol, such as diameter for a circle, width and height for a rectangle, and direction, major and minor axies for an ellipse.

doc/rst/source/explain_symbols.rst_
seisman: I think you misunderstood what I say. The new documentation for **-Se** says:

>  **-Se**\ *direction*/*major_axis*/*minor_axis*
>        **e**\ llipse. If not given, then *direction* (in degrees counter-clockwise from horizontal),
>         *major_axis*, and *minor_axis* must be found after the *coordinates* [and *value*] columns. This option yields
>        a Cartesian ellipse whose shape is unaffected by the map projection.  If only a single
>        *size* is given then we plot a degenerate ellipse (circle) with given diameter.

I think the documentation means the following command works:
```
echo 0 0 | gmt plot -R-3/3/-3/3 -Jm1i -Se1c -Gred -B -pdf t
```
but users may think that the below command works (it doesn't):
```
echo 0 0 1c | gmt plot -R-3/3/-3/3 -Jm1i -Se -Gred -B -pdf t
```

doc/rst/source/explain_symbols.rst_
seisman: > I think the bets would be a single entry called symbol-parameters which may be one or more parameters needed to define the symbol, such as diameter for a circle, width and height for a rectangle, and direction, major and minor axies for an ellipse.

It sounds good to me, but I didn't check if it works for other symbols (some symbols I never used before).

PaulWessel:
I think this looks pretty good now, and with only minor duplication between **ternary** and plot*.

PaulWessel:
Let me know if you still feel there is lack of clarity, @seisman .

doc/rst/source/explain_symbols.rst_
```diff
@@ -97,52 +100,52 @@
...
         a linear projection we assume the dimensions are given in the same units as **-R**.
         For allowable geographical units, see `Units`_ [Default is k for km].
         The shape of the rectangle will be affected by the properties of the map projection.
 
-    **-Sr**
+    **-Sr**\ *width*/*height*
         **r**\ ectangle. If *width*/*height* are not given, then these
-        dimensions must be found in data columns 3 and 4.  Alternatively, append **+s**
+        dimensions must be found after the location [and *value*] columns.  Alternatively, append **+s**
         and then the diagonal corner coordinates are expected in columns 3 and 4 instead.
```
seisman, line 139: **-Sr** still mentions columns 3 and 4.

doc/rst/source/explain_symbols.rst_
```diff
@@ -175,7 +178,7 @@
         specific fonts (size and type) and control outline and the fill properties. Note there
         is no mechanism to perfectly center the string; see **-D** to make a simple global adjustment.
 
-    **-Sl**
+    **-Sl**\ *size*\ **+t**\ *string*\ [**+f**\ *font*]
```
seisman, line 181: **+j***justify* is valid, right?

doc/rst/source/explain_3D_symbols.rst_
```diff
@@ -0,0 +1,31 @@
...
+        is measured relative to *base* [Relative to origin].
+        The facet colors will be modified to simulate shading.  Use **-SO** to disable 3-D illumination.
+        Normally a single *z* value is considered.  For multi-band columns, append
+        **+z**\ *nbands* (and provide *nbands* *z*-columns via the input; these must be monotonically
+        increasing up the column) or append **+Z**\ *nbands* (and expect *nbands* of *dz* increments
+        in data columns that must be summed to yield actual *z* values).  The multi-band column symbol
+        requires **-C** and will use the band number (0, 1, *nbands*\ -1) to assign the band color.
+
+    **-Su**\ *size*\ [**c**\|\ **i**\|\ **p**\|\ **u**]
+        c\ **u**\ be (3-D).  The *size)* sets length of all sides. Append
```
seisman, line 28:
```suggestion
        c\ **u**\ be (3-D).  The *size* sets length of all sides. Append
```

doc/rst/source/explain_symbols_ternary.rst_
```diff
@@ -14,138 +14,112 @@
     those values to suitable symbol sizes via the **-i** mechanism. The general
     input expectations are::
```
seisman, line 15: Double colons mean that the following lines are codes, so the documentation is rendered like 
image

We should use a single colon here:
```suggestion
    input expectations are:
```

doc/rst/source/explain_symbols_ternary.rst_
```diff
@@ -1,5 +1,5 @@
 **-S**\ [*symbol*][*size*]
-    Plot individual symbols.
+    Plot individual symbols in a ternary diagram.
     If present, *size* is symbol size in the unit set in
     :doc:`gmt.conf` (unless **c**, **i**, or **p** is appended). If the symbol
```
seisman, line 4:
```suggestion
    If present, *size* is symbol size in the unit set by
    :term:`PROJ_LENGTH_UNIT` (unless **c**, **i**, or **p** is appended). If the symbol
```

seisman (approved):
It looks good. Just a few tiny issues.

seisman:
BTW, the "Segment Header Parsing" section only mentions -G, -W and -Z, but as far as I know, -L is also valid.

doc/rst/source/explain_symbols.rst_
PaulWessel: OK fixed

doc/rst/source/explain_symbols.rst_
PaulWessel: Yep, now added

PR #4199, opened 2020-09-11
Fix GMT_base_symbols3D.ps

seisman:
GMT_base_symbols3D.sh fails. It seems the PS file was generated using SI units.


seisman:
The same fix in #4198.

PR #4223, opened 2020-09-16
Adapt to new format of Ghostscript registry key.

joa-quim:
Ghostscript now writes a registry ey with the patch number, E.g. 9.53.1 while before there only up to minor (9.53). This broke the ability to find that key in the Windows registry. This PR fixes it but at the expenses of not searching anymore to possible multiple GS versions installed. Now it takes the first one it picks.


seisman:
The CI now passes.

> Now it takes the first one it picks.

Is it guaranteed that the newest version is the first one?

joa-quim:
> Is it guaranteed that the newest version is the first one?

No, the first one will whatever the ``RegOpenKeyEx()`` comes out with.

seisman (approved):

PR #4228, opened 2020-09-18
Improve docmentation for -t option

seisman:

1. transparency=0 is also valid.

2. Improve the -t description

**Old:**
image

**New:**
image

### DIFF
--- a/doc/rst/source/plot3d_common.rst_
+++ b/doc/rst/source/plot3d_common.rst_
@@ -196,7 +196,9 @@ Optional Arguments
 .. include:: explain_-qi.rst_
 
 .. include:: explain_-t.rst_
-If no transparency is appended then we will read it from the last column per data record instead.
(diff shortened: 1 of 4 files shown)


doc/rst/source/plot3d_common.rst_
```diff
@@ -196,7 +196,9 @@ Optional Arguments
 .. include:: explain_-qi.rst_
 
 .. include:: explain_-t.rst_
-If no transparency is appended then we will read it from the last column per data record instead.
+..
+
+   If no transparency is appended then we will read it from the last column per data record instead.
```
seisman, line 201: Perhaps better to say "the last numerical column"?
```suggestion
   If no transparency is appended then we will read it from the last numerical column per data record instead.
```

PaulWessel (approved):

PR #4238, opened 2020-09-19
More flexible scale origin selection in -L

### LINKED ISSUE #4227: Make +cslat in basemap +L optional.
In order to draw a simple map scale (basemap +L), I think it would be good to make optional to define the latitude where the scale is calculated (slat; as in done for the longitude). The default could be the central parallel of the map. 

**Are you willing to help implement and maintain this feature?** 
I would like to help but I am not sure if I could.

PaulWessel:

Until now, the **+c**[_lon_/]_lat_ modifier in **-L** used to specify the origin for the map scale computation has been required.  This PR relaxes that by introducing two new flavors:

Here, **+c** with no arguments sets the scale origin to the mid-point of the map frame
No **+c** given at all sets the scale origin to match the placement location specified by the refpoint

I have updated the docs and added a new test to make sure this works.  Should be good to go I tope.
Closes #4227


joa-quim (approved):
Please add a summary of this to a log file. I'm becoming very lost of new features added since I made the Julia wrappers and on how I should update them depending on GMT versions

PaulWessel:
Were is the ChangeLog? I think it is hand-made at time of a release?

PaulWessel:
OK, ChangeLog is an rst file in doc/source.  We have made lots of changes since 6.1.1 but they are all listed in the git log.  Usually, we run through the git log since last release and select items that should go in the ChangeLog.  it is not done piece-by-piece.  I understand your issue though. Need to find the script/command that does the extraction from git.  Hi @seisman, perhaps easy for you got write a short admin/git_release_history.sh that accepts a tag (say 6.1.1) as the starting point and reports changes since then on stdout?  I think we still need to sort through the output and I think I did some awk/grep to make it simpler to read but cannot find it...

PR #4243, opened 2020-09-21
Update earth-age.rst_

PaulWessel:
Paper in press, has doi.


seisman (approved):
We may also need to update the earth_age grids. 

`grdinfo` says:

```
Remark: Obtained by Gaussian Cartesian filtering (111.2 km fullwidth) from age.2020.1.GTS2012.1m.nc [Seton et al., 2020; in review]
```

PR #4244, opened 2020-09-22
Improve the Developer documentation for how to debug

PaulWessel:

In addition to debug the GMT C from command line calls, I have added sections (for macOS) on how to debug PyGMT and GMT.jl scripts.  I am holding back on GMT/MEX as we will explore adding it as a supplement due to the compiled nature of the interface (not a wrapper).
We learned that Xcode 12.0 does not work with the julia process but 11.7 does.

### DIFF
--- a/doc/rst/source/debug.rst
+++ b/doc/rst/source/debug.rst
@@ -125,44 +125,63 @@ Applications).  Xcode may change as versions change; the images below is for Xco
 Debug PyGMT in Xcode on macOS
 ------------------------------
 
-Install PyGMT following the official instructions:
+Install PyGMT following the official instructions (if you already have the dev version then you may just
+need to cd into your pygmt dir and call git pull):
 
-#. Add conda-forge channel
-   conda config --prepend channels conda-forge
-   **Note**: The next step is different from the PyGMT official instructions, because we want to use the GMT dev version
-   conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging
+#. Add conda-forge channel::
 
-#. Activate the pygmt environment
-   conda activate pygmt
+    conda config --prepend channels conda-forge
+    **Note**: The next step is different from the PyGMT official instructions, because we want to use the GMT dev version
+    conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging
 
-#. Install Pygmt in editable/development mode
-   cd pygmt
-   make install
+#. Activate the PyGMT environment::
 
-#. Compile GMT using Xcode
-   Tell PyGMT where to find the GMT library by setting the environmental variable GMT_LIBRARY_PATH
-   export GMT_LIBRARY_PATH=~/Gits/gmt/gmt/build/xcode/src/Debug
+    conda activate pygmt
 
-#. Open Xcode
-   Run a python console, attach the process id in Xcode, and run PyGMT codes in the Python console.
-   Set a stop point in Xcode, say in GMT_Call_Module or GMT_Create_Session and Xcode will stop at the breakpoint.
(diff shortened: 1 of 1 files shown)


doc/rst/source/debug.rst
```diff
@@ -119,3 +121,60 @@ Applications).  Xcode may change as versions change; the images below is for Xco
...
+    conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging
+
+#. Activate the PyGMT environment::
+
+    conda activate pygmt
+
+#. Install Pygmt in editable/development mode::
+
+    cd pygmt
+    make install
```
seisman, line 144: @weiji14 Do you think if it's easier to simply run the following command to install pygmt and its dependencies, instead of creating a virtual environment using conda: 
```
cd pygmt
pip install -e .
```

weiji14: Yes it would be easier, but this would contaminate the system python. I.e. ok for the occasional Python user like Paul, but not for those of us who might work with different virtual environments. But we can put `conda activate pygmt` as **optional** with this caveat in mind.

I would also go with `pip install --editable .` instead of `pip install -e .` just for readability, again for casual Python users :wink:

PaulWessel: Could you of you make the changes to the debug.rst file to cover these cases?  Then we can call it a wrap.

doc/rst/source/debug.rst
```diff
@@ -119,3 +121,60 @@ Applications).  Xcode may change as versions change; the images below is for Xco
    .. figure:: /_images/xcode-8.*
       :width: 100%
       :align: center
+
+Debug PyGMT in Xcode on macOS
+------------------------------
+
+**Note**: Tested with Xcode 12.0.  Install PyGMT following the official instructions (if you already have the dev version then you may just
+need to cd into your pygmt dir and call git pull):
```
weiji14, line 129:
```suggestion
**Note**: Tested with Xcode 12.0.  Install PyGMT following the official instructions at https://www.pygmt.org/dev/install.
If you already have the dev version then you may just need to `cd` into your pygmt dir and call git pull:
```

doc/rst/source/debug.rst
weiji14, line 144:
```suggestion
#. Add conda-forge channel and activate virtual environment (optional)::

    conda config --prepend channels conda-forge
    **Note**: The next step is different from the PyGMT official instructions, because we want to use the GMT dev version
    conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging

    # Activate the PyGMT environment
    conda activate pygmt

#. Install Pygmt in editable/development mode::

    cd pygmt
    pip install --editable .
```

Something like this?

doc/rst/source/debug.rst
```diff
@@ -119,3 +121,59 @@ Applications).  Xcode may change as versions change; the images below is for Xco
...
+#. Add conda-forge channel and activate virtual environment (optional)::
+
+    conda config --prepend channels conda-forge
+    **Note**: The next step is different from the PyGMT official instructions, because we want to use the GMT dev version
+    conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging
+
+    # Activate the PyGMT environment
+    conda activate pygmt
+
+#. Install Pygmt in editable/development mode::
```
seisman, line 140:
```suggestion
#. Install PyGMT in editable/development mode::
```

seisman (commented):
The PyGMT instructions now look good.

PaulWessel:
Will need a review, though, in order to merge.

seisman (approved):

PR #4249, opened 2020-09-23
Anticipate unit modifier even if -R gets a gridfile

### LINKED ISSUE #4248: The -R option fails to parse trailing unit selection +u if argument is a grid file
In ex28 we have a command like this:

`gmt grdimage @Kilauea.utm.nc -I+d -Jx1:160000 --FONT_ANNOT_PRIMARY=9p`

which works because we rely on some under-the-hood magic that guesses the grid is in UTM meters.  Ideally, it should be

`gmt grdimage @Kilauea.utm.nc+ue -I+d -Jx1:160000 --FONT_ANNOT_PRIMARY=9p`

but this causes a failure since the parser is not removing the +ue modifier (or expecting it), just passing it as part of the filename to the remote reader, which sternly says:

```
gmt [NOTICE]:   -> Download cache file: @Kilauea.utm.nc+ue
gmt [ERROR]: Libcurl Error: HTTP response code said error
```

The solution is to handle the modifier in the section of the code that deals with **-R**_gridfile_.

PaulWessel:
Addresses and closes #4248.

### DIFF
--- a/doc/rst/source/explain_-R_full.rst_
+++ b/doc/rst/source/explain_-R_full.rst_
@@ -29,9 +29,7 @@
     **-I** is used to create the corresponding region.
     Alternatively, specify the name of an existing grid file and the
     **-R** settings (and grid spacing and registration, if applicable) are copied from
-    the grid. When **-R** is used without any further arguments, the
-    values from the last use of **-R** in a previous GMT command
-    will be used.
+    the grid.
     For calendar time coordinates you may either give (a) relative time
     (relative to the selected :term:`TIME_EPOCH` and in the selected
     :term:`TIME_UNIT`; append **t** to **-JX**\|\ **x**), or (b) absolute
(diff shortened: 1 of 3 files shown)


seisman:
I saw the following errors/warnings:
```
$ gmt grdimage @Kilauea.utm.nc+ue -I+d -Jx1:160000 --FONT_ANNOT_PRIMARY=9p -pdf map
gmt [NOTICE]:   -> Download cache file: @Kilauea.utm.nc+ue
gmt [ERROR]: Libcurl Error: HTTP response code said error
gmt [WARNING]: You can turn remote file download off by setting GMT_AUTO_DOWNLOAD off.
grdimage [NOTICE]:   -> Download cache file: @Kilauea.utm.nc
```

PaulWessel:
Thanks, tracing that down now.

PaulWessel:
Wait a minute, you cannot add **+ue** to some random grid.  That is a **-R** modifier.  However, you can add **+Uk** or something to scale me to km (or the other way, cannot recall now) but that also fails for the same reason the invalid **+u**e does.

PaulWessel:
OK, this should hopefully do the trick.

seisman (approved):

Issue #4235, opened 2020-09-18
Problem controlling frame annotation in subplot


thatfaultguy:
**Description of the problem**

I am using the inside MAP_FRAME_TYPE and would like change the frame annotation between subplots with the first subplot having annotated ticks and the following two subplots with just the tick marks. If I specify the -B option at the subplot begin line, then this seems to override the individual -B commands in the subplots.  

Furthermore, the ticks and annotations are behind the grd file (and only visible because it is transparent - I think), how can I bring them to the front? I don't want to use annotations outside the plot as it would create more white space between the subplot frames.

**Full script that generated the error**

link to data files used in the script:
https://drive.google.com/drive/folders/1SnAVMq4OleuOQMrr4jAg3BY4LcLZZbOF?usp=sharing

```
# PASTE CODE HERE
#!/bin/bash

cpt="roma0.cpt"
obs_file1="D118input.grd"
model_file1="D118_lmodel.grd"
residual_file1="D118_lresidual.grd"
fault="fault_1f_uniform_long.txt"


gmt begin insar_plot png
gmt set MAP_FRAME_TYPE inside 
    gmt subplot begin 3x3 -Fs3i -A 
        gmt subplot set 
            gmt grdimage $obs_file1 -R170/172/61.8/62.5 -C$cpt -t40 
            gmt plot $fault -W1p,black -BWeSn+t"Observed"
        gmt subplot set 
            gmt grdimage $model_file1 -R170/172/61.8/62.5 -C$cpt -t40 
            gmt plot $fault -Bwesn+t"Modelled" -W1p,black 
        gmt subplot set 
            gmt grdimage $residual_file1 -R170/172/61.8/62.5 -C$cpt -t40
            gmt plot $fault -Bwesn+t"Residual" -W1p,black
    gmt subplot end 
gmt end show 
```

**Full error message**

```
# PASTE ERROR MESSAGE HERE
```

**Actual outcome**

![insar_plot](https://user-images.githubusercontent.com/67931182/93647870-454d0780-f9be-11ea-8ebe-43d83a3fd3eb.png)

**Expected outcome**

I was hoping that only the first subplot would be annotated with the other subplots having only tickmarks. also the annotations should be completely opaque and at the 'top' layer

**System information**

* Operating system:macOS
* GMT version (6.1.1):

joa-quim:
Hi, this is a question for the forum (not an issue that I can spot)

[...]

PR #4256, opened 2020-09-26
Improve subplot when MAP_FRAME_TYPE=inside

PaulWessel:

See #4235 for context.  This small PR deals with the calculation of panel size and margins in the event **MAP_FRAME_TYPE**=_inside_ before **gmt subplot begin** is run.  Previously we did not check, but with all annotations and ticks inside the panel, the margin is simply what is set via **-M**.  I added a simple 4-panel check to ensure this works.  This PR does not address the desire to have inside basemaps plot on top of the panel content, which will require much more work and changes.

### DIFF
--- a/src/subplot.c
+++ b/src/subplot.c
@@ -760,7 +760,7 @@ EXTERN_MSC int GMT_subplot (void *V_API, int mode, void *args) {
 		uint64_t seg;
 		double x, y, width = 0.0, height = 0.0, tick_height, annot_height, label_height, title_height, y_header_off = 0.0;
 		double *cx = NULL, *cy = NULL, *px = NULL, *py = NULL, y_heading, fluff[2] = {0.0, 0.0}, off[2] = {0.0, 0.0}, GMT_LETTER_HEIGHT = 0.736;
-		double master_scale = (GMT->current.setting.map_frame_type == GMT_IS_INSIDE) ? 0.0 : 1.0;
+		double master_scale = (GMT->current.setting.map_frame_type == GMT_IS_INSIDE) ? 0.0 : 1.0;	/* THe 0 helps wipe any dimensions outside the panel to zero */
 		char **Bx = NULL, **By = NULL, *cmd = NULL, axes[3] = {""}, Bopt[GMT_LEN256] = {""};
 		char vfile[GMT_VF_LEN] = {""}, xymode = 'r';
 		bool add_annot, no_frame = false;
--- a/src/subplot.c
+++ b/src/subplot.c
@@ -760,6 +760,7 @@ EXTERN_MSC int GMT_subplot (void *V_API, int mode, void *args) {
 		uint64_t seg;
 		double x, y, width = 0.0, height = 0.0, tick_height, annot_height, label_height, title_height, y_header_off = 0.0;
 		double *cx = NULL, *cy = NULL, *px = NULL, *py = NULL, y_heading, fluff[2] = {0.0, 0.0}, off[2] = {0.0, 0.0}, GMT_LETTER_HEIGHT = 0.736;
+		double master_scale = (GMT->current.setting.map_frame_type == GMT_IS_INSIDE) ? 0.0 : 1.0;
 		char **Bx = NULL, **By = NULL, *cmd = NULL, axes[3] = {""}, Bopt[GMT_LEN256] = {""};
 		char vfile[GMT_VF_LEN] = {""}, xymode = 'r';
 		bool add_annot, no_frame = false;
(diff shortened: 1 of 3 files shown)


seisman:
I changed your test to:
```
gmt subplot begin 2x2 -Fs3i -M5p -A -SCb -SRl -Bwstr -R0/80/0/10 -T"Inside annotations" --MAP_FRAME_TYPE=inside
```
and then it doesn't work.

PaulWessel:
Well, --PAR=value only applies to a single command, so when the later basemap calls come there is no such thing, despite us computing the correct spacing during begin. Hm, In the next PR I am adding a new line to the hidden subplotinfo file that remembers that inside was active, and we could use that to add --MAP_FRAME_TYPE=inside when we plot the frame.  That way the par setting will live until subplot end.  I agree this is what a user would want thought.
However, the current PR works correctly if used with gmt set, so lets accept that one and then I can add more along these lines.

seisman (approved):

PaulWessel:
Thanks, will work on this probably more tomorrow.

PaulWessel:
Question for @joa-quim and @seisman : In **psxy** and possibly other modules, we lay down the basemap _before_ plotting symbols/lines, because we want the latter to overprint any gridlines.  However, if frametype _inside_ is in effect, not only will symbols overprint any gridlines, but they will overprint annotations.  I think we agree that when _inside_ is used we want these things to be on top of the plot.  With the current scheme, even if I figure out a way to add the requested subplot **-B** options to the last plot call in a panel, it may _still_ get obscured by the **plot** module.  I am also not sure if annotations and ticks and gridlines should all be considered at the same level with _inside_.  But I guess they should all go on top?

Solution: Ensure all plot modules switch to calling _gmt_map_basemap_ after all plotting if inside is in effect.  Note some modules, like **grdimage**, only do this at the end anyway.

joa-quim:
Yes, in ``grdimage`` the grid goes on top, otherwise it wouldn't be visible. But for plot they have to go **under** the symbols/lines. We had this discussion when we found that ``plot`` and ``plot3d`` were doing different things.
See this Matlab [example](https://www.mathworks.com/help/matlab/ref/scatter3.html). It would be awful if grids were on top. And we don't want ``plot`` and ``plot3d`` do different things. Inner annotations, though, should preferably go on top of everything.

PaulWessel:
OK, so sounds like many modules should call gmt_map_basemap twice: before and after, but inside there will be tests to only do one or only do part (say gridlines are always on bottom except for images).  Need some temp variable to remember that girdlines were or were not plotted earlier.

PaulWessel:
My first try on this.  Getting 60 failed tests because now gridlines are behind things like psvelo ellipses, histogram bars, etc.
I am guessing we want that, i.e., gridlines on the back except for images (grdimage, psimage, grdview) and pscoast.  ANything else that should have gridlines on top?

PaulWessel:
What about grdcontour and pscontour when we will with color? Kind of like images. no? But just regular contours are like lines, so gridlines on back.

joa-quim:
Yep, those two should display a binary behavior. Grid on top when fill color and bellow when contours only.

PaulWessel:
And histograms should cover up gridlines as well?

joa-quim:
Both Matlab and Matplotlib put the grid lines on top, though our logic would say that they should go under because a bar is a kind of a symbol.
What histogram should have is this ``BarWidth`` option (in Matlab)
```
>> A = [0 0 1 1 1 0 0 0 0 NaN NaN 1 0 0 0 1 0 1 0 1 0 0 0 1 1 1 1];
C = categorical(A,[1 0 NaN],{'yes','no','undecided'});
>> histogram(C,'BarWidth',0.5)
>> histogram(C,'BarWidth',0.5); grid on
```

And the hexagonal binning? It probably has gridlines under but since it makes an image the lines should go on top.

PaulWessel:
We could introduce **-E**_width_[**unit**][**+o**_off_[**unit**]] where _width_ changes the plotted width of each bar.  If units are applied then that is a firm width, else it is assumed to be in x-units.  Optionally the modifier _off_ could shift the bar from its normal position.  That way you could plot more than one histogram for the same binning scheme by shrinking size and offsetting them.

Hexagonal binning are like images so need gridlines on top, but is done via call to plot.  I will need to see how to fix that.

PaulWessel:
hexagonal binning is just a separate plot with hexagon symbols.  So plot cannot know that perhaps you might want gridlines on top.  Users can always override this behavior at any time by adding a basemap call.

PaulWessel:
How about solar? Kind of like pscoast in plotting a large polygon so gridlines on top, @joa-quim ?

joa-quim:
Yes, solar is like pscoast.

joa-quim:
> We could introduce -Ewidth[unit][+ooff[unit]]

The horizontal offset would be useful in bar plots too (``psxy -Sb``)

PaulWessel:
Hm, maybe.  There we have at least -D to do that, for the most common case of shifting by a plot dimension offset.

PR #4260, opened 2020-09-28
subplot set fails if not given index|row,col

PaulWessel:
Per the subplot documentation, **gmt subplot set** without any _row,col_ or _index_ shall go to the next logical panel, but in reality it fails parsing by trying to read the next option.  This PR checks if the next option is a non-option argument (hence flagged as "input file") and only then tries to parse.  Otherwise we default to the next panel.


PaulWessel:
Pinging @joa-quim and @seisman

seisman (approved):

PR #4263, opened 2020-09-28
Plot bar-symbol errorbars on top of symbol

PaulWessel:
Unlike regular (smaller) symbols with error bars, a bar symbol usually goes all the way from the axis to the value.  Plotting an error bar behind it means the lower bound of the bar will be obscured.  Here we plot the errorbar on top if bar symbols are used.  This is now clarified in the docs.  Tests pass but we have no test for this so need @joa-quim to try this before approving as he pointed this out.


joa-quim (approved):
Added a test script but it misses the orig PS

Issue #4265, opened 2020-09-28
Consistency on placement of gridlines and basemaps


PaulWessel:
**Description of the desired feature**

There are several aspects to consider regarding when to place gridlines, basemap frame and annotations+ticks on a plot.  We recently learned that when **MAP_FRAME_TYPE**=_inside_ then it is important that the annotations and ticks are plotted at the end of the module, whereas we typically lay down the annotations first. **Note:** The user can always achieve any other result by using a single **basemap** command before or after the data plotting, but we should have good and consistent default actions.

The normal rule has been that we always lay down the full basemap (frame, annotations+ticks, optional gridlines) prior to plotting anything else in the module.  This rule is ignored for the following cases:

1. For modules plotting filled polygons (data, coast, solar, etc.) the gridlines shall be plotted at the end of the module (if 2-D).
2. For modules plotting images (grdimage, grdview if 2-D, image, grdcontour with fill) the gridlines shall be plotted  at the end of the module.

If **MAP_FRAME_TYPE**=_inside_ then annotations shall be plotted at the end of the module.

Unresolved is the drawing of the base frame itself. For some modules like **plot -N**, we will plot symbols that normally can overprint the frame.  To me, it looks ugly to have the frame plot on top of these points.  Also, symbols that may start at the baseframe, such as vertical or horizontal bars, will overprint the frame by at least half the frame pen (and more if there is bar outline).  In contrast, here I think plotting the baseframe last looks better.  Two options:

Always draw baseframe at the start.  This means items can and will plot on top of the frame.
Always draw baseframe at the end.  This means symbols at or on the border will have a fat line cutting through them.

Whatever we decide, some tests may need to be recoded to accomplish what we wanted to show.  Comments, @joa-quim and @seisman ?

PaulWessel:
One thought: It may be that there is no single system that always will work and users may have to add extra commands, which is tedious.  How about we make this entire thing an option? E.g., the **-B** frame setting could take an option **+l** for last and then we plot it at the end, else the beginning.

PaulWessel:
Also, it seems data should take precedence over frames, so plotting data overlapping on frames should be more important.

seisman:
I always prefer a simple and consistent system, which is easy to document (for us) and to understand (for users). 

Thus, for base frame, I prefer to always have it at the start. If users don't like it, it's easy to call the basemap module after.

> the **-B** frame setting could take an option **+l** for last and then we plot it at the end, else the beginning.

[...]

Issue #4181, opened 2020-09-07
Double frame on grdview perspective plot


weiji14:
**Description of the problem**

There appears to be a double-frame on the x and y-axis when making a perspective plot using `grdview` with the `-Jz` flag. Originally detected at https://github.com/GenericMappingTools/pygmt/pull/589#discussion_r484320626.

| Actual (incorrect) plot | Expected (correct) plot |
|---|---|
| ![test_grdview_with_perspective_and_zaxis_frame-expected](https://user-images.githubusercontent.com/23487320/92370449-6a37b400-f14e-11ea-968f-6a520bb7d6aa.png)| ![test_grdview_with_perspective_and_zaxis_frame](https://user-images.githubusercontent.com/23487320/92370444-69068700-f14e-11ea-8717-f7194f26bb58.png) |

**Full script that generated the error**

```
gmt begin doubleframe png
    gmt grdcut @earth_relief_01d_g -R-116/-109/-47/-44 -Gtempgrid.nc
    gmt grdview tempgrid.nc -p225/30 -Jz0.005 -Bzaf -Vd
gmt end
```

**Full error message**

<details>

```
gmt [DEBUG]: Obtained the ppid from parent: 176
gmt [DEBUG]: Enter: gmtinit_new_GMT_ctrl
gmt [DEBUG]: GMT->session.SHAREDIR = /srv/conda/envs/notebook/share/gmt
gmt [DEBUG]: GMT->session.HOMEDIR = /home/jovyan
gmt [DEBUG]: GMT->session.USERDIR = /home/jovyan/.gmt [created]
gmt [DEBUG]: GMT->session.CACHEDIR = /home/jovyan/.gmt/cache [created]
gmt [DEBUG]: GMT: 0. Will try to find subdir=postscriptlight stem = PSL_custom_fonts suffix=.txt
gmt [DEBUG]: GMT: 1. gmt_getsharepath trying current dir
gmt [DEBUG]: GMT: 2. gmt_getsharepath trying USERDIR /home/jovyan/.gmt
gmt [DEBUG]: GMT: 3. gmt_getsharepath trying USERDIR subdir /home/jovyan/.gmt/postscriptlight
gmt [DEBUG]: GMT: 4. gmt_getsharepath trying SHAREDIR subdir /srv/conda/envs/notebook/share/gmt/postscriptlight
[...]

PR #4274, opened 2020-09-30
Realigning order of basemap frame, gridlines, and annotations/ticks for all modules

PaulWessel:

See #4265 for contest.  We want to add consistency to how GMT places the "basemap" on top of or below a plot within the same module (obviously, stacking modules can override any default arrangements).  In general, we now place the optional gridlines before plotting data, but include images and polygons (e.g., via **pscoast**) in which case gridlines goes on top.  We generally lay down the frame on top to ensure any lines or symbols at or near the boundary do not "eat" into the frame width, but if the module has a no-clip option like **-N** then we draw it first so symbols can straddle that line.  Annotations and tick-marks are usually laid down before the plot but in the case of **MAP_FRAME_TYPE**=_inside_ then we must do it at the end instead.  There are also some exceptions for true 3-D plots.

Implementation is done via a new function _gmt_set_basemap_orders_ which sets, for each of the discussed items, what the order of plotting should be.  For instance, in **pscoast** it may say

`gmt_set_basemap_orders (GMT, clipping ? GMT_BASEMAP_FRAME_BEFORE : GMT_BASEMAP_FRAME_AFTER, GMT_BASEMAP_GRID_AFTER, GMT_BASEMAP_ANNOT_AFTER);`

where clipping options may require the frame to be drawn first instead of last, but otherwise we draw items later.  Each plot module has such a call and the arguments vary.  The function itself can override the settings based on some of the conditions discussed above.

After implementation, we now have ~180+ "failures" in the tests.  Most of them are of these types:

Some plots that had gridlines on top now has gridlines on the bottom, avoiding cutting through symbols and bars
Some plots with polygons or bars near a border (e.g., histograms) now have a clean frame line without notches
Many base frames have subtle changes because tick marks are not plotted and are visible whereas in some cases they were covered up.

I have labeled this PR WIP because I think we need to actually look at all these to see if anything got lost in the translation that we want to preserve.


PaulWessel:
I know this is a big tasks to look at those 180 PNG and PDFs but I think we need all of us to do this in case there are some subtle regressions.  I've stared at these for a while and the results look good to me.  Let me know of you see anything that should not change, @joa-quim and @seisman.

joa-quim:
I got *only* 133 failures in total.

PaulWessel:
That is a new one - less failures on Windows than macOS (186 right now).

joa-quim:
Ex16 is partially eating the frame

PaulWessel:
Yes, I need to fix the setting for grdview.  It draws before but shoud be after (unless 3-D etc).

seisman:
- `doc/script/GMT_inset.sh`: Tiny changes in the coastline 
- `ex04`, `ex05`: Tiny changes in the title string
- `ex08`: The z-axis at the back should be invisible  
- `ex11`: this example is messy now.
- `ex15`: in panel (d), some pixels overlap with the frame. The old PS looks better to me.
- `ex16`: @joa-quim already mentioned above

joa-quim:
grdcut/origin. The inner pijama frame is overlapped
image

PaulWessel:
> * `doc/script/GMT_inset.sh`: Tiny changes in the coastline
No idea why coastline is affected, but perhaps I removed a pen reset - will look
> * `ex04`, `ex05`: Tiny changes in the title string
Unsure about what that is
> * `ex08`: The z-axis at the back should be invisible
working on that
> * `ex11`: this example is messy now.
Yes must be a major error I missed
> * `ex15`: in panel (d), some pixels overlap with the frame. The old PS looks better to me.
Fixing that in mask
> * `ex16`: @joa-quim already mentioned above
Same, fixing in grdview

PaulWessel:
> grdcut/origin. The inner pajama frame is overlapped

But that was always the case.  We do -B in the first command and then pile on.  This needs moving the -B to the last call.

PaulWessel:
I think we have many tests where we plot numerous overlays and often start with **-Baf**.  To follow our plan I think these need to move to the last command.  Finally, we do not yet have a solution for subplots since it will add -B settings to the **first** command in the panel, not last.  Changing this is harder and will be pursued after this PR.

joa-quim:
grdimage/grdclip

image

joa-quim:
grdview/categorical

image

joa-quim:
grdview/grdview image is eating half-frame. Probably the same issue as above.

joa-quim:
- psxy/diagonal Where is that green coming from?
image

PaulWessel:
Thanks, I am probably missing a final stroke in the -B stuff before psxy changes the pen to green.
Will work on changes after the next zoom starting in 5 minutes.

PaulWessel:
Some failures are due to us plotting ticks after the frame.  This exposes bugs that were always there, but covered up by the frame.  Using a thinner frame lets us see the rogue ticks in both PR and master.  Here is the very top of the pac-man like conic basemap from test/grdimage/grdclip.sh resulting from this PR (which triggers a failure):

map2

and this is what master has:

map1

Easy to see on the right that in master the rogue tick is normally covered by the white thick frame but for this PR the tick is on top.  The solution is to fix the tick angle bug but leave ticks on top, since by selecting a plain frame the ticks are very obvious.

joa-quim:
One more but don't know if anything can be done to this
```
julia> imshow(G, view=(45,90), xlabel="X", ylabel="Y", fmt=:png, Vd=1)
        grdview  -JX12c/0 -Baf -Bza -Bx+lX -By+lY -p45/90 -Qi100 -JZ8 -R-3/3/-3/3 -C -P -K > C:\TMP\GMTjl_tmp.ps
```
image

However, using ``grdimage`` instead of ``grdview`` and not rotated this time
```
julia> imshow(G,  xlabel="X", ylabel="Y", fmt=:png, Vd=1)
        grdimage  -JX12c/0 -Baf -BWSen -Bx+lX -By+lY -R-3/3/-3/3 -C -P -K > C:\TMP\GMTjl_tmp.ps
```
image

joa-quim:
and with ``grdview`` the axes are even a bit eaten at some places

image

PaulWessel:
It seems like apart from the modules with a -N, we should always end with frame plus ticks plus annotations, while gridlines depends on what is plotted. I will review all those BEFORE vs AFTER settings again.

PaulWessel:
I have made quite a few changes to this PR.  THe problem I have is that there will be > 100 PS changes.  I am hoping to do that this way:

1. For the modules with quite a few tests, I will run ctest -R module and examine just that subset (e.g., grdimage or pscoast).
2. If all the changes are benign (i.e., as expeted, ticks on top of frame, etc) I will update those only in a commit to this branch

I will then repeat this until I need to find another scheme to just handle a few examples at the time.  Is this an OK approach, @seisman?  I dont want to look at 160 images at once.

PaulWessel:
With the new requirement that ticks shall follow frame, we get 280 failures.  Here is one example of a plan basemap "failure":

rose_mag2

Before we proceed, shall we discuss this a bit? For plain frames the scheme of ending with the frame never prevented ticks from being seen, so visually there is no change despite the failure.  For fancy frames though, given our stupid fat frame default width, all ticks are covered up by the frame.  Again, not seen as a problem since the black/white checkerboard acts as ticks as well, but if you shrink the width of the fancy frame then you will notice that ticks are not visible in the white parts of the fancy frame, but now they are (and gives similar failures to the above case).  And as the **-B** doc example with both primary and secondary annotatinos and ticks showed, the white frame completely masks the ticks.  So I think drawing the ticks on top is better, but I would like to hear the opinions of @joa-quim and @seisman.

joa-quim:
I’m away from the computer for some hours so can’t test and reply much but I’m a bit confused with this ticks. It would be nice to have examples of the other, worst?, case. 
It’s silly to plot ticks with different color than frame but in that case ticks on top look bad.

PaulWessel:
Agreed, I will make a few examples of plan and fancy with colored ticks so we can truly see.  We may find we want to treat fancy separate from plain, for instance.  More in a bit.

PaulWessel:
A quick check with colors makes it clear that, at least for plain basemaps, we should place the ticks before the frame, always.  For oblique ticks, it is also clear that we must always have a frame pen equal to or larger than the tick pen, otherwise a corner of the tick could protrude off the frame.  With sensible pen the starting point of the tick is covered up.  We may wish to add a warning if anyone sets a fat tick pen, for instance.
I will do some tests with fancy and primary/secondary.  I note the B_geo_2.sh does not seem to draw secondary ticks despite them being requested so need to get to the bottom of that first.  However, many balls in air simultaneously (cooking, expanding psevents to do lines/polygons, and actually not working).

PaulWessel:
So plain frames are solved and will stay as they were: ticks overlain by frame.

Looking at fancy basemaps. We place ticks with annotations but we never did place ticks for frame intervals - that we left for the checkerboard to indicate. So for me, this means there is no great reason to place the ticks on top and make 100 failures.
The only case that might be different is when we do fancy with both primary and secondary annotations - which I think we only do once or so.  Here is the GMT_-B_geo2.sh output.  First master:

GMT_-B_geo_2_old

The key thing here are the secondary ticks (which are drawn) but are partially covered up by the white frame.  E.g., at 1:15W.  In the current new scheme this figure looks like this:

GMT_-B_geo_2_new

The key difference is that the ticks are drawn last and hence cuts through the white.
One possible solution would be to delay ticks for the special case of fancy with both primary and secondary annotations, else stick to the old scheme of laying down ticks first.  Any comments on this, @seisman and @joa-quim ?

PaulWessel:
I should add I am not sure it is worth having one exception.  If so then this big detour can end and we are just back to placing frame before data plus the gridline placements.

joa-quim:
But we don't want the ticks to cut through the whites as they don't cut through the blacks (irrelevant because invisible). The pajama stripes were set to a certain length, so they shouldn't be sliced. 
I guess this means, no work needed for this case.

PaulWessel:
OK, I agree, so then we do grid, tick, frame in that order (except some grids on top).

PaulWessel:
Back to 82 failures now and most are expected due to the basemap coming at a different time.  I will review them and commit updates in batches.

PaulWessel:
I have completed my review and update of all the ~60 failures after deciding (or rediscovering) that we do want to place ticks beneath the frame.  With that settled, what remains was the initial idea of having clean base-frames without having symbols eat into eat (e.g., histograms) and by default placing gridlines behind plotted data, with exceptions for polygons and landmasses.  In the process I ended up fixing several scripts that had the **-B** options placed on the wrong overlay to achieve what we agreed should be the correct order of things.  From **rose** examples I fixed an inconsistency in the scale bar and now document what pens arne fonts are used for this bit, and the same for wiggle. Some rose examples plotted the basemap 3-4 times on top of each other, now fixed. This PR leaves three examples (that pass) with what I will consider issues, but they are the same problems in master (so not a reason to complain here):

1. grdcontour/polcont.sh: The frame half-circle and the diameter are not connected into one line, leaving a notch at the two ends due to the joint settings.
2. psxy/poleline.sh: The very lowest gridline shaves a tiny piece of the SW corner but the steps we use along the line is too large for this tiny cut, causing a bad 2nd latitude annotation and no gridline.  This goes away if I shrink **MAP_LINE_STEP** but I have to do it by quote a bit (0.05p vs default 0.75p) so it slows down. To be worked on more.
3. psxyz/3dbars.sh: The 3-D frame is plotted, but then some bars are plotted and they overprint thefront vertical axis which gives the wrong depth view.

I will open a new issue on those once this one is approved.  So @joa-quim and @seisman please have a look at the corrections - I have manually gone through them one by one and committed in batches.

joa-quim:
I don't get it why I have 4 tests that fail because of the ``@earth_relief_05m`` grid
```
143 - doc/scripts/GMT_tut_11.sh (Failed)
150 - doc/scripts/GMT_tut_18.sh (Failed)
177 - doc/examples/ex08/ex08.sh (Failed)
432 - test/grdimage/grdcyclic.sh (Failed)
```
I deleted the grid from cache but no difference.

PaulWessel:
And your gmt_data_server.txt is up to date?

joa-quim:
It should update itself, no? It says
```
98
# 6.1.0 or later GMT version required
#
# Master table with information about all the remote data sets available on the GMT Data server
# Changes related to data on the gmtserver must be added into this file before they are available to users.
# Set your editor to TAB width = 4 for alignments.
#
# Updated Sept 21, 2020
```

PaulWessel:
Yes, that is what I have as well.  What sort of failures?

joa-quim:
Minuscule differences in the contours, not even visible but to ``
GMT_tut_11
gm``.

seisman:
The following results may not make sense to me (some may not be related to changes in this PR).

`test/pshistogram/altwidth.ps`:
image

`doc/anim03/anim03.sh`:
image

`test/psxy/antwrap.sh`: plot polygon first then gridlines?
image

`test/psxy/badpols.sh`: is this what we want?
image

PaulWessel:
altwidth.ps: This is the problem with subplot that auto-adds -B to first command.  Then 2nd histogram comes and plots onto the axis. anim03: WIll have to look - never noticed that.  Antwrap: We decided to plot gridlines on top of large polygons so I will fix this one, may be order of things, badpols: Premature placement of -B which should be on the last command per plot.  Thanks, will fix these later today,

PaulWessel:
Hi @seisman, I have now implemented what we had agreed on earlier (gridlines should be on top of polygons from plot, plot3d, solar) so that changes the antwrap.ps result.  The badpols.sh simply had -Baf before plotting, now adjusted.  THe change to plot and plot3d affected a handful of other results since now the gridlines are on top.  The altwidth.sh cannot be fixed until subplot can do frame as last overlay in a panel, and the anim03 is not related to this PR but seems related to an grdimage issue to be determined.

seisman:
@PaulWessel There are so many conversations in this PR, and we made several decisions in different threads. It's a little difficult for me to follow what the decisions are.

Could you please summarize the basic rules (ticks->gridlines-data->basemap, if my understanding is right),  exceptions (e.g., data before gridlines for pologyons, images), and known issues (e.g., subplot), in the first post of this PR?

seisman:
- ex06 and `test/psrose/*.sh`: gridline shoud be on top of rose histogram?
- ex33, upper panel: should plot basemap after data
- ex41: 
image
- ex50: 
image
- `test/psxy/ErrorEllipseGeo.sh`:
image
- `test/psxy/filler.sh`, `test/psxyz/filler.sh`
- `test/psxy/kara.sh`
- `test/psxyz/3dbars.sh`: the bar should be behind the thin "frame"?
image

PaulWessel:
Regarding the rules we follow, I think these are it, with some exceptions noted later:

1. Gridlines should go first since data are more important that gridlines
2. Annotations, if outside, can go before plotting or after, usually not critical, but since ticks are done at same time we place them before the frame
3. Frame is normally at the end to cover up any lines or symbols touching the domain box.

Exceptions:

1. For polygons, images, and filled contour plots we lay gridlines on top
2. For MAP_FRAME_TYPE_inside we also do annotations at the end of the module

This scheme treats histogram bars as symbols and hence they will cover up gridlines. Your example above of **psrose** is a type of histogram so even though the rose diagram is a polygon I think we should be consistent on treating them all as histogram bars.  The 3D example is on my list as well.  Clearly, we need to be able to plot the background 3-D part separately form the foreground lines since otherwise we will always end up with this dumb problem.

I will work on the other examples you mentioned.

PaulWessel:
OK, I have improved the case for psxy[z] and updated 23 originals.

seisman:
Remaining "problematic" figures after your fixes:

- ex50 row 7: still eat half frame 
- test/psxy/ErrorEllipseGeo.sh: compared to `test/psxy/ErrorEllipseCart.ps`, why are there blue lines near the boundaries.
- test/psxyz/filler.sh: the psxy one looks good, but psxyz one doesn't
- test/psxyz/3dbars.sh: 
> The 3D example is on my list as well. Clearly, we need to be able to plot the background 3-D part separately form the foreground lines since otherwise we will always end up with this dumb problem.

Perhaps I should open a separate PR for the 3dbars plot, so that we won't forget it.

PaulWessel:
Yes, please make an issue re 3-d frames so I can get back to that one later.

PaulWessel:
The ones you listed minus 3dbars should be OK now.

seisman:
This is the latest `test/psxy/ErrorEllipseCart.ps`:
ErrorEllipseCart

This is the latest `test/psxy/ErrorEllipseGeo.ps`:
ErrorEllipseGeo

Two blue lines show near the frame boundary in `test/psxy/ErrorEllipseGeo.ps` but not in `test/psxy/ErrorEllipseCart.ps`. Is that what we want?

PaulWessel:
No, but this is a "but" in the geographic ellise drawer, nothing to do with -B.  It draws the entire ellipse with outline then clipping, instead of drawing the fill first, then draw a line (that will get chopped at the border).  Can open a separate issue for that.

PaulWessel:
@seisman if u have time, please run tests to see if the psxy change not only fixes the geo ellipse but not causing trouble elsewhere [I am at clinic so running on MacBook air...].  If ok I will make a more permanent fix and probable in psxyz as well.

seisman:
> @seisman if u have time, please run tests to see if the psxy change not only fixes the geo ellipse but not causing trouble elsewhere [I am at clinic so running on MacBook air...]. If ok I will make a more permanent fix and probable in psxyz as well.

Now I have ~280 failures.

PaulWessel:
Me too. trying again.

PaulWessel:
OK, now down to 18 failures.  Looks lik pen width differences related to the difference of plotting online _after_ plotting fill versus plotting a filled outlined polygon?

seisman:
Yes, most differences are tiny, but anim03 is wrong. Here is my anim03 diff:
image

seisman:
The new `test/psxy/ErrorEllipseGeo.ps` looks good, and the PS file should be updated too.

PaulWessel:
I added the ellipse ps. I don't understand anim03.  It is like it ignores -p entirely if I want to place frame on top...

PaulWessel:
Just a note: anim03.sh (before we lost the 3-D view on it, TBD) also had that scalloped baseframe in master.  So that is a separate issue from this PR once I fix the 3-D view.  I ran it in master and cut out from the PS the entire image placement and the scalloping went away.  It seems related to the clip path.  Here is the view after removing the image and letting the clip path be stroked in red pen:

anim03

Looks like unfortunate round-off but too large to be integer PS units - must be something in the original sampling of meridians and parallels I think.

PaulWessel:
OK, the cause is this: The image-derived clippath (tracing the pixel coordinates) results in a rough pixellated clip path.  Since it basically tracks the domain projected rectangle the clip path exceeds the domain in some places.  Because the frame is placed before the image, the image is overlain within the clip path and will paing its near-white onto the clip path, resulting in the jagged appearance.  The solution is obviously to plot the frame at the end, which is what this PR does, except from the (temporary) problem at we end up with a 2-D non-projected map...

PaulWessel:
Hi @seisman, I fixed the anim03.sh issue.  I updated that PS and it passes the testing.  However, I noted a weird thing I have seen before but cannot recall if it just went away after reboot or not: The new PS file looks great in ghostview (gv) and test passes, but if I tried open anim04.ps then Preview would show the black checkerboard at the wrong angle.  Using psconvert first did the same.  So that got me worried, but I think it is OK; see what happens on your end as well.  Please run the tests: I did not (yet) update the tiny changes so wanted to run that by you first if you see anything else odd.  My failures are:

```
	187 - doc/examples/ex18/ex18.sh (Failed)
	193 - doc/examples/ex24/ex24.sh (Failed)
	343 - test/gmtspatial/nn.sh (Failed)
	346 - test/gmtspatial/sphareacentroid.sh (Failed)
	347 - test/gmtvector/meanvec.sh (Failed)
	395 - test/grdcut/origin.sh (Failed)
	409 - test/grdfill/nnfill.sh (Failed)
	495 - test/grdvector/vectors.sh (Failed)
	524 - test/mapproject/oblmerc_down.sh (Failed)
	525 - test/mapproject/oblmerc_up.sh (Failed)
	660 - test/psevents/events.sh (Failed)
	914 - test/sph/sph_5.sh (Failed)
```

and I think they are all tiny issues with lines.  Let me know if you are OK with me updating these.

PaulWessel:
Apologies for fixing #4181 in this PR but I wanted to be safe it did not break other things.  The double frame is now gone.

seisman:
> : The new PS file looks great in ghostview (gv) and test passes, but if I tried open anim04.ps then Preview would show the black checkerboard at the wrong angle.

Yes, we saw it before and we thought it may be a Preview bug.

seisman:
I have two more failing tests than you, but there are just tiny differences:
```
	596 - test/psbasemap/asymm_grid.sh (Failed)
	640 - test/pscoast/placement.sh (Failed)
````

**Comment by PaulWessel** [comment]
So OK with me updating the orig PS in this PR then?

**Comment by seisman** [comment]
Yes to me.

**Comment by PaulWessel** [comment]
Hi @seisman ane @joa-quim, I think we have reached the end of the line on this PR.  This clearly improved the quality of the plots.  The one thing remaining is related to subplot panels where the -B happens on the first panel but perhaps should always be last.  That will require more work since we cannot know we are last until we either move to a new frame or we are done.  That is a separate PR.

**Comment by seisman** [comment]
```
../../src/psrose.c:799:10: warning: unused variable 'dim' [-Wunused-variable]
                double dim[PSL_MAX_DIMS];
```

PaulWessel:
Hi @seisman, seems a CI failure but nothing to be concerned about?  @joa-quim, may be helpful if you could run all the test one more time on this branch.  I think all is well now with expected result apart form the notes on subplot panel orders.

seisman:
> Hi @seisman, seems a CI failure but nothing to be concerned about?

Yes, just an occasional network connection issue.

joa-quim:
Same number of failures (16). Some, ``doc/scripts/GMT_tut_11.sh, 18, etc`` fail because the grid is slightly different. Don't get it.

seisman:
I just make a commit 784347f22634b292d9c192bfdc4e0ae2e43046a4 to trigger full tests in CI.

seisman:
This PR is good to me, but please check this file `test/psxy/clipping5.ps`. 

image

PaulWessel:
OK, soon after finishing another PR.

PaulWessel:
So clipping5.sh is flagged as **GMT_KNOWN_FAILURE**.  I think it is not improved or degraded by this PR though; its solution lies elsewhere.

seisman:
> So clipping5.sh is flagged as **GMT_KNOWN_FAILURE**. I think it is not improved or degraded by this PR though; its solution lies elsewhere.

OK, I didn't notice that. The CI looks good. One failure on Linux and 7 failures on Windows. They are all known failures.

seisman (approved):

seisman:
I just disabled the full tests in e529b14880e377df294a2ee759afb8163de5c9f3. Now this PR looks good to me.

PR #4276, opened 2020-09-30
Fix another bunch of CIDs

joa-quim:


PaulWessel (approved):

PaulWessel:
I did not run tests before approving them... @joa-quim probably did not run the tests before submitting?

joa-quim:
No I didn't but ex11 fails with the same garbage as had already failed in frames branch. ``splinefill`` passes and ``rgb_grids`` exists with a parsing error. Trying to figure it out.

PR #4281, opened 2020-10-01
Fid bad CID fix that broke the r,g,b grids case

joa-quim:


seisman:
@PaulWessel just made some fixes in #4280, which causes some conflicts with this PR.

joa-quim:
In your PR, delete the line around 7964, the one with ``gmt_M_str_free (input);`` and I can delete this PR

PaulWessel:
I am correctly freeing the input string and not correctly passing infile to the function that returns, so I think master has this covered as is?

joa-quim:
Yes, looks good. So I'll close this PR, right?

PaulWessel:
Yes, think so.  Sorry, did not notice you were working on that

PR #4289, opened 2020-10-06
Let events handle time-sensitive lines and polygons

PaulWessel:

So far, **events** can plot various symbols that each have their own duration, with optional enhancements like rise time, decay, and fading.  However, there is also a need to be able to plot lines and polygons on a temporal basis in animations.  This PR adds this capability.  A new option **-A** selects lines or polygons, whereas the existing **-S** option selections symbols.  We support two modes for lines and polygons via **-A**:

1. **-Ar**: Lines only, and these lines are given with three columns, _x y time_, and for the current time passed to **events** (via **-T**) we determine which _records_ of the line falls inside the event duration and just plot that segment.  For now, the line can only have a single pen/color, but a likely extension to this would be to assign a color based on a CPT via time, since the entire segment can only have one color.
2. **-As**: Polygons or lines, but per _segment_.  Here we read multi-segment files and each segment must have a header option **-T**_string_ that indicates the time (and possibly end time or duration, depending on **-L**) for the segment.  Thus, when the current plot time falls within the event duration we plot the entire segment (as line or polygon).  Unlike **-Ar**, this mode allows color to be assigned via **-Z**_val_ in the header (and via CPT lookup) and users can supply **-W** and **-G** options as well.  This mode also supports variations in transparency during the optional rise and fad sections of the even and passes a **-t**_transparency_ via the output segment header that **plot** understands.

As an example of basic operations, here is a simple movie that shows how line animation may work:

line

Likewise, here is a simple movie showing two static polygons (of course, sophisticated applications could produce lots of segments with motion in space during time as well):

pol

This PR updates the code, documentation, and adds a simple test for each modes to the test/psevents folder.

### DIFF
--- a/doc/rst/source/events_common.rst_
+++ b/doc/rst/source/events_common.rst_
@@ -155,7 +155,7 @@ Optional Arguments
     Optionally, append symbol *size* with unit from
     (**c**\|\ **i**\|\ **p**\ ).  If no *size* is given then we read an event-specific size
     from the data file's third column (fourth if **-C** is used). **Note**: Not all the symbols that
-    are available in :doc:`psxy` can be used here.  At the moment we only support the basic
+    are available in :doc:`plot` can be used here.  At the moment we only support the basic
     symbols and custom symbols; bars, vectors, ellipses, fronts, decorated and quoted lines cannot
     be specified.  Symbols sizes read from a data file will be assumed to be in the unit controlled
     by :term:`PROJ_LENGTH_UNIT` unless they have **c**, **i**, or **p** appended.  If you are only
--- a/doc/rst/source/events_common.rst_
+++ b/doc/rst/source/events_common.rst_
@@ -29,18 +29,6 @@ Required Arguments
 .. |Add_-Rz| unicode:: 0x20 .. just an invisible code
 .. include:: explain_-Rz.rst_
 
-.. _-S:
-
-**-S**\ *symbol*\ [*size*]
-    Specify the symbol to use for the event.  Optionally, append symbol *size* with unit from
-    (**c**\|\ **i**\|\ **p**\ ).  If no *size* is given then we read an event-specific size
-    from the data file's third column (fourth if **-C** is used). **Note**: Not all the symbols that
-    are available in :doc:`psxy` can be used here.  At the moment we only support the basic
-    symbols and custom symbols; bars, vectors, ellipses, fronts, decorated and quoted lines cannot
-    be specified.  Symbols sizes read from a data file will be assumed to be in the unit controlled
-    by :term:`PROJ_LENGTH_UNIT` unless they have **c**, **i**, or **p** appended.  If you are only
-    plotting labels then **-S** is not required.
-
 .. _-T:
 
 **-T**\ *now*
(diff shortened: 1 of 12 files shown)


doc/rst/source/events_common.rst_
```diff
@@ -54,14 +42,29 @@ Optional Arguments
...
+**-A**\ **r**\|\ **s**
+    When no **-S** is given we expect to read lines or polygons.  Two different forms for input are
+    considered and specified via **-A**\ : (1) Choose **-Ar** if your input data are *lines*, the data
+    format is given as *x, y, time*, and the "event" is the portion of that line limited by the current
+    time (*now*), event duration and rise/fade periods (via **-L** and **-Es**), or (2) choose **-As**
+    to expect to read each complete segment (i.e., a polygon or line with just *x, y*) from a multisegment
+    file and get a common time for each segment via a required **-T**\ *string* specified in the segment
+    header.  Here, the entire polygon is the "event" if inside the limits of current time (*now*), event
+    duration and rise/fade periods.  If **-L**\ [**t**] is set then the *string* must be either
+    *begin*/*length\|\ end* or *begin*,*length\|\ end* (if *begin* is absolute time).  **Note**:
```
seisman, line 56:
```suggestion
    *begin*/*length*\|\ *end* or *begin*,\ *length*\|\ *end* (if *begin* is absolute time).  **Note**:
```

src/gmt_io.c
```diff
@@ -7113,6 +7115,10 @@ int gmt_parse_segment_header (struct GMT_CTRL *GMT, char *header, struct GMT_PAL
 
 	if (!header || !header[0]) return (0);
 
+	if (gmt_parse_segment_item (GMT, header, "-t", line)) {	/* Found a potential -t option for transparency */
+		transparency = atof (line) * 0.01;
+		got_transparency = true;	/* Must apply to any fill and pen settings */
+	}
```
seisman, line 7121: Does this mean **plot** and **plot3d** can understand **-t** in the segment header? If yes, you may want to add **-t** to https://docs.generic-mapping-tools.org/dev/plot.html#segment-header-parsing.

PaulWessel: Yes, I was debating if it should be a hidden trick for psevents to work or an expansion of capability for all.  I guess it is a reasonable extension so I will update the docs.

PaulWessel: OK, done

doc/rst/source/events_common.rst_
```diff
@@ -142,6 +148,19 @@ Optional Arguments
...
     Files will be named *prefix*\ _symbols.txt and *prefix*\ _labels.txt.
 
+.. _-S:
+
+**-S**\ *symbol*\ [*size*]
+    Specify the symbol to use for the event [Default plots lines or polygons; see **-A**].
+    Optionally, append symbol *size* with unit from
+    (**c**\|\ **i**\|\ **p**\ ).  If no *size* is given then we read an event-specific size
+    from the data file's third column (fourth if **-C** is used). **Note**: Not all the symbols that
+    are available in :doc:`psxy` can be used here.  At the moment we only support the basic
```
seisman, line 158:
```suggestion
    are available in :doc:`plot` can be used here.  At the moment we only support the basic
```

seisman (approved):

PR #4294, opened 2020-10-07
Fix two PS files that are not correctly finalized

seisman:
These two PS files are good, but they are not correctly finalized (-K, -O).
This PR updates the two PS files and also the script pssac/pssac_W.sh.


PaulWessel (approved):
I am about to do a search for scripts that ends with those empty psxy -T calls.  I think we should eliminate those since they all add an unnecessary process call that just slows ctest down.  RIp out and take out the previous -K.

PR #4298, opened 2020-10-07
save the PS file including the path

MarceloBanik:
save the PS file including the path
see https://forum.generic-mapping-tools.org/t/postscript-file-creation-on-gmt6

Fixes #

**Reminders**

### DIFF
--- a/src/psconvert.c
+++ b/src/psconvert.c
@@ -2424,7 +2424,9 @@ EXTERN_MSC int GMT_psconvert (void *V_API, int mode, void *args) {
 
 		if (GMT->current.setting.run_mode == GMT_MODERN) {
 			if (Ctrl->T.ps) {	/* Under modern mode we can also save the PS file by renaming it */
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Looks great, thanks.  Just a GMT coding style change:  Since out_file is an array, we would typically write out_file[0] = '\0'; instead.

src/psconvert.c
```diff
@@ -2424,7 +2424,9 @@ EXTERN_MSC int GMT_psconvert (void *V_API, int mode, void *args) {
 
 		if (GMT->current.setting.run_mode == GMT_MODERN) {
 			if (Ctrl->T.ps) {	/* Under modern mode we can also save the PS file by renaming it */
-				strncpy (out_file, Ctrl->F.file, PATH_MAX-1);
+			        *out_file = '\0'; /* truncate string to build new output file */
```
seisman, line 2427: As suggested by @PaulWessel:
```suggestion
			        out_file[0] = '\0'; /* truncate string to build new output file */
```

PR #4300, opened 2020-10-08
meca: Fix plotting of beachballs whose nodal lines cross the center

### LINKED ISSUE #2397: psmeca plots wrong beachball for a moment tensor with large isotropic component
For a long time, psmeca plotted incorrect beachballs for focal mechanism with large isotropic component. The issue was fixed in #882, with more comprehensive tests provided in #661. Currently, psmeca should work for most moment tensors.
 
Here is a failing case, originally reported in http://gmt.soest.hawaii.edu/issues/1229.

A minimal example to reproduce the issue:
```
gmt meca -JX6c/6c -R123/165/-60/-15 -Sm1c -Baf -pdf meca << END
146.203 -38.2828 10.2 -0.0000 -2.2350 -0.5587 -1.2374 -0.3892 -0.5304 24 151.203 -37.2828 xyz
END
```

The output is:
image

I haven't figured out why it fails. Just open an issue in case I forget it.

### LINKED ISSUE #4225: gmt meca -Sm throws error if Mrr=0

When Mrr=0 (as is the case in some solutions in the Harvard CMT catalogue), gmt meca -Sm breaks and doesn't plot the focal mechanism. I get an error that "Assertion failed: (n < 2)" in ../src/seis/utilmeca.c, line 714. This is a problem in GMT >= 6.1.0, but not a problem in GMT 6.0.0. If I change Mrr to 0.01 then the mechanism plots fine. I don't think this is a problem for any other of the components of the moment tensor.

```
gmt meca -Sm0.5c -JM5c -R20/30/35/45  -Gred -W1 -T0 -pdf test1 << EOF
25.06 38.38 15 -0.00 1.08 -1.08 -0.75 -0.07 -0.46 24 X Y
EOF
```

**Full error message**

```
Assertion failed: (n < 2), function meca_ps_tensor, file ../src/seis/utilmeca.c, line 714.
Abort trap: 6
```

**Actual outcome**

nothing

**Expected outcome**

image

Cheers
Tom

**System information**

* Operating system: macOS Mojave 10.14.1

seisman:

To plot beachballs, we first need to determine positions of nodal lines in the Earth coordinate systems. The nodal lines may be split into at most 3 segments, due to the existence of gaps in azimuth.

The figure below uses the focal mechanism reported in #4225. The segments of the nodal line are coded by different colors. The old codes incorrectly find a gap between the last red point and the first blue point, and then break the line into two segments. The issue only happens when the nodal line crosses the center of the beachball. The new codes fix the problem by checking that the takeoff angle far away from the beachball boundary.

| Old | New |
|---|---|
|image| image|

This PR fixes #2397 and fixes #4225. I add a new test. All tests still pass.

### DIFF
--- a/src/seis/utilmeca.c
+++ b/src/seis/utilmeca.c
@@ -709,7 +709,13 @@ double meca_ps_tensor (struct GMT_CTRL *GMT, struct PSL_CTRL *PSL, double x0, do
 			x1[i] = xc; y1[i] = yc;
 		}
 		else {
-			if (fabs (fabs (az - azp) - M_PI) < D2R * 10.) {
+			if (fabs (fabs (az - azp) - M_PI) < D2R * 10.0 && takeoff > 80.0 * D2R) {
+				/* Find a gap, so start a new segment. A gap must meet the following two criteria:
+				 * 1. Current azimuth diffs from previous azimuth by ~180 degree
+				 * 2. Takeoff angle is ~90 degree so that the data fall at the beachball boundary
+				 *
+				 * Note: the choices of 10.0 and 80.0 are arbitary and may NOT working for extreme cases.
+				 */
 				azi[n][1] = azp;
 				assert (n < 2);
 				azi[++n][0] = az;
--- a/test/seis/meca_unusual_beachball.ps
+++ b/test/seis/meca_unusual_beachball.ps
@@ -1,0 +1,1144 @@
+%!PS-Adobe-3.0
+%%BoundingBox: 0 0 612 792
+%%HiResBoundingBox: 0 0 612.0000 792.0000             
+%%Title: GMT v6.2.0_ae02510_2020.10.08 [64-bit] Document from psmeca
+%%Creator: GMT6
+%%For: unknown
+%%DocumentNeededResources: font Helvetica
+%%CreationDate: Thu Oct  8 12:12:53 2020
+%%LanguageLevel: 2
+%%DocumentData: Clean7Bit
+%%Orientation: Landscape
+%%Pages: 1
+%%EndComments
+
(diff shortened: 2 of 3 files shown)


src/seis/utilmeca.c
```diff
@@ -709,7 +709,13 @@ double meca_ps_tensor (struct GMT_CTRL *GMT, struct PSL_CTRL *PSL, double x0, do
...
-			if (fabs (fabs (az - azp) - M_PI) < D2R * 10.) {
+			if (fabs (fabs (az - azp) - M_PI) < D2R * 10.0 && takeoff > 80.0 * D2R) {
+				/* Find a gap, so start a new segment. A gap must meet the following two criteria:
+				 * 1. Current azimuth diffs from previous azimuth by ~180 degree
+				 * 2. Takeoff angle is ~90 degree so that the data fall at the beachball boundary
+				 *
+				 * Note: the choices of 10.0 and 80.0 are arbitary and may NOT working for extreme cases.
+				 */
 				azi[n][1] = azp;
 				assert (n < 2);
```
seisman, line 720: A non-related issue is that `assert (n < 2);` may not work on macOS. When n=2, the `assert` statement does nothing on macOS.

joa-quim (approved):
Tee ``assert`` is C function. It should work everywhere. Now, and I don't remember which, it's not activated in debug builds (or the other way around).

seisman:
> Tee `assert` is C function. It should work everywhere. Now, and I don't remember which, it's not activated in debug builds (or the other way around).

Yes, you're right. `assert` works for **debug** build but **NOT** for **release** build. I believe this is not what we want, right?

Issue #4303, opened 2020-10-08
colorbar triangles are unclosed


seisman:
Here is a screenshot of `doc/scripts/GMT_App_M_2.ps` after I changed the foreground color to yellow: 

![image](https://user-images.githubusercontent.com/3974108/95515746-1c4fdf00-098c-11eb-99c8-77ab32a32433.png)

Closed triangles may look better.

PaulWessel:
So I immediately implemented the "fix" of duplicating the first coordinates and plotting a closed 4-point triangle instead.  But then I learned why I had not done that before: 

![Untitled](https://user-images.githubusercontent.com/26473567/95521152-97050400-0964-11eb-8457-594ab5cdcd43.png)

Because of the beveled joints it now sticks outside the width of the whole bar.  The solution is to fiddle with the geometry and adjust the two x-coordinates so that it fits perfectly.  This depends on angles and pen widths.  I think that is the best solution, no?

seisman:
> The solution is to fiddle with the geometry and adjust the two x-coordinates so that it fits perfectly.

Yes if it's not too much work.

PaulWessel:
I think we only need to fiddle with -L<gap> is used, right? The regular setup works as intended and it does shrink its coordinates to fit perfectly.  FOr the one with the gap we do want to close the triangle and change the coordinates further so that it does not stick outside.

seisman:
> I think we only need to fiddle with -L is used, right? The regular setup works as intended and it does shrink its coordinates to fit perfectly. FOr the one with the gap we do want to close the triangle and change the coordinates further so that it does not stick outside.

Yes

PR #4305, opened 2020-10-08
Make psscale -D+e triangles that align with the bat width even if gap is selected

PaulWessel:

See #4303 for context.  This PR adjusts the y-coordinates of the triangle base so that the mitered ends are exactly aligned with the width of the rest of the rectangular bar.  Only affects GMT_App_M_2.sh I think.


seisman:
Looks great to me. Update original PS files?

seisman (approved):

PR #4315, opened 2020-10-10
Improve CMake scripts if sphinx-build is not found

seisman:
sphinx-build is required for building documentations. If sphinx-build is
not found, there is no need to make the verbatims and convert PS to PNG
formats. Thus I add the `if (SPHINX_FOUND)` test to make sure we don't 
try to define some broken targets when if sphinx-build is not available.

In the old CMake scripts, if sphinx-build is not found, cmake gives the
following error:
```
ninja: error: 'doc/examples/_docs_rst_mkdir_verbatim', needed by 'doc/rst/source/_verbatim/anim01.txt', missing and no known rule to make it
```
The error message means that CMake wants to create the verbatim
`anim01.txt`, but the directory `doc/rst/source/_verbatim` (i.e., the
target `doc/examples/_docs_rst_mkdir_verbatim`) is not found. The error
message is very confusing.

This RP should fix the issue. If sphinx-build is not found, it should
give the warning like `Target docs_html is not defined.`

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -32,6 +32,7 @@ install (FILES
 	COMPONENT Runtime)
 
 # reST documentation
+find_package (Sphinx)
 add_subdirectory (rst)
 set (RST_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/rst/source)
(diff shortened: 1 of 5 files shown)


PaulWessel (approved):
Looks good!

PR #4330, opened 2020-10-15
Move deprecation messages up to INFORMATION

joa-quim:
Julia generated code raises this options
```
        histogram @v3206_06.txt  -R-6000/0/0/30 -Bpy+u" %" -Bpy+lFrequency -BWSne+glightblue+tHistograms -Bpx+l"Topography (m)" -Gorange -W1 -T250 -Z1
histogram [WARNING]: Axis sub-item f set more than once (typo?)
histogram [WARNING]: Axis sub-item a set more than once (typo?)
histogram [WARNING]: Cannot tell if -T250 -W1 is new or deprecated syntax; selected new.
```
The first two are due to a separate ``-Bpy+u" %" -Bpy+lFrequency ``, very hard, to impossible, to join the the julia parsing function. And note that there are any *a* or *f* set, so the message is not correct anyway.

The third warning is due to a missing (but not mandatory) unit in **-W1**. Also this is not controlable since users are not obliged to specify a unit.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -3580,7 +3580,7 @@ GMT_LOCAL int gmtinit_set_titem (struct GMT_CTRL *GMT, struct GMT_PLOT_AXIS *A, 
 
 	if (phase != 0.0) A->phase = phase;	/* phase must apply to entire axis */
 	if (I->active) {
-		GMT_Report (GMT->parent, GMT_MSG_WARNING, "Axis sub-item %c set more than once (typo?)\n", flag);
+		GMT_Report (GMT->parent, GMT_MSG_INFORMATION, "Axis sub-item %c set more than once (typo?)\n", flag);
 		return (GMT_NOERROR);
 	}
 	if (unit == 'c' || unit == 'C') {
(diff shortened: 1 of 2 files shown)


PaulWessel:
I agree the -B tests are not accurate so should not be warnings.  Not so sure about -W.  Giving -W without units looks weird to me and reminiscent of GMT 3 where pen was in 1/300 of an inch.  Can't you slap a 'p' if there is no other letter there at the end?

joa-quim:
With lots of work, maybe in one case or two but hardly in the general case. It's a no-so-simple parsing function. 
https://github.com/GenericMappingTools/GMT.jl/blob/master/src/common_options.jl#L971
And the default line width is point, so **-W1** is perfectly legal.

PaulWessel (approved):

PaulWessel:
Second thought on this, @joa-quim.  I think these warnings are helpful if we are running the command line interface as old GMT 4 scripts are likely.  However, if API->external is true then we can simply skip them.  I think that is a better solution, no?

joa-quim:
I don't like so much the idea of polluting perfectly regular GMT6 syntax because of possible use of a GMT4 script, but ok.

PR #4332, opened 2020-10-15
Fix issue in grdimage for constant intensity

PaulWessel:
Sloppy implementation not checking if pointer to intensity header existed before assigning things.  Since we had no test for constant intensity I added one as well.


seisman (approved):

PaulWessel:
Ignore CI errors again?

seisman:
Yes.

PR #4345, opened 2020-10-17
Prevent signed/unsigned warnings by improving code

PaulWessel:
Mostly using the same declaration of variables in comparisons, plus adding casts to macros that for OpenMP reasons need to be signed int regardless of original variable.

Also removed an unused function argument in _gmt_getscale_.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -6717,7 +6717,7 @@ int gmtlib_validate_id (struct GMTAPI_CTRL *API, int family, int object_ID, int 
 	for (i = 0, item = GMT_NOTSET; item == GMT_NOTSET && i < API->n_objects; i++) {
 		S_obj = API->object[i];	/* Shorthand only */
 		if (!S_obj) continue;								/* Empty object */
-		if (direction != GMT_NOTSET && S_obj->direction != direction) continue;		/* Not the same direction */
+		if (direction != GMT_NOTSET && (int)S_obj->direction != direction) continue;		/* Not the same direction */
 		if (direction == GMT_IN && S_obj->status != GMT_IS_UNUSED && object_ID == GMT_NOTSET) continue;		/* Already used this input object once */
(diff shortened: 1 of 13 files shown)


PaulWessel:
Could anyone approve this one for me, please?

PaulWessel:
CI fails: Why is Linux saying this:

```
../src/grdmix.c:514:5: note: in expansion of macro â€˜gmt_M_grd_loopâ€™
2020-10-17T22:12:01.3461422Z      gmt_M_grd_loop (GMT, G, row, col, node)
2020-10-17T22:12:01.3461963Z      ^~~~~~~~~~~~~~
2020-10-17T22:12:01.3462573Z ../src/gmt_grd.h:163:47: error: invalid controlling predicate
2020-10-17T22:12:01.3463538Z  #define gmt_M_row_loop(C,G,row) for (row = 0; (int)row < (int)G->header->n_rows; row++)
```

joa-quim:
Incomprehensible error message but the windows build fails too.
```
..\src\grdmix.c(512): error C3017: termination test in OpenMP 'for' statement has improper form

```

PaulWessel:
OK, so the OPenMP parser does not like

`for (col = 0; (int)col < (int)columns; col++)`

which I changed from

`for (col = 0; col < (int)columns; col++)`

Wonderful.  And the whole reason we were doing this in the first place is that the old Windows OpenMP implementation cannot handle unsigned loop variables.  So we are fucked and will have to live with these warnings, making it hard to see real warnings.

joa-quim:
Linux was complaining too and unsigned ints have been a warnings pain for a long time, not OMP only.

joa-quim (approved):

seisman:
We can disable the sign comparison warnings by adding the `-Wno-sign-compare` compiler flag. That's what I'm doing locally.

PaulWessel:
True, but having it on found at least one place where I used unsigned int to store a value from a function that returned -1...  So good to have it all show.

PR #4346, opened 2020-10-18
Better definition and use of file modifiers

PaulWessel:
List them specifically in gmt_constants.h and use the macros instead of hard-coded strings in several places:

```
#define GMT_GRIDFILE_MODIFIERS "onsuU"
#define GMT_CPTFILE_MODIFIERS "hiuU"
```

Also, when all possible file modifiers are needed (say for _gmt_access_), make a unique string on the run from the established constants.


joa-quim (approved):

PaulWessel:
Weird, I am sure I removed the WIP yesterday but there it was.  Now gone.

Issue #4344, opened 2020-10-17
grdimage and pscoast crashes when plotting with stereographic projection


anbj:
When doing ``gmt grdimage @earth_relief_30m -JS0/90/20c -P`` this happens:

```
$ gmt grdimage @earth_relief_30m -JS0/90/20c -P
grdimage [ERROR]: South boundary cannot be -90.0 for north polar stereographic projection
grdimage [ERROR]: Internal Failure = error
ERROR: Caught signal number 11 (Segmentation fault) at
/lib/x86_64-linux-gnu/libc.so.6(+0x15c181)[0x7f946382e181]
[0x0]
Stack backtrace:
/usr/local/lib/libgmt.so.6(sig_handler+0x26e)[0x7f9466161eee]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12730)[0x7f9465c04730]
/lib/x86_64-linux-gnu/libc.so.6(+0x15c181)[0x7f946382e181]
/lib/x86_64-linux-gnu/libc.so.6(+0x4f021)[0x7f9463721021]
/lib/x86_64-linux-gnu/libc.so.6(_IO_vfprintf+0x576)[0x7f9463721f66]
/lib/x86_64-linux-gnu/libc.so.6(vsnprintf+0x90)[0x7f946374a7a0]
/usr/local/lib/libgmt.so.6(gmtlib_report_func+0x125)[0x7f946627baf5]
/usr/local/lib/libgmt.so.6(gmt_err_func+0x45)[0x7f946621c9c5]
/usr/local/lib/libgmt.so.6(GMT_grdimage+0x1321)[0x7f94663cbee1]
/usr/local/lib/libgmt.so.6(GMT_Call_Module+0x3ee)[0x7f946617f41e]
gmt(main+0x260)[0x55df6c79e420]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb)[0x7f94636f609b]
gmt(_start+0x2a)[0x55df6c79f49a]
```

Looks like it's related to the implied ``-R`` (=global, including the south pole?).

Likewise with ``pscoast``:

```
$ gmt pscoast -JS0/90/20c -W -R-180/180/-90/90
pscoast [ERROR]: South boundary cannot be -90.0 for north polar stereographic projection
pscoast [ERROR]: Internal Failure = error
ERROR: Caught signal number 11 (Segmentation fault) at
[...]

PR #4347, opened 2020-10-19
Better handling of error conditions due to map_setup call

PaulWessel:
This PR is addressing the issues raised in #4344. We had a convoluted set of error checks and reporting that gave both correct error messages and confusing wrong references to other things due to a mix of internal and "official" error codes.  This PR changes this to a simpler reporting.  I was unable to reproduce the SEGV in the OP examples but I believe it was related to bad array indices into the gmt_error_string[] array.  This hopefully goes away - if not we will follow up.


anbj (approved):
Prints error message and quits, and segv is gone - perfect.

seisman (approved):

seisman:
```
$ gmt info file-do-not-exist
gmtinfo [ERROR]: Cannot find file file-do-not-exist
```

The above command returns error code 71 before this PR, but now it returns 72. Is it expected?

PaulWessel:
Impressed you noticed!  I added one more error code and I added it alphabetically so the numbers did change.  Since we do not even publish those values and what they are, does it matter?

seisman:
A PyGMT test fails due to the change of the error code. It doesn't matter. It's a bad idea to check those non-zero status codes. Will fix it in PyGMT.

joa-quim:
I notice the same in a Julia test. Had to fix the error number as I can't break the error message into a non-numeric string.

PaulWessel:
Would you prefer that I move that new error constant to the end of the list instead of having it sorted by name?  I think though as long as we are not documenting what these constants are as part of the API then they are basically not defined for users to use.

joa-quim:
Better not care about this. I'll have to find a way of not reporting the error number, which doesn't inform much anyway.

Issue #4341, opened 2020-10-17
grdimage. Weird warning message


Esteban82:
When I run the commands below I get the message:

_grdimage (WARNING): Modifier +E detected but not a valid modifier! - ignored_

```
	SET	CUT="FFT+E+n.nc"
	gmt grdmath -R0/256/0/256 -I1 -r 0 1 NRAND = %CUT%
	gmt grdimage -R -JX8c %CUT% >> out.ps
```

grdimage try to interpret the the "+E+n" from the name of the file. If I remove it from the name, the warning message do not appear. Is it a normal behaviour that the module try to interpret the name of the file? Should I be carefull with the names that I choose?

Terminal screen capture with debug information: 
![grdimage_+E+n](https://user-images.githubusercontent.com/17908951/96327689-2ddf5980-1012-11eb-9b31-67123b167933.png)

* Operating system: Win64
* GMT version: `gmt 6.1.1`

PaulWessel:
This is a good point.  GMT do have certain modifiers used in grid file names (+n+s+o comes to mind, as do +u), but the parsing is not robust enough.  So we probably should warn about using what looks like modifiers in actual file names.

PaulWessel:
We possibly do not want that message to be a WARNING as well.

PaulWessel:
Also, the parser should be improved to not worry about "modifiers" until after the file extension.  E.g.

FFT+E+n.nc should give no messages as no modifiers are found after ".nc", while
FFT+E.nc+n should only find one modifier (+n) since appended to the file.

I will make an issue out of this.

PaulWessel:
Should be fixed, let me know if you catch other cases.

Esteban82:
I got the same warning message with: 

`	gmt grdimage -R -J -O -K $DEM -Crainbow >> $OUT -Itemp_EJ7.6_Mapa_DEM+Grav_AireLibre_shadow.nc`

Warning message: 
`grdimage [WARNING]: Modifier +G detected but not a valid modifier! - ignored`

joa-quim:
If we can't use divisions in filenames why should we be allowed to use additions? 😄

PaulWessel:
Multiplications are also not allowed.  On macs, colons (:) are not allowed.

PR #4354, opened 2020-10-22
Strengthen parsing of grid and CPT filename modifiers

PaulWessel:

Both grids and CPT files can take optional modifiers, such as specifying scaling and offsets (for grids) or setting a CPT hinge (for CPT files).  However, the detection of these modifiers in the filename-string was not very robust, and we tended to use more general-purpose modifier check-functions developed for other GMT (non-file) options.  While proper modifiers were always parsed correctly, the lax implementation resulted in several problems:

1. If the user's filename happened to contain a substring that looked like a possible modifier (e.g., junk+E44.grd) we would get inappropriate complaints about bad modifiers.  See #4341 for an example.
2. Worse, if a filename contained such a substring that actually matched a valid modifier, but it was never meant to and occured before the filename extension, then we _still_ parsed it, chopped of the rest, and then could not find the file.
3. Finally, the code had many places where ad-hoc fixes looking for specific modifiers were hard-wired in (e.g., strstr (file, "+h")) that make code maintenance difficult.
4. Our documentation on grids and CPT were unclear on how any of this worked.

This PR addresses all those issues, adds documentation to clarify that users's CPT files **must** end with extension .cpt (and hence we only check for modifiers after the extension) while for grid files we **recommend** the use of extensions .grd or .nc (likewise only process modifiers after any of those), but otherwise it is the user's responsibility to understand that gridfiles like my.data+o44 will have z offset by 44 even if the user meant it to stand for something else..

### DIFF
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -2114,7 +2114,7 @@ packed integer grids (=\ *ID*\ **+s**\ *a* is a shorthand for
 if you tend to use what *could* look like modifier-sequences to GMT (e.g., using
 filenames like data.grid+o4) you can prevent any confusion by using either the
 GMT-recommended ".grd" or ".nc" as grid file extensions (e.g., data.my+o4.grd).
-Since valid modifiers are *appended* to a file name,finding such an extension simplifies parsing.
+Since valid modifiers are *appended* to a file name, finding such an extension simplifies parsing.
 
 Note that the GMT netCDF and native binary grids store the grid scale and offset
 in the file, hence if you specify these attributes when writing a file then upon reading the grid
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -1025,6 +1025,12 @@ modifiers to your CPT names when used in GMT commands.  The **+u**\ *unit*
 modifier will scale z *from unit to* meters, while **+U**\ *unit* does
 the inverse (scale z *from meters to unit*).
 
+**Note**: Users are allowed to name their CPT files anything they want, but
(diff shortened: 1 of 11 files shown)


src/gmt_constants.h
```diff
@@ -274,8 +274,8 @@ enum GMT_enum_basemap {
 #define GMT_GRIDFILE_MODIFIERS "onsuU"
 
 /* Modifiers for CPT files:
- * +h<hinge> to override soft-hinge value in CPT
- * +i<dz> is used to round auto-deteremined min/amax range to a multiple of dz.
+ * +[h<hinge>] to override soft-hinge value in CPT
```
seisman, line 277:
```suggestion
 * +h[<hinge>] to override soft-hinge value in CPT
```

doc/rst/source/cookbook/features.rst
```diff
@@ -2104,6 +2110,12 @@ You may the scale as *a* for auto-adjusting the scale and/or offset of
 packed integer grids (=\ *ID*\ **+s**\ *a* is a shorthand for
 =\ *ID*\ **+s**\ *a*\ **+o**\ *a*).
 
+**Note**: Users are allowed to name their grid files anything they want.  However,
+if you tend to use what *could* look like modifier-sequences to GMT (e.g., using
+filenames like data.grid+o4) you can prevent any confusion by using either the
+GMT-recommended ".grd" or ".nc" as grid file extensions (e.g., data.my+o4.grd).
+Since valid modifiers are *appended* to a file name,finding such an extension simplifies parsing.
```
seisman, line 2117:
```suggestion
Since valid modifiers are *appended* to a file name, finding such an extension simplifies parsing.
```

seisman (approved):

PR #4370, opened 2020-10-24
Fix duplicate array values for most modules

### LINKED ISSUE #4368: histogram doesn't give correct region if list of bins is given
Example input data: [test.txt](https://github.com/GenericMappingTools/gmt/files/5433548/test.txt)

Using `-T-3/7/1`, the automatic x range is from -3 to 7, determined by the **-T** option. It looks reasonable.
```
gmt histogram test.txt -T-3/7/1 -Gblue -Bxa1f -Byaf -png map
```

image

Instead of using `-T-3/7/1`, the following command gives a comma-separated list of coordinates. In this case, the automatic x range is from -1 to 6. It seems the x range is determined from the input data, not the **-T** option.
```
gmt histogram test.txt -T-3,-2,-2,-1,-1,1,1,2,2,3,3,4,4,5,5,6,6,7 -Gblue -Bxa1f -Byaf -png map
```
image

PaulWessel:
See #4368 for context.  If an array is given via a comma-separated list or a file, for most modules that acquire such an array it is required that it be monotonically increasing and not have duplicated values.  This we never checked for. So most of the PR involves adding a new flag that forces sorting and removal of duplicates and then passing that flag to most places where _gmt_parse_array_ is called.  The only exception I could find is **gmtmath** which is fine with a list of numbers.  A minor fix in histogram had to do with updating the wesn for x after parsing the list. Closes #4368.

### DIFF
--- a/src/filter1d.c
+++ b/src/filter1d.c
@@ -394,7 +394,7 @@ static int parse (struct GMT_CTRL *GMT, struct FILTER1D_CTRL *Ctrl, struct GMT_O
 	}
 
 	if (Ctrl->T.active)	/* Do this one here since we need Ctrl->N.col to be set first, if selected */
-		n_errors += gmt_parse_array (GMT, 'T', t_arg, &(Ctrl->T.T), GMT_ARRAY_TIME | GMT_ARRAY_DIST | GMT_ARRAY_ROUND, Ctrl->N.col);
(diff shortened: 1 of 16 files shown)


seisman (approved):
I confirm that this PR fixes issue #4368, but I didn't run the full tests.

PaulWessel:
I ran the full set and nothing new there.

PaulWessel:
What is the issue with home-brew and is there anyone working to fix whatever it is that breaks the CI?

seisman:
It looks like an issue of Azure Pipelines, but I don't see any issue reports or fixes from either homebrew or Azure Pipelines.

PR #4380, opened 2020-10-26
Let mapproject report correct data format if -V is given

PaulWessel:
**mapproject** made assumptions about the data columns and would try to print verbose min/max messages using format %g.  However, that will fail if absolute time is one of the coordinates (we get some large integer UNIX seconds):

`mapproject [INFORMATION]: Input extreme values: Xmin: 463634246 Xmax: 463634249 Ymin: -0.00285797172888 Ymax 0.00044`

Now, we use _gmt_ascii_format_one_ and the correct type columns to format strings individually, to get the correct result:

`mapproject [INFORMATION]: Input extreme values: Xmin: 1984-09-10T03:17:26 Xmax: 1984-09-10T03:17:29 Ymin: -0.00285797172888 Ymax 0.00044`

Also, I fixed a message that said lon-lat when data were not geographic.  The initial transform message also was fixed:

`mapproject [INFORMATION]: Transform 1984-09-10T03:15:00/1984-09-10T03:45:00/-15/15 -> 0/20/0/10 [cm]`

and it works the same if **-I** is used:

```
mapproject [INFORMATION]: Input extreme values:  Xmin: 1.62222222229 Xmax: 1.65555555609 Ymin: 4.99904734276 Ymax 5.00014666667
mapproject [INFORMATION]: Output extreme values: Xmin: 1984-09-10T03:17:26 Xmax: 1984-09-10T03:17:29 Ymin: -0.00285797172 Ymax 0.000440000009998
```

### DIFF
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -785,7 +785,7 @@ EXTERN_MSC int GMT_mapproject (void *V_API, int mode, void *args) {
 	bool line_start = true, do_geo_conv = false, double_whammy = false, first = true;
 	bool geodetic_calc = false, datum_conv_only = false, along_track = false, projected = false;
 
-	unsigned int i = 0, col, speed_col = 2, first_z_col = 0;
+	unsigned int i = 0, col, speed_col = 2, first_z_col = 0, i_col_x, i_col_y, o_col_x, o_col_y;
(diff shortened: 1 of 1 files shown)


PaulWessel:
Pinging @joa-quim and @seisman for approval.

seisman (approved):

PR #4395, opened 2020-10-30
Multi-band bars in plot and plot3d

### LINKED ISSUE #4391: psxyz does not documment the -Sb...+Z|z option
That happens because both ``psxyz`` and ``psxy`` both use the ``explain_symbols.rst_`` file that does not mention it.

Also, the -Sb...+Z|z option that was introduced mainly to simplify ex10 is very handy and would be a nice to have in ``psxy``. As is one have to do a **lot** of gymnastic to produce a figure like this in plain shell GMT

```
julia> bar([0. 1 2 3; 1 2 3 4], stack=1, show=1,  fmt=:png, fill=["red" "green" "blue"])
```

GMTjl_tmp

PaulWessel:

The mechanism that allows for multi-band columns in **plot3d** has now been generalized to also apply to the 2-D bars (horizontal and vertical) in both **plot** and **plot3d** (and thus closes #4391).  This is still WIP because of a bug that fails in the _3dbars.sh_ script as well as a question tp @joa-quim and @seisman:

While **+z**|**Z** made sense for allowing more z-values in a multi-band vertical column, does it make sense for horizontal and vertical bars? Or shall we use **+x**|**X** and **+y**|**Y** for horizontal and vertical bar, respectively?


joa-quim:
I would prefer to have the same modifiers for all the x,y,z cases. It could be **+w|W**

PaulWessel:
OK, we will need to deal with +Z|z via backwards compatibility.  What does W stand for? Better with +V|v for values?

PaulWessel:
Actually, may make more sense to use +v for values and +i for increments.  Nothing mnemonic about upper case V or W.

joa-quim:
Thought on **w**idth

And, do we really need that *nbands* in **Z**nbads? It could get that count from the number of columns.

PaulWessel:
Yes, I thinks we could skip nbands. Needs a bit of rewriting.

PaulWessel:
Changed to **+v**|**i**. Will work on eliminating _nbands_ later.

PaulWessel:
There are complications in schemes to eliminate _nbands_:

1. When **-S** is given without symbol code, we expect it as the first text in the trailing text.  With different symbols, the number of columns vary per record.
2. When **-S** is given, we read data on a rec by rec basis.  To set this up, we pre-compute the number of numerical columns needed so we can call _GMT_Set_Columns_ **before** opening the file.

There are no great solutions here given other complications:

1. From externals, the concept of varying numerical columns is no good as they will want to pass a matrix or a set of vectors.  In fact, have @joa-quim or @seisman tried to call plot with the symbol code given by trailing text? Is that something we should disallow altogether for external calls (i.e., when memory data are given)?
2. We have in the past talked about yanking the rec-by-rec scheme and read a full data set (as done for lines and polygons).  But as long as **-S** can skip the symbol code, we _still_ would have to handle variable records and the only way we could do that via a data set is to only specify 2 columns (x, y) and let the rest be trailing text we then parse as needed (kind of like **pstext**).  But again, probably not what you want from externals.

So my bottom line, specifying _nbands_ may be required as long as symbol code is optional.

joa-quim:
No, I never tried to plot from symbols defined in a text column. Building a memory array of that nature doesn't seem useful and even less useful thinking that users would do so instead of loading a file.

PaulWessel:
This PR has now been completed, with any bugs fixed.  We use **+z**|**i** in both 2-D and 3-D bars/columns to indicate multi-band coloring.  The docs have been updated with updated figures showing the multi-band bars for both **plot** and **plot3d** and the multi-band column for **plot3d**.  I allow the (now undocumented) **+z**|**Z** to quietly mean **+z**|**i**.
New test scripts for both **plot** and **plot3d** showing the use of multi-band 2-D bars with or without *base* have been added as well.  I think it is good to go, @joa-quim and @seisman, please have a look.

joa-quim:
It all looks nice ... except keep using *inches* in tests

PaulWessel:
True.  And overnight I had the idea of adding a **+h** modifier which would take the bar width and divide it into vertical slices and each individual slice would have its own height.  Without **+h** we get verticals.  That would seem like a nice enhancement as well, no?

joa-quim:
You mean the horizontal bar-gropus like this?

```
julia> bar([0 1 2 3; 1 2 3 4], fmt=:png, show=1)
```
GMTjl_tmp

PaulWessel:
Yes

PaulWessel:
So each cluster is one multi-bar symbol and provided by one data record with n y-values that takes the given bar width and splits it into multiple skinnier bars in a cluster.

joa-quim:
Then sure, a must have. And make it **+h**[pct] where *pct* by default is 1.0, meaning all bar in the soub-groups touch each other. Lower values open a space between them. See https://www.mathworks.com/help/matlab/ref/bar.html

PaulWessel:
Yes, can do that.  This only applies to the 2-D flat bars, not 3-D column. So enhanced syntax will be

**-Sb**|**b**_width_[_unit_][**+b**|**B**[_base_]][**+i**|**v**_nbands_][**+h**[_gap_]]

where the new thing is the **+h** horizontal mode with an option to set aside a fraction of the bar width for _gaps_ between the bars.  I agree that some sort of fraction or percent is good since we may not know the bar width (if x-units is used, for instance as it depends on projection).  Question is shall it be a percentage (0-100%) or a fraction (0-1)?

joa-quim:
Fraction, like in Matlab https://www.mathworks.com/help/matlab/ref/bar.html#bthf_ju-width

PaulWessel:
Yes I thought of that too, a gap < 1 is clearly a fraction and not percent.

PaulWessel:
With bars being either vertical (**-Sb**) or horizontal (**-SB**) the **+h** is not a great mnemonic.  I hvae settled on **+s** for side-by-side.

joa-quim:
With ~1e4 options + modifiers I don't care anymore about mnemonics. That's why long-verbose is so the way.

PaulWessel:
I agree, and Meghan can start on that part pretty quickly since it is mostly deciding (with our input) and the long names and then build a few structures with the translation.

PaulWessel:
OK, @joa-quim and @seisman. I have implemented -**Sb**|**b**_width_[_unit_][**+b**|**B**[_base_]][**+i**|**v**_nbands_][**+s**[_gap_]] where **+s** stands for side-by-side individual, thinner bars (as opposed to the vertical placement on a single bar.  The *width* is then split among the _nbands_ bars after any desired _gap_ is removed.  The _gap_ is measured as a fraction (or percent) of *width* [0].  The explanatory figure in the docs now looks like this:

GMT_base_symbols4

Added two extra tests for this feature (for **psxy** and **psxyz**).  All pass, please have a look and hopefully approve.

joa-quim:
Sorry to push it still a bit further. Will autolegend know about the group colors meaning?

PaulWessel:
I doubt it since it requires a CPT as well as multi-band entries.  I won't tackle that right now.

joa-quim (approved):

PR #4400, opened 2020-10-31
Animation 13 (seismic waveforms)

PaulWessel:
**A new movie example with time-series**

This animation exercises the new **events** **-Ar**_dpi_ option implemented earlier.  The movie runs for me, as explained below.  For this draft there are some limitations and issues:

1. The data file is currently just placed in the same directory, but the plan is to place it in the cache before merging into master.
2. There is very little meta-data plotted, such as a location map, information about the quake, and scaling factor and units fo the y-axes.
3. This is a short time-series (76 seconds) so in real time it is only 76/24 or a bit over 3 seconds long.  That makes it a bit jerky given that 1s is a large step in time.  We can (a) slow it down by picking another interval (e.g., 0.1s), (b) smooth it by picking a higher number of frames per second (60?), or (c) pick a longer time-series from another earthquake.
4. Since all panels have the same annotation/ticks the **-B** stuff should go on the **subplot begin** line.  However that fails, probably because of the **-Bp** and **-Bs** options, and if so a "feature" in **subplot**.  Because of this I only plot the clock annotation and not a date, but date could probably go in a side bar with the location map.

The script has a commented out **movie** call: You first run the _anim13.sh_ script, then copy/paste the **gmt movie** command and run that afterwards.  Because of **-Z** you must run _anim13.sh_ before each **movie** command.

Here is the anim13.png:

anim13

This is the animated GIF that is built:

anim13


seisman:
> 3\. This is a short time-series (76 seconds) so in real time it is only 76/24 or a bit over 3 seconds long.

It's unclear to me what "real time" means. I thought real time means the 76-s time-series will become a 76-s movie.

PaulWessel:
Yes you are right.  THe 76 s data should play in 76 seconds.  With 24 frames per second I need to select output times using 1/24s intervals.

PaulWessel:
Will post the updated git soon.  But my experience opens up an issue for us: **movie -T**_info_ can either take

1. Number of frames (and time is just 0, 1, 2,...)
2. _min/max/inc_, and we create an equidistant list of times (this can be absolute time with **-f**0T
3. A file with time and perhaps other columns

In anim13.sh I am creating a single-column list of times with **gmt math**, and the problem is two-fold:

1. For "real-time" animation here I want 24 frames to equal 1 second, so dt = 0.0416666666667, not the best value to use in a min/max/inc setup.
2. The default **FORMAT_CLOCK_OUT** is hh:mm:ss so I get 24 identical times with the same second unless I am smart enough to add **--FORMAT_CLOCK_OUT**=hh:mm:ss.xxxxx (hint: was not).

I can solve this issue by using **-T**_min/max/inc_ but at other times one really may need to build such times from **gmt math** and then those two issues are problematic.  Some solutions:

1. The GMT_wide **-T**_min/max/inc_ machinery could add another modifier (say **+i**) to indicate that the increment should be inverted (we already have **+n** to say the increment is really the number of points).  That would let me do `2020-10-06T05:55:00/2020-10-06T05:56:15/24+i` which would avoid awkward manual divisions and lack of significant decimals.
2. Modules that accept **-T** (math, regress, filter1d, ...) could check if T is absolute time and unit is seconds, then look at the increment and if < 1 either automatically set a suitable format_clock_out format (I implemented that in the **psevents -Ar**_dpi_ scheme) or at least warn of loss of precision in the output.  I think if we can figure it out it would be better to set the right precision.

PaulWessel:
Updated movie is too big (212 Mb) now that it has lots of frames (1801) - you can run it though and see what we should enhance.  Apart from things like adding a map and meta data, we can

1. Decide if the curves should have colors (e.g., get more intense when there are larger amplitudes)
2. Modify that leading blob - maybe too big right now (-M options)
3. Add a fading out of the color some time after arrival (let the red go to darker red)

PaulWessel:
Updated to use the new 24+i for 1/24 increment and selected red, green, blue colors for the three traces.  Takes 12 minutes to render on 24 cores.  Suggestions for improvements, please.

PaulWessel:
Here is the latest MP4: [anim13.mp4.zip](https://github.com/GenericMappingTools/gmt/files/5473624/anim13.mp4.zip)

PaulWessel:
Any suggestions to make this better, @seisman or @joa-quim?

joa-quim:
It would be funny to have a counters in km since the beginning of the P waves till first S's showing the epicentral distance from station.

PaulWessel:
Made a few more changes - now the still frame looks like this:

anim13

Maybe some text legend etc in the blank spaces is good enough?

PaulWessel:
Added some text in the side-bar.  Maybe @seisman could consider a focal mechanism in the available space etc?  Think I am reaching my artistic limit on this animation.

anim13

seisman:
When I run the script, I saw an empty image poped out before looping over frames.

seisman:
The event information is also incorrect. See the USGS page for the correct information:
https://earthquake.usgs.gov/earthquakes/eventpage/us6000c5zm/executive

PaulWessel:
Yes, I get that too, so I wonder if the gmt show part for the master frame somehow comes before we finish making the PNG? Not sure how but I have to quit Preview then open up the png separately.

PaulWessel:
Please edit the metadata - I may have googled the wrong quake.

seisman:
This is the new image after merging #4444 
image

seisman:
Here are a few things that I think could be improved.

1. The positioning of the map, beachball and event information is not intuitive. I tried many times to put the beachball at the right place.
2. Zoom in the waveforms, you'll see:

image

It's not as good as I expect. Perhaps something wrong with the psevent interpolation?

PaulWessel:
Regarding positioning:

Yes, plotting the meca there is probably tricky.  Here, it acts more like a legend item.   Perhaps meca could have a way of not requiring -R -J and/or not using the lon lat?  Sorry, have go cocking so will think more about this.
THe psevent interpolation is set for the desired movie resolution so zooming in is not an option.  If you want to zoom in you need to run this as a 4k movie instead and see what it looks like at normal view (not zoom.=)

PaulWessel:
We have had a regression on that s.xxx format needed for this case - will have to find where that is coming from.  You can see it clearly if you use the -M without asking for png and you get a PDF (for the HD resolution settings).
I will also post an issue to address how to place a meca symbol in a legend.

src/psevents.c
```diff
@@ -658,9 +658,10 @@ EXTERN_MSC int GMT_psevents (void *V_API, int mode, void *args) {
 			if (bad) {
 				GMT_Report (API, GMT_MSG_WARNING, "Interpolation  of time columne with -F%c failed to give monotonically increasing values in %d places. Please use --GMT_INTERPOLANT=linear instead\n", Fmode, bad);
```
seisman, line 659:
```suggestion
				GMT_Report (API, GMT_MSG_WARNING, "Interpolation of time columne with -F%c failed to give monotonically increasing values in %d places. Please use --GMT_INTERPOLANT=linear instead\n", Fmode, bad);
```

doc/examples/anim13/waveform_AV.DOL.txt
```diff
@@ -0,0 +1,3753 @@
+# Waveform for earthquake on 2020-10-06 05:54:50 (UTC) at 54.844°N 159.860°W at 30.4 km depth with M 5.9 - 68 km SE of Sand Point, Alaska
```
seisman, line 1:
```suggestion
# Waveform for earthquake on 2020-10-06 05:54:50 (UTC) at 54.844°N 159.860°W at 30.4 km depth with M 5.9 - 68 km SE of Sand Point, Alaska
# Recorded by station AV.DOL at 55.15°N, 161.86°W.
```

doc/examples/anim13/anim13.sh
```diff
@@ -0,0 +1,69 @@
...
+DPI=200i
+SPLINE=linear
+EOF
+# 1. Create a dense point-file from the waveform data
+cat << EOF > pre.sh
+gmt begin
+	# 1a. Create dense point-files from the waveform data components
+	gmt psevents waveform_AV.DOL.txt -R\${PLOT_DOMAIN} -JX20cT/3.5c -Ar\${DPI} -i0,1+s1e-6,0 -f2T --GMT_INTERPOLANT=\${SPLINE} > E.txt
+	gmt psevents waveform_AV.DOL.txt -R\${PLOT_DOMAIN} -JX20cT/3.5c -Ar\${DPI} -i0,2+s1e-6,0 -f2T --GMT_INTERPOLANT=\${SPLINE} > N.txt
+	gmt psevents waveform_AV.DOL.txt -R\${PLOT_DOMAIN} -JX20cT/3.5c -Ar\${DPI} -i0,3+s1e-6,0 -f2T --GMT_INTERPOLANT=\${SPLINE} > Z.txt
```
seisman, line 30:
```suggestion
	gmt events waveform_AV.DOL.txt -R\${PLOT_DOMAIN} -JX20cT/3.5c -Ar\${DPI} -i0,1+s1e-6,0 -f2T --GMT_INTERPOLANT=\${SPLINE} > E.txt
	gmt events waveform_AV.DOL.txt -R\${PLOT_DOMAIN} -JX20cT/3.5c -Ar\${DPI} -i0,2+s1e-6,0 -f2T --GMT_INTERPOLANT=\${SPLINE} > N.txt
	gmt events waveform_AV.DOL.txt -R\${PLOT_DOMAIN} -JX20cT/3.5c -Ar\${DPI} -i0,3+s1e-6,0 -f2T --GMT_INTERPOLANT=\${SPLINE} > Z.txt
```

doc/examples/anim13/anim13.sh
```diff
@@ -0,0 +1,69 @@
...
+			printf "%s -2.6\n%s 2.8\n" \${MOVIE_COL0} \${MOVIE_COL0} | gmt plot -W0.5p
+			gmt subplot set 1 -A"N"
+			gmt plot waveform_AV.DOL.txt -Bpxafg -Bsxa1D -Bpyaf -BWSrt -W0.5p,darkgray -i0,2+s1e-6
+			gmt events N.txt -Sc0.5p -Ggreen -Es+d0.5+r0.5 -Ms4+c1 -Mi0.5 -Mt+c0 -T\${MOVIE_COL0}
+			printf "%s -2.6\n%s 2.8\n" \${MOVIE_COL0} \${MOVIE_COL0} | gmt plot -W0.5p
+			gmt subplot set 2 -A"E"
+			gmt plot waveform_AV.DOL.txt -Bpxafg -Bsxa1D -Bpyaf -BWSrt -W0.5p,darkgray -i0,1+s1e-6
+			gmt events E.txt -Sc0.5p -Gblue -Es+d0.5+r0.5 -Ms4+c1 -Mi0.5 -Mt+c0 -T\${MOVIE_COL0}
+			printf "%s -2.6\n%s 2.8\n" \${MOVIE_COL0} \${MOVIE_COL0} | gmt plot -W0.5p
+		gmt subplot end
```
seisman, line 65: No need to indent the whole "gmt subplot" block.

PaulWessel: No need to do any indention (not Python) but seems OK to keep that separate.
BTW, I removed the ss.xxxxx I had on the movie call and now I get whole seconds so I need to add the same fix I did in events to movie.  Maybe tomorrow before this is all resolved.

seisman:
Usually it's useful to also show the station location on map, but for this example, the event and station are too close.
image

doc/examples/anim13/anim13.sh
seisman: > No need to do any indention (not Python) but seems OK to keep that separate.

Yes, my above comment is a little misleading. What I mean is the indentation of Line 52 (`gmt set`) and the "gmt subplot" block. To me, it's more readable that "gmt set" and "gmt subplot begin" are in the same indentation level.

This is the current indentation style:
```
gmt begin 
    gmt set 
        gmt subplot begin
            gmt subplot set 0
            ...
        gmt subplot end
gmt end 
```

What I prefer to is:
```
gmt begin 
    gmt set 
    gmt subplot begin
        gmt subplot set 0
        ...
    gmt subplot end
gmt end 
```

PaulWessel: Yes, my bad.  I will fix since doing the other thing as well.

PaulWessel:
Should we try to center the beachball in the sidebar?  Could be related to #4445 though.

PaulWessel:
Hi @seisman, we should get this one on the road.  Let's just punt on any fancy mechanism to simplify placement of the beachball in the legend area for now and instead add this to our set of animations, as it shows animations of lines (via the new -A option to events).  Shall we wait to post the MP4 on the YouTube channel until 6.2 is released?  Or post now and let the release docs update when that time comes but let nightly-build docs have the full info? I think the latter, no?

anim13

Esteban82 (approved):

meghanrjones:
I think it would be better to add waveform_AV.DOL.txt as a remote file rather than including it in the git repo. That way anyone can make the script by copying the text without needing to worry about the data.

PaulWessel:
We could place a title in that empty space left of the clock.  Any suggestions for a title?

seisman:
Can we put the station information in the title, and also put a black triangle on the map?

PaulWessel:
OK, I placed the station name on top and plotted both the red quake and the black triangular station on the map.  Looks OK?

anim13

seisman:
I tried to run the script locally, and it failed because the data file was still not available in the cache. Then I manually downloaded the file and ran it again:

```
movie [INFORMATION]: FFmpeg 4.3.1 found.
movie [INFORMATION]: Paper dimensions: Width = 24c Height = 13.5c
movie [INFORMATION]: Pixel dimensions: 1920 x 1080
movie [INFORMATION]: Building opaque PNG images.
movie [ERROR]: Working directory /var/folders/mz/d3wtwjtx4fg8qlds5lgt1c240000gn/T/anim13 already exist and -Z was not specified - exiting
```
I understand why it fails, and after deleting the directory, it works well.

My question is, why does `movie` always use the **same** temporary directory?

PaulWessel:
That is a good point.  I guess it started before -W when -N set a local work dir based on the movie name.  It made sense to keep it the same.  I think for -W there really is no reason why it has to be called /tmpdir/moviename.  However, note that it may be full of thousands of PNG files that we do not want to delete, as the idea was that you may wish to reformat the movie from the original PNGs and since they may take hours to create we did not want to auto-delete those directories.  Hence a fixed name was simpler.  So the message you got is to protect you from accidentally deleting hours worth of work.

seisman:
Perhaps something wrong with the input times? I'm getting this movie:

https://user-images.githubusercontent.com/3974108/111682327-99fcfe00-87fa-11eb-9531-9c1273fda73f.mp4

PaulWessel:
Hm, and sure you are running this after building the branch since it has code-fixes related to time formatting.

seisman:
> Hm, and sure you are running this after building the branch since it has code-fixes related to time formatting.

Good point! I didn't realize that there are code changes in this PR. Will run it again.

PaulWessel:
Hopefully that works for you - works well here and I think we should post the movie on YouTub to get that link so I can add it to the anim13.sh script.  Then hopefully approve and merge.

PaulWessel:
OK, with the waveform file safely in the cache (which involved a detour to fix issues on the server) and a title and station nicely plotted, may we agree to

1. This is the final version of anim13.sh
2. OK to upload to YouTube now so I can get a link to place in the anim13.rst file
3. After that, get approvals and merge

Pinging @seisman and @meghanrjones on that.

doc/examples/anim13/anim13.sh
```diff
@@ -0,0 +1,71 @@
...
+# Movie of seismic waveforms from 2020-10-06 Alaska earthquake
+# Epicenter 54.844N 159.842W
+# Depth 31.0 km
+# UTC time 2020-10-06 05:54:50
+# Recorded at station Dolgoi Island, Mount Dutton, Alaska
+#
+# USGS event page: https://earthquake.usgs.gov/earthquakes/eventpage/us6000c5zm/executive
+#
+# DEM:  @earth_relief_06m
+# Data: @waveform_AV.DOL.txt
```
seisman, line 12: Need to explain the meaning of each column in the data `waveform_AV.DOL.txt`, as it's not a commonly used format in seismology community

doc/examples/anim13/anim13.sh
```diff
@@ -0,0 +1,71 @@
...
+		gmt subplot set 0 -A"Z"
+		gmt plot @waveform_AV.DOL.txt -Bpxafg -Bsxa1D -Bpyaf -BWsrt -W0.5p,darkgray -i0,3+s1e-6
+		gmt events Z.txt -Sc0.5p -Gred -Es+d0.5+r0.5 -Ms4+c1 -Mi0.5 -Mt+c0 -T\${MOVIE_COL0}
+		printf "%s -2.6\n%s 2.8\n" \${MOVIE_COL0} \${MOVIE_COL0} | gmt plot -W0.5p
+		gmt subplot set 1 -A"N"
+		gmt plot @waveform_AV.DOL.txt -Bpxafg -Bsxa1D -Bpyaf -BWsrt -W0.5p,darkgray -i0,2+s1e-6
+		gmt events N.txt -Sc0.5p -Ggreen -Es+d0.5+r0.5 -Ms4+c1 -Mi0.5 -Mt+c0 -T\${MOVIE_COL0}
+		printf "%s -2.6\n%s 2.8\n" \${MOVIE_COL0} \${MOVIE_COL0} | gmt plot -W0.5p
+		gmt subplot set 2 -A"E"
+		gmt plot @waveform_AV.DOL.txt -Bpxafg -Bsxa1D -Bpyaf -BWSrt -W0.5p,darkgray -i0,1+s1e-6
```
seisman, line 64: It seems these commands plot static elements that don't change with time. I'm wondering if it's possible to move them to `pre.sh`?
```
gmt plot @waveform_AV.DOL.txt -Bpxafg -Bsxa1D -Bpyaf -BWsrt -W0.5p,darkgray -i0,3+s1e-6
gmt plot @waveform_AV.DOL.txt -Bpxafg -Bsxa1D -Bpyaf -BWsrt -W0.5p,darkgray -i0,2+s1e-6
gmt plot @waveform_AV.DOL.txt -Bpxafg -Bsxa1D -Bpyaf -BWSrt -W0.5p,darkgray -i0,1+s1e-6
```

doc/examples/anim13/anim13.sh
PaulWessel: OK, might you be able to help so that it makes sense to other seismologists?

doc/examples/anim13/anim13.sh
PaulWessel: Heh, I spend some time last night trying that.  I could get close.  If I do a **subplot** with just the static and **-A** stuff and then another **subplot** in main just doing the animated parts then it would not line up.  The reason is that the **-B** choices affects spacing between panels.  So, I will need to plot the annotations exactly the same, twice, so they overprint, otherwise it shifts the spacing.  So possible ways of dealing this this might be:

Add a modifier or flag to **subplot** that says "use the -B and -SC -SR in computing spacing and gaps etc, then do not plot any basemap".  This would be helpful in the case like this animation.  Because it involves 1-3 separate options it may need a separate option to turn this feature on.  E.g., one of -D, -E, -L, -N, -Q. Maybe

**-D** Use the selections and settings for basemap annotations (via default settings, **-B** and **-S**) to determine panel dimensions and spacings but do _not_ draw any panel frames and annotations [draw frames and annotations as requested].

I am hoping this is an easy thing to implement (just write out no frame info to the subplot hidden files).  Seems worth trying, no?

PaulWessel:
I guess the broader question is this: Should it be possible to call subplot twice and have the two results align even if one is not drawing frames and annotations? I think ideally the answer is yes.  This test script illustrate the problem:

```
gmt begin test
	gmt subplot begin 3x1 -R0/10/0/5 -Fs10c/6c -M6p -A -SCl -SRb -Bafg -T"My Subplot"
		gmt basemap -c
		gmt basemap -c
		gmt basemap -c
	gmt subplot end
	# Not overlay data
	gmt subplot begin 3x1 -R0/10/0/5 -Fs10c/6c -M6p -B+n
		echo 5 2.5 | gmt plot -Sc2c -Gred -c
		echo 5 2.5 | gmt plot -Sc2c -Ggreen -c
		echo 5 2.5 | gmt plot -Sc2c -Gblue -c
	gmt subplot end
gmt end show
```

yielding a misaligned plot (all circles should be at 5/2.5)

test

Perhaps @joa-quim will have comments on this as well [see previous post above regarding -D].

seisman:
> I guess the broader question is this: Should it be possible to call subplot twice and have the two results align even if one is not drawing frames and annotations? I think ideally the answer is yes.

I agree.

PaulWessel:
Took just 2 minutes to add -D and test it and it solves the problem.  Users just need to use the same settings and -S -B but add -D and it aligns.  I will make a PR for this with test for master, then we can merge in to anim13 and redesign the script to plot the static stuff in pre.sh

meghanrjones:
> I guess the broader question is this: Should it be possible to call subplot twice and have the two results align even if one is not drawing frames and annotations? I think ideally the answer is yes. This test script illustrate the problem:
> 
> ```
> gmt begin test
> 	gmt subplot begin 3x1 -R0/10/0/5 -Fs10c/6c -M6p -A -SCl -SRb -Bafg -T"My Subplot"
> 		gmt basemap -c
> 		gmt basemap -c
> 		gmt basemap -c
> 	gmt subplot end
> 	# Not overlay data
> 	gmt subplot begin 3x1 -R0/10/0/5 -Fs10c/6c -M6p -B+n
> 		echo 5 2.5 | gmt plot -Sc2c -Gred -c
> 		echo 5 2.5 | gmt plot -Sc2c -Ggreen -c
> 		echo 5 2.5 | gmt plot -Sc2c -Gblue -c
> 	gmt subplot end
> gmt end show
> ```
> 
> yielding a misaligned plot (all circles should be at 5/2.5)
> 
> test
> 
> Perhaps @joa-quim will have comments on this as well [see previous post above regarding -D].

I agree that this is worthwhile functionality. Independently of subplots, I think that more support for underlays in modern mode would be nice. Something like:

```
export GMT_SESSION_NAME=underlay
gmt begin
gmt basemap -JM6i -R-30/30/-30/30 -Bafg

unset GMT_SESSION_NAME

gmt begin -Dunderlay,overlay # duplicate underlay session to overlay
gmt plot ex1.txt -S -Cchrome
gmt end show

gmt begin -Dunderlay,overlay # duplicate underlay session to overlay
gmt plot ex2.txt -S -Cchrome
gmt end show

gmt begin -Dunderlay,overlay # duplicate underlay session to overlay
gmt plot ex3.txt -S -Cchrome
gmt end show
```
Does this seem possible?

PaulWessel:
> I agree that this is worthwhile functionality. Independently of subplots, I think that more support for underlays in modern mode would be nice.

This would be a way to recycle bits of PostScript such as one can do in classic mode, i.e., make a common overlay add.ps and use it in different plots.  We can discuss this and what the technological challenges might be.  It is clearly different from what -D does in the subplot-only context.

PaulWessel:
Quick vote for @joa-quim @seisman @meghanrjones: Which makes most sense to you mnemonically:

1. -D Determine panel dimensions/sizes but do not draw/annotate them
2. -N No frames: Determine panel dimensions/sizes but do not draw/annotate them

meghanrjones:
> Quick vote for @joa-quim @seisman @meghanrjones: Which makes most sense to you mnemonically:
> 
>     1. -D Determine panel dimensions/sizes but do not draw/annotate them
> 
>     2. -N No frames: Determine panel dimensions/sizes but do not draw/annotate them

I like **-D**

PaulWessel:
Alright.  After merging in **subplot -D** I rewrote the script to do what @seisman suggested.  So now I think we are ready, no?  It did not shave more than a minute off by not plotting a few lines but it is the principle of plotting static once and variables in the frames.  Still need your OK to post the movie to YouTube so we can finalize the URL.  OK?

seisman:
OK to upload as long as the new script generates the same animation as the old one (I don't rerun the animation, but the one generated using the old script is pretty good).

PaulWessel:
I just reran after the update and it looks exactly the same.

PaulWessel:
Will upload after a forced walk for exercise.

meghanrjones:
> OK, with the waveform file safely in the cache (which involved a detour to fix issues on the server) and a title and station nicely plotted, may we agree to
> 
>     1. This is the final version of anim13.sh
> 
>     2. OK to upload to YouTube now so I can get a link to place in the anim13.rst file
> 
>     3. After that, get approvals and merge
> 
> 
> Pinging @seisman and @meghanrjones on that.

Yes, sounds good.

seisman:
Perhaps we still need to change `gmt end show` to `gmt end`, because I still see an empty image popped up when running the animation.

PaulWessel:
I also get a Preview failure that pops up, and I must close it, quit Preview then type open anim13.png and it comes up fine.  It is as if the gmt docs anim13.png under the hood happens before the file is properly closed.
But I can remove the show.

doc/examples/anim13/anim13.sh
```diff
@@ -0,0 +1,74 @@
...
+# Movie of seismic waveforms from 2020-10-06 Alaska earthquake
+# Epicenter 54.844N 159.842W
+# Depth 31.0 km
+# UTC time 2020-10-06 05:54:50
+# Recorded at station Dolgoi Island, Mount Dutton, Alaska
+#
+# USGS event page: https://earthquake.usgs.gov/earthquakes/eventpage/us6000c5zm/executive
+#
+# DEM:  @earth_relief_06m
+# Data: @waveform_AV.DOL.txt
```
seisman, line 12:
```suggestion
# Data: @waveform_AV.DOL.txt
# 
# The data file @waveform_AV.DOL.txt contains 4 columns: 
# UTC times, and three-component movement in east-west, north-south and up-down directions
```

seisman (approved):

seisman:
Shall we tweet the animation, although it required GMT 6.2.0?

doc/examples/anim13/anim13.sh
```diff
@@ -0,0 +1,77 @@
...
+	gmt text -R0/4.5/0/13 -Jx1c -F+j+f -B0 -X-5.25c -Y-0.5c <<- EOF
+	2.25 12.5 CM 18p 2020-10-06
+	2.25 11.7 CM 9p SE of Sand Point, Alaska
+	2.25 11 CM 10p 54.844@.N, 159.842@.W
+	2.25 10.5 CM 10p 31.0 km, Mw 6.0
+	2.25 10.0 CM 10p 2020-10-06 05:54:50 UTC
+	EOF
+	gmt grdimage -R190W/130W/30/75 @earth_relief_06m -Ba30+f -JM3.5c -I+d -Cterra --MAP_FRAME_TYPE=plain -X0.75c -Y0.5c -BWNbr --MAP_ANNOT_OBLIQUE=lat_parallel --FONT_ANNOT_PRIMARY=9p
+	echo 161.86W 55.15N | gmt plot -St0.2c -GBlack
+	echo 159.842W 54.844N | gmt plot -Sa0.25c -Gred -Wfaint
```
meghanrjones, line 56:
```suggestion
	echo 159.851W 54.851N | gmt plot -Sa0.25c -Gred -Wfaint
```

doc/examples/anim13/anim13.sh
```diff
@@ -0,0 +1,77 @@
...
+	gmt meca -Sm2.5c -M -G110/168/255 -T0 -W0.5p -Yh+0.5c <<- EOF
+	-159.70 54.81 39 0.82 -0.68 -0.14 0.63 0.34 -0.40 25
```
meghanrjones, line 58: @seisman, why was the moment tensor coordinates different than the event coordinates? These may need to be updated based on https://earthquake.usgs.gov/earthquakes/eventpage/us6000c5zm/moment-tensor?source=us&code=us_6000c5zm_mww, but I want to check with you about proper coordinates first.

seisman: I think I used the focal mechanism from the [GCMT solution](https://www.globalcmt.org/cgi-bin/globalcmt-cgi-bin/CMT5/form?itype=ymd&yr=2020&mo=10&day=6&oyr=1976&omo=1&oday=1&jyr=1976&jday=1&ojyr=1976&ojday=1&otype=nd&nday=1&lmw=5.8&umw=10&lms=0&ums=10&lmb=0&umb=10&llat=-90&ulat=90&llon=-180&ulon=180&lhd=0&uhd=60&lts=-9999&uts=9999&lpe1=0&upe1=90&lpe2=0&upe2=90&list=0). 

I agree that it would be better to use consistent coordinates for the event and beachball, but note that the USGS page provides focal mechanism in a different format (strike, dip, rake), so we need to change `-Sm2.5c` to `-Sc2.5c`.

meghanrjones: OK, I will add a suggestion using the USGS data and will confirm that it appears the same as their figure (I can't remember the last time I needed to plot a beachball - certainly for some class - so this is exciting)

meghanrjones: I couldn't get the fill to work as expected -  any recommendations here?
```
gmt begin
	gmt meca -R190W/130W/30/75 -JM3.5c -Sc2.5c -M -G110/168/255 -T0 -W0.5p -Yh+0.5c <<- EOF
-159.851 54.851 40.5 245 29 93 61 61 88 1.048 18
EOF
gmt end show
```

seisman: I think you need to remove **-T0**.

> For double couple mechanisms, the -T option renders the beach ball transparent by drawing only the nodal planes and the circumference. For non-double couple mechanisms, -T0 option overlays best double couple transparently.

meghanrjones: > I think you need to remove **-T0**. It's documented, but I'm not sure if it makes sense or not.
> 
> > For double couple mechanisms, the -T option renders the beach ball transparent by drawing only the nodal planes and the circumference. For non-double couple mechanisms, -T0 option overlays best double couple transparently.

I can assume it makes sense to seismologists - just was not clear to me right away.

meghanrjones, line 58:
```suggestion
	gmt meca -Sc2.5c -M -G110/168/255 -W0.5p -Yh+0.5c <<- EOF
	-159.851 54.851 40.5 245 29 93 61 61 88 1.048 18
```
Looks like this:
image

seisman: Looks good.

doc/rst/source/animations/anim13.rst
```diff
@@ -0,0 +1,18 @@
+.. _anim13:
+
+(13) Animating Time-Series
+==========================
```
meghanrjones, line 4:
```suggestion
(13) Animating time-series (seismograms)
=========================================
```
One final recommendation

doc/rst/source/animations/anim13.rst
```diff
@@ -0,0 +1,18 @@
...
+
+Here we demonstrate how one can animate a time-series so that the line grows in length and temporarily
+change pen thickness and color/intensity as it is being written.  The trick we use is to convert the
```
meghanrjones, line 7:
```suggestion
Here we demonstrate how one can animate a time-series so that the line grows in length and the
pen thickness and color/intensity changes as it is being written.  The trick we use is to convert the
```

doc/examples/anim13/anim13.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+#
+# Movie of seismic waveforms from 2020-10-06 Alaska earthquake
+# Epicenter 54.844N 159.842W
```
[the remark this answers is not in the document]
PaulWessel: I think we dont need to update the animation - cannot tell difference of 0.01 degree locations on that small map.  But will update script.

PaulWessel:
Sorry, I guess if the focal mechanism changed and magnitude then we do need to update the animation.  Perhaps a good practice for @meghanrjones to go through that step?  let me know.

meghanrjones:
> Sorry, I guess if the focal mechanism changed and magnitude then we do need to update the animation. Perhaps a good practice for @meghanrjones to go through that step? let me know.

Yes, I can update it.

PaulWessel:
Any reason not to merge this now? I assume the YouTube link does not change when you update?

meghanrjones:
The link seems broken in the Vercel preview.
image

seisman:
> The link seems broken in the Vercel preview.

Need to update `doc/rst/source/animations.rst`.

doc/rst/source/animations.rst
```diff
@@ -71,6 +71,11 @@ includes the animations) does not take excessive time.
 
   :doc:`/animations/anim12`
 
+- .. youtube:: XXXXXXXXXX
```
meghanrjones, line 74:
```suggestion
- .. youtube:: eq_-2nYYf-s
```

meghanrjones (approved):
Fine to merge after committing the last suggestion (link update). I am making the animation now.

meghanrjones:
Actually, it might need a new url:

> Replace or delete your video
> 
> You can delete any video you uploaded to your YouTube channel. You can’t replace a video since any new video uploads will get a new URL, but you can change the existing video.

meghanrjones:
Video and links are updated, so it should be good to merge after the checks finish. Let me know if I should post on twitter as a '6.2 preview' - otherwise we can share after the release.

PaulWessel:
Hm, maybe getter to go with the RC1 release since easier for non-dev types to try out a bundle or installer?

meghanrjones:
> Hm, maybe getter to go with the RC1 release since easier for non-dev types to try out a bundle or installer?

Sounds good.

Issue #3766, opened 2020-07-29
Background and foreground colors for categorical CPTs


seisman:
**Description of the problem**

As per the [documentation](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#categorical-cpts):

> For categorical data, background color or foreground color do not apply. 

The actual behavior is inconsistent with the documentation.

**Full script that generated the error**

```bash
gmt begin map png

cat > types.cpt << EOF
0   yellow   ;desert
1   blue     ;forest
2   red      ;iceland
B   white
F   black
N   gray
EOF

gmt plot -R0/12/0/6 -Jx1c -Baf -BWSen -Sc0.8c -W1p -Ctypes.cpt << EOF
1  3  -1
2  3  0
3  3  1
4  3  1.5
5  3  2
6  3  3
7  3  3.1
8 3  NaN
EOF

gmt colorbar -Ctypes.cpt -DJBC+e+n
gmt end show
```

**Actual outcome**

![image](https://user-images.githubusercontent.com/3974108/88821913-35b8fb80-d191-11ea-82e4-633d29c8d387.png)

**Expected outcome**

The figure doesn't match the documentation:

1. It seems background and foreground colors (white and black) still works for categorical CPTs with **numeric keys**. If yes, we need to document it better, otherwise it's a bug.
2. If YES for point 1, then here is another bug. The color of the circle with z=3 should have a foreground color (black), but now it's red. If I change z=3 to z=3.1, it's black. 
3. z=1.5 is in blue, but for categorical data, z=1.5 is an undefined value.

**System information**

* Operating system: macOS
* GMT version (`gmt --version`): 6.2.0_2cce034_2020.07.28

PaulWessel:
I do not think we have checks that might apply to categorical data.  We should check that the input z value exactly matches the entries in the CPT and not the z = 1.5 which should give an error.  People who want z from 1-2 to give a constant color should use an discrete CPT, not a categorical one.  So we should start with implementing that test.  Likewise I think back and foreground does not make too much sense either but I suppose for numerical categories we could allow that.  For key-based categorically we clearly should not allow it.  I will see what checks I can add.

PaulWessel:
[...]

PR #4407, opened 2020-11-01
Improve and enforce categorical CPT usage

PaulWessel:

See #3766 for context.  We have never been careful about ensuring a unique categorical key matches those in the CPT.  For instance, many categorical CPTs uses integers as keys, such as 

```
0   yellow   ;desert
1   blue     ;forest
2   red      ;iceland
```

and in that case, only look-up values of 0,1, and 2 should return a valid color index.  Because these are the only categorical, it makes no sense of returning background or foreground colors for other keys, just the NaN key.  Furthermore, the **+e** modifier to **-D** that plots the fore and background color on colorbars should not apply to categorical CPTs.  This PR implements these checks and yields warnings for violations.  We should probably strengthen the CPTdiscussion in the cookbook about this.  I am sure there are people who think categorical CPTs can be used similarly to discrete CPTs and thus expects values like 1.5 to still result in blue in the CPT above.

Letting this be WIP since have not had a change to run all tests to see if we find other issues or bad usage among our own examples and tests, but @seisman can test this on the example in the original issue post #3766.

### DIFF
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -1056,8 +1056,8 @@ etc) it would be meaningless to compute average values (what would 1.5
 mean?). For such data a special format of the CPTs are provided.
 Here, each category is assigned a unique key, a color or pattern, and an
 optional label (usually the category name) marked by a leading
-semi-colon. Keys must be monotonically increasing but do not need to be
-consecutive. The format is
+semi-colon. Keys (if numerical) must be monotonically increasing but do
+not need to be consecutive. The format is
 
 +-----------------+--------+--------------+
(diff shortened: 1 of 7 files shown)


PaulWessel:
Only failure is that _sph_5.sh_ gives a slightly different sent of noise along edges.  Please check and see if you agree that this is OK and we can upload a new PS, @joa-quim or @seisman.

seisman:
This PR looks good to me.

> We should probably strengthen the CPTdiscussion in the cookbook about this.

Yes, we should do that.

PaulWessel:
I checked the documentation and it is actually correct regarding the fact there is no back/fore-ground colors for categorical data, only NaN.  So this is good to go now.

seisman (approved):

Issue #4408, opened 2020-11-02
Improve output of absolute time coordinates


PaulWessel:
**Description of the desired feature**

While GMT writes proper ASCII output absolute time coordinates in ISO format, there are some **awkward limitations**.  The output format is controlled by **FORMAT_DATE_OUT** and **FORMAT_CLOCK_OUT**. The problem occurs when the time increments are very small.  The default **FORMAT_CLOCK_OUT** is _hh:mm:ss_ so if the time coordinates have a higher precision than 1 second then those decimals are lost unless the user is aware of this problem and adds **--FORMAT_CLOCK_OUT**=_hh:mm:ss.xxxxx_, where _xxxxx_ may be the required number of decimals to do the data justice.  This is not a good system as it requires us to constantly add **--FORMAT_CLOCK_OUT** to commands to make sure we do not loose precision in ASCII output. Most users will not have the wherewithal to know this and get frustrated when their output is ruined.

**Solutions:**

1. If any of the output columns contain absolute time, we could scan that column to find the widest sub-second remainder and automatically set a _hh:mm:ss.xxxxx_ that can handle the maximum precision needed.
2. We could do this only when GMT creates an array to make sure the array can be printed with adequate precision.

(1) means we must scan all the rows and determine `sub = S->data[tcol][row] - floor (S->data[tcol][row])` and keep track of the **widest nonzero** _sub_.  Say we have three records with UNIX times 1601514000.5, 1601514010.75, 1601514020.00.  These are ~10 seconds apart but to preserve precision we clearly need to use _ss.xx_.    We would find _sub_ = 0.5 and 0.75 and learn we need 2 decimals. (2) means we only do this for a fixed increment _inc_ and we then apply the same scheme to _inc_ to learn how many decimals.  E.g., if we set up an equidistant time _inc_ of 2 msec then we should find 3 decimals, _ss.xxx_.

[...]

PR #4409, opened 2020-11-03
Let GMT set a better output format for fractional seconds

PaulWessel:
See #4408 for background.  This PR implements it for table output.  If **-V** is used you may see a message like this:

`gmtmath [INFORMATION]: To prevent loss of time-series precision we have changed FORMAT_CLOCK_OUT to hh:mm:ss.xxxxx`

Note this action is only taken if **FORMAT_CLOCK_OUT** has not been changed from its default value.

I am not sure what to do about rec-by-rec tools which have no way of detecting the format.  We should probably

1. Add some more language under the **FORMAT_CLOCK_OUT** documentation as to under what circumstances the user may wish to change this setting (i.e., dense time-series data).
2. Clarify that GMT may override the default format when data-precision loss is likely (and how we do it).

Thoughts on this?


joa-quim (approved):

joa-quim:
The rec-by-rec could also analyze the first 20 (or 10) recs and change to ``.xx``. Then only the first 20 recs would have had an eventual loss and issuing a warning.

PaulWessel:
True, but the problem is that test would be run forever since it would go in GMT_Put_Record.

PaulWessel:
Looking at where _GMT_Put_Record_ is used, I can only think of a few modules where this would be an issue:

1. **filter1d**: Filtering time-series and **-I** specifies sub-second interval.
2. **greenspline**: 1-D of interpolating time-series and **-I** specifies sub-second interval.

In those cases we can just check in those modules and update **FORMAT_CLOCK_OUT** if necessary.

Of course, there may be folks who will be running blockmean on (time,distance,value) data and could run into this as well, but I dont see a good way to deal with this, other than improve the **FORMAT_CLOCK_OUT** documentation.

PaulWessel:
Oh, and sample1d as well.

PaulWessel:
Not **sample1d** since it uses _GMT_Write_Data_.  So I only added a check for **filter1d** as the most likely of the modules to deal with time-series.  I think that is a good start - and the docs have been updated to indicate the issue.

PR #4432, opened 2020-11-07
Fix several psternary problems and add enhancements

PaulWessel:

This PR started with trying to address the [forum request](https://forum.generic-mapping-tools.org/t/how-to-reverse-the-axis-of-the-psternary-plot/978/3) to plot the axes in reverse direction.  This is now an option via the new **-D** selection.  However, in the process of testing this I discovered that the points plotted and the axes orientations where out of sync.  The initial implementation of **psternary** basically assumed that the axes were positive in a clock-wise orientation but instead they were plotted in a counter-clockwise orientation.  A separate PR fixed the ignoring of -N (#4431).  This PR addresses therefore several things:

1. It adds **-D** to reverse the axes orientations.
2. It fixes the mixup between projecting (_a,b,c_) to (_x,y_) for the two -D settings.
3. It adds a new test (_corners.sh_) that ensures end-member points (1,0,0), (0,1,0) and (0,0,1) correctly plots.
4. It adds _-D_ to the existing _psternary.sh_ script so that the axes are in the correct order for that example.
5. It adjusts the placement of vertices labels (**-L**) which is affected by **-D**.
6. It allows any of the **-L** labels to be given as - which skips that label.

The psternary.ps original has been updated and all tests pass.  Below is the corners.png result showing correct calibrations:

corners

### DIFF
--- a/doc/rst/source/psternary.rst
+++ b/doc/rst/source/psternary.rst
@@ -13,11 +13,10 @@ Synopsis
 .. include:: common_SYN_OPTs.rst_
 
 **gmt psternary** [ *table* ]
-[ **-JX**\ *width* ]
+[ **-JX**\ [-]\ *width* ]
 [ |-R|\ *amin/amax/bmin/bmax/cmin/cmax* ]
 [ |SYN_OPT-B| ]
 [ |-C|\ *cpt* ]
-[ |-D| ]
 [ |-G|\ *fill* ]
 [ |-K| ]
 [ |-L|\ *a*\ /*b*\ /*c* ]
--- a/doc/rst/source/psternary.rst
+++ b/doc/rst/source/psternary.rst
@@ -17,6 +17,7 @@ Synopsis
 [ |-R|\ *amin/amax/bmin/bmax/cmin/cmax* ]
 [ |SYN_OPT-B| ]
 [ |-C|\ *cpt* ]
+[ |-D| ]
 [ |-G|\ *fill* ]
 [ |-K| ]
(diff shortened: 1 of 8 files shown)


seisman:
> 1. It adds **-D** to reverse the axes orientations.

What happens if -JX is given a negative size?

PaulWessel:
My first thought was just to use **-JX**-_width_.  There are some pros/cons here we can discuss:

1. Given **-M** (which requires no **-R -J**) it seemed more flexible to make a separate option for this effect.
2. If we accepted **-JX**-_width_, we would need to strip off that sign, set the internal Ctrl->D.active variable, and proceed with **-JX**_width_ due to how **psternary** was crafted (we do our own annotations, etc, since there are three axes and no GMT ternary-projection or basemap).

So perhaps the best option is to allow **-JX**-_width_as well (since many will be familiar with the effect elsewhere) but then do the internal switcharoo as described above.

If nothing else, we would need to catch a negative width and say something, but I think the proposal above is better.

PaulWessel:
Now -JX-10c has the same effect as -D -JX10c.

PaulWessel:
I am also open to a discussion if we should _not_ add **-D** since eventually, **psternary** may be asked to plot polygons (see the forum request for what this may be) and even contours/images.  Wasting option letters on things like **-D** now will make it harder to claim them for more important features later.  So perhaps **-JX**_-width_ is all we really should do?

PaulWessel:
Hi @seisman I ended up removing -D and doing it all via -JX-width to be consistent with other operations and to save an option for another day.

seisman (approved):

PR #4447, opened 2020-11-11
Let build script fix all the SCM CPTs that have either soft or hard hinge

PaulWessel:

As we stretch these to go from -1/+1 we need to add a zero-slice in there for all these, not just oleron.cpt.

Fixes the hinge problem discussed [here](https://forum.generic-mapping-tools.org/t/two-color-cpt-at-1-and-1-with-no-centre-value/1025/7).


weiji14:
Should `roma` and `batlow` have a soft hinge at 0 too like `oleron`?

image

PaulWessel:
That is a good question.  Fabio is not clear on this concept but roma is certainly like seis and has that white patch as a "hinge", but batlow does not look like having a hinge to me.

weiji14:
> That is a good question. Fabio is not clear on this concept but roma is certainly like seis and has that white patch as a "hinge", but batlow does not look like having a hinge to me.

Ok, `roma` is a diverging colourmap according to http://www.fabiocrameri.ch/roma.php, but [`batlow`](http://www.fabiocrameri.ch/batlow.php) isn't. So maybe set the hinge on `roma` only.

PaulWessel:
Agree, I will do that.

PaulWessel:
OK, roma added to this PR.

src/gmt_cpt_masters.h
```diff
@@ -69,23 +69,23 @@
...
 "tokyo       : Perceptually uniform colormap without black or white, by Fabio Crameri [C=RGB]",
 "tokyoS      : Perceptually uniform categorical colormap without black or white, by Fabio Crameri [C=RGB]",
 "topo        : Sandwell/Anderson colors for topography [R=-7000/+7000, H, C=HSV]",
 "turku       : Perceptually uniform colormap by Fabio Crameri [C=RGB]",
 "turkuS      : Perceptually uniform categorical colormap by Fabio Crameri [C=RGB]",
 "turbo       : Google's Improved Rainbow Colormap for Visualization [C=RGB]",
-"vik         : Perceptually uniform bimodal colormap, light, by Fabio Crameri [S,C=RGB]",
-"vikO        : Perceptually uniform bimodal cyclic colormap, light, by Fabio Crameri [S,C=RGB]",
+"vik         : Perceptually uniform bimodal colormap, light, by Fabio Crameri [S, C=RGB]",
+"vikO        : Perceptually uniform bimodal cyclic colormap, light, by Fabio Crameri [S, C=RGB]",
```
weiji14, line 88:
```suggestion
"vikO        : Perceptually uniform bimodal cyclic colormap, light, by Fabio Crameri [C=RGB]",
```
Are soft hinges possible with cyclic colormaps :thinking:?

PaulWessel:
Maybe @joa-quim or @seisman can approve this one now?  I assume the test failures are CI related...

joa-quim (approved):

Issue #4445, opened 2020-11-11
Ability to place seismological/geological symbols in legends


PaulWessel:
**Description of the desired feature**

Modules like **meca**, **velo**, etc. plot advanced symbols in seismology and geodesy and use a geographic region (**-R**), map projection (**-J)** and input _longitude, latitude_ to place these features on a map.  There are times when we also would like to plot such a symbol outside a map, such as in a legend.  In legends, there usually is no geographic region nor projection, and at best there is a Cartesian **-R -J** set up for placing information.  Because the above-mentioned tools reads longitude, latitude of where a feature is it becomes a guessing game to pick **-R -J** and input coordinates to align the symbol, say, centered in a legend.

One way to handle this would be to add a new option to these modules that, like **-D** for colorbars or scales, deals with this issue.  Examining the various modules I see that the option **-I** (for information) is available [or **-Z** if **coupe** is can be excluded].  Thus, a solution may be this:

1. Add **-I**[**g**|**j**|**J**|**n**|**x**]_refpoint_ for specifying where the special symbol should be plotted (with the usual ways of _refpoint_ values like coordinates or codes)
2. When **-I** is used the input longitude, latitude is ignored and placement is given via **-I** instead.

For instance, to center a focal mechanism in the middle of a 5x10 cm legend area we could do

```
gmt meca -R0/10/0/5 -Jx1c -Sm2.5c -M -G110/168/255 -T0 -W0.5p -Yh+0.5c -IdCM <<- EOF
-159.70 54.81 39 0.82 -0.68 -0.14 0.63 0.34 -0.40 25
EOF
```

and the actual earthquake location (`-159.70 54.81 `) is simply ignored.  I think the proposed mechanism will make it easier to add special symbols such as GPS velocity with error symbols and focal mechanism to user legends.

Comments, @joa-quim and @seisman? Apart from meca and velo, does it make sense to extend this solution to **polar**, **coupe**, and **sac**?

seisman:
Is it worth wasting the **-I** option for such rare cases? I probably would manually modify the longitude and latitude to position it at the specific location.

PaulWessel:
[...]

PR #4449, opened 2020-11-12
WIP Testing meca -I to place symbols in a legend or box

PaulWessel:

See #4445.  This PR tests this concept.  For example you can try this command and change the **-I** choices:

`echo -159.70 54.81 39 0.82 -0.68 -0.14 0.63 0.34 -0.40 25 | gmt meca -R0/10/0/5 -Jx1c -B0 -Sm2.5c -M -G110/168/255 -T0 -W0.5p -IjTR+o0.25c -png map`

map

Few things to know:

1. Regardless of input file, only the first record is plotted - we then skip the rest
2. The symbol size is what **-S** gives - no scaling by magnitude
3. The **-N** is turned on since **-IJ** will place things outside rather than inside the box
4. While **-Ix** should not need **-R -J** I have not implemented that yet since affects **THIS_MODULE_NEEDS**
5. The **-C** option should be disallowed or ignored when **-I** is used.

We should make a test page of all the 9 meca symbols using the 9 adjustment points.

### DIFF
--- a/src/seis/psmeca.c
+++ b/src/seis/psmeca.c
@@ -66,7 +66,7 @@ struct PSMECA_CTRL {
 		bool active;
 		struct GMT_FILL fill;
 	} G;
-	struct PSMECA_I {	/* -D[g|j|n|x]<refpoint>[+o<dx>[/<dy>]] */
+	struct PSMECA_I {	/* -D[g|j|J|n|x]<refpoint>[+o<dx>[/<dy>]] */
 		bool active;
 		struct GMT_REFPOINT *refpoint;
 		double off[2];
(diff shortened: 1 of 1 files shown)


PaulWessel:
Hi @seisman, I think you voted this down when I put this together.  If that is still the sentiment then I can delete this branch, yes?

seisman:
If I understand it correctly, after implementing the new legend code proposed in https://github.com/GenericMappingTools/gmt/issues/4445#issuecomment-751553334, the **-I** option implemented in this PR is no longer needed. If anyone wants to place a beachball at the top-right corner, he can always pass the new legend code to `legend`.

PR #4454, opened 2020-11-16
Introducing and using struct GMT_CUBE for modules needing 3-D grids

PaulWessel:

This PR (so far) introduces a new struct **GMT_CUBE** for holding a grid header (for the x-y information), a few complementary variables for the third dimension, and a data pointer holding the data cube (float) [I have deleted half-baked and abandoned branches _three-D-grids_ (#2966) and _new-3D-io-push_].  I have hooked temporary functions up so that the API functions _GMT_Create_Data_, _GMT_Read_Data_, and _GMT_Write_Data_ can accept the **GMT_IS_CUBE** family.  I have then rewritten parts of **grdinterpolate** and **greenspline** (the 3-D case; I also optimized this module a bit) so they use these new containers and the standard i/o functions.  Tests pass after rewriting _gspline_4.sh_ test for 3-D gridding to handle the new 3-D output.  This branch is still WIP because there are a few more things to do:

1.  Need to integrate the above temporary functions properly so that these containers are _registered_ with the API.  This allows for the _automatic garbage collection_ to work as well as the _virtual file mechanism_ for passing memory data cubes to other modules.
2. Once that is done then I can finish the **GMT_IS_CUBE** entry in _GMT_Destroy_Data_ as well.
3. I need to add that case to _GMT_Duplicate_Data_.
4. I need to add two cases to _GMT_Convevert_Data_: Grids to Cube and Cube to Grids.

With all this in place I can start, hopefully with help from @joa-quim, to change the current case for _GMT_Write_Data_ to write a 3-D netCDF file instead of a list of 2-D grids (Reading the 3-D grid works fine but can be optimized further).

@GenericMappingTools/core may wish to look at gmt_resources.h and search for DATACUBE and comment if they have concerns about variable names, etc.


PaulWessel:
Also, I will need to add new documentation to the API after the implementation.

PaulWessel:
I have now completed the update discussed above and added some tests.  The new **GMT_CUBE** container is now fully supported by the API.  Because only **grdinterpolate** and **greenspline** actually deal with 3-D cubes the testing is limited, but I added _testapi_cube.c_ which tests aspects such as (1) reading in a cube, (2) writing out the cube, (3) opening a virtual file on a memory cube both for input and output and successfully pass then to a **grdinterpolate** command, then writing out the result.  There is a minor memory leak that I will hut down and fix soon but only visible to those who compile with **MEM_DEBUG**.  I have also updated the API documentation.  I have marked the **GMT_CUBE** as experimental in that (at least until 6.2.0 release) we reserve the right to make any changes we see fit.

The key limitation to work on now is to handle the writing of a 3-D output cube [currently we are still limited to doing this by way of a stack of grids].  Please review and hopefully accept this PR, @joa-quim and @seisman.  All tests, including revised and new ones, pass.

seisman:
There are some compilation errors on Linux and Windows:
```
../src/greenspline.c:2474:69: error: ‘Z’ undeclared (first use in this function)
 #pragma omp parallel for private(V,row,col,ij,p,r,C,part,wp) shared(Z,dimension,yp,header,xp,X,Ctrl,GMT,alpha,Lz,norm,data,par,nz_off,z_level,nm,normalize)
```

PaulWessel:
> There are some compilation errors on Linux and Windows:
> 
> ```
> ../src/greenspline.c:2474:69: error: ‘Z’ undeclared (first use in this function)
>  #pragma omp parallel for private(V,row,col,ij,p,r,C,part,wp) shared(Z,dimension,yp,header,xp,X,Ctrl,GMT,alpha,Lz,norm,data,par,nz_off,z_level,nm,normalize)
> ```

Sorry, those were seen as comments for clang... Now fixed.

PaulWessel:
Need @joa-quim or @seisman to OK this branch.  Tests all pass.

src/testapi_cube.c
```diff
@@ -0,0 +1,38 @@
+#include "gmt_dev.h"
```
seisman, line 1: Is it OK to use `gmt.h` instead of `gmt_dev.h`?
```suggestion
#include "gmt.h"
```

test/api/apicubeplot.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+#
+# Test the C API for reading via a matrix and making a plot
+ps=apicubeplot.ps
+if [ ! -f cube.nc ]; then
+	cp "${src}"/../grdinterpolate/cube.nc .
+fi
+testapi_cube
```
seisman, line 8: Like #4483, you have to add a wrapper for `testapi_cube` in `test/gmtest.in`.

seisman (approved):
I didn't read the C codes. The GMT_CUBE struct looks good to me. 

There are a few typos.

src/testapi_cube.c
PaulWessel: Yes, committed separately since another change was needed as well.

test/api/apicubeplot.sh
PaulWessel: OK, did that now.

PR #4460, opened 2020-11-18
make/grd2cpt: Ensure that if -C is given it has an argument

### LINKED ISSUE #4459: Check if the CPT name is empty
Here is the command to reproduce the issue:
```
$ gmt grd2cpt @earth_relief_01d -C -R0/10/0/10 > 123.cpt
grd2cpt [ERROR]: CPT gmtapi_colors2cpt_77472.cpt has no z-slices!
[Session gmt (0)]: Error returned from GMT API: GMT_CPT_READ_ERROR (8)
[Session gmt (0)]: Error returned from GMT API: GMT_CPT_READ_ERROR (8)
```

It looks like a user error, but at least we should improve the error message.

PaulWessel:
While there are default CPTs, these only go into effect if **-C** is _not_ given.  If **-C** _is_ given with _no_ argument it is an error, now caught by both **makecpt** and **grd2cpt**.  Closes #4459.

### DIFF
--- a/src/grd2cpt.c
+++ b/src/grd2cpt.c
@@ -421,11 +421,9 @@ static int parse (struct GMT_CTRL *GMT, struct GRD2CPT_CTRL *Ctrl, struct GMT_OP
 		}
 	}
 
-	if (GMT->current.setting.run_mode == GMT_CLASSIC) {
-		if (Ctrl->H.active) {
-			n_errors++;
-			GMT_Report (GMT->parent, GMT_MSG_ERROR, "Unrecognized option -H\n");
-		}
+	if (Ctrl->H.active && GMT->current.setting.run_mode == GMT_CLASSIC) {
+		GMT_Report (GMT->parent, GMT_MSG_WARNING, "Option -H: Only available in modern mode - ignored in classic mode\n");
+		Ctrl->H.active = false;
 	}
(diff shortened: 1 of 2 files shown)


src/grd2cpt.c
```diff
@@ -421,10 +421,14 @@ static int parse (struct GMT_CTRL *GMT, struct GRD2CPT_CTRL *Ctrl, struct GMT_OP
 		}
 	}
 
-	if (Ctrl->H.active && GMT->current.setting.run_mode == GMT_CLASSIC) {
-		n_errors++;
-		GMT_Report (GMT->parent, GMT_MSG_ERROR, "Unrecognized option -H\n");
+	if (GMT->current.setting.run_mode == GMT_CLASSIC) {
+		if (Ctrl->H.active) {
+			n_errors++;
+			GMT_Report (GMT->parent, GMT_MSG_ERROR, "Unrecognized option -H\n");
```
joa-quim, line 427: It would be better to say that "-H option is not valid in classic mode". Or even better, just ignore it silently.

PaulWessel: OK, and I changed it to a warning.  Not a fan of silence for wrong use.

seisman (approved):

src/grd2cpt.c
joa-quim: OK, but remember that sometimes the commands were not written by humans.

joa-quim (approved):

PR #4461, opened 2020-11-18
Fix cmake abspath conversions

rtburns-jpl:
CMAKE_INSTALL_*DIR variables are not necessarily relative
to the install prefix - they may be absolute, in which case
appending them to CMAKE_INSTALL_PREFIX is not valid.

Fixes installation paths when provided GNU_INSTALL_*DIRs are not relative to CMAKE_INSTALL_PREFIX, as used by some tooling - e.g. `-DGNU_INSTALL_LIBDIR=/usr/lib/`
See https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path for more info.

**Reminders**


seisman:
Will review this PR in the next few days.

cmake/modules/JoinPaths.cmake
```diff
@@ -0,0 +1,23 @@
+# This module provides function for joining paths
+# known from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
+# Copyright 2020 Jan Tojnar
+# https://github.com/jtojnar/cmake-snips
+#
+# Modelled after Python’s os.path.join
```
seisman, line 8: Avoid using non-ASCII characters.
```suggestion
# Modelled after Python's os.path.join
```

seisman (commented):
It looks good, except a tiny fix.

seisman:
@rtburns-jpl After merging this PR, we can't build the GMT macOS bundle anymore. Please see the issue report in https://github.com/GenericMappingTools/gmt/issues/4481. 

It's unclear to me how CMake builds bundles, so we have to revert the changes in this PR (https://github.com/GenericMappingTools/gmt/pull/4484) now.

If you know more about CMake and know how to fix it, we're glad to re-merge your changes again.

PR #4464, opened 2020-11-19
Use modern gs transparency command

### LINKED ISSUE #4453: Compatibility with the old .setopacityalpha and new .setfillconstantalpha/.setalphaisshape PostScript
The issue was discussed in another issue #3282. Here I open a separate issue for better discussions.

---

Just build their master branch. The transparency bug in 9.51 and 9.52 is fixed. Now I can see transparency in PDF, JPG and PNG.

However, I get the following warnings for transparent plots:
```
   **** WARNING: .setopacityalpha is deprecated (as of 9.53) and will be removed in a future release
   **** See .setfillconstantalpha/.setalphaisshape for the improved solution
```

_Originally posted by @seisman in https://github.com/GenericMappingTools/gmt/issues/3282#issuecomment-626022540_

---

This is a threat for us, so we need to figure out how to rewrite the PSL_transp PostScript function that is defined in the PSL_strings.h include file.  I have yet to find any documentation on .setfillconstantalpha and how to use it.  Just replacing .setopacityalpha with . setfillconstantalpha did not seem to wark (no errors, but no transparency).  We will need to research this more to better document how we do this, possibly asking the gs people for advice.

_Originally posted by @PaulWessel in https://github.com/GenericMappingTools/gmt/issues/3282#issuecomment-627601404_

---

Comparing the 9.52 documentation and the "master/dev" documentation, it seems .setfillconstantalpha/.setalphaisshape are added in 9.53.

- 9.52: https://www.ghostscript.com/doc/9.52/Language.htm#Transparency
- dev: https://www.ghostscript.com/doc/current/Language.htm#Transparency

_Originally posted by @seisman in https://github.com/GenericMappingTools/gmt/issues/3282#issuecomment-627625579_

---

> Just replacing .setopacityalpha with . setfillconstantalpha did not seem to wark (no errors, but no transparency).

It works for me with gs 9.53 (built from their master a few days ago). It doesn't work for you possibly because you're still using 9.50?

_Originally posted by @seisman in https://github.com/GenericMappingTools/gmt/issues/3282#issuecomment-627627733_

---

Yes.  So we cannot drop it in yet, but also cannot wait until they drop support.  I guess we need a way to issue the right code based on gs version.  Have to think about how that should be done, maybe with a bit-flag passed into PSL_plotinit.

_Originally posted by @PaulWessel in https://github.com/GenericMappingTools/gmt/issues/3282#issuecomment-627628719_

PaulWessel:
Yet, we check if our gs is old and if so we stitch in the older transparency command.  At some point we should remove that crutch.  Closes #4453.


seisman:
Running this command:
```
gmt basemap -R0/10/0/10 -JX10c -Baf -t90 -png test
```

**master**
image

**This branch**
image

PaulWessel:
And manually editing the PS and putting back the old command gives the correct answer.  Since all our plotting commands are between the two PSL_transp calls I wonder if they have a bug...

PaulWessel:
The problem is that the old .setopacityalpha has been replaced with **two** settings: .setfillconstantalpha and .setstrokeconstantalpha.  So, there are actually two options facing us:

1. We can set both and then GMT should behave exactly as before
2. We allow **-t** and `@transp` arguments to take optional **+s** (stroke) or **+f** (fill) to allow for enhanced flexibility.

I am not sure when this is useful but I note that the gs developers felt this is useful.  Perhaps fading a color in a polygon but leaving lines solid is not a bad idea, so I think option 2 is better - but requires gs 9.53 of course to have an effect.

PaulWessel:
Following up on last post, there are actually several issues to decide here going forward:

**A. Backwards compatibility**

1. I think **-t**_transp_ should set _both_ stroke and fill transparency, as it currently does.
2. Note that **plot**, **text** and **plot3d** can take  **-t** with no arguments and then read transparency from data column.  We then read _one_ value but assign it to both types of transparency, to be consistent with (1).

**B. New features**

As discussed, extend the syntax for **-t** to now be **-t**[_ftransp_[/_stransp_]][**+s**][**+f**] where

1. _ftransp_ and _stransp_ are fill and stroke transparencies, respectively.  If only _ftransp_ is given it means _stransp = ftransp_ unless there are modifiers.
2. Modifiers **+f** and **+s** limits the single given transparency to either fill or stroke
3. If no transparencies are given then we read one from file (_ftransp = stransp_) unless one or both modifiers are given.

Comments or concerns, @joa-quim and @seisman ?

PaulWessel:
Note that our transparency macro supports both ghostscript and Adobe Destiller:

```
/PSL_transp {						% Set transparencies
  /.setfillconstantalpha where {pop .setblendmode .setfillconstantalpha}{		% Using ghostscript
  /pdfmark where {pop [ /BM exch /CA exch dup /ca exch /SetTransparency pdfmark}	% Or Adobe
  {pop pop} ifelse} ifelse				% Or skip if neither are supported
}!
```

and looking (again) and the pdfmark documentation I can now remember that CA and ca sets the two different types of transparencies and we are just duplicating them (dup) since we only pass on number to PSL_transp.  Going forward, we will need to pass two numbers in order to set different values, e.g.,

`0.2 0.5 /Normal PSL_transp`

I think a first step is simply to change the macro to take two numbers and then change the code that writes these commands to write two numbers (duplicates).  Then the above enhancements can take place next.

Information on pdfmark can be found [here](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/pdfmarkReference_v9.pdf).  The example they give on top of page 41 is actually what we get (using the path twice) in GMT, so perhaps we need to be more careful when transparency applies to both path and fill.

PaulWessel:
I have implemented the dual transparencies and fixed the _PSL_transp_ macro for gs 9.53.  I also let **psconvert** handle version mismatches on the fly: New PS with old gs as well as Old PS (our PS origs) and new gs.  I have tested the transparency for both ghostscript 9.53 and 9.50 and Adobe Distiller as well as none available, so the new macro works.  I have updated the docs and codes and all tests pass.  I will need help with the new explain_-tv.rst_ and explain_-tv_full.rst_ since the link from the synopsis do not work and I get

`common_SYN_OPTs.rst_:63: WARNING: undefined label: -tv_full (if the link has no caption the label must precede a section header)`

Please give the update GMT PS a spin.

doc/rst/source/explain_-tv_full.rst_
```diff
@@ -0,0 +1,12 @@
+.. _-tv_full:
+
+**-t**\ [*transp*\ [/*transp2*]][**+f**][**+s**]
+    Set transparency level for an overlay, in [0-100] percent range. [Default
+    is 0, i.e., opaque].  Only visible when PDF or raster format output is selected.
+    Only the PNG format selection adds a transparency layer in the image (for further processing).
+    If given, *transp* applies to both fill and stroke, but you can limit the transparency to
+    one of them by appending **+f** or **+s** for fill or stroke, respectively.  Alternatively,
+    append /*transp2* to set separate transparencies for fills and strokes.  If no transparencies
+    are given then we expect to read them from the last data column(s).  Use the modifiers
```
seisman, line 10:
```suggestion
    are given then we expect to read them from the last numerical column(s).  Use the modifiers
```

src/gmt_constants.h
```diff
@@ -419,6 +419,14 @@ enum GMT_enum_colorswap {
 	GMT_USE_FILL_RGB  = 1,	/* Take pen color from that of the current fill */
 	GMT_USE_PEN_RGB = 2};	/* Take fill color from that of the current pen */
 
+/*! Index for fill/stroke transparency value */
+enum GMT_enum_transp {
+	GMT_FILL_TRANSP     = 0,	/* transp[GMT_FILL_TRANSP] is used for filling */
+	GMT_PEN_TRANSP      = 1,	/* transp[GMT_PENTRANSP] is used for stroking */
```
seisman, line 425:
```suggestion
	GMT_PEN_TRANSP      = 1,	/* transp[GMT_PEN_TRANSP] is used for stroking */
```

src/postscriptlight.c
```diff
@@ -417,7 +417,15 @@ typedef struct {
...
 
-/* These are used when the PDF pdfmark extension for transparency is used. */
+/* These are used when the PDF pdfmark or Ghostscript extensions for transparency is used:
+ * Adobe:       https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/pdfmarkReference_v9.pdf
+ * Ghostscript: https://www.ghostscript.com/doc/current/Language.htm#Transparency
+ *
+ * From gs 9.53 their transparency model takes two transparencies (stroke and fill) while before
+ * it only took one.  The pdfmark took two but we simply duplicated it since GMT itself only dealt
+ * with one transparency for both.  From GMT 6.2.0 we will allow these two transparencies to be set
+ * individually if the user so selects.  NOte: We do not support any of the soft masks/shapes stuff.
```
seisman, line 427:
```suggestion
 * individually if the user so selects.  Note: We do not support any of the soft masks/shapes stuff.
```

seisman:
These two commands give me the same figures:
```
gmt basemap -R0/10/0/10 -JX10c -Baf -t90+s -png test
gmt basemap -R0/10/0/10 -JX10c -Baf -t90+f -png test
```

PaulWessel:
Thanks, I can see why and will fix shortly.

PaulWessel:
Try these and contrast:

```
gmt basemap -R0/10/0/10 -JX10c -Baf -t60+f -pdf test --FONT_ANNOT_PRIMARY=16p,Helvetica,red=0.25p,blue
gmt basemap -R0/10/0/10 -JX10c -Baf -t60+s -pdf test --FONT_ANNOT_PRIMARY=16p,Helvetica,red=0.25p,blue

```

PaulWessel:
Or your original examples.

seisman (approved):
Looks good to me, but also see the tiny fixes in my previous review.

PaulWessel:
For @joa-quim : This PR prevents those warnings when gs 9.53 is run on previously built PS made before this PR, such as our PS originals in GitHub.

seisman:
> I will need help with the new explain_-tv.rst_ and explain_-tv_full.rst_ since the link from the synopsis do not work and I get
> 
> `common_SYN_OPTs.rst_:63: WARNING: undefined label: -tv_full (if the link has no caption the label must precede a section header)`
> 
> Please give the update GMT PS a spin.

The warning is because, `explain_-tv_full.rst_` is not included anywhere.

PaulWessel:
OK, yes, the other _full are included by gmt.rst but here -t is slightly different in the -tv.  Is there a way to include it but not show it?  Or perhaps just include full in the three module rst files instead of the short one?

PaulWessel:
Perhaps that is the best solution so it is clear without clicking that see more link that there are a few changes here.

seisman:
> Perhaps that is the best solution so it is clear without clicking that see more link that there are a few changes here.

Sounds OK to me.

doc/rst/source/text_common.rst_
```diff
@@ -226,6 +226,6 @@ Optional Arguments
 
 .. include:: explain_-qi.rst_
 
-.. include:: explain_-tv.rst_
+.. include:: explain_-tv_full.rst_
```
seisman, line 229: Can we remove the file `explain_-tv.rst_` now?

PaulWessel: yes, otherwise it will not be referenced.  WiIl do.

PR #4495, opened 2020-11-26
Fix #4493

### LINKED ISSUE #4493: grdimage can't read selected bands from GeoTIFF
See [forum post](https://forum.generic-mapping-tools.org/t/how-to-plot-satellite-optical-image-of-geotif-format-and-show-the-true-color/225/13) for reference.

Debugging shows that gdalread is called 4 times, three of them in the *info* mode. The second time it carries the bands info but that is lost in subsequent calls and when it finally reads the data, the entire bands set is read. This happens because ``grdimage`` has this block that consumes and looses the bands request
```
	if (!Ctrl->D.active && ftype == GMT_IS_GRID) {	/* See if input could be an image of a kind that could also be a grid and we don't yet know what it is.  Pass GMT_GRID_IS_IMAGE mode */
		if ((I = GMT_Read_Data (API, GMT_IS_IMAGE, GMT_IS_FILE, GMT_IS_SURFACE, GMT_CONTAINER_ONLY | GMT_GRID_IS_IMAGE, NULL, Ctrl->In.file, NULL)) != NULL) {
			gmtlib_read_grd_info (GMT, Ctrl->In.file, I->header);	/* Re-read header as grid to ensure orig_datatype is set */
```

joa-quim:
When stripping the +b.... in GMT_Register_IO() it latter always reset a '=' instead of the required '+'

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -8024,7 +8024,7 @@ int GMT_Register_IO (void *V_API, unsigned int family, unsigned int method, unsi
 			file = strdup (resource);
 			if (direction == GMT_IN) {	/* For input we can check if the file exists and can be read. */
 				char *p = NULL;
-				bool not_url = true;
+				bool not_url = true, is_plus_sign = false;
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Thanks, got caught up in other changes.

Issue #4497, opened 2020-11-28
White lines and white MAP_FRAME_PEN don't go along.


joa-quim:
```
printf "0.0 0\n1.1 1" | gmt plot -JX12c/8c -Baf -BWSen+tBLA -R-0.04/1.12/-0.04/1.04  --MAP_FRAME_PEN=thicker,white --PS_PAGE_COLOR=gray -W1,white -png lixo
```

![lixo](https://user-images.githubusercontent.com/537321/100492771-71f06e80-3127-11eb-8038-14c7f85934c3.png)

PaulWessel:
Probably a case for **GMT_THEME**=_inverse_.
Yet, that W axis should be white like the test.

PaulWessel:
Interestingly, the basemap command work:

`gmt basemap -JX12c/8c -Baf -BWSen+tBLA -R-0.04/1.12/-0.04/1.04  --MAP_FRAME_PEN=thicker,white --PS_PAGE_COLOR=gray -png lixo`

So somewere we must be reverting to black before stroking.  Will look.

joa-quim:
> Probably a case for GMT_THEME=inverse.

That's were it started (*dark* actually).

And if I take the ``darkmode.conf`` make it a ``gmt.conf`` and  run
```
printf "0.0 0\n1.1 1" | gmt plot -JX12c/8c -Baf -BWSen+tBLA -R-0.04/1.12/-0.04/1.04 --PS_PAGE_COLOR=gray -png lixo
```
then it cashes.

joa-quim:
The crash occurs in ``gmtremote_data_load()``. It should not have to visit that function when plotting a local file.

```
printf "0.0 0\n1.1 1" | gmt plot -JX12c/8c -Baf -BWSen+tBLA -R-0.04/1.12/-0.04/1.04 --PS_PAGE_COLOR=gray -png lixo
```

PaulWessel:
Working to fix the initial problem and have done so and find my fixes reveals 4 new failures that are actually corrections.  One is mgd77/cm4.sh which I think you wrote.  In it, you plot 3 axes WSn and then you plot the E axis with a separate setting: --MAP_DEFAULT_PEN=+1p,red.  Per the documentation, the leading + means

[...]

PR #4498, opened 2020-11-29
Pen reset

PaulWessel:

Properly reset pen width, style, color when basemap is called.  Depending on circumstances, past settings may still be the default which messes things up.  This could be due to the use of subplot panels with tags, movie labels, but also psxy/psxyz calls where the pen setting was set outside the gsave/grestore for the meat of the plots, thus lived on.

This fixes the first of the problems in #4497; it would be better to create separate issues for separate problems for tracking and closing purposes.


PaulWessel:
I should add that the 4 failing tests all were failures in master: the panels with tags should have no outline here but the first one had outline.  The gmtlogo example overrode the baseframe pen form the requested gray to solid black, and the cm plot asked for a red east frame but got black.

joa-quim (approved):

PR #4505, opened 2020-11-29
Let grd2xyz -Wa have optional modifier to choose the length unit used

### LINKED ISSUE #4479: Let grd2xyz -Wa save node weigths in other units
Copy from the [forum: How to select unit for area of nodes in grd2xyz -Wa?
](https://forum.generic-mapping-tools.org/t/how-to-select-unit-for-area-of-nodes-in-grd2xyz-wa/1071)

I want to create an histogram from a geographic grid, so I used grd2xyz to extract the z-values.
And also I use the -Wa argument to compute weights equal to the area each node represents. I think that this values are in km2. Is it possible to choose another unit (m2, sq miles, etc)? Should I use gmtmath?

I know that for an histogram and can do it with the weights in km, but since it is a very large file, I want to store the result as integer in binary (probable unsigned short 2-byte integer) and I can’t do it is the units are in km2.

Allow other units t one used, say via **-Wa+ue**for meter^2.

PaulWessel:

We add the optional modifier **+u**_unit_ to **-Wa** (with default unit **k**) so we can control the area units.  The documentation has been updated to explain a bit more that the area of gridline-registered grids have special cases at sides, corners and at poles.  Closes #4479. Perhaps @Esteban82 may want to give it a spin as well?


seisman (approved):
Looks good, but I don't test it.

PR #4519, opened 2020-11-30
Update api.rst

sarzeaud:
Add a trick to dereference API->error when GMT_Create_Data fails.

Fixes #

**Reminders**


PaulWessel (approved):
Also note there is GMT_Error_Message (API) available.

sarzeaud:
I had a case where API->error was not GMT_NOERROR and GMT_Error_Message returned an empty string (GMT_Destroy_Data with a wrong reference). That is why I think it is also important to have access to API->error.

PaulWessel:
Fair enough. Perhaps we should have a mechanism within the API itself to do that (we are trying to hide gmt_private.h and the full GMT stuff (gmt_dev.h)) with the API.

sarzeaud:
I simple trick could be, when it appears that the string that GMT_ERROR_Message should return has 0 length, to return the string "Error code: %d" with the code.

Issue #4518, opened 2020-11-30
API does not produce the same image as script


sarzeaud:
The API can be used to create an image file with exactly the same instructions as a script. However, with or without the A option on the ``gmt begin`` line, the figure is not cropped when ``gmt end show`` is called.

Take this simple script:

<pre><code>
gmt begin toto_script jpg
gmt basemap -BWESN -Bxa30mg30m -Bya20mg20m -JM7.27/42.27/16.25c -R5.5/41.425/9.0/43.1r
gmt end show
</code></pre>

resulting in this image:

![toto_script](https://user-images.githubusercontent.com/44192622/100641501-d5e98180-3337-11eb-915f-61ff6c4da213.jpg)

The same instructions via the API:

<pre><code>
int main () {
void *API;

/* Initialize the GMT session */
if ((API = GMT_Create_Session ("GMT_plot", 2, 0, NULL)) == NULL)
   return EXIT_FAILURE;
GMT_Call_Module(API, "begin", GMT_MODULE_CMD, "toto_api jpg");
GMT_Call_Module(API, "basemap", GMT_MODULE_CMD, "-BWESN -Bxa30mg30m -Bya20mg20m -JM7.27/42.27/16.25c -R5.5/41.425/9.0/43.1r");
GMT_Call_Module(API, "end", GMT_MODULE_CMD, "show");
/* Destroy session */
if (GMT_Destroy_Session (API)) 
  return EXIT_FAILURE;
}
</code></pre>

produce a different image, not cropped:

![toto_api](https://user-images.githubusercontent.com/44192622/100641495-d550eb00-3337-11eb-9214-ec62ce570c6f.jpg)

Running ``gmt end show -Vd`` reveals the difference between script and API processing:

From script:
`     end [DEBUG]: psconvert: '/home/sarzeaud/.gmt/sessions/gmt_session.14414/gmt_0.ps-' -Tj -Ftoto -A`
From C program (-A option is missing):
`     end [DEBUG]: psconvert: '/home/sarzeaud/.gmt/sessions/gmt_session.14414/gmt_0.ps-' -Tj -Ftoto`

The result is the same whether or not we use the A option on the ``gmt begin`` line.

**System information**

* Operating system: Debian stretch
* GMT version (`gmt --version`): 6.2.0_d266c30_2020.11.25

PaulWessel:
[...]

PR #4525, opened 2020-12-01
Enforce mode setting from GMT_Create_Session in API-driven programs

PaulWessel:

See #4518 for background, as reported by @sarzeaud.

The key problem was that when modules are not driven by gmt.c (i.e., command-line calls), the single _GMT_Create_Session_ call must set the type of the session.  Pass **GMT_SESSION_NORMAL** and you create a classic session, and pass **GMT_SESSION_RUNMODE** to select modern mode.  Apart from the wrappers (MATLAB, Julia, Python), stand-alone C/C++ programs must follow that rule.  Now, sessions started as classic cannot later call modern-mode only modules such as **begin**, **figure**, **subplot**, **inset**, and **end**.

I have added a new testapi_map.c program derived from the example in #4518.  By default it creates a PostScript plot and the test for this passes, but if **testapi_map** is run manually on the command line with a valid graphics extension (e.g., testapi_map jpg) then it will now produce the cropped JPG missing from the original posting.

I suggest @joa-quim and @seisman test this branch in GMT.jl and PyGMT before approving.


sarzeaud:
Hi Paul,
I have just tested your patch here with the GMT_SESSION_RUNMODE flag. It works BUT options that could follow after the file extension are not taken into account. I have tested with A+m0.2c to have a thin margin around the crop but it is not done.
Can you imagine a way to do that, or must I forget it?

PaulWessel:
Will have a look. GMT modern mode was invented way after the API was designed.

joa-quim:
It looks a bit scary but didn't affect GMT.jl

PaulWessel:
It is a bit tricky with starting the modern mode at the GMT_Create_Session in this case, since that means the session is created before the args to gmt begin is processed.  We need to rethink what role the mode has in the use of the API.

sarzeaud:
Do you plan to merge this branch into master before 6.2?

PaulWessel:
I am not sure as I am not happy with it overall.  I think I want to revisit this but I am juggling several other branches with the team right now. Still want to fix this one before 6.2 though but only 24 hours per day, still.

sarzeaud:
Thank you. Tell me if I can help.

PR #4527, opened 2020-12-01
Added the possibility to read variables in NetCDF-4 groups

remkos:

This is to introduce the possibility to read variables from netCDF-4 files that contain variables in groups.
For example, a group "ku" with variables "time", "latitude", and "longitude".

This required introducing "," as an optional separator for variables on the command line. Because, otherwise, the "/'" separator for the variables on the command line would be seen as part of the group/variable naming convention.
 
Example 1: In case of a traditional file without groups (NetCDF-3 or NetCDF-4): file.nc?time/latitude/longitude and file.nc?time,latitude,longitude are equivalent, and will both work.

Example 2: In case we have a group "ku".
* file.nc?ku/time will first check if the group 'ku' exists; if so, then open group 'ku' to find 'time'.
* file.nc?ku/time,ku/latitude,ku/longitude will automatically assume that there is a group 'ku'.

This has been tested with success on both files with and without groups, reading either single or multiple variables, in one group or spread over multiple groups.

Fixes #

**Reminders**

If approved, I will add examples to documentation and tests.

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -3906,7 +3906,7 @@ GMT_LOCAL void * gmtio_nc_input (struct GMT_CTRL *GMT, FILE *fp, uint64_t *n, in
 			}
 			for (k = 0; k < GMT->current.io.count[v][1]; ++col, ++k) {	/* For each column in variable v [typically 1 unless 2-D array] */
 				start[1] = k, count[1] = 1;	/* Read only k'th column in 2-D array or the only column [k = 0] in 1-D array */
-				nc_get_vara_double (GMT->current.io.ncid, GMT->current.io.varid[v], start, count, GMT->hidden.mem_coord[col]);	/* Read column */
+				nc_get_vara_double (GMT->current.io.grpid[v], GMT->current.io.varid[v], start, count, GMT->hidden.mem_coord[col]);	/* Read column */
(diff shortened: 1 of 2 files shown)


PaulWessel (changes requested):
Hi @remkos, I first thought this had to do with grids but I see this is netcdf tables.  I installed the branch and ran the tests and it fails to pass gaps_time.sh which deals with netcdf tables:

```
gmtmath (gmtio_nc_fopen): NetCDF: No group found. [gtec_tx_daily.nc]
gmtmath [ERROR]: Cannot open file gtec_tx_daily.nc?time/gtec
gmtmath (gmtio_nc_fopen): NetCDF: No group found. [gtec_tx_daily.nc]
gmtmath [ERROR]: Cannot open file gtec_tx_daily.nc?time/gtec
psxy [WARNING]: File <stdin> is empty!
```

Perhaps you can sort out what goes wrong there since there are no groups I think.
Also, there probably needs to be some updated in the [docs](https://docs.generic-mapping-tools.org/dev/cookbook/file-formats.html#table-data).

remkos:
Indeed. This applies (for the time being) only to tables. I will look into grids later, though I do not expect groups to be that common there.
The above commit should fix the bug. Somehow I did not test it properly.

PaulWessel (approved):
All tests pass for me, nice.  Please add docs and example as you see fit.

PR #4534, opened 2020-12-03
Need -J if specifying an oblique region

### LINKED ISSUE #4526: Problems cutting grid with -R[..]+r
```
gmt grdcut -R-5/53/90/81+r @earth_relief_02m -Gtest.nc
```
This does not create ``test.nc``. If I omit the ``-G`` I don’t get any complains, which is strange since ``-G`` is required.

This should be possible to do, per https://docs.generic-mapping-tools.org/latest/gmt.html#r-full.

PaulWessel:
The error message was mislabeld as DEBUG and hence did not show without -Vd.  Closes #4526.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -14411,7 +14411,7 @@ struct GMT_CTRL *gmt_init_module (struct GMTAPI_CTRL *API, const char *lib_name,
 					API->GMT->hidden.func_level = level;	/* Reset to what it should be */
 					if (GMT->common.R.oblique) {	/* Must do gmt_mapsetup here to get correct region for building tiles */
 						if (!opt_J) {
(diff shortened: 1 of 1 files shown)


seisman (approved):

joa-quim:
It doesn't look very intuitive to me that one need a projection to select a region provide the the two opposite corners of a rectangle *vs* not needing it the rectangle is given in its more common form.

PaulWessel:
That is true, but this is how we do it throughout GMT when an oblique form of -R is used.  I am not sure if issuing a warning and setting -JQ is a good solution or not but it would "fix" the case if we read the user's mind correctly and they just used that -R form to set corners of a rectangle that happens to align with meridians and parallels.

seisman:
Again, please ignore the CI failures if your local build passes.

PaulWessel:
I will merge this on since consistent with other treatment.  If we are to add code that second-guess the users intentions then that is a separate issue I think.

PR #4541, opened 2020-12-04
Add some more debug statements for cube

PaulWessel:
So we can see what is going on.  Maybe @seisman can run this with -Vd on gspline_4.sh again?


seisman (approved):
We can merge it into master branch, then I can start the tests.

PR #4543, opened 2020-12-04
Be able to exclude tests that are not fail on all platforms

### LINKED ISSUE #1121: Add OS-specific GMT_KNOWN_FAILURE flag
Some tests are known to pass on one OS but fail on others. It would be good if we can add a OS-specific flag, like:

```
# GMT_KNOWN_FAILURE_ON_WINDOWS
```

seisman:

Some tests are known to fail, and we have the `GMT_KNOWN_FAILURE` flag to exclude those tests.

A few tests pass on macOS, but fail on Linux or Windows. For example, `gmtmex/WL_example_3.sh` fails on Linux and Windows, and `grdvector/bothg.sh` fails on Windows.

This PR allows us to skip a few tests on specific platforms, by adding flags `GMT_KNOWN_FAILURE_WINDOW`, `GMT_KNOWN_FAILURE_LINUX`, and/or `GMT_KNOWN_FAILURE_MACOS`, so that we can make all CI jobs green.

**Please note that these flags only work for CI and have no effects for local tests.**

I also add these new flags to some long-standing failures. Now the only two failures are gspline_4.sh and gdal_nn.sh.

Closes #1121.

**Reminders**

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -57,7 +57,7 @@ jobs:
 
           - name: Linux - Build docs + Deploy docs + Package
             os: ubuntu-latest
-            run_in_pr : true
+            run_in_pr : false
             BUILD_DOCS  : true
             DEPLOY_DOCS : true
             PACKAGE     : true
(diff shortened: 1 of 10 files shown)


seisman:
Ping @PaulWessel

PaulWessel (approved):
Sorry forgot, looks good.

PR #4545, opened 2020-12-05
Fix the infamous gdal_nn test crash/failures

joa-quim:


PaulWessel (changes requested):
Add free Ctrol->fname_in in your Free_ function as per the email.

PaulWessel (approved):

Issue #4500, opened 2020-11-29
Have a distinct color set to use in plot (and themes)


joa-quim:
Like in Matlab and others, it would be useful to have a set of distinctive colors to be used in a multi-lines plot. Something like a
```
COLOR_SET=green,darkgreen,lightgreen,moregreen,etc
```
which could be referred as **-W1,color1**, **-W2,color2**, etc or even automatically via some clever trickery in modern mode.

PaulWessel:
Could they come from the categorical CPTs? But yes, I can see a user want to control which colors rather than whatever the CPT has.  So perhaps

**COLOR_SET**=_categorical-cpt_|_list-of-colors_

with a GMT defaults of categorical(.cpt).  As for usage, this is presumably (like in matlab) used when you give many lines and just want them different.  So perhaps -Wpenwidth,auto, where the auto-keyword means substitude the ordered list from COLOR_SET as we cycle through the segments. I can see an issue with segment versus table (file) though.  Let's say your "red" line is broken into several segments due to data gaps, yet you still want all of those segments red.  We could do the above on a per table or per segment, or perhaps need to handle this with -wpenwidth,auto[-segment|table] with default to table?

joa-quim:
*=categorical-cpt* would complicate things because user would have to have a categorical cpt to make it work. I think it's preferable to have it simple like (colors used by Matlab)

**COLOR_SET**=#0072BD,#D95319,#EDB120,#7E2F8E,#77AC30,#4DBEEE,#A2142F

the *auto* is good for modern but wont work for tables in classic, but it would for segments.

PaulWessel:
User already has categorical.cpt because it is one of the GMT master cpt - hence it is suitable default.  But we can discuss what is a suitable default here.

Not sure I follow you on the classic/modern point - I see no reason to limit this to modern (except the auto legend stuff).

joa-quim:
I don't want to limit to modern (on the contrary, classic is what I use most of times in the Julia post-modern) but on classic *auto* cannot work because it doesn't know what color was already used in a previous plot, and setting it in gmt.history does not seem a good idea.

PaulWessel:
I see, so you want this to go across many gmt calls to plot? Why not do them all at once?  That is when the auto-color is handy.  If you do them separately then might as well do -W separately as well.

joa-quim:
I use to have the ideas in the shower but this time I had it half-asleep on way to bed.
[...]

PR #4547, opened 2020-12-06
Sequential auto-colors for polygon fills or line pens in plot and plot3d

PaulWessel:

This PR is the initial implementation of the ideas discussed in #4500.  So far, this has been implemented and should be tested by @joa-quim who requested it:

- A new GMT defaults called **COLOR_SET** [categorical].  It takes either a comma-separated list of colors or the name of a categorical CPT.
- Ability to specify a color fill via **-G** or **-W** by giving the color as _auto_[-_segment_] or _auto-table_ in **plot**.  The first will increment the internal color sequence ID for each input segment while the second will increment the ID for each input table.  In either case we cycle around if we run out of colors.
- If a flavor of auto is selected, the **COLOR_SET** argument is read in as a CPT and we pick colors from it cyclically.
- The initial transparency is kept, so **-G**_auto_@75 retains 75% transparency as the colors cycle.
- This works for both classic and modern mode.

So far, this is restricted to a single process (i.e., the ID starts over at 0).  The plan is to allow auto to be able to continue the ID across several processes (but will be reset at start of each subset panel, for instance.  We will also add a new entry to the auto-legend if **-l** is used.

After building from this branch you can try

`gmt plot @fz_07.txt -R-60/20/-30/30 -JM6i -B -W0.25p,auto -pdf map`

or

`gmt plot @fz_07.txt -R-60/20/-30/30 -JM6i -B -W0.25p,auto -pdf map --COLOR_SET=red,green,blue`

Sorry, I don't have a handy command for **-G**_auto_ but I did test it on some dummy polygons.

Some questions based on this implementation:

1. Should **-G** with no argument mean **-G**_auto_ or is that confusing?
2. If yes, should **-W**_width_,[,_style_] also mean color = auto or is that confusing (e.g., -W0.25p, or -W0.5,,dashed)?
3. I decided to set the color and increment the ID at the top of the loop.  If the segment is entirely outside the region then that color will not show up.  I think this makes sense if you are plotting 6 lines and on one plot segment 4 is outside and not drawn - you don't want segment 5 to now get segment 4's color, right?

Any comments on the current implementation?  I have not touched the docs yet (except for adding **COLOR_SET** to gmt.conf.rst).

### DIFF
--- a/doc/rst/source/plot3d_common.rst_
+++ b/doc/rst/source/plot3d_common.rst_
@@ -88,7 +88,9 @@ Optional Arguments
     Note that the module will search for **-G** and **-W** strings in all the
     segment headers and let any values thus found over-ride the command line settings.
     If **-Z** is set, use **-G+z** to assign fill color via **-C**\ *cpt* and the
-    *z*-values obtained.
+    *z*-values obtained.  Finally, if *fill* = *auto*\ [*-segment*] or *auto=table* then
+    we will cycle through the fill colors implied by :term:`COLOR_SET` and change on a per-segment
+    or per-table basis.  Any *transparency* setting is unchanged.
 
 .. _-I:
 
@@ -154,7 +156,9 @@ Optional Arguments
     **-C**). If instead modifier **+cf** is appended then the color from the cpt
     file is applied to symbol fill.  Use just **+c** for both effects.
     If **-Z** is set, then append **+z** to **-W** to assign pen color via **-C**\ *cpt* and the
-    *z*-values obtained.
+    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto=table* then
+    we will cycle through the pen colors implied by :term:`COLOR_SET` and change on a per-segment
+    or per-table basis.  The *width*, *style*, or *transparency* settings are unchanged.
 
 .. _-X:
 
--- a/doc/rst/source/plot3d_common.rst_
+++ b/doc/rst/source/plot3d_common.rst_
@@ -88,7 +88,7 @@ Optional Arguments
     Note that the module will search for **-G** and **-W** strings in all the
     segment headers and let any values thus found over-ride the command line settings.
     If **-Z** is set, use **-G+z** to assign fill color via **-C**\ *cpt* and the
-    *z*-values obtained.  Finally, if *fill* = *auto*\ [*-segment*] or *auto=table* then
+    *z*-values obtained.  Finally, if *fill* = *auto*\ [*-segment*] or *auto-table* then
     we will cycle through the fill colors implied by :term:`COLOR_SET` and change on a per-segment
     or per-table basis.  Any *transparency* setting is unchanged.
 
--- a/doc/rst/source/plot3d_common.rst_
+++ b/doc/rst/source/plot3d_common.rst_
@@ -194,6 +194,9 @@ Optional Arguments
 
 .. include:: explain_-icols.rst_
 
+.. |Add_-l| unicode:: 0x20 .. just an invisible code
+.. include:: explain_-l.rst_
+
 .. |Add_perspective| unicode:: 0x20 .. just an invisible code
 .. include:: explain_perspective.rst_
 
--- a/doc/rst/source/plot3d_common.rst_
+++ b/doc/rst/source/plot3d_common.rst_
@@ -156,7 +156,7 @@ Optional Arguments
     **-C**). If instead modifier **+cf** is appended then the color from the cpt
     file is applied to symbol fill.  Use just **+c** for both effects.
     If **-Z** is set, then append **+z** to **-W** to assign pen color via **-C**\ *cpt* and the
-    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto=table* then
+    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto-table* then
     we will cycle through the pen colors implied by :term:`COLOR_SET` and change on a per-segment
     or per-table basis.  The *width*, *style*, or *transparency* settings are unchanged.
 
--- a/doc/rst/source/plot_common.rst_
+++ b/doc/rst/source/plot_common.rst_
@@ -126,7 +126,7 @@ Optional Arguments
     Note that this module will search for **-G** and **-W** strings in all the
     segment headers and let any values thus found over-ride the command line settings.
     If **-Z** is set, use **-G+z** to assign fill color via **-C**\ *cpt* and the
-    *z*-values obtained. Finally, if *fill* = *auto*\ [*-segment*] or *auto=table* then
+    *z*-values obtained. Finally, if *fill* = *auto*\ [*-segment*] or *auto-table* then
     we will cycle through the fill colors implied by :term:`COLOR_SET` and change on a per-segment
     or per-table basis.  Any *transparency* setting is unchanged.
 
--- a/doc/rst/source/plot_common.rst_
+++ b/doc/rst/source/plot_common.rst_
@@ -126,7 +126,9 @@ Optional Arguments
     Note that this module will search for **-G** and **-W** strings in all the
     segment headers and let any values thus found over-ride the command line settings.
     If **-Z** is set, use **-G+z** to assign fill color via **-C**\ *cpt* and the
-    *z*-values obtained.
+    *z*-values obtained. Finally, if *fill* = *auto*\ [*-segment*] or *auto=table* then
+    we will cycle through the fill colors implied by :term:`COLOR_SET` and change on a per-segment
+    or per-table basis.  Any *transparency* setting is unchanged.
 
 .. _-I:
 
@@ -199,7 +201,9 @@ Optional Arguments
     at the end of the pen specification.
     See the `Vector Attributes`_ for more information.
     If **-Z** is set, then append **+z** to **-W** to assign pen color via **-C**\ *cpt* and the
-    *z*-values obtained.
+    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto=table* then
+    we will cycle through the pen colors implied by :term:`COLOR_SET` and change on a per-segment
+    or per-table basis.  The *width*, *style*, or *transparency* settings are unchanged.
 
 .. _-X:
 
--- a/doc/rst/source/plot_common.rst_
+++ b/doc/rst/source/plot_common.rst_
@@ -201,7 +201,7 @@ Optional Arguments
     at the end of the pen specification.
     See the `Vector Attributes`_ for more information.
     If **-Z** is set, then append **+z** to **-W** to assign pen color via **-C**\ *cpt* and the
-    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto=table* then
+    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto-table* then
     we will cycle through the pen colors implied by :term:`COLOR_SET` and change on a per-segment
     or per-table basis.  The *width*, *style*, or *transparency* settings are unchanged.
 
--- a/doc/rst/source/explain_-l_full.rst_
+++ b/doc/rst/source/explain_-l_full.rst_
@@ -28,4 +28,7 @@
     just subdivides the available width among the specified columns.
     The upper-case modifiers reflect legend codes described in :doc:`legend`, which provide more details and
     customization.  If **legend** is not called explicitly we will call it implicitly when finishing the plot
-    via :doc:`end`.
+    via :doc:`end`.  **Note**: If auto-coloring is used for pens or fills and **-l** is set then *label* may
+    contain a C-format for integers (e.g., %3.3d) or just # and we will use the sequence number
+    with the format to build the label entries.  Alternatively, give a list of comma-separated labels, or give
+    no label if your segment headers contain label settings.
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -89,6 +89,9 @@ COLOR Parameters
     **COLOR_NAN**
         Color used for the non-defined areas of images (i.e., where z = NaN) [127.5].
 
+    **COLOR_SET**
+        Default comma-separated list of colors (or a CPT name) for automatic, sequential color assignments [categorical].
+
 .. _DIR Parameters:
 
 DIR Parameters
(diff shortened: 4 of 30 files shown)


PaulWessel:
Update on implementation:

1. Extended it to **psxyz** as well
2. Now sequential ID survives the session via gmt.history
3. Added two basic tests that demonstrate it works for lines and fills

Please test, @joa-quim.

PaulWessel:
Documentation updated and test added to demonstrate that subplot panels reset the sequential ID.

PaulWessel:
Question for @joa-quim: Should -W0.5p,auto-table -Gauto@50 be allowed: Polygon fill will change per segment while polygon outline will change per table?

joa-quim:
Ui, many questions.

> should -W0.5p,auto-table -Gauto@50 be allowed: Polygon fill will change per segment while polygon outline will change per table?

For simplicity, yes. User asked for it and this seems a potentially very confusing case. Several tables, with multi-segments and an *auto* scheme. Please know what you are doing.

> We plot max n_colors items. Plotting 300 lines with 4 colors will give 4 legend entries.

Yes, if user gets surprised, use more colors in the *auto* scheme.

> Should -G with no argument mean -Gauto or is that confusing?

I'm leaning to yes. Again, if surprises, RTFM.

> If yes, should -Wwidth,[,style] also mean color = auto or is that confusing (e.g., -W0.25p, or -W0.5,,dashed)?

That would be a breaking change, no? The default line color has been black since the black hole times.

> If the segment is entirely outside the region then that color will not show up. I think this makes sense if you are plotting 6 lines and on one plot segment 4 is outside and not drawn - you don't want segment 5 to now get segment 4's color, right?

Hmm, I think it should be the other way. Users are not supposed to know the color sequence, just that there is one. It makes little sense to me to jump one color only because a certain segment/table does not fit into region. What ever condition leads to the plot and auto colors are used we are expecting that the order is maintained.

And to end with, **COLOR_SET** *[categorical]* is not best default. The idea of the color set is to have defined a default color set with a *best* distinctive colors to have in a *xy* plot and *categorical* is not that (first 3 colors are *red, green, blue*. My proposal is to use the same as Matlab. *i.e.*: ``#0072BD,#D95319,#EDB120,#7E2F8E,#77AC30,#4DBEEE,#A2142F``

joa-quim:
And for testing. The two commands you showed work fine but this one creates a corrupted PS
```
gmt psxy -JX12c/0 -Baf -BWSen -R-9/-6.7/39.6/41.24 -Gauto -P lixo.dat > lixo.ps
```
[lixo.zip](https://github.com/GenericMappingTools/gmt/files/5649831/lixo.zip)

PaulWessel:
Please update and try again.  It work for me ,but since I now switched to your MATLAB color set I get a memory free error from the CPT made from a list.  But the PS seems oK.

joa-quim:
Nope, still get a bad PS
[lixo.ps.zip](https://github.com/GenericMappingTools/gmt/files/5649881/lixo.ps.zip)

PaulWessel:
Yes, yours is but mine is not.  Arg. Anyway, working on labels now so will look at the whole thing when done.  I assume your psxy command has no issues if you use -Gpink instead, right?

joa-quim:
No, no problems if I use *-Gred*

PaulWessel:
Not bad on very first test: `gmt plot p.txt -R-1/6/-1/6 -B -Gauto@50 -png map -l"Poly %d"+jTL`

map

PaulWessel:
Working on enhancing **-l**_label_.  How about this:

1. _label_ is a simple string [current behavior so far]
2. _label_ is a format statement for an integer and we add auto index from sequential color hunt [as above]
3. _label_ is a list of labels and we cycle through them until exhausted
4. _label_ is not given but we use maybe **+Lh** to say "pick label from segment header" (same as **grdcontour -A**)

Question regarding 3: What do we use as list separator?  Commas? Slashes?

joa-quim:
``"Poly %d"`` Could it be ``"Poly #d"``?
Remember your own words of the gray hair C programmers.

PaulWessel:
It could, but then I may want "Polygon-%3.3" and I dont want to implement `#3.3d` etc.  But perhaps we could tolerate "Polygon #" and let that be the same as "Polygon %d". I dont think the #d adds anything.

PaulWessel:
Hm, already have +L for legend labels.  Need a flag for a variable header taken from segment header.  Maybe it is good enough to not give any header and if there is a segment header then we use it?

joa-quim:
> I dont think the #d adds anything.

Right, that's silly because it keeps the ``d``, but the ``#`` alone is what meant

joa-quim:
Using a segment header without explicitly asking for it seems calling for surprises.

PaulWessel:
OK, but is that worse than giving no labels at all, just 4 colored lines? You can always give "Line #" but if you get nothing and you have segment headers, what makes sense then?
I will test this for now - we can change it if we think of a better scheme.

PaulWessel:
List, C-format, and `#` works, e.g.

`gmt plot p.txt -R-1/6/-1/6 -B -Gauto@50 -png map -l"Area 51,Pentagon,University of Hawaii,Jail"+jTL`

map

joa-quim:
Well, I said *without explicitly asking for it*. If user requests it, then OK. Segment headers easily make long strings for labels.

PaulWessel:
Sorry, not segment headers (the whole thing) but segment _labels_, e.g.

`> Some shit here, -Z33 -L"My label"`

would use "My label" if -L is there.

PaulWessel:
This also ties into shapefiles.

joa-quim:
Ah, that's OK to have be default.

PaulWessel:
Implemented the legend labeling for auto-color sequences, added new tests, including classic carry-over with the sequence.  Only think not implemented (that I can remember) is the mechanism to reset the sequence in classic mode.

PaulWessel:
Also @joa-quim , please try the lixo.dat plygons that gave bad PS earlier.  I hope to have fixed it by initializing more variables in psxy.

joa-quim:
The polygons case is working now.

PaulWessel:
I think this PR is ready for serious scrutiny and approval.  I have added 6 test to demonstrate new capabilities.

**1. autocolorline:** Ensure that the color sequence resets per panel.
autocolorline

**2. autocolorliines:** Testing the plotting of lots of segments with default and custom **COLOR_SET**.
autocolorlines

**3. autocolorpols**: Same for a few fake polygons.
autocolorpols

**4. autocolorpolsclassic**: Similar as above but across two **psxy** calls and the color sequence is continued.  Because the sequence information is placed in gmt.history this is automatically limited for modern mode but for classic we may (possibly) need a trigger to reset the sequence.
autocolorpolsclassic

**5. autolegendcolor:** 2x2 subplot of same fake color polygons with 4 ways to automate legend entries.
autolegendcolor

**6. autolegendcolorL:** Same for lines.
autolegendcolorL

PaulWessel:
Pinging @GenericMappingTools/core and @GenericMappingTools/gmt-contributors to examine this new feature.

meghanrjones:
Nice, I will try this out today.

joa-quim:
The **autocolorpolsclassic** plots from a multi-segment only. What is it that needs to be stored in *history*?
I see that *history* has these two guys. Is it it?
```
@L	2
@S	2 0
```

PaulWessel:
> The **autocolorpolsclassic** plots from a multi-segment only. What is it that needs to be stored in _history_?

You may recall you were concerned that in classic we would not be able to add to a plot and stay in sequence.  The only way to continue the sequence across processes is to pass that info via gmt.history.  For modern those are self-limiting but in classic there is just one.

> I see that _history_ has these two guys. Is it it?
> ```
> @L	2
> @S	2 0
> ```

Just the S, the L has to do with PostScript layer number I think.

joa-quim:
Yes, but that was for when multiple tables were used in a plot. With a single table color should always start from first. As is if one runs **autocolorpolsclassic** multiple times we get three different plots (because there are only 3 cyclic colors).

And I think that even if in the same PS each new plot should start from first color, but here we would have to trust in -X -Y to tell us when that happens (like you said at the start of this discussion).

PaulWessel:
I thought I had implemented what you wanted, but can change.  Seems you do not want any connection between modules if in classic mode, so only use gmt.history.* for modern mode to do this, and in any case reset counters to 0 0 if -X -Y is given nonzero values?

joa-quim:
I guess I'm expecting it to behave like in Matlab. When we start a new figure, colors always start from first. More than one module can contribute to a figure but new figure new life. Otherwise we risk current situation where second (and third ...) script runs give different results.

PaulWessel:
OK.  So starting new figure in modern is automatic reset.  But anyone who thinks classic (even in modern) can use -X -Y to move around and make many figures in the same session.  That is why the only scheme I can think of is detecting -X -Y changes.  So OK with the idea that is the color ID sequences is not 0 0 and -X and/or -Y indicate  shift, we reset to 0 0 then?

PaulWessel:
Specifically, **psxy** and **psxyz** will now call _gmt_init_next_color_ (GMT) to determine if we should start over at 0,0:

```
void gmt_init_next_color (struct GMT_CTRL *GMT) {
	/* Reset the sequential color IDs if starting a new plot or we detect overlay shift via -X -Y */
	bool reset = false;
	if (!GMT->common.O.active)	/* Start of a new plot means reset counters read from history to 0 0 */
		reset = true;
	else if (fabs (GMT->common.X.off) > 0.0 || fabs (GMT->common.Y.off) > 0.0)	/* Overlay but we are moving focus */
		reset = true;
	if (reset)
		GMT->current.plot.color_seq_id[0] = GMT->current.plot.color_seq_id[1] = 0;
}
```

OK for @joa-quim ?

joa-quim:
Sure, looks fine.

joa-quim:
But I have still another issue. The default is now
```
COLOR_SET                      = #0072BD,#D95319,#EDB120,#7E2F8E,#77AC30,#4DBEEE,#A2142F
```
but when a script uses 
```
COLOR_SET red,green,blue
```
that is saved in gmt.conf. And how do we change it back to the default? Sure, the user can do (or delete ``gmt.conf``)
```
COLOR_SET #0072BD,#D95319,#EDB120,#7E2F8E,#77AC30,#4DBEEE,#A2142F
```
but we cannot expect that users know that. Perhaps have the default be ``COLOR_SET = default`` and internally ``default`` be set to those colors. Then resetting the default would be easy.

PaulWessel:
Yes, I can add a COLOR_SET = default option.

PaulWessel:
Will update the autocolorpolsclassic PS shortly.

joa-quim (approved):

meghanrjones:
It seems like this feature is mentioned in the -G option of plot and plot3d without needing to see the full documentation for fill attributes. I think this is a good choice for making users aware of the auto option. Do you think it's worthwhile also adding it to the docs for the psxy -G option?

PaulWessel:
I think it is there, no? Both psxy.rst and plot.rst include plot_common.rst where is where **-G** is mentioned.  Or did you mean something else?

meghanrjones:
Yes, that is what I meant. I must have made an error in building the docs, because I tried to test using ```gmt docs psxy``` and did not see it.

meghanrjones:
Does test/psxy/autocolorpolsclassic.ps need to get updated for these latest features?

PaulWessel:
I just built the docs with bdoc and then build again with bd to place files so **gmt docs psxy** loads the right things.  I see the **-G** stuff.
And I thought I updated _autocolorpolsclassic.ps_ after making the change, the tests work for me. One git pull behind?

meghanrjones (approved):

PaulWessel:
Great, any concerns with this one, @seisman ?

meghanrjones:
Yup...not my night apparently.

doc/rst/source/plot3d_common.rst_
```diff
@@ -88,7 +88,9 @@ Optional Arguments
     Note that the module will search for **-G** and **-W** strings in all the
     segment headers and let any values thus found over-ride the command line settings.
     If **-Z** is set, use **-G+z** to assign fill color via **-C**\ *cpt* and the
-    *z*-values obtained.
+    *z*-values obtained.  Finally, if *fill* = *auto*\ [*-segment*] or *auto=table* then
```
seisman, line 91: Typo?
```suggestion
    *z*-values obtained.  Finally, if *fill* = *auto*\ [*-segment*] or *auto-table* then
```

doc/rst/source/plot3d_common.rst_
```diff
@@ -154,7 +156,9 @@ Optional Arguments
     **-C**). If instead modifier **+cf** is appended then the color from the cpt
     file is applied to symbol fill.  Use just **+c** for both effects.
     If **-Z** is set, then append **+z** to **-W** to assign pen color via **-C**\ *cpt* and the
-    *z*-values obtained.
+    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto=table* then
```
seisman, line 159: Typo?
```suggestion
    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto-table* then
```

doc/rst/source/plot_common.rst_
```diff
@@ -126,7 +126,9 @@ Optional Arguments
     Note that this module will search for **-G** and **-W** strings in all the
     segment headers and let any values thus found over-ride the command line settings.
     If **-Z** is set, use **-G+z** to assign fill color via **-C**\ *cpt* and the
-    *z*-values obtained.
+    *z*-values obtained. Finally, if *fill* = *auto*\ [*-segment*] or *auto=table* then
```
seisman, line 129:
```suggestion
    *z*-values obtained. Finally, if *fill* = *auto*\ [*-segment*] or *auto-table* then
```

doc/rst/source/plot_common.rst_
```diff
@@ -199,7 +201,9 @@ Optional Arguments
     at the end of the pen specification.
     See the `Vector Attributes`_ for more information.
     If **-Z** is set, then append **+z** to **-W** to assign pen color via **-C**\ *cpt* and the
-    *z*-values obtained.
+    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto=table* then
```
seisman, line 204:
```suggestion
    *z*-values obtained.  Finally, if pen *color* = *auto*\ [*-segment*] or *auto-table* then
```

seisman:
Tests pass on macOS, but fail on Linux.

**autolegendcolor**:

image

**autolegendcolorL**

image

PaulWessel:
That is the one which has -l"Line A,Line B,Line C,N/A"+jBR, which then is passed to the new function _gmt_get_word_ that returns the k'th word from the comma-separated list.  Wonder why it fails for Linux, but it is a brand new function so...

PR #4548, opened 2020-12-06
Fix wrong entries in OpenMP pragmas

PaulWessel:
Hoping this is the cause of failures in Linux and Windows since they presumably are running with OpenMP.  Also rename badly named local functions...  let me know if this does the trick, @seisman.

### DIFF
--- a/src/greenspline.c
+++ b/src/greenspline.c
@@ -2380,14 +2380,14 @@ EXTERN_MSC int GMT_greenspline (void *V_API, int mode, void *args) {
 	else {	/* Output on equidistant lattice */
 		uint64_t nz_off, nxy;
 		unsigned int layer, wmode = GMT_ADD_DEFAULT;
-		double *xp = NULL, *yp = NULL, wp, V[4];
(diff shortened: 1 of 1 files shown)


seisman:
Still fails.

PaulWessel:
OK, but is it true the CI uses OpenMP?

seisman:
I think yes. It says:
```
*  OpenMP support             : enabled
```

PaulWessel:
So I think the mistake is to set `V[GMT_Z] = cube->z[layer]` outside the OpenMP loop since _V_ is private and thus duplicated for each thread but I doubt this also magically sets `V[GMT_Z] = cube->z[layer]`.  So _V[GMT_Z]_ is probably zero always which matches the fact that all levels give the same wrong result.  Please try this one, @seisman .

seisman:
Great! The test now passes on all platforms.

seisman (approved):

PaulWessel:
Make a mental note that when a module fails on a platform with OpenMP and passes on macOS we should suspect bad pragmas...not the first time.

PR #4555, opened 2020-12-08
mgd77list: Verify if field names are valid

sarzeaud:

This is the first step to go to a mgd77list which stay successful when the requested fields (-F option) are missing in the file.

Add the function MGD77_Verify_Columns in mgd77.c to check if all the field names given to the -F option are valid.
Add a call to this function at the end of the parse function of mgd77list.

I wonder if we should not add a special "nan" field if the user wants to make room for further computations...

Fixes #

**Reminders**


PaulWessel (approved):
Look good to me.

PR #4560, opened 2020-12-09
Control the message "GMT [ERROR]: GMT_Encode_Options: earthtide cannot set -G when called externally"

joa-quim:
It should not be issued when modules like the blocks or earthtide use G=filename. Only when no arg and the result is expected to return to the memory of external caller.


PaulWessel (approved):
Taking this on faith of course.

joa-quim:
I'm going to merge this because it solves the ``earthtide`` case but the blocks still screw
```
julia> blockmean(d, region=[0 2 0 2], inc=1, grid="aiai.grd", reg=true, S=:n, Vd=1)
        blockmean  -Sn -I1 -Az -Gaiai.grd -R0/2/0/2 -r
ERROR: get_grid: programming error, output matrix is empty
```
Apparently it's waiting for a grid to be returned, but none should be and it actually wrote the grid file. So, it looks a different issue.

Issue #4557, opened 2020-12-08
Plot title with subtitle on a newline


weiji14:
**Description of the desired feature**

There should be a way to plot a title on two lines instead of just one. Carried on from discussion at https://forum.generic-mapping-tools.org/t/how-to-plot-title-with-a-subtitle-on-a-newline/1121

### Current state

Using newline `\n` doesn't work on GMT 6.1.1:

```bash
gmt basemap -R0/1/0/1 -JX1 -Baf+t"Title\nSubtitle" -png map
```

![image](https://user-images.githubusercontent.com/23487320/101553425-7830f580-3a19-11eb-8284-5b03192bc203.png)

### Desired state

The "subtitle" should be on a second line.

It might be nice to have slightly different font sizes for the main title and subtitle. But making the newline `\n` work will do for now :smile:

**Are you willing to help implement and maintain this feature?** No

PaulWessel:
Think you can use @%font%...@%% to change the font to your liking.  I don't think we want a fixed system unless we assume there is at most one \n and it truly is a title/subtitle 1-2 system.  But I do think your suggestion makes it simpler to implement (again, assuming anything after the first \n is a subtitle set at a smaller size, say 0.8*FONT_TITLE or a new FONT_SUBTITLE defaulting to 0.8...

PaulWessel:
Perhaps @joa-quim and @seisman have opinion on the font-setting issue for multiple lines? I am bored and eager to code this up before afternoon coffee.

seisman:
Sometimes people just want to split a long title into two lines. The second line doesn't have to be a subtitle. I believe it will surprise people that the second line has different font size.

joa-quim:
I agree with Dongdong, but easy to have both. Use a double ``\n`` (``\n\n``) to have a subtitle.

PaulWessel:
Yes, I like that, or perhaps \r for continuation of title (same font) and \n for smaller subtitle.  None of those characters are actually used when we typeset the PS so we could just assign those two functions and be done with it (a second \n just means the subtitle continues on more than one line with no further font changes.

joa-quim:
Can we make it ``r`` => ``\continue`` and ``\n`` => ``\subtitle``? (or ``#continue`` if ``\c`` causes troubles)
``\r`` & ``\n`` are very natural to our *C-ish* thinking but we want to go for *longoptions*

PaulWessel:
[...]

PR #4562, opened 2020-12-10
Support for multi-line plot titles and introducing subtitles (also supporting multi-lines)

PaulWessel:

See #4557 for context.  This PR implements two related features:

1. It introduces a new **+s**_subtitle_ modifier to the **-B** frame settings.  The **+s** requires the presence of **+t**.
2. It introduces `@^` (and `#break;`) as markers that break a title or subtitle into multiple lines.

I have added two examples that demonstrates these (_longtitles_hat.sh.sh_ and _longtitles_break.sh_).

Given that `@^` is pretty unique for breaking the line, do we still need `#break;`? Let me know if I can remove that before we merge.

longtitles_hat

If you feel the subtitle could be a tad smaller (it is 20p while the title is 24p) let me know.  Below is what it looks with 24p/18p, which I think looks better personally:

longtitles_hat

Later, I will introduce the same to subplot **-T**_heading_ option as welo, but currently its implementation requires me to make some changes first.


joa-quim:
> Given that @^ is pretty unique for breaking the line, do we still need #break;? 

It's not that I dislike ``@^`` by itself, but I don't like so much creating an escape sequence that it's only to be used in one single case (breaking title strings). So, for me we should only have ``#break`` (though I still prefer ``<break>``).

``24p/18p`` Look better to me.

PaulWessel:
Happy to entertain opinions on this.  I can imagine a near future where @^ could work in any other setting as well (say axes lables), so in that case perhaps @^ is not a bad idea anyway.

PaulWessel:
Now implemented with 18p size.

meghanrjones:
<break> and/or @^ seem fine to me. I agree with Joaquim that <break> is very intuitive.

doc/rst/source/cookbook/options.rst
```diff
@@ -281,13 +281,13 @@ but most examples of its usage are actually quite simple. We distinguish
...
 specifications may be specified. The frame settings covers things such
-as which axes should be plotted, canvas fill, plot title, and what type
+as which axes should be plotted, canvas fill, plot title (and subtitle), and what type
 of gridlines be drawn, whereas the Axes settings deal with annotation,
 tick, and gridline intervals, axes labels, and annotation units.
 
 The Frame settings are specified by
 
--  **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+i**\ [*val*]][**+n**][**+o**\ *lon/lat*][**+t**\ *title*][**+w**\ [*pen*]][**+x**\ *fill*][**+y**\ *fill*][**+z**\ *fill*]
+-  **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+i**\ [*val*]][**+n**][**+o**\ *lon/lat*][**+s**\ *subtitle*][**+t**\ *title*][**+w**\ [*pen*]][**+x**\ *fill*][**+y**\ *fill*][**+z**\ *fill*]
```
meghanrjones, line 290:
```suggestion
-  **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+i**\ [*val*]][**+n**][**+o**\ *lon/lat*][**+t**\ *title*][**+s**\ *subtitle*][**+w**\ [*pen*]][**+x**\ *fill*][**+y**\ *fill*][**+z**\ *fill*]
```
Small change, but I think it makes more sense for **+t** to come first, since it is required for **+s**

doc/rst/source/explain_-B_full.rst_
```diff
@@ -3,7 +3,7 @@
 **-B**\ [**p**\|\ **s**]\ *parameters*
     Set map Frame and Axes parameters. The Frame parameters are specified by
 
-    **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+i**\ [*val*]][**+n**][**+o**\ *lon/lat*][**+t**\ *title*][**+w**\ [*pen*]][**+x**\ *fill*][**+y**\ *fill*][**+z**\ *fill*]
+    **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+i**\ [*val*]][**+n**][**+o**\ *lon/lat*][**+s**\ *subtitle*][**+t**\ *title*][**+w**\ [*pen*]][**+x**\ *fill*][**+y**\ *fill*][**+z**\ *fill*]
```
meghanrjones, line 6:
```suggestion
    **-B**\ [*axes*][**+b**][**+g**\ *fill*][**+i**\ [*val*]][**+n**][**+o**\ *lon/lat*][**+t**\ *title*][**+s**\ *subtitle*][**+w**\ [*pen*]][**+x**\ *fill*][**+y**\ *fill*][**+z**\ *fill*]
```
same change as above

PaulWessel: But the rest is alphabetical

meghanrjones: OK, suggestion deleted

joa-quim:
Ok, so leave ``@^`` and ``#break`` (the idea for longoptions is that they _talk_ and ``@^`` doesn't)

joa-quim (approved):

PaulWessel:
OK, will merge for now as is.  If there were a consensus that `<break>` is a better choice than `#break; `then we can change it before 6.2 release.

PR #4568, opened 2020-12-13
Better check for earthtide and block* grid output selections

PaulWessel:
This should allow externals to use -Gfile to write directly to file.  Hopefully, @joa-quim can check if this works as expected now.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12371,7 +12371,7 @@ struct GMT_RESOURCE * GMT_Encode_Options (void *V_API, const char *module_name, 
 	unsigned int output_pos = 0, input_pos = 0, mod_pos;
 	int family = GMT_NOTSET;	/* -1, or one of GMT_IS_DATASET, GMT_IS_GRID, GMT_IS_PALETTE, GMT_IS_IMAGE */
 	int geometry = GMT_NOTSET;	/* -1, or one of GMT_IS_NONE, GMT_IS_TEXT, GMT_IS_POINT, GMT_IS_LINE, GMT_IS_POLY, GMT_IS_SURFACE */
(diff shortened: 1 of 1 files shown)


joa-quim:
Nope, this doesn't work;
```
earthtide(G="lixo.grd")
```
still returns the Grid object to Julia and doesn't write it to disk.

PaulWessel:
Can you tell me what exactly is the command passed to GMT_Call_Module in this case?

joa-quim:
earthtide  -Cz -Glixo.grd

PaulWessel:
Think this will fix it though.  In your case k remained 1 so it added -G?.

joa-quim:
No, it doesn't. You can check it with Matlab too. This  returns instead of writing
```
gmtmex('earthtide -Cx -Glixo.grd')
```
It does work though if no **-C** is specified

PaulWessel:
OK, and now fell into your typical trap (cooking).  So tomorrow.

PaulWessel:
Working on this now - see the problem and will fix shortly.

PaulWessel:
Problem was that the -C check mean the -G was not done (else if).  Now we do -C, and then we do -G etc.  Now, k ends up being zero and it works for me.  Please give it a spin, @joa-quim.

PaulWessel:
I think we should have **earthtide** fill out the z unit info correctly . Right now it says 

```
lixo.grd: Remark: 
lixo.grd: z_min: -0.0772267207503 z_max: 0.0775824487209 name: z
```

i.e., no information on what the units are (meters?) and no remark that this is x-component of deformation (only inferred via command -Cx) and name probably should be "x/east tie"?  I can do this if you can confirm the units.

joa-quim:
Yes, units are meters.

joa-quim (approved):
OK, my tests pass now. Also for the blocks

PR #4569, opened 2020-12-14
Revamp gmtgravmag3d to use the IO and therefore allow calling from externals.

joa-quim:
Also changed a bit of the syntax but it should be backward compatible, except even older syntax that was removed. Hence the tests update.

### DIFF
--- a/doc/rst/source/supplements/potential/gmtgravmag3d.rst
+++ b/doc/rst/source/supplements/potential/gmtgravmag3d.rst
@@ -12,9 +12,8 @@ Synopsis
 
 .. include:: ../../common_SYN_OPTs.rst_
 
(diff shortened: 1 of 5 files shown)


PaulWessel (approved):

seisman:
@joa-quim @PaulWessel 
This PR introduced a few warnings:
```
src/potential/gmtgravmag3d.c:355:6: error: Uninitialized variable: n_cols [uninitvar]
 if (n_cols < 3 || n_cols == 7 || n_cols > 8) {
     ^
src/potential/gmtgravmag3d.c:348:52: note: Assuming condition is false
 if ((fp = gmt_fopen (GMT, Ctrl->T.xyz_file, "r")) == NULL) return (-1);
                                                   ^
src/potential/gmtgravmag3d.c:355:6: note: Uninitialized variable: n_cols
 if (n_cols < 3 || n_cols == 7 || n_cols > 8) {
```

PR #4570, opened 2020-12-14
Fix inconsistencies in pstext syntax and docs

PaulWessel:
While we do have some backwards-compatible handling deprecated variants like **-Z+** and **-Z**_level_, these should not be mentioned in the synopsis/usage..


src/pstext.c
```diff
@@ -603,11 +603,11 @@ static int parse (struct GMT_CTRL *GMT, struct PSTEXT_CTRL *Ctrl, struct GMT_OPT
 			case 'Z':
 				/* For backward compatibility we will see -Z+ as the current -Z
 				 * and -Z<level> as an alternative to -p<az>/<el>/<level> */
-				if (opt->arg[0] == '+' && !opt->arg[1])
+				if (opt->arg[0] == '+' && !opt->arg[1])	/* Deprecated -Z+ optino */
```
seisman, line 606:
```suggestion
				if (opt->arg[0] == '+' && !opt->arg[1])	/* Deprecated -Z+ option */
```

seisman (approved):

PR #4586, opened 2020-12-18
Update gmtdefaults description

### LINKED ISSUE #4585: Possible gmtdefaults documentation error
The documentation for [gmtdefaults](https://docs.generic-mapping-tools.org/dev/gmtdefaults.html) says:
> (2) in classic mode you may use any text editor to edit the file gmt.conf in your home, ~/.gmt or current directory (if you do not have this file, run gmtset -D to get one with the system default settings)

I think "gmtset -D" should be "gmtdefaults -D > gmt.conf". ```gmtset -D``` outputs a blank gmt.conf file, whereas ```gmtdefaults -D``` writes the system default settings to stdout, which is the stated behavior. If my interpretation is correct, I can submit a PR with the change.

meghanrjones:

Updates gmtdefaults documentation to create gmt.conf using gmtdefaults rather then gmtset. Also specifies that gmtset changes settings for the duration of the current session in modern modern mode.

I could not find an similar example in the docs to determine whether "gmtdefaults -D > gmt.conf" should be formatted as in-line code or not. Please let me know if I should change that. 

Fixes #4585 

**Reminders**

### DIFF
--- a/doc/rst/source/gmtdefaults.rst
+++ b/doc/rst/source/gmtdefaults.rst
@@ -23,13 +23,14 @@ Description
 **-D** is used. There are three ways to change some of the settings: (1)
 Use the command :doc:`gmtset`, (2) in classic mode you may use any text editor to edit the file
 :doc:`gmt.conf` in your home, ~/.gmt or current directory (if you do not
-have this file, run :doc:`gmtset` **-D** to get one with the system default
-settings), or (3) override any parameter by specifying one
+have this file, run :doc:`gmtdefaults` **-D** > gmt.conf to get one with the 
+system default settings), or (3) override any parameter by specifying one
 or more **-**\ **-PARAMETER**\ =\ *VALUE* statements on the command line of any
 GMT command (**PARAMETER** and *VALUE* are any combination listed
-below). The first two options are permanent changes until explicitly
-changed back, while the last option is ephemeral and only applies to the
-single GMT command that received the override. GMT can provide
+below). In classic mode, the first two options are permanent changes until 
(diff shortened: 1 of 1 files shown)


meghanrjones:
One other small inconsistency in the documentation for modules like ```defaults```, ```begin```, and ```end``` is that the description from explain_help.rst_ is not completely accurate because calling the function without no arguments does not give the same result as calling with **-?**. I do not know if it would be worthwhile to fix this with a separate explain_help.rst_ file for those modules.

doc/rst/source/gmtdefaults.rst
```diff
@@ -23,13 +23,14 @@ Description
...
+system default settings), or (3) override any parameter by specifying one
 or more **-**\ **-PARAMETER**\ =\ *VALUE* statements on the command line of any
 GMT command (**PARAMETER** and *VALUE* are any combination listed
-below). The first two options are permanent changes until explicitly
-changed back, while the last option is ephemeral and only applies to the
-single GMT command that received the override. GMT can provide
+below). In classic mode, the first two options are permanent changes until 
+explicitly changed back, while the last option is ephemeral and only applies to 
+the single GMT command that received the override. In modern mode, changes made using
+`gmtset` stay in effect for the duration of the current session. GMT can provide
```
seisman, line 33:
```suggestion
:doc:`gmtset` stay in effect for the duration of the current session. GMT can provide
```

seisman (approved):

seisman:
> calling the function without no arguments does not give the same result as calling with **-?**

They are very tiny issues to me, and only very few people will notice them.

PR #4618, opened 2021-01-02
Save the GMT->common.R structure in a new GMT->hidden.common_R so that it can be accessed by externals

joa-quim:
In order to be able to guess a reasonable projection from the map limits it is necessary to know, for example, what **-RDE** has turned to in terms of lon/lat. Currently I use the trick of asking ``gmt2kml`` the boundingbox only and then parse the output kml. But this is cumbersome and slow.

This PR adds a new struct, ``GMT->common.R`` to ``GMT->hidden`` and a new function ``void *gmtlib_get_common_R (void *V_API)`` that allows easy access to the **R** settings from Julia. Possibly a similar solution for ``common.J`` would be interesting too but I have not yet add a need for it.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -15101,6 +15101,18 @@ EXTERN_MSC void * gmtlib_get_ctrl (void *V_API) {
 	return API->GMT;	/* Pass back the GMT ctrl pointer as void pointer */
 }
 
+EXTERN_MSC void *gmtlib_get_common_R (void *V_API) {
+	/* For external environments that need to access GMT->common.R */
+	struct GMTAPI_CTRL *API = NULL;
+	struct COMMON_R R;
+
+	if (V_API == NULL) return_null (V_API, GMT_NOT_A_SESSION);
+	API = gmtapi_get_api_ptr (V_API);
+	gmt_M_memcpy (&R, &API->GMT->hidden.common_R, 1, struct COMMON_R);
+
+	return &R;	/* Pass back the GMT ctrl pointer as void pointer */
+}
+
 int64_t gmt_eliminate_duplicates (struct GMTAPI_CTRL *API, struct GMT_DATASET *D, uint64_t cols[], uint64_t ncols, bool text) {
 	/* Scan dataset per segment and eliminate any duplicate records as identified by having no change in all the specified cols.
 	 * If no change then we skip the duplicate records.  No segment will be eliminated since first record always survives.
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -15101,16 +15101,15 @@ EXTERN_MSC void * gmtlib_get_ctrl (void *V_API) {
 	return API->GMT;	/* Pass back the GMT ctrl pointer as void pointer */
 }
 
-EXTERN_MSC void *gmtlib_get_common_R (void *V_API) {
+EXTERN_MSC int gmtlib_get_common_R (void *V_API, struct COMMON_R *R) {
 	/* For external environments that need to access GMT->common.R */
 	struct GMTAPI_CTRL *API = NULL;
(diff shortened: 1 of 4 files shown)


PaulWessel (commented):
Not sure about the return &R of a local structure declared inside a function.  Don't you either need to pass in a pointer to a valid structure **or** allocate and return the pointer to one?

joa-quim:
I see your point but it actually worked pretty well with the ``return &R;`` (and was way easier to implement in the Julia side). But made it ``int gmtlib_get_common_R (void *V_API, struct COMMON_R *R)``

PaulWessel:
OK, and just so I understand why you want this: You want to implement the auto-J in Julia even though we will do exactly the same in GMT C?  Would it not be better to help getting that implementation done?

joa-quim:
This is not for auto-J in the sense that Kristof asked. It's for not needing **-J** at all and let it guess from coords only. Also, if implemented in Julia it will available to all GMT >= 6.0 versions and not only future 6.2

joa-quim:
... unless you were willing to make **-J** optional at times. Seems too drastic as a new feature.

PaulWessel (approved):
I will approve this to not be a roadblock for you.  However, I do have serious concerns:

1. Adding these tricks adds more places where we have to be careful with changes as it can break Julia.  Not a good design.
2. The excuse that you can implement something in Julia now (so it works with GMT 6.0) instead of strengthening GMT core so it works for all (GMT plus all externals) is weak and will just lead to repeated work down the road.
3. This is using the Julia wrapper to make changes Julia-specific in GMT core rather than considering generic solutions across all wrappers.  We should look for changes in GMT that benefit all externals.  Your impatience with getting this to work in Julia _today_ severely breaks that idea.
4. We already have GMT_Get_Common which can give you w/e/s/n and that is a cross-all function anyone can use.

PaulWessel:
-J is already optional; it is just that we only have two default choices: -JX or -JQ regardless of -R.

joa-quim:
There is nothing Julia specific in this PR. Only a way store and allow inquiring last *w/e/s/n* which is cleared from the lib memory at the end of a module run (in ``gmt_end_module()``). This can be used by any external.

If ``GMT_Get_Common`` can be make to work then that's even better but as is, it wouldn't work. 
```
if (par) gmt_M_memcpy (par, GMT->common.R.wesn, 4, double);
```
because if consulted after a module finish the ``GMT->common.R.wesn`` would have been wiped out in ``gmt_end_module``. The way to make it work is to ask instead to
```
if (par) gmt_M_memcpy (par, GMT->hidden.common_R.wesn, 4, double);
```

PaulWessel:
I see.  Well, the full -R has lots of loose parts; would you be OK with just betting w/e/s/n via GMT_Get_Common instead? We would place that in the API struct and have _GMT_Get_Common_ return those values.

joa-quim:
Yes, *wesn* and perhaps *inc* would be the only important pieces. But this made me realize that probably many things in ``GMT_Get_Common`` won't work as advertised for the exact same reason. Everything that is reset by ``gmt_end_module`` will not have anything interesting to be exported by ``GMT_Get_Common``. And there is no time to inquire them before ``gmt_end_module`` gets called because all happens inside the call
```
GMT_Call_Module(API, g_module, GMT_MODULE_OPT, LL)
```

PaulWessel:
Yes, that seems true - not tested much of course.  OK, I can imagine a fix to this where we keep a snapshot of the whole common args and report from that instead.

PaulWessel:
In middle of testing another branch but I can implement a fix for the GMT_Get_Common soon.

joa-quim:
Closing as addressed by #4621

PR #4621, opened 2021-01-02
Fix GMT_Get_Common so it works outside a module

PaulWessel:
_GMT_Get_Common_ can report back the current settings of GMT common options like **-R -I**.  However, because modules destroy the GMT struct copy we instead make a snapshot at the end of a module of the full GMT common structure and let _GMT_Get_Common_ report values from it instead.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12920,55 +12920,55 @@ int GMT_Get_Common (void *V_API, unsigned int option, double par[]) {
 	GMT = API->GMT;
 
 	switch (option) {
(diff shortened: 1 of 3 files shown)


joa-quim (approved):
OK, at least for the **-R** it works, though long in [-180 180] is stored in [0 360] for **+r**
```
julia> coast(region="-10/36/-7/41+r");
julia> GMT_Get_Common(GMT.API, 'R')
([348.91442543022674, 353.94138021920253, 35.90000000000003, 41.099999999999994], 4)
```

PaulWessel:
How longs are inside GMT is tricky, but they are unique.

PR #4622, opened 2021-01-04
Add MAP_FRAME_PERCENT to control fancy frame fat pen width

PaulWessel:
See discussion on the [forum](https://forum.generic-mapping-tools.org/t/noaa-nautical-charts-border-style/1156/21).  This PR just allows the fat pen to change via **MAP_FRAME_PERCENT** [100].

`gmt basemap -R0/30/0/30 -JM15c -B -png t --MAP_FRAME_PERCENT=25`

The initial test shows that the cross-frame ticks were not needed when the setting was fixed at 100% but now there seems to be a need for tick lines?  While ticks are drawn for plain frame, no frame ticks are drawn for fancy frames since the ends of the fat lines were good enough.  Accepting comments, @joa-quim and @KristofKoch.

t

### DIFF
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -651,6 +651,10 @@ MAP Parameters
     **MAP_FRAME_PEN**
         Pen attributes used to draw plain map frame [thicker,black].
 
+    **MAP_FRAME_PERCENT**
+        Percentage of the fancy frame width to use for the internal checkerboard
+        frame lines [100].
+
     **MAP_FRAME_TYPE**
         Choose between **inside**, **plain** and **fancy** (thick boundary,
         alternating black/white frame; append **-rounded** for rounded corners)
--- a/src/gmt_defaults.h
+++ b/src/gmt_defaults.h
@@ -133,6 +133,7 @@ struct GMT_DEFAULTS {
 	double map_annot_offset[2];		/* Distance between primary or secondary annotation and tickmarks [5p/5p] */
 	double map_annot_min_angle;		/* If angle between map boundary and annotation is less, no annotation is drawn [20] */
 	double map_annot_min_spacing;	/* If an annotation is closer that this to an older annotation, the annotation is skipped [0.0] */
+	double map_frame_percent;		/* Percentage of fancy map frame width to actually draw [100] */
 	double map_frame_width;			/* Thickness of fancy map frame [5p] */
 	double map_grid_cross_size[2];	/* Size of primary & secondary gridcrosses.  0 means draw continuous gridlines */
 	double map_heading_offset;		/* Distance between top of panel title and base of subplot heading [18p] */
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -215,6 +215,7 @@ static struct GMT_parameter GMT_keyword_active[]= {
 	{ 0, "MAP_DEGREE_SYMBOL"},
 	{ 0, "MAP_FRAME_AXES"},
 	{ 0, "MAP_FRAME_PEN"},
+	{ 0, "MAP_FRAME_PERCENT"},
 	{ 0, "MAP_FRAME_TYPE"},
 	{ 0, "MAP_FRAME_WIDTH"},
 	{ 0, "MAP_GRID_CROSS_SIZE_PRIMARY"},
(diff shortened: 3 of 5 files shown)


KristofKoch:
Hi @PaulWessel thank you for implementing it – I gave it a shot and had two issues with it:

- affects also the outer thinner pen - I expected it to affect only the thick pen
- tickmarks through the width of the frame are interrupted by the stripes – I expected a smooth line from frame border to border

mfp25_2

```
gmt basemap -R0/30/0/30 -JM15c -B -pdf t1 --MAP_FRAME_PERCENT=25
gmt basemap -R0/30/0/30 -JM15c -B -pdf t2
```

Generally ticks are generally desirable to have. For greater flexibility it would be great to be able to control them inside and outside the frame. Ok, I can always overlay a second frame without ticks if I only want outer ticks and none within the frame.

I tried to get ticks in the frame on both ends of the black segments by using `-Ba5f1g1` but `f` seems not supported:

t3

```
gmt basemap -R0/15/0/10 -JM10c -Ba5f1g1 -png t3 --MAP_FRAME_PERCENT=25 --MAP_TICK_LENGTH_PRIMARY=10p/5p
```

KristofKoch:
After giving it some further thought I came up with this:

skinny-fancy-JL


[skinny-fancy-JL.pdf](https://github.com/GenericMappingTools/gmt/files/5764711/skinny-fancy-JL.pdf) (14 KB)

It still isn't a single border and I have to work around some limitations of the current _skinny-fancy_ implementation. But it comes very close to the NOAA navigational chart frame style I'm trying to reproduce.

```
gmt begin skinny-fancy-JL pdf
  gmt basemap -R6.9/9.1/53.8/54.8 -JL8/54.5/54.25/54.75/17c --MAP_FRAME_TYPE=plain -Bp30m -Bs1d --MAP_FRAME_PEN=6.6p,black --MAP_TICK_LENGTH_PRIMARY=7p/3p
  gmt basemap --MAP_FRAME_TYPE=plain -B0 --MAP_FRAME_PEN=6p,white
  gmt basemap --MAP_FRAME_TYPE=fancy -Bf10mg1 --MAP_FRAME_PERCENT=40 --MAP_FRAME_WIDTH=3p --MAP_TICK_LENGTH_PRIMARY=10p/5p
  gmt basemap --MAP_FRAME_TYPE=plain -Bf5m --MAP_TICK_LENGTH_PRIMARY=3p/3p --MAP_FRAME_PEN=0.3p,black
  gmt basemap --MAP_FRAME_TYPE=inside -Bf1m --MAP_TICK_LENGTH_PRIMARY=3p/3p --MAP_FRAME_PEN=0.3p,black
gmt end show
```
GMT version: 6.2.0_434185e_2021.01.03

joa-quim:
The problem with the thin lines is that their thickness are computed after ``fat_pen`` and they need to be compensated for the fat_line thinning.
```
thin_pen = 0.1 * fat_pen;
```
The other problem is due to the fact that there are white stripes being drawn and those eat half the with of the vertical tick.

Also, using ``MAP_FRAME_TYPE=inside`` is no solution because it also *eats* part of the map.

KristofKoch:
> The problem with the thin lines is that their thickness are computed after `fat_pen` and they need to be compensated for the fat_line thinning.
> 
> ```
> thin_pen = 0.1 * fat_pen;
> ```

But this is an easy one, or am I missing something?

```
thin_pen = 0.1 * fat_pen / ( MAP_FRAME_PERCENT * 0.01 );
```

> The other problem is due to the fact that there are white stripes being drawn and those eat half the with of the vertical tick. 

This is what I suspected. Seems to be tied to `-Bf` somehow?

> Also, using MAP_FRAME_TYPE=inside is no solution because it also eats part of the map. 

This is correct and not desired. But it eats only 3.3p into the map as would a plain frame with a 6.6p pen so I deemed it acceptable.

PaulWessel:
OK, thin pen corrected, the other issues remain for now.

KristofKoch:
Awesome - thank you, @PaulWessel! I assume this being a rather niche use case. I'm very grateful you implemented this in its current form. 

Do you think something like a _map frame cheat sheet_ would be useful for others?

PaulWessel:
In between working on the proposal I can see what I can do with drawing the frame ticks, at least when percent < 100.  We could of course use a discussion on various map frame options once we get the basic stuff implemented in C.

joa-quim (approved):
OK, let's merge this and continue later.

seisman:
Many tests fail after merging this PR into master.

PaulWessel:
Yikes, I ran all tests locally.  OK, will check after todays many zooms...  I assume somehow it is not initialized to 100.

PaulWessel:
Fixed by #4625.

PR #4626, opened 2021-01-07
Fix GMT_Set_Default() "API_GRID_LAYOUT" case

joa-quim:
Besides the fix, accept also "?R" and "?C" as synonyms of "columns" & "rows" to be consistent with the "API_IMAGE_LAYOUT" use case.


PaulWessel (approved):
Fine with me.

PR #4636, opened 2021-01-09
Minor under-the-hood changes related to long-options

PaulWessel:
I fixed the issue renaming `read-cols` to incols and `write-cols` to `incols` and `outcols` as these are a bit special since both are **-q** in short-form but must map to separate  **-qi** and **-qo**.  Also, the description of long directive and argument was a bit off in the _gmt_common_longoptions.h_ file,  I also renamed _gmtinit_kw_replace_ to _gmtinit_translate_to_short_options_ and added a dummy inverse function _gmtinit_translate_to_long_options_ we will use to test the translation tables.


joa-quim:
Hmm `incols`` and outcols`` are **i** and **o**, no?

src/gmt_common_longoptions.h
```diff
@@ -3,7 +3,7 @@
 	 * General short-option syntax:
 	 *		-<short_option>[<short_directives>][+<short_modifiers>[<argument>]]
 	 * General long-option syntax:
-	 * 		--<long_option>[=<long_directives][+<long_modifiers>[=<argument>]]
+	 *		--<long_option>[=[<long_directives>:]<arg>][+<long_modifier1>[=<arg1>]][+<long_modifier2>[=<arg2>]]...
```
meghanrjones, line 6: The new developer docs have
```
--long-option[=directive][+modifier1[=arg1]][+modifier2[=arg2]][...]
```
can you explain to me why there is ```:]<arg>]``` after the long_directive here?

PaulWessel: Yes, I had forgotten about that.  I need to dig a bit deeper since I clearly did not provide an example when that is needed.  I think it is for options like -Qd77 which has both a directive (d) and a value (77), hence --longqname=distance:77

PaulWessel:
> Hmm `incols` and `outcols` are **-i** and **-o**, no?

Sorry, I meant **-qi** and **-qo** which is inrows and outrows (not cols).

PaulWessel:
@meghanrjones, I have updated the docs.  My plan on all of this is to compliment the current function that translates long-options to short-options with the inverse: A function that converts short-options to long-options.  This will allow me to fully test the translations we can add some ifdefs in the code for testing to have all modules run its options through the short-to-long-to-short loop.  This will be the closes to @leouieda's unit-tests as we will get on the parsing.

doc/rst/source/devdocs/long_options.rst
```diff
@@ -69,20 +69,21 @@ Clearly, these translation tables will need to address not only the longer *name
...
-where *option* and any *modifier* are single characters and there may be none, one,
+where *option*, *directive*, and any *modifier* are single characters and there may be none, one,
 or more modifiers following the initial option and the
-optional *directive*.  As we saw in the case of **-i**, the sequence of "optional directive
+optional *directive* (and optional *arg*).  As we saw in the case of **-i**, the sequence of "optional directive
 followed by optional modifiers" may in fact be repeated by separating these sequences with
 a comma.  The corresponding long-format syntax format is represented this way::
 
-    -long-option[=directive][+modifier1[=arg1]][+modifier2[=arg2]][...]
+    -long-option[=[directive:]arg][+modifier1[=arg1]][+modifier2[=arg2]][...]
```
meghanrjones, line 80:
```suggestion
    --long-option[=[directive:]arg][+modifier1[=arg1]][+modifier2[=arg2]][...]
```

src/gmt_common_longoptions.h
PaulWessel: Updated that in the web page as well.

PaulWessel:
Ping @GenericMappingTools/core to approve this harmless PR.

joa-quim (approved):

PR #4639, opened 2021-01-09
Change Jz/Z as an optional argument

Esteban82:
-JZ/z for grdview is optional (as can be seen for the examples and -T argument) so I change it. I'm not sure if I did it right.


PaulWessel (approved):
THanks!

PR #4640, opened 2021-01-09
Update longbasemap test obliquepole option

meghanrjones:

The test modern/longbasemap.sh fails after the last commit. This updates the obliquepole option.


PaulWessel:
Oops, I actually fixed that as part of my other PR since I noticed it too.  Already merged it in so perhaps this one is not needed.

PR #4648, opened 2021-01-12
Clarify that grdimage -Q also works for images

PaulWessel:
If an image (instead of a grid and a CPT) is given with **-Q** then we turn black pixels transparent.

If anyone feels we should be more flexible and turn a specific pixel-value transparent instead then we could break this backwards compatibility and examine **COLOR_NAN** instead (128).  To make black pixels transparent one would have to add **--COLOR_NAN**=black.  Thoughts?  I don't think it is given that the user always want to turn black in to transparency...


PaulWessel:
Ping @joa-quim

joa-quim:
Yes, we should be able to set any color transparent, and there is a way for indexed images but I can't find it right now. Also **COLOR_NAN** is not the best name to pick a transparent color for images.

PaulWessel:
Well, it is not a new name, it is one of our defaults.  The reason it might make sense is that we talk about making NANs in the grid transparent, but for images we must decide what that means.  Using the NaN color is not a bad reference I think.  Better than having to add yet another setting like **COLOR_TRANSP** just for this purpose.

joa-quim:
BTW, ``grdimage`` doesn't accept **-D** & **-Q**

PaulWessel:
Hm, maybe my last revision prevented that? THe -D is optional I think as grdimage will usually discover it is an image.

joa-quim:
Nope. It's an older thing and probably only affecting the externals. I have this in my ``grdimage.jl``

```
if (!occursin("-D", cmd))  cmd *= " -D"  end	# GMT bug. It says not need but it is.
```

PaulWessel:
Right, externals may need to set -D.

PaulWessel:
May need an API->external if-test to allow -D then. I will be busy next few ours with Chair-duties...

PaulWessel:
Hi @joa-quim, anything we need to do with this one?  I guess allow -D if external?

joa-quim:
Yes, we have to allow -D for externals too.

Meanwhile I finally found I what I had in mind for alternative way of setting transparency. But it seems it's also when calling gdalwrite. Don't know where to document it other than (where it is)

https://github.com/GenericMappingTools/gmt/blob/master/src/gmt_gdalwrite.c#L360

PaulWessel:
We have this in the parser:

```
#ifdef HAVE_GDAL
			case 'D':	/* Get an image via GDAL */
				if (!API->external)			/* For externals we actually still need the -D */
					GMT_Report (API, GMT_MSG_COMPAT,
					            "Option -D is deprecated; images are detected automatically\n");
				Ctrl->D.active = true;
				Ctrl->D.mode = (opt->arg[0] == 'r');
				break;
#endif
```

So -D is accepted from externals.  The problem is this check:

```
n_errors += gmt_M_check_condition (GMT, Ctrl->D.active && Ctrl->Q.active,
	                            "Option -Q: Cannot use -D when doing colormasking\n");
```

I think that just needs to go since that was the case that we fixed (without -D explicitly given, but an image).  So for the same to work form externals (where -D will be set for that case) we cannot have that check at all.  Agreed?

joa-quim:
Yes. I actually have already quickly checked that removing that check worked but had to drop that change because git was f my mind all the time about non-committed changes when i changed branches.

PaulWessel:
> Well, it is not a new name, it is one of our defaults. The reason it might make sense is that we talk about making NANs in the grid transparent, but for images we must decide what that means. Using the NaN color is not a bad reference I think. Better than having to add yet another setting like **COLOR_TRANSP** just for this purpose.

Hi @joa-quim, we still need to deal with this if we are to allow any other color than 0,0,0 be the transparent pixel.  Another possibility is to have **-Q**[_r/g/b_] which takes an optional r/g/b to be the transparent pixel value for images.

joa-quim:
Yes, I can see that we may want to turn white pixels transparent.

PaulWessel:
OK, will implement -Qrgb in this PR.

PaulWessel:
Tested with that sub_low_byte.tif file for -Q, -Q0 (the same, -Qblack (the same) and -Q1 (the other near-black pixel) and it works.

joa-quim:
OK, while you are at it but a different issue, try plotting that ``sub_low_byte.tif`` with the polar projection and **-n+a**. The *anti-aliasing* is responsible for more than half the time of the command run.

PaulWessel:
Not sure what we can do about that.  But a separate issue from this PR though.  OK to approve?

joa-quim (approved):

PaulWessel:
If you give -n+a then you may end up with an aliased image.  Perhaps that is OK if you are in a hurry, but I would rather not.  Especially for things like polar projection where at one end (near pole) you definitively need to average all those nodes that suddenly plot very close, while at the far end you want the interpolation.

joa-quim:
Couldn't see the difference (by eye) between with or without averaging, but seems strange that averaging takes longer that the projection itself.

PaulWessel:
Well, the image is 3561x4051 and we are basically doing an averaging of 3 values (instead of 1 for grids).

PaulWessel:
Out of curiosity I will add a debug option to save the counter used in the averaging and write a grid.  THen we can at least play with settings and see if we can modify some of the defaults.  This will be under #ifdef DEBUG only so not accessible to most people.

PR #4660, opened 2021-01-16
Let grdimage deal also with indexed images sent in by externals.

### LINKED ISSUE #4657: grdimage lost ability to deal with indexed images from externals
That happened in the ``grdimage`` rewrite where, from memory (mine), indexed images are previously expanded to RGB after reading. But now from externals (where the Image object lands), grdimage now does this and doesn't look anymore to the colormap contents. I used to be able to set transparency via colormap but that is lost too.

```
gray_only = (I->header->n_bands == 1);	/* Got a grayscale image */
```

joa-quim:
Rename gmtapi_expand_index_image() to gmtlib_expand_index_image() and export it
Fixes #4657

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -4198,6 +4198,7 @@ GMT_LOCAL int gmtapi_import_ppm (struct GMT_CTRL *GMT, char *fname, struct GMT_I
 	return GMT_NOERROR;
 }
 
+#ifdef HAVE_GDAL
 GMT_LOCAL bool gmtapi_expand_index_image (struct GMT_CTRL *GMT, struct GMT_IMAGE *I_in, struct GMT_IMAGE **I_out) {
 	/* In most situations we can use an input image given to a module as the dataset to
 	 * plot.  However, if the image is indexed then we must expand it to rgb since we may
@@ -4288,6 +4289,7 @@ int gmtlib_ind2rgb (struct GMT_CTRL *GMT, struct GMT_IMAGE **I_in) {
 	}
 	return GMT_NOERROR;
 }
+#endif
 
 /*! . */
 GMT_LOCAL struct GMT_IMAGE *gmtapi_import_image (struct GMTAPI_CTRL *API, int object_ID, unsigned int mode, struct GMT_IMAGE *image) {
@@ -4568,12 +4570,14 @@ GMT_LOCAL struct GMT_IMAGE *gmtapi_import_image (struct GMTAPI_CTRL *API, int ob
 
 	if (done) S_obj->status = GMT_IS_USED;	/* Mark as read (unless we just got the header) */
 
+#ifdef HAVE_GDAL
 	if (no_index && gmtapi_expand_index_image (API->GMT, I_obj, &Irgb)) {	/* true if we have a read-only indexed image and we had to allocate a new one */
 		if (GMT_Destroy_Data (API, &I_obj) != GMT_NOERROR) {
 			return_null (API, API->error);
 		}
 		I_obj = Irgb;
 	}
+#endif
 
 	if (!via) S_obj->resource = I_obj;	/* Retain pointer to the allocated data so we use garbage collection later */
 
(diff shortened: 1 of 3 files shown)


PaulWessel (commented):
Before I approve this: Should not all of this happen in gmtapi_import_image? Seems like we should always expect indexed images to rgb since we have no operator that just works directoy in the index. Instead of doing it in grdimage.  What about other modules that read images, won't you have to duplicate that step there too? Seems like it needs to be centralized.

joa-quim:
I don't remember if any other module would need this. The problem is that the code flow followed by
```
julia> imshow(It, Vd=1)
        grdimage  -R0/350/0/258 -JX12c/0 -Baf -BWSen -D -P -K > C:\TMP\GMTjl_tmp.ps
```
does not pass in ``gmtapi_import_image``

PaulWessel:
Well, **grdmix** might, for instance; not sure about **grdview**.
Help me understand why in your example, grdimage is getting an image without visiting _GMT_Read_Image_ and hence _gmtapi_import_image_?

joa-quim:
In ``GMT_Read_Data`` it slips through 
```
return (gmtapi_pass_object (API, S_obj, family, mode, wesn));
```
and then directly gets the image with ``gmtapi_get_image_data``

PaulWessel:
So that is where we need to intervene then, I think.  Otherwise we are stuck repeating the same code in other places.

joa-quim:
But we only want to convert to RGB if image is going to be sent to pslib. If I want to save it with gdalwrite for example I don't want it to be converted.

joa-quim:
We can turn this into a function, so the amount of possible repetitions would be negligible.
```
		if (I->header->n_bands == 1 && I->n_indexed_colors > 0) {	/* Indexed image, must convert to RGB */
			struct GMT_IMAGE *Irgb = NULL;
			gmtlib_expand_index_image (GMT, I, &Irgb);	/* true if we have a read-only indexed image and we had to allocate a new one */
			if (GMT_Destroy_Data (API, &I) != GMT_NOERROR)
				Return (API->error);
			I = Irgb;
		}
```

PaulWessel:
OK, that is not too bac.  Please make it a function so that it is good to go elsewhere as needed.  Please add some comments so we remember why it was added and what it does.

joa-quim:
F. It was supposed to have been an easy thing to do.

PaulWessel (approved):
Seems to not have broken anything, so good!

seisman:
No, it breaks if GDAL is not installed.

PaulWessel:
Thanks for checking that - I used GDAL.  Add some ifdefs, @joa-quim ?

joa-quim:
Yes, it was because of some IFDEFS I had removed. It is silly that we end up requiring GDAL to do this opts that have nothing to do with GDAL.

seisman:
The CI test without GDAL now passes.

joa-quim:
BTW,  ``set (GMT_EXCLUDE_GDAL TRUE)`` doesn't work. Raises error
```
/mnt/c/progs_cygw/GMTdev/gmt5/master/src/gmt_gdalread.h:26:10: fatal error: gdal.h: No such file or directory
```

PaulWessel:
Why? We have

```
#ifdef HAVE_GDAL
#	include "gmt_gdalread.h"      /* GDAL support */
#endif
```

Oh, I think it is your gmt_ogrread.c file that complains.

joa-quim:
Nope, even if I wrap it under a #ifdef HAVE_GDAL I still get the same error.

PaulWessel:
Not reproducible here by turning GMT_EXCLUDE_GDAL true.  Maybe you have some other code in there that gets added and thus complains.  It cannot go into those ifdefs if HAVE_GDAL is not set.

PaulWessel:
Not done yet:

../src/grdimage.c:1296:46: warning: incompatible pointer types passing 'struct GMT_IMAGE **' to parameter of type 'struct GMT_IMAGE *'; remove & [-Wincompatible-pointer-types]
                if (!Ctrl->A.active && gmtlib_ind2rgb(GMT, &I)) {
                                                           ^~
../src/grdimage.c:1106:72: note: passing argument to parameter 'I_in' here
EXTERN_MSC int gmtlib_ind2rgb (struct GMT_CTRL *GMT, struct GMT_IMAGE *I_in);
                                                                       ^

joa-quim:
Shit forgot that change. 
Can't see warnings under the awful flood of GCC warnings nor on Win where all OMP variables show up as unused vars.

PR #4662, opened 2021-01-16
Add Vincenty direct problem solution to gmt vector -Tt

PaulWessel:

In addition the the spherical solution, this PR adds the ellipsoidal (Vincenty-1975) solution to the direct geodetic problem: Given a point, a bearing, and distance, determine the end point.  The **-j** option is used to control which solution is returned. This relates to this [posting](https://forum.generic-mapping-tools.org/t/how-to-calculate-coordinates-from-bearing-distance/1217/8) on the Forum.  After this PR, the second point that is 23 nautical miles in azimuth 310 from (50N, 8E) is

Spherical [Default or **-j**g]:
```
echo 8 50 | gmt vector -Tt310/23n
7.54112215577	50.24533433

```
Ellipsoidal [**-je**]
```
echo 8 50 | gmt vector -Tt310/23n -je
7.54254355708	50.2452568932

```

As @joa-quim reported, the proj/geod reports

7.542543557083350 50.245256893178066

Same if I mess with **FORMAT_FLOAT_OUT**.

### DIFF
--- a/doc/rst/source/gmtvector.rst
+++ b/doc/rst/source/gmtvector.rst
@@ -121,7 +121,8 @@ Optional Arguments
     from the third and fourth data column in the file, and **x** for cross-product.
     If **-T** is not given then no transformation takes place; the
     output is determined by other options such as **-A**, **-C**,
-    **-E**, and **-N**.
+    **-E**, and **-N**. **Note**: For **-Tt** and geographic coordinates we will
+    perform a great circle calculation unless **-je** is selected.
 
 .. _-V:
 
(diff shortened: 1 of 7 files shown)


joa-quim (approved):

joa-quim:
Well, 
```
As @joa-quim reported, the proj/geod reports

7.542543557083350 50.245256893178066
```
was from my Matlab *Dead Reckon* (the name of this operation) routine, but same as ``geod``

KristofKoch:
You … I … don't know what to say. Thank you, @PaulWessel! Test drove the branch as well with equator crossings and around the poles – works great!

PaulWessel:
As for all Vincenty stuff it may get super-slow if you are requesting a point that is close to the antipode of the start point, but that is not a likely scenario for mst.

PR #4665, opened 2021-01-17
Let developers save the antialias counter when projecting grids/images

PaulWessel:
For non-rectangular projections we must do an anti-alias step to ensure we avoid aliasing.  Anti-aliasing can be turned off via **-n+a**.  However, for exploring the function of anti-aliasing, developers can now add **-n+A** to save the averaging counter to a grid for examination.  The modifier is only accessible for developers building with **-DDEBUG** so cannot accidentally be used by users.  Because the anti-alias operation takes time it is good to be able to study this in more detail.

Below is the grid of counters for the sub_low_byte.tif we have examined due to the transparency pixel stuff.  The image shows that close to the pole there are many tiny places where up to ~10 input pixels are averaged for every output pixels, while at the south end it is typically just 1 or sometimes 2.

t

### DIFF
--- a/src/gmt_common.h
+++ b/src/gmt_common.h
@@ -211,9 +211,10 @@ struct GMT_COMMON {
 		bool active;
 		struct GMT_LEGEND_ITEM item;
 	} l;
-	struct n {	/* -n[b|c|l|n][+a][+b<BC>][+c][+t<threshold>] */
+	struct n {	/* -n[b|c|l|n][+a][+b<BC>][+c][+t<threshold>] (and +A for debugging) */
 		bool active;
 		bool antialias;		/* Defaults to true, if supported */
+		bool save_debug;	/* Write antialias counters to tmp grid */
 		bool truncate;		/* Defaults to false */
 		unsigned int interpolant;	/* Defaults to BCR_BICUBIC */
 		bool bc_set;		/* true if +b was parsed */
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -2230,6 +2230,11 @@ int gmtinit_parse_n_option (struct GMT_CTRL *GMT, char *item) {
 
 	while ((gmt_strtok (&item[k], "+", &pos, p))) {
 		switch (p[0]) {
+#ifdef DEBUG
+			case 'A':	/* Debug antialias will save the counter to a grid */
+				GMT->common.n.save_debug = true;
+				break;
+#endif
 			case 'a':	/* Turn off antialias */
 				GMT->common.n.antialias = false;
 				break;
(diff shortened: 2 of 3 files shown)


joa-quim (approved):

seisman:
It doesn't compile well on all platforms.

joa-quim:
Maybe I should open another issue but before that here it goes (I'm using Julia here just to be able to accurately time the execution time)
```
julia> @time imshow("sub_low_byte.tif", J=:guess, Vd=1)
        grdimage sub_low_byte.tif  -JD80.0608274538835/82.75349518138835/79.12611688231391/90.0082517795372/12c -Baf -BWSen -R39.27833884/120.84331606777/77.312427732777/88.19456263 -C -P -K > C:\TMP\GMTjl_tmp.ps
  6.930420 seconds (4.97 k allocations: 456.703 KiB)

julia> @time imshow("sub_low_byte.tif", J=:guess, Vd=1, n="+a")
        grdimage sub_low_byte.tif  -JD80.0608274538835/82.75349518138835/79.12611688231391/90.0082517795372/12c -Baf -BWSen -n+a -R39.27833884/120.84331606777/77.312427732777/88.19456263 -C -P -K > C:\TMP\GMTjl_tmp.ps
  4.069992 seconds (476.15 k allocations: 30.116 MiB, 0.15% gc time, 2.44% compilation time)
```
So the anti-alias makes it run ~70% slower and our ``gm`` test wouldn't see any difference from the files.

```
gm compare -density 200 -maximum-error 0.005 -highlight-color magenta -highlight-style assign -metric rmse -file lixo.png GMTjl_tmp_no_a.ps GMTjl_tmp_a.ps
Image Difference (RootMeanSquaredError):
           Normalized    Absolute
          ============  ==========
     Red: 0.0009242512       60.6
   Green: 0.0011457076       75.1
    Blue: 0.0013581637       89.0
   Total: 0.0011563585       75.8
```

Tried a cleverness of reverting ``if (GMT->common.n.antialias)`` to ``if (!GMT->common.n.antialias)`` in both ``gmt_grd_project`` and ``gmt_img_project`` and run the tests. But though seeing a significant speed up with no apparent increase in number of failures, had to interrupt it because there were numerous crashes popping up. Don't see why this would cause crashes but did not investigate.

PaulWessel:
If we can find some reasonable criteria to determine when the aliasing is not  very significant then perhaps we could change the default behavior.  However, I am not sure how to do that.  Your test above depends on the data set. A data with lots of variability closer to the pole would be much more aliased than a smoother data set.

joa-quim:
Also tried with a Mercator example, and same thing/ Speed up and no detectable difference
```
julia> @time imshow("faa.grd", J=:merc, Vd=1, n="+a")
        grdimage faa.grd  -JM12c -Baf -BWSen -n+a -R-24.2/-1.8166666666667/29.116666666667/47.066666666667 -C -P -K > C:\TMP\GMTjl_tmp.ps
  1.488302 seconds (4.96 k allocations: 453.625 KiB)

julia> @time imshow("faa.grd", J=:merc)
  2.469542 seconds (458.80 k allocations: 29.014 MiB, 0.30% gc time, 3.92% compilation time)
```

joa-quim:
That's why I tried the entire test dataset but some other shit came in between.

PR #4672, opened 2021-01-20
Clarify that placing -Sq in segment headers does not eliminate need to give -Qs on command line

PaulWessel:
See [forum post](https://forum.generic-mapping-tools.org/t/psxy-quoted-line-sq-options-in-segment-header/1227) for context. In review it was not clear that **-Sq** is still needed on the command line to tell **psxy** we are plotting a symbol; just having these in the segment header is too late as now we are plotting lines...
I have updated the docs to make this requirement clear.

### DIFF
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -336,8 +336,8 @@
         with the same pen as set for the line (i.e., via **-W**). To use an
         alternate pen, append **+p**\ *pen*. To skip the outline, just use
         **+p** with no argument.  To make the main front line invisible, add **+i**.  **Note**:
-        By placing **-Sf** options in the segment header you can change the front
-        types on a segment-by-segment basis.
+        By placing **-Sf** options in the segment headers that differ from the one on
+        the command line you can change the front types on a segment-by-segment basis.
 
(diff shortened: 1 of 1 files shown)


doc/rst/source/explain_symbols.rst_
```diff
@@ -336,8 +336,8 @@
         with the same pen as set for the line (i.e., via **-W**). To use an
         alternate pen, append **+p**\ *pen*. To skip the outline, just use
         **+p** with no argument.  To make the main front line invisible, add **+i**.  **Note**:
-        By placing **-Sf** options in the segment header you can change the front
-        types on a segment-by-segment basis.
+        By placing **-Sf** options in the segment headers that differ from the one on
+        the command line you can change the front types on a segment-by-segment basis.
```
joa-quim, line 340: "font" not "front"

meghanrjones: **-Sf** is for drawing a front, so I think this is probably an intentional word choice

meghanrjones (approved):

doc/rst/source/explain_symbols.rst_
joa-quim: Yes, likely.

PR #4681, opened 2021-01-22
Add test script showing failure of clipping in TM projection

PaulWessel:
See [forum](https://forum.generic-mapping-tools.org/t/how-to-avoid-the-funky-look-of-jt-and-se/1224) for background.  This PR just adds a minimal failing script to remind us to address this problem.

### DIFF
--- a/test/psxy/tissot_TM.sh
+++ b/test/psxy/tissot_TM.sh
@@ -3,4 +3,4 @@
 # Demonstrate failure to clip ellipses against periodic TM boundary in y.
 # See https://forum.generic-mapping-tools.org/t/how-to-avoid-the-funky-look-of-jt-and-se/1224
 ps=tissot_TM.ps
-echo 30 -60 2000k | gmt psxy -R0/360/-80/80 -JT-15/15/16c -Bafg -SE- -Gred@50 -Wred -P > $ps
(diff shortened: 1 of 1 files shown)


meghanrjones:
Why track this with a test rather than a new issue?

PaulWessel:
I should certainly open an issue too.  The thinking was that this is a particularly tricky problem since in this projection, it is the y-direction that has the periodic longitude boundary, so I wanted to elevate its status and make sure it is kept alive in our minds.

PaulWessel:
The other reason is that these types of bugs have in the past been tricky to eradicate and later code changes often reactivates such bugs.  The test will then start to fail and hence notify us if that happens.

seisman:
You don't provide a correct PS file for this test, so it always fails, right?

PaulWessel:
yes, no truth yet but I may be able to fake one soon

meghanrjones:
My preference would be to use a label on the issue to elevate the status (e.g., 'priority: high') rather than a test without a postscript comparison. It should be considered a known failure, which I think could make it relatively easily forgotten. I would not be as concerned if there's a correct answer accompanying the test.

PaulWessel:
OK, I am fine with that - we can delete the PR if you want.  I am working on the problem right now and just need to get it to work...

PR #4687, opened 2021-01-23
Improve color space documentation via intensity illustration

PaulWessel:

The Color Spaces section in the cookbook appendix now has a figure that illustrations how intensities affects the color via darkening and whitening while keeping the hue constant:

GMT_color_hsv

### DIFF
--- a/doc/scripts/GMT_color_hsv.sh
+++ b/doc/scripts/GMT_color_hsv.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+#
+# Show a slice through RGB cube for h = 60 (yellow) and illustrate effect of intensity changes
+# Cheating and rotating the slice to the horizontal
+ps=GMT_color_hsv.ps
+x=4
+r=217
+b=54
+smin=1.0
+smax=0.1
+vmin=0.3
+vmax=1.0
+y=`gmt math -Q $x 2 SQRT MUL =`
+gmt grdmath -I1 -R0/255/0/255 Y 256 MUL X ADD = rgb_cube.grd
+gmt grdmath -I1 -R0/255/0/255 Y 255 DIV = v.grd
+gmt grdmath -I1 -R0/255/0/255 1 X Y DIV SUB X Y LE MUL = s.grd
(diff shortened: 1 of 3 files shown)


joa-quim (approved):

doc/scripts/GMT_color_hsv.sh
```diff
@@ -0,0 +1,94 @@
...
+y=`gmt math -Q $x 2 SQRT MUL =`
+gmt grdmath -I1 -R0/255/0/255 Y 256 MUL X ADD = rgb_cube.grd
+gmt grdmath -I1 -R0/255/0/255 Y 255 DIV = v.grd
+gmt grdmath -I1 -R0/255/0/255 1 X Y DIV SUB X Y LE MUL = s.grd
+gmt math -T0/255/1 -N3/0  T 256 MUL 0.5 SUB -C2 256 ADD  = | awk '{printf "%s\t%d/%d/0\t%s\t%d/%d/255\n", $2, $1, $1, $3, $1, $1}' > rgb_cube.cpt
+gmt psxy -R0/8/0/8 -Jx1i -P -K -X0.75i -Y2i -T > $ps
+cat << EOF >> $ps
+gsave
+-54.735611 rotate
+EOF
```
meghanrjones, line 22: @PaulWessel, I'm working on updating all the cookbook scripts to modern mode which will help simplify the CMake build setup for the docs. I am stuck on this one because of the cheat of using PostScript to rotate the slice to the horizontal. Do you know if there's a way to accomplish this figure using pure GMT?

PaulWessel:
Yes, probably.  It is a rotation of -54.73 degrees so presumably one could build a rectangular grid (the yellow iceberg you see would be circumscribed by the rectangular grid and then the rotation is applied to the coordinates of the original grid to yield x', y' and then those must be gridded.  Maybe project -T -C can be used to rotation those nodes, e.g.

gmt grd2xyz rgb_cube.grd | gmt project -C<coordinates of W> -B<coordinates of Y> > a.txt
Same for the other grids
Now grid the rotated points in a rectangular region (maybe via nearneighbor?)
Set clip path etc.

I.e., not trivial I think

PR #4688, opened 2021-01-23
Fix typo for reverse doc string

willschlitzer:
The doc string for `-I` previous said "so the latter much be compatible with the changed". I'm assuming that "much" should be "must" so I submitted this PR.


meghanrjones (approved):

meghanrjones:
Thanks for submitting the fix, @willschlitzer.

PR #4691, opened 2021-01-24
Add doc illustrations of how CMYK relates to RGB and DEM slopes to intensity

PaulWessel:
This documentation PR adds two figures rescued from the (in)famous UH GMT course notes by translating from GMT 3 to 6. One is used when discussing the relationship between surface slopes and intensity and then artificial illumination:

GMT_slope2intensity

while the other is used when discussing the CMYK color system

GMT_cmyk

I wonder if we perhaps should rename the old Appendix that is called _Color Space: The Final Frontier,_ which was sort of a funny name at the time, but perhaps now it might be better to call it _Color Systems and Artificial Illumination_?


joa-quim (approved):
Yes, rename it. It's not that final. We also have the Lambertian illumination whose description is not included here.

PR #4710, opened 2021-01-28
Testing a GMT_THEME off solution

PaulWessel:
Goal is to allow classic mode users to override specifics in a modern-mode theme settings without having the **GMT_THEME** be parsed again at the end and thus overwrite the changes.

@meghanrjones, see if this sub-branch off gmt-themes works as you expect.  It does allow a classic user to say

```
gmt defaults -D > gmt.conf
gmt set GMT_THEME modern FONT_ANNOT_PRIMARY auto,Times-Roman,black
gmt psbasemap -RDE+R10m -Bpa30m -Bsa1d -JL10.5/51/49/53/17c > test.ps
```

and get autoscaling with selected annotation font.  In order to allow that 2nd command to work I now update all settings immediately if **GMT_THEME** is encountered - before we waiting until all other settings had been parsed before we overrode everything.  I think it makes more sense that later options to gmt set overrides whatever the earlier ones set.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -9640,6 +9640,7 @@ GMT_LOCAL int gmtinit_update_theme (struct GMT_CTRL *GMT) {
 	char theme_file[PATH_MAX] = {""};
 
 	if (!GMT->current.setting.update_theme) return GMT_NOERROR;	/* Nothing to do */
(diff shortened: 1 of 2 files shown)


meghanrjones (approved):
This works much better

PR #4715, opened 2021-01-29
Themes autoaxes

PaulWessel:

For modern theme, sets **MAP_FRAME_AXES** = "auto" as the default, and this is then updated when we know more about the projection, type of plot, perspective, etc.  This worked well for the _GMT_tut_18_ and _polarpanels_ cases.  The PR waits with setting the frame parameters if "auto" is set, and then when all the auto items are updated we examine if the lot is perspective (we then determine quadrant and selects what the WrStZ equivalent is), polar (we then check if S and N should be flipped), or else we use the WrStZ default for modern.  These are still subject to the override if a geographic fancy border was selected.  Please check this out @meghanrjones and let me know if there are related cases that still need work.  I managed to make a conflict as clearly my brain cannot handle branches to other branches...but will sort that out later.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -18491,6 +18491,7 @@ void gmt_auto_offsets_for_colorbar (struct GMT_CTRL *GMT, double offset[], int j
 
 	char side, axis, B_delim[2] = {30, 0}, p[GMT_BUFSIZ] = {""};	/* Use ASCII 30 RS Record Separator between -B strings */
 	char file[PATH_MAX] = {""};
+	char *frame_axes = (!strcmp (GMT->current.setting.map_frame_axes, "auto")) ? "WrStZ" : GMT->current.setting.map_frame_axes;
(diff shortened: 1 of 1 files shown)


meghanrjones:
There's some problems with the axes labels overlapping the adjacent subplots, which seem more prevalent with the auto MAP_FRAME_AXES. Check out test/psxy/autocolorpols/autocolorpols.sh or test/pslegend/autolegendcolorL.sh for examples.

PaulWessel:
The problem is a combination of the issues addressed in two separate branches: This one and the subplot-axes.  With both merged I think it will be OK.

PaulWessel:
I think it was a mistake of me to make these sub-branches for merging with gmt-themes since in this case the two needs to work together to fix issues in gmt-themes branch.  Please approve both so I can merge into gmt-themes and address any remaining problems.

meghanrjones (approved):

PR #4717, opened 2021-01-29
Update documentation for gmt themes

meghanrjones:

Updates the gmt.conf documentation to include theme dependent defaults and a description of auto scaling. Currently the text inside brackets link to a table with theme dependent settings, but this could be done in other ways if it is not currently clear.

@PaulWessel you should be able to build the docs now.


PaulWessel:
This looks great, @meghanrjones.  However, since it is a bit new let's also get approval from @seisman and @joa-quim before merging.  Meanwhile, two questions and possibly requests:

1. **MAP_FRAME_AXES**: Maybe we need to explain that if auto then we will normally supply WrStZ but if a specific projection of perspective view makes that a bad choice then we supply what we believe is best.
2. At the top of gmt.conf.rst it says "Note that default distances and lengths below are given in both cm or inch ; the chosen default depends on your choice of default unit (see PROJ_LENGTH_UNIT)".  I think this is outdated and should be removed.  It looks like almost all of the dimensions are now listed in points but users can supply their arguments in any of inch, cm, or points.  
3. Looks like **FONT_SUBTITLE** still needs to be updated to say [theme-dependent]?
4. MAP_TICK_PEN_PRIMARY: I cannot recall why this is not auto like GRID_PENs and FRAME_PEN?

meghanrjones:
>     1. **MAP_FRAME_AXES**: Maybe we need to explain that if auto then we will normally supply WrStZ but if a specific projection of perspective view makes that a bad choice then we supply what we believe is best.
Agreed, I will add some documentation after we merge the auto functionality for MAP_FRAME_AXES.
>     2. At the top of gmt.conf.rst it says "Note that default distances and lengths below are given in both cm or inch ; the chosen default depends on your choice of default unit (see PROJ_LENGTH_UNIT)".  I think this is outdated and should be removed.  It looks like almost all of the dimensions are now listed in points but users can supply their arguments in any of inch, cm, or points.
Agreed, I will update this separately because it is not themes-specific. 
>     3. Looks like **FONT_SUBTITLE** still needs to be updated to say [theme-dependent]?

Thanks, updated.
>     4. MAP_TICK_PEN_PRIMARY: I cannot recall why this is not auto like GRID_PENs and FRAME_PEN?

It was applied in gmt_init.c, just not listed in share/themes. I added it to the documentation and themes .conf files.

joa-quim:
I think we should clarify this, which in fact is not exactly true.
> Note that default distances and lengths below are given in both cm or inch; the chosen default depends on your choice of default unit (see PROJ_LENGTH_UNIT).

The default is in fact **cm only** because that's what in hardwired in code. But users can create a ``gmtdefaults`` file with US units.

I like the table modern modes setting table. It goes to the point. However, I wonder if it should be in a separate page instead of a part of the ``gmt.conf``. Not a big issue right now but when we start to add more themes the ``gmt.conf.rst`` may grow to undesirable size. Better to a single document dedicated to *themes*.

In a separate matter, the those **[NOTICE]** are annoying. I'm getting tons of these when running the Julia tests.
```
[NOTICE]: Computed primary annotation font size: 9.26667 p  Dimension scaling: 0.926667
```

PaulWessel:
> The default is in fact cm only because that's what is hardwired in code. But users can create a gmtdefaults file with US units.

Looks like most things like offsets and ticklengths are given in points.

meghanrjones:
> > The default is in fact cm only because that's what is hardwired in code. But users can create a gmtdefaults file with US units.
> 
> Looks like most things like offsets and ticklengths are given in points.

Yes, the output from gmtdefaults for lengths and distances is points regardless of whether the installation was SI or US. I do not think this is currently clear in the gmt.conf and gmtdefaults documentation. I will submit a PR with some changes, but I think the ultimate solution is to write a how-to guide for changing the configuration parameters and to link to that from the documentation for gmt.conf, gmtdefaults, and gmtset. 

@joa-quim, the table for the themes settings is in a separate ReST file (explain_themes_settings.rst_) that is included in gmt.conf.rst, so it would be very easy to move this to a different location if the number of themes grows. Do you think it should be moved now?

joa-quim:
We have a mix of points and cm. Fig sizes are in **cm**, but none are ever in **in** (uff, finally) unless specifically selected either on command line or in a defaults file.

PaulWessel:
Yes, the documentation is, or should be, either in points for relative small lengths and fonts, and cm for things that can be larger.  If we find exceptions than each of us can do a quick PR to fix them.

joa-quim:
The fact that the theme is in a separate rst file is invisible to the user. Sorry, I was not clear. I think it would be better that it opens as a different document. Right, for now, with only one theme, it's fine to be includes but if we are going to extend it we could start right now. Not a big deal though.

meghanrjones:
> The fact that the theme is in a separate rst file is invisible to the user. Sorry, I was not clear. I think it would be better that it opens as a different document. Right, for now, with only one theme, it's fine to be includes but if we are going to extend it we could start right now. Not a big deal though.

Would you prefer the page to be free-standing, like [proj specifications](https://docs.generic-mapping-tools.org/dev/proj-codes.html), or in the cookbook, like [built-in cpts](https://docs.generic-mapping-tools.org/dev/cookbook/cpts.html), both of which are included in the [Quick Links table](https://docs.generic-mapping-tools.org/dev/index.html#quick-links)?

joa-quim:
> Would you prefer the page to be free-standing, like proj specifications, or in the cookbook, like built-in cpts, both of which are included in the Quick Links table?

Yes, I think it would be better. More expansible.

meghanrjones:
OK, it should be better now. I also added links to the auto scaling cookbook feature from all the parameters that can be automatically sized, and updated that page for recent changes (and the pending scale based on both height and width change).

PaulWessel:
This looks nice to me.  @joa-quim, @seisman  OK?

seisman:
Looks good.

joa-quim:
Yes, looks nice.

PaulWessel:
OK to merge, @meghanrjones

PR #4719, opened 2021-01-29
Fix bug in -Td+zTZ where time zone was always set to 0

### LINKED ISSUE #4695: Time zone not changing when passing datetime to solar function
Using the **+z** modifier, I'm attempting to change the time zone for the **-T** parameter for `solar`.  However, changing the time zone does not result in a different display of the day/night terminator as would be expected. This is an error I found when attempting to wrap the function in PyGMT, and @joa-quim asked me to open an issue at [this forum post](https://forum.generic-mapping-tools.org/t/syntax-for-setting-time-zone/1249).

The scripts that I'm passing will be with my PyGMT wrapped version of `solar`, but I believe it will adequately demonstrate the issue. Using identical `Figure.coast` commands, these different solar plots result in identical figures.

#### fig.solar(Td="+d2021-01-24T10:57:00+z-8", fill="navyblue@80")

**Actual outcome**

Screenshot from 2021-01-24 11-00-50

**Expected outcome**

The plots to have different day/night terminators.

**System information**

* Operating system: Ubuntu 20.04
* PyGMT version: 0.2.2.dev

joa-quim:
Fix #4695

### DIFF
--- a/src/pssolar.c
+++ b/src/pssolar.c
@@ -1,6 +1,6 @@
 /*--------------------------------------------------------------------
  *
- *	Copyright (c) 1991-2020 by the GMT Team (https://www.generic-mapping-tools.org/team.html)
+ *	Copyright (c) 1991-2021 by the GMT Team (https://www.generic-mapping-tools.org/team.html)
  *	See LICENSE.TXT file for copying and redistribution conditions.
  *
  *	This program is free software; you can redistribute it and/or modify
--- a/src/pssolar.c
+++ b/src/pssolar.c
@@ -1,6 +1,6 @@
 /*--------------------------------------------------------------------
  *
- *	Copyright (c) 1991-2021 by the GMT Team (https://www.generic-mapping-tools.org/team.html)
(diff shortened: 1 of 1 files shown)


src/pssolar.c
```diff
@@ -1,6 +1,6 @@
 /*--------------------------------------------------------------------
  *
- *	Copyright (c) 1991-2021 by the GMT Team (https://www.generic-mapping-tools.org/team.html)
+ *	Copyright (c) 1991-2020 by the GMT Team (https://www.generic-mapping-tools.org/team.html)
```
meghanrjones, line 3: Why change this back to 2020?

meghanrjones:
Should we add a test for this?

joa-quim:
Wasn't me!!!!!!

PaulWessel (approved):

joa-quim:
We can add **+z1** to ``-T+d2012-6-12T18:00:00`` in ``pssolar/shading.sh`` but it will need a new PS orig.

Issue #4703, opened 2021-01-25
Label with shadow and rounded corner in movie (-Lc)


Esteban82:
Request based on the [similar issue of the forum](https://forum.generic-mapping-tools.org/t/label-with-shadow-and-rounded-corner-in-movie-lc/1262).

Basically, let label of movie have the same option than -F of legend (and others): Primary and secondary border, round corner, shadow.

PR #4724, opened 2021-01-31
Add new pstext option to cast a shade beneath a text box

PaulWessel:

This PR is made in support of #4703 by adding new option **-S**[_dx_/_dy_/][_shade_] to **pstext** so that either single or paragraph text may enjoy the optional shade-effect.  In the process, I did

1. Report an error if **-C+tc**|**C** is used with regular (i.e., not **-M**) text
2. Add a test to demonstrate that the option works for both types of text
3. Fix a minor issue in another test (book.sh)
4. Implement **-S** so that the backwards compatibility for GMT4 **-S**_pen_ still is honored.

As an example, this suggestion is now listed in the documentation:

`gmt text -R0/10/0/5 -Jx1c -F+f32p+cCM+tWELCOME -B -Gyellow -Wfaint -S -C+tO -pdf map`

with this result:

map

### DIFF
--- a/src/pstext.c
+++ b/src/pstext.c
@@ -162,7 +162,7 @@ static void Free_Ctrl (struct GMT_CTRL *GMT, struct PSTEXT_CTRL *C) {	/* Dealloc
 	gmt_M_free (GMT, C);
 }
 
-GMT_LOCAL void pstext_output_words (struct GMT_CTRL *GMT, struct PSL_CTRL *PSL, double x, double y, char *text, struct PSTEXT_INFO *T) {
+GMT_LOCAL void pstext_output_words (struct GMT_CTRL *GMT, struct PSL_CTRL *PSL, double x, double y, char *text, struct PSTEXT_INFO *T, struct PSTEXT_CTRL *Ctrl) {
 	double offset[2];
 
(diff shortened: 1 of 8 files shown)


src/pstext.c
```diff
@@ -356,6 +368,8 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Message (API, GMT_TIME_NONE, "\t-N Do Not clip text that exceeds the map boundaries [Default will clip].\n");
 	GMT_Option (API, "O,P");
 	GMT_Message (API, GMT_TIME_NONE, "\t-Q For all text to be (l)lower or (u)pper-case [Default leaves text as is].\n");
+	GMT_Message (API, GMT_TIME_NONE, "\t-S Plot a shadow behind the text box.  Requires -G<color> to be given as well.\n");
```
joa-quim, line 371: If it requires **-G** as well what is the **-S**_shade_ parameter used to?

PaulWessel: Text box may be white, yellow, or whatever, but the shade it casts need to be different for most cases.  See plot above.

joa-quim (approved):

joa-quim:
Yes, realized that only after writing my comment.

PaulWessel:
Need this to do so from movie but this is actually a nice option in text as is.

PR #4727, opened 2021-01-31
Allow movie labels to have drop-shadow and rounded rectangular box

### LINKED ISSUE #4703: Label with shadow and rounded corner in movie (-Lc)

PaulWessel:
See #4703 for background.  This PR adds the capability of **-L** to select a _rounded_ rectangular text box via new modifier **+r.**  It also allows you to add a _drop-down shadow box_ (of same dimension and shape) with selected offset [4p/-4p] and shade [gray50] via **+h**[_dx_/_dy_][_shade_].  Closes #4703.  Below is a frame from animation 7 after changing **-Lf** to **-Lf+g**_lightgreen_**+h**2p/-2p/_darkgreen_**+r**.  Hopefully @Esteban82 will want to test this out as well.

anim07

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -189,7 +189,7 @@ Optional Arguments
 
 .. _-L:
 
-**-L**\ *labelinfo*
+**-L**\ *labelinfo*\ [*modifiers*]
     Automatic labeling of individual frames.  Repeatable up to 32 labels.  Places the chosen label at the frame perimeter:
     **e** selects the elapsed time in seconds as the label; append **+s**\ *scale* to set the length
     in seconds of each frame [Default is 1/*framerate*],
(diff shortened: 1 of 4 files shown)


joa-quim (approved):

Esteban82:
> Hopefully @Esteban82 will want to test this out as well.

Many thanks @PaulWessel. It work perfect. Just a comment. I was surprised that you use +h for the shade effect. I was imaging that you would use +s like in [legend +F](https://docs.generic-mapping-tools.org/dev/legend.html#f).

PaulWessel:
I think **+s** was already taken for scale in **-L**, no?

Esteban82:
Yes, true. I didn't see it.

PR #4729, opened 2021-01-31
Clarify units in plot and plot3d -D option

PaulWessel:
These are plot distance shifts in cm, inch, or points.

### DIFF
--- a/doc/rst/source/plot_common.rst_
+++ b/doc/rst/source/plot_common.rst_
@@ -70,6 +70,7 @@ Optional Arguments
 **-D**\ *dx*/*dy*
     Offset the plot symbol or line locations by the given amounts *dx/dy*
     [Default is no offset]. If *dy* is not given it is set equal to *dx*.
+    You may append dimensional units from **c**\ \|\ **i**\ \|\ **p** to each value.
 
 .. _-E:
 
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -78,6 +78,7 @@ two ways to ensure that GMT understands which unit you intend to use:
 
 #. Set the parameter :term:`PROJ_LENGTH_UNIT` to the desired unit. Then,
    all dimensions without explicit units will be interpreted accordingly.
+   By default, GMT always initializes :term:`PROJ_LENGTH_UNIT` to cm.
 
 The latter method is less robust as other users may have a different
 default unit set and then your script may not work as intended. For portability,
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -918,11 +918,11 @@ Projection Parameters
         The default **Vincenty** is accurate to about 0.5 mm.
 
     **PROJ_LENGTH_UNIT**
-        Sets the unit length. Choose between **c**\ m, **i**\ nch, or
-        **p**\ oint [c (or i)]. Note that, in GMT, one point is defined
-        as 1/72 inch (the PostScript definition), while it is often
-        defined as 1/72.27 inch in the typesetting industry. There is no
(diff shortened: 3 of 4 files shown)


doc/rst/source/plot_common.rst_
```diff
@@ -70,6 +70,7 @@ Optional Arguments
 **-D**\ *dx*/*dy*
     Offset the plot symbol or line locations by the given amounts *dx/dy*
     [Default is no offset]. If *dy* is not given it is set equal to *dx*.
+    You may append dimensional units from **c**\ \|\ **i**\ \|\ **p** to each value.
```
joa-quim, line 73: Please clarify also which *c* or *p* is the default unit. That is **-D4** means 4 cm or 4 points?

PaulWessel:
Well, that is always subject to **PROJ_LENGTH_UNIT** settings so for US users it may be inches.  System default is of course cm which is explained in the cookbook.

joa-quim:
OK, so that means the default is **cm**. I find the description in **PROJ_LENGTH_UNIT** is misleading (as I mentioned when we talked about it in the themes-doc). The point is that if the user does not choose to change the value of **PROJ_LENGTH_UNIT** the default is **cm** because that's hardwired in the code but the parameters description does not say that.

PaulWessel:
We should look at the section on Units in the cookbook.  We should clear up any confusion of course, but since you can select the overall default unit to be inch or points, I do not think we can say cm is the default unit for options like **-**D.  That would be a lot of places.  We _can_ say that a particular setting default is 2c since it is 2cm no matter what **PROJ_LENGTH_UNIT**.

seisman:
If I understand it correctly, the default unit is cm (or inch), except for *pen* (in point), right? We may clarify it in the Units section.

PaulWessel:
If a user has a gmt.conf in their home or !/.gmt area that says PROJ_LENGTH_UNIT inch then that changes all default units to inch.  For fonts the default unit is always points.  E.g., 12,TImes is same as 12p,Times while 1i,TImes is the same as 72p,TImes.

seisman:
> If a user has a gmt.conf in their home or !/.gmt area that says PROJ_LENGTH_UNIT inch then that changes all default units to inch

```
gmt begin map
echo 2 5 | gmt plot -R0/10/0/10 -JX10c -Baf -Sc2c -W1
gmt set PROJ_LENGTH_UNIT cm
echo 4 5 | gmt plot -R0/10/0/10 -JX10c -Baf -Sc2c -W1
gmt set PROJ_LENGTH_UNIT inch
echo 6 5 | gmt plot -R0/10/0/10 -JX10c -Baf -Sc2c -W1
gmt end show
```

image

PaulWessel:
Yes, sorry, pen and fonts are usually set in points so that is the default.  Other dimensions (offsets, shifts, etc) are under PROJ_LENGTH_UNIT control.

joa-quim:
If I get a virgin computer, go run and install GMT and do ``plot -Sc1 ...`` I get a circle with **1 cm** diameter. So **cm** is the *defacto* default.

Right, one can change it via **PROJ_LENGTH_UNIT** but that does not change the fact that cm are the default. (except for fonts and pens). And we are not telling this to the users.

PaulWessel:
I will add a sentence to the Unit description on the cookbook to say that clearly.

PaulWessel:
Now stating GMT initializes PROJ_LENTH_UNIT to cm.

joa-quim:
Sorry to insist but the CookBook seems to be much less consulted by the users. I think gmt.conf ``PROJ_LENGTH_UNIT`` it's the right place to say that.

PaulWessel:
That is find. I will update that as well.

joa-quim (approved):

PR #4737, opened 2021-02-02
Ensure 64-bit indices used in loops over nodes

PaulWessel:
Turns out it is easy to fall into the trap where one uses simple-minded stuff like

```
int row, col, band;
uint6_t node = row * col * band;
```

and relearns the hard way that the right hand side is evaluated at int precision and thus gives junk when overflowing the int 4-byte range...  The alternative is to write cumbersome expressions like

`uint6_t node = ((uint64_t)row) * ((uint64_t )col) * ((uint64_t)band);`

or simply use int64_t for the row, col, band.  That is the solution here.

This should help with the problems reported on the [forum](https://forum.generic-mapping-tools.org/t/grdimage-earth-day-30s/1288/2) as well.

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -110,7 +110,7 @@ struct GRDIMAGE_CONF {
 	bool int_mode;				/* true if we are to apply illumination */
 	unsigned int *actual_row;	/* Array with actual rows as a function of pseudo row */
 	unsigned int *actual_col;	/* Array of actual columns as a function of pseudo col */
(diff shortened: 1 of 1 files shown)


joa-quim:
> uint64_t node = ((uint64_t)row) * ((uint64_t )col) * ((uint64_t)band);

I'm sure there is a CID somewhere warning about this. At least I remember to have fixed several of these cases. And, BTW, in the above expression only one needed to be casted and the others get automatically promoted by the compiler.

joa-quim (approved):

Issue #4706, opened 2021-01-26
Add support for periodic time intervals


PaulWessel:
**Description of the desired feature**

I propose we add in support to convert absolute time coordinates into periodic time related to daily, weekly, monthly, and annual cycles.  See #4507 for the initial impetus, including plots demonstrating the use and need for this feature, even though that proposal was ultimately too generic (create custom periodicities).

Here are the principles I propose to follow:

1. The implementation will be added to the lowest-level i/o processing since input may come from ASCII files, native binary files, netCDF table files, or virtual files.  They all share those low-level functions so that is where the conversion must take place.  Any other solution will duplicate efforts and open up possibilities for bugs.
2. I believe the most suitable way of informing GMT of periodic time is via the **-f** option, since it is used to set column type, and (except when -**J** indicates absolute time) is needed to properly read files and interpret virtual tables.  The **-f** currently has the format **-f**_col_**?** where current column type flags **?** are **x** (longitude), **y** (latitude), **T** (absolute ASCII time), **t** (relative, binary time), **p**[_unit_] (projected coordinate, append unit if not meter), or **f** (floating point).  There is also the two shorthands **-fg** (for **-f**0**x**,1**y**) and **-fc** (**-f**0-1**f**) [actually, I think **p** is like that too, including two columns for map projected data].  Given these letters, and because the periodicity applies equally to absolute and relative times + epoch (since the latter is the only non-ASCII way of feeding GMT), I propose we use **-fT**|**t**[**y**|**m**|**w**|**d**] which is simple to parse.
3. The four cycles will lead to a conversion from what is initially seconds relative to the epoch to new repeating Cartesian coordinates as follows:

- daily: Yields coordinates in the 0.00000 - 86399.xxxxx range (fractional seconds)
- weekly: Yields coordinates in the 0.00000 - 6.xxxxxx range (fractional  weekdays)
- monthly: Yields coordinates in the 0.00000 - 11.xxxxx range (fractional month)
- yearly: Yields coordinates in the 0-1 range (fractional year).

4. Like the behavior of longitude with a Cartesian scaling told to deal with longitude (**-Jx**1c**d**/0.4c for example), the periodic coordinates need to wrap around given the prevailing **-R** selection.  For example, **-R**-6/6/.... would plot June through June and month = 3.44 needs to wrap and plot where -2.56 is on that axis.

[...]

PR #4746, opened 2021-02-04
Add new common option -w for wrapped time or custom Cartesian coordinates

PaulWessel:

See #4706 for the background for this PR - but note there are a few subtle differences in implementation from that long discussion. This PR introduces a new GMT common option that is used to convert an input column (typically _x_ as absolute time) to a cyclical coordinate given a specific period.  The syntax is

**-wy**|**a**|**w**|**d**|**h**|**m**|**s**|**c**_period_[/_phase_][**+c**_col_]

The **+c** modifier can be used to select another column than the default first [0 or x].  The cycle conversions work as follows:

- **y** selects yearly normalized cycle.  Time is converted to a single year with coordinates 0-0.9999999.
- **a** selects an annual cycle. Similar to **y**, but here time is consolidated by calendar month _m_ (_m_ is integers 0-11) so that all times falling within month _m_ will get a coordinate in the range _m_-_m_.99999999... The converted data lend themselves directly to analysis requiring calendar month processing.
- **w** selects a weekly cycle, where all times converted to the 0-6.99999999... weekday range.
- **d** selects a daily cycle, with all times converted to the 0-23.999999 hours range.
- **h** selects an hourly cycle, with all times converted to the 0-59.999999 minutes range.
- **m** selects a minute cycle, with all times converted to the 0-59.999999 seconds range.
- **s** selects a second cycle, with all times converted to the 0-0.9999999 seconds range.
- **c** selects a custom period cycle for Cartesian data, with the coordinates converted to a normalized 0-0.9999999 range. This code requires you to append the _period_ in question, and optionally a _phase_ if not zero [0].

When periodicity is implied via **-w**, lines exiting on one side of the plot will wrap around to the other side, similar to how lines on a map behave at periodic boundaries.  In addition, a similar plot wrapping has been implemented for the _y_-coordinate if it is selected to be the cyclic coordinate.  Selecting cyclicity for other coordinates does not support line-wrapping (e..g, for _z_) but otherwise behaves the same way as described above.

All modules that read table data and that are not specific to geographic data have been given the new **-w** option, and the documentation has been updated accordingly.  Section 4 in the cookbook (standard command line options) has a new and long section explaining the purpose of this option including five examples with code and illustrations using remote data sets.  I have started the process where the short explanations for a common option (source files explain_-_opt_.rst_) now links to the much longer discussion in the Cookbook.  I suggest we do the same for all common options.

Because this is a new option, all existing tests run as before, and the new examples in the cookbook represent a small set of scripts that test **-w**.  I hope to add at least one more example illustrating weekly cycles either using Covid-19 testing or highway traffic, and a few more tests to our test suite later.


PaulWessel:
Finished the docs by adding a final example showing off the weekly and daily wrapping using traffic data from NY:

GMT_cycle_6

```
gmt begin GMT_cycle_6
	gmt set TIME_WEEK_START Monday
	gmt subplot begin 2x2 -Fs15c/8c -M14p/6p -A -T"Verrazzano-Narrows Bridge @%34%\337@%% Brooklyn Traffic [2018-2020]" -BWSrt
 		# Raw time-series
		gmt plot @NY_traffic.txt -Wfaint,red-By+l"Vehicles/hour" -gx21600 -f0T -c
		# Weekly wrapped time-series
 		gmt plot @NY_traffic.txt -R0/7/0/17000 -Wfaint,red -ww -gx0.25 -c
		# Weekly histogram normalized for number of hours for a weekday in the 3-year data
		n_week_hours=$(gmt convert @NY_traffic.txt -ww | gmt select -Z0/0.999+c0 | gmt info -Fi -o2)
 		gmt histogram @NY_traffic.txt -R0/7/2000/4000 -Wthick -Gred -T1 -Z0+w -By+l"Vehicles/hour" -i0,1+d${n_week_hours} -ww -c
		# Daily histogram normalized for number of days in the 3-year data
		n_mondays=$(gmt convert @NY_traffic.txt -wd | gmt select -Z-0.5/0.5+c0 | gmt info -Fi -o2)
 		gmt histogram @NY_traffic.txt -R0/24/0/8000 -Wthick -Gred -T1 -Z0+w -i0,1+d${n_mondays} -wd -c
 	gmt subplot end
gmt end show
```

PaulWessel:
Hi @GenericMappingTools/core, let me know if you have any concerns about my syntax decisions etc above.  Since this is a new option its behavior is decoupled from all other things, meaning no other examples or GMT usage is affected since this stuff requires -w.  I would like to merge this in asap as the 7+ examples seem to indicate it is stable enough for being added.

PaulWessel:
Note the **TIME_WEEK_START** = Monday was required since our tests run under the US defaults which as Sunday as first day of the week...otherwise it would not be needed.

WalterHFSmith:
This is beautiful, thanks.

w

doc/scripts/GMT_cycle_6.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# Exploring NY traffic across V-N bridge, 2018-2020
+# First just show data, with gaps > 6 hours not connected
+# Notice May 2018 spikes, data gaps in the summer of 2018, Covid drop.
+gmt begin GMT_cycle_6
+	gmt set TIME_WEEK_START Monday
+	gmt subplot begin 2x2 -Fs15c/8c -M14p/6p -A -T"Verrazzano-Narrows Bridge @%34%\337@%% Brooklyn Traffic [2018-2020]" -BWSrt
+ 		# Raw time-series
+		gmt plot @NY_traffic.txt -Wfaint,red-By+l"Vehicles/hour" -gx21600 -c
```
meghanrjones, line 9:
```suggestion
		gmt plot @NY_traffic.txt -Wfaint,red -By+l"Vehicles/hour" -gx21600 -c
```

PaulWessel: oops

meghanrjones:
This is great! I am wondering how you decided which table-based modules should and should not get the new **-w** option? For example, I image that a user could want to use **-w** with gmtinfo.

PaulWessel:
I think I gave it to all except those that it makes no sense for, e.g., **gmtflexure**, **talwani2d**, etc. But if I missed some we can easily add: Just need to

1. Add w to the list in **THIS_MODULE_OPTIONS**
2. Add another [%s] and **GMT_w_OPT** to the appropriate usage GMT_Message string
3. Add w to the comma-separated list in the _GMT_Option_ closest to w in usage function
4. Add `[ |SYN_OPT-p| ] `to the _module.rst_ page
5. Add` .. include:: explain_-l.rst_` to the the end of the _module.rst_ page (or module_common.rst_ if so).

meghanrjones:
OK, I will make a PR to this branch with **-w** in gmtinfo, rose, and any other modules based on those steps.

src/gmt_init.c
```diff
@@ -2473,6 +2472,63 @@ bool gmtinit_parse_t_option (struct GMT_CTRL *GMT, char *item) {
...
+
+	switch (arg[0]) {	/* Look at which valid code we got */
+		case 's': GMT->current.io.cycle_operator = GMT_CYCLE_SEC; break;
+		case 'm': GMT->current.io.cycle_operator = GMT_CYCLE_MIN; break;
+		case 'h': GMT->current.io.cycle_operator = GMT_CYCLE_HOUR; break;
+		case 'd': GMT->current.io.cycle_operator = GMT_CYCLE_DAY; break;
+		case 'w': GMT->current.io.cycle_operator = GMT_CYCLE_WEEK; break;
+		case 'a': GMT->current.io.cycle_operator = GMT_CYCLE_ANNUAL; break;
+		case 'y': GMT->current.io.cycle_operator = GMT_CYCLE_YEAR; break;
+		case 'p': GMT->current.io.cycle_operator = GMT_CYCLE_CUSTOM;
```
meghanrjones, line 2502:
```suggestion
		case 'c': GMT->current.io.cycle_operator = GMT_CYCLE_CUSTOM;
```
Docs say that this is option **-wc**

PaulWessel: Thanks, coding changed over time as I felt c for cycle made more sense.

meghanrjones (approved):
To confirm, **-wc** is not supported for time?

PaulWessel:
Correct, for Cartesian only as is.

meghanrjones:
OK, and there's no way for GMT to determine whether the input is cartesian versus time? For example, to warn about using **-wy** on cartesian data or **-wc** on time data.

PaulWessel:
It is for ascii data where we learn if there ie ABSTIME formats. Sorry, in the middle of zoom so cannot think and type.  Let me think about -wc for time data before we decide.

PaulWessel:
OK, I cannot see anything in the code that would precent a user from setting a custom period of -wc848484/99 and do that with absolulte time data (which is internally seconds).  It will be treated like Cartesian in that sense (no special annotation tricks).  So let us NOT state that it is restricted to any particular data type.

meghanrjones:
OK, I think when making that change it's worth stating in the docs that TIME_UNIT does not apply to **-wc** and including the explanation about absolute time data being seconds internally because the user doesn't get an error if something like ```gmt convert @NY_traffic.txt -wc12d/1d > new_c.txt``` is attempted.

PaulWessel:
Thanks, yes good idea.  I will have to hold that thought until 4-5 pm here unfortunately but will do that.  I think with -wc the data is still parsed in and TIME_UNIT sets what the internal data looks like.  But from there on it is like any other Cartesian data.  So yes, -w1d/6h is not going to work or be allowed [I think].

meghanrjones:
OK, one last comment before leaving this be. It would be nice if **-wc** checked that the *period* is a float or integer. I find it weird, for example, that this gives an output without error: ```gmt convert @NY_traffic.txt -wc800000s8 > new_c.txt```

PaulWessel:
OK, I think we are about ready now.  Your case and similar commands will run into stern checking:

```
gmt convert @NY_traffic.txt -wc800000/1d > new_c.txt
gmtconvert [ERROR]: Cannot convert 1d to floating point as it contains invalid characters (d).
gmtconvert [ERROR]: Option -w parsing failure. Correct syntax:
	-wa|y|w|d|p<period][/<phase>][+c<col>]
	-w Wrapped selected column [0] with specified cyclicity:
	   Absolute time: Append y|a|w|d|h|m|s for year, annual (by month), week, day, hour, minute, or second cycles.
	   Alternatively append c<period>[/<phase>] for custom cyclicity.
	   Select another column than x via +c<col>.
gmtconvert [ERROR]: Offending option -wc800000/1d
```

PR #4748, opened 2021-02-04
Give -i a complimentary +ddivide modifier to simplify scaling with inverse values

PaulWessel:
It is just **way** easier to divide by 60 using **+d**60 than to scale by 0.01666666 via **+s**0.0166666666, especially when you also may have to compute what the inverse of the factor is in the script, such as 7 or 52, or 3300.


joa-quim (approved):
I thought this had been done already. At least I have a recall of the subject.

PaulWessel:
You are probalbly remembering -Tmin/max/inc+i for inverse inc.  I decided it was too strange to do -i0+s10+i vs -i0+d10.

PR #4749, opened 2021-02-04
If -f indicates abstime then dont fuss that -J does not

PaulWessel:
We used to get this annoying message from _gmt_map_setup_ if **-f** had indicated absolute time but **-JX** did not have that 'T':

`Option -JX|x: Your x-column contains absolute time but -JX|x...T was not specified!`

It was just a warning, but then it continued, taking no corrective action, and screwing things up.  This PR now does two things:

1. Sets the missing linear projection type to **GMT_TIME** and relegating that message to **GMT_DEBUG**.
2. Fixes a related issue in _gmt_auto_frame_interval_ so that it knows the axis is absolute time so that a 6 month interval is not attempted to be used as 6 seconds and giving 1000000 annotations....

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -3292,9 +3292,9 @@ GMT_LOCAL void gmtio_assign_col_type_if_notset (struct GMT_CTRL *GMT, unsigned i
 		}
 	}
 	else {	/* logical equal physical, so no column shopping */
-		if (!GMT->current.io.col_set[GMT_IN][col])
+		if (GMT->current.io.col_set[GMT_IN][col] == GMT_IS_UNKNOWN)
 			gmt_set_column_type (GMT, GMT_IN, col, type);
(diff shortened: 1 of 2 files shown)


joa-quim (approved):

PaulWessel:
I should say that this is the simple fix.  A related issue will have to come later, which is when modern mode auto-detect **-R** from a dataset, and that dataset has abstime and there are no -f or -J to tell us.  The call to **gmtinfo** will detect that this is temporal data, but of course the wesn returned back is binary so in seconds since 1970.  The calling module does not know this, so it plots the seconds as a Cartesian plot.  I need a way to pass the abstime flag back...   As example,

`gmt plot NY_traffic.txt -Wfaint,red -B -pdf t`

should default to 15cx15cm plot (it does) and get -R from the file (it kinda does) but ends up with NaNs for west/east..

PR #4752, opened 2021-02-05
Return col-types of x/y when gmtinfo is called by gmt_init_modules

PaulWessel:
When we must compute a missing **-R** from a dataset we call **gmtinfo**.  Unfortunately, it does not return information such as what the x and y data types are, which is important, expecially for absolute time.  This PR creates a back-door **-C**coltypes argument that returns the needed quantities in 2 extra columns which allows the proper data types to be set in the calling module.  This PR now allows this command to run correctly and auto-select absolute time range and tick intervals without passing any **-f** or **-J** options but using the built-in defaults:

`gmt plot @NY_traffic.txt -Wfaint,red -B -pdf t`

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -13441,8 +13441,10 @@ GMT_LOCAL int gmtinit_get_region_from_data (struct GMTAPI_CTRL *API, int family,
 						return API->error;	/* Failure to make new option or append to list */
 				}
 			}
-			if ((tmp = GMT_Make_Option (API, 'C', "coltypes")) == NULL || (head = GMT_Append_Option (API, tmp, head)) == NULL)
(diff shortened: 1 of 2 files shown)


joa-quim (approved):
What I don't like so much is that we are wasting a letter. If it's to be hidden, numbers can be used as optional characters to. E.g. **-1** can be made a valid option too.

PaulWessel:
We are not wasting a letter.  **-C** has ben there forever and means column output. We are just piggy-backing off this option with a long word.

PaulWessel:
But, I get your point.  I will change this to some -number to leave the **-C** uncluttered.

joa-quim:
Ah, sorry. Overloading is ok.

PR #4759, opened 2021-02-06
Improved custom symbols

### LINKED ISSUE #4755: Have EPS custom symbols in decorated lines

I'm a heavy user of `-S~:+i+sk` (Thank you again, @PaulWessel!) with custom symbols written [the GMT way](https://docs.generic-mapping-tools.org/latest/cookbook/custom-symbols.html#macro-commands). I just learnt that one awesome feature of `plot -Sk` doesn't work: I can't use an EPS custom symbol with `-S~`:

`plot [ERROR]: Cannot use EPS custom symbol in the decorated line specification: circle_star`

When my symbols get a bit complicated (more curved lines than I'm willing to calculate the parameters for, parts of an ellipsoid or similar), and no logic is required in the symbol, I like to take the lazy route via an EPS custom symbol. 

Therefore I'd greatly appreciate if you could kindly make EPS custom symbols usable in decorated lines. 

_Some code to play with:_

```
cat > path1.txt << END
-1.5 -1
 0    1
 1.5  1.2
END

gmt begin eps_symbol_line
  gmt plot path1.txt -W4p,black,8_4
  gmt plot path1.txt -S~n1:+i+skcircle_star/2c
gmt end show
```

[Download circle_star.eps.zip](https://github.com/GenericMappingTools/gmt/files/5935747/circle_star.eps.zip) (~1 KB, just a quick'n'dirty example for an EPS symbol)

PaulWessel:

This PR improves and simplifies the handling of custom symbols.  In the past, giving an EPS file as a custom symbol mean we handled it separately.  The downside of that was that it could not be used as a symbol used with decorated lines.  By adding a new operator **P** for "_Placing EPS file_" to the macro language, plotting an EPS can be part of custom symbols, and if a single EPS is given as the symbol we simply convert under the hood to a custom symbol placing the EPS.  With those changes, decorated lines can use custom symbols, finally shedding some light on _how_ the chicken crossed the road (the jury is still out on the "why").

chicken_road

Closes #4755.


joa-quim (approved):
Just for reference. https://www.galloportugal.com/

KristofKoch:
I got some trouble building it:

```
[  0%] Built target gen_gmt_suppl_supplements_headers
[  3%] Built target script2verbatim
[  3%] Generating gmt_dimensions.h
[  6%] Built target gshhg_version
[  6%] Built target gen_gmt_headers
[  6%] Generating gmt_dimensions.h
make[2]: *** No rule to make target `/usr/lib/libz.dylib', needed by `src/libpostscriptlight.6.2.0.dylib'.  Stop.
make[1]: *** [src/CMakeFiles/pslib.dir/all] Error 2
make: *** [all] Error 2
[  3%] Generating gmt_dimensions.h
[  3%] Built target gen_gmt_headers
[  3%] Built target gen_gmt_suppl_supplements_headers
[  3%] Generating gmt_dimensions.h
make[2]: *** No rule to make target `/usr/lib/libz.dylib', needed by `src/libpostscriptlight.6.2.0.dylib'.  Stop.
make[1]: *** [src/CMakeFiles/pslib.dir/all] Error 2
make: *** [all] Error 2
$
```

PaulWessel:
Libz is not us, maybe wipe build and try again?

joa-quim:
I think it's a GDAL dependency. Is it homebrew? Brew's formula for GDAL is crazy.

KristofKoch:
Wiping `build` did the trick – maybe upgrading the OS was the culprit (first build after upgrade). Now on to the test drive …

KristofKoch:
Seems to work great!

EZE-FRA

Some thoughts (not sure if EPS specific):

* Have aa way to define padding around the symbol where the line is omitted - currently the line runs "through" (under) the symbol blocking details (see nose and tail of aircraft)
* Have the symbol align in the direction of the line. First coordinate is Frankfurt (upper right), second Buenos Aires (lower left) -> aircraft symbol has nose pointing in SW direction towards Buenos Aires.

joa-quim:
Aha. I had that impression on your first version of the plain but now I can confirm it. The lights in the pilot's cabin are on.

joa-quim:
> Have a way to define padding around the symbol where the line is omitted

I think that is dangerous because it would have to be a rectangular clipping box and in this case it would clear the lon,lat grid as well.

PaulWessel:
There is no way to add some fancy clipping as it is with text labels (quoted lines), but you can certainly plot white to "white things out" beneath the symbol by making that part of your symbol or pre-plot some other blob before you plot your plane.

KristofKoch:
The old trick by giving it a white outline still works but is not ideal (landmass, coast lines). Found that `-S~+s+p` doesn't work (somewhat expected). Maybe clarify in the docs?

EZE-FRA2

And yes, @joa-quim - lights on the flight deck are on. But why would it be a problem if the gridlines are cut as well? Nothing a `basemap` call can't fix?

PaulWessel:
What do you mean by +s+p?  You mean you expected an error message like "no symbol specified"?

joa-quim:
No problem at all with the lights. The problem I referred was if GMT clipped around the symbol it would clipt a lot more than in your last example and no help from basemap since the symbol is plot after the coastlines, grids, etc.

PaulWessel:
> * Have the symbol align in the direction of the line. First coordinate is Frankfurt (upper right), second Buenos Aires (lower left) -> aircraft symbol has nose pointing in SW direction towards Buenos Aires.

So isn't that what it does via +a?

> +aangle
> Force annotations at a fixed angle, +an for line-normal, or +ap for line-parallel [Default].

We do compute the line azimuth and that is what causes the rotation.

KristofKoch:
> What do you mean by +s+p? You mean you expected an error message like "no symbol specified"?

I wanted to modify the EPS symbol outline with a pen definition via `+p`. Something like `-S~n1:+skb748_bw/8c+p2p,white`. Maybe a bit too ambitious.

> So isn't that what it does via +a?

Not really. See panels a) and c) in example below for the current `+ap` behavior. 

In a), the line starts in the West and ends in the East so the symbol points to the East in the direction of the line. Exactly what I want.

In panel c) on the other hand the line starts in the East and ends in the West so I'd like to have the symbol point towards the West. Basically indicating the direction the line was constructed.

I would find this useful to indicate a direction of travel or similar things. See the globe plot above - as I understand it there is no easy way to get the aircraft symbol to fly from Frankfurt to Buenos Aires.

What do you think of such a behavior @PaulWessel?



symbol_orientation

Some code to play with:

```
cat > dir.def << END
-0.5   0.25 M
 0.5   0    D
-0.5  -0.25 D
-0.5   0.25  0.3 c -W- -Gred
-0.5  -0.25  0.3 c -W- -Ggreen
END

cat > pathWE+ap.txt << END
>
-0.5   0.4
 0.5   0.6
>
-0.5   0
 0.5   0
>
-0.5  -0.4
 0.5  -0.6
END

cat > pathEW+ap.txt << END
>
 0.5   0.6
-0.5   0.4
>
 0.5   0
-0.5   0
>
 0.5  -0.6
-0.5  -0.4
END

cat > pos.txt << END
0 0
END

gmt begin symbol_orientation png
  gmt subplot begin 3x2 -Fs8c -M5p/60p -A -Bwsentr -Bag1d -R-1/1/-1/1 -JM8c
    gmt subplot set
      gmt basemap -B+t"West to East"+s"@%10%-S~n1:+sk+ap@%%"
      gmt plot pathWE+ap.txt -W2p,darkgrey,8_4 -S~n1:+skdir/1c+p2p,black+gwhite+ap
      gmt plot pathWE+ap.txt -SqN-1:+i+l" START "
      gmt plot pathWE+ap.txt -SqN+1:+i+l" END "
    gmt subplot set
      gmt basemap -B+t"West to East"+s"@%10%-S~n1:+sk+an@%%"
      gmt plot pathWE+ap.txt -W2p,darkgrey,8_4 -S~n1:+skdir/1c+p2p,black+gwhite+an
      gmt plot pathWE+ap.txt -SqN-1:+i+l" START "
      gmt plot pathWE+ap.txt -SqN+1:+i+l" END "
    gmt subplot set
      gmt basemap -B+t"East to West"+s"@%10%-S~n1:+sk+ap@%%"
      gmt plot pathEW+ap.txt -W2p,darkgrey,8_4 -S~n1:+skdir/1c+p2p,black+gwhite+ap
      gmt plot pathEW+ap.txt -SqN-1:+i+l" START "
      gmt plot pathEW+ap.txt -SqN+1:+i+l" END "
    gmt subplot set
      gmt basemap -B+t"East to West"+s"@%10%-S~n1:+sk+an@%%"
      gmt plot pathEW+ap.txt -W2p,darkgrey,8_4 -S~n1:+skdir/1c+p2p,black+gwhite+an
      gmt plot pathEW+ap.txt -SqN-1:+i+l" START "
      gmt plot pathEW+ap.txt -SqN+1:+i+l" END "
    gmt subplot set
      gmt basemap -B+t"Basic Symbol"+s"@%10%-Sk@%%"
      gmt plot pos.txt -Skdir/1c -W2p,black -Gwhite
  gmt subplot end
gmt end show
```

PaulWessel:
OK, about the outline first: Since the EPS file is just a black box we place at the given coordinate, scaled to the size, and possibly rotated, we have no information about shape.  There is no outline to draw, etc.  Of course, in your cases you could make a symbol that is a mix of placing an EPS plus drawing an outline of background. If we need more documentation to make this clear then let me know where it should go.

PaulWessel:
As for the orientation.  I agree that panel c) is wrong.  We probably do not compute the angle properly.  it should always follow the line coordinates.  That is a good test case I can play with.

KristofKoch:
That was my reasoning as well that you just place the black box EPS and there is no way to interfere with it via the usual options. But with you one never knows so I gave it a try. ;-) 

You want me to open an issue on the orientation thing?

PaulWessel:
Yes, please do so we can better track it.

KristofKoch:
Opened #4763 for it.

Issue #4763, opened 2021-02-07
plot -S~:+sk+ap – Symbol should always follow along the line


KristofKoch:
**Description of the problem**

When placing symbols on lines with `plot -S~:+sk+ap` the symbol is not always oriented in the direction of line construction. 

**Full script that generated the error**

```
cat > dir.def << END
-0.5   0.25 M
 0.5   0    D
-0.5  -0.25 D
-0.5   0.25  0.3 c -W- -Gred
-0.5  -0.25  0.3 c -W- -Ggreen
END

cat > pathWE+ap.txt << END
>
-0.5   0.4
 0.5   0.6
>
-0.5   0
 0.5   0
>
-0.5  -0.4
 0.5  -0.6
END

cat > pathEW+ap.txt << END
>
 0.5   0.6
-0.5   0.4
>
 0.5   0
-0.5   0
>
 0.5  -0.6
-0.5  -0.4
END

cat > pathWE_v+ap.txt << END
-0.5  0.5
 0    0.5
 0   -0.5
 0.5 -0.5
END

cat > pathEW_v+ap.txt << END
 0.5 -0.5
 0   -0.5
 0    0.5
-0.5  0.5
END

cat > pathWE_n+ap.txt << END
-0.5  0.5
-0.5 -0.5
 0.5  0.5
 0.5 -0.5
END

cat > pathEW_n+ap.txt << END
 0.5 -0.5
 0.5  0.5
-0.5 -0.5
-0.5  0.5
END

cat > pathWE_z+ap.txt << END
-0.5  0.5
 0.5  0.5
-0.5 -0.5
 0.5 -0.5
END

cat > pathEW_z+ap.txt << END
 0.5 -0.5
-0.5 -0.5
 0.5  0.5
-0.5  0.5
END

gmt begin symbol_orientation png
  gmt subplot begin 4x2 -Fs8c -M5p -A -Bwsentr -Bag1d -R-1/1/-1/1 -JM8c -T"Symbol Orientation @%10%-S~:+sk+ap@%%"
    gmt subplot set
      gmt basemap
      gmt plot pathWE+ap.txt -W2p,darkgrey,8_4 -S~n1:+skdir/1c+p2p,black+gwhite+ap
      gmt plot pathWE+ap.txt -SqN-1:+i+l" START "
      gmt plot pathWE+ap.txt -SqN+1:+i+l" END "
    gmt subplot set
      gmt basemap
[...]

PR #4765, opened 2021-02-07
Ensure decorated symbols follow the direction of the line (with +ap|n)

PaulWessel:

When the decorated symbol option was added, it was largely based on the quoted text (and contour) machinery.  In those cases, text needs to be readable so only the orientation of the line was required.  However, with symbols we will want to allow the symbols to be aligned with the direction of the line if **+ap** is specified.   This is what #4763 is all about.  The solution was to be more careful when computing the least-squares slope by also considering sums of x and y for the points involved.  This PR fixes the initial issue and I turned that issue into two new tests (one for geographic projections (which resamples lines) and one for Cartesian (no resampling)).

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -1790,9 +1790,10 @@ GMT_LOCAL uint64_t gmtsupport_determine_circle (struct GMT_CTRL *GMT, double x0,
 }
 
 /*! . */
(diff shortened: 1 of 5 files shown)


KristofKoch (approved):
Tested it with several different cases, worked like a charm. Thank you, @PaulWessel

joa-quim (approved):

PR #4766, opened 2021-02-08
Change 18p pen name from "obese" to "wide"

liamtoney:
**Description of proposed changes**

The [GMT Code of Conduct pledge](https://github.com/GenericMappingTools/gmt/blob/0819a199bc04bc651ba67030c9b138a4c3745737/CODE_OF_CONDUCT.md#our-pledge) states that

> ...we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, **body size**, disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, religion, or sexual identity and orientation.

With this in mind, I propose that the name "obese" for the 18p pen width be changed. I have suggested "wide" as a replacement, but that was just the first option that came to mind and seemed reasonable. I am sure there are others.

I am happy to discuss this further. GMT developers: Thank you so much for developing this amazing tool! This simple PR's purpose is only to make GMT a little more welcoming to **all** users.

PR #4767, opened 2021-02-08
Let existing scripts using pen=obese we processed correctly

PaulWessel:
Although #4766 change its name to "wide", we need to let existing scripts be processed to retain backward compatibility without discussing this in the docs.

Just checking with you all if this is OK, or if in this case a harder non-backwards-compatible change is preferred?

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -937,12 +937,12 @@ int gmt_getfonttype (struct GMT_CTRL *GMT, char *name) {
 
 	if (!name[0]) return (-1);
 	if (!isdigit ((unsigned char) name[0])) {	/* Does not start with number. Try font name */
-		int ret = -1;
+		int ret = GMT_NOTSET;
 		for (i = 0; i < GMT->session.n_fonts && strcmp (name, GMT->session.font[i].name); i++);
 		if (i < GMT->session.n_fonts) ret = i;
 		return (ret);
 	}
-	if (!isdigit ((unsigned char) name[strlen(name)-1])) return (-1);	/* Starts with digit, ends with something else: cannot be */
+	if (!isdigit ((unsigned char) name[strlen(name)-1])) return (GMT_NOTSET);	/* Starts with digit, ends with something else: cannot be */
 	return (atoi (name));
 }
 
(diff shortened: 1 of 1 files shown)


liamtoney (commented):
I'll let others chime in on this.

meghanrjones:
Given the comment description and the documentation update, I think this resolution is fine but agree that we should let anyone veto it. Just a reminder that if anyone prefers to comment about this privately, contact info for Paul and Leo are provided in the [GMT code of conduct](https://github.com/GenericMappingTools/gmt/blob/master/CODE_OF_CONDUCT.md).

joa-quim:
I confess to have troubles understanding why one call say *fat pen* but should not use *obese pen*. Furthermore, in Portuguese *obese* is the educated word whilst is *fat* that can be used in an offensive way though normally is not offensive, only if intention is put on it.

But we are talking about lines in plots, not people.

PaulWessel:
Yes, we are talking about lines. but obese in the Merriam-Webster dictionary says " having excessive body fat" so it is specifically about people.  So I am fine with changing to "wide".  I also think the resolution of allowing backwards support for anyone's script that use -Wobese be processed correctly (since we have allowed this option for years).  If anyone feel we should go further and refuse to honor -Wobese then I am happy to enterain such voices either here or privately.  I do not think I have ever used this one myself since I would have to look up what it is in terms of points anyway.  I do -W30p etc instead.

PaulWessel:
I should add that I added obese just to get an even number of items in the pen [table](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#tbl-pennames)...

liamtoney:
> I confess to have troubles understanding why one call say _fat pen_ but should not use _obese pen_.

I actually think that _fat[er|est]_ could be changed as well. I didn't suggest it in #4766 because I think it would break more scripts, and I think that _obese_ is "more" inappropriate. I'm biased: I rarely, if ever, use the pen names, instead opting for e.g. `-W10p` (as @PaulWessel mentions).

> Furthermore, in Portuguese _obese_ is the educated word whilst is _fat_ that can be used in an offensive way though normally is not offensive, only if intention is put on it.

I agree that in vacuum, _obese_ is not necessarily inappropriate. I just don't believe it has a place in **this** context.

> But we are talking about lines in plots, not people.

Yes. But **people** use the code, and — in particular — read the docs.

joa-quim:
I never used -Wobese either and had no intention even before this discussion, so for me it can go.  Just don't appreciate that much this route.

PaulWessel:
At least the "fat" name has the minor benefit of also being applied to other things than humans, eg., a fat book, fat wallet, a fat bank account.

joa-quim:
and is often used in papers fig captions

liamtoney:
> At least the "fat" name has the minor benefit of also being applied to other things than humans, eg., a fat book, fat wallet, a fat bank account.

Agreed. And the point here is not, of course, to get lost in the semantics and hair splitting. It's about being as welcoming as we can be. I feel that the proposed change as-is is a good start — we can revisit other aspects later if needed.

PaulWessel:
OK, that is all good.  Since there is nothing left mentioning obese in usage messages or documentation or scripts we supply I think we are  fine for now.  Closing this issue but anyone should feel OK opening it again or a new one.  Perhaps there are other aspects that can have issues too.  I do remember a few years ago we got email from a student of native american ancestry who objected to the two color names Indianred and Navajowhite.  I was able to show that these names are in the X11 RGB table (so not ours) and that they derive from the continent of India (indianred) and that the other has to do with the color of some characteristic pottery.  However, now I an not so sure those answers that welcoming...

PaulWessel:
As discussed, I will close this and merge as is.

PR #4768, opened 2021-02-08
Ensure output format for time is changed if worked on

PaulWessel:
See this forum [post](https://forum.generic-mapping-tools.org/t/deviating-time-unit-default-in-gmt-math-q-mode-just-a-note/1315/2) for background.  This PR implements consistency in how columns with input absolute time is treated on output:

1. If more than one column is involved  and **-C** is _not_ set to include the time column then output will be absolute time
2. If **-Q** is used then the time column is modified by calculations and we set output to relative time.
3. If **-C** involves the time column then format is changed to relative time.

Note of these changes take place if the user has used **-f** or **-fo** to set specific format.

### DIFF
--- a/doc/rst/source/gmtmath.rst
+++ b/doc/rst/source/gmtmath.rst
@@ -705,7 +705,7 @@ contains absolute time, then the default output format for any *other* columns c
 absolute time will be reset to relative time.  Likewise, in scalar mode (**-Q**) the
 time column will be operated on and hence it also will be formatted as relative
 time.  Finally, if **-C** is used to include "time" in the columns operated on then
-we likewise will reset that column's formatting. The user can override this behavior with a
+we likewise will reset that column's format to relative time. The user can override this behavior with a
 suitable **-f** or **-fo** setting.  **Note**: We cannot guess what your operations on the
 time column will do, hence this default behavior.  As examples, if you are computing time differences
 then clearly relative time formatting is required, while if you are computing new absolute times
--- a/doc/rst/source/gmtmath.rst
+++ b/doc/rst/source/gmtmath.rst
@@ -701,7 +701,7 @@ Absolute Time Column(s)
 -----------------------
 
 If input data have more than one column and the "time" column (id set via **-N** [0])
-in contains absolute time, then the default output format for any *other* columns containing
+contains absolute time, then the default output format for any *other* columns containing
 absolute time will be reset to relative time.  Likewise, in scalar mode (**-Q**) the
 time column will be operated on and hence it also will be formatted as relative
 time.  Finally, if **-C** is used to include "time" in the columns operated on then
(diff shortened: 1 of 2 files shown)


doc/rst/source/gmtmath.rst
```diff
@@ -697,6 +697,21 @@ in the data then only those columns will be updated, hence the unspecified colum
 On the other hand, if you load the file first and then issue **-C**\ *cols* then the unspecified
 columns will have been loaded but are then ignored until you undo the effect of **-C**.
 
+Absolute Time Column(s)
+-----------------------
+
+If input data have more than one column and the "time" column (id set via **-N** [0])
+in contains absolute time, then the default output format for any *other* columns containing
```
meghanrjones, line 704:
```suggestion
contains absolute time, then the default output format for any *other* columns containing
```

doc/rst/source/gmtmath.rst
```diff
@@ -697,6 +697,21 @@ in the data then only those columns will be updated, hence the unspecified colum
...
 
+Absolute Time Column(s)
+-----------------------
+
+If input data have more than one column and the "time" column (id set via **-N** [0])
+contains absolute time, then the default output format for any *other* columns containing
+absolute time will be reset to relative time.  Likewise, in scalar mode (**-Q**) the
+time column will be operated on and hence it also will be formatted as relative
+time.  Finally, if **-C** is used to include "time" in the columns operated on then
+we likewise will reset that column's formatting. The user can override this behavior with a
```
meghanrjones, line 708:
```suggestion
we likewise will reset that column's format to relative time. The user can override this behavior with a
```

meghanrjones (approved):
This is good. I think it would also be helpful to make a time-input tutorial and/or add a 'time units' section to the [gmt units section](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#gmt-units) of the cookbook that provides some examples of different time inputs and transformations. This can be a separate issue and would be a good fit for the 'good first issue' and 'help wanted' categories.

PaulWessel:
Thanks.  One unfinished business in GMT was to implement [ISO time duration formatting](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals).  There is a syntax for this but until we add it the only output for relative time will be float (in **TIME_UNIT** [s]).  I imagine this would be another type to join t and T in -f, say e for elapsed time.

meghanrjones:
> Thanks. One unfinished business in GMT was to implement [ISO time duration formatting](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals). There is a syntax for this but until we add it the only output for relative time will be float (in **TIME_UNIT** [s]). I imagine this would be another type to join t and T in -f, say e for elapsed time.

Should I make an issue to track this? I didn't realize that the ISO time duration formatting was in the works.

PaulWessel:
Sure, make it a feature request. "in the works" == "Paul has considered it and found the Wiki page but got side-tracked"

PR #4769, opened 2021-02-08
Add -w option to more modules

meghanrjones:

Adds **-w** to ```rose``` and ```gmtinfo```. To confirm whether all necessary modules are complete, modules that have a *table* as a required input but do not have **-w** as an option are listed below.

- [ ] clip
- [ ] contour
- [ ] filter1d
- [ ] fitcircle
- [ ] gmt2kml
- [ ] gmtconnect
- [ ] gmtsimplify
- [ ] gmtspatial
- [ ] gmtsplit
- [ ] mapproject
- [ ] project
- [ ] spectrum1d
- [ ] sph2grd
- [ ] sphdistance
- [ ] sphinterpolate
- [ ] sphtriangulate
- [ ] ternary

**Reminders**

### DIFF
--- a/doc/rst/source/gmtinfo.rst
+++ b/doc/rst/source/gmtinfo.rst
@@ -35,6 +35,7 @@ Synopsis
 [ |SYN_OPT-qi| ]
 [ |SYN_OPT-r| ]
 [ |SYN_OPT-s| ]
+[ |SYN_OPT-w| ]
 [ |SYN_OPT-:| ]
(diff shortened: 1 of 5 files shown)


PaulWessel:
I think this is OK.  If we get requests to add -w to, say contour then we can do so, but none of these seems sensible to me, and many (sph*, etc) are only used with geographical data.

PaulWessel (approved):

Issue #4771, opened 2021-02-08
math -T with time


delaviel-anger:
> gmt -v 6.2.0_bb0b2d9_2020.12.14 on macOS Big Sur

When trying to create a vector of timestamps (e.g 1 over 4 Jan 1st from 1980 to 2019)
```
gmt math -o0 -T1980-01-01T/2019-12-31T/4y T =
00T00:00:00.000000000
00T00:00:00.000000000
00T00:00:00.000000000
00T00:00:00.000000000
00T00:00:00.000000000
00T00:00:00.000000000
00T00:00:00.000000000
00T00:00:00.000000000
00T00:00:00.000000000
00T00:00:00.000000000
```

However the *inc***d** version works :
```
gmt math -o0 -T1980-01-01T/2019-12-31T/1461d T =
1980-01-01T00:00:00
1984-01-01T00:00:00
1988-01-01T00:00:00
1992-01-01T00:00:00
1996-01-01T00:00:00
2000-01-01T00:00:00
2004-01-01T00:00:00
2008-01-01T00:00:00
2012-01-01T00:00:00
2016-01-01T00:00:00
```
After some tests, it seems that only `d` is properly working.
`w` acts like seconds
`o` and `y` have the output bug

Thanks :)

meghanrjones:
This bug was introduced by this [merge commit](https://github.com/GenericMappingTools/gmt/commit/3ebebec8554f7368f8300ff5f169901b096be357) (PR: https://github.com/GenericMappingTools/gmt/pull/4409). I'll take a look and will ping for help if needed.

meghanrjones:
Fixed by #4778

meghanrjones:
>  `w` acts like seconds

Turns out weeks isn't supported, #4778 updates the documentation accordingly.

PR #4778, opened 2021-02-09
Limit precision for FORMAT_CLOCK_OUT reformatting

meghanrjones:

This is the simplest fix (do not go all the way to nanoseconds in ```gmtio_get_precision_width```) for [issue 4771](https://github.com/GenericMappingTools/gmt/issues/4771), but still needs improvement.

**Description of problem**
In ```gmt_io.c```, if ```gmtio_get_precision_width``` returns k > 6, the new **FORMAT_CLOCK_OUT** (hh:mm:ss.xxxxxxx) will be >16 characters. This is a problem because all strings used to format date/clock output must be 16 characters or less, as shown here:

```
void gmt_format_abstime_output (struct GMT_CTRL *GMT, double dt, char *text) {
	char date[GMT_LEN16] = {""}, tclock[GMT_LEN16] = {""};
```

**To do**
- [x] Add a return in ```gmtio_check_abstime_format``` for cases like ```gmt math -o0 -T1980-01-01T/2019-12-31T/4y T =``` which are not at a risk of time-series precision loss
- [x] Add new tests

Fixes # 4471

**Reminders**


PaulWessel (approved):
THis is a good fix. THanks!

src/gmt_io.c
```diff
@@ -4646,8 +4646,8 @@ GMT_LOCAL int gmtio_get_precision_width (struct GMT_CTRL *GMT, double x) {
 GMT_LOCAL void gmtio_check_abstime_format (struct GMT_CTRL *GMT, struct GMT_DATASET *D) {
 	bool abstime_found = false;
 	unsigned int col, row;
-	int w_max = 0, this_w;
-	double sub;
+    int w_max = 0, this_w;
+    double sub, sub_max = 0;
```
seisman, line 4650: Just a reminder: you're using spaces while @PaulWessel uses tabs for indentations.

src/gmt_io.c
```diff
@@ -4667,12 +4667,13 @@ GMT_LOCAL void gmtio_check_abstime_format (struct GMT_CTRL *GMT, struct GMT_DATA
...
 		for (row = 0; row < MIN (S->n_rows, 20); row++) {	/* Maximum 20 rows are examined */
 			sub = S->data[col][row] - floor (S->data[col][row]);	/* Fractional second */
+            if (sub > sub_max) sub_max = sub;
 			if (gmt_M_is_zero (sub)) continue;	/* Not checking zeros for width */
 			if ((this_w = gmtio_get_precision_width (GMT, sub)) > w_max)
 				w_max = this_w;
 		}
 	}
-	if (w_max) {	/* Need to append w_max "x" to the default format */
+	if (w_max && (sub_max >= 1E-6)) {	/* Need to append w_max "x" to the default format */
```
meghanrjones, line 4676: @PaulWessel, can you provide your opinion on whether this is added criteria is a good idea? The benefit is that for most cases like ```gmt math -T``` this will prevent precision being added to FORMAT_CLOCK_OUT when the maximum error is less than the least significant digit. The risk is that we're only testing for the first 20 rows, so if the error systematically grows over the course of the record it could be problematic.

src/gmt_io.c
PaulWessel: Ah yes, that problem.... Yes, lets stick to tabs and developers can configure their editor to give whatever width they like in the viewer

joa-quim: We all use <tabs> only for indentation

src/gmt_io.c
PaulWessel: I think this is a good check. The 20 is a bit arbitrary anyway and we cannot know what lies down the road...

src/gmt_io.c
meghanrjones: Thanks for the reminder @seisman

meghanrjones:
I think this is good to be merged now. @PaulWessel or @seisman, could you check that no changes need to be made to the new tests?

PaulWessel:
Sorry, checking now - had a zoom

PaulWessel:
Looks like the test passes and works fine.  Good to go I think,

PR #4781, opened 2021-02-09
Copy gmt.conf if movie or batch detects one in current directory

PaulWessel:

This PR seeks to let **batch** and **movie** be able to use any gmt.conf files found in the directory when these commands are run.  The reason we need to do something specific is that the GMT commands contained in the scripts are run in different directories than where batch or movie starts.  Thus, we must actually place the gmt.conf file in all the directories where the scripts are run form.  The algorithm implement is this:

1. If gmt.conf is found in the current directory then we keep the full path to this file in a variable
2. As we build the pre/post-flight and parallel scripts from input scripts we check if **gmt begin** is given the **-C** option, and if not then we copy in the gmt.conf it it was found in step 1.

The documentation has been updated and also points out that users cannot use the preflight script to create a gmt.conf since it will be created in a hidden directory that is deleted upon completion.

### DIFF
--- a/doc/rst/source/batch.rst
+++ b/doc/rst/source/batch.rst
@@ -177,6 +177,14 @@ as well as any new files produced by *mainscript* or the optional scripts set vi
 No path specification is needed to access these files.  Other files may
 require full paths unless their directories were already included in the :term:`DIR_DATA` setting.
 
+Custom gmt.conf files
+---------------------
+
+If you have a gmt.conf file in the top directory with your main script prior to running **batch** then it will be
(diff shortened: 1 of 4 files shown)


joa-quim (approved):

delaviel-anger:
If I may add ... It's the same problem with the cpt file :)

PaulWessel:
Yes, it may be - I will look at that.  The complication is that many will use **-W** to place the temporary stuff in on a temporary disk and then it is not in ...

PR #4793, opened 2021-02-11
Add the polynomial seamount shape to grdseamount arsenal

PaulWessel:

While a Gaussian model for seamounts is very popular, its height as a function of radius only asymptotically approaches zero.  An alternative is a polynomial shape which tracks the Gaussian shape very well yet goes to zero at the basal radius.  Its equation is

h(r) = (1+r)^3(1-r)^3/(1+r^3)

and has been used in seamount studies.  This PR adds this model as a new selection in grdseamount via **-Co**.  I have updated the documentation and added a test script that also compares it to a Gaussian:

smt_poly

Several of us (Sandwell, Kiim, me) are working on seamount problems these days and we need this feature added.


src/potential/grdseamount.c
```diff
@@ -456,6 +516,13 @@ GMT_LOCAL double grdseamount_gauss_solver (double in[], double f, double v, bool
 	return (phi);
 }
 
+GMT_LOCAL double grdseamount_poly_solver (double in[], double f, double v, bool elliptical) {
+	/* Return effective phi given volume fraction from a polynomial seamount is not yet implemented */
+	gmt_M_unused (in), gmt_M_unused (f), gmt_M_unused (elliptical);
+	/* NOT IMPLEMENTED YET */
+	return (v);
```
meghanrjones, line 523: Are you planning to implement this in this PR or separately?

PaulWessel:
Separately if I can figure it out, but will numerically be so close to Gaussian that I may leave it as is or fake it.

joa-quim (approved):

src/potential/grdseamount.c
```diff
@@ -410,6 +422,54 @@ GMT_LOCAL void grdseamount_gaussian_area_volume_height (double a, double b, doub
 	*z = (*V) / (*A);
 }
 
+GMT_LOCAL double poly_smt_func (double r) {
+	/* Polynomial radial function similar to a Gaussian */
+	return (pow ((1.0 + r) * (1.0 - r), 3.0) / (1.0 + pow (r, 3.0)));
```
meghanrjones, line 427: Can you confirm that this is the correct functional form? I want to double check because the function given in the PR proposal is different. (h(r) = (1+r)^3(1-r)^2/(1+r^3))

PaulWessel: Yikes,  OK I will fix but after an hour of zoom.  This explains why my updated plots in doc/script looks good (gmt math on the right equation) while the test plot seems to have too much difference with a Gaussian (using the flawed output).
OR, you could fix this as per the equation and run tests and problaby need to update the test/potential/? plot.

meghanrjones: I actually think that both the test and the docs might use the correct equation and it's just the PR description that had the wrong one. This is what the seamount would look like with h(r) = (1+r)^3(1-r)^2/(1+r^3):
smt_poly

PaulWessel:
gotta be ^3 everywhere, not 2.

PaulWessel:
Actually, I think it was right all along, no?

PaulWessel:
Fixing the proposal...

meghanrjones:
> Fixing the proposal...

Yes, all the code was correct. Sorry for the false alarm. Probably no one will look at the PR again, but better to be accurate anyways.

meghanrjones (approved):

Issue #4762, opened 2021-02-07
Standardize on a maximum usage message width


PaulWessel:
**Description of the desired feature**

When you type gmt module you get the usage message.  These vary in character length and some are really long, probably wrapping around on most people's laptop and making it harder to read.  A quick script made this histogram of these lengths:

![hist](https://user-images.githubusercontent.com/26473567/107159755-2c75cc00-6936-11eb-9a65-abed254cc231.png)

I think we should

1. Decide on a maximum width for these (100 characters, 120?)
2. Go through all usage messages and do the manual wrapping
3. This is a chore than almost anyone can help with so we can ask for volunteers.

KristofKoch:
Default terminal size is 80 x 24 characters. Why not stick to this size? Yes, more work (lots of red east of 80 …). But I'd volunteer. Where the 80 x 24 comes from? This size is seemingly dating back to IBM punched cards. 

Another note about ideal line length and characters per line:

> Generally for printed text it is widely accepted that line length fall between 45–75 characters per line (cpl), though the ideal is 66 cpl (including letters and spaces).

_Bringhurst, R. (1992). Horizontal Motion. The Elements of Typographic Style, pp 25-36. Point Roberts, WA: Hartley & Marks._

PaulWessel:
Those are good points, @KristofKoch.  My vote should count very little on this since I have a 42 inch 4K TV as may monitor and my terminals are not 80x24!

joa-quim:
110-120 chars. Remember that the listings do not start at 0 but at several <tabs> long. For example the longest line in the image is at position 105 and that on a Windows terminal with "thick" chars.

I never agreed with those very short widths defended with arguments on human perception, comfort of reading etc. This not about relaxed reading a novel but about (repeatedly) reading a couple of lines in order to understand a certain point. And now all command windows can be enlarged to not produce an excessive line wrapping.

![image](https://user-images.githubusercontent.com/537321/107162567-14d12e80-699c-11eb-9743-eab430b91371.png)

KristofKoch:
That's a really nice setup! When I put two terminals adjacent to each other I can go 92 characters wide on each.  I don't have a 42" monitor though … Maybe that 80 count isn't so bad after all? 

If you accept me for doing this job I'd appreciate some guidance on where to find the messages, @PaulWessel.

KristofKoch:
[...]

Issue #477, opened 2019-03-05
Coding styles and code formatting


seisman:
With more people contributing to the GMT codes, it's necessary to define coding styles for the GMT projects to helps maintain consistent coding styles for multiple contributors across various editors and IDEs.

For example, I prefer spaces for indentation while GMT uses tabs. So I need to convert spaces to tabs before committing my codes to keep the consistence. Sometimes, I forgot to do the conversion before committing my codes, making GMT codes mixing tabs and spaces (sorry for that).

We may use some tools (e.g. `clang-format`) to define our coding styles and automatically format the GMT codes.

leouieda:
I usually like the auto-formatting route because it takes the burden off of our shoulders. It really sucks when you commit a file only to realize that your editor auto-converted all the tabs to spaces and your diff is the whole file.

PaulWessel:
I prefer tabs to indicate indent.  The editor can be set to translate that into any number of spaces.  You cannot go the other way (have the editor know that your # of spaces is a tab for indent).  I think we are consistently using tabs across GMT and I fix stuff when I notice spaces.

PaulWessel:
A related issue is the length of a line. Apparently, some people (I am looking at you, @joa-quim) have monitors that are less than 40 inches and thus will break a tiny 200-character line into two lines, sometimes with the backslash.  Why am I buying 40" SamSung 4K TVs as monitors if people are cramping my style and forcing 80-character punchcard widths on me?

seisman:
I agree that it's unnecessary to force the line length to <=80 characters for GMT. 

Clang-format won't break the input long lines if the option `ColumnLimit` is 0.

See https://clang.llvm.org/docs/ClangFormatStyleOptions.html for all clang-format's options.

joa-quim:
@seisman don't believe in all you read here. No one is defending a 80 chars limit, not even 92 (which probably covers all that other code standards defend). But there should be a limit well behind 200 chars. Specially if they are used to write code with
```
for() ...; for() ...; if () ...
```
instructions in a single line.
I think something between 120 and 130 is ok. Unfortunately we still have many lines longer than this. (the fig trims at 200 but there several that go beyond 300).

![histo_LL](https://user-images.githubusercontent.com/537321/54879695-e0017000-4e33-11e9-91c6-a167ffb31603.png)

joa-quim:
Actually our code lines are way longer because the script that counted their lengths counted tabs as one char only.

[...]

PR #4796, opened 2021-02-12
Improve usage documentation and break synopsis lines to fit terminal

PaulWessel:
See #4762 for background.  This PR adds a new testbreak.c program that can be used for exploring auto-breaking.  Since it is not a GMT module, it gets built and placed in build/src so you need to run it from there.  Its syntax is

testbreak [_length_ [-]]

Here, _length_ is the nominal width of the synopsis output [80].  If you give a _length_ then you may also give an optional - to select @KristofKoch's fancy UTF-8 carriage-return symbol ⏎ for line continuation; otherwise you get @joa-quim's sad .....
Behavior: We try to make output records around the _length_.  If adding the next word greatly exceeds _length_ then we try to break it at a ] and append a "line to be continued" marker, and indent 2 spaces on the continuation line.

Examples:

testbreak

```
WRAP: 80 characters

usage: somemodule <grid> [-A] [-C] [-D[+x<xname>][+y<yname>][+d<dname>][+s<scale>]...
	  ...[+o<offset>][+n<invalid>][+t<title>][+r<remark>][+v<name>]] [-E[a|e|h|l|r|t|v]] 
	[-G<outgrid>] [-J<args>] [-L[+n|p]] [-N<table>] 
	[-R<west>/<east>/<south>/<north>[+r]] [-S] [-T] [-V[<level>]] 
	[-bi[<ncol>][t][w][+l|b]] [-di<nodata>] [-e[~]<pattern>] [-f[i|o]<info>] [-h[i|o]...
	  ...[<nrecs>][+c][+d][+m<segheader>][+r<remark>][+t<title>]] [-i<cols>[+l][+d<divide>]...
	  ...[+s<scale>][+o<offset>][,...][,t[<word>]]] 
	[-wa|y|w|d|p<period][/<phase>][+c<col>]] [-:[i|o]] [--PAR=<value>]
```

testbreak 100 -
```
WRAP: 100 characters

usage: somemodule <grid> [-A] [-C] [-D[+x<xname>][+y<yname>][+d<dname>][+s<scale>][+o<offset>][+n<invalid>]⏎
	  [+t<title>][+r<remark>][+v<name>]] [-E[a|e|h|l|r|t|v]] [-G<outgrid>] [-J<args>] [-L[+n|p]] 
	[-N<table>] [-R<west>/<east>/<south>/<north>[+r]] [-S] [-T] [-V[<level>]] [-bi[<ncol>][t][w][+l|b]] 
	[-di<nodata>] [-e[~]<pattern>] [-f[i|o]<info>] [-h[i|o][<nrecs>][+c][+d][+m<segheader>][+r<remark>]⏎
	  [+t<title>]] [-i<cols>[+l][+d<divide>][+s<scale>][+o<offset>][,...][,t[<word>]]] [-wa|y|w|d|p<period]⏎
	  [/<phase>][+c<col>]] [-:[i|o]] [--PAR=<value>]
```

The idea is that we would just make one lone text string at top of each usage() function then call this gmt_usage_line function to display according to our width rules.  It could even be a GMT parameter [**GMT_SYNOPSIS_WIDTH** = 120].


PaulWessel:
Does anyone find this test useful and have suggestions for improvements? @KristofKoch ? @joa-quim ?

joa-quim:
Did not play with this yet.

KristofKoch:
Surely a beginners question – but how do I start the program?

joa-quim:
After building from this branch just type ``testbreak`` in the command line.

KristofKoch:
Mh, then I'm doing something wrong:

```
$ gmt --version
6.2.0_0545e0a_2021.02.11
$ testbreak
-bash: testbreak: command not found
$ gmt testbreak

ERROR: No module named testbreak was found. […]
```

PaulWessel:
Remember the binary is in build/src which is probably not in your path.  It is not a gmt module.

PaulWessel:
OK, looks like nobody gave this a spin.  Given we have a lot of hideously formatted usage messages like

```
usage: gmt velo [<table>] -J<args> -R<west>/<east>/<south>/<north>[+r] [-A<vecpar>] [-B<args>] [-D<sigscale>]
	[-G<fill>] [-L] [-N] [-S<symbol><args>[+f<font>]]
	[-U[<label>][+c][+j<just>][+o<dx>/<dy>]] [-V] [-W<pen>] [-X[a|c|f|r]<xshift>]
	[-Y[a|c|f|r]<yshift>] [-c[<row>,<col>|<index>]] [-di<nodata>] [-e[~]<pattern>] [-h[i|o][<nrecs>][+c][+d][+m<segheader>][+r<remark>][+t<title>]]
	[-i<cols>[+l][+d<divide>][+s<scale>][+o<offset>][,...][,t[<word>]]] [-p[x|y|z]<azim>[/<elev>[/<zlevel>]][+w<lon0>/<lat0>[/<z0>][+v<x0>/<y0>]]
	[-qi[~]<rows>[,...][+c<col>][+a|f|s]] [-t<transp>[/<transp2>[+f|s]] [-:[i|o]] [--PAR=<value>]
```

and 150 of these, it seems something automatic and configurable is better than manual grunt work that has to be updated each time another option is added?  Please give it a try and see if you think this is an improvement.  Adding @GenericMappingTools/core now since I forgot.

meghanrjones:
If this gets converted to [a GMT_Wrap function](https://github.com/GenericMappingTools/gmt/issues/4762#issuecomment-775437261), could it be implemented for users to define their preferred default message line length in ConfigUserAdvanced.cmake?

PaulWessel:
Sure, if we decide that adding a new defaults like **GMT_SYNOPSIS_WIDTH** is too much tinkering then we can make that a CMake setting instead with our final default value [say 120].

meghanrjones:
> Sure, if we decide that adding a new defaults like **GMT_SYNOPSIS_WIDTH** is too much tinkering then we can make that a CMake setting instead with our final default value [say 120].

Either way seems fine. Will gmt_usage_line work for the detailed option descriptions in addition to the synopsis or will that require a separate function? Also, did you base the example synopsis in testbreak.c on any particular GMT module? I am interested in comparing the testbreak output to the current GMT output.

PaulWessel:
Looks like it was based on grdedit.  This was only for the synopsis.  The individual descriptions is tricker because some are done by a functino and others are inline.  But perhaps that can still be handled and the break-function could be told to honor \n to break when we want to break, etc.

meghanrjones:
> Looks like it was based on grdedit. This was only for the synopsis. The individual descriptions is tricker because some are done by a functino and others are inline. But perhaps that can still be handled and the break-function could be told to honor \n to break when we want to break, etc.

Thanks for the explanation. This is nice, I think that giving the user the ability to define their preferred message length is a worthwhile improvement. When the ⏎ symbol is not used, I find it slightly more readable to use only three periods on the continued line, rather than two spaces and three periods.

It will still be necessary to define the preferred line length for the reformatted synopsis and long descriptions in the source code if GMT_Message() gets refactored to GMT_Wrap() in some places; it would be nice to close [this issue](https://github.com/GenericMappingTools/gmt/issues/477) at the same time by adding the decided convention to [CODE_CONVENTIONS.md](https://github.com/GenericMappingTools/gmt/blob/master/CODE_CONVENTIONS.md). I agree with @joa-quim's suggestion of 120-130 chars.

PaulWessel:
Yes, I agree. 120-130 is probably good, but once we implement a parameter then we can play with it easier and get a sense for an optional one - perhaps we can even think of a way to determine the "best" length given the lines we have.

As is my habit for silly things like this, I just coded away until it worked.  Not much design planning, and it is possible there are corner cases I did not consider that won't be easy to shoehorn in etc.  But the basics are there I think.  If we could simply transfer all those GMT_Message lines to a single string, e.g., 

```
	GMT_Message (API, GMT_TIME_NONE, "\t<grid> is file to be modified.\n");
	GMT_Message (API, GMT_TIME_NONE, "\n\tOPTIONS:\n");
	GMT_Message (API, GMT_TIME_NONE, "\t-A Adjust dx/dy to be compatible with the file domain or -R.\n");
	GMT_Message (API, GMT_TIME_NONE, "\t-C Remove the command history from the header.\n");
```

to 


```
"\n\tOPTIONS:\n"
	"\t-A Adjust dx/dy to be compatible with the file domain or -R.\n"
	"\t-C Remove the command history from the header.\n"
```

and then call `GMT_Wrap_Message (API, GMT_TIME_NONE, msg);`

KristofKoch:
Not exactly helping but I still can't get `testbreak` to run. My apologies.

PaulWessel:
Looks like I forgot to give the magic information that you must enable this in your cmake/ConfigUserAdvanced.cmake:

`set (DO_API_TESTS ON)
`

otherwise it does not compile all the test*.c codes that only developers care about.  Sorry about that.

meghanrjones:
> Not exactly helping but I still can't get `testbreak` to run. My apologies.

Also, after building this branch you can add it to your `PATH` like this (assuming your build directory is named `build`):

``
export PATH='<path-to-gmt-repo>/build/src/':$PATH
``

KristofKoch:
Ahhhh, with the magic information I got it to work. Thanks, @PaulWessel. And thank you @meghanrjones for the reminder.

Unfortunately it doesn't break cleanly after 80 chars for me:

testbreak-80

PaulWessel:
True, it was written with the count as "guidance" and break on the next nearest opportunity.  Breaking at exactly 80 means in the middle of arguments which I think is an affront to all human decency...

KristofKoch:
The terminal window in the screenshot above is 80 chars wide. My understanding was that I give the terminal width and therefore the count acts as limit and not as guidance. Maybe break < 77 for the `…` variant and < 79 for the `⏎` variant (to allow for the extra chars)? Basically the last opportunity to break prior to reaching the count?

PaulWessel:
I can probably change the code to keep track of the last break point and when we exceed the count go there.

KristofKoch:
@PaulWessel do you see a chance for the deluxe version? Get the current size of the terminal window and break accordingly? I'm not a C-programmer but maybe https://stackoverflow.com/questions/1022957/getting-terminal-width-in-c#1022961 is a starting point? 

If feasible this would end the discussion between the 80x24 supporters and the "I have a monster display and I want to use it" crowd as well without resorting to a fixed break count at build time or as a variable.

PaulWessel:
We could try.  First will need to see if @joa-quim can try the simple code in the link you provided and see if it even works for Window.  It worked fine in macOS.

PaulWessel:
While the ISO code for the fancy return did not work for @joa-quim, I would be astonished if there is not some other glyph that Windows has that we could pick if on Windows.

PaulWessel:
Hi @KristofKoch I updated using your link info so if you give - as the width it will use the terminal width.  Still, it will overshot but I will try to undershoot in the next revision.

PaulWessel:
Now should not overshoot.  Try testbreak - - to (a) pick up terminal width and (b) use fancy ISO return.

KristofKoch:
Very cool, @PaulWessel! Some minor issues:

testbreak

1. Why the break symbol? It is a clean break.
2. Ugly break – not catering for the inserted break symbol
3. More of a readability issue - might a space before the carriage return symbol increase readability?
4. Why the continuation points for clean breaks and for carriage return symbol breaks?

PaulWessel:
> * Why the break symbol? It is a clean break.

Bug that needs fixing.
> * Ugly break – not catering for the inserted break symbol

I probably only counted for one (ISO break) but need to count 2 for these
> * More of a readability issue - might a space before the carriage return symbol increase readability?

It might,  should be easy to try
> * Why the continuation points for clean breaks and for carriage return symbol breaks?

Bug that needs fixing.

This code was written as code should not be written: Rapidly hacking away with no clear design plan.  So while these worts need fixing, just focus on the overall behavior for now.  I.e., is something like this a good solution that we really want to implement in GMT? Also, we need to awake @joa-quim to figure out how Windows handles the ioctl call (badly, presumably) and what the equivalent action would be for Windows.

joa-quim:
Because I knew it
```
fatal error C1083: Cannot open include file: 'sys/ioctl.h': No such file or directory
```

Not very inspirating

https://unix.stackexchange.com/questions/254182/ioctl-library-for-windows

PaulWessel:
Sure, but I expect there to be an #ifdef WIN32 that we can use to instead call a Windows library function to get the width of the terminal.  that is all we need, plus ideally a glyph code to print the return signal.  I am sure all this exist.

KristofKoch:
Neither a C nor a Windows guy but another hopefully useful tidbit of information:

https://stackoverflow.com/questions/6812224/getting-terminal-size-in-c-for-windows

PaulWessel:
OK @joa-quim I implemented what @KristofKoch found for Windows.  See if it works for Windows now.

joa-quim:
Funny, also found this and made a working version. But yours has a few less lines

https://docs.microsoft.com/en-us/windows/console/scrolling-a-screen-buffer-s-window

PaulWessel:
And does it work?

joa-quim:
Yes.
Bur replace
```
n_columns = csbi.srWindow.Right - csbi.srWindow.Left + 1;
```
by
```
n_columns = csbi.srWindow.Right;
```
``csbi.srWindow.Left == 0``

PaulWessel:
OK, and the +1 is not needed?

joa-quim:
Don't see why. ``csbi.srWindow.Right`` gives the full width.

PaulWessel:
OK, committed.  Now need a glyph.  Not sure waht is good.  Maybe limited: https://graphicdesign.stackexchange.com/questions/22128/what-font-and-characters-are-used-for-the-line-break-symbol-used-in-microsoft/49731

joa-quim (approved):

PaulWessel:
While this one still needs some work, I think I will just rewrite that function based on the lessons learned.  But I would really like to just a glyph for the break-the-line.  Might @joa-quim be able to figure out how to print such a char from C and have it look fine in the Windows terminal?
Note I removed the leading tab that we use with 2 spaces since (a) tab in terminals is typically 8 characters which is 10% of @KristofKoch's window width (!) and (b) while it is configurable in some terminals, I suspect nobody does that (in editors, yes, but not terminal windows).

KristofKoch:
> [H]ow to print [a carriage-return symbol] from C and have it look fine in the Windows terminal?

I had a look around and it doesn't look too promising for Windows: 

* UTF-8 is [not supported out of the box](https://akr.am/blog/posts/using-utf-8-in-the-windows-terminal)
* The [Windows code page 1252](https://en.wikipedia.org/wiki/Windows-1252#Character_set) has no carriage-return symbol

Those are the options you can choose from until Windows supports UTF-8:

Windows-1252-infobox

Maybe @joa-quim has some tricks up his sleeve?

If you want to avoid spending three chars for dots `...` it seems the horizontal ellipsis `…` (Alt+0133 in Windows world) is supported saving two chars in width. @joa-quim can you confirm this?

Horizontal ellipsis `…` mock-up:
```
usage: somemodule <grid> [-A] [-C] [-D[+x<xname>][+y<yname>][+d<dname>]…
  …[+s<scale>][+o<offset>][+n<invalid>][+t<title>][+r<remark>][+v<name>]] 
  [-E[a|e|h|l|r|t|v]] [-G<outgrid>] [-J<args>] [-L[+n|p]] [-N<table>] 
  [-R<west>/<east>/<south>/<north>[+r]] [-S] [-T] [-V[<level>]] [-bi[<ncol>]…
  …[t][w][+l|b]] [-di<nodata>] [-e[~]<pattern>] [-f[i|o]<info>] [-h[i|o]…
  …[<nrecs>][+c][+d][+m<segheader>][+r<remark>][+t<title>]] [-i<cols>[+l]…
  …[+d<divide>][+s<scale>][+o<offset>][,...][,t[<word>]]] 
  [-wa|y|w|d|p<period][/<phase>][+c<col>]] [-:[i|o]] [--PAR=<value>]
```

> 8 characters which is 10% of @KristofKoch's window width

Not a problem anymore. I switched to a more convenient 1000x5 terminal size. See screenshot below. Gone are the hassles of line breaks etc. ;)

1000x5

joa-quim:
I never really understood what is utf8. In principle it should be the last bit (last 128 chars) of the extended ASCII set. But then we get the different code pages and ... I give up understanding,

My Matlab prints this (the known "Sargent" symbol) that looks good for the purpose

>> char(171)

ans =

    '«'

PaulWessel:
If we can plot the 1-char ellipses symbol for Windows then we can use that as a line breaker for all:

```
  -T<long>/<lat>[+d<distance>][+e<eigenvalue>]…
  …[+p<pen>][+g<fill>][+w<weights>]
```

I think ending one line with … and starting the next with … is a reasonable scheme - it does not have to be different symbols. We would also couple this with not printing tabs but use a smaller number of spaces for indents.

Hi @joa-quim, are you able to craete that symbol via printf in C?  Please edit this and replace the values in brk and see if it can work for you:

```
#include <stdio.h>
int main () {
	char *message = "-T<long>/<lat>[+d<distance>][+e<eigenvalue>]";
	char *brk = "\xe2\x80\xa6";

	fprintf (stderr, "  -T<long>/<lat>[+d<distance>][+e<eigenvalue>]%s\n  %s[+p<pen>][+g<fill>][+w<weights>]\n", brk, brk);
}

```

PaulWessel:
I think the simplest algorithm to handle a single input string with no line-return characters is to split into words, where words are separated either by spaces (between options) or when we find "][" which is the juxtaposition of two optional items.  These are the places were we can break.  So we can split the line into these words, then keep outputting words as long as the next word fits the stated width, and if not issue the ellipses+newline and then another ellipses and start outputting to next line.  Should be simple.  As example of a single long synopsis long from grdtrack:

`usage: gmt grdtrack -G<grid1> -G<grid2> ... [<table>] [-A[f|m|p|r|R][+l]] [-C<length>/<ds>[/<spacing>][+a|v][+l|r] [-D<dfile>] [-E<line1>[,<line2>,...][+a<az>][+c][+d][+g][+i<step>][+l<length>][+n<np][+o<az>][+r<radius>]] [-F[+b][+n][+r][+z<z0>]] [-N] [-R<west>/<east>/<south>/<north>[+r]] [-S[<method>][<modifiers>]] [-T<radius>>[+e|p]] [-V[<level>]] [-Z] [-b[i|o][<ncol>][t][w][+l|b]] [-e[~]<pattern>] [-f[i|o]<info>] [-g[a]x|y|d|X|Y|D|[<col>]z<gap>[+n|p]] [-h[i|o][<nrecs>][+c][+d][+m<segheader>][+r<remark>][+t<title>]] [-i<cols>[+l][+d<divide>][+s<scale>][+o<offset>][,...][,t[<word>]]] [-je|f|g] [-n[b|c|l|n][+a][+b<BC>][+c][+t<threshold>]] [-o<cols>[,...][t[<word>]]] [-q[i|o][~]<rows>[,...][+c<col>][+a|f|s]] -s[<cols>][+a][+r]] [-wa|y|w|d|p<period][/<phase>][+c<col>]] [-:[i|o]] [--PAR=<value>]`

The function would need a flag to distinguish between synopsis line and text as part of option descriptions.  These will also exceed widths and need to also honor the terminal width.

PaulWessel:
New algorithm implemented that seems to work very well.  Also, the program now expects a single quoted sentence to break.  Here is the output when feeding it the above **grdtrack** synopsis for a terminal width of 90:

```
usage: gmt grdtrack -G<grid1> -G<grid2> ... [<table>] [-A[f|m|p|r|R][+l]]
  [-C<length>/<ds>[/<spacing>][+a|v][+l|r] [-D<dfile>] [-E<line1>[,<line2>,...][+a<az>]…
   …[+c][+d][+g][+i<step>][+l<length>][+n<np][+o<az>][+r<radius>]] [-F[+b][+n][+r]…
   …[+z<z0>]] [-N] [-R<west>/<east>/<south>/<north>[+r]] [-S[<method>][<modifiers>]]
  [-T<radius>>[+e|p]] [-V[<level>]] [-Z] [-b[i|o][<ncol>][t][w][+l|b]] [-e[~]<pattern>]
  [-f[i|o]<info>] [-g[a]x|y|d|X|Y|D|[<col>]z<gap>[+n|p]] [-h[i|o][<nrecs>][+c][+d]…
   …[+m<segheader>][+r<remark>][+t<title>]] [-i<cols>[+l][+d<divide>][+s<scale>]…
   …[+o<offset>][,...][,t[<word>]]] [-je|f|g] [-n[b|c|l|n][+a][+b<BC>][+c]…
   …[+t<threshold>]] [-o<cols>[,...][t[<word>]]] [-q[i|o][~]<rows>[,...][+c<col>]…
   …[+a|f|s]] -s[<cols>][+a][+r]] [-wa|y|w|d|p<period][/<phase>][+c<col>]] [-:[i|o]]
  [--PAR=<value>]
```

A few notes:

1. Uses the ellipsis character to indicate break at end of line, then indents an extra space and prints the ellipsis again before continuing with the rest of the split line.
2. Only works for Linux/macOS until @joa-quim edits the ifdef WIN part and puts in the right code for ellipses
3. When you run it, it first breaks the words up and prints all the words for debugging purposes - not shown here.  @KristofKoch, this one should stay in its lane and not exceed the terminal setting.

joa-quim:
Don't understand what you want me to do. I cannot fabricate a symbol and even don't understand how to make this work. Had to cange the line
```
char *brk = "\xe2\x80\xa6";`
```
to something else otherwise it apparently looped for some seconds before printing

```
src\testbreak.exe -
WRAP: 153 characters

Found 1 words
00: [0] -

-
```

PaulWessel:
I thought between your Matlab text and Krifstoff's win char list you had a ellipses symbol you could pick

PaulWessel:
He says "Alt+0133 in Windows world".  We need that into a C-string, probably via hex like for the others.

joa-quim:
But how do I make ``testbreak`` print anything?

PaulWessel:
try to give it a string in quotes, e.g.

`testbreak "There was a man who lived by the sea.  He never ate any food, just breathed air until his stomach was full.  Happily, it enjoyed the sea breezed and the sun until he suddenly died of starvation."`

or

`testbreak "usage: gmt grdtrack -G<grid1> -G<grid2> ... [<table>] [-A[f|m|p|r|R][+l]] [-C<length>/<ds>[/<spacing>][+a|v][+l|r] [-D<dfile>] [-E<line1>[,<line2>,...][+a<az>][+c][+d][+g][+i<step>][+l<length>][+n<np][+o<az>][+r<radius>]] [-F[+b][+n][+r][+z<z0>]] [-N] [-R<west>/<east>/<south>/<north>[+r]] [-S[<method>][<modifiers>]] [-T<radius>>[+e|p]] [-V[<level>]] [-Z] [-b[i|o][<ncol>][t][w][+l|b]] [-e[~]<pattern>] [-f[i|o]<info>] [-g[a]x|y|d|X|Y|D|[<col>]z<gap>[+n|p]] [-h[i|o][<nrecs>][+c][+d][+m<segheader>][+r<remark>][+t<title>]] [-i<cols>[+l][+d<divide>][+s<scale>][+o<offset>][,...][,t[<word>]]] [-je|f|g] [-n[b|c|l|n][+a][+b<BC>][+c][+t<threshold>]] [-o<cols>[,...][t[<word>]]] [-q[i|o][~]<rows>[,...][+c<col>][+a|f|s]] -s[<cols>][+a][+r]] [-wa|y|w|d|p<period][/<phase>][+c<col>]] [-:[i|o]] [--PAR=<value>]"`

Then resize your terminal and run again.

joa-quim:
Adding the ``SetConsoleOutputCP`` line seems to work.

```
#ifdef WIN32
	char *brk = "\xe2\x80\xa6";
	SetConsoleOutputCP(CP_UTF8);
#else
```

PaulWessel:
OK great, and windows.h is all I need for that?

PaulWessel:
I guess this also opens up for the carriage-return symbol then, being UTF8.

joa-quim:
That's what it seems. There is no other exotic include.

PaulWessel:
How a bout this, @KristofKoch ?

```
usage: gmt grdtrack -G<grid1> -G<grid2> ... [<table>] [-A[f|m|p|r|R][+l]]
  [-C<length>/<ds>[/<spacing>][+a|v][+l|r] [-D<dfile>] [-E<line1>[,<line2>,...][+a<az>]⏎
   …[+c][+d][+g][+i<step>][+l<length>][+n<np][+o<az>][+r<radius>]] [-F[+b][+n][+r]⏎
   …[+z<z0>]] [-N] [-R<west>/<east>/<south>/<north>[+r]] [-S[<method>][<modifiers>]]
  [-T<radius>>[+e|p]] [-V[<level>]] [-Z] [-b[i|o][<ncol>][t][w][+l|b]] [-e[~]<pattern>]
  [-f[i|o]<info>] [-g[a]x|y|d|X|Y|D|[<col>]z<gap>[+n|p]] [-h[i|o][<nrecs>][+c][+d]⏎
   …[+m<segheader>][+r<remark>][+t<title>]] [-i<cols>[+l][+d<divide>][+s<scale>]⏎
   …[+o<offset>][,...][,t[<word>]]] [-je|f|g] [-n[b|c|l|n][+a][+b<BC>][+c][+t<threshold>]]
  [-o<cols>[,...][t[<word>]]] [-q[i|o][~]<rows>[,...][+c<col>][+a|f|s]] -s[<cols>][+a]⏎
   …[+r]] [-wa|y|w|d|p<period][/<phase>][+c<col>]] [-:[i|o]] [--PAR=<value>]
```

KristofKoch:
@PaulWessel I like that carriage return + ellipsis style. Sorry, work interfered with GMT the last couple of days.

PaulWessel:
While we are converging on what to do with synopsis lines, what about the rest of the option usage lines?  As an example, here is what **grdtrack** shows for **-S**

```
	-S In conjunction with -C, compute a single stacked profile from all profiles across each segment.
	   Append which method should be used when performing the stacking:
	   a = mean, m = median, p = mode, l = lower, L = lower of +ve values, u = upper, U = upper of -ve values [a].
	   The modifiers control what is being written; choose one or more among
	     +a : Append stacked values to all cross-profiles.
	     +d : Append stack deviations to all cross-profiles.
	     +r : Append data residuals (data - stack) to all cross-profiles.
	     +s[<file>] : Save stacked profile to <file> [stacked_profile.txt].
	     +c<fact> : Compute envelope as +/- <fact>*deviation [2].
	   Note: Deviations depend on method and are L1 scale (m), st.dev (a), LMS scale (p), or half-range (u-l)/2.
```

The longest line here is 108 chars plus the leading indent (TAB + 3 spaces), so appears as 119 characters wide.  Do we want to treat these the same way as synopsis, meaning if the user's terminal is 80 chars wide it should instead look like this (a mockup):

```
	-S In conjunction with -C, compute a single stacked profile from all profiles
 	   across each segment. Append which method should be used when performing
 	   the stacking: a = mean, m = median, p = mode, l = lower, L = lower of +ve
 	   values, u = upper, U = upper of -ve values [a]. The modifiers
 	   control what is being written; choose one or more among:
	     +a : Append stacked values to all cross-profiles.
	     +d : Append stack deviations to all cross-profiles.
	     +r : Append data residuals (data - stack) to all cross-profiles.
	     +s[<file>] : Save stacked profile to <file> [stacked_profile.txt].
	     +c<fact> : Compute envelope as +/- <fact>*deviation [2].
	   Note: Deviations depend on method and are L1 scale (m), st.dev (a), LMS
	   scale (p), or half-range (u-l)/2.
```

Note that these texts may have embedded new-lines which will need to be honored if ending a line.

KristofKoch:
Hi @PaulWessel, I like your `grdtrack -S` mockup. In my opinion the forced `⏎  …` line break is just a special variation of a hyphenation char. So yes, I agree with you. The more formatted option lines should break at terminal width as well but still retain their indentation and line breaks. I don't know how much of a problem clean breaking of nested enumerations might be:

_80 characters wide:_
```
-O some option : Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed 
   diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, 
   sed diam voluptua:
     +a : At vero eos et accusam et justo duo dolores et ea rebum. Stet clita 
        kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.
     +longeone : Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed 
        diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam 
        erat, sed diam voluptua. 
     +c : At vero eos et accusam et justo duo dolores et ea rebum. Stet clita 
        kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.
-P some other option: Lorem ipsum dolor sit amet, consetetur sadipscing elitr, 
   sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam 
   erat, sed diam voluptua:
```

_100 characters wide:_
```
-O some option : Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod 
   tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua:
     +a : At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no    
        sea takimata sanctus est Lorem ipsum dolor sit amet.
     +longeone : Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed 
        diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam 
        voluptua. 
     +c : At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no
        sea takimata sanctus est Lorem ipsum dolor sit amet.
-P some other option: Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy 
   eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua:
```

PaulWessel:
I think as long as the module calls the new _GMT_Line_Message_ separately for the synopsis (one long string to be wrapped) versus keeping each option as a separate call and sub-modifiers as separate lines, each call given initial and different indent values, then it should be OK.

PaulWessel:
I have now implemented these ideas into a new API function called GMT_Usage.  Currently, it is only used in **gmtsimplify** to print the synopsis line:

```
(base) pwessel@macnut:~/GMTdev/gmt-dev/src-> gmt simplify -
gmt gmtsimplify [core] 6.2.0 [64-bit] - Line reduction using the Douglas-Peucker algorithm

usage: gmt gmtsimplify [<table>] -T<tolerance> [-V[<level>]] [-b[i|o][<ncol>][t][w][+l|b]]
  [-d[i|o]<nodata>] [-e[~]<pattern>] [-f[i|o]<info>] [-g[a]x|y|d|X|Y|D|[<col>]z<gap>[+n|p]]
  [-h[i|o][<nrecs>][+c][+d][+m<segheader>][+r<remark>][+t<title>]] [-i<cols>[+l][+d<divide>]⏎
   …[+s<scale>][+o<offset>][,...][,t[<word>]]] [-o<cols>[,...][t[<word>]]] [-q[i|o]⏎
   …[~]<rows>[,...][+c<col>][+a|f|s]] [-:[i|o]] [--PAR=<value>]
```

If @joa-quim could build and run that command on Windows and perhaps @seisman or @KristofKoch can try Linux to see that all is well, then I can start testing it on the GMT_Option output as well.

KristofKoch:
Hi @PaulWessel, the synopsis line breaks well for me (size 80x24):

Bildschirmfoto 2021-03-17 um 20 55 38

Great work!

PaulWessel:
Thanks @KristofKoch.  I will finish the rest of this for gmtsimplify so we have a template for what to do across 150 other modules and then we can post some help wanted signs.

KristofKoch:
You can count me in @PaulWessel

KristofKoch:
Quick question after your latest commits @PaulWessel: Do you consider the line-break behaviour now final? Or are you still tweaking?

PaulWessel:
Still testing.  I needed to also break on very long asdf/asdf/asdf/asdf/asdf/asdf/asdf/adsf at a slash [e.g., for general perspective command], and the indent values etc can still change.  I am almost done with recasting all the standard option usage messages (-R, -J, -a, -i, etc) and will finish gmtspatial soon.

PaulWessel:
Hi @KristofKoch and others.  I have now wrapped all the common GMT options plus the entirety of two modules: **simplify** and **plot**.  Some of the settings for this is now defined in _gmt_constants.h_; see **GMT_INDENT*** and **GMT_LINE_***.
Please build and tests these two modules. We still have inconsistencies in how an option is presented when there are modifiers.  Contrast **-h** with plot's **-F -L** for instance.  I have also fixed numerous spellings, punctuations and unclosed parentheses in the usages.

Comments welcome.

PaulWessel:
I need some feedback on how we want to format typical option descriptions for options with modifiers.  We do not have a fixed system for doing this.  Some places we do this

**-G**_val1_[**+m**_val2_][**+q**|**n**][**+z**_val3_]
   Turn on nasty effects based on _val1_.  Choose between these optional modifiers: **+m** makes things worse, append how much via _val2_, **+q** or **+n** adds Gaussian or linear noise respectively, while **+z** sets the overall pain level to _val3_.

while in other places we do this:

**-G**_val1_[**+m**_val2_][**+q**|**n**][**+z**_val3_]
   Turn on nasty effects based on _val1_.  Choose between these optional modifiers:
**+m** makes things worse, append how much via _val2_.
**+q** or **+n** adds Gaussian or linear noise respectively.
**+z** sets the overall pain level to _val3_.

While not shown here, the entire text after the -G line will be indented by 3 spaces, and then in the second example each modifier descriptive line (which may wrap) is indented 2-3 more spaces.

I think we should pick one format and revised this throughout as we roll out the auto-wrapper.  To me, the second example adds better separation and makes it easier to see the different modifiers.  While it will never be as clear as the RST documentation I think there is value in being consistent.

Thoughts on this, @GenericMappingTools/core ?

meghanrjones:
> I need some feedback on how we want to format typical option descriptions for options with modifiers. We do not have a fixed system for doing this. Some places we do this
> 
> **-G**_val1_[**+m**_val2_][**+q**|**n**][**+z**_val3_]
> Turn on nasty effects based on _val1_. Choose between these optional modifiers: **+m** makes things worse, append how much via _val2_, **+q** or **+n** adds Gaussian or linear noise respectively, while **+z** sets the overall pain level to _val3_.
> 
> while in other places we do this:
> 
> **-G**_val1_[**+m**_val2_][**+q**|**n**][**+z**_val3_]
> Turn on nasty effects based on _val1_. Choose between these optional modifiers:
> **+m** makes things worse, append how much via _val2_.
> **+q** or **+n** adds Gaussian or linear noise respectively.
> **+z** sets the overall pain level to _val3_.
> 
> While not shown here, the entire text after the -G line will be indented by 3 spaces, and then in the second example each modifier descriptive line (which may wrap) is indented 2-3 more spaces.
> 
> I think we should pick one format and revised this throughout as we roll out the auto-wrapper. To me, the second example adds better separation and makes it easier to see the different modifiers. While it will never be as clear as the RST documentation I think there is value in being consistent.
> 
> Thoughts on this, @GenericMappingTools/core ?

I prefer the second option as well because it is easier to distinguish between the modifiers. I also like that is more similar to the ReST documentation, in particular after the implementation of @Esteban82's suggestions for formatting.

PaulWessel:
```
usage: gmt gmtsimplify [<table>] -T<tolerance> [-V[<level>]] [-b[i|o][<ncol>]⏎
   …[t][w][+l|b]] [-d[i|o]<nodata>] [-e[~]<pattern>] [-f[i|o]<info>]
  [-g[a]x|y|d|X|Y|D|[<col>]z<gap>[+n|p]] [-h[i|o][<nrecs>][+c][+d]⏎
   …[+m<segheader>][+r<remark>][+t<title>]] [-i<cols>[+l][+d<divide>]⏎
   …[+s<scale>][+o<offset>][,...][,t[<word>]]] [-o<cols>[,...][t[<word>]]]
  [-q[i|o][~]<rows>[,...][+c<col>][+a|f|s]] [-:[i|o]] [--PAR=<value>]

  -h[i|o][<n>][+c][+d][+m<segheader>][+r<remark>][+t<title>]
     Input/output file has [0] header record(s) [OFF]. Optionally, append i for
     input or o for output only and number of header records [0]. Note: -hi
     turns off the writing of all headers on output since none will be read.
     Optionally, append modifiers:
     +c Add header record with column information [none].
     +d Delete headers before adding new ones [Default will append headers].
     +m Insert a new segment header and <segheader> content after the headers
        [none].
     +r Add a <remark> comment to the output [none].
     +t Add a <title> comment to the output [none]. Note: <remark> and <title>
        may contain \n to indicate line-breaks.
     For binary files, <n> is considered to mean number of bytes instead of
     records.
```

So this seems like a good scheme for me.  What you see for the **-h** option is the result of a few _GMT_Usage_ calls with different indent levels, as follows:

1. Synopsis gets indent level 0 hence it starts in the column 1, but as the line wraps all subsequent lines go to the next indent level.
2. The actual indents for each level is given by an array an is {0, 2, 5, 7, 10, 12, 15}
3. These indents was chosen so that the indents for options is 3 since "-T " is 3 characters
4. Similarly, the indent after a modifier "+g " is also 3 for the same alignment reasons.
5. To continue a usage message but immediately indent in as if it already wrapped I pass the negative level.
6. Breaking in the middle of something trigger the return and continue characters and an extra space in the indent for continue.

Hence, this is the actual C code that writes the above:

```
GMT_Usage (API, 0, "usage: %s [<table>] -T<tolerance> [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s]\n",
	name, GMT_V_OPT, GMT_b_OPT, GMT_d_OPT, GMT_e_OPT, GMT_f_OPT, GMT_g_OPT, GMT_h_OPT, GMT_i_OPT, GMT_o_OPT, GMT_q_OPT, GMT_colon_OPT, GMT_PAR_OPT);
...
GMT_Usage (API, 1, "-h[i|o][<n>][+c][+d][+m<segheader>][+r<remark>][+t<title>]");
GMT_Usage (API, -1, "Input/output file has [%d] header record(s) [%s]. "
	"Optionally, append i for input or o for output only and number of header records [0]. "
	"Note: -hi turns off the writing of all headers on output since none will be read.  Optionally, append modifiers:",
	GMT->current.setting.io_n_header_items, GMT_choice[GMT->current.setting.io_header[GMT_IN]]);
GMT_Usage (API, 2, "+c Add header record with column information [none].");
GMT_Usage (API, 2, "+d Delete headers before adding new ones [Default will append headers].");
GMT_Usage (API, 2, "+m Insert a new segment header and <segheader> content after the headers [none].");
GMT_Usage (API, 2, "+r Add a <remark> comment to the output [none].");
GMT_Usage (API, 2, "+t Add a <title> comment to the output [none]. "
	"Note: <remark> and <title> may contain \\n to indicate line-breaks. ");
GMT_Usage (API, -1, "For binary files, <n> is considered to mean number of bytes instead of records.");

```

Questions for you:

1. Do you like this layout better?
2. Do you agree with the 1-space indent before the ellipses on the broken lines or might it be OK to drop it since the break/ellipses characters make it clear enough that we are wrapping?
3. This particular option (-h) lists its full syntax on the first line then goes on to explain the option, but for others I do not think we list things.  We should be consistent.  Sometimes the synopsis does a short-cut and says [-Qparams] while in other places we say [-Qlong/list+mmod+gmod] and spell it out.  So we are very inconsistent.  If you are actually looking in the terminal I think it would be helpful to see the full syntax on that start line rather than the synopsis line.  if so, do we then show the full option in both places?  Having -Qparam in the synopsis does not really tell me much since I still have to run the full usage to remember what params are, but having  [-Qlong/list+mmod+gmod] may be enough for me.  Thoughts?
4.

PaulWessel:
Last, would you want more indention before the modifiers above or OK to start at the current level of indent?

meghanrjones:
This looks great! I think the single spaces can go. I don't find the cases of -Qparams to be useful in the terminal or html docs and would prefer more details.

Here's a side by side comparison of old (left) versus new (right) `simplify` on 80 character terminals. Much nicer!
image

meghanrjones (approved):

PaulWessel:
Second thought on indent for modifiers.  Here is what the rewritten -l usage looks like now:

```
  -l[<label>][+D<pen>][+G<gap>][+H<header>][+L[<code>/]<txt>][+N<cols>][+S<size>[/⏎
     …<height>]][+V[<pen>]][+f<font>][+g<fill>][+j<just>][+o<off>][+p<pen>][+s<scale>]⏎
     …[+w<width>]
     Add symbol, line or polygon to the legend. Optionally, append <label> and any of the
     legend codes:
     +D Draw horizontal line with specified <pen>.
     +G Add a vertical <gap> of given height.
     +H Place a <header> in the legend.
     +L Place a <label> with optional justification [L].
     +N Set current number of columns for symbols.
     +S Set the current symbol <size>.
     +V Enable vertical lines between columns, optionally append <pen>].
     +f Set <font> used for the label.
     +g Set a frame fill <fill> [white].
     +j Set the justification of the legend [BL].
     +o Shift legend placement from reference point by <off> [0.2c].
     +p Draw frame outline, optionally append <pen> [1p].
     +s Set an overall symbol scale [1].
     +w Set a specific legend width [auto].
```

I wonder if it is better or clearer if I indent the modifiers:

```
  -l[<label>][+D<pen>][+G<gap>][+H<header>][+L[<code>/]<txt>][+N<cols>][+S<size>[/⏎
     …<height>]][+V[<pen>]][+f<font>][+g<fill>][+j<just>][+o<off>][+p<pen>][+s<scale>]⏎
     …[+w<width>]
     Add symbol, line or polygon to the legend. Optionally, append <label> and any of the
     legend codes:
       +D Draw horizontal line with specified <pen>.
       +G Add a vertical <gap> of given height.
       +H Place a <header> in the legend.
       +L Place a <label> with optional justification [L].
       +N Set current number of columns for symbols.
       +S Set the current symbol <size>.
       +V Enable vertical lines between columns, optionally append <pen>].
       +f Set <font> used for the label.
       +g Set a frame fill <fill> [white].
       +j Set the justification of the legend [BL].
       +o Shift legend placement from reference point by <off> [0.2c].
       +p Draw frame outline, optionally append <pen> [1p].
       +s Set an overall symbol scale [1].
       +w Set a specific legend width [auto].
```

PaulWessel:
I think I am answering this myself: Much cleaner with the extra indent for listings of modifiers.  I am doing all the common usage messages and upgrading some of the very terse GMT_?_OPT constants in gmt_synopsis.h to show all the modifiers, e.g., the very long **-l** option.

PaulWessel:
In an effort to document GMT_Usage and improve what the indent level means I made this graph:

indents

The items you see were placed by several `GMT_Usage (API, level, text)` calls (only showing the **-i** part here):

```
GMT_Usage (API, 1, "%s", GMT_i_OPT);
GMT_Usage (API, -1, "Set alternate input column order and optional transformations [Default reads all columns in original order]. "
	"Append list of columns; t[<word>] selects the trailing text; append <word> to pick a word from the text. Use -in to select numerical input only. "
	"Optional modifiers per column or column group:");
GMT_Usage (API, 3, "+d Divide column by appended <divisor>.");
GMT_Usage (API, 3, "+l Take log10 of column.");
GMT_Usage (API, 3, "+o Add to column the appended <offset>.");
GMT_Usage (API, 3, "+s Multiply column by appended <scale>.");
```

 level is a signed integer and the behavior is this:

- _level_: Place the start of the given text at the indicated indent position in the picture.  Once the line wraps then the continuation of the line is indented one level further and stays at that next level until the end of text.
- _-level_: Behaves similarly excepts it immediately indents to the next level from the first line onwards.

Thus, setting a modifier explanation that may go over multiple lines requires level 3 since we want it to wrap and indent those extra 3 spaces.  If there is some trailing paragraph for this option after modifiers then we use level = -1 since that immediately indents to 2 and stays there.  I still call this "1" since it is part of the option level (1) but the initial start at 1 has already happened earlier, hence -1 to get that immediate indent.

Does this make sense?  So level = 2 means add a paragraph that starts at the level 2 indent but any wrapped lines will be aligned at 3.  -2 means immediately indent to level 3.  I think we need clear instructions if we are to request help from volunteers on this.

PaulWessel:
Alternatively, perhaps it makes more sense if -level also starts at the level indent but then indents no further even when wrapping.  So then my -1 examples above should be -2.  I.e., in all cases |level| indicates the starting ident, with only non-negative levels increasing once for subsequent lines.

PaulWessel:
Again, I may answer my own question here: I think letting |indent| always mean the same indent for the first line is simpler to think about than the first explanation.  So the minus sign simply turns off further indents for this line while a positive level auto-indents once the first line is set.  Clear I think.

meghanrjones:
I think the final decisions should get documented in a new section in [CODE_CONVENTIONS.md](https://github.com/GenericMappingTools/gmt/blob/master/CODE_CONVENTIONS.md), in addition to the comments in gmt_api.c.

PaulWessel:
That is a good idea.  I will first update my implementation to follow my level documentation and then I can format something for inclusion in CODE_CONVENTIONS.md - that will be helpful when recruiting volunteers as well.

doc/rst/source/explain_-aspatial.rst_
```diff
@@ -1,2 +1,2 @@
-**-a**\ [*col*\ =]\ *name*\ [,\ *...*] :ref:`(more ...) <-aspatial_full>`
+**-a**\ [[*col*\ =]\ *name*\ [,\ *...*]] :ref:`(more ...) <-aspatial_full>`
```
meghanrjones, line 1: Can **-a** really be used without *col*=*name*?

PaulWessel: That is what the docs say ...

meghanrjones: I interpret the quote below from the docs as at least one instance of *col*=*name* is required, but this change in the PR indicates that it is optional.

> give one or more comma-separated associations col=name

PaulWessel:
I will check it on some files I have but headed out for some walking right now.
I have literally fixed dozens of minor errors in the docs as I have rewritten lots of the usage messages for the common options, trying to standardize on how we do modifiers (one per line, etc).  It is a tough slow and there is near repetition between gmt_syntax() and gmt_explain_options - I think I need to eliminate that...

PaulWessel:
```
gmt convert -a Wobbe++_2012_G3.gmt | head
C32n.1n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C32n.1n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C24n.3n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C24n.3n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C20n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C20n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
C28n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
```

meghanrjones:
> ```
> gmt convert -a Wobbe++_2012_G3.gmt | head
> C32n.1n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C32n.1n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C24n.3n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C24n.3n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C20n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C20n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> C28n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> ```

OK - I just noticed that it says that later on in the paragraph but think that it could be made more clear in the first two sentences. I can format the ReST docs for **-a** to be more clear as part of #4969.

doc/rst/source/gmtconvert.rst
```diff
@@ -116,11 +116,11 @@ Optional Arguments
 
 .. _-F:
 
-**-F**\ [**c**\|\ **n**\|\ **r**\|\ **v**][**a**\|\ **f**\|\ **s**\|\ **r**\|\ *refpoint*]
+**-F**\ [**c**\|\ **n**\|\ **p**\|\ **v**][**a**\|\ **f**\|\ **s**\|\ **r**\|\ *refpoint*]
```
meghanrjones, line 119: Is this a documentation fix or something that's being changed in the source code as part of this PR? I haven't build this branch, but am unable to get -Fp to work (working from the base code in (test/psxy/qoutedlinedash.sh)

PaulWessel: I decided it was dumb to use **r** twice in this syntax so I switch r for refpoint to **p** for refpoint.  I changed the code to honor either choice.  I did not test anything though but it was a trivial change so I assume it works - in this branch.

PaulWessel:
> > ```
> > gmt convert -a Wobbe++_2012_G3.gmt | head
> > C32n.1n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C32n.1n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C24n.3n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C24n.3n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C20n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C20n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C28n	y	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > C28n	o	N/A	Wobbe++_2012_G3	10.1029/2011GC003742	Magnetic Anomaly
> > ```
> 
> 
> OK - I just noticed that it says that later on in the paragraph but think that it could be made more clear in the first two sentences. I can format the ReST docs for **-a** to be more clear as part of #4969.

Thanks, that would be good.  Also, I need to check this further since clearly the above is only outputting the trailing text and no lon,lat values...  But it does seem to pick up all the aspatial info, so that part seems to work.

PaulWessel:
As this branch rewrites many of the syntax/usage functions in gmt_init.c it may become tricker to merge new changes into master that then may be in conflict with this branch, requiring manual fixing.  I think, not sure though.
One thing I notice is that we have the function GMT_Option () which prints out the syntax fo any of the common functions.  It is used in all the usage messages, for instance.  However, if the parser detects an error during parsing, it will call gmt_syntax () to report what the syntax should be.  Unfortunately, the content of gmt_syntax is a bit out of sync with GMT_Option (which is a front to gmt_explain_options).  It is of course crazy to have to maintain two separate functions that are largely the same.  I think the initial impulse for this way back was that gmt_syntax would be a bit briefer (especially for things like -B and -J), but I am not sure that is so important if the user has a syntax error in their script.  So I suggest we eliminate gmt_syntax and instead use GMT_Option so we only have to maintain one correct set of usage messages.
I assume nobody is going to argue against that, @joa-quim @seisman @meghanrjones ?  If not I will do this in this branch which is the place to do it.

meghanrjones:
> As this branch rewrites many of the syntax/usage functions in gmt_init.c it may become tricker to merge new changes into master that then may be in conflict with this branch, requiring manual fixing. I think, not sure though.
> One thing I notice is that we have the function GMT_Option () which prints out the syntax fo any of the common functions. It is used in all the usage messages, for instance. However, if the parser detects an error during parsing, it will call gmt_syntax () to report what the syntax should be. Unfortunately, the content of gmt_syntax is a bit out of sync with GMT_Option (which is a front to gmt_explain_options). It is of course crazy to have to maintain two separate functions that are largely the same. I think the initial impulse for this way back was that gmt_syntax would be a bit briefer (especially for things like -B and -J), but I am not sure that is so important if the user has a syntax error in their script. So I suggest we eliminate gmt_syntax and instead use GMT_Option so we only have to maintain one correct set of usage messages.
> I assume nobody is going to argue against that, @joa-quim @seisman @meghanrjones ? If not I will do this in this branch which is the place to do it.

Sounds good. This seems to be turning into the new themes branch...

joa-quim:
OK, this is master so things may already a bit different here.When I type ``gmt psxy`` I see a very long man for **-S** and short ones for **-B -J -R**. It is normal the **-S** is long because it's the *main one* in this module. But **-J** is much more variable. It's quite long in *grdproject* and *grdedit* (for example), medium in *psxy* and tinny in *grdcut*. So it we have more then 2 sorts of **-J** explanations. I always have felt that a very long **-J** is not really helpful on the on line help (and it doesn't even mention proj4 or EPSGs). I think the version displayed by *psxy* is OK and could be the one that go into the unique GMT_Option(). For comparison, **-B** is never very long.

PaulWessel:
RIght. gmt_explain_options have both the codes 'B' and 'b' which corresponds to the full and shortened usage messages.  Same for 'J' and 'j'.  In the GMT_Options interface these are translated to "B" and "B-" and "J" and "J-". Various modules passes either B or B- etc to GMT_Options to get the version they want.  The tiny one in grdcut is not served by GMT_Option but by hardwired text in that module.  So all this is OK-ish and served by GMT_Option and hence gmt_syntax can go away I think.  I am experimenting now with a "new" gmt_syntax that simply prints the error message and calls GMT_Option instead.  Seems OK but still testing.

PaulWessel:
I have hallowed out gmt_syntax and simply call GMT_Option instead, only after making sure -B -J gets the reduced messages regardless of module.

joa-quim:
I always wished that we could do ``basemap -B`` and get only the help for that option. While this would very complicated to program fore all options, I think it wouldn't be very complicated to do for the global options now that they are controlled by a unique function.

PaulWessel:
Yes, possible.  The only change today is that they are controlled by a single unique function instead of two slightly different one doing lots of duplication.  I guess implementation would be something like this:

1. Do a quick loop over the options and see if there is only one and it has no arg.
2. If that option is a common option then call GMT_Option on it, then exit

So should not be too hard.  gmt grdtrack -E would not work without major work so no.

PaulWessel:
For your amusement: If I insert this section near the end of GMT_Create_Options then @joa-quim gets what he wants.  Obviously, the "BJR" string need to have all common option letters".

```
	if (n_args == 1 && strchr ("BJR", head->option)) {
		char string[3] = {""};
		string[0] = head->option;
		GMT_Option (API, string);
		API->error = GMT_MODULE_USAGE;
		GMT_Destroy_Options (API, &head);
		return (NULL);
	}
```

Shall we go there?

joa-quim:
A further usage of it would be for the externals that could simply ask GMT to print the on-line help of the common options ... after the long-options are in action.

PaulWessel:
OK.  I will make a PR for this.

joa-quim:
Well, ``pygmt`` and the MEX can make an immediate use of it since they doesn't expand the options modifiers and use the hard-rock GMT syntax.

PaulWessel:
I think at this point I will need @GenericMappingTools/core and @KristofKoch to build this branch and run

```
gmt simplify
gmt plot
gmt basemap
```

in a terminal of their desired width and let me know if you see anything you do not like, any inconsistencies, and suggestions for improvements, etc.  For any common options not used by these three you can do other modules that use them and only look at that section since these three are the only ones I have completed.

joa-quim:
I see strange things. The synopsis is not formatted and ``gmt psbasemap``, ``gmt basemap`` do different things. Also the body help strings do not seem correctly formatted, displaying sort as well long lines.

image

meghanrjones (commented):
Some first comments (I can review again after these are fixed):
`gmtsimplify`
* *table* is listed as optional on terminal and in synopsis, but as a required argument here: https://docs.generic-mapping-tools.org/dev/gmtsimplify.html#required-arguments

`plot`
* synopsis is missing -T

`basemap`
* There are two **-D** options in the synopsis, while there are none in the html docs or when `gmt basemap -^` is used


It's inconsistent in `gmt psxy` and `gmtbasement` whether the full usage is printed at the start of the full description. For example, for ** plot -A** it's `-A Suppress drawing...` and for **plot -E** it's
```
-E[x|y|X|Y][+a][+c[l|f]][+n][+p<pen>][+w<width>]
     Draw (symmetrical)...
```
I prefer the version for `-E`

Common options
- **-bi** and **-bo** - Html docs say that **d** is double precision float while terminal docs say **D** is double precision float
- **-e** - the **+f** option is not included in the ReST documentation synopsis https://docs.generic-mapping-tools.org/dev/gmt.html#e-full
- **-f** - could add that **-f** can be used multiple times

PaulWessel:
Hi @meghanrjones, did you make sure you checked out the latest version of this branch? I cannot match your gmt_message issues above since they are mostly gone since yesterday.

meghanrjones:
> Hi @meghanrjones, did you make sure you checked out the latest version of this branch? I cannot match your gmt_message issues above since they are mostly gone since yesterday.

Sorry, I thought that this comment was about Joa-quim's screenshots. I will check in a minute - committing some changes on a different branch now.

PaulWessel:
No worries, I will apply some of the fixes that do match.

doc/rst/source/explain_-icols_full.rst_
```diff
@@ -1,12 +1,12 @@
-**-i**\ *cols*\ [**+l**][**+d**\ *divide*][**+s**\ *scale*][**+o**\ *offset*][,\ *...*][,\ **t**\ [*word*]]
+**-i**\ *cols*\ [**+l**][**+d**\ *divisor*][**+s**\ *scale*][**+o**\ *offset*][,\ *...*][,\ **t**\ [*word*]]
```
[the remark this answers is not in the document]
PaulWessel: OK, got those.

src/gmt_synopsis.h
```diff
@@ -104,20 +110,25 @@
...
+#define GMT_i_OPT	"-i<cols>[+l][+d<divisor>][+s<scale>][+o<offset>][,...][,t[<word>]]"
 #define GMT_j_OPT	"-je|f|g"
-#define GMT_l_OPT	"-l<label>[<mods>]"
+#define GMT_l_OPT	"-l[<label>][+D<pen>][+G<gap>][+H<header>][+L[<code>/]<txt>][+N<cols>][+S<size>[/<height>]][+V[<pen>]][+f<font>][+g<fill>][+j<just>][+o<off>][+p<pen>][+s<scale>][+w<width>]"
+#ifdef DEBUG
+#define GMT_n_OPT	"-n[b|c|l|n][+A][+a][+b<BC>][+c][+t<threshold>]"
+#else
 #define GMT_n_OPT	"-n[b|c|l|n][+a][+b<BC>][+c][+t<threshold>]"
+#endif
 #define GMT_o_OPT	"-o<cols>[,...][t[<word>]]"
```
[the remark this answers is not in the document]
PaulWessel: Not sure how best to write this.  if there is no <cols> then there cannot be a leading comma.  So -i2,3,t2 is find but -i,t2 is not, it must be -it2.

meghanrjones:
I probably missed the `-T` for `plot` earlier because it is there now. For basemap, this is what I get where there are two **-D** options with | but without a carriage return symbol.

image

meghanrjones:
This is a screenshot from `gmt plot` (just pulled and built) of the inconsistencies around the synopses (-A, -B, -C, -D are formatted one way and -E and -F are another)

image

PaulWessel:
The two basemap -D are separated by " | " so it is a clean word break and no return symbol.  If I lump them together in a single -D with no repeated -D and just "|" then it will break at some ][ point but would be very long and complicated.  Finally, I need to figure out why they are even showing up since basemap is modern mode and those old -D are for insets in psbasemap...

src/gmt_synopsis.h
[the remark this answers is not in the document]
meghanrjones: I agree that it's tricky - I think it makes slightly more sense to include the comma in the docs.

meghanrjones:
> The two basemap -D are separated by " | " so it is a clean word break and no return symbol. If I lump them together in a single -D with no repeated -D and just "|" then it will break at some ][ point but would be very long and complicated. Finally, I need to figure out why they are even showing up since basemap is modern mode and those old -D are for insets in psbasemap...

Yes, I think the fact that they are listed for modern basemap is the real problem.

meghanrjones:
OK, I checked again and the `-T` is not here (seems to be terminal width dependent):

image

PaulWessel:
Odd.  OK, will check.

PaulWessel:
Looks like **gmt plot -** and **gmt plot** behaves differently.  Same with basemap for -D.

PaulWessel:
Looks like when there are no args the modern mode flag is never set (or probably set but lost in the shuffle between GMT and the copy of GMT.....
I think GMT either runs in modern mode or in classic and never a mix, so I think the run_mode parameter should live in API-> and not in GMT->.  That is the problem here.

PaulWessel:
I am considering to add a blank line before each section in the usage listing.  This is what it would look like in the terminal.  It seems cleaner to me and easier to separate the sections.  What do you think, people?

Untitled

KristofKoch:
Hi @PaulWessel I like your idea. I think it improves readability quite a bit. But it doesn't work for me in 3b80696 with `gmt plot`? (Apologies for my absence … work again)

Screenshot 3b80696 gmt plot

PaulWessel:
Sorry I did not commit those yet, just posted the example. Getting 2nd COVID-19 shot so back home in about an hour.

PaulWessel:
OK, I think we want those extra lines as they clearly makes it easier to see the paragraphs.  I  have update the code for the common options and the three modules gmtsimplify, plot, and basemap.

KristofKoch:
Nice, @PaulWessel! I really like it. While reading through the reformatted `plot` output I got another idea on how to improve readability: **Bold text** to visually separate commands from text. Both you and I use some kind of text styling on GitHub and in the Forum to indicate what is a command and what is regular text. Why not in the shell as well?

Bold text

Try for yourself: 
`printf "\e[m\nAlternatively, append c<period>[/<phase>] for custom cyclicity.\n\nAlternatively, append \e[1mc<period>[/<phase>] \e[mfor custom cyclicity.\e[m\n\n"`

I don't know about Windows. Maybe @joa-quim can paste `printf "\e[mnormal text \e[1mbold text\e[m\n"` into a terminal and see if he gets something like 

normal and bold text

And yes, I smell feature creep. It has nothing to do with "Explore how to break synopsis lines". Your thoughts?

KristofKoch (approved):
Really happy with the new layout!

PaulWessel:
Hi @KristofKoch.  @meghanrjones and I decided that even if we started now it would take some time to implement this new scheme across all modules without seriously delaying GMT 6.2.0.  So I think we will continue work on this branch but not let is stop us from releasing 6.2.0.  We sill start with RC1 soon but probably after our community meeting on Thursday.  Thus. we can continue to experiment with bold/italics styles - it is an interesting suggestion and I guess there are escape sequences to do the switcharoo.  Your examples above did not do anything for me though (macOS), so this depends on other things than UTF-8.

meghanrjones:
> Hi @KristofKoch. @meghanrjones and I decided that even if we started now it would take some time to implement this new scheme across all modules without seriously delaying GMT 6.2.0. So I think we will continue work on this branch but not let is stop us from releasing 6.2.0.

I agree

KristofKoch:
I agree not to delay 6.2 for this, @PaulWessel and @meghanrjones.

The bold font doesn't work for you @PaulWessel? I'm on MacOS as well and it works for me in the default _Terminal.app_ and in my usually used _iTerm.app_.

**Terminal.app**
terminal_app

**iTerm.app**
term_app

PaulWessel:
Turns out it does not work when your Terminal's font was set to Courier Bold...
If I select another non-bold font then it works.

PaulWessel:
We will merge this into master and prioritize updating the 150 modules for 6.3.  All comment options have been updated already.  I just now replaced OPTIONS: with OPTIONAL ARGUMENTS and added REQUIRED ARGUMENTS to the three modules I have worked on.  Please have a final check - one question: I think we may want to change the indentation of the ...ARGUMENTS line since it is using a tab which is no longer used by the others.  Perhaps 2 spaces will work here?

meghanrjones:
> We will merge this into master and prioritize updating the 150 modules for 6.3. All comment options have been updated already. I just now replaced OPTIONS: with OPTIONAL ARGUMENTS and added REQUIRED ARGUMENTS to the three modules I have worked on. Please have a final check - one question: I think we may want to change the indentation of the ...ARGUMENTS line since it is using a tab which is no longer used by the others. Perhaps 2 spaces will work here?

Yes, I agree that it would look better aligned evenly with either 'usage' or the arguments, but not indented further than everything else.

PaulWessel:
OK, done, will wait for internal tests to give green light, then merge.

PR #4800, opened 2021-02-12
Fix mistakes in subplot for axes not annotated by -S

PaulWessel:

When **-S** is used to control which axes should be annotated, the axes not selected for annotation shall be drawn (or now) following any specific **-B** given to an individual panel, a **-B** given to subplot begin, or the prevailing **MAP_FRAME_AXES** setting.  The bug was that it only checked if the lower case _wesn_ was specified whereas in the case where just _S_ should be annotated, a _N_ should be interpreted as n.

This PR updates five wrong PS due to the bug and completes the addition of nine subplot_x.y.ps files that had not yet been added due to the bug.


meghanrjones:
These look good. How would a user specify to not draw the frame or annotations for a given set of axes, if they want to return to the previous look? Is there a way to use something like -B+n for this?

PaulWessel:
Presumably they would use MAP_FRAME_AXES=WSne or in modern theme I think WSrtZ

PaulWessel:
I mean with the modern theme some of these plots will revert to what they looked like before this PR, but that was just because of this bug.  What I submitted is the "correct" plots given the defaults in master, in particular MAP_FRAME_AXES being WESNZ.

meghanrjones:
> Presumably they would use MAP_FRAME_AXES=WSne or in modern theme I think WSrtZ

Wouldn't this still draw the frames, just not the ticks/annotations? Is there a specification for not drawing the frames either? Similar to --MAP_FRAME_TYPE=graph, but for specific axes and without the arrows. I think this should be possible for subplots, but could be addressed separately from this PR.

PaulWessel:
I think **MAP_FRAME_TYPE**="" will only draw the annotated axes selected by -S.  The graph type is interesting.  Currently we make no space in subplot for the extension - I ahd to add -M12p to make this look OK:

```
gmt begin subplot_2.0
	gmt set FONT_HEADING 28p,Helvetica,black MAP_FRAME_AXES "" MAP_FRAME_TYPE graph
    gmt subplot begin 2x2 -Fs8c -SCb -SRl -R-2/2/0/10 -T"2.0: -SCb" -M12p
       	gmt basemap -c
        gmt basemap -c
        gmt basemap -c
      	gmt basemap -c
    gmt subplot end
gmt end show
```

But that we can fix downstream since the same in master.

meghanrjones (approved):

PaulWessel:
Thanks, now we probably need to merge master into the themes branch.  There will likely be conflicts and you handled those better than me last time - want to give that a try?

meghanrjones:
Yes, I'll give it a go.

Issue #4795, opened 2021-02-12
no display on gmt plot -L past dateline


delaviel-anger:
Hi

Trying to use this 
```
gmt plot  -R110/280/${day1}/${day2} -JX10cd/-10cT -Wthinnest,black -GP300/8 -L -t50 <<- END
$a \${day1}
280 \${day1}
280 \${day2}
$a \${day2}
END
```
As soon as `$a` is greater than 180, it doesn't work.

(It's done within `gmt batch`)

update : it appears to be a problem with the -L option
```
gmt plot  -R110/280/${day1}/${day2} -JX10cd/-10cT -Wthinnest,black -GP300/8 -Sc -t50 <<- END
$a \${day1}
280 \${day1}
280 \${day2}
$a \${day2}
END
```
works
(I managed a quick work around using `-Sr+s`)

PaulWessel:
Fixed via #4802.

PR #4802, opened 2021-02-13
Distinguish between geographic and separate lon or lat wrapping

PaulWessel:
We were abusing the macro _gmt_M_is_geographic_ to determine if x is longitude but that macro also checks that y is latitude.  In many of the instances we simple wanted to use _gmt_M_x_is_lon_ or _gmt_M_y_is_lat_ separately, which was the trouble behind #4795.  The _gmt_M_is_geographic_  macro should only be used when we are dealing with mapping issues (i.e., when map projections are set).  Closes #5795.


joa-quim (approved):
Many changes. Let's see what future brings.

PaulWessel:
Yes, but all tests pass.

PR #4808, opened 2021-02-14
Let pscoast -E+l|L trigger dataset output via KEYS for externals

PaulWessel:
Cryptic I know, but mean to help externals (MEX, Julia, Python) run coast **-E+l**|**L** and receive back a dataset with names of countries and states.  Need @joa-quim to test it though.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -1422,6 +1422,9 @@ GMT_LOCAL unsigned int gmtapi_add_existing (struct GMTAPI_CTRL *API, enum GMT_en
 #define K_PRIMARY			0
 #define K_SECONDARY			1
 
+#define K_OR			0
+#define K_AND			1
+
 #define API_PRIMARY_INPUT		'{'
 #define API_PRIMARY_OUTPUT		'}'
 #define API_SECONDARY_INPUT		'('
@@ -1609,7 +1612,7 @@ GMT_LOCAL char ** gmtapi_process_keys (void *V_API, const char *string, char typ
 					change_type = false;
 					for (kk = 3; s[k][kk]; kk++) {	/* Examine characters in the modifier string */
 						if (strchr ("-+", s[k][kk])) {	/* Start of all (+) versus at least one (-) */
-							kase = (s[k][kk] == '-') ? 0 : 1;	/* Set kase and go to next letter */
+							kase = (s[k][kk] == '-') ? K_OR : K_AND;	/* Set kase and go to next letter */
 							continue;
 						}
 						count[kase]++;	/* How many AND and how many OR modifiers (depending on kase) */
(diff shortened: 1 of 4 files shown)


joa-quim:
Nope, still prints to stdout. The line you added is correctly visited and ``type`` set to ``D`` but not enough.

PaulWessel:
Could you please try to set the KEYS to

`#define THIS_MODULE_KEYS	">?},}DE-lL"`

build and test.  This says "if -E is given and if either +l or +L are given then switch primary output to D"

If it does not work with } then try ).

joa-quim:
No, none works. Even tried to put the ``}`` or ``)`` at the end of the KEY (this is new type, no?)
If you want to test it in Julia, it's only do
```
coast(E="+l");
```

PaulWessel:
Working on this, trying to re-understand the magic.  Will document it if I do.

PaulWessel:
Seems related to the fact that -M uses GMT_Put_Record and this goes into a memory location while -E+l uses GMT_Write_Data to NULL which is straight to stdout.   I think the simplest solution for me is to use GMT_Put_Record for -E as well and then it should hopefully work.  GMT_Encode_Options do end up with >D} so all should be OK...

PaulWessel:
OK @joa-quim this works for me with MEX.  The only odd thing is a Matlab thing where `"AO\tAngola"` in Matlab shows as `"AO->Angola"`, i.e. the tab is printing as a pointer.  Probably not an issue in Julia though.

src/gmt_api.c
```diff
@@ -1609,15 +1612,15 @@ GMT_LOCAL char ** gmtapi_process_keys (void *V_API, const char *string, char typ
...
+							kase = (s[k][kk] == '-') ? K_OR : K_AND;	/* Set kase and go to next letter */
 							continue;
 						}
 						count[kase]++;	/* How many AND and how many OR modifiers (depending on kase) */
 						modifier[1] = s[k][kk];	/* Set current modifier */
 						if (strstr (opt->arg, modifier)) given[kase]++;	/* Match found with given option */
 					}
 					/* Only change the key if we found all the AND modifiers and at least one of the OR modifiers (if any were given) */
-					if ((count[0] == 0 || (count[0] && given[0])) && count[1] == given[1]) change_type = true;
+					if ((count[K_OR] == 0 || (count[K_OR] && given[K_OR])) && count[K_AND] == given[K_AND]) change_type = true;
```
joa-quim, line 1624: I find that these makes it harder to follow the code. One will never know what is ``K_OR`` without going search code somewhere else.

PaulWessel: IN my case it was the opposite: 0 and 1 were meaningless but now I know if it is boolean OR or AND we are doing with the modifiers.  But I agree it is hard to follow regardless.

joa-quim:
If you do
```
>> D.text{1}

ans =

    'AO	Angola'
```
and
```
>> D.text{1}(3)

ans =

    '	'
```

but the mystery now is that in Julia it returns nothing and prints nothing.

PaulWessel:
OK, forgot how cells work.  Depending on wrapping: I added a new item to the struct **GMT_DCW_SELECT** in gmt_dcw.h if that might affect you.

joa-quim:
No issues there. I don't wrap ``gmt_dcw.h``

joa-quim (approved):

Issue #4805, opened 2021-02-13
Various concerns about pslegend


PaulWessel:
Here are some issues related to **pslegend** that I would like some feedback on

**1. Combining auto- and manual legend**

Right now, if the user uses **-l** to automate basic legend info, but then wants to add some stuff manually, she is out of luck.  There seems to be interest in allowing both the hidden file and regular input to be merged in making the final legend. To do this would likely require an option to let us know that there are two sources.  Perhaps **-M** to merge hidden legend to specified input?

**2. Usage message and modern mode**

Also right now, there are special if-tests in the usage message to output information about the new **-T** option that can save the hidden legend data to a specified file, but only in modern mode.  Since nobody is starting a modern session with `gmt begin` only to then type `gmt legend + ` to see the usage message, I think I need to remobe such checks and instead say

`-T Write hidden legend specification file to <file> (modern mode only).`

Thoughts on this, @GenericMappingTools/core ?

PR #4809, opened 2021-02-14
Let legend handle both hidden and given information

PaulWessel:
Allow merging of hidden and given legend info.  See #4805 for background.  Also fixes the silly usage message restriction.  Tested by modifying the _autolegend.sh_ script and PS.


joa-quim (approved):

RKT361218:
how to toggle the position of legend except TL,BL,.... etc

PR #4811, opened 2021-02-15
Example for anim12 written in Batch script

Squadzone:
for Windows 
execute with command prompt
tested with gmt 6.1.1

### DIFF
--- a/doc/examples/anim12/anim12.bat
+++ b/doc/examples/anim12/anim12.bat
@@ -1,0 +1,61 @@
+REM Blending the NASA day and night views from the Blue and Black Marble mosaic
+REM images using a day-night mask that goes through a full 24-hour cycle.  In addition,
+REM we adjust the colors using the intensities derived from the slopes of the earth
+REM relief grid.  We spin around at 24 frames per second where each frame advances by 1 minute.
+REM Because we are not plotting anything, just manipulating images via grids, we must instead
+REM create the frames and the final movie via the batch module.
+REM
+REM DEM:    @earth_relief_04m
+REM Images: @earth_day_HD.tif @earth_night_HD.tif from the GMT cache server
+REM 
+REM The finished movie is available in our YouTube channel as well:
+REM https://youtu.be/X8TojLs0NYk
+REM The movie took ~4 minutes to render on a 24-core MacPro 2013.
+REM To make a UHD (4k) movie, comment/uncomment the SIZE, INC and FILT settings
+set SIZE=HD
+set INC=11.25
+set FILT=21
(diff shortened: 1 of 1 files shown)


doc/examples/anim12/anim12.bat
```diff
@@ -0,0 +1,61 @@
...
+REM relief grid.  We spin around at 24 frames per second where each frame advances by 1 minute.
+REM Because we are not plotting anything, just manipulating images via grids, we must instead
+REM create the frames and the final movie via the batch module.
+REM
+REM DEM:    @earth_relief_04m
+REM Images: @earth_day_HD.tif @earth_night_HD.tif from the GMT cache server
+REM 
+REM The finished movie is available in our YouTube channel as well:
+REM https://youtu.be/X8TojLs0NYk
+REM The movie took ~4 minutes to render on a 24-core MacPro 2013.
```
seisman, line 13: Quick comments:

1. Remove this sentence or show the minutes on your Windows machine.
2. Add "author/contributor" or similar information to give the contributors (you) more credit?
3. We won't show the batch script in the online documentation, but perhaps we could add a note saying that a batch version is available?

Squadzone: 1. Well, maybe someone with good PC specs with windows can test it? im using old pc 7-core windows 10 and takes ~15minutes to do this job :(
2. im fine with it :)
3. agreed, since most user in my place using windows so better if has example using batch

Squadzone: https://github.com/Squadzone/gmt_example_batch_script

seisman:
@joa-quim I guess we don't want to maintain batch scripts for our animation examples, right?

joa-quim:
Agree that we don't want to maintain (worst, create versions of) animations batch scripts but I think it would be good to have at least one example to show that Windows users can also write their animations with plain `.bat` files.

seisman:
> I think it would be good to have at least one example to show that Windows users can also write their animations with plain `.bat` files.

https://docs.generic-mapping-tools.org/dev/movie.html

image

Is this one enough?

joa-quim:
Well, yes but it wouldn't harm either if we had one of the animations as a batch file. And specially in this case that the user already made it.

seisman:
I guess we need a Windows user to run the batch scripts to make sure they're correct.

seisman:
Actually, the batch version is more difficult to read and understand, because it uses `echo` extensively to write GMT commands into the pre, main and post scripts.

seisman:
I'm closing this PR.

Issue #4810, opened 2021-02-15
Problem cleanup after execute gmt batch in windows 10 64bit using Command prompt


Squadzone:
this is code inside batch_cleanup.bat that generate by gmt batch

```
@echo off
REM Start of script
REM Cleanup script removes working directory with job files
rd /s /q D:/tes_example/temp
```

**Full error message**

```
Invalid switch - "tes_example".
batch [ERROR]: Running cleanup script batch_cleanup.bat returned error 1 - exiting
```

**System information**

* Operating system: Windows 10 64bit
* GMT version (`gmt --version`): 6.1.1

joa-quim:
@PaulWessel it has to be
```
rd /s /q D:\tes_example\temp
```

Squadzone:
thanks

PR #4812, opened 2021-02-15
Use correct DOS paths when issuing batch commands in movie and batch

PaulWessel:
See #4810 for background.  This PR ensures that any removal of directories and files of the form _directory/file_ or _directory/subdir_ considers if these should be reformatted to use DOS backslash.  That is not needed in calls like _chdir_ but _is_ needed on the command line.  Hopefully, @Squadzone can give this PR a spin as well.


PaulWessel:
I should add that I have only coded this up but not able to actually test under Windows.  It looks like the issue only applies to the case when **-W** is used to set an alternative directory, since otherwise the rm commands do not contain directory separators.

joa-quim (approved):

PR #4816, opened 2021-02-16
Fix grdcut example in remote dataset docs

meghanrjones:

This updates the ``grdcut`` example command in the remote datasets documentation because ``grdcut`` does not work without either **-R**, **-S**, or **-Z**.

Fixes #

**Reminders**

### DIFF
--- a/doc/rst/source/datasets/remote-data.rst
+++ b/doc/rst/source/datasets/remote-data.rst
@@ -130,7 +130,7 @@ Should you need a single grid from any of our tiled dataset, e.g., to feed into 
 not depend on GMT, you can create that via :doc:`/grdcut`.  For instance, to make a global grid from the
 eight tiles that make up the 2m x 2m gridline-registered data, try::
 
-    gmt grdcut @earth_relief_02m_g -Gearth_at_2m.grd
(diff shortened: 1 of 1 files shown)


joa-quim:
Perhaps here the example should be using ``grdconvert`` instead of ``grdcut``. It  seems a it odd *cutting to the same size*

meghanrjones:
> Perhaps here the example should be using `grdconvert` instead of `grdcut`. It seems a bit odd _cutting to the same size_

Yes, but this is the simplest way to achieve the stated goal (getting a single grid from the tiled dataset). ``grdconvert @earth_relief_02m_g -Gearth_at_2m.grd -Rg`` would not do this and any additional arguments would make it unnecessarily complicated.

joa-quim:
Hmm, I thought all the modules would do the mosaic if need, but you are right
```
grdconvert @earth_relief_02m_g -Gearth_at_02m.grd=sf
grdconvert (grdconvert.c:245(GMT_grdconvert)): Unknown grid format id number [C:\TMP/=tiled_76_GX.a56344]
```

PaulWessel (approved):
Please open a separate feature request for **grdconvert** since that ought to work (but will require some under-the-hood stuff).

PR #4823, opened 2021-02-17
Improve grdseamount -Q behavior

PaulWessel:

This improves the behavior for **-Q..+d** for polynomial seamounts.  It now has the correct equations for volume and works correctly for no flattening.  The flattening case is still a work in progress and may be improved at a later time.

Also adds missing check on flattening range (0-1).  No tests affected.


PaulWessel:
And just like that I figured out my missing solution for **-Q...+d -Co** so now all cases work and are exact, and I have added a growth.sh test for each of the four seamount types.  I believe this PR is complete and good to go.  For instance, here is the growth of a seamount using the new polynomial function with increments approximated by discs (without the **+d** the increments are the exact difference between time horizons):

growth_o

meghanrjones (approved):

PR #4829, opened 2021-02-19
Fix typo in first example (add xt in extension)

Esteban82:
BTW, I think that the following example could be add. 

If you instead just want to save the data in a file, use:  

  

    gmt pshistogram  @v3206_06.txt -T250 -Io > data.txt 

If you agree, I will add it. However, I can't managed to do it with GMT Modern Mode One-line Commands. It should be possible to be done in modern mode?

### DIFF
--- a/doc/rst/source/histogram.rst
+++ b/doc/rst/source/histogram.rst
@@ -55,7 +55,7 @@ Examples
 
 .. include:: oneliner_info.rst_
 
-To draw a histogram of the remote data v3206_06.t containing seafloor depths,
+To draw a histogram of the remote data v3206_06.txt containing seafloor depths,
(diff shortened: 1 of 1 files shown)


seisman:
> If you instead just want to save the data in a file, use:
> 
> ```
> gmt pshistogram  @v3206_06.txt -T250 -Io > data.txt 
> ```

It is a good example. Please add it.

> If you agree, I will add it. However, I can't managed to do it with GMT Modern Mode One-line Commands. It should be possible to be done in modern mode?

I think it's a bug. Please open an issue.

seisman:
Perhaps better to add the example after the bug is fixed?

joa-quim:
Please don't. WE have so may bugs to fix and this is not really a bug. **-Io|O** is for saving files not to plot anything

PR #4849, opened 2021-02-24
Allow GMT_Put_Vector to convert other text items than datetime

PaulWessel:
See [this post](https://github.com/GenericMappingTools/pygmt/issues/647) for background.  While we added support for **GMT_DATETIME** last year, this PR generalizes the conversion from text to numbers via the type **GMT_TEXT**.  The strings may be either Cartesian numbers, datetime, longitude, or latitude, and we also allow these to be converted to other internal representations than **GMT_DOUBLE** (e.g., **GMT_FLOAT**, **GMT_LONG**, etc.) via logical addition (e.g., pass type as  **GMT_TEXT**|**GMT_FLOAT**).

I have left the prior scheme of **GMT_DATETIME** for datetime so we are backwards compatible, but one now only needs to select **GMT_TEXT** and GMT figures out what you passed, hopefully.  We assume each column has the same type and there is no error checking if you pass both time and longitude as part of a single column.  I added _testapi_putvector.c_ and calling script _apiputvector.sh_ which demonstrates that it works.

I think a quick test and this can be approved and merged.

### DIFF
--- a/doc/rst/source/api.rst
+++ b/doc/rst/source/api.rst
@@ -1369,8 +1369,8 @@ or ISO datetime strings and we do the conversion to internal numerical values an
 allocate a vector to hold the result in the given ``col``.  By default that vector
 will be assigned to type **GMT_DOUBLE** but you can add another primary data type
 for the conversion if you prefer (e.g., **GMT_TEXT**\|\ **GMT_LONG** to get final 
-internal absolute time in integer seconds). For these four special data types GMT
-allocates internal memory to hold the concerted data and ``vector`` is not used
+internal absolute time in integer seconds). For the special data type **GMT_TEXT** GMT
+allocates internal memory to hold the converted data and ``vector`` is not used
 any further.
 
 To extract a custom vector from an output :ref:`GMT_VECTOR <struct-vector>` you can use
--- a/doc/rst/source/api.rst
+++ b/doc/rst/source/api.rst
@@ -1371,7 +1371,7 @@ a vector to hold the result in the given ``col``.  By default that vector will b
 assigned to type **GMT_DOUBLE** but you can add another primary data type for the
 conversion if you prefer (e.g., **GMT_LONG**\|\ **GMT_DATETIME** to get final 
 internal absolute time in integer seconds). For these four special data types GMT
-allocates internal memory to hold the concerted data and ``vector'' is not used
+allocates internal memory to hold the concerted data and ``vector`` is not used
 any further.
 
 To extract a custom vector from an output :ref:`GMT_VECTOR <struct-vector>` you can use
--- a/doc/rst/source/api.rst
+++ b/doc/rst/source/api.rst
@@ -1363,9 +1363,11 @@ For vectors the same principles apply:
 where ``V`` is the :ref:`GMT_VECTOR <struct-vector>` created by GMT_Create_Data_, ``col`` is the vector
 column in question, ``type`` is one of the
 recognized data :ref:`types <tbl-types>` used for this vector, and ``vector`` is
-a pointer to this custom vector.  In addition, ``type`` may be also **GMT_DATETIME**, in which case
-we expect an array of strings with ISO datetime strings and we do the conversion to internal
-GMT time and allocate a vector to hold the result in the given ``col``.
+a pointer to this custom vector.  In addition, ``type`` may be also **GMT_TEXTLON**, **GMT_TEXTLAT**,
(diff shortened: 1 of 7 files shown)


seisman:
Just built this branch and ran the full PyGMT tests, the **GMT_DATETIME** implementation works like before.

PaulWessel:
Great, maybe @weiji14 could test Lon lat inputs

seisman:
As for the new data types, **GMT_TEXTLON**, **GMT_TEXTLAT**, and **GMT_TEXT**, does it mean that PyGMT has to know the data types before passing to GMT?

For example, if the string is `12.34E`, PyGMT must check if it's text, longitude, latitude or datetime, and then choose the correct `type` when calling `GMT_Put_Vector`. It means more work for PyGMT (and other external wrappers) to detect the data types. 

I'm wondering if we can simply pass any strings (e.g., `12.34`, `12.34E`, `12.34N`, `12:30:30`, `2020-01-01T01:00:00` to GMT giving a general data type (`type=GMT_TEXT`) and GMT is very clever to parse them and automatically determine the data type.

PaulWessel:
We could automatically determine type and give an error if more than one, e.g., both time and latitude since a column must only have one type

PaulWessel:
@seisman, this PR is now more align with your suggestion.

PaulWessel:
Pinging @weiji14 and @seisman to consider this PR given it has a test that passes.

seisman:
It works well with a simple test (see https://github.com/GenericMappingTools/pygmt/pull/975 for the example script and output). Need to do more testing later.

doc/rst/source/api.rst
```diff
@@ -1360,12 +1360,19 @@ For vectors the same principles apply:
...
+:ref:`types <tbl-types>` used for this vector, and ``vector`` is a pointer to the
+user's read-only custom vector.  In addition, ``type`` may also be  **GMT_TEXT**,
+in which case we expect an array of strings with numbers, longitudes, latitudes,
+or ISO datetime strings and we do the conversion to internal numerical values and
+allocate a vector to hold the result in the given ``col``.  By default that vector
+will be assigned to type **GMT_DOUBLE** but you can add another primary data type
+for the conversion if you prefer (e.g., **GMT_TEXT**\|\ **GMT_LONG** to get final 
+internal absolute time in integer seconds). For these four special data types GMT
+allocates internal memory to hold the concerted data and ``vector`` is not used
+any further.
```
seisman, line 1374: Fix typos.
```suggestion
internal absolute time in integer seconds). For the special data type **GMT_TEXT** GMT
allocates internal memory to hold the converted data and ``vector`` is not used
any further.
```

src/gmt_api.c
```diff
@@ -14466,56 +14468,139 @@ int GMT_Change_Layout_ (unsigned int *family, char *code, unsigned int *mode, vo
...
+	return vector;
+}
+
 int GMT_Put_Vector (void *V_API, struct GMT_VECTOR *V, unsigned int col, unsigned int type, void *vector) {
 	/* Hooks a users custom vector onto V's column array and sets the type.
 	 * It is the user's responsibility to pass correct type for the given vector.
-	 * We also check that the number of rows have been set earlier. */
+	 * We also check that the number of rows have been set earlier.
+	 * We also allow special text-based arrays for longitude, latitude, datetime or Cartesian data to be passed
+	 * which may be logically OR'ed with desired array type (e.g., GMT_LONG|GMT_TEXT.
```
seisman, line 14518:
```suggestion
	 * which may be logically OR'ed with desired array type (e.g., GMT_LONG|GMT_TEXT).
```

src/gmt_api.c
```diff
@@ -14466,56 +14468,139 @@ int GMT_Change_Layout_ (unsigned int *family, char *code, unsigned int *mode, vo
...
+	 * Note: We do not check for data loss in the conversion (e..g, GMT_UCHAR|GMT_TEXT) */
```
seisman, line 14519:
```suggestion
	 * Note: We do not check for data loss in the conversion (e.g., GMT_UCHAR|GMT_TEXT) */
```

seisman (approved):
Looks good to me.

seisman:
@weiji14 Could you please bump the GMT dev version again in the conda-forge channel, so that we can better test the new **GMT_TEXT** data type.

weiji14:
Ok, I'll do that in a bit

PR #4851, opened 2021-02-25
Improve gridline spacing for parallels

PaulWessel:
When the user selects automatic gridline interval (e.g., **-Bag**) we determine a suitable interval separately for lon and lat, then simply use the largest.  That works well for large map regions, but if you are doing a small map near a pole, and especially for oblique maps where the pole is inside the map frame, the longitude range of 360 may swamp the latitude range. This leads to large parallel spacings of 15-30 degrees even if the map is just a few degrees of latitude around a pole, and hence no parallels are drawn.  This PR checks for these cases and allows for separate intervals when the situation requires it (for annotations, ticks, and gridlines).  Thankfully, only 8 PostScript files needed to be updated; they mostly were polar views with basically no parallel gridlines except the polar cap.  The condition to stick with the single larger interval is used if the latitude range is at least 25% of the longitude range (e.g., a polar view from Equator to pole, for instance.  With this change we get behavior like this:

a

where latitude range is < 5 and longitude range is 360.


PaulWessel:
FYI, here is what it looks like without this PR:

a

joa-quim (approved):

Issue #4869, opened 2021-02-28
Better option coordinate among symbol modules mega, coupe, polar, velo


PaulWessel:
These modules were initially created by IPGG in Paris in the 1990s and eventually donated to us for continued upkeep.  Because no guidelines for option selection existed at the time, I have had to make major changes over the years, such as preventing that these modules eat up available common options for the rest of GMT (e.g., **-o**, etc).  Yet, we still have some discordant notes in the system.  Today we fixed **coupe** **-Z**_cpt_ to be **-C**_cpt_ like in the rest of GMT.  However, **meca** also uses **-Z** that way, and **-C** is used and does something else.  I have looked at the four modules, determined available option codes, and looked for simplicities. Here is my assessment and proposal for an update.  The update will be backwards compatible, of course.

1. CPT color are optionally needed by **meca**, **coupe**, might be a very useful addition for **velo**, and possibly for **polar** if there is a use for this.  Thus, I wish to make space for **-C** for all four, wether they presently use a CPT or not.
2. **meca** and **polar** have a similar **-C** option, but with some differences: In **meca**, **-C** forces us to read two extra columns with an alternative lon,lat location where the symbol will be plotted, and then a line is drawn from that location to the original location.  Optionally, a small circle (pen only) is drawn at the original location.  For **polar**, there really is no lon,lat input so **-D** is used to specify where on the map you want this hemisphere plot to be centered.  Its **-C** option then takes a second location and then the same thing happens: hemisphere plot is placed at **-C**_lon/lat_ and a line with optional circle goes to the **-D**_lon/lat_ point. Here, I think it makes most sense to let the functionality of **-C** be just modifiers to **-D** instead, thus leaving **-C** free.  For **meca** I think it is best not to clutter **-S** with more modifiers but simply introduce **-A** as the new option to take over for the old **-C**.  That way can use **+a** with **polar** **-D** for some mnemonic value.
4. I guess because of the projection onto the crossection, **coupe** does not have a similar **-C** facility to adjust location of symbol.  Might this be conceivable useful to allow for, @seisman ? If it is, then uniformity across these modules would imply we use the common **-Z** letter (and **+z** in polar) instead of **-A**.
5. Of course, the other issue with the existing **-C** option is that one takes a pen while the other takes a **+p**_pen_ modifier.  Both take an optional **+s**_size_ to affect the circle size.

Proposal:

1. Make all these module have **-C**_cpt_ or have the capacity to add one
[...]

PR #4870, opened 2021-02-28
pspolar: Merge -C with -D and improve parsing

PaulWessel:
See #4869 for background.

To make room for a possible **-C**_cpt_ option across many modules, we let the optional alternate point be set via a modifier to **-D** (which sets the primary point) instead. The new parsing allows for long/lat to be given in the usual geographic syntax accepted elsewhere in GMT (i.e., ddd:mm:ssF as well as decimal).

A few concerns remain (apart from the letter used for the modifier in **-D** – currently **+z**):

1. The description of **-M** is strange: "Sets the size of the beach ball to plot polarities in".  We are not plotting beachballs here but the lower half of the focal sphere, no?  Copy/paste issue? I can understand the **+m** modifier to scale size to magnitude for beachballs.  Does this still make sense for a focal sphere and we just update the language?
2. The default size of the small circle at the point set via **-D** when an alternate location is given is listed as 0.015i. But the code is using **GMT_DOT_SIZE** which is actually 0.005i.  So that is a factor of 3 between usage documentation and actual values, and the documentation is quiet about it.  The 0.005i is 0.36 points so I put in 0.5 points so the circle can be seen.  However, I never make these plots so I do not know what is a useful default, but being smalle than the pen thickness of 0.25p cannot be useful. **Note**: This will be the same issue for **meca** (and possibly **coupe** if that capability is added). I also note that **meca** uses the same 0.005i default diameter but says in the documentation it is 0 (presumably means no circle).  So there needs to be a common default and common language used in both usage and docs across these modules.

Hoping @seisman can opine on these issues.

### DIFF
--- a/doc/rst/source/supplements/seis/polar.rst
+++ b/doc/rst/source/supplements/seis/polar.rst
@@ -14,7 +14,7 @@ Synopsis
 
 **gmt polar**
 [ *table* ]
-|-D|\ *lon/lat*\ [**+z**\ *lon*/*lat*\ [**+p**\ *pen*][**+s**\ *size*]]
+|-D|\ *lon/lat*
 |-J|\ *parameters*
 |SYN_OPT-R|
 |-M|\ *size*\ [**+m**\ *mag*]
--- a/doc/rst/source/supplements/seis/polar.rst
+++ b/doc/rst/source/supplements/seis/polar.rst
@@ -12,12 +12,14 @@ Synopsis
 
 .. include:: ../../common_SYN_OPTs.rst_
 
-**gmt polar** [ *table* ] |-D|\ *lon/lat* |-J|\ *parameters*
+**gmt polar**
+[ *table* ]
+|-D|\ *lon/lat*\ [**+z**\ *lon*/*lat*\ [**+p**\ *pen*][**+s**\ *size*]]
+|-J|\ *parameters*
 |SYN_OPT-R|
 |-M|\ *size*\ [**+m**\ *mag*]
 |-S|\ *<symbol><size>*
 [ |SYN_OPT-B| ]
-[ |-C|\ *lon*/*lat*\ [**+p**\ *pen*][**+s**\ *pointsize*] ]
 [ |-E|\ *fill* ]
 [ |-F|\ *fill* ]
 [ |-G|\ *fill* ]
@@ -58,7 +60,7 @@ Examples
 
 Use special format derived from HYPO71 output::
 
-    gmt polar -R239/240/34/35.2 -JM8c -N -Sc0.4 -D239.5/34.5 -M5 -Qh -pdf test <<END
+    gmt polar -R239/240/34/35.2 -JM8c -N -Sc0.4 -D239:30E/34:30N -M5 -Qh -pdf test <<END
     #Date Or. time stat azim ih
     910223 1 22 0481 11 147 ipu0
     910223 1 22 6185 247 120 ipd0
--- a/doc/rst/source/supplements/seis/polar_common.rst_
+++ b/doc/rst/source/supplements/seis/polar_common.rst_
@@ -9,7 +9,8 @@ will plot stations on focal mechanisms on a map.
 Parameters are expected to be in the following columns:
 
     **1**,\ **2**,\ **3**:
-        station\_code, azimuth, take-off angle
+        *station-code azimuth take-off-angle*
+        (must all be numerical values)
     **4**:
         polarity:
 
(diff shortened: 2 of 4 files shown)


PaulWessel:
Pinging @seisman for comments on the questions above.

seisman:
> 1. The description of **-M** is strange: "Sets the size of the beach ball to plot polarities in".  We are not plotting beachballs here but the lower half of the focal sphere, no?  Copy/paste issue? I can understand the **+m** modifier to scale size to magnitude for beachballs.  Does this still make sense for a focal sphere and we just update the language?

Yes, changing it to "Sets the size of the focal sphere to plot polarities in" should be good.

PaulWessel:
OK done.  For my education, is this true?: The data plotted are all observations from a single earthquake observed at various stations at different azimuths and distances.  The focal sphere is typically plotted at the location of the earthquake, given via -D (but see +z for offsetting its location).  If so, should I add something like that to the docs?

PaulWessel:
Any opinion on the size of that tiny circle? I think if default pen is 0.25p we should at least make the circle 0.5p.  Just tested that with plot and it gives a tiny donut:

`echo 0 0 | gmt plot -R-1/1/-1/1 -JX5c -Sc0.5p -W0.25p -B -pdf t`

seisman:
> OK done. For my education, is this true?: The data plotted are all observations from a single earthquake observed at various stations at different azimuths and distances. The focal sphere is typically plotted at the location of the earthquake, given via -D (but see +z for offsetting its location). If so, should I add something like that to the docs?

Yes, your understanding is correct. It's good to add these explainations.

> 2\. The default size of the small circle at the point set via **-D** when an alternate location is given is listed as 0.015i. But the code is using **GMT_DOT_SIZE** which is actually 0.005i.  So that is a factor of 3 between usage documentation and actual values, and the documentation is quiet about it.  The 0.005i is 0.36 points so I put in 0.5 points so the circle can be seen.  However, I never make these plots so I do not know what is a useful default, but being smalle than the pen thickness of 0.25p cannot be useful. **Note**: This will be the same issue for **meca** (and possibly **coupe** if that capability is added). I also note that **meca** uses the same 0.005i default diameter but says in the documentation it is 0 (presumably means no circle).  So there needs to be a common default and common language used in both usage and docs across these modules.

> Any opinion on the size of that tiny circle? I think if default pen is 0.25p we should at least make the circle 0.5p. Just tested that with plot and it gives a tiny donut:
> 
> `echo 0 0 | gmt plot -R-1/1/-1/1 -JX5c -Sc0.5p -W0.25p -B -pdf t`

I'm thinking about do we really need the line and the circle in **polar**.

We have an example for **polar** in `test/pspolar/pspolar_01.sh`. In the below figure, the left panel is plotted using **meca** and **polar**, and the right panel is plotted using **polar** only. I think we all agree that using **polar** only doesn't provide any useful information.

image

In **meca**, we have the **-C** option to shift the beachball location, add a line connecting the old and new locations, and also add a circle at the old location. The below figure shows how **-C** works. In the left panel, I need to add **-C** to both **meca** and **polar** so that the beachball and the polarity symbols can match. In the right panel, **-C** is added to **polar**, but again, without a beachball, the right panel seems useless. 

image

So I think the new syntax should be `-D<lon>/<lat>[+z<lon>/<lat>]` only. If anyone wants to connect the old and new locations, they should do it in **meca**, not in **polar**. Does it make sense to you?

seisman:
> Any opinion on the size of that tiny circle? I think if default pen is 0.25p we should at least make the circle 0.5p. Just tested that with plot and it gives a tiny donut:
> 
> `echo 0 0 | gmt plot -R-1/1/-1/1 -JX5c -Sc0.5p -W0.25p -B -pdf t`

Back to **meca**, the default circle size is too small. It seems `0.5p` is also too small (see the figure below, using the default pen and 0.5p circle):
image

PaulWessel:
Thanks @seisman, I agree it seems pointless to let **pspolar** draw that circle and line. There is likely a lot of copy/paste in these various codes and docs rom way back.  But if we did -D<lon>/<lat>[+z<lon>/<lat>], what purpose does the initial lon/lat serve?  It is simply ignored then, right, since it is no longer used in drawing line or circle.

PaulWessel:
Since we are fixing bugs anyway, I guess we can pick what the default circle should be. 0.5p is very tiny.  Perhaps 0.1c is better?

seisman:
> Thanks @seisman, I agree it seems pointless to let **pspolar** draw that circle and line. There is likely a lot of copy/paste in these various codes and docs rom way back. But if we did `-D<lon>/<lat>[+z<lon>/<lat>]`, what purpose does the initial lon/lat serve? It is simply ignored then, right, since it is no longer used in drawing line or circle. Should we just simply use `-Dlon/lat` for wherever we want to place the sphere?

Yes, that's a good point. `-Dlon/lat` is enough.

seisman:
> Since we are fixing bugs anyway, I guess we can pick what the default circle should be. 0.5p is very tiny. Perhaps 0.1c is better?

0.1c seems good on a 7-inch map.
image

PaulWessel:
Your 0.1c circle seems filled?  How did that happen?

seisman:
> Your 0.1c circle seems filled? How did that happen?

I think the color is controlled by **-G**.

PaulWessel:
Finally, with **+z** no longer needed then meca could instead use **-A** for "alternative location" of symbol, unless you think coupe may one day want to specify alternative locations.  But being projected, not clear how to me.  Anyway, the options now are

1. Use -Z in meca for what -C now does, and in the future we could use -Z in coupe for the same thing
2. Use -A in meca for what -C now does, and decide there will never be such a need in coupe.

I think you can better make that choice than I!

seisman:
I don't think it's possible to specify new locations in **coupe**. So **-A** seems good to me

PaulWessel:
Great, I will make it so in meca.  I think I did all the things we discussed in pspolar so I will remove the WIP - maybe a final check from you and we are done?

PaulWessel:
I still wonder about the 0.1c circle.  In GMT4 the default was 0 (no circle, just the line). In GMT5-6.1.1 the circle size was 0.005i which is so small that unless the user selected a faint line it would be covered by the line.  So effectively zero.  Should we just state that the circle is optional; append +s<size> to plot it - otherwise just draw the line?

PaulWessel:
If you look at test meca_input_columns.sh you can see some examples of drawing the line and there is a barely discernible bump at the end for the 0.005i circle.  I think this is not a good look.  We should not plot it by default and let user select a size if they want it.

seisman:
> Should we just state that the circle is optional; append +s to plot it - otherwise just draw the line?

Sounds good to me.

PaulWessel:
OK, done.

seisman (approved):

Issue #4821, opened 2021-02-17
Recent changes (maybe #4813) break some tests on Windows


seisman:
Recent changes (maybe #4813) break some tests on Windows:
```
The following tests FAILED:
	124 - doc/scripts/GMT_movie_progress.sh (Failed)
	234 - doc/examples/anim01/anim01.sh (Failed)
	235 - doc/examples/anim02/anim02.sh (Failed)
	236 - doc/examples/anim03/anim03.sh (Failed)
	237 - doc/examples/anim04/anim04.sh (Failed)
	238 - doc/examples/anim05/anim05.sh (Failed)
	485 - test/grdmath/daynight.sh (Failed)
	575 - test/movie/PS_layers.sh (Failed)
	576 - test/movie/movie_indicator_de.sh (Failed)
	577 - test/movie/movie_indicator_f_hor.sh (Failed)
	578 - test/movie/movie_indicator_f_ver.sh (Failed)
Errors while running CTest
	618 - test/psbasemap/JOb_pole.sh (Failed)
```

The error messages are:
```
 1/19 Test  #124: doc/scripts/GMT_movie_progress.sh .....***Failed   13.60 sec
Set GMT_SESSION_NAME = 48886
cp: cannot stat 'D:agmtgmtbuilddocscriptsGMT_movie_progressgmt.conf': No such file or directory
ERROR: D:/a/gmt/gmt/build/test/gmtest:210
memtrack errors: 0
exit status: 1
```

PaulWessel:
Hopeing someone can run these manually on a local Windows system and see what is happening.

Squadzone:
@PaulWessel 

can you take a look this
https://github.com/Squadzone/gmt/commit/e58ecd6fbc03801e93d196d1cafc098ec3f95dcc

joa-quim:
We are aware of that. When time allows.  

Sent from my iDedo

PaulWessel:
@joa-quim, what should I do with this one? The ifdef applies to the Win executables and I guess the issue is that these can be run from cmd or git bash and there should be a different behavior?

joa-quim:
I may have a clue. The error message says **cp:** ``cannot stat`` but on **cmd** it's **copy**, not **cp**. But because the script is run from bash, the **cp** command exists but does not recognize the backslash as a file separator.

PaulWessel:
[...]

PR #4871, opened 2021-02-28
Let movie and batch use input script mode to decide the OS

PaulWessel:
When running the windows-installer versions of **movie** or **batch** the user may wish to write scripts in bash or batch.  Thus, commands for moving and copying files as well as directory separators should be determined based on the script type and not just the fact it is a windows binary.  Hopefully this PR will address #4821.

Left in WIP mode for now since we need more cross-platform testing for the Windows executables.

### DIFF
--- a/src/batch.c
+++ b/src/batch.c
@@ -598,6 +598,7 @@ EXTERN_MSC int GMT_batch (void *V_API, int mode, void *args) {
 				GMT_Report (API, GMT_MSG_ERROR, "Unable to make preflight script %s executable - exiting.\n", pre_file);
 				fclose (Ctrl->In.fp);
 				Return (GMT_RUNTIME_ERROR);
+			}
 		}
 		/* Run the pre-flight now which may or may not create a <timefile> needed later via -T, as well as needed data files */
 		if (Ctrl->In.mode == GMT_DOS_MODE)	/* Needs to be "cmd /C" and not "start /B" to let it have time to finish */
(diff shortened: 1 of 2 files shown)


joa-quim:
You forgot to wrap some of these
```
error C2065: 'S_IRWXU': undeclared identifier
```

PaulWessel:
Boy, I forgot to even look for teh #ifndef branches in general...

joa-quim:
Even worse now. I have a lot more of those errors.
You don't need to set executable bit on bash files to run them on Windows.

PaulWessel:
What are the errors? An this is running a batch-script movie on Windows, right? 
OK, I agree I goofed here,  Those are not command line statements but system calls (chmod) and those should be under ifdefs.  It is just the runtime - scripting commands that we need to deal with.

joa-quim:
Compile errors
```
error C2065: 'S_IRWXU': undeclared identifier
```

PaulWessel:
Hopefully getter again.

joa-quim:
OK, it builds now. But I don't find any ``globe.bat`` or any other animation batch

PaulWessel:
https://docs.generic-mapping-tools.org/dev/movie.html#examples

joa-quim:
Batch finish correctly and sh is on its way and mine is dinner table

PaulWessel:
Great.  Dinner sounds good - just had breakfast a few hours ago...

PaulWessel:
Just a comment on your point about

`/* Needs to be "cmd /C" and not "start /B" to let it have time to finish */`

This is for the preflight script which must finish before we launch the frames.  That is why start would be a disaster (the first frame would start right away and try to use data created by the preflight but it has not finished yet...

joa-quim (approved):

PR #4876, opened 2021-03-01
Let meca take -I for intensity-based color changes

PaulWessel:
For **meca** to be used in the context of animations via events, it will need the **-I** option used to intensify the current color.  This PR implements this along the lines of **plot**.  Some simple examples. First the case of a beachball with color from a CPT but no intensity variations:

```
gmt makecpt -Cturbo -T0/8 > t.cpt
echo 0 0 6.0 10 35 129 5 0 0 Mixed | gmt meca -Sa10c -Ct.cpt -R-1/1/-1/1 -JM14c -B -L1p -png a
```

a

Now let apply a constant intensity of 0.5:

`echo 0 0 6.0 10 35 129 5 0 0 Mixed | gmt meca -Sa10c -Ct.cpt -R-1/1/-1/1 -JM14c -B -L1p -I0.5 -png b`

b

Finally, we will read the intensity from the data record following the required columns:

`echo 0 0 6.0 10 35 129 5 -0.5 0 0 Mixed | gmt meca -Sa10c -Ct.cpt -R-1/1/-1/1 -JM14c -B -L1p -I -png c`

c

The only thing I don't really like is the fact that the alternative lon lat is in the file and becomes part of the trailing text but the code checks for this hence the correct label is extracted.  However, it means the intensity must come before those two coordinates, if present.

### DIFF
--- a/doc/rst/source/supplements/seis/meca.rst
+++ b/doc/rst/source/supplements/seis/meca.rst
@@ -23,6 +23,7 @@ Synopsis
 [ |-E|\ *fill*]
 [ |-F|\ *mode*\ [*args*] ]
 [ |-G|\ *fill*]
+[ |-I|\ [*intens*] ]
 [ |-L|\ [*pen*] ]
 [ |-M| ]
 [ |-N| ]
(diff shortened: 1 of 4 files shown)


PaulWessel:
Hi @seisman, sorry to rush you but I wish to do the same to **coupe** and it would be simpler if the **meca** PR was approved and merged first. I would do it the same was as I did in **meca** so should have no side-effects for users. The plan once this is done is to add the per-record transparency mechanism that plot -t has, and finally the size scaling (somehow) so that I can try to make a movie of a focal mechanism instead of just a colored circle.

seisman (approved):
Looks good to me.

PR #4878, opened 2021-03-01
Format basemap argument modifiers as lists 

meghanrjones:

This PR implements @Esteban82's [suggestion for improving the GMT documentation](https://github.com/GenericMappingTools/gmt/discussions/4678#discussioncomment-335383) by formatting the modifiers of `basemap` arguments as lists. In addition, this PR adds cross-referencing between the arguments and formats the ReST files lines to 120 characters rather than 80 - 120+ chars based on [this discussion](https://github.com/GenericMappingTools/gmt/issues/477). 

Fixes #

**Reminders**


meghanrjones:
Below are screenshots of the current and revised documentation of **-F** and **-L** for `basemap` to save efforts in building the docs. What do you think, @Esteban82 and @joa-quim?

Current version:
image

Revised version:
image

PaulWessel:
I like that layout. Reading it I had a few further comments:

1. The roses, scales etc are referred to as map "embellishments" in the cookbook.  Assuming that is a good common word, should we replace references to map "features" in the above section with embellishment and cross-link to that cookbook section?
2. If RST allows, would it look cleaner if we did not have the leading bullets and just started with the plus symbol for modifier instead?
3. Should we recycle the figure for these embellishments and place then at the start of these sections (they are also shown in the cookbook)?

doc/rst/source/basemap_common.rst_
```diff
@@ -43,26 +41,26 @@ Optional Arguments
...
+    * **+g**\ *fill* to fill the box [default is no fill]
+    * **+c**\ *clearance* where *clearance* is either *gap*, *xgap*\ /\ *ygap*, or *lgap*\ /\ *rgap*\ /\ *bgap*\ /\
+      *tgap* where these items are uniform, separate in x- and y-direction, or individual side spacings between map
+      inset/scale/rose and border respectively.
+    * **+i** to draw a secondary, inner border as well. We use a uniform *gap* between borders of 2\ **p** and the
+      :term:`MAP_DEFAULT_PEN` unless other values are specified.
+    * **+r** to draw rounded rectangular borders instead, with a 6\ **p** corner radius. You can override this radius by
+      appending another value.
+    * **+s** to draw an offset background shaded region. Here, *dx*/*dy* indicates the shift relative to the foreground
+      frame [default is 4\ **p**/-4\ **p**] and *shade* sets the fill style to use for shading [default is gray50].
```
seisman, line 63: Just a quick comment. Personally, I prefer to use `-` (not `*`) for bullet lists, because `*` is heavily used for italic and bold. 
```suggestion
    - **+p**\ *pen* to specify a different pen width
    - **+g**\ *fill* to fill the box [default is no fill]
    - **+c**\ *clearance* where *clearance* is either *gap*, *xgap*\ /\ *ygap*, or *lgap*\ /\ *rgap*\ /\ *bgap*\ /\
      *tgap* where these items are uniform, separate in x- and y-direction, or individual side spacings between map
      inset/scale/rose and border respectively.
    - **+i** to draw a secondary, inner border as well. We use a uniform *gap* between borders of 2\ **p** and the
      :term:`MAP_DEFAULT_PEN` unless other values are specified.
    - **+r** to draw rounded rectangular borders instead, with a 6\ **p** corner radius. You can override this radius by
      appending another value.
    - **+s** to draw an offset background shaded region. Here, *dx*/*dy* indicates the shift relative to the foreground
      frame [default is 4\ **p**/-4\ **p**] and *shade* sets the fill style to use for shading [default is gray50].
```

meghanrjones:
> I like that layout. Reading it I had a few further comments:
> 
>     1. The roses, scales etc are referred to as map "embellishments" in the cookbook.  Assuming that is a good common word, should we replace references to map "features" in the above section with embellishment and cross-link to that cookbook section?
> 
Good idea!
>     2. If RST allows, would it look cleaner if we did not have the leading bullets and just started with the plus symbol for modifier instead?
This page describes the list styles that can be easily implemented using the rtd theme: https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html. Probably a table is the closest option to this idea, but maintaining tables is bit more complicated than lists. 
>     3. Should we recycle the figure for these embellishments and place then at the start of these sections (they are also shown in the cookbook)?
Figures in the arguments section would be a dramatic change in my opinion, so I think we should ping others before implementing this option. I think that figures at the start or links to figures in the cookbook would be equally fine.

seisman:
> > ```
> > 2. If RST allows, would it look cleaner if we did not have the leading bullets and just started with the plus symbol for modifier instead?
> > ```
> 
> This page describes the list styles that can be easily implemented using the rtd theme: [sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html](https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html). Probably a table is the closest option to this idea, but maintaining tables is bit more complicated than lists.

It's possible to hide the leading bullets using a pure CSS solution.
 
> > ```
> > 3. Should we recycle the figure for these embellishments and place then at the start of these sections (they are also shown in the cookbook)?
> > ```
> 
> Figures in the arguments section would be a dramatic change in my opinion, so I think we should ping others before implementing this option. I think that figures at the start or links to figures in the cookbook would be equally fine.

Perhaps it's time to add the [sphinx-togglebutton](https://sphinx-togglebutton.readthedocs.io/en/latest/) extension. I think it's pretty useful to add more figures and examples in the arguments sections, without too much distraction.

meghanrjones:
> > > ```
> > > 2. If RST allows, would it look cleaner if we did not have the leading bullets and just started with the plus symbol for modifier instead?
> > > ```
> > 
> > 
> > This page describes the list styles that can be easily implemented using the rtd theme: [sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html](https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/lists_tables.html). Probably a table is the closest option to this idea, but maintaining tables is bit more complicated than lists.
> 
> It's possible to hide the leading bullets using a pure CSS solution.

I'll try this out.

> > > ```
> > > 3. Should we recycle the figure for these embellishments and place then at the start of these sections (they are also shown in the cookbook)?
> > > ```
> > 
> > 
> > Figures in the arguments section would be a dramatic change in my opinion, so I think we should ping others before implementing this option. I think that figures at the start or links to figures in the cookbook would be equally fine.
> 
> Perhaps it's time to add the [sphinx-togglebutton](https://sphinx-togglebutton.readthedocs.io/en/latest/) extension. I think it's pretty useful to add more figures and examples in the arguments sections, without too much distraction.

Looks good to me.

seisman:
> > It's possible to hide the leading bullets using a pure CSS solution.
> 
> I'll try this out.

I can help if you don't know much about CSS. It may take me a few minutes to make it work.

joa-quim:
The small images in the ``plot`` man page look very nice. As long as they are small, wide and short, like those in ``plot`` and the the compass and map scale in the CookBook, I think they would be helpful to have next to the **-F** & **-L** description.

meghanrjones:
> > > It's possible to hide the leading bullets using a pure CSS solution.
> > 
> > 
> > I'll try this out.
> 
> I can help if you don't know much about CSS. It may take me a few minutes to make it work.

I understand how to include a list without bullets using CSS. However, it's a bit messy because the bold and italics ReST styles do not work, such that the modifiers would need to be bolded using CSS properties. My opinion is that it's worth keeping the bullets in order to keep the same font property formatting and allow cross referencing to other arguments, unless you know of a way to combine ReST and raw html functionality.

seisman:
> I understand how to include a list without bullets using CSS. However, it's a bit messy because the bold and italics ReST styles do not work, such that the modifiers would need to be bolded using CSS properties. My opinion is that it's worth keeping the bullets in order to keep the same font property formatting and allow cross referencing to other arguments, unless you know of a way to combine ReST and raw html functionality.

It's unclear to me what your solution is.

What I meant is adding the following lines to `doc/rst/_static/style.css`:
```css
.rst-content .section #optional-arguments ul li {
    list-style: none;
}
```
It will look like:
image

meghanrjones:
> > I understand how to include a list without bullets using CSS. However, it's a bit messy because the bold and italics ReST styles do not work, such that the modifiers would need to be bolded using CSS properties. My opinion is that it's worth keeping the bullets in order to keep the same font property formatting and allow cross referencing to other arguments, unless you know of a way to combine ReST and raw html functionality.
> 
> It's unclear to me what your solution is.
> 
> What I meant is adding the following lines to `doc/rst/_static/style.css`:
> 
> ```css
> .rst-content .section #optional-arguments ul li {
>     list-style: none;
> }
> ```
> 

Yes, this is much better. I was clearly trying a wrong way. Thank you for the help. Here's the section without the bullets. @PaulWessel, do you think this is cleaner?

image

seisman:
If we decide to use the new lists without bullets, we may also need to adjust the vertical spacing between items, the left side margin, and even the first-line text-indent.

meghanrjones:
@seisman, do you know if there is a way to change the formatting for spacing between open parentheses and links? For example, there appears to be a gap here between the parentheses and the **-D** and **-L** links:

> image

seisman:
> @seisman, do you know if there is a way to change the formatting for spacing between open parentheses and links? For example, there appears to be a gap here between the parentheses and the **-D** and **-L** links:
> 
> > image

Check this file https://github.com/GenericMappingTools/gmt/blob/master/doc/rst/source/common_SYN_OPTs.rst_.

You need to change 
```
.. |-D| replace:: |h-D| **-D**\ |h-x|
```
to
```
.. |-D| replace:: |h-D|\ **-D**\ |h-x|
```

It works but I don't check carefully if there are any other side-effects.

Esteban82:
I just read everything. Very nice @meghanrjones. I agree with joaquim idea to include small, wide and short figures.

meghanrjones:
All the ReST formatting is complete. I opened a separate issue in #4913 for the images. Regarding how to format the lists (bullet points, indentation, etc.), I think it would be best to merge this and open a separate PR to discuss changing the CSS formatting for lists in the argument section. Comparing the formatting differences will be easier using Vercel preview if the basic structure for the lists are already included in the documentation.

PaulWessel:
Looks slick!  One issue, under **-Lg** there is mention of modifiers **+o**_dx_ but the optional _dy_ is missing, making it look a bit odd?

meghanrjones:
> Looks slick! One issue, under **-Lg** there is mention of modifiers **+o**_dx_ but the optional _dy_ is missing, making it look a bit odd?

Should be better now, if I understood your comment correctly.

PaulWessel:
Yep, perfect.

PaulWessel (approved):

PR #4886, opened 2021-03-02
Ensure degenerated Cartesian ellipses get correct azimuth

### LINKED ISSUE #4877: psxy -SE bug when using linear projection
I have been having some issues with using the -SE option with psxy and linear projections. I've specifically been having issues with ellipses, but I think the problem can be best described with this example:

echo "0 0 0.1" | gmt psxy -JX10c/10c -R-0.1/0.1/-1/1 -SE- -Gred -Wthinner,black -Bxa0.05 -Bya0.5 -BSWne > tmp.ps
gmt psconvert -Tf -Z -P -A tmp.ps

This should plot a circle with a diameter of 0.1 units. As you can see from the attached plot, this command isn't working how I would expect. It seems the x and y direction are flipped (i.e. in this case the shape should be rotated by 90 deg).

[tmp.pdf](https://github.com/GenericMappingTools/gmt/files/6062185/tmp.pdf)

**System information**

* Operating system: ubuntu

PaulWessel:
Depending on projection, needed to set the degenerate ellipse azimuth correctly since for Cartesian axes with different scalings it may plot as an ellipse, not a circle, so direction matters even if it is a "circle".

Also, pen, fill was not activated for this particular case, leading to an invisible ellipse (which seems broken since 6.1.1).  With this PR, the following command works:

`echo "0 0 0.1" | gmt plot -JX10c/10c -R-0.1/0.1/-1/1 -SE- -Gred -Wthinner,black -B -BSWne -png t`

and gives the expected thin ellipse stretched in the x-direction given the different ranges:

t

Closes #4877.

### DIFF
--- a/src/psxy.c
+++ b/src/psxy.c
@@ -1571,9 +1571,9 @@ EXTERN_MSC int GMT_psxy (void *V_API, int mode, void *args) {
 
 			if ((S.symbol == PSL_ELLIPSE || S.symbol == PSL_ROTRECT) && !S.par_set) {	/* Ellipses and rectangles */
 				if (S.n_required == 0)	/* Degenerate ellipse or rectangle, got diameter via S.size_x */
-					in[ex2] = in[ex3] = S.size_x, in[ex1] = 0.0;	/* Duplicate diameter as major and minor axes and set azimuth to zero  */
+					in[ex2] = in[ex3] = S.size_x, in[ex1] = (gmt_M_is_cartesian (GMT, GMT_IN)) ? 90.0 : 0.0;	/* Duplicate diameter as major and minor axes and set azimuth to zero  */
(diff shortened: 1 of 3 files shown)


meghanrjones:
Is it worthwhile to add a test for this?

PaulWessel:
Yes, since this case was not covered.  Maybe you can make one that does 2x2 panels of degenerate ellipses for Cartesian projections with different scalings, including equal, to just cover the bases.

meghanrjones:
> Yes, since this case was not covered. Maybe you can make one that does 2x2 panels of degenerate ellipses for Cartesian projections with different scalings, including equal, to just cover the bases.

OK, I will push a new test directly to this branch rather than creating a new PR off it for simplicity.

PaulWessel:
Yep, that is good.

PR #4887, opened 2021-03-02
Ensure automatic y-range works for log counts and log projections

PaulWessel:
See discussion on the [forum](https://forum.generic-mapping-tools.org/t/histogram-with-y-range-not-specified/1384/8).  The **-R**w/e/0/0 option set to determine the y-range automatically cannot become **-R**w/e/0/n when we are doing logarithmic cases.  In those cases it needs to be 1/n.

### DIFF
--- a/src/pshistogram.c
+++ b/src/pshistogram.c
@@ -137,6 +137,9 @@ struct PSHISTOGRAM_INFO {	/* Control structure for pshistogram */
 	struct GMT_ARRAY *T;
 };
 
+#define LOG10_2 0.301029995664
+#define LOG10_5 0.698970004336
+
 static void *New_Ctrl (struct GMT_CTRL *GMT) {	/* Allocate and initialize a new control structure */
 	unsigned int k;
 	struct PSHISTOGRAM_CTRL *C = NULL;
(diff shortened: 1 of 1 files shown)


Esteban82 (approved):

PaulWessel:
I changed it to WIP. Although the latest log10 checks will prevent a crash due to a bad w/e/s/n, the automatic w/e/s/n algorithm assumes linear projections.  If log10 then we should do something better, so will work on that later today.

PaulWessel:
OK, this revision handles logarithmic axes as well as -Z settings involving log and tries to come up with better ranges while having safety valves against getting a log10 (0).  No change in testing.

meghanrjones (commented):
The y-range minimum at 1 is problematic in some instances for the log cases of **-Z**. For example, much of the data is cut off in this plot:
```
gmt begin Histo png
gmt histogram @v3206_06.txt -T100 -Jx0.002c/1c -Bxaf -By1 -Gblack -Z3 -R-6000/0/0/0
gmt end show
```

PaulWessel:
Thanks, went overboard here.  It should be 0 not 1 since we do log (1+x) in the bins.  Now that I have added a check for **-J** being log the else case does not have to worry about this any more.

meghanrjones (approved):

PR #4890, opened 2021-03-03
Explore separate MAP_ANNOT_OBLIQUE settings for oblique and other projections

PaulWessel:
This PR to be merged into gmt-themes sets the new default for **MAP_ANNOT_OBLIQUE** to be **lon_horizontal,lat_horizontal,tick_extend** but we then proceed to reset this to the previous default **anywhere** if the projection/region is not oblique, _unless_ the user manually changed the setting.

With this scheme I get happy results for things like

```
gmt basemap -R-200/200/-200/200+uk -JOa8/88/46/15c -T8/87/3c -Bxa5fg -Bya1fg -BNSWE -png a
gmt basemap -R0/20/0/20 -JM6i -B -png b
gmt basemap -R-180/-20/0/90 -JPoly/10c -Bx30g10 -By10g10 -png c
gmt basemap -Rg -JH0/15c -Ba30fg -BWSNE -png d
```

See what you think, @meghanrjones.  Note I also upped the **ANNOT_MIN_SPACING** default to 28p (~1 cm) since the annotations for plot **a** are horizontal and it needs more space to not overprint widely.  Also note that my caveat about **subplot** seems to have been premature.  To place latitudes for rectilinear projections parallel to the border we require **-SR**...**+p** anyway.  That seems to work still with this PR:

```
gmt begin t pdf
  gmt subplot begin 2x2 -Fs3i -M5p -A -SCb -SRl+p -Bwstr -R-11/11/-11/11 -JM3i
    gmt basemap
    gmt basemap -c
    gmt basemap -c
    gmt basemap -c
  gmt subplot end
gmt end show
```

Apart from setting the new defaults in gmt_init.c, the reset takes place in _gmtmap_reset_oblique_settings_ in gmt_map.c.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -10388,7 +10388,7 @@ unsigned int gmtlib_setparameter (struct GMT_CTRL *GMT, const char *keyword, cha
 			ival = gmtinit_parse_map_annot_oblique (GMT, value);
 			if (ival >= GMT_OBL_ANNOT_LON_X_LAT_Y && ival < GMT_OBL_ANNOT_FLAG_LIMIT)  {
 				GMT->current.setting.map_annot_oblique = ival;
-				GMT->current.setting.map_annot_oblique_set = 1;
+				GMT->current.setting.map_annot_oblique_set = true;
 			}
 			else
 				error = true;
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -6552,6 +6552,7 @@ void gmt_conf_SI (struct GMT_CTRL *GMT) {
 		gmtinit_conf_modern (GMT);
 	else
 		gmtinit_conf_classic (GMT);
+	//GMT->current.setting.map_annot_oblique_set = false;
 }
 
 /*! . */
@@ -18734,7 +18735,7 @@ int gmt_manage_workflow (struct GMTAPI_CTRL *API, unsigned int mode, char *text)
 			if (error) return (error);		/* Bail at this point */
 			gmt_reset_history (API->GMT);	/* No old classic history shall affect a new modern mode session */
 
-			gmt_conf_SI (API->GMT);				/* Get the original system defaults */
+			gmt_conf_SI (API->GMT);			/* Get the original system defaults */
 			if (!clean_start) {
 				/*  Overload any user-supplied defaults via a gmt.conf file but reset PS_MEDIA to the original system default */
 				if (gmt_getdefaults (API->GMT, NULL) == GMT_NOERROR)	/* Ingested a gmt.conf file */
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -6099,7 +6099,8 @@ GMT_LOCAL void gmtinit_conf_classic (struct GMT_CTRL *GMT) {
 	GMT->current.setting.given_unit[GMTCASE_MAP_ANNOT_OFFSET_PRIMARY] = 'p';
 	GMT->current.setting.given_unit[GMTCASE_MAP_ANNOT_OFFSET_SECONDARY] = 'p';
 	/* MAP_ANNOT_OBLIQUE */
-	GMT->current.setting.map_annot_oblique = GMT_OBL_ANNOT_ANYWHERE;
+	//GMT->current.setting.map_annot_oblique = GMT_OBL_ANNOT_ANYWHERE;
+	GMT->current.setting.map_annot_oblique = GMT_OBL_ANNOT_LON_HORIZONTAL | GMT_OBL_ANNOT_LAT_HORIZONTAL | GMT_OBL_ANNOT_EXTEND_TICKS;
 	/* MAP_ANNOT_MIN_ANGLE */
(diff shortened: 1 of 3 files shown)


meghanrjones (commented):
The **-C** option for begin does not work with this implementation for MAP_ANNOT_OBLIQUE (it still works for other parameters). For example this produces a plot with lon_horizontal,lat_horizontal for MAP_ANNOT_OBLIQUE:
```
gmt set MAP_ANNOT_OBLIQUE=lon_horizontal,lat_horizontal
gmt begin test ps -C
        gmt coast -R0/360/-80/80 -JT180/-30/6i -Glightgray -A1000 -Ba60g60
gmt end show
```

meghanrjones:
Increasing **MAP_ANNOT_MIN_SPACING** brings back [this problem](https://github.com/GenericMappingTools/gmt/pull/3344#issuecomment-753524842) for secondary and primary axes:
```
gmt coast -W -RDE+R10m -Bpa30m+f -Bsa1d -JL10.5/51/49/53/17c -N1/1p,darkred -Saliceblue -png mapDEmodern --GMT_THEME=modern
```

PaulWessel:
Right.  So clearly, if users insist on specifying their own very small **-B** intervals then this can happen anywhere for any nonzero default min spacing.  If we shrink it from 1cm to 24p then it is OK, but then we get a bit of overlap in the **-Joa** plot.

In any case, it seems we should report to the user when we skip an annotation due to crowdedness and remind them they can control this via that setting.  I don't think we do any reporting now but being told 5 annotations were suppressed due to a proximity to other annotations set by spacing = xxxp would be helpful I think.

meghanrjones:
> Right. So clearly, if users insist on specifying their own very small **-B** intervals then this can happen anywhere for any nonzero default min spacing. If we shrink it from 1cm to 24p then it is OK, but then we get a bit of overlap in the **-Joa** plot.
> 
> In any case, it seems we should report to the user when we skip an annotation due to crowdedness and remind them they can control this via that setting. I don't think we do any reporting now but being told 5 annotations were suppressed due to a proximity to other annotations set by spacing = xxxp would be helpful I think.

Yes, I agree that the main problem is when the user does not know why the annotations are not plotted. A warning is a good solution.

PaulWessel:
I should probably add that in master since I do not now where this particular PR will go in the end.

PaulWessel:
> The **-C** option for begin does not work with this implementation for MAP_ANNOT_OBLIQUE (it still works for other parameters). For example this produces a plot with lon_horizontal,lat_horizontal for MAP_ANNOT_OBLIQUE:
> 
> ```
> gmt set MAP_ANNOT_OBLIQUE=lon_horizontal,lat_horizontal
> gmt begin test ps -C
>         gmt coast -R0/360/-80/80 -JT180/-30/6i -Glightgray -A1000 -Ba60g60
> gmt end show
> ```

I guess I am confused now.  This PR sets the new default for  **MAP_ANNOT_OBLIQUE**=lon_horizontal,lat_horizontal,tick_extend.  If you then set MAP_ANNOT_OBLIQUE=lon_horizontal,lat_horizontal then the only difference would be the ticks?  I am looking at the plot and do not now what the problem is...

meghanrjones:
> > The **-C** option for begin does not work with this implementation for MAP_ANNOT_OBLIQUE (it still works for other parameters). For example this produces a plot with lon_horizontal,lat_horizontal for MAP_ANNOT_OBLIQUE:
> > ```
> > gmt set MAP_ANNOT_OBLIQUE=lon_horizontal,lat_horizontal
> > gmt begin test ps -C
> >         gmt coast -R0/360/-80/80 -JT180/-30/6i -Glightgray -A1000 -Ba60g60
> > gmt end show
> > ```
> 
> I guess I am confused now. This PR sets the new default for **MAP_ANNOT_OBLIQUE**=lon_horizontal,lat_horizontal,tick_extend. If you then set MAP_ANNOT_OBLIQUE=lon_horizontal,lat_horizontal then the only difference would be the ticks? I am looking at the plot and do not now what the problem is...

You are right, sorry, I mixed up the new settings.

src/gmt_init.c
```diff
@@ -6099,7 +6099,8 @@ GMT_LOCAL void gmtinit_conf_classic (struct GMT_CTRL *GMT) {
 	GMT->current.setting.given_unit[GMTCASE_MAP_ANNOT_OFFSET_PRIMARY] = 'p';
 	GMT->current.setting.given_unit[GMTCASE_MAP_ANNOT_OFFSET_SECONDARY] = 'p';
 	/* MAP_ANNOT_OBLIQUE */
-	GMT->current.setting.map_annot_oblique = GMT_OBL_ANNOT_ANYWHERE;
+	//GMT->current.setting.map_annot_oblique = GMT_OBL_ANNOT_ANYWHERE;
```
meghanrjones, line 6102: Why comment rather than remove?

meghanrjones (approved):

src/gmt_init.c
PaulWessel: Forgot, will fix.

Issue #4903, opened 2021-03-04
CPT files with labels should optionally apply to the interval


PaulWessel:
**Description of the desired feature**

This issue derives partly from this [post](https://forum.generic-mapping-tools.org/t/colorbar-and-oblique-annotation/1424/6) on the forum.  Per our [documentation](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#regular-cpts), the CPT may have a flag **L**|**U**|**B** to indicate that the optional label should be placed on the lower, upper, or both ends of the color slice.  However, if the color slice represents an interval (e.g., a day, year, geologic period) then one would actually want the label to be placed at the center of the slice interval.  It would seem the only way to accomplish this is to allow the flag **I** for Interval.  Then, tools like **colorbar** can know what to do with custom annotations for unequal intervals.  The closest thing we have to doing this is **colorbar -Li**, which does use the label as an interval annotation, but it also turns all color intervals to equal lengths, which is not what we want here.  As an example of a slice in such as CPT file:

`66.0    lightgreen 72.1    lightgreen I ; Maastrichtian`

since that is the name of a geologic interval.

An alternative to the flag would be to instead enhance **-L** to take to take a modifier to **not** make the intervals equal size.  Thoughts, @GenericMappingTools/core ?

WalterHFSmith:
Could you extend the options LUB to include LUBC, where C means the label goes at the center? 
Would it need a text string after it, since it might not be a numerical value?

e.g, 65 r g b 251 r g b C Mesozoic

PaulWessel:
It could be C, but elsewhere in GMT we use I for interval (e.g., for custom annotations) so I think I should stick with the same letter.  The label is detected via the ; indicator if there is one.

PR #4914, opened 2021-03-05
Let colorbar -S+a<angle> place slanted annotations when -B is not used

PaulWessel:

See this forum [post](https://forum.generic-mapping-tools.org/t/colorbar-and-oblique-annotation/1424/6) for some background information.

**colorbar** has its own annotation machinery for the cases when equidistant annotations (as supplied via **-B**) is not desired.  Due to crowding, there are times it is best to place the annotations at an angle.  This PR implements **-S+a**_angle_ and adds a test that shows how it looks for all possible combinations of horizontal/vertical, positive and negative angles, and flipping the side of the annotations:

slant_annot

There is a related issue of how/where to place annotations if they are _interval_ annotations but this requires further decisions on CPT format, i.e., #4903.


doc/rst/source/colorbar_common.rst_
```diff
@@ -182,8 +182,9 @@ Optional Arguments
 
 .. _-S:
 
-**-S**\ [**+c**\|\ **n**\ ][**+s**]
+**-S**\ [**+a**\ *angle*][**+c**\|\ **n**\ ][**+s**]
     Control various aspects of color bar appearance when **-B** is *not* used.
+    Append **+a** to place annotations at the given *angle* [no slanting].
```
meghanrjones, line 187:
```suggestion
    Append **+a** to place annotations at the given *angle* [default is no slanting].
```

meghanrjones (approved):
Nice! It would be eventually nice to have a **+j** option for centering the annotations on the ticks or **+o** to have long annotations closer to the ticks.

PaulWessel:
> Nice! It would be eventually nice to have a **+j** option for centering the annotations on the ticks or **+o** to have long annotations closer to the ticks.

Yes, TBD how to implement that part.  You can move things closer to ticks by reducing **MAP_ANNOT_OFFSET** - is that what you mean?

PR #4918, opened 2021-03-06
WIP: Simplify documentation for common options

### LINKED ISSUE #4917: Use a list to improve readability of the -R documentation

There are many ways to specify a region, from w/e/s/n via country codes to grid files.  However, in the RST docs the option is all one long paragraph.  I suggest we split it into bulleted sections like the recent -F-box upgrade (#4915).

meghanrjones:

This (WIP) PR organizes the documentation for -R to increase readability. It also consolidates some information from the [common option page](https://docs.generic-mapping-tools.org/dev/std-opts.html) and the [cookbook common options](https://docs.generic-mapping-tools.org/dev/cookbook/options.html). 

Fixes #4917 

**Reminders**

### DIFF
--- a/doc/rst/source/cookbook/options.rst
+++ b/doc/rst/source/cookbook/options.rst
@@ -76,49 +76,7 @@ importance (some are used a lot more than others).
 Data domain or map region: The **-R** option
 --------------------------------------------
 
-The **-R** option defines the map region or data domain of interest. It
-may be specified in one of five ways, two of which are shown in Figure
-:ref:`Map region <gmt_region>`:
-
-#. **-R**\ *xmin*/*xmax*/*ymin*/*ymax*. This is the standard way to
-   specify Cartesian data domains and geographical regions when using
-   map projections where meridians and parallels are rectilinear.
-
-#. **-R**\ *xlleft*/*ylleft*/*xuright*/*yuright*\ **+r**. This form is
-   used with map projections that are oblique, making meridians and
-   parallels poor choices for map boundaries. Here, we instead specify
-   the lower left corner and upper right corner geographic coordinates,
-   followed by the modifier **+r**. This form guarantees a rectangular map
-   even though lines of equal longitude and latitude are not straight lines.
-
-#. **-R**\ *gridfile*. This will copy the domain settings found for the
-   grid in specified file. Note that depending on the nature of the
-   calling program, this mechanism will also set grid spacing and
-   possibly the grid registration (see
-   Section `Grid registration: The -r option`_).
-
-#. **-R**\ *code1,code2,...*\ [**+e**\|\ **r**\|\ **R**\ [*incs*]]. This indirectly supplies
-   the region by consulting the DCW (Digital Chart of the World) database and derives
-   the bounding regions for one or more countries given by the codes.
-   Simply append one or more comma-separated countries using the two-character
-   ISO 3166-1 alpha-2 convention (e.g., https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2).
-   To select a state within a country (if available), append .state, e.g, US.TX for Texas.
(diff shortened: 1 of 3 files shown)


meghanrjones:
There's a lot of redundancy in the text of the various doc/rst/source/explain_-\*\_full.rst_ files and doc/rst/source/cookbook/options.rst. For example, I only noticed a couple small differences for **-B** and much of the information for **-R** was repeated (although organized differently). I think it would be better to consolidate these in the doc/rst/source/explain_-\*\_full.rst_ files and then include those files in the cookbook before going on to provide more examples and images. This is a prototype of that organization for the **-R** option. What do you think of this structure @PaulWessel? Is there a reason to keep the cookbook completely isolated from the other source ReST files? (Note: I haven't completed the documentation in explain_-R_full.rst yet; just wanted to discuss the include structure first).

seisman:
Just a quick comment, `explain_-*_full.rst_` are only included in the gmt manpage (`gmt.rst`). They're useful when we provided documentation in the manpage format in GMT 5.

Now, we no longer provide manpages. Perhaps we should just directly link to the cookbook?

meghanrjones:
> Just a quick comment, `explain_-*_full.rst_` are only included in the gmt manpage (`gmt.rst`). They're useful when we provided documentation in the manpage format in GMT 5.
> 
> Now, we no longer provide manpages. Perhaps we should just directly link to the cookbook?

Yes, I think this would be even better for maintenance and would help reduce the knowledge gap between users who know how essential the cookbook is and users who do not (yet).

joa-quim:
I don't think it's a good idea to link **-R** to the CookBook. The CookBook is a heavy page and we should no assume that all users have beefy computers and very fast internet.

meghanrjones:
> I don't think it's a good idea to link **-R** to the CookBook. The CookBook is a heavy page and we should no assume that all users have beefy computers and very fast internet.

But anyone who is concerned about internet usage can use `gmt docs`. It is also possible to block the images from loading if `gmt docs` for some reason doesn't work, which is the main reason for the cookbook being larger. When I build the docs, gmt.html is 123 KB and cookbook/options.html is 147 KB, so the base versions aren't so different.

joa-quim:
Yes, the CookBook page is heavy because of the images. Imagining a trick to make it faster by not loading them seems unreasonable to me. I believe people use internet dos pages more then ``gmt docs`` (and would bet that many don't even know about it).

PaulWessel:
> we should no assume that all users have beefy computers and very fast internet.

We know that not all users have beefy computers and very fast internet.  However, GMT is funded by the NSF to provide scientific software to US scientists, all of whom do have as beefy computers and fas ast internet as they decide to have.  We cannot let lowest-common-demoninator thinking affect how we design and build GMT.  We are building it for people like ourselves, not for folks with low RAM, dial-up, and rubber bands.  And I am sure @joa-quim could have written this paragraph for me if I had asked him to!

joa-quim:
I would understand this if we had any need/gain to do it. Otherwise, I don't.

meghanrjones:
> I would understand this if we had any need/gain to do it. Otherwise, I don't.

The need is that there are hundreds of lines (at least) of redundant documentation that should not need to be updated independently any time a new feature gets added. While the solution in this PR is an option, it is more convoluted than directly linking to the common options descriptions in the cookbook.

seisman:
> I would understand this if we had any need/gain to do it. Otherwise, I don't.

It also means fewer source files in the `doc/rst/source` directory and may make it easier for new contributors to make changes to documentations.

meghanrjones:
I'm closing this PR and will try out the different ways for consolidating the documentation. First, it is necessary to make sure that there isn't information included in the gmt manpage but not in the cookbook (e.g., #4922 for the -R option).

PR #4920, opened 2021-03-06
Allow +a<angle> for y-axis as well as x-axis

### LINKED ISSUE #4575: -By +angle
Sorry for the failed pull request. I'm not very familiar with Github yet... At any rate :

>For the y- and z-axes, arbitrary angles are not allowed but +an and +ap specify annotations normal [Default] and parallel to the axis.

## Somehow related to this :
I have a plot with Y calendar axis going down (January at the top, December at the bottom). I wanted to add a secondary Y-axis with the "year", but it takes to much space horizontally and the `-Bys+ap` gives the "year" written from bottom to top (opposite direction / you tilt your head to the left to read it).
With an arbitrary angle, one could easily do `-Bys+a270` to have it correctly written... 

Is it possible to add the feature ?
Thanks

PaulWessel:

Until now, setting annotations at arbitrary angles via **-B**...**+a**_angle_ has been a feature restricted to the x-axis.  This PR opens that up to the y-axis (and by extension - but not tested - the z-axis].  All tests still pass and a new test using the arbitrary angle on the y-axis looks fine:

xyslant

Closes #4575.

### DIFF
--- a/doc/rst/source/cookbook/options.rst
+++ b/doc/rst/source/cookbook/options.rst
@@ -359,13 +359,12 @@ but you may also split this into two separate invocations for clarity, i.e.,
     default setting :term:`FORMAT_GEO_MAP`). However, for other plots you can add
     specific units by adding **+u**\ *unit*.  If any of these text strings contain
     spaces or special characters you will need to enclose them in quotes.
-    Cartesian x-axes also allow for the optional **+a**\ *angle*, which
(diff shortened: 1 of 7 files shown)


meghanrjones (approved):
Nice! The output from some scripts might change due to this PR, similar to the revisions necessary for the slanted.sh test, but it seems a worthwhile break for allowing custom y-axis annotation angles.

PR #4922, opened 2021-03-07
Update cookbook -R docs

meghanrjones:

This PR updates the documentation for [-R in the cookbook](https://docs.generic-mapping-tools.org/dev/cookbook/options.html#option-r) to ensure that all the information from [-R in the gmt module docs](https://docs.generic-mapping-tools.org/dev/gmt.html#r-full) is included. Specifically, information about -Rxmin/xmax/ymin/ymax/zmin/zmax was missing. It also adds **-Rg** and **-Rd** to the list of methods and formats the modifiers for the **-R***code* method as a list. Also missing was the note about guessing the units for UTM, TM, and Stereographic projections.

Fixes #

**Reminders**


meghanrjones:
While updating the -R cookbook documentation, I have formatted the lines to be 120 characters rather than 80 chars based on the conversation in [issue 477](https://github.com/GenericMappingTools/gmt/issues/477) and the general inconvenience of 80 char lines. However, the topic of [that issue](https://github.com/GenericMappingTools/gmt/issues/477) was code, rather than documentation, so please comment if it is preferable to keep the char length shorter in ReST files or if the line length changes make reviewing too difficult - in which case I will not format to 120 characters for the other options.

meghanrjones:
This is the link to the Vercel preview for this section: https://gmt-3te6k52ic-gmt.vercel.app/cookbook/options.html#option-r
And the current version: https://docs.generic-mapping-tools.org/dev/cookbook/options.html#option-r

meghanrjones:
@PaulWessel, this preview shows the latest version for -R with your recommendations applied: https://gmt-op280cgbm-gmt.vercel.app/cookbook/options.html#data-domain-or-map-region-the-r-option

PaulWessel:
Thanks, that is better. What do you think of these thoughts:

1. I am having second-thought about the **J** given that we have -J and I suppose it could be confusing.  Maybe **C** for code is better?
2. The radian section has a mix of pi and the math pi.  Should we stick with math pi in the text
3. The sentence Add W, E, S, or N to the coordinate limits (e.g., -R0/360/-90/90N): Give we have -Rg I think better to show an example like -R15W/30E/10S/15N instead,

meghanrjones:
> 
>     1. I am having second-thought about the **J** given that we have -J and I suppose it could be confusing.  Maybe **C** for code is better?

**C** would work, but my preference would be *justify* because that is used most other places in the docs for these two character combos. I could just add an example or two to clarify the difference between *justify* and *x0*.
>     2. The radian section has a mix of pi and the math pi.  Should we stick with math pi in the text

The issue is that the user will need to type 'pi', so using the symbol might be confusing. For example:
```gmt psbasemap -R-12pi/12pi/0/100 -JX6i/7i -Bxa4pi -Byaf -png test```
So, I could either make them all 'pi' or check that it's only pi when providing an example that can be given verbatim to gmt.

>     3. The sentence Add W, E, S, or N to the coordinate limits (e.g., -R0/360/-90/90N): Give we have -Rg I think better to show an example like -R15W/30E/10S/15N instead.

Yes, done.

meghanrjones:
I made some revisions to the radians section and the 5th **-R** option that will hopefully make things clearer.

PaulWessel (approved):

PR #4925, opened 2021-03-07
Update cookbook -B docs

meghanrjones:

This PR formats the [cookbook **-B** description](https://docs.generic-mapping-tools.org/dev/cookbook/options.html#option-b) using lists, rather than simply paragraphs, and adds some information that was included in the [gmt manpage](https://docs.generic-mapping-tools.org/dev/gmt.html#b-full) but not the cookbook section. 

**Preview here:** https://gmt-q7oppih8n-gmt.vercel.app/cookbook/options.html#option-b

Here is an example of the updated formatting for the frame settings:

image

Fixes #

**Reminders**


meghanrjones:
This is the link to the Vercel preview for this section: https://gmt-pdv5f3age-gmt.vercel.app/cookbook/options.html#option-b
This is the current version of this section: https://docs.generic-mapping-tools.org/dev/cookbook/options.html#map-frame-and-axes-annotations-the-b-option

The only section from the GMT manpage that I did not add to the cookbook is the very long list of GMT parameters that can affect the appearance of map boundaries.

PaulWessel:
Looks good! Some comments:

1. Under Calendar time coordinates there is a malformed italics section showing `*yyyy*`.
2. Same under radians.
3. For item 6 in -R (the code x0 etc, it always looks add to have the run-on codex0.  Might it be better to use a bold **J** instead, and explain **J** is a 2-char combo... or us a two char code **XX**?  This mechanism is something Kurt Feigl was very keen on as he uses that a lot in other software...
4. The general G as in -R0/360G/-90/90N took me by surprised.  I suspect this is something Remko added a long time ago.  Given there is also -Rgw/e/s/n, is this really necessary?  Also, at least the example should not be 0/360/-90/90 since we have -Rg for that - pick a subset.

meghanrjones:
> Looks good! Some comments:
> 
>     1. Under Calendar time coordinates there is a malformed italics section showing `*yyyy*`.
> 
>     2. Same under radians.
> 
>     3. For item 6 in -R (the code x0 etc, it always looks add to have the run-on codex0.  Might it be better to use a bold **J** instead, and explain **J** is a 2-char combo... or us a two char code **XX**?  This mechanism is something Kurt Feigl was very keen on as he uses that a lot in other software...
> 
>     4. The general G as in -R0/360G/-90/90N took me by surprised.  I suspect this is something Remko added a long time ago.  Given there is also -Rgw/e/s/n, is this really necessary?  Also, at least the example should not be 0/360/-90/90 since we have -Rg for that - pick a subset.

Thanks, I will address all these things. Also, to clarify this PR updates the -B option and #4922 updates the -R option. I can do the rest in a single PR if that's preferable, but thought it would be better to do the big ones (-R, -J, and -B) individually in case I get sidetracked in the process.

I'm not sure where you are referred to for point 4. I see **-Rg** but not **-RG**, can you provide a snippet to search for?

PaulWessel:
It is the -R0/360G/-90/90N listed at the end of the the "Geographic coordinates:" section

PaulWessel:
I am thinking we leave whatever code makes this work, but drop the docs as I think it is a useless format.

meghanrjones:
@PaulWessel, I'm not requesting an urgent review, but just want to make sure you are aware of this PR in case it got lost in the WIP PRs.

PaulWessel (approved):
Sorry, it looks good to me but the vercel preview shows a
<<<<<<< HEAD
in the middle of the radian stuff in -R.  Possibly a line that did not get removed? or vercel got confused?

meghanrjones:
> Sorry, it looks good to me but the vercel preview shows a
> <<<<<<< HEAD
> in the middle of the radian stuff in -R. Possibly a line that did not get removed? or vercel got confused?

Thanks, I must have missed a merge conflict. Fixed now.

PR #4926, opened 2021-03-07
Add the sphinx-togglebutton extension and a figure for MAP_FRAME_TYPES

seisman:
**Description of proposed changes**

This PR adds the [`sphinx-togglebutton`](https://sphinx-togglebutton.readthedocs.io/en/latest/) extension (~20 KB) to the repository, and also an example figure for MAP_FRAME_TYPE showing how the togglebutton works.

**Preview link:** https://gmt-p0akp8sm5-gmt.vercel.app/gmt.conf.html#term-MAP_FRAME_TYPE

It shows a toggle button on the right:

image

Click the button, and you will see:

image

Fixes #

**Reminders**

- [ ] Make sure that your code follows our style. Use the other functions/files as a basis.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Describe changes to function behavior and arguments in a comment below the function declaration.
- [ ] If adding new functionality, add a detailed description to the documentation and/or an example.

PR #4927, opened 2021-03-08
Add dropdowns showing the code for cookbook figures

meghanrjones:

This PR adds dropdown panels (using sphinx-panels) that show the code used to generate the figures in the 'features' section of the cookbook. Relates to #4926, which shows a different sphinx extension for generating dropdown panels. Also relates to the documentation discussion initiated by @KristofKoch [here](https://github.com/GenericMappingTools/gmt/discussions/4678#discussioncomment-312265).

**Preview link**: https://gmt-1r6r2gxux-gmt.vercel.app/cookbook/features.html#overview-and-the-gmt-conf-file

It shows a toggle button:

image

which can be clicked on to show the code:

image

Fixes #

**Reminders**

### DIFF
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -232,6 +232,10 @@ E.g., if your ``gmt.conf`` file has *x* offset = 3\ **c** as default, the
 
 .. _gmt_defaults_a:
 
+.. dropdown:: Click to see source code for figure below.
+
+   .. literalinclude:: /_verbatim/GMT_Defaults_1a.txt
+
 .. figure:: /_images/GMT_Defaults_1a.*
    :width: 500 px
    :align: center
@@ -240,6 +244,10 @@ E.g., if your ``gmt.conf`` file has *x* offset = 3\ **c** as default, the
 
 .. _gmt_defaults_b:
 
+.. dropdown:: Click to see source code for figure below.
+
+   .. literalinclude:: /_verbatim/GMT_Defaults_1b.txt
+
 .. figure:: /_images/GMT_Defaults_1b.*
    :width: 500 px
    :align: center
@@ -247,6 +255,10 @@ E.g., if your ``gmt.conf`` file has *x* offset = 3\ **c** as default, the
    More GMT parameters that affect plot appearance.
 
 .. _gmt_defaults_c:
+
+.. dropdown:: Click to see source code for figure below.
+
+   .. literalinclude:: /_verbatim/GMT_Defaults_1c.txt
 
 .. figure:: /_images/GMT_Defaults_1c.*
    :width: 500 px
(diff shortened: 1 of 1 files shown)


KristofKoch:
Great work @meghanrjones! I'm not sure about the prominence of the dropdowns. In my eyes they draw too much attention from the reader. I agree with @seisman in #4926 and would prefer the [`sphinx-togglebutton`](https://sphinx-togglebutton.readthedocs.io/en/latest/#toggle-any-content-with-the-toggle-directive). Compare for yourself in this crude mock-up:

dropdown vs toggle

Additionally I would place the script beneath the plot as this would aid the flow of reading:

your version:

* explaining text
* script for a figure I haven't seen yet
* illustrating figure
* figure caption

proposed version:

* explaining text
* illustrating figure
* figure caption
* script for figure

What do you think?

meghanrjones:
> Great work @meghanrjones! I'm not sure about the prominence of the dropdowns. In my eyes they draw too much attention from the reader. I agree with @seisman in #4926 and would prefer the [`sphinx-togglebutton`](https://sphinx-togglebutton.readthedocs.io/en/latest/#toggle-any-content-with-the-toggle-directive). Compare for yourself in this crude mock-up:
> 
> dropdown vs toggle
> 
> Additionally I would place the script beneath the plot as this would aid the flow of reading:
> 
> your version:
> 
>     * explaining text
> 
>     * script for a figure I haven't seen yet
> 
>     * illustrating figure
> 
>     * figure caption
> 
> 
> proposed version:
> 
>     * explaining text
> 
>     * illustrating figure
> 
>     * figure caption
> 
>     * script for figure
> 
> 
> What do you think?

I agree with placing the dropdown/toggle-button after the code. I also like the more subtle style of the togglebutton, but think that a more descriptive message for the button is essential (i.e., 'Click to show the source script' rather than 'Click to show'). I don't think this is included in the sphinx-togglebutton extension, but since the source code is so simple with a permissive license perhaps we could add the functionality or open an issue in sphinx-togglebutton. @seisman, do you know if changing the 'Click to show' text is already possible with that extension?

seisman:
> do you know if changing the 'Click to show' text is already possible with that extension?

Yes, the extension provides a configuration to change the hint text (see https://sphinx-togglebutton.readthedocs.io/en/latest/#configuration), but the text is the same for all toggle buttons.

I think we may want to use different texts for different cases. For example:

- "Click to show the source script"
- "Click to show an example"

It's possible to extend the directive to support a custom hint text, e.g., 
```
.. toggle:: Click to show an example
   
   IMAGE.

   SOURCE.
```

meghanrjones:
> > do you know if changing the 'Click to show' text is already possible with that extension?
> 
> Yes, the extension provides a configuration to change the hint text (see https://sphinx-togglebutton.readthedocs.io/en/latest/#configuration), but the text is the same for all toggle buttons.
> 
> I think we may want to use different texts for different cases. For example:
> 
>     * "Click to show the source script"
> 
>     * "Click to show an example"
> 
> 
> It's possible to extend the directive to support a custom hint text, e.g.,
> 
> ```
> .. toggle:: Click to show an example
>    
>    IMAGE.
> 
>    SOURCE.
> ```

Great, let's move forward with using the togglebutton for showing the scripts.

PR #4931, opened 2021-03-08
Shorten -J docs using or syntax

### LINKED ISSUE #2267: Avoid near-duplication in -J explanation
The **-J** option explanation has pairs of near-identical syntax cases, where we show one with lower-case code (and _scale_) and one with upper-case code (and _width_).  E.g., for Mercator we list

**-Jm**[_lon0/[lat0_/]]_scale_ or **-JM**[_lon0/[lat0_/]]_width_ (Mercator [**C**])

Given that we start **-J** by explaining the purpose of the lower and upper letter codes and that they mean we give map scale or map width, might we not be able to use the single statement

**-Jm**|**M**[_lon0/[lat0_/]]_scale_|_width_

to avoid the repetition that gets seriously long for some of the azimuthal projections?  Would you prefer

**-Jm**|**M**[_lon0/[lat0_/]]_scaleorwidth_

instead? I bring this up since I just fixed a copy/paste error for **-JO** due to this duplication.

meghanrjones:

Shortens the syntax for **-J** in the [gmt manpage](https://docs.generic-mapping-tools.org/dev/gmt.html#j-full) to use **lowercase**|**uppercase** and *scale*|*width*. 

**Preview link**: https://gmt-1vi3itdn8-gmt.vercel.app/gmt.html#j-full

Fixes #2267 

**Reminders**

### DIFF
--- a/doc/rst/source/explain_-J_full.rst_
+++ b/doc/rst/source/explain_-J_full.rst_
@@ -29,14 +29,13 @@ respectively):
 
 .. _-Jc_full:
 
-    **-Jc**\ *lon0/lat0/scale* or **-JC**\ *lon0/lat0/width* (Cassini).
+    **-Jc**\|\ **C**\ *lon0/lat0/scale*\|\ *width* (Cassini).
 
         Give projection center *lon0/lat0* and *scale* (**1:**\ *xxxx* or UNIT/degree).
 
 .. _-Jcyl_stere_full:
 
-    **-Jcyl_stere**/[*lon0/*\ [*lat0/*]]\ *scale* or
-    **-JCyl_stere**/[*lon0/*\ [*lat0/*]]\ *width* (Cylindrical Stereographic).
+    **-Jcyl_stere**\|\ **Cyl_stere**\ /[*lon0/*\ [*lat0/*]]\ *scale*\|\ *width* (Cylindrical Stereographic)
 
         Give central meridian *lon0* (optional), standard parallel *lat0*
         (optional), and *scale* along parallel (**1:**\ *xxxx* or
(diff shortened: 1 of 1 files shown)


PaulWessel (commented):
Looks great . Only comment is related to *scale*.  Do we need to make it even clearer in the first 1-2 sentences under -J that if you use upper case then width is just that, otherwise a scale is needed and there are further instruction what the scale means per proection?  I am trying to unconfuse people who want to use upper case but then read

        `Give projection center *lon0/lat0* and *scale* (**1:**\ *xxxx* or UNIT/degree).`

and wonder what happen to width?

meghanrjones:
> Looks great . Only comment is related to _scale_. Do we need to make it even clearer in the first 1-2 sentences under -J that if you use upper case then width is just that, otherwise a scale is needed and there are further instruction what the scale means per proection? I am trying to unconfuse people who want to use upper case but then read
> 
> ```
>     `Give projection center *lon0/lat0* and *scale* (**1:**\ *xxxx* or UNIT/degree).`
> ```
> 
> and wonder what happen to width?

I will revise to make it more clear. I think putting 'width' in all caps would also make the relationship more obvious. For example: **-Jm**|**M**[_lon0/[lat0_/]]_scale_|_WIDTH_

meghanrjones:
> Looks great . Only comment is related to _scale_. Do we need to make it even clearer in the first 1-2 sentences under -J that if you use upper case then width is just that, otherwise a scale is needed and there are further instruction what the scale means per proection? I am trying to unconfuse people who want to use upper case but then read
> 
> ```
>     `Give projection center *lon0/lat0* and *scale* (**1:**\ *xxxx* or UNIT/degree).`
> ```
> 
> and wonder what happen to width?

Here's how the formatting looks now. I don't think it's worth spending much time revising the preamble under **-J***parameters* right now as that may get wrapped into the cookbook. I can open a separate issue for it in the case that the common options do not get merged with the cookbook common options.

image

PaulWessel (approved):

PR #4937, opened 2021-03-10
Add autosectionlabel sphinx extension and rename duplicate headers

meghanrjones:

This PR adds the [sphinx autosection label](https://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html), which is also described in the [sphinx documentation on cross-referencing](https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#automatically-label-sections). It will enable linking directly to section headers in various files, rather than needing to set up explicit, unique references.
For example, this:
```
:ref:`cookbook/features:The background panel`
```
Rather than adding ``.. _Background-panel:`` above 'The background panel' in cookbook/features.rst section and then linking using:
```
:ref:`The background panel <Background-panel>`
```
I think this will make adding cross-references easier moving forward. It also will make it more obvious where the references being linked to. The downside is that duplicate headers within individual files cause warnings. Those have been renamed as part of this pull request, notably adding the GMT version to each section header in the changelog.

Fixes #

**Reminders**

### DIFF
--- a/doc/rst/conf.py.in
+++ b/doc/rst/conf.py.in
@@ -9,7 +9,7 @@ sys.path.append(os.path.abspath('_extensions'))
 
 # -- General configuration -----------------------------------------------------
 needs_sphinx = '1.8'
-extensions = ['sphinx.ext.mathjax', 'jinja', 'youtube', 'sphinx_panels', 'sphinx_togglebutton', 'sphinx_copybutton']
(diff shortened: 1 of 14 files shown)


PaulWessel (approved):
Does this extension require new code inclusion like what @seisman had to do for me since I am not using conda but macport?

meghanrjones:
> Does this extension require new code inclusion like what @seisman had to do for me since I am not using conda but macport?

No, this one is bundled with Sphinx

seisman (approved):

PR #4939, opened 2021-03-10
Old bug was only partly fixed 3 years ago

PaulWessel:
There are a factor of 0.5 that should have been 1.0.  @remkos fixed this 3 years ago but there was two places where the 0.5 issue appeared.  This PR fixes the 2nd instance.  In the process I replace sqrt(2) with constant M_SQRT2.

I am leaving this as WIP since the change causes our sph2grd/EGM96.sh to fail (which is expected) but I would like the original poster to supply me a test to make sure we know get the correct results.


PaulWessel:
Hi @remkos, see the [old forum](http://gmt.soest.hawaii.edu/issues/1127) for details.  Looks to me that the two ortho sections in gmt_stat.c need to look like this:

`pmm /= (M_SQRT2 * M_SQRTPI);`

With that change, I can run **sph2grd** and **grdmath** separately and get a surface integral that equals unity:

```
echo 10 5 1 0 > t.txt
gmt sph2grd t.txt -Rg -I1 -Nm -Gsh.grd
res="1" # deg - resolution of SH grid
pi="3.141592"
echo "Integrated Ylm**2 volume = "$(gmt grd2xyz sh.grd | awk 'BEGIN {sum=0} {sum+=(sin((90-$2)*'$pi'/180)*'$res'*'$res'*('$pi'/180)*('$pi'/180)*($3^2))} END {printf "%.6f\n", sum}')
gmt grdmath -Rg -I1 10 5 YLM = sh.grd POP
echo "Integrated Ylm**2 volume = "$(gmt grd2xyz sh.grd | awk 'BEGIN {sum=0} {sum+=(sin((90-$2)*'$pi'/180)*'$res'*'$res'*('$pi'/180)*('$pi'/180)*($3^2))} END {printf "%.6f\n", sum}')
```

Both of the sums equal ~1.  In master the **sph2grd** grid sums to 0.5 and the **grdmath** grid sums to 2.
Do you agree with this change?

PaulWessel:
Looks like I will have to make a decision on this as is.  For the record, in master the above two **sph2grd** and awk commands that does the surface integral numerically yields 0.5 and 2, respectively.  Clearly not 1.  With my fix, both integrals add to ~1.

PaulWessel:
Any concerns with this one, @joa-quim @meghanrjones @seisman ? Seems pretty clear but neither @remkos nor the original poster on the old forum has responded in many weeks.

meghanrjones:
> Any concerns with this one, @joa-quim @meghanrjones @seisman ? Seems pretty clear but neither @remkos nor the original poster on the old forum has responded in many weeks.

I do not have any concerns here.

PaulWessel:
All tests pass, hopefully someone can have a look and approve.

meghanrjones:
> All tests pass, hopefully someone can have a look and approve.

I held off on approving based on the original comment that a test is pending. Do you think that your **sph2grd** and **grd2xyz** comparison could work as a test?

meghanrjones (approved):

PaulWessel:
OK, I added a test that passes.

meghanrjones:
> OK, I added a test that passes.

Looks good.

PR #4942, opened 2021-03-11
Add toggle-buttons showing source scripts to cookbook

meghanrjones:

This PR adds toggle-buttons to show the source scripts used to generate figures in the cookbook, when the source code isn't already included.

This is what it looks like:

image

And toggled open:

image

Fixes #

**Reminders**


PaulWessel (approved):
Looks good!

PR #4950, opened 2021-03-12
CI: Build docs and run tests based on changed files

seisman:

We now have 4 CI workflows:

- `build.yml`: Build the GMT source for **every commit in PR and in master branch**
- `docker.yml`: Build the GMT source on other Ubuntu and Debian versions [**only runs in the master branch**]
- `docs.yml`: Build the GMT documentation and deploy it to gh-pages branch [**only runs in the master branch**]
- `tests.yml`: Run the full tests [**only runs in the master branch**]

As all the documentation (ReST files and images) are stored in the `doc` directory, it wastes CI resources to build the documentation if only GMT C files are changed. Similarly, if only ReST files are changed, we also don't need to run the full tests. 

This PR improves the `docs` and `tests` workflows, so that they only run if a changed file match the specified patterns.

The rules are:

- If any changed files match `doc/**`, the `docs` workflow will run
- If any changed files match `src/**`, 'test/**', 'doc/examples/**', and 'doc/scripts/**', the `tests` workflow will run.

Do the rules make sense to you? Do I miss anything?

**Reminders**

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,6 +9,10 @@ on:
     branches:
       - master
       - 6.[0-9]+
+    paths:
+      - 'doc/**'
+      - '.github/workflows/**'
+      - 'ci/**'
   release:
     types:
       - published
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,6 +9,13 @@ on:
     branches:
       - master
       - 6.[0-9]+
+    paths:
+      - 'src/**'
+      - 'test/**'
+      - 'doc/examples/**'
+      - 'doc/scripts/**'
+      - '.github/workflows/**'
+      - 'ci/**'
 
 defaults:
   run:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ on:
     paths:
       - 'src/**'
       - 'test/**'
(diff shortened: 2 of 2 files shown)


PaulWessel:
Should we be concerned about changes to share/cpt/*.cpt ?  Rare, but does happen.

seisman:
> Should we be concerned about changes to share/cpt/*.cpt ? Rare, but does happen.

You mean running tests if any "share/cpt/*.cpt" change? It makes sense.

PaulWessel:
yes, it could affect the tests. Maybe anything in share actually.  IF we get a typo in share/localization/gmt_de.locale [and we actually ahd a test for those ...] then tests should be run again.  I think you can check for any share/** changes since they are super-infrequent (perhaps except for cpt) but it would cover us that one time we change them.

seisman:
> yes, it could affect the tests. Maybe anything in share actually. IF we get a typo in share/localization/gmt_de.locale [and we actually ahd a test for those ...] then tests should be run again. I think you can check for any share/** changes since they are super-infrequent (perhaps except for cpt) but it would cover us that one time we change them.

Sounds reasonable. Done in 5bd885cbd4272c83b501a94f8e7499de793e60ab.

PaulWessel (approved):

meghanrjones (approved):

PR #4952, opened 2021-03-13
Update GMT TM PostSCript

PaulWessel:
For some reason did not fail test. @seisman why would the GMT_TM.sh test not fail when the orig had crooked annotations and the new one horizontal.  Surely a large RMS.  @meghanrjones even made the RMS smaller to see but still no fail?

### DIFF
--- a/doc/scripts/GMT_TM.ps
+++ b/doc/scripts/GMT_TM.ps
@@ -1,11 +1,11 @@
 %!PS-Adobe-3.0
 %%BoundingBox: 0 0 612 792
 %%HiResBoundingBox: 0 0 612.0000 792.0000
-%%Title: GMT v6.1.0_41e2997-dirty_2020.05.26 [64-bit] Document from pscoast
+%%Title: GMT v6.2.0_1afb40d_2021.03.12 [64-bit] Document from coast
 %%Creator: GMT6
 %%For: unknown
 %%DocumentNeededResources: font Helvetica
-%%CreationDate: Wed May 27 12:50:12 2020
+%%CreationDate: Fri Mar 12 14:49:51 2021
 %%LanguageLevel: 2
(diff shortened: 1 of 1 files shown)


seisman:
https://github.com/GenericMappingTools/gmt/blob/acbef76c415069162b5cf5d2198d80f8fd3ed7bf/test/gmtest.in#L15-L16

It's a modern mode script, but the variable `modern` is incorrectly given a 0 value.

seisman:
https://github.com/GenericMappingTools/gmt/blob/acbef76c415069162b5cf5d2198d80f8fd3ed7bf/test/gmtest.in#L284-L287

Perhaps we should change these lines to:
```
 if [ "X$ps" = "X" ]; then 
 	ps="${local_script%.sh}.ps" 
 fi
```
If the variable is empty, we should assume that the PS file is the same as the script file.

PaulWessel:
Ah, yes, that is a good solution.

seisman:
> Ah, yes, that is a good solution.

It's unclear to me whether the solution will have a bad effect on those pass-fail tests.

PaulWessel:
Hm, will need to look at the test script to see what a $ps variable does.  I think it tries to see if the file exist before doing things.  Anyway TGIF seminar in 2 minutes here....

seisman:
I'll run the full tests and see how it works.

PaulWessel:
Maybe @meghanrjones can look at the new failures and decide if it is mostly more of the same (update PS only).

meghanrjones:
> Maybe @meghanrjones can look at the new failures and decide if it is mostly more of the same (update PS only).

Yes, will do.

meghanrjones:
Closing this one because GMT_TM is included in #4962

PR #4958, opened 2021-03-13
Consolidate decisions on annotating pole in one function

PaulWessel:
Add _gmtplot_skip_pole_lat_annotation_ which prevents +/- 90 latitude from being annotation for polar projections as well as Eckert 4 (too gentle slope near pole).

This means 

`gmt basemap -Rd -JKf12c -Bafg  -png map`

no longer tries to place 90 by using the slope of the boundary near the poles (which is too gentle and we get crazy offset).

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -2584,10 +2584,12 @@ GMT_LOCAL void gmtplot_consider_internal_annotations (struct GMT_CTRL *GMT, stru
 }
 
 bool gmtplot_skip_pole_lat_annotation (struct GMT_CTRL *GMT, double lat) {
-	if (fabs (lat) < 90.0) return false;	/* Not at the poles so OK to annotate */
 	/* Here, latitude is -90 or +90 and the question is do we annotation that latitude? */
+	if (GMT->current.proj.projection_GMT == GMT_ECKERT4   && fabs (lat) > 85.0) return true;	/* Slope too gentle near poles to adjust via boost */
+	if (GMT->current.proj.projection_GMT == GMT_HAMMER    && fabs (lat) > 85.0) return true;	/* Slope too gentle near poles to adjust via boost */
+	if (GMT->current.proj.projection_GMT == GMT_MOLLWEIDE && fabs (lat) > 88.0) return true;	/* Slope too gentle near poles to adjust via boost */
+	if (fabs (lat) < 90.0) return false;	/* Not at the poles so OK to annotate */
 	if (GMT->current.proj.polar) return true;	/* Cannot since inside map */
-	if (GMT->current.proj.projection_GMT == GMT_ECKERT4) return true;	/* Slope too gentle near poles to adjust via boost */
 	return false;
 }
 
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -2583,6 +2583,14 @@ GMT_LOCAL void gmtplot_consider_internal_annotations (struct GMT_CTRL *GMT, stru
 	PSL_settextmode (PSL, PSL_TXTMODE_HYPHEN);	/* Back to leave as is */
 }
 
+bool gmtplot_skip_pole_lat_annotation (struct GMT_CTRL *GMT, double lat) {
+	if (fabs (lat) < 90.0) return false;	/* Not at the poles so OK to annotate */
+	/* Here, latitude is -90 or +90 and the question is do we annotation that latitude? */
+	if (GMT->current.proj.polar) return true;	/* Cannot since inside map */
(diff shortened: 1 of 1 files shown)


PaulWessel:
The check on Van grinten should not be needed since for such projections there is no "south" or "north" boundary if S = -90 and N = +90.  This is why there is no pole annotation for Hammer, Mollweide, etc.

meghanrjones (commented):
The annotations start to look bad at >85 latitude. Do you think it's better to set this at 90 because that is the only likely 'automatic' annotation selection?

PaulWessel:
Yes, I see. Not sure what you are asking though: You want to label the 90 degree latitude for Eckert 4 somehow?

src/gmt_plot.c
```diff
@@ -2583,6 +2583,14 @@ GMT_LOCAL void gmtplot_consider_internal_annotations (struct GMT_CTRL *GMT, stru
 	PSL_settextmode (PSL, PSL_TXTMODE_HYPHEN);	/* Back to leave as is */
 }
 
+bool gmtplot_skip_pole_lat_annotation (struct GMT_CTRL *GMT, double lat) {
+	if (fabs (lat) < 90.0) return false;	/* Not at the poles so OK to annotate */
```
meghanrjones, line 2587:
```suggestion
	if (fabs (lat) < 86.0) return false;	/* Not at the poles so OK to annotate */
```
I was thinking of this type of situation which might motivate a cutoff < 90 (``gmt basemap -Rd -JKf12c -Ba11fg -png map``), but perhaps it should be allowed even if bad looking

PaulWessel: I agree that there is no simple way to place latitudes > 85 for -JKf so we should prevent it.  But only for this projection since the others that have the pole as a straight line have a much larger "take-off" angle at the end of that line so the algorithm I implemented works.  It is just -Jf that as such a tiny angle from 90 to 89 that we end up with small/tiny and it blows up.  So if you agree I will modify the check to do < 85 for Eckert 4 but not the others.

PaulWessel: Hm, actually, you may have a point.  85 does not look great for hammer either.

PaulWessel: Not out of woods: Here is what I see when testing with

`gmt basemap -Rd -JH100c -Ba2fg  -png map --MAP_ANNOT_MIN_SPACING=0 --FONT_ANNOT_PRIMARY=9p`

changing -J to the different misc. projections.

1. Mollweide looks passible up to 88 for me
2. Hammer, which is almost the same, does not.  It needs to stop at 85.
3. Eckert 6 is OK in the sense of the offset, but the y-change from 85 to 90 is so tiny it will overprint except for our new MIN_ANNOT_OFFSET setting.  Set that to 0 and it looks ugly.  I think that may be OK.
4. THe Robinson, WInkel, Sinusoidal, Van Grinten all look OK

There is of course some crowding with those I deem work OK, but that is handled by the min spacing part.  So I suggest adding these specific limits for Mollweide [88], Hammer [85], and Eckert 4 [85] with the others at 90.

meghanrjones: Yes, seems limited to those two projections.

meghanrjones: > Not out of woods: Here is what I see when testing with
> 
> `gmt basemap -Rd -JH100c -Ba2fg -png map --MAP_ANNOT_MIN_SPACING=0 --FONT_ANNOT_PRIMARY=9p`
> 
> changing -J to the different misc. projections.
> 
>     1. Mollweide looks passible up to 88 for me
> 
>     2. Hammer, which is almost the same, does not.  It needs to stop at 85.
> 
>     3. Eckert 6 is OK in the sense of the offset, but the y-change from 85 to 90 is so tiny it will overprint except for our new MIN_ANNOT_OFFSET setting.  Set that to 0 and it looks ugly.  I think that may be OK.
> 
>     4. THe Robinson, WInkel, Sinusoidal, Van Grinten all look OK
> 
> 
> There is of course some crowding with those I deem work OK, but that is handled by the min spacing part. So I suggest adding these specific limits for Mollweide [88], Hammer [85], and Eckert 4 [85] with the others at 90.

Seems fine.

meghanrjones:
In the description example (`gmt basemap -Rd -JKf12c -Bafg -png map`), why is there only a tick mark for the central meridian? Is this how it should be?

meghanrjones (approved):

PaulWessel:
I think for all other meridians the angle of the tick is almost horizontal, which was the problem with the rogue 90 annotation too.  So for at least this misc projection (and probably for those that have a polar line), we should default to normal ticks unless the user sets MAP_ANNOT_OBLIQUE manually.

PaulWessel:
So I will merge this one but need to start another branch dealing with those projections and tick_normal.

PR #4957, opened 2021-03-13
Only set undefined axes when drawn

meghanrjones:
**Description of proposed changes**

Fixes these two examples:
```
gmt basemap -R0/1/0/1 -JX5c -B+t"ABC" -png map
```

```
gmt begin colorbar png
gmt makecpt -Cbatlow -T0/10/1
gmt basemap -R20/30/-10/10 -B
gmt colorbar
gmt end show
```

Fixes #4955 (partly)

**Reminders**

- [ ] Make sure that your code follows our style. Use the other functions/files as a basis.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Describe changes to function behavior and arguments in a comment below the function declaration.
- [ ] If adding new functionality, add a detailed description to the documentation and/or an example.

Issue #4955, opened 2021-03-13
Potential issues or breaking changes with the GMT modern theme?


seisman:
Running this command:
```
gmt basemap -R0/1/0/1 -JX5c -B+t"ABC" -png map
```

GMT v6.1.1 doesn't plot any axes, only the title, but GMT master branch plots the 4 axes. 

| v6.1.1 | master |
|---|---|
| ![image](https://user-images.githubusercontent.com/3974108/111018261-dc9b8200-8385-11eb-97f0-4dd14f7001c8.png) | ![image](https://user-images.githubusercontent.com/3974108/111018239-bece1d00-8385-11eb-8eee-1f3548780637.png) |

The changes are a big surprise to me and look like a bug.

seisman:
```
gmt basemap -Rg -JW15c -B -png map
```

| v6.1.1 | master |
|---|---|
| ![image](https://user-images.githubusercontent.com/3974108/111018668-238a7700-8388-11eb-88aa-c207b49dc4ff.png) | ![image](https://user-images.githubusercontent.com/3974108/111018682-34d38380-8388-11eb-8be1-6ee69ff8694b.png) |

seisman:
```
gmt begin colorbar png
gmt makecpt -Cbatlow -T0/10/1
gmt basemap -R20/30/-10/10 -B
gmt colorbar
gmt end show
```

I know the east and north annotations and ticks are not plotted by default, but the color bar is not annotated, too.

| v6.1.1 | master |
|---|---|
| ![image](https://user-images.githubusercontent.com/3974108/111018775-dce94c80-8388-11eb-9586-1517d0d790ce.png) | ![image](https://user-images.githubusercontent.com/3974108/111018786-ef638600-8388-11eb-9643-e7513629466a.png) |

seisman:
The automatic annotation intervals are not that good for small regions.

```
gmt basemap -JM6c -R-125/-121/47/48 -B -png map
```

| v6.1.1 | master |
|---|---|
| ![image](https://user-images.githubusercontent.com/3974108/111019014-8250f000-838a-11eb-95c0-840bf37e8b43.png) | ![image](https://user-images.githubusercontent.com/3974108/111019009-79f8b500-838a-11eb-9c6f-68189591e66f.png) |

seisman:
This one is pretty bad:

```
gmt coast -Rd -JKf12c -Bafg -Ggreen -png map
```

[...]

PR #4960, opened 2021-03-14
Draw ticks normal to line when a pole plots as a line

PaulWessel:
This only affects the non-oblique regions for Robinson, Winkel-Tripel and the two Eckert projections

There are many (> 10) failures because the default changes.  Please see if you agree than all the ones using these 4 projections look better with this fix, e.g ex46. ex25, ex22, ex20, urlquakes, etc etc.


PaulWessel:
Here is the list of failures that I attribute specifically to the Winkel, Robinson, Eckert upgrade:

ex46
ex25
ex22
ex20
urlquakes
pssolar_fill
pssolar_fill2
select
closed
GMT_tut_6

There are many other failures that are unrelated to this change in tick-normal.  If you agree then I can update these PS files.

meghanrjones:
> Here is the list of failures that I attribute specifically to the Winkel, Robinson, Eckert upgrade:
> 
> ex46
> ex25
> ex22
> ex20
> urlquakes
> pssolar_fill
> pssolar_fill2
> select
> closed
> GMT_tut_6
> 
> There are many other failures that are unrelated to this change in tick-normal. If you agree then I can update these PS files.

Looking now

meghanrjones (approved):
I agree that the annotations look better for those scripts. I think that the best appearance would be with slanted ticks and a constant annotation distance from the frame, but perhaps that is not an option.

I'm not sure it's worth updating the PS files until all the issues in #4955 are resolved since they are likely to change again - up to you.

PaulWessel:
Yes, and no.  I worry that if we do not update the PS files but just merge in the fix that it is a bit hard to keep track of what is actually fixed and what is still out there.

PaulWessel:
> I agree that the annotations look better for those scripts. I think that the best appearance would be with slanted ticks and a constant annotation distance from the frame, but perhaps that is not an option.

It cannot work for some since the tick angle is crazy horizontal for some of them, and they we would have to invent some threshold angle instead that would not represent map trends etc.  I think the clean normal is pretty decent and at least is a big improvement over the bumpy lineup with the central meridian sticking down.

meghanrjones:
> > I agree that the annotations look better for those scripts. I think that the best appearance would be with slanted ticks and a constant annotation distance from the frame, but perhaps that is not an option.
> 
> It cannot work for some since the tick angle is crazy horizontal for some of them, and they we would have to invent some threshold angle instead that would not represent map trends etc. I think the clean normal is pretty decent and at least is a big improvement over the bumpy lineup with the central meridian sticking down.

OK, agreed.

PaulWessel:
Looking at ex16 which is master now has missing CPT annotations but running master produces them.  So looks like we merged in some PS files with issues related to CPT annotations but whatever the problem was it is no longer an issue?  E.g., ex16, categoricalcpt.ps.

meghanrjones:
> Looking at ex16 which is master now has missing CPT annotations but running master produces them. So looks like we merged in some PS files with issues related to CPT annotations but whatever the problem was it is no longer an issue? E.g., ex16, categoricalcpt.ps.

Yes, sorry. I fixed in #4957 but did not update PS files.

PaulWessel:
OK, you or me updating those two?
After than I see several with slight differences in map pen thickness (?), and GMT_cassini has a different interval but was not updated? Then GMT_TM looks OK in the failure so update that too?

meghanrjones:
> OK, you or me updating those two?
> After than I see several with slight differences in map pen thickness (?), and GMT_cassini has a different interval but was not updated? Then GMT_TM looks OK in the failure so update that too?

The GMT_cassini is probably related to https://github.com/GenericMappingTools/gmt/pull/4954 and the map pen thicknesses very likely could be related to that as well. I'll go through all the PS files that failed after #4954 and #4957 now and make a PR with the updates.

PR #4962, opened 2021-03-14
Update PS files for failing tests after #4954 and #4957

meghanrjones:

Updates PostScript files for tests that starting failing after #4954 and #4957. The PS files for two failing tests were not updated because of missing annotations:

	 98 - doc/scripts/GMT_gnomonic.sh (Failed)
	133 - doc/scripts/GMT_perspective.sh (Failed)

Fixes #

**Reminders**


PaulWessel (approved):

PaulWessel:
OK, down to those two.  I have a family zoom the next 2 hours so will resume then.  Let me now if you figure anything out.

PaulWessel:
Clearly the perspective view one is affected by the default WSrt but here S is curved and not used so we want N instead.  I think not annotating E is OK. Without checking, not sure how hard to detect the type of boundaries here, so one cheap solution is to revert to WESN in these cases.  I think it is hard to know exactly how the window view affects what the borders look like.  Once plotted, the user can select axes manually but at least we show what it looks like for alll.
Gnomonic: Not sure why it turned off but presumably related to WESN.

meghanrjones:
> Clearly the perspective view one is affected by the default WSrt but here S is curved and not used so we want N instead. I think not annotating E is OK. Without checking, not sure how hard to detect the type of boundaries here, so one cheap solution is to revert to WESN in these cases. I think it is hard to know exactly how the window view affects what the borders look like. Once plotted, the user can select axes manually but at least we show what it looks like for alll.
> Gnomonic: Not sure why it turned off but presumably related to WESN.

Yes, the east and north axes are plotted on the Gnomonic plot, so WESN works but WEsn does not show any annotations (before or after themes).

PaulWessel:
Gnomonic was the last projection to even get any annotations (some years ago) as it was messy.  Our simplest fix for these two projections would be to default to WESN.  It should be documented though, maybe on the themes page, and I suspect these two are used so infrequently that maybe a warning message would not be bothersome, saying that due to the nature of this projection, we have reset to WESN (again, unless the user changed MAP_ANNOT_AXES themselves).

meghanrjones:
> Gnomonic was the last projection to even get any annotations (some years ago) as it was messy. Our simplest fix for these two projections would be to default to WESN. It should be documented though, maybe on the themes page, and I suspect these two are used so infrequently that maybe a warning message would not be bothersome, saying that due to the nature of this projection, we have reset to WESN (again, unless the user changed MAP_ANNOT_AXES themselves).

It will only default to WESN for MAP_FRAME_AXES=auto, correct? This seems fine and we can add it to the **MAP_FRAME_AXES** info here: https://docs.generic-mapping-tools.org/dev/cookbook/features.html#automatic-gmt-settings

PaulWessel:
Yes, let me look at that setup.  Need to know the projection so that is the same place where font sizes etc are set.

PR #4963, opened 2021-03-15
Adjust auto axes for gnomonic and general perspective

PaulWessel:
This addresses the difficulty of forcing **WrStZ** on the gnomonic and general perspective projection - here we reset to **WESNZ** unless the user set something manually.  Updating the plots since they now have W,E,S,N in the annotations.


meghanrjones (approved):
Looks good. I just pushed a change with a documentation update.

Issue #4866, opened 2021-02-27
Let events plot more complicated symbols via meca, velo, etc.


PaulWessel:
**Description of the desired feature**

I believe @Esteban82 asked this on the forum some time ago: How to make movies using **events** to plot focal mechanisms?  I have an interest in getting this feature implemented as well and here are some thoughts on how this could work.  First, a brief overview of how regular symbols are plotted.  These are simple (circle, star, etc) so only one parameter.  In contrast, focal mechanisms plotted via **gmt meca** may take 7+ parameters.  Here is the workflow:

1. Input data need 1-2 extra columns appended at the end: The time of the event (required) and the optional duration of it (if variable.  If constant then we can set it via **-L**).
2. Given the current frame time and the various parameters set via **-E**, **-M**, **-L**, determine _which_ symbols should be visible (we skip the rest) and determine the individual symbol scale, color intensity and transparency for each one of them.
3. Write a temporary data file with all the events that must be plotted, and append the three extra columns setting symbol scale, color intensity, and transparency.
4. Call **gmt plot** on this temporary file

To "generalize" this workflow to apply to more complicated symbols via **meca** and others, I think this may be the path forward:

First, modules like **meca** and **velo** will need new options to allow the passing via extra data columns of symbol size (or more like an overall scale), color intensity, and transparency.  This could be achieve this way:

1. Symbol size/scale: Since these modules have their own **-S** option to set many aspects of the symbol type and size, we can add another modifier that basically means "then scale the final symbol size by the factor in the first extra data column". Let's call this modifier **+z**.
2. Intensity: All of these modules can be given the same **-I**[_intens_] option found in **plot** and then use the second extra column for an intensity value to modify the present color.
3. Transparency: I believe the modules already have an appropriate **-t** option since it is a common option but I will need to see if it actually works the same way as in **plot** (**-t** without arg means read transparency from last column).

With those changes the modules will be capable of issuing the correct plot code when driven by **events** as well as when run independently.

The second challenge is to tell **events** you want one of these symbols.  I think it is better not to clutter the present **-S** option (standard symbols) but instead use **-Z** to select these more advanced symbols.  Because **meca**, **velo**, **coupe** etc take many options and they are not all identical, it is not a simple thing to do, especially the many data formats and conventions allowed.  My preferred solution to this goes like this:

[...]

PR #4964, opened 2021-03-15
Add -H to plot for symbol and pen attribute scaling

PaulWessel:

See #4866 for background.  This PR adds **-H** to **plot** which can be used to scale the symbol size _as well as_ the symbol pen outline attributes (width, style) by the provided scales in the data file.  Documentation has been updated with a **Data Column Order** explanatory note (as was done for **meca**, **coupe**, and **velo**), and a simple test _scalesymb.sh_ was added to demonstrate that it works with both a fixed symbol size or a variable one (i.e., size read from file, then multiplied by scale read from file).

This feature is required for properly scaled animations via **events** but the feature can also be used for other **plot** purposes.


PaulWessel:
Decided to combine both psxy and psxyz in the same PR since similar.

meghanrjones:
Does **-H** scale just the pen or the pen and the size? I think it would be more flexible to have **-H** only scale the pen and leave the size scaling to **-S**, but the documentation for **-H** says it scales both the size and the pen.

meghanrjones:
> Does **-H** scale just the pen or the pen and the size? I think it would be more flexible to have **-H** only scale the pen and leave the size scaling to **-S**, but the documentation for **-H** says it scales both the size and the pen.

Second comment - if it only scales the pen, could this be done as a modifier to **-W** rather than a new flag. It seems like a small feature for one of the three remaining upper case letters for psxy.

joa-quim:
I got more convinced by your example than the feature description. Even so that I wonder if that should not be a command line flag to reproduce this effect. After all, if it's possible to achieve via file it should also be possible via command option.

PaulWessel:
Remember that plot can accept a fixed _size_ (**-Sc**1c) as well as variable _size_ (**-Sc**); in the latter case we read actual dimensions (including units if appended) from the next column after the optional z-column (**-C**).  This is a separate thing from an overall _scale_ which is nondimensional and scales the current symbol size (which is either fixed or variable, as above) and also adjusts pen attributes related to dimension (i..e, width and gap/dash lengths). To do this effect requires the _scale_ column which is not the same as a _size_ column and both may be needed for some applications.  So no, it cannot just be a modifier to **-W** since it does affect the symbol size.
I am not sure what @joa-quim is saying since this is a command flag and is doable via command option - as the test scripts do.

PaulWessel:
I chose **-H** since available across the 5 modules that need this effect.

joa-quim:
Yes, I got that but what I'm saying is that we could as well want to do ``-Sc1c -W2p -H5`` and have that same symbol be  scaled differently on different plot calls for example. It's an extension of the ``-S<symb><size>``, ``-S<symb>`` case.

PaulWessel:
So **-H**_scale_ would imply a constant scale and no scale column, whereas **-H** would read the scale from file?  I can implement that very easily.

joa-quim:
Yes, exactly.

PaulWessel:
OK, **-H**_scale_ implemented, tests still pass and the new feature works fine, e.g.

`echo 0 0 | gmt plot -R-2/2/-2/2 -JX4i -B -Sc1i -W1p -Gred -H3 -pdf t`

joa-quim (approved):

meghanrjones (approved):

PR #4970, opened 2021-03-16
Upgrade events and meca, coupe, velo to handle new scaling column

PaulWessel:

Along the lines of the upgrades to **plot** and **plot3d**, this PR lets **meca**, **coupe**, and **velo** have the same new **-H**[_scale_] option.  This allows these modules, like **plot** and **plot3d**, to scale a symbol and its pen using **-H**.  The **events** module was upgraded (and simplified in the process) to no longer worry about if **-S** has a symbol size of not.  If it does not then the user presumably has a size column, otherwise it is a constant.  Regardless of that, **events** will add a scaling column and pass **-H** to the module for plotting symbols so that as these change size, both the symbol size and the pen attributes change accordingly.

Both **psxy** and **psxyz** needed some more work to deal with ellipses and rectangles since those dimensions were not in _S.size_x_.  These upgrades were required to make the events examples and movies work.

All tests pass and I checked that _anim08_ and _anim09_ still work.


meghanrjones (approved):

PaulWessel:
While it should not affect any user, just pining @Esteban82 so he knows there has been some code changes and let me know if anything changed in his movies.

Esteban82:
I re run my animation and I haven't any problem or modification on it.

PR #4971, opened 2021-03-16
Address limitation of slice annotations in colorbar

PaulWessel:
As pointed out on the [forum]( https://forum.generic-mapping-tools.org/t/colorbar-label-two-ends-with-different-texts/1464 ) we have no solution for annotating the end of the last slice when **-L** is used.  This PR addresses that shortcoming by allowing the last slice to have an optional _second_ label following a second semi-colon.  For instance, in the case of the forum example, the last slice would be

3000 43/30/80 4500 43/30/80 B ; 3000 ; >4500

This yields a plot like this with this PR:

`gmt colorbar -Cmy.cpt -Dx2i/2i+w4i+h -L -png t`

t

After implementing this feature, the _GMT_App_M_2_ test failed but by adding "Precambrian" as the 2nd label for the last slice we get a more consistent plot, and when **-L** is given an argument (e.g., **-L**i or **-L**0.1) the second annotation is not used since the single or first annotation is then centered on each slice.

GMT_App_M_2

The documentation has been updated to explain this effect.  The implementation was chosen as it has no side-effects on the API.

### DIFF
--- a/doc/rst/source/cookbook/cpts.rst
+++ b/doc/rst/source/cookbook/cpts.rst
@@ -103,10 +103,9 @@ are centered. The example below also shows how to annotate ranges using
 **-Li** (in which case no name labels should appear in the CPT),
 and how to switch the color bar around (by using a negative length).
 
-Finally, note that if the last slice should have both lower and upper
-custom labels then supply two semicolon-separated labels. Furthermore,
-the annotation code should be **L** for all slices except the last, which
-must be set to **B**.
+**Note**: If the last slice should have both lower and upper
+custom labels then you must supply *two* semicolon-separated labels and set the
+annotation code to **B**.
 
 .. figure:: /_images/GMT_App_M_2.*
    :width: 600 px
--- a/doc/rst/source/cookbook/cpts.rst
+++ b/doc/rst/source/cookbook/cpts.rst
@@ -104,7 +104,9 @@ are centered. The example below also shows how to annotate ranges using
 and how to switch the color bar around (by using a negative length).
 
 Finally, note that if the last slice should have both lower and upper
-custom labels then supply two semicolon-separated labels.
+custom labels then supply two semicolon-separated labels. Furthermore,
+the annotation code should be **L** for all slices except the last, which
+must be set to **B**.
 
 .. figure:: /_images/GMT_App_M_2.*
    :width: 600 px
--- a/doc/rst/source/cookbook/cpts.rst
+++ b/doc/rst/source/cookbook/cpts.rst
@@ -103,6 +103,9 @@ are centered. The example below also shows how to annotate ranges using
 **-Li** (in which case no name labels should appear in the CPT),
 and how to switch the color bar around (by using a negative length).
 
+Finally, note that if the last slice should have both lower and upper
+custom labels then supply two semicolon-separated labels.
+
 .. figure:: /_images/GMT_App_M_2.*
    :width: 600 px
    :align: center
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -1294,7 +1294,9 @@ be omitted to determine the annotation and tick interval automatically
 make :doc:`/colorbar`, when used with the
 **-L** option, place the supplied label instead of formatted *z*-values.
 **Note**: The last slice may have two semicolon-separated labels to set
-labels for both the lower *and* upper boundary.
+labels for both the lower *and* upper boundary. Furthermore,
+the annotation code **A** should be **L** for all slices except the last,
+which must be set to **B**.
 
 The background color (for *z*-values < :math:`z_0`), foreground color (for *z*-values >
 :math:`z_{n-1}`), and not-a-number (NaN) color (for *z*-values =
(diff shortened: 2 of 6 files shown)


meghanrjones:
This seems correct (when all labels are upper (U) there is no lower-most label):

```
cat << EOF > colors.cpt
0	white	200 white U ; <200
200 230/230/188 300 230/230/188 U ; 200
300 195/203/160 500 195/203/160 U ; 300
500 155/175/172 800 155/175/172 U ; 500
800 115/149/184 1200 115/149/184 U ; 800
1200 78/124/195 2000 78/124/195 U ; 1200
2000 42/94/151 3000 42/94/151 U ; 2000
3000 43/30/80 4500 43/30/80 U ; 3000

B	white
F	44/26/76
N	white
EOF

gmt begin sizes
    gmt colorbar -Dx8c/1c+w12c/0.5c+jTC+h -L -Ccolors.cpt
gmt end show
```
sizes

This does not (when all labels are lower (L), the top most label still gets plotted on both the upper and lower bounds):

```
cat << EOF > colors.cpt
0	white	200 white U	; <200
200 230/230/188 300 230/230/188 L ; 200
300 195/203/160 500 195/203/160 L ; 300
500 155/175/172 800 155/175/172 L ; 500
800 115/149/184 1200 115/149/184 L ; 800
1200 78/124/195 2000 78/124/195 L ; 1200
2000 42/94/151 3000 42/94/151 L ; 2000
3000 43/30/80 4500 43/30/80 L ; 3000

B	white
F	44/26/76
N	white
EOF

gmt begin sizes
    gmt colorbar -Dx8c/1c+w12c/0.5c+jTC+h -L -Ccolors.cpt
gmt end show
```

sizes

PaulWessel:
Yes, the L|U|B code can interfere badly here.  But your case does not specify two labels for the last slice.  I guess in this case we should not place the second 3000 unless a second label is given.

meghanrjones:
> Yes, the L|U|B code can interfere badly here. But your case does not specify two labels for the last slice. I guess in this case we should not place the second 3000 unless a second label is given.

Yes, I would expect no second label. This bug existed before this PR, but since it is closely related it seems like the right time to fix it. But, let me know if you think it should be fixed in a separate PR.

PaulWessel:
No, makes sense in this PR.  I will see.  it will involve some more checking on the L|U|B code

PaulWessel:
I am not sure I agree with the OP's use of <200 and >4500 since but I think for these types of labels the code must be L on all and B on the last.  We have code that deals with the last slice end annotation but we do not have the reverse: dealing with the first slices bottom annotation before looping over slices.  So the solution is to just improve the docs on what code needs to be used for this case.  I.e., for non-categorical CPTs, use L for all slices and B fore last, and supply two labels for the last.  These are only used with -L.  Once you give -L an argument (i or a gap) then the first label is used , centered on the slice.

meghanrjones:
> I am not sure I agree with the OP's use of <200 and >4500 since but I think for these types of labels the code must be L on all and B on the last. We have code that deals with the last slice end annotation but we do not have the reverse: dealing with the first slices bottom annotation before looping over slices. So the solution is to just improve the docs on what code needs to be used for this case. I.e., for non-categorical CPTs, use L for all slices and B fore last, and supply two labels for the last. These are only used with -L. Once you give -L an argument (i or a gap) then the first label is used , centered on the slice.

You mean there should not be an option to leave the upper-most interval boundary un-annotated?

PaulWessel:
I think if the last slice has B then it must have two labels. Else we should not set the same label twice since that is clearly wrong.  We should never plot "3000" twice.

meghanrjones:
> I think if the last slice has B then it must have two labels. Else we should not set the same label twice since that is clearly wrong. We should never plot "3000" twice.

OK

PaulWessel:
OK, I have clarify the docs and simplified the code a bit.  No more duplicate 3000. Users are told to use code L for all slices except last, which must be U and have two labels.

meghanrjones (approved):
Looks good. Note that even though the cookbook states that the last slice must be set to **B**, the two intervals are drawn without warning if the last slice is set to **L** - the only difference is that the last tick mark is not drawn.

PaulWessel:
Right, I see that now.  Odd, I see the code drawing the ticks.  I cannot think of a reason why we would not want to draw the tick (this is when there is no -B).  Even if there is no annotation there I think the tick should be there, no?

PaulWessel:
The code is deciding to draw a longer annotation tick or a shorter frame tick, but for the last it is doing neither.

meghanrjones:
> The code is deciding to draw a longer annotation tick or a shorter frame tick, but for the last it is doing neither.

Yes, it's inconsistent because it still does an annotation tick on the bottom most tick when all labels are set to upper.

PaulWessel:
I need to do some more work as GMT_agefig.sh because the @age_chrons_GTS2012_2020 file used for age grids only has U flags and my code broke that....  It needs to check for just U as well, so I will need to revise my docs anyway.  Stay tuned.

PaulWessel:
OK done.  The age crhon CPT is good since it is all code U since they do now want to label the "zero chron" at 0 Ma, only the ones that follow.  So in their case the last slice only has U and that is fine.
In cases where you want to annotate the very first part then you need to use L everywhere except the last which needs B.

PR #4991, opened 2021-03-18
Adjust auto frame interval for themes

### LINKED ISSUE #4955: Potential issues or breaking changes with the GMT modern theme?

meghanrjones:

This PR relaxes the increase in annotation/frame/grid intervals from the gmt themes branch to minimize the instances in which 1 or fewer annotations are plotted on an axis.

For example:
``gmt basemap -JM6c -R-125/-121/47/48 -B -png map``

Before:
image
After:
image

Fixes #4955

**Reminders**


PaulWessel (approved):
I approve, but just checking/letting you know that the main reason that scale increased to 1.75 had to do with annotations in modern changing from 37 to 37N etc.  However, I do think there was more, hm, gut/guess involved than science in picking that factor.  It looked OK for some examples I looked at but clearing here we are because of it.  Looks like you checked lots of examples so I am happy with that.

meghanrjones:
> I approve, but just checking/letting you know that the main reason that scale increased to 1.75 had to do with annotations in modern changing from 37 to 37N etc. However, I do think there was more, hm, gut/guess involved than science in picking that factor. It looked OK for some examples I looked at but clearing here we are because of it. Looks like you checked lots of examples so I am happy with that.

Yes, I appreciated that there were comments in the code explaining that decision. In the PostScript files that needed updating with this change, I found that there were more that went from an insufficient number of annotations to a reasonable number than there were going from a good distance between annotation text to a bad distance. You are right, it's subjective. 

If you didn't see, PyGMT just implemented a separate 'data version control' for their test images. Maybe if that works well we can use it and then update the frame selections as much as we need without the downsides related to repo size - https://github.com/GenericMappingTools/pygmt/pull/1036.

PaulWessel:
We should discuss doing the same for GMT.  If I understand correctly it would limit the inclusion of large files into the GMT repo.

meghanrjones:
> We should discuss doing the same for GMT. If I understand correctly it would limit the inclusion of large files into the GMT repo.

Agreed. It will be good to see how it goes for PyGMT before putting effort into migrating the baseline images. So, we could discuss at the community meeting or in a GitHub issue around that time.

PR #4993, opened 2021-03-19
Allow subplot -D to compute correct layout without drawing frames

PaulWessel:
This is essential if we are to allow **subplot** to take an overlay from a later **subplot** command since they both need the same computed dimensions and spacings which are controlled by gmt.conf, **-B -C -M -S**.  Now, if you keep those settings the same but add -D then dimensions and gaps stay the same but frames are not drawn needlessly the second time.

Added a test to demonstrate overlay works.  Here the basemaps are drawn by one **subplot** command and then the second one only plots the data (circle at 4,3) and uses the same setups plus **-D**:

subplotoverlay

### DIFF
--- a/doc/rst/source/subplot.rst
+++ b/doc/rst/source/subplot.rst
@@ -125,7 +125,7 @@ Optional Arguments (begin mode)
 **-D**
     Use the prevailing defaults settings (via gmt.conf or **--PAR**\ =\ *value*) and the selections made
     via **-B**, **-C**, **-M** and **-S** to determine the panel sizes (if using **-Ff**) and panel spacings only, but
-    do *not* draw and annotated any frames.  This option is useful if you wish to lay down a partial subplot
+    do *not* draw and annotate any frames.  This option is useful if you wish to lay down a partial subplot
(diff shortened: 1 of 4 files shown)


doc/rst/source/subplot.rst
```diff
@@ -119,6 +120,16 @@ Optional Arguments (begin mode)
     to all subplots, while settings under **set** only apply to the selected (active) subplot.  **Note**: Common options **-X**
     and **-Y** are not available during subplots; use **-C** instead.
 
+.. _-D:
+
+**-D**
+    Use the prevailing defaults settings (via gmt.conf or **--PAR**\ =\ *value*) and the selections made
+    via **-B**, **-C**, **-M** and **-S** to determine the panel sizes (if using **-Ff**) and panel spacings only, but
+    do *not* draw and annotated any frames.  This option is useful if you wish to lay down a partial subplot
```
seisman, line 128:
```suggestion
    do *not* draw and annotate any frames.  This option is useful if you wish to lay down a partial subplot
```

seisman (approved):

meghanrjones (approved):

PR #5006, opened 2021-03-22
WIP Let a module given a single common option with no arg just print that option's usage message

PaulWessel:
So` gmt basemap -J` will print just the -J syntax and not give an error. same with `gmt xyz2grd -i`.
I left it at WIP since it is undocumented and should be discussed somewhere...


joa-quim (approved):

PaulWessel:
Where should this be documented, @meghanrjones?  Maybe at the end of the listing of common options?  E.g., a very quick way to see the specific syntax for one of the common options is to run a gmt module with just a single common option with no arguments.  While this would normally yield an error message, here it will produce the syntax for that option on the terminal. For example, gmt coast -B will print the usage message for the -B option."

meghanrjones:
Is the plan to eventually expand this to all options or will this only ever be for common options? If it's just for common options, I'm not really seeing why it's implemented to require calling a module that uses the given common option. I would think that `gmt --show-common -B` would be more consistent with the other parts of the command line documentation, in which case it could get listed with the other `gmt` options.

PaulWessel:
I agree it may make more sense in gmt.c, but then it will not be available to the externals (they do not use gmt.c).

meghanrjones:
> I agree it may make more sense in gmt.c, but then it will not be available to the externals (they do not use gmt.c).

OK, with the current implementation I think it could get documented here for the ReST documentation (in `doc/rst/source/gmt.rst`) stating that 'if *module-options* is a common option alone, gmt will report the usage description for that option'  after `Starts a given GMT module with the module-options that pertain to that particular module.`:

image

On the terminal, maybe it could be mentioned after `if <module-options> is '=' we call exit (0) if module exist and non-zero otherwise.` when calling gmt.

PaulWessel:
This branch is giving me nothing but **trouble**.  It is difficult to get it to work as intended for modern module names and classic names.  I am wasting my time and I suggest we abandon this branch.  Why not just

`gmt docs gmt -R`

if it is important to just see what -R looks like, @joa-quim ?

joa-quim:
Well, that is the same as asking why not just do ``gmt docs module`` instead of using the module's online help. But if things are giving more work than it worth just drop it, no problems.

PaulWessel:
Yes, it is causing lots of band-aid fixing to even make it work, and so far there are still corner cases.  SOme modules can run with just one option, say gmt convert -V and it would get caught in this and start printing -V syntax instead.  So lots of pitfalls.  I am pulling da plug.

PR #5007, opened 2021-03-23
Let grdcut take polygon and clip to polygon

### LINKED ISSUE #5004: grdcut: add option to cut grid based on polygon
Based on discussion in https://forum.generic-mapping-tools.org/t/how-to-trim-a-contour-map/1477.

It would be a nice addition to ``grdcut`` to be able to easily cut a grid based on an input polygon.

E.g.

```
gmt grdcut in.grd -Gout.nc -Fcut_polygon.gmt
```

PaulWessel:
The output grid will reflect the bounding box of the polygon (clipped to the input grid bounds if the polygon exceeds them) and all nodes outside the polygon will be set to NaN.  Closes #5004.  Need @anbj to give it a spin.  I tried

```
gmt coast -M -EES > spain.txt
gmt grdcut @earth_relief_10m -R-20/5/25/45 -Gt.grd
gmt grdcut t.grd -Fspain.txt -V -Gnew.grd
gmt grdimage new.grd -JM5i -Baf -pdf map

```
and lo and behold it worked on first test. But does not work directly with  @earth_relief_10m yet, hence WIP for now.

### DIFF
--- a/doc/rst/source/grdcut.rst
+++ b/doc/rst/source/grdcut.rst
@@ -12,8 +12,11 @@ Synopsis
 
 .. include:: common_SYN_OPTs.rst_
 
-**gmt grdcut** *ingrid* |-G|\ *outgrid*
+**gmt grdcut**
+*ingrid*
+|-G|\ *outgrid*
 |SYN_OPT-R|
+[ |-F|\ *polygonfile* ]
 [ |-J|\ *parameters* ]
 [ |-N|\ [*nodata*] ]
 [ |-S|\ *lon/lat/radius*\ [**+n**] ]
@@ -51,6 +54,13 @@ Required Arguments
 
 Optional Arguments
 ------------------
+
+.. _-F:
+
+**-F**\ *polygonfile*
+    Specify a multisegment closed polygon file.  The output grid region will
+    reflect the bounding box of the polygon and all nodes outside the
+    polygon will be set to NaN.
 
 .. _-J:
 
--- a/doc/rst/source/grdcut.rst
+++ b/doc/rst/source/grdcut.rst
@@ -16,7 +16,7 @@ Synopsis
 *ingrid*
 |-G|\ *outgrid*
 |SYN_OPT-R|
-[ |-F|\ *polygonfile*\ [**+c**] ]
+[ |-F|\ *polygonfile*\ [**+c**][**+i**] ]
 [ |-J|\ *parameters* ]
 [ |-N|\ [*nodata*] ]
 [ |-S|\ *lon/lat/radius*\ [**+n**] ]
@@ -57,10 +57,11 @@ Optional Arguments
 
 .. _-F:
 
-**-F**\ *polygonfile*\ [**+c**]
+**-F**\ *polygonfile*\ [**+c**][**+i**]
     Specify a multisegment closed polygon file.  All grid nodes outside the
-    polygon will be set to NaN.  Optionally, append **+c** to crop the grid
-    region to reflect the bounding box of the polygon.
+    polygon will be set to NaN.  Append **+i** to invert that and set all
+    nodes inside the polygon to NaN instead. Optionally, append **+c** to
+    crop the grid region to reflect the bounding box of the polygon.
 
 .. _-J:
 
--- a/doc/rst/source/grdcut.rst
+++ b/doc/rst/source/grdcut.rst
@@ -154,6 +154,13 @@ distance of 500 km from the point 45,30 try::
 
     gmt grdcut bathy.nc -Gsubset_bathy.nc -S45/30/500k -V
 
+To create a topography grid with data only inside France and set it
+to NaN outside France, based on the 10x10 minute DEM, try::
+
+    gmt coast -EFR -M > FR.txt
+    gmt grdcut @earth_relief_10m -FFR.txt+c -GFR_only.grd
+    gmt grdimage FR_only.grd -B -pdf map
+    
 See Also
 --------
 
--- a/doc/rst/source/grdcut.rst
+++ b/doc/rst/source/grdcut.rst
@@ -16,7 +16,7 @@ Synopsis
 *ingrid*
 |-G|\ *outgrid*
 |SYN_OPT-R|
-[ |-F|\ *polygonfile* ]
+[ |-F|\ *polygonfile*\ [**+c**] ]
 [ |-J|\ *parameters* ]
 [ |-N|\ [*nodata*] ]
 [ |-S|\ *lon/lat/radius*\ [**+n**] ]
@@ -57,10 +57,10 @@ Optional Arguments
 
 .. _-F:
 
-**-F**\ *polygonfile*
-    Specify a multisegment closed polygon file.  The output grid region will
-    reflect the bounding box of the polygon and all nodes outside the
-    polygon will be set to NaN.
+**-F**\ *polygonfile*\ [**+c**]
+    Specify a multisegment closed polygon file.  All grid nodes outside the
+    polygon will be set to NaN.  Optionally, append **+c** to crop the grid
+    region to reflect the bounding box of the polygon.
 
 .. _-J:
 
--- a/doc/rst/source/supplements/module_supplements_purpose.rst_
+++ b/doc/rst/source/supplements/module_supplements_purpose.rst_
@@ -4,9 +4,9 @@
 
 .. |gpsgridder_purpose| replace:: Interpolate GPS strains using Green's functions for elastic deformation
 
-.. |velo_purpose| replace:: Plot velocity vectors, crosses, and wedges
+.. |velo_purpose| replace:: Plot velocity vectors, crosses, anisotropy bars and wedges
 
-.. |psvelo_purpose| replace:: Plot velocity vectors, crosses, and wedges
+.. |psvelo_purpose| replace:: Plot velocity vectors, crosses, anisotropy bars and wedges
 
 .. |gshhg_purpose| replace:: Extract data tables from binary GSHHG or WDBII data files
 
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12897,6 +12897,13 @@ struct GMT_RESOURCE * GMT_Encode_Options (void *V_API, const char *module_name, 
 		else info[ku].pos -= nn[info[ku].direction][K_SECONDARY];	/* Move any primary objects to start of list for this direction */
 	}
 	GMT_Report (API, GMT_MSG_DEBUG, "GMT_Encode_Options: Found %d inputs and %d outputs that need memory hook-up\n", input_pos, output_pos);
+#if 0
+	{	/* Left here for simple debugging with messages on the GMT side */
+		text = GMT_Create_Cmd (API, *head);
+		GMT_Report (API, GMT_MSG_NOTICE, "GMT_Encode_Options: Revised command before memory-substitution: %s\n", text);
+		GMT_Destroy_Cmd (API, &text);
+	}
+#endif
 	/* Just checking that the options were properly processed */
 	if (gmt_M_is_verbose (API->GMT, GMT_MSG_DEBUG)) {
 		static char *omode[2] = {"Primary", "Secondary"};
--- a/src/gmt_grdio.c
+++ b/src/gmt_grdio.c
@@ -828,7 +828,8 @@ void gmtlib_grd_get_units (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *header)
 		/* Change name of variable and unit to lower case for comparison */
 		gmt_str_tolower (string[i]);
 
-		if ((!strncmp (string[i], "longitude", 9U) || strstr (string[i], "degrees_e")) && (header->wesn[XLO] > -360.0 && header->wesn[XHI] <= 360.0)) {
+		/* PW: This test was <= 360 but if you have grids from 340-375 and it says longitude then we should not make it Cartesian */
+		if ((!strncmp (string[i], "longitude", 9U) || strstr (string[i], "degrees_e")) && (header->wesn[XLO] > -360.0 && header->wesn[XHI] < 720.0)) {
 			/* Input data type is longitude */
 			gmt_set_column_type (GMT, GMT_IN, i, GMT_IS_LON);
 		}
--- a/src/grdcut.c
+++ b/src/grdcut.c
@@ -42,6 +42,10 @@ struct GRDCUT_CTRL {
 		bool active;
 		char *file;
 	} In;
+	struct GRDCUT_F {	/* -Fpolfile */
+		bool active;
+		char *file;
+	} F;
 	struct GRDCUT_G {	/* -G<output_grdfile> */
 		bool active;
 		char *file;
@@ -85,6 +89,7 @@ static void *New_Ctrl (struct GMT_CTRL *GMT) {	/* Allocate and initialize a new 
 static void Free_Ctrl (struct GMT_CTRL *GMT, struct GRDCUT_CTRL *C) {	/* Deallocate control structure */
 	if (!C) return;
 	gmt_M_str_free (C->In.file);
+	gmt_M_str_free (C->F.file);
 	gmt_M_str_free (C->G.file);
 	gmt_M_free (GMT, C);
 }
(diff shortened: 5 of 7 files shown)


anbj:
Aloha Paul,

Why problem with ``@earth_relief_10m``? I tried that as well, and gave same results as with ``@earth_relief_30m`` below.

```
gmt grdcut -Rg @earth_relief_30m -Gworld.nc
gmt pscoast -ENO -M > no.gmt
gmt grdcut -Fno.gmt -Gpolycut.nc world.nc 
```
Works! 

But; observe x-range - 0-360. Y-range is a bit off; y_max: 90 seems a bit too much. Norway cannot claim the North Pole.

```
gmt grdinfo polycut.nc 
polycut.nc: Title: Produced by grdcut
polycut.nc: Command: grdcut -Fno.gmt -Gpolycut.nc world.nc
polycut.nc: Remark: Obtained by Gaussian Cartesian filtering (55.6 km fullwidth) from SRTM15+V2.1.nc [Tozer et al., 2019; http://dx.doi.org/10.1029/2019EA000658]
polycut.nc: Pixel node registration used [Geographic grid]
polycut.nc: Grid file format: nf = GMT netCDF format (32-bit float), CF-1.7
polycut.nc: x_min: 0 x_max: 360 x_inc: 0.5 (30 min) name: longitude n_columns: 720
polycut.nc: y_min: 57.5 y_max: 90 y_inc: 0.5 (30 min) name: latitude n_rows: 65
polycut.nc: v_min: -111.5 v_max: 1409.5 name: elevation (m)
polycut.nc: scale_factor: 1 add_offset: 0
polycut.nc: format: netCDF-4 chunk_size: 360,65 shuffle: on deflation_level: 3
```

Tidbit(?)

Due to QGIS requiring proper header, I injected some OGR-stuff;

```
sed -i '1i # @VGMT1.0' no.gmt

gmt grdcut -Fno.gmt -Gpolycut.nc world.nc 
grdcut [ERROR]: OGR parsing incomplete (is file missing OGR statements?) - abort
/usr/local/lib/libgmt.so.6(GMT_grdcut+0x561)[0x7fb96fbf6ef1]
[0x0]
Stack backtrace:
/usr/local/lib/libgmt.so.6(sig_handler+0x26e)[0x7fb96f9962ce]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12730)[0x7fb96f449730]
/usr/local/lib/libgmt.so.6(GMT_grdcut+0x561)[0x7fb96fbf6ef1]
/usr/local/lib/libgmt.so.6(GMT_Call_Module+0x3ee)[0x7fb96f9b489e]
gmt(main+0x260)[0x558ffccf2430]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb)[0x7fb96cf3a09b]
gmt(_start+0x2a)[0x558ffccf34aa]
```

Alright let's add ``@G`` to indicate that it's a ``MULTIPOLYGON`` :

```
sed -i '1d' no.gmt
sed -i '1i # @VGMT1.0 @GMULTIPOLYGON' no.gmt
gmt grdcut -Fno.gmt -Gpolycut.nc world.nc
```

And that works.

joa-quim:
I think the default should not be to clip to the polygon's BoundingBox. Probably more often then not we don't want the grid size to change. But there are other situations where that may be wished. So I think the default should be to keep the original **-R**egion but optionally clip to polygon BB.

PaulWessel:
Thanks @anbj - should be fixed now.  I had a cut/paste issue on MIN/MAX and I did not bother to be careful with the longitudes.  Works for me now.

PaulWessel:
I agree with you @joa-quim.  Will think of a mechanism to allow that option. Might make most sense to give **-F** a modifier **+c** for crop grid to polygon bounding box [leave region alone].

PaulWessel:
OK, **-F**_polfile_[**+c**] is implemented and tested like this:

```
gmt coast -M -EES > spain.txt
gmt grdcut @earth_relief_30m -Fspain.txt -Gt.grd
gmt grdimage t.grd -JQ5i -Baf -pdf t

gmt grdcut @earth_relief_30m -Fspain.txt+c -Gt.grd
gmt grdimage t.grd -JQ5i -Baf -pdf t
```

joa-quim:
Yes, something around those lines, though I confess that I don't like **-F**_+crime_ options since the modifier character looks like it's part of the name. The alternative is to put the **+c** at the end.
Also, needs new KEYs for **-F**

PaulWessel:
Will do KEYS.  Modifier is at end of course.  -Fspain.txt+c.

joa-quim:
You also forgot one your favorites. **+i** to invert the sense of the clipping

PaulWessel:
I did forget!  OK, on it.

PaulWessel:
OK, please give it a spin with the above commands and add +i.  Works fine for me.

joa-quim:
```
gmt grdcut @earth_relief_30m -Fspain.txt+c+i -Gt.grd
grdcut [ERROR]: Must specify only one of the -R, -S or the -Z options
```

PaulWessel:
Well, you do have to build the branch, no?

joa-quim:
Yep, I rebuilt but forgot to update after changing the branch. It's working now.

PaulWessel:
I think only one of -R or +c can be given else an error. I can add that.  There is no conceivable reason to give both.

joa-quim:
Yes **-R** and **+c** are self exclusive but **-i** not
```
gmt grdcut @earth_relief_30m -Fspain.txt+i -Gt.grd -R-10/3/36/47
grdcut [ERROR]: Must specify only one of the -F, -R, -S or the -Z options
```

PaulWessel:
Should work with -R +i now.

joa-quim:
But I'm getting this error in Julia coming from the **-F** parsing in ``grdcut``, from the line

if (GMT_Get_FilePath (GMT->parent, GMT_IS_DATASET, GMT_IN, GMT_FILE_REMOTE, &(Ctrl->F.file))) n_errors++;

```
julia> G = grdcut("@earth_relief_30m", clip=(polygon=ES, crop2cutline=true, invert=true),  Vd=1);
        grdcut @earth_relief_30m  -F+c+i
grdcut [ERROR]: No filename provided
```
and yes, no file name provided because ``ES`` is a GMTdataset
```
ES = coast(M=true, E=:ES);
```

PaulWessel:
But presumably this results in a virtual file that holds the ES polygon and you pass it via -F.  grdcut does not know anything about ES.

joa-quim:
Yes, it should be a virtual file name, not an empty one

image

joa-quim:
The -F<file> must be read with ``GMT_Read_Data()`` so that virtual files are recognized, no?

PaulWessel:
Yes, the F->filename is passed got GMT_Read_Data so if it is a virtual file it should all work as is I think.

joa-quim:
No, it's not using GMT_Read_Data() and it's not working
```
			case 'F':
				Ctrl->F.active = true;
				if (opt->arg[0]) {
					if ((c = gmt_first_modifier (GMT, opt->arg, "ci"))) {	/* Process any modifiers */
						unsigned int pos = 0;	/* Reset to start of new word */
						char p[PATH_MAX] = {""};
						while (gmt_getmodopt (GMT, 'F', c, "ci", &pos, p, &n_errors) && n_errors == 0) {
							switch (p[0]) {
								case 'c': Ctrl->F.crop = true; break;
								case 'i': Ctrl->F.invert = 1; break;
								default: break;	/* These are caught in gmt_getmodopt so break is just for Coverity */
							}
						}
						c[0] = '\0';	/* Chop off all modifiers so range can be determined */
					}
					Ctrl->F.file = strdup (opt->arg);
					if (c) c[0] = '+';	/* Restore modifier */
					if (GMT_Get_FilePath (GMT->parent, GMT_IS_DATASET, GMT_IN, GMT_FILE_REMOTE, &(Ctrl->F.file))) n_errors++;
				}
```

PaulWessel:
But there is no reading there.  It is here:

```
	else if (Ctrl->F.active) {	/* ONly read in subset corresponding to the bounding box of the polygon */
		if ((G = GMT_Read_Data (API, GMT_IS_GRID, GMT_IS_FILE, GMT_IS_SURFACE, GMT_CONTAINER_ONLY, NULL, Ctrl->In.file, NULL)) == NULL) {
			Return (API->error);	/* Get header only */
		}
		if ((D = GMT_Read_Data (API, GMT_IS_DATASET, GMT_IS_FILE, GMT_IS_POLY, GMT_READ_NORMAL, NULL, Ctrl->F.file, NULL)) == NULL) {
			Return (API->error);
		}
```

joa-quim:
But there is the checking in GMT_Get_FilePath() like I showed you with the debugger screen capture.

PaulWessel:
Yes, but if it sees a virtual file name it does not try to do stuff.  We have those same GMT_Get_FilePath everywhere,

PaulWessel:
It has

`	if (gmt_M_file_is_memory (file)) return GMT_NOERROR;	/* Memory files are always fine */
`

joa-quim:
That line comes after the line where it errored but the ``file == ""`` which is the true source of the problem.

PaulWessel:
Maybe move the memory check line ahead of that - it has its own check for file not being NULL.  Please try.

joa-quim:
Can't help ``file = ""`` and would fail in the ``gmt_M_file_is_memory ()`` macro too. And to help with my MEX is broken, probably some change in gmt.dll

PaulWessel:
OK, so the problem lies on the external size, not getting and placing the virtual file in the -F option before passing to Call_Module.  Hm, there is nothing special about this -F that is different from, say, the -F in gmtselect or grdrotator.  Meaning, there is no special case in GMT_Encode_Options to deal with anything unique here.  Wait, since -F is not a primary input, don't you need to pass -F? to have that expanded?

joa-quim:
No, I have many other like this. Never need any ``?``. But BTW, this works
```
julia> G = grdcut("@earth_relief_30m", clip=ES,  Vd=1);
        grdcut @earth_relief_30m  -F
```
So its the presence of the modifiers that makes it screw. Don't remember if I have any other case like this (but I would say yes)

Regarding the MEX. I see that it has an import symbol called ``GMT_get_V`` but I can't find any function with that name in current GMT. Gone with the wind?

joa-quim:
> Regarding the MEX. I see that it has an import symbol called GMT_get_V but I can't find any function with that name in current GMT. Gone with the wind?

Yes, it was renamed ``to gmt_get_V`` and that broke the ``gmtmex``

PaulWessel:
OK, I will build the mex.  I am not sure what happened to GMT_get_V, so will look at that too.  I know we handle things like -L<code><file> like in surface, but not sure if we have cases like -F<file>+mods and how we handle that.

PaulWessel:
Will fix the mex to handle the old name.  Should never have been called GMT_ anyway...

PaulWessel:
Good timing since I just installed Matlab 2021a so we will see if it works.

joa-quim:
I would rather prefer that GMT maintain a  GMT_get_V alias to not breaking the. These type of breaking are common in unix and that's why compatibilities don't last more than 6 months. On Windows they last for decades.

PaulWessel:
Meanwhile, with latest matlab and latest xcode I cannot select the process to connect to....  So need to try the previous matlab and possibly the previous xcode.  Sigh.

joa-quim:
See what I said about compatibility and 6 months?

PaulWessel:
Re-learing about KEYS and the = and =f variants...

PaulWessel:
Alright @joa-quim I figured out that the keys for -Ffile[+mod] needs to be **FD(=**.  With that I got this to run fine:

```
D = gmt ('read', '-Td spain.txt'); 
G = gmt ('grdcut', '@earth_relief_30m -F+c', D);
```

If you want to build the mex, please update the gmtmex to deal with the get_V stuff.

PaulWessel:
Should work well for @anbj on the command line now.

anbj:
Will do testing.

One thing though, as mentioned above, about the OGR-related stuff.

If you add the OGR/GMT header, you get a crash:

```
gmt grdcut -Rg @earth_relief_30m -Gworld.nc           <--------- prepare a test grid

gmt pscoast -ENO -M > no.gmt                                   <--------- prepare a test polygon

sed -i '1i # @VGMT1.0' no.gmt                                   <---------- add minimal OGR header for QGIS compatability

$ gmt grdcut -Fno.gmt -Gpolycut.nc world.nc
grdcut [ERROR]: OGR parsing incomplete (is file missing OGR statements?) - abort
/usr/local/lib/libgmt.so.6(GMT_grdcut+0x561)[0x7fecf4bc5ef1]
[0x0]
Stack backtrace:
/usr/local/lib/libgmt.so.6(sig_handler+0x26e)[0x7fecf49652ce]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12730)[0x7fecf4418730]
/usr/local/lib/libgmt.so.6(GMT_grdcut+0x561)[0x7fecf4bc5ef1]
/usr/local/lib/libgmt.so.6(GMT_Call_Module+0x3ee)[0x7fecf498389e]
gmt(main+0x260)[0x56155414f430]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb)[0x7fecf1f0909b]
gmt(_start+0x2a)[0x5615541504aa]
```

Remove the header, and replace it with one  which also defines what geometry type is in the file and it works:

```
sed -i '1d' no.gmt                                                                  <-------- remove the header added above
sed -i '1i # @VGMT1.0 @GMULTIPOLYGON' no.gmt        <-------- add new header, including geometry
gmt grdcut -Fno.gmt -Gpolycut.nc world.nc
```

joa-quim (approved):
Works fine now in Julia too

PaulWessel:
Great.  Adding a small test then will merge.

PaulWessel:
> 1. I'm having problems using `-F[pol]+c`. Following your example char by char, Paul:
> 
> ```
> gmt coast -M -EES > spain.txt
> gmt grdcut @earth_relief_30m -Fspain.txt -Gt.grd
> gmt grdcut @earth_relief_30m -Fspain.txt+c -Gt.grd
> grdcut [ERROR]: Cannot find file spain.txt+c
> ```
> 
> 1. Adding `-F[pol]+i` does not have any apparent effect. File size differs by two bytes though:
> 
> ```
> gmt grdcut @earth_relief_30m -Fspain.txt -G1.grd
> gmt grdcut @earth_relief_30m -Fspain.txt+i -G2.grd
> 
> ls -l | grep -e 1.grd -e 2.grd
> -rw-r--r--  1 anbj anbj   20388 Mar 24 09:03 1.grd
> -rw-r--r--  1 anbj anbj   20390 Mar 24 09:03 2.grd
> ```
> 
> 1. Not directly related, but still (as mentioned above):
> 
> If you add the OGR/GMT header, you get a crash:
> 
> ```
> gmt grdcut -Rg @earth_relief_30m -Gworld.nc       <--------- prepare a test grid
> 
> gmt pscoast -ENO -M > no.gmt                      <--------- prepare a test polygon
> 
> sed -i '1i # @VGMT1.0' no.gmt                     <---------- add minimal OGR header for QGIS compatibility
> 
> $ gmt grdcut -Fno.gmt -Gpolycut.nc world.nc
> grdcut [ERROR]: OGR parsing incomplete (is file missing OGR statements?) - abort     <-------- Look at this!
> ERROR: Caught signal number 11 (Segmentation fault) at                               <-------- muchos problemos
> /usr/local/lib/libgmt.so.6(GMT_grdcut+0x561)[0x7fecf4bc5ef1]
> [0x0]
> Stack backtrace:
> /usr/local/lib/libgmt.so.6(sig_handler+0x26e)[0x7fecf49652ce]
> /lib/x86_64-linux-gnu/libpthread.so.0(+0x12730)[0x7fecf4418730]
> /usr/local/lib/libgmt.so.6(GMT_grdcut+0x561)[0x7fecf4bc5ef1]
> /usr/local/lib/libgmt.so.6(GMT_Call_Module+0x3ee)[0x7fecf498389e]
> gmt(main+0x260)[0x56155414f430]
> /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb)[0x7fecf1f0909b]
> gmt(_start+0x2a)[0x5615541504aa]
> ```
> 
> Remove the header, and replace it with one which also defines what geometry type is in the file and it works:
> 
> ```
> sed -i '1d' no.gmt                                <-------- remove the header added above
> sed -i '1i # @VGMT1.0 @GMULTIPOLYGON' no.gmt      <-------- add new header, including geometry
> gmt grdcut -Fno.gmt -Gpolycut.nc world.nc         <--------- nada problemos
> ```

I think you need to update the branch and build again.  The +c +i stuff works across all platforms now.

The OGR stuff: First , nothing specific to this branch, but second, you are making a broken OGR file and GMT complaints.  Is your issue that it should gracefully quit and not crash?

anbj:
> I think you need to update the branch and build again. The +c +i stuff works across all platforms now.

That might have been the case. I'm sorry!

> The OGR stuff: First , nothing specific to this branch, but second, you are making a broken OGR file and GMT complaints. Is your issue that it should gracefully quit and not crash?

Yes. I believe Joaquim once said that all crashes aare bugs. A graceful exit is what I'm looking for.

PR #4996, opened 2021-03-19
Consolidate source text for cookbook and manpage -R docs

meghanrjones:
**Description of proposed changes**

This PR makes a few changes to the documentation:

**Fixes links in gmt.rst**
This PR adds references for each option formatted like ``.. _-B:`` in gmt.rst so that these links work (https://docs.generic-mapping-tools.org/dev/gmt.html#the-common-gmt-options)

**Consolidates the cookbook and the gmt manpage -R documentation into one ReST file** (`explain_-R.rst_`) which is then included in both `gmt.rst` and `cookbook/options.rst`

**Formats the gmt manpage -R documentation (`gmt.rst`) as a list**

> image

**Changes the format for links to more information in the module docs**

> *Before*
> 
> image

> *After*
> 
> image

**Combines ``explain_-R.rst_`` and ``explain_-R_full.rst_``**
Rather than having separate documents, the modules now use the include directive's start-after and end-before options to specify which information to include.

**Moves ``.. _-*_full:`` references from the individual ``_explain_-*_full.rst_`` to ``gmt.rst``**
This will make future updates easier.

Fixes #4917 
Relates to #4969

**Reminders**

- [ ] Make sure that your code follows our style. Use the other functions/files as a basis.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Describe changes to function behavior and arguments in a comment below the function declaration.
- [ ] If adding new functionality, add a detailed description to the documentation and/or an example.

Issue #4969, opened 2021-03-15
Consolidating the source text for common options documentation


meghanrjones:
**Description of the desired feature**

*Originally discussed at https://github.com/GenericMappingTools/gmt/pull/4918*

The GMT common options are documented in the [gmt manpage](https://docs.generic-mapping-tools.org/dev/gmt.html#the-common-gmt-options) and the cookbook [standardized command line options chapter](https://docs.generic-mapping-tools.org/dev/cookbook/options.html#). The source files for the common options documentation should be consolidated due to the amount of redundant information in these files, which can lead to errors when adding new features or enhancements. 

Here is the current structure of these two documentation sources:
- GMT manpage - primary source is `doc/rst/source/gmt.rst`, which has include directives for separate ReST files for each standard option (named `doc/rst/source/explain_-*.rst_`).
- GMT cookbook - all documentation is located in `doc/rst/source/cookbook/options.rst`

Here are two options for consolidating the documentation:

**Option 1**
Move the **syntax** and **description** (without examples) sections from `doc/rst/source/cookbook/options.rst` to the `explain_-*.rst_` files and then include those files both in the cookbook and the gmt manpage.

*Benefits of this method*
- No need for PyGMT to change their documentation (which currently links to sections in https://docs.generic-mapping-tools.org/dev/gmt.html)
- Users that are bandwidth limited can view https://docs.generic-mapping-tools.org/dev/gmt.html for predominantly text-based information (missing examples; comparatively limited documentation for **-J**).

**Option 2**
Only include a simple table with the common options syntax in https://docs.generic-mapping-tools.org/dev/gmt.html that links to full options in cookbook

*Benefits of this method*
- Simpler documentation structure (could encourage new collaborators)
- 39 fewer files in `doc/rst/source`

For some context about the data usage discussion, here's the amount transferred when loading the various pages:
- https://docs.generic-mapping-tools.org/dev/gmt.html#the-common-gmt-options (750 KB)
- https://docs.generic-mapping-tools.org/dev/cookbook/options.html (1.7 MB)
- https://docs.generic-mapping-tools.org/dev/std-opts.html (640 KB)

**Are you willing to help implement and maintain this feature?** Yes

meghanrjones:
[...]

PR #5012, opened 2021-03-24
Consolidate source text for cookbook and manpage -U docs

meghanrjones:

This PR consolidates the source documentation for -U for the cookbook and reference docs (similar to -R in #4996).

Addresses #4969

Fixes #

**Reminders**


PaulWessel (approved):
Looks good. However, I noted a few things: 

1. As I read the **-U** docs, it is a bit unclear that **+c** is used instead of giving a label.  I think it is actually more like _label_ OR **+c**, and in old GMT it was _label_|**c** I think.  So maybe this can be made clearer, and maybe we should use **-U**[_label_|**+c**][...]
2. Peaking below (and for another PR), the **-a** seems to require **-a**_name_ if I remove the optional [] stuff.

meghanrjones:
> Looks good. However, I noted a few things:
> 
>     1. As I read the **-U** docs, it is a bit unclear that **+c** is used instead of giving a label.  I think it is actually more like _label_ OR **+c**, and in old GMT it was _label_|**c** I think.  So maybe this can be made clearer, and maybe we should use **-U**[_label_|**+c**][...]
> 
>     2. Peaking below (and for another PR), the **-a** seems to require **-a**_name_ if I remove the optional [] stuff.

Both fixed in https://github.com/GenericMappingTools/gmt/pull/5012/commits/21551e92a208e32ced4f1e2776408f6ff1f3a12d

image

image

PaulWessel:
Great, thanks.

PR #5014, opened 2021-03-24
Clarify interplay of -I and -C

### LINKED ISSUE #4938: Misleading description for the -I argument of gmt info

I find the description for **-I** argument of `gmt info` may be a little misleading for new GMT users (originally posted at https://github.com/GenericMappingTools/pygmt/issues/1035).

It seems that only when the **-C** argument is combined with **-I**, then the output will be rounded up/down for as many columns as there are increments provided in **-I**. Without **-C**, **-I** only outputs the min/max values for the first two columns no matter how many increments we input. So, we'd better always add **-C** when we use **-I** unless only the first two columns are needed.

---

**Test**

I use [Iris flower dataset](https://en.wikipedia.org/wiki/Iris_flower_data_set) downloaded from https://github.com/mwaskom/seaborn-data/raw/master/iris.csv to test **-I** and **-C**.

```
# Download the sample iris data:
$ wget https://github.com/mwaskom/seaborn-data/raw/master/iris.csv
```

```
# Output with -I and -C:
$ awk -F "," 'FNR>1 {print $4,$1,$3}' iris.csv | gmt info -C -I1/0.2/0.3
0	3	4.2	8	0.9	7.2

# Output with only -I:
$ awk -F "," 'FNR>1 {print $4,$1,$3}' iris.csv | gmt info -I1/0.2/0.3   
-R0/3/4.2/8

# Output with only -C:
$ awk -F "," 'FNR>1 {print $4,$1,$3}' iris.csv | gmt info -C            
0.1	2.5	4.3	7.9	1	6.9

# Output with neither -I nor -C:
$ awk -F "," 'FNR>1 {print $4,$1,$3}' iris.csv | gmt info   
<Standard Input>: N = 150	<0.1/2.5>	<4.3/7.9>	<1/6.9>
```

---

**GMT DOC**

1. The description of `gmt info` at [GMT DOC](https://docs.generic-mapping-tools.org/latest/gmtinfo.html#description) :
> As another option, info can find the extent of data in the first two columns rounded up and down to the nearest multiple of the supplied increments given by -I.

This is right. When we only use **-I** without **-C**, only the rounded min/max values of the first columns will be output.

> If -C is combined with -I then the output will be in column form and rounded up/down for as many columns as there are increments provided in -I.

This is exactly right. When we use both **-I** and **-C**, we can output the rounded min/max values for each column provided in **-I**.

2. The detailed description of **-I** at [GMT DOC](https://docs.generic-mapping-tools.org/latest/gmtinfo.html#i):
> Report the min/max of the first n columns to the nearest multiple of the provided increments (separate the n increments by slashes), and output results in the form -Rw/e/s/n (unless -C is set). If only one increment is given we also use it for the second column (for backwards compatibility).

The first part of the first sentence assumes **-C** is used, while the second part assumes **-C** is not used.

---

**Suggestion**

It's better to revise the description of **-I** option at [GMT DOC](https://docs.generic-mapping-tools.org/latest/gmtinfo.html#i) to explicitly indicate the difference for the two cases, i.e., **-C** is used and is not used. I'd like to help this issue if you think we need to revise a little about **-I**.

**System information**

* Operating system: CentOS 7

PaulWessel:
Closes #4938.

### DIFF
--- a/doc/rst/source/gmtinfo.rst
+++ b/doc/rst/source/gmtinfo.rst
@@ -111,10 +111,11 @@ Optional Arguments
 .. _-I:
 
 **-I**\ [**b**\|\ **e**\|\ **f**\|\ **p**\|\ **s**]\ *dx*\ [/*dy*\ [/*dz*...][**+e**\|\ **r**\|\ **R**]
-    Report the min/max of the first *n* columns to the nearest multiple
-    of the provided increments (separate the *n* increments by slashes),
-    and output results in the form **-R**\ *w/e/s/n* (unless **-C** is
-    set). If only one increment is given we also use it for the second
+    Compute the *min*\ /*max* values of the first *n* columns to the nearest multiple
+    of the provided increments (separate the *n* increments by slashes) [2].
+    By default, output results in the form **-R**\ *w/e/s/n*, unless **-C** is
+    set in which case we output each *min* and *max* value in separate output columns.
+    If only one increment is given we also use it for the second
     column (for backwards compatibility). To override this behavior, use
     **-Ip**\ *dx*. If the input *x*- and *y*-coordinates all have the
     same phase shift relative to the *dx* and *dy* increments then we
--- a/doc/rst/source/gmtinfo.rst
+++ b/doc/rst/source/gmtinfo.rst
@@ -112,7 +112,7 @@ Optional Arguments
 
 **-I**\ [**b**\|\ **e**\|\ **f**\|\ **p**\|\ **s**]\ *dx*\ [/*dy*\ [/*dz*...][**+e**\|\ **r**\|\ **R**]
     Compute the *min*\ /*max* values of the first *n* columns to the nearest multiple
-    of the provided increments (separate the *n* increments by slashes) [2].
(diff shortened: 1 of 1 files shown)


doc/rst/source/gmtinfo.rst
```diff
@@ -111,10 +111,11 @@ Optional Arguments
 .. _-I:
 
 **-I**\ [**b**\|\ **e**\|\ **f**\|\ **p**\|\ **s**]\ *dx*\ [/*dy*\ [/*dz*...][**+e**\|\ **r**\|\ **R**]
-    Report the min/max of the first *n* columns to the nearest multiple
-    of the provided increments (separate the *n* increments by slashes),
-    and output results in the form **-R**\ *w/e/s/n* (unless **-C** is
-    set). If only one increment is given we also use it for the second
+    Compute the *min*\ /*max* values of the first *n* columns to the nearest multiple
+    of the provided increments (separate the *n* increments by slashes) [2].
```
meghanrjones, line 115:
```suggestion
    of the provided increments (separate the *n* increments by slashes) [default is 2 columns].
```

meghanrjones (approved):

PR #5015, opened 2021-03-24
Force projection for global images when rectangular projections are selected

### LINKED ISSUE #4814: Using Jq not at meridian 0 with earth_day didn’t work properly
Using Jq not at meridian 0 with earth_day didn’t work properly.

```
gmt begin Jq_earth_day png
gmt grdimage -Jq-60/0.0625c -Rd -Ba30 -BNSEW @earth_day_01d
gmt coast -Wthin,red
gmt end
```

**Actual outcome**
Jq_earth_day

**Expected outcome**
Replacing earth_day by earth_relief, then the image is centered at meridian -60
Jq_earth_relief
Replacing Jq by Jy works for earth_day and earth_relief.

**System information**

* Operating system: Ubuntu 20.04.2 LTS
* GMT version (`gmt --version`): 6.2.0_50eafd2_2021.02.10

PaulWessel:
See #4814 for background.  The problem is that when a rectangular projection is selected then there is no need to reproject a grid to plot it.  However, for images the reading does not shuffle the region, hence we do need to run it through the projection step.  Perhaps this may change in the future, but for now this catches those cases.  THis is why it worked when the OP specified the projection explicitly.  And it works for other projections like **-Jh** etc. All tests passes as before.  Closes #4814.

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -1257,6 +1257,10 @@ EXTERN_MSC int GMT_grdimage (void *V_API, int mode, void *args) {
 #endif
 
 	if (Ctrl->D.active) {	/* Main input is a single image and not a grid */
+		if (I && !need_to_project) {
+			gmt_M_memcpy (wesn, GMT->common.R.active[RSET] ? GMT->common.R.wesn : I->header->wesn, 4, double);
+			need_to_project = (gmt_whole_earth (GMT, I->header->wesn, wesn));	/* Must project global images even if not needed for grids */
+		}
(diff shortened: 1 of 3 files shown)


meghanrjones:
Is a test needed here?

PaulWessel:
yes that is a good idea of course.  I will hack one up from the issue.

PaulWessel:
OK, added the test for this issue.

meghanrjones (approved):

PR #5018, opened 2021-03-25
Prevent makecpt crash due to lack of nodes

PaulWessel:

If iT is used with _inc_**+n** to indicate we want _inc_ points along the z-axis, we must check that this results in at least 2 nodes since we need that many to form a single CPT slice.  Before this PR, commands like

gmt makecpt -Cjet -Z -T0/100/1+n

would result in a crash because downstream it was assumed there surely would be at least 2 nodes.  I have plugged that crash-site and now added a test in **makecpt** itself to stop right there if n < 2.


meghanrjones (commented):
...so many failures

PaulWessel:
Sorry, you need to update - I fixed a stupid thing.  My test now only gave me one (the new one from earlier today) that may not be relevant.

meghanrjones:
Is the change pushed? git says I'm up to date

PaulWessel:
Ooops. now.

meghanrjones (approved):

Issue #4367, opened 2020-10-24
Vertical exaggeration for Profiles


Esteban82:
This is a feature **suggest** in order to improve profiles. I have my commands to do it (in DOS, sorry, I will change to bash soon).

1. When I run basemap/plot the map scale is inform in the terminal. But in the case of a profile (-JX), that scale is the horizontal one and may NOT be the same as the vertical scale.

2. From the horizontal and vertical scale, a vertical exaggeration can be calculated.

Suggestions: The vertical scale and exaggeration could be shown in the terminal. 

Is it possible to save that information (horizontal and vertical scales, and vertical exaggeration) so it can be plot in the figure? I have to do the maths, show the results in the terminal and them plot it with:

```
REM	Factor for the Horizontal (FH) and Vertical (FV) to convert the graphic units (cm) to reals (m, km). 
REM	km-cm=100000, m-cm=100
	SET	FH=100000
	SET	FV=100

REM  	Save the variables to do the math
	echo %Max% > "temp_Max"
	echo %Min% > "temp_Min"
	echo %H%   > "temp_H"
	echo %KM%  > "temp_KM"
	echo %L%   > "temp_L"

REM	Show in Terminal
REM	--------------------------------------------------------------
	echo Horizontal Scale=
	gmt math "temp_KM" "temp_L" DIV %FH% MUL = 
	gmt math "temp_KM" "temp_L" DIV %FH% MUL = "temp_Esc_Hz"

	echo Vertical Scale =  
	gmt math "temp_Max" "temp_Min" SUB "temp_H" DIV %FV% MUL =
	gmt math "temp_Max" "temp_Min" SUB "temp_H" DIV %FV% MUL = "temp_Esc_Ve"

	echo Vertical Exaggeration=
	gmt math "temp_Esc_Hz" "temp_Esc_Ve" DIV =


REM	Add info in the figure
REM	-------------------------------------------------------------
REM	Data from terminal
	echo Esc. Hz. = 1:37.797.333 | gmt pstext -R -J -O -K -F+cBL+f9p -Gwhite -W1 >> %OUT%
	echo Esc. Ve. = 1:460.000    | gmt pstext -R -J -O -K -F+cBC+f9p -Gwhite -W1 >> %OUT%
	echo Ex.Vert. = 82           | gmt pstext -R -J -O -K -F+cBR+f9p -Gwhite -W1 >> %OUT%
```
[...]

PR #5019, opened 2021-03-25
WIP Anim14

Esteban82:
I share with you my idea for animation 14 to show the ability to make a movie with meca and coupe. There are many details that can be improve in the script but now I would like to know your general opinion before I go on improving the scripts and the animation.

About the data: I used the Global CMT web page and I add a column with the time.

Here I shared a [link of the movie](https://www.dropbox.com/s/3hy6wz4hlffiwa9/anim14.mp4?dl=0).


Esteban82:
> Is the red NW-SE line related to the distance coordinates in the lower panel?

The red line in the location of the profile of the lower panel. so the distance would be the same. Maybe I could use the -Ab sintaxis (for coupe -Ablon1/lat1/strike/p_length/dip/p_width/dmin/dmax) so I could use p_length to set the axis label directly.


> I think that it could look smoother to iterate by the month number rather than month name in the upper right corner (i.e. 03-2013 rather than March-2013). What do you think?

Yes, here I show one frame per month but in the final animation it would one per week or per day (so the movie would be around 1 minute). 


> Did you add the time column using GMT? If so, I think it would be nice to include that step in the script. 

No, I did it on excel. But it should be possible to do it with a unix tool. The date is in the name of the event (YYYYOODDHHMM) 200501061335A. And I only use data from 2005 onwards. For older events the names are different. 

> In the `WIP: explain section` it would also be helpful I think to eventually include the query info for the GMT Catalog Search.

Yes, good idea.

PaulWessel:
Nice job, @Esteban82!   I also have a couple of suggestions beyond what @meghanrjones listed:

1. You have 	DEM="/home/federico/E:/Facultad/Datos_Geofisicos/Batimetria/SRTM15+v21/SRTM15+V2.1.nc"	but for the final version it should probably just use @earth_relief_15s instead.
2. I very recently updated the coupe -A syntax to use modifiers instead of those endless slashes - please see if you can get that to work.
3. The final version should probably combine the several gmt set calls into one.  Fewer total script lines the better.
4. Eventually translate to english comments - we can help with cleaning that up.

Esteban82:
Thanks @PaulWessel. I will see it.

PaulWessel:
> > Is the red NW-SE line related to the distance coordinates in the lower panel?
> > Did you add the time column using GMT? If so, I think it would be nice to include that step in the script.
> 
> No, I did it on excel. But it should be possible to do it with a unix tool. The date is in the name of the event (YYYYOODDHHMM) 200501061335A. And I only use data from 2005 onwards. For older events the names are different.

I am sure @seisman can pull out a clever **sed** command for this...  FYI, I have contacted the CMT people and they are willing to at least look into allowing their web tools to have an option to select time in ISO format as an extra column, but did not have time at the moment ...

seisman:
> > > Is the red NW-SE line related to the distance coordinates in the lower panel?
> > > Did you add the time column using GMT? If so, I think it would be nice to include that step in the script.
> > 
> > 
> > No, I did it on excel. But it should be possible to do it with a unix tool. The date is in the name of the event (YYYYOODDHHMM) 200501061335A. And I only use data from 2005 onwards. For older events the names are different.
> 
> I am sure @seisman can pull out a clever **sed** command for this... FYI, I have contacted the CMT people and they are willing to at least look into allowing their web tools to have an option to select time in ISO format as an extra column, but did not have time at the moment ...

It's possible to use **awk**, but the command will be very long. Will see if there is a short command for this.

PaulWessel:
Perhaps GMT can sniff around in the trailing text somehow and get this time.  But not sure how to tell coupe/meca that.

Esteban82:
For the data there is also the option to use the [ndk files](https://www.ldeo.columbia.edu/~gcmt/projects/CMT/catalog/allorder.ndk_explained). Then with [ndk2meca](http://www-udc.ig.utexas.edu/external/becker/software/ndk2meca.awk) can be converted for GMT (which should be modified to include the time data).

PaulWessel:
The ndk2meca script is miles long.  Ekstrom may or may not find time to add the time-out feature to his website (code is pretty old I think).  So, options:

Time is not needed in **meca** or **coupe**; it is entirely something just needed by **events** in animating those features.  Given that time is actually present by default in the CMT output, perhaps another solution than to have to insert a time column is to just read it from the trailing text as needed? If **events** cannot make sense of that time-stamp then we must expect a time column instead?  I.e., if first record has timestamp in trailing test then use tralling text for all time records, else expect the extra time column?  What do you think, @Esteban82 and @seisman ?  I can then cancel the request with Ekstrom and all will work well, and we would not have to worry about introducing yet another "meca" file format just for the rare case of animation.

PaulWessel:
Just looking at your data file, first record. Has time as 200501081845A.  is it normal that there are no seconds on the time? Or is this variable and other records might have ss at the end?

Esteban82:
Right @PaulWessel. The names of the focal mechanism after 2005 have that format without seconds. The events from 1976-2005 have this format:  111304G, which correspond to MMDDYY. In the NDK file there is information of the seconds. 
Later I will repond your first question.

PaulWessel:
Thanks @Esteban82.  I imagine there will be times when seismologist really want/need to know the rupture time down to the second, hence we must preserve the ability to read a time-column.  However, for lots of animations like yours, clearly nearest second is more than enough.  So I think the ability to get the time from the current record would avoid the gymnastics you had to do in Excel to get time into that column.

Esteban82:
I found an old command to convert the ndk to gmt and I modified to include the date and time to the output:

 `gawk '/^PDE/ {Date=$2; Time=$3; Lat=$4; Long=$5; Depth=$6; getline; Name=$1; getline; getline; getline; strike1=$12; dip1=$13; rake1=$14; strike2=$15; dip2=$16; rake2=$17; print Long, Lat, Depth, strike1, dip1, rake1, strike2, dip2, rake2, Date, Time, Name}' jan76_dec17.ndk > jan76_dec17.gmt`

The output now has the date and the time with SS.X. 
91.01 38.15 27.0 322 37 127 98 61 65 1977/01/01 21:39:41.3 B010177B

I plotted with coupe and it works. Do you now how could I get the correct format for the time ti use it with events?

PaulWessel:
OK let me know.  Coupe by itself does not use tiem so events is the test.  I suspect you need to replace those slashes with hyphens, i.e., 1977-01-01T21:39:41.3 for a fully ISO format.  But, we could possibly handle that internally in events if it comes to that.

Esteban82:
I had to add the exponent to the data so the command that works is (the field of Name could be remove). It would be good if you check that the format is ok. 

`gawk '/^PDE/ {Date=$2; Time=$3; Lat=$4; Long=$5; Depth=$6; getline; Name=$1; getline; getline; Exp=$1; getline; strike1=$12; dip1=$13; rake1=$14; strike2=$15; dip2=$16; rake2=$17; print Long, Lat, Depth, strike1, dip1, rake1, strike2, dip2, rake2, Exp, Date "T" Time, Name}' jan76_dec17.ndk > jan76_dec17.gmt`

So now I can plot all the events from 1976-2017 in all the world:

[Movie](https://www.dropbox.com/s/440rbyar0jd5aki/anim14_2.mp4?dl=0)

anim14_2

seisman:
> So I think the ability to get the time from the current record would avoid the gymnastics you had to do in Excel to get time into that column.

I don't think **events** should do any special operations adding a time column. Please note that both **meca** and **coupe** can accept focal mechanisms in many different formats. Even for the CMT format, people may still get the focal mechanisms from other agencies. So there is no guarantee that the last column is a string like "YYMMDDhhmmA".

Esteban82:
I forgot to mention that I replaced the / for - in the date.

PaulWessel:
OK, I listen to seismologists.  No gymnastics, we expect users to add time column.  The documentation can give some hints for how that could be done, and maybe Ekstrom will help us down the road to make it an optional format.

Esteban82:
I managed to get correct output format from all the ndk files with this: 

```
gawk '/^PDE/ {Date=$2; Time=$3; Lat=$4; Long=$5; Depth=$6; getline; Name=$1; getline; getline; Exp=$1; getline; strike1=$12; dip1=$13; rake1=$14; strike2=$15; dip2=$16; rake2=$17; print Long, Lat, Depth, strike1, dip1, rake1, strike2, dip2, rake2, Exp, Date "T" Time, Name}' 0_Data/*.ndk |	sed -i 's/\//-/g' meca.gmt
gmt convert "meca.gmt" -N10+a > "out.txt"
```
The output look like this: 
91.01	38.15	27	322	37	127	98	61	65	25	1977-01-01T21:39:41.3	B010177B


However I noticed that if I joint the 2 commands like this: 

`gawk '/^PDE/ {Date=$2; Time=$3; Lat=$4; Long=$5; Depth=$6; getline; Name=$1; getline; getline; Exp=$1; getline; strike1=$12; dip1=$13; rake1=$14; strike2=$15; dip2=$16; rake2=$17; print Long, Lat, Depth, strike1, dip1, rake1, strike2, dip2, rake2, Exp, Date "T" Time, Name}' 0_Data/*.ndk |	sed 's/\//-/g' | gmt convert -N10+a > "out.txt"`

I get an error in the last column. If this a bug in gmt convert?

91.01	38.15	27	322	37	127	98	61	65	25	1977-01-01T21:39:41.3	C200807270118A

Esteban82:
I would like to know your opinion. For the map of the animation I was thinking to show all the focal mechanism in black/gray and to plot in color only the ones that are shown in the profile (see figure). What do you think? Also I think it would be better NOT to show the seismic events. 

EJ21 2_Mapa_Sismicidad_II

PaulWessel:
I agree the coloring/gray-shade is a great way to convey what events are being shown in the cross-section.  And by seismic events you mean the non-CMT small circles, right?  Yes, maybe cleaner without them.

PaulWessel:
I cannot test the **gmtconvert** issue above since the ndk files are not in this branch.  Maybe add them for now - we can always remove and finalize before merging the official version.

PaulWessel:
@Hi @Esteban82, this single command works for me (I assume you were using **gmtconvert** to get tabs?):

gawk '/^PDE/ {Date=$2; Time=$3; Lat=$4; Long=$5; Depth=$6; getline; Name=$1; getline; getline; Exp=$1; getline; strike1=$12; dip1=$13; rake1=$14; strike2=$15; dip2=$16; rake2=$17; print Long, Lat, Depth, strike1, dip1, rake1, strike2, dip2, rake2, Exp, Date "T" Time, Name}' jan76_dec17.ndk | sed -e 's/\//-/g' -e 's/ /\t/g'  > meca.gmt

Esteban82:
Right, sorry @PaulWessel.The problem is that I downloaded all the ndk files available (the one I uploaded and then 35 more files, one per each month from 2018 to 2020). Here is the link with the[ 2018 year](https://www.ldeo.columbia.edu/~gcmt/projects/CMT/catalog/NEW_MONTHLY/2018/). So, when I use gawk with *.ndk (instead of jan76_dec17.ndk) the records in the output were mixed (due to name of the ndk file) and I have to order them by date.

Maybe for the animation, i could use the jan76_dec17.ndk only. But it seems that there is an issue with gmt convert.

PaulWessel:
Can you tell me what the gmt convert error you see is? It runs for me on that single file.  I took a few records from that file and changed the order then ran gmt convert -N and it sorted them.

meghanrjones:
> I would like to know your opinion. For the map of the animation I was thinking to show all the focal mechanism in black/gray and to plot in color only the ones that are shown in the profile (see figure). What do you think? Also I think it would be better NOT to show the seismic events.

Yes, I agree it looks good to only color ones shown in the profile and it would be better with only the beach balls shown. I would prefer a colormap like 'inferno', 'hot',' or 'magma' for the depth coloring.

Esteban82:
@PaulWessel  in the https://github.com/GenericMappingTools/gmt/pull/5019#issuecomment-809651917 I get an error in the last column (the name of the event, C200807270118A instead of B010177B).

Esteban82:
Another detail. Instead of plotting a label with the date and a bar progress, I think it would be better instead to only a have a progress bar with the dates (month and year instead of %).  Is it possible?

I try a bit with the test file (movie_indicator_f_ver.sh) and I see the message:

`movie [WARNING]: Cannot handle elapsed time with progress indicator (f) yet - skipped`

PaulWessel:
Not possible yet with elapsed time.  But you can use absolute time (i.e., your actual earthquake times).

Esteban82:
> Not possible yet with elapsed time. But you can use absolute time (i.e., your actual earthquake times).

Rigth @PaulWessel. I think I will managed to add the time in the progress bar.

About the time elapsed, I managed to do this with days in 2018 just changing the TIME_EPOCH. Is that the idea of elapsed time ?

movie_indicator_f_ver

with this script:
`gmt movie map.sh -CHD -T2018-01-01T/2019-01-1T/1d -Fnone -Ml,png -Nmovie_indicator_f_ver --FONT_TAG=10p,Helvetica,black --TIME_UNIT=d --TIME_EPOCH=2018-01-01 --FORMAT_DATE_MAP=mm-yyyy -Pf+w4.0i+ac0+jML+o0.5i/0 -Pe+w4.0i+ac0+jML+o1.0i/0 -Pd+w4.0i+ac0+jML+o3.0i/0
`

PaulWessel:
While this may work, it probably won't work across year boundaries.  By elapsed time we need the ability to handle the ISO convention for [elapsed time](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals); currently GMT does not do that so a proper implementation is held up until that can be added; not for 6.2.0 unfortunately.  So for now what you do is fine but perhaps not general.

Esteban82:
@PaulWessel a do it with absolute time. I get this figure. I think I will use the first option (-Pf). In the other two figures is it possible to get the dates in the format of mm-yyyy (instead of fraction of year)? I try with FORMAT_DATE_OUT, FORMAT_DATE_MAP and FORMAT_DATE_IN but it didn't work. 


movie_indicator_f_ver

PaulWessel:
Hi @Esteban82, I can have a look.  Is the example above a small ovie example just with the progress bars? If so could you please share so I don't have to start from scratch?

Esteban82:
Hi @PaulWessel 
The script is:
```

ps=movie_indicator_f_ver.ps
cat << EOF > map.sh
gmt begin
	gmt plot -R0/5/0/5 -Jx1i -T -X0.1i -Y0.1i -B0
gmt end
EOF

gmt movie map.sh -C5.2ix5.2ix100 -T1981-09-01T/2018-01-1T/1y -Fnone -Ml,png -Nmovie_indicator_f_ver -Pf+w4.0i+ac0+jML+o0.5i/0 -Pe+w4.0i+ac0+jML+o1.0i/0 -Pd+w4.0i+ac0+jML+o3.0i/0 --TIME_UNIT=y --TIME_EPOCH=0000-01-01 --FORMAT_DATE_OUT=mm-yyyy --FORMAT_DATE_MAP=mm-yyyy --FORMAT_DATE_IN=mm-yyyy
```

PaulWessel:
Thanks @Esteban82, see #5088 for the problem.  Now it works (and would have worked if **-T** was given a file with times).  Note: see the note for **-L** in the man page regarding DATE and CLOCK settings.  You probably only want date and not clock so make sure you use `--FORMAT_DATE_MAP=mm-yyyy --FORMAT_CLOCK_MAP=-`

Esteban82:
Hi @PaulWessel Paul. I think I have a bug here with -Pf. I run the [previous script](https://github.com/GenericMappingTools/gmt/pull/5019#issuecomment-812933703). As you can see in the figure there is no bar for the -Pf option. The are two bars (-Pd and e) are ok.

In the terminal I get this error:
```
basemap [ERROR]: Option -R parsing failure. Correct syntax:
	-R<xmin>/<xmax>/<ymin>/<ymax>[/<zmin>/<zmax>]
	  Append +r if giving lower left and upper right coordinates
	-Rg or -Rd for global domain
	-R<grdfile> to take the domain from a grid file
basemap [ERROR]: Offending option -R01-09-1981/01-01-2018/0/1
plot [ERROR]: Failed to call basemap with args -R01-09-1981/01-01-2018/0/1 -JX4i/0.0001i -Baf -BN -X0 -Y0 --MAP_FRAME_PEN=3p,black --FONT_ANNOT_PRIMARY=+9p,Helvetica,black --GMT_HISTORY=false ???

```


movie_indicator_f_ver

PaulWessel:
Yikes.  I understand what is going wrong.  The -Pf is tricky as it needs to pass a working -Rstring and I did that but I used the FORMAT_DATE_MAP settings in making the string, and of course that is no good in -R...  Stand by.

PaulWessel:
I have a fix - running the 1100 tests first to make sure no side-effects, then will post a PR.

PaulWessel:
Make sure you never use **FORMAT_DATE_OUT** and **FORMAT_CLOCK_OUT** with movie as these may be used internally to prepare proper ISO timestamps for use in **-R** arguments for calls to other modules.  I do not check for that.  You can use **the_IN** versions if your data require it and of course the **_MAP** version for formatting the plotting.

Esteban82:
Here is the master frame of my current version of the animation.
There are some details still to improve. 

[Link Movie](https://www.dropbox.com/s/bb70alyyda2ve1r/anim14.mp4?dl=0)

Map:
1. Maybe I could reorder the scale, colorbar and progress bar. What do you think?
2. Should I include an inset map with the location of the area?
3. Should I remove the annotations of the frame? If so, I could enlarge a bit the map and remove the white areas outside the map (that do not look very well in animation in my opinion).
4. Should I reduce a bit the size of the focal mechanisms?
5. How could I improve the location the profile?

Profile: 
1. I like that there is a continuation of events in almost all the profile.
2. Also here should I reduce the sizes?
3. There are many white spaces. I will work to add some of the features discussed in #4367. Most likely coordinates and the angles. 

anim14_a4

meghanrjones:
For the profile, I think it would be nice to have it close to normal to the trench. By rough estimation, something like this would have a near continuous distribution of focal mechanisms:
image

I agree with you that an inset would work instead of the outside annotations.

For the ordering of the scale, colorbar and progress bar, I think it could look nice with the progress bar inside the same white box as the depth colorbar with the colorbar to the left of the progress bar. Then the width of the scale bar box and the colorbar box could be the same.

Also, regarding the timing for this PR, we are freezing the code for 6.2rc1 to get that candidate out so that people will hopefully give feedback soon. This animation can be included in the next release candidate for 6.2, if that plan is OK with you.

Esteban82:
> For the profile, I think it would be nice to have it close to normal to the trench.

Mmm, geologically I would agree. But the issue is that the profile would be E-W. And I think I would be better to show a diagonal profile could be done with GMT without any problem. Besides the diagonal profile matches the direction of the subducting slab in that area. 

> This animation can be included in the next release candidate for 6.2, if that plan is OK with you.

Ok, I have no problem.

meghanrjones:
> > For the profile, I think it would be nice to have it close to normal to the trench.
> 
> Mmm, geologically I would agree. But the issue is that the profile would be E-W. And I think I would be better to show a diagonal profile could be done with GMT without any problem. Besides the diagonal profile matches the direction of the subducting slab in that area.
> 

OK, diagonal is good then.
> > This animation can be included in the next release candidate for 6.2, if that plan is OK with you.
> 
> Ok, I have no problem.

Great!

PaulWessel:
Nice job, @Esteban82!  A few comments from me:

1. My eyes correlates the events in the profile with those on the map and I notice the shift: Your distance axis on the bottom is much longer than the actual profile.  I wonder if you can make these more or less similar by either (a) cut off the area between 61-64W where there are no events, or (b) let the max distance on the profile be a few hundred km longer so that there is more (vertical, on the plot) alignment between events on the map and in the profile.  The latter is probably simpler since (a) would mess up your HD format shape.
2. I agree with @meghanrjones that it would be nice if the three embellishments were all inside a common white box.

Esteban82:
As I said before, in the profile, I would like to add the vertical exaggeration. When I run basemap -Vi I know that the:
`Linear projection implies x-axis distance exaggeration relative to the y-axis by a factor of 0.671316`

How could I save that value so I could plot it in the profile?

I know how to do it with gmt math but I think it would be better to use basemap to avoid adding extra commands in the script.

BTW, if I use a negative value in -J (like -JX23/-3) I get a negative value of vertical exaggeration.

Esteban82:
Here is the current version. 

I add some info in the profile, including an inset with the distortion of the angles due to its vertical exaggeration. Is it clear? 

Pending issues: 
1. Add three embellishments inside a common white box.
2. Improve vertical alignment between events on the map and in the profile.
3. Reduce the width profile so that the "Depth" label appears.

Doubts: 
1. Improve the location map? How?
2. Improve the profile trace on the map? How?


anim14_hot

PaulWessel:
Hi @Esteban82, looks very nice.  Maybe improve the contrast between the green topo and the greenish beachballs by picking another CMT color scheme?

Esteban82:
I modify the -Mi (in events) to avoid the greenish effect. I share the figures with the 3 cpt proposed by @meghanrjones (hot, inferno and magma). I like the first one. 

I couldn't put the three embellishments inside a common white box. First I remove the map scale (the distances can be know from the profile). I don't know how to do it. I try with and empty legend. But it is difficult to place the colorbar and progress bar together in part due that the each one have different anchor (the colorbar to the map and the progress bar to the canvas). 

Any comments on the master frame? Soon I will share the movie. 

anim14_hot
anim14_inferno
anim14_magma

PaulWessel:
I agree it may be too complicated to make all those embellishments line up inside a common box.  Let's not bother - your movie ie great as is - and you get to pick the colortable so _hot_ is fine with me.

meghanrjones:
Agreed, it looks great!

Esteban82:
[Link](https://www.dropbox.com/s/5u6sk9gs5u54mvd/anim14_hot_movie.mp4?dl=0) to the movie. I have to modify it to last 1 minute.

BTW now = am using the @earth_relief_03s grid and I get the following warning message. Is there any problem if there is warning message for an official animation? 
`grdblend [WARNING]: File @S40W070.earth_relief_15s_p.nc coordinates are phase-shifted w.r.t. the output grid - must resample`

PaulWessel:
The 3s and 1s SRTM tiles only come in gridline registered format, while the others come in both, except earth_relief_15s which is pixel as well.  I was hoping to convince David Sandwell to also make a 15s gridline version but not yet, hence we get this warning.
We may wish to move this message to INFORMATION rather than WARNING since there is nothing you can do about it, or did wrong.

PaulWessel:
Hi @Esteban82.  I am trying to run the movie but you ahve personal data files in there.  Perhaps you ahve not pushed commits to the branch in a while? I need to make this movie this week for the EGU talk:

```
grdimage [ERROR]: File /home/federico/E:/Facultad/Datos_Geofisicos/Batimetria/SRTM15+v21/SRTM15+V2.1.nc was not found
grdimage [ERROR]: Cannot find file /home/federico/E:/Facultad/Datos_Geofisicos/Batimetria/SRTM15+v21/SRTM15+V2.1.nc
```

Esteban82:
Hi @PaulWessel. I have some issues with the script as I describe in #5142.  Here is a file with the script and the data (up to 2021)
[anim14_Paul.zip](https://github.com/GenericMappingTools/gmt/files/6330161/anim14_Paul.zip). Do you want to me to update the script in the branch? I ask you because there are some things that I need to improve in the script and I don't know if it would be better to fix that issues first.

PaulWessel:
Thanks @Esteban82.  No, it is OK that you wait to update the script.  I do not have time to try to deal with #5142 until after EGU so I will use your zipped script/data for now. Thanks again!

Esteban82:
You are welcome. BTW check the last line in the script that suspend the pc. In the script you probably need to see the options for the movie (i.e. movie -T and events -Es+r+p+d+f).

Esteban82:
@PaulWessel Did the script worked? BTW, it would be possible to see your talk (live or later)?

PaulWessel:
Yes, worked fine.  Thanks!  I think the talk is under EGU lock.  I will ask them if it can be opened up but I will also record a practice talk just in case.

Esteban82:
> Yes, worked fine. Thanks! I think the talk is under EGU lock. I will ask them if it can be opened up but I will also record a practice talk just in case.

Link to the talk https://meetingorganizer.copernicus.org/EGU21/session/41216#Presentations

Esteban82:
For me, this is the[ final version](https://www.dropbox.com/s/wmuyr4uvhiaj5td/anim14_image_7d_i0.6_f11_r21%2Bp42%2Bd35%2Bf.mp4?dl=0) of the animation. I modified a bit the animations parameters.

meghanrjones:
> For me, this is the[ final version](https://www.dropbox.com/s/wmuyr4uvhiaj5td/anim14_image_7d_i0.6_f11_r21%2Bp42%2Bd35%2Bf.mp4?dl=0) of the animation. I modified a bit the animations parameters.

Looks great!

PaulWessel:
Great.  What remains is to clean up the script a bit (remove unused things) and do a translation to english.  We can help with that but if you want to take a first crack at it please do.  We will also need a statement on what the movie is about - this goes into the short blurbs, like doc/rst/source/animations/anim13/rst.  Could you copy/paste and edit and add a anim14.rst file to this branch, then add the new entry in doc/rst/source/animations.rst?  Once all ready then @meghanrjones or I can run off the movie and post on YouTube.

Esteban82:
I updated the script. At the end I will revised the comments in the first lines (3 to 17).
Issues:
1) I try to include from the line 43 to 53 in pre.sh but if I do that, I doesn't work. Does it matter? How could I do that? Or should I leave how it is now?
2) I try to use something like -Yh+0.3c in line 72 (to avoid creating $Y1) but I didn't get the correct shift. 
3) Is the rest the script clear?

PaulWessel:
> 1. I try to include from the line 43 to 53 in pre.sh but if I do that, I doesn't work. Does it matter? How could I do that? Or should I leave how it is now?

If you want to use those variables in pre.sh then they need to either be run by pre.sh or you can stick these things in an include.sh and use `-Iinclude.sh` in movie.

> 2. I try to use something like -Yh+0.3c in line 72 (to avoid creating $Y1) but I didn't get the correct shift.

Hm, I should look at that - it should work as you indicate.

> 3. Is the rest of the script clear?

Generally yes, but I will look harder after a family zoom... One thing you can probably do for me is to add units where they are implied so that is is clear.  E.g., -W1p instead of -W1, -M0.15c instead of -M0.15, etc. so that the animation will run the same way for anyone who has changed their **PROJ_LENGTH_UNIT** to inch.

Esteban82:
Hm, I should look at that - it should work as you indicate.

        I try to use something like -Yh+0.3c in line 72 (to avoid creating $Y1) but I didn't get the correct shift.

@PaulWessel I think that there is an bug related with gmt inset (which is the previous command in the script). I remove it and it works fine. And also it works when I move it upwards in the script (so it wasn't the previous command.

PaulWessel:
Thanks, that is helpful.  I am sure it is like for subplot (which does work that way) that I need to reset some dimension after gmt inset end.

Esteban82:
I could add all the units except for the variable W. If a I add the unit (W=22.73c) then I got this figure where the vertical exaggeration and the dimensions of the inset are wrong (both are calculated with gmt math from $W).

@PaulWessel It seems that there is a BUG with gmt math -Q and DIV or I do not understand how it works
I try the following commands:

`gmt math -Q 1c 1c ADD = --PROJ_LENGTH_UNIT=c` = 2
`gmt math -Q 1c 1c DIV = --PROJ_LENGTH_UNIT=c` = 2.54 !!??
`gmt math -Q 1c 1c DIV = --PROJ_LENGTH_UNIT=i` = 1      !!??

Also, I haven't any luck to include the lines 42-52 in `include.sh`.  But maybe we could leave as it is.

meghanrjones:
@Esteban82, meca.gmt should now be available as a remote file using @meca.gmt. Could you please update the script to use the remote file format for reproducibility?

Also, I think you will need to change the file permissions to a+x and use -Fmp4 in the final script. Can you confirm whether this is correct @PaulWessel?

PaulWessel:
yes, those are true.  Also, need to remove superfluous quotes around filenames with no variables or spaces, e.g. "meca.gmt", etc.

Esteban82:
Done @meghanrjones. I am affraid I don't know how to change the permission to a+x.

PaulWessel:
I must merged master into this branch so we are getting very close to merging this final branch back to master.

Esteban82:
Is this important? https://github.com/GenericMappingTools/gmt/actions/runs/797990491

PaulWessel:
I don't think so (yet).  Note, I just added some updates (fix a few spellings etc) and added the anim14.rst file and where it is called in animations.rst.  @Esteban82, could you please git pull and make any changes you would like to see in the anim14.rst file? @meghanrjones can run the animation and

1. Update the timing information in the script
2. Upload the animation to our YouTube Channel
3. Update the URL in the anim14.sh, anim14.rst files

THen I think we can remove the WIP from the title, run all tests etc, and be ready to merge, no?

Esteban82:
I just revised the rst file (fix a typo).  Remember that the timing of the animation can be reduced after #5142. But It can be updated later.

PaulWessel:
Thanks. Yep, I will look for an implementation of that, or at least explain a workaround, but this wont be done for 6.2.0.

PaulWessel:
Took 1 hour 35 minutes on my Mac Pro...

meghanrjones (approved):
Great work, @Esteban82!

Esteban82:
You are welcome. Should I Squash and merge?

PaulWessel:
yes, please have the honor since it is your branch.

PR #5025, opened 2021-03-25
Update coupe_common.rst_. Default value of -A+w

Esteban82:
The default value for -A+w is it 0, right? I add it. Not sure if the description is clear enough.

### DIFF
--- a/doc/rst/source/supplements/seis/coupe_common.rst_
+++ b/doc/rst/source/supplements/seis/coupe_common.rst_
@@ -46,7 +46,7 @@ Required Arguments
      limiting the length of the cross-section, *dip* is the dip of the plane
      on which the cross-section is made [90], *width* is the width in km of the
      cross-section on each side of a vertical plane or above and under an
-     oblique plane, and *dmin* and *dmax* are the distances min and max from
+     oblique plane [0], and *dmin* and *dmax* are the distances min and max from
      horizontal plane in km, along steepest descent direction. Add **+r** to get the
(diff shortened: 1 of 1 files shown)


PaulWessel:
I am not sure, adding @seisman to opine.

seisman:
In the new syntax, it seems *p_width* is uninitialized if **+w***width* is not given. Does *p_width*=0 mean that no beachball will be plotted?

PaulWessel:
All Ctrl variables are zero an NULL when initialized, so yes p_width will be zero unless set.  I just don't know if that means nothiing is plotted.

seisman:
@Esteban82 What will happen if **+w***width* is not given?

https://github.com/GenericMappingTools/gmt/blob/fcd0f36ab2c545eda3f3f0d91a557882d18daf1d/src/seis/pscoupe.c#L1132-L1134

As I understand the codes, the beachball will always be skipped when p_width is uninitialized (=0).

Esteban82:
Without +w no beachball was plot.

seisman:
I think the default behavior is useless. Can we change it to infinity (means all beachballs are plotted by default)?

PaulWessel:
Yep, that is much better. I can do that.

PaulWessel:
Hi @Esteban82 I guess you can delete this branch as I did the +w default in another.  Please give it a try to ensure the default (infinity) works as you expect.

Esteban82:
Should I close this PR, right?

Issue #4992, opened 2021-03-18
histogram -N odd behaviour


gd-a:
[For reference](https://forum.generic-mapping-tools.org/t/histogram-with-new-w-cyclic-option/1470/11?u=planetgus)

The equivalent normal distribution is not plotted correctly : 

```
gmt begin test01 png
gmt histogram  -JX10c test.txt  -T1 -BWSen -Bxa1Of1O+l"Months" -Byaf+l"Frequency"+u"%" -D+r -Z1 -N0 -wa -Glightblue -Wthinnest,black --FORMAT_TIME_PRIMARY_MAP=c
gmt end
```

![](https://forum.generic-mapping-tools.org/uploads/default/original/2X/2/267b68628cc9df923bb844021d99ef4add3fb90b.png)

[test.txt|attachment](https://forum.generic-mapping-tools.org/uploads/short-url/yoKybUabavuVCJQZZCyobon9EV.txt)

meghanrjones:
I think @PaulWessel will disagree with me here, but in looking at the code and documentation I am not convinced that this behavior should be unexpected. GMT is taking the mean (6.14 for these data) and standard deviation (4.47 for these data) of f(x) for 0<x<12 and plotting the equivalent normal distribution.

gd-a:
Maybe, but wouldn't it be more relevant to have the curve matching the data ?
Also If I shift the bar plot to be centred on winter (new cyclic feature), the curve changes...

meghanrjones:
I view it as the `R=-5/7/0/12` is basically applying a phase shift to the data (range of 0-12 before and -5-7 after), which is making the mean change depending on whether the plot is centered on summer or winter. I'm not sure if that interpretation is correct - the option to use negative numbers in region with -w is not fully documented but rather used in one of the **-w** examples.

PaulWessel:
Will look once back from vaccinations

PaulWessel:
Thanks @meghanrjones, I agree that the issue has to do with the **-R** setting and the wrapping.  The first (wrong) plot does compute a mean half-way between the peaks etc.  It just does not know that this data is periodic and one cannot take the mean  like that when **-w** is on.  I am not sure what is the best, but an analogy with a circle would suggest I need to fit a _von Mises_ distribution to this rather than the normal distribution.  I.e., basically it is statistics of data on a circle.  I know how to do those things, so will see how to implement it.

PR #5028, opened 2021-03-26
Add VPDF for Von Mises PDF to gmtmath

PaulWessel:

This adds a new operator VPDF that creates a PDF for the Von Mises circular (wrapped) normal distribution for periodic data.  I added a test to ensure it works.  I also removed the number of operators listed in the doc since it is always out of sync with reality anyway.


PaulWessel:
FYI: The **VPDF** and PR was added for several reasons:

1. Given the ability to wrap data into periodic functions (**-w**) the statistics of the wrapped functions as used in **histogram -N** must follow the very different rules of such distribution (see failure in #4992).  Here, von Mises analysis will be required.  Having this capability in gmt **math** is helpful.
2. **psrose** does not have a an option similar to **histogram -N** but it could.  In that case, the curve would be a von Mises fit.
3. I have done quite a bit with these distros as well as the 3-D equivalent (Fisher distribution on a sphere) via my data analysis and book so it is pretty much ready to go apart from implementation.

So **VPDF** may soon show up in grdmath as another PR, as will **IPDF** for Fisher distribution.

PaulWessel:
Might @joa-quim have time to check this branch and approve?  It is pretty isolated and cannot break other things (than itself - but the test works).

joa-quim:
some missing commit?
```
grdmath.c.obj : error LNK2019: unresolved external symbol gmt_vonmises_pdf referenced in function grdmath_VPDF
src\gmt_w64.dll : fatal error LNK1120: 1 unresolved externals
```

PaulWessel:
That branch awaits the gmtmath PR first; it is not a PR yet

joa-quim:
OK, wrong branch. This one builds fine

joa-quim (approved):

PR #5030, opened 2021-03-27
Add VPDF for Von Mises PDF to grdmath

PaulWessel:

Allows evaluation of the Von Mises PDF.  Also removes the out-of-date number of operators and fixes formatting for **DAYNIGHT** operator.  It is simple enough that I did not bother to add another test for this.  Here is a case showing how the PDF is very broad (and wrapps over 360) when kappa is small then focuses on the mean (125) when kappa gets bigger:

```
gmt grdmath -R0/360/1/20 -I1 X 125 Y VPDF = t.grd
gmt grdimage t.grd -JX15cd/8c -Bxaf+l"Angle" -Byaf+l"@~k@~" -B+t"Von Mises PDF as function of @~k@~" --FORMAT_GEO_MAP=+D -png map
```

map

### DIFF
--- a/doc/rst/source/grdmath.rst
+++ b/doc/rst/source/grdmath.rst
@@ -151,11 +151,11 @@ Optional Arguments
 Operators
 ---------
 
-Choose among the following operators. "args" are the number of input
+Choose among the following operators. "Args" are the number of input
 and output arguments.
 
 +---------------+-------+--------------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


PaulWessel:
Probably lat for you, @joa-quim but this is the other branch.  The third for **FISHER** distribution may come after your bedtime.

joa-quim (approved):

PR #5033, opened 2021-03-29
fix two typos

anbj:
Please revise the fix;

This is the change I made:
```
that could contain either an image or a data set (e.g., geotiff) we determine
```

But it could also be meant to say:
```
that could contain either an image of a data set (e.g., geotiff) or a data set we determine
```


meghanrjones (approved):
Thanks, @anbj!

PR #5043, opened 2021-03-30
Consolidate source text for cookbook and manpage -a docs

meghanrjones:

This PR consolidates the source documentation for -a for the cookbook and reference docs.

Addresses #4969

**Reminders**

### DIFF
--- a/doc/rst/source/cookbook/options.rst
+++ b/doc/rst/source/cookbook/options.rst
@@ -638,57 +638,8 @@ Plot positioning and layout: The **-X** **-Y** options
 OGR/GMT GIS i/o: The **-a** option
 ----------------------------------
 
-**Syntax**
-
-**-a**\ [[*col*\ =]\ *name*\ ][,\ *...*]
-
-**Description**
-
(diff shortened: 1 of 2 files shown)


PaulWessel (approved):
A few minor things: The gmtlib_ogr_get_type function accepts either DOUBLE or double, etc.  Does it look cleaner if we use lower case here?
THe listing of the col values D, G, L is missing a comma after I in one of the two places.
SHould stuff that follows **Note**: start with capital letter? We have a few places where it is lower case, e,g, "if".

meghanrjones:
> A few minor things: The gmtlib_ogr_get_type function accepts either DOUBLE or double, etc. Does it look cleaner if we use lower case here?
> THe listing of the col values D, G, L is missing a comma after I in one of the two places.
> SHould stuff that follows **Note**: start with capital letter? We have a few places where it is lower case, e,g, "if".

Yes, I agree with all points. Updated in https://github.com/GenericMappingTools/gmt/pull/5043/commits/529f76c21f8aa8459eeddf2245ec56840cabc549.

PR #5056, opened 2021-03-31
Document rose -N option in reST files

meghanrjones:

This PR adds the **-N** option to the reST documentation for `rose`. 

**Reminders**

### DIFF
--- a/doc/rst/source/psrose.rst
+++ b/doc/rst/source/psrose.rst
@@ -22,6 +22,7 @@ Synopsis
 [ |-K| ]
 [ |-L|\ [*wlabel*\ ,\ *elabel*\ ,\ *slabel*\ ,\ *nlabel*] ]
 [ |-M|\ *parameters* ]
+[ |-N| ]
 [ |-O| ] [ |-P| ]
 [ |-Q|\ *alpha* ]
 [ |-R|\ *r0*/*r1*/*az0*/*az1* ]
--- a/doc/rst/source/rose.rst
+++ b/doc/rst/source/rose.rst
@@ -22,6 +22,7 @@ Synopsis
 [ |-J|\ **X**\ *diameter* ]
 [ |-L|\ [*wlabel*\ ,\ *elabel*\ ,\ *slabel*\ ,\ *nlabel*] ]
 [ |-M|\ *parameters* ]
+[ |-N| ]
 [ |-Q|\ *alpha* ]
 [ |-R|\ *r0*/*r1*/*az0*/*az1* ]
 [ |-S| ]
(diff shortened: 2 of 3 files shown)


meghanrjones:
@PaulWessel, the **-N** usage says that it's only valid for **-Sn***radius*, but there is no **n***radius* modifiers listed for **-S** in the html or command-line docs. I can look into the code regarding **n***radius*, or if you know off-hand what those modifiers do then I can add that to the docs.

PaulWessel:
Will look after a zoom that starts in 1 minute...

PaulWessel (approved):

PaulWessel:
> @PaulWessel, the **-N** usage says that it's only valid for **-Sn**_radius_, but there is no **n**_radius_ modifiers listed for **-S** in the html or command-line docs. I can look into the code regarding **n**_radius_, or if you know off-hand what those modifiers do then I can add that to the docs.

The **-S**[**n**]_radius_ was the old syntax, so this text will have to change.  I see the **-N** requires **-S** basically, so silly to have two options needed to accomplish normalization of radii, then take square root of the normalized radii.  I think this scheme would allow some simplification plus adding a **-N** option like **histogram** has:

1. Let **-N** with no argument be a backwards compatible way to select sqrt(normalized-radii).  This would be marked as a deprecated option.
2. As an alternative to **-N**, introduce **-S**[**+a**] for what presently requires **-S -N,** with **+a** meaning "area normalization"
3. Introduce **-N0**[**+p**_pen_] for drawing the best-fitting Least-squares von Mises pdf fit on top of the polar histogram (there are no robust fits yet for these distros, but if we add those later then best to have the same **-N0**|**1**|**2** syntax as in histogram).

Thoughts?

meghanrjones:
> > @PaulWessel, the **-N** usage says that it's only valid for **-Sn**_radius_, but there is no **n**_radius_ modifiers listed for **-S** in the html or command-line docs. I can look into the code regarding **n**_radius_, or if you know off-hand what those modifiers do then I can add that to the docs.
> 
> The **-S**[**n**]_radius_ was the old syntax, so this text will have to change. I see the **-N** requires **-S** basically, so silly to have two options needed to accomplish normalization of radii, then take square root of the normalized radii. I think this scheme would allow some simplification plus adding a **-N** option like **histogram** has:
> 
>     1. Let **-N** with no argument be a backwards compatible way to select sqrt(normalized-radii).  This would be marked as a deprecated option.
> 
>     2. As an alternative to **-N**, introduce **-S**[**+a**] for what presently requires **-S -N,** with **+a** meaning "area normalization"
> 
>     3. Introduce **-N0**[**+p**_pen_] for drawing the best-fitting Least-squares von Mises pdf fit on top of the polar histogram (there are no robust fits yet for these distros, but if we add those later then best to have the same **-N0**|**1**|**2** syntax as in histogram).
> 
> 
> Thoughts?

Yes, I agree with all three points.

PR #5059, opened 2021-04-01
Consolidate -p docs

meghanrjones:

This PR consolidates the source documentation for -p for the cookbook and reference docs.

Addresses #4969

**Reminders**


PaulWessel:
I think they are mostly relevant for animations, but they may also be for someone who is doing a subplot of 3 plots with different views.  I think what you have there is fine - should we learn more down the road we can adjust.

PR #5062, opened 2021-04-01
Implement histogram -N when -w is active

### LINKED ISSUE #4992: histogram -N odd behaviour

PaulWessel:

This addresses the issue raised in #4992.

Once **-w** is used the data are forced to be periodic and can be thought of as directional data on a circle with 360 degrees periodic range.  Thus, to compute the mean and standard deviation we must do it via parameter estimation for the [von Mises distribution](https://en.wikipedia.org/wiki/Von_Mises_distribution).  Earlier PRs added that distribution PDF to GMT.  This PR uses it in **histogram** after obtaining the parameters _mu_ and _kappa_ via the mean resultant vector.  The example now works and I made it into a test (fig below).  Note that the data are not particularly symmetric about the mean and that explains the less than ideal fit, especially for the chosen 1 month interval.  The second plot uses 0.1 months and it is clearer that the fit is reasonable and that the data have some outlying parts:

wrappedhisto

Closes #4992.


meghanrjones:
Nice! I think it would be worth mentioning this method for calculating the mean and standard deviation when **-w** is used in the documentation for **-N***0*.

Also, we should check with @gd-a about whether the data shared with issue https://github.com/GenericMappingTools/gmt/issues/4992 can be included in the repo. If not, we can make up some other data.

PaulWessel:
OK, true.  Hey @gd-a, would you mind if we use this dataset for this test - i.e., it is included in the test directory tree.  If not we can generate something instead.
I will update the docs.

gd-a:
All good 👍

meghanrjones:
I agree that this would make a nice example. Should we make `wrappedhisto.txt` a remote file now to save work later?

PaulWessel:
OK, I can add it to the cache. Whether or not we make ex51 is a different matter I think as we have plans to revamp the docs and focus more on a different sort of (simpler) examples.  But I will do the cache for now and we can always turn this test into an example down the road.

meghanrjones:
> OK, I can add it to the cache. Whether or not we make ex51 is a different matter I think as we have plans to revamp the docs and focus more on a different sort of (simpler) examples. But I will do the cache for now and we can always turn this test into an example down the road.

Yes, I think just adding to the examples section of the histogram docs could be helpful, since that is where more examples with remote files are most urgently needed. But, that's for another time and does not need to be in this PR. I will review this more thoroughly now.

meghanrjones:
How frequently does the server get updated from the GitHub repo?

PaulWessel:
The crontab job runs every hour.  However, we had an issue last time we did this so will need to check.  If you cannot get the file from the server cache then I may need to log in and run something manually.

meghanrjones:
> The crontab job runs every hour. However, we had an issue last time we did this so will need to check. If you cannot get the file from the server cache then I may need to log in and run something manually.

Curl download fails and I do not see it here : https://oceania.generic-mapping-tools.org/cache/

PaulWessel:
I think you have login access to gmtserver.soest.hawaii.edu.  THe crontab runs hourly and runs srv_git_update.sh.  If the git fetch shows updates then the script calls srv_do_updates.sh which then should sync up the cache dir.  I do not understand why this is not working - wanna have a look and see what you can learn?

PaulWessel:
gmtserver updated and the modified test script accessing @wrappedhist.txt works - all tests pass.  Hoping @seisman can approve as Meghan is off this evening.

seisman (approved):

seisman:
The macOS failure is unrelated.

PR #5063, opened 2021-04-01
Consolidate -s docs

meghanrjones:

This PR consolidates the source documentation for -s for the cookbook and reference docs.

Addresses #4969

**Reminders**


PaulWessel (approved):
For the +a entry: I think the stuff in the brackets don't make sense here, no?

meghanrjones:
> For the +a entry: I think the stuff in the brackets don't make sense here, no?

You're right. Is "[Default skips record only if values in all specified *cols* equal NaN]" the correct statement?

PaulWessel:
Yes, that sounds about right.  I think that is what we want to happen...

PR #5079, opened 2021-04-02
Add psrose option to draw von Mises distribution

### LINKED ISSUE #5064: Upgrade normalization options for rose module
1. Let **-N** with no argument be a backwards compatible way to select sqrt(normalized-radii).  This would be marked as a deprecated option.
2. As an alternative to **-N**, introduce **-S**[**+a**] for what presently requires **-S -N,** with **+a** meaning "area normalization"
3. Introduce **-N0**[**+p**_pen_] for drawing the best-fitting Least-squares von Mises pdf fit on top of the polar histogram (there are no robust fits yet for these distros, but if we add those later then best to have the same **-N0**|**1**|**2** syntax as in histogram).

_Originally posted by @PaulWessel in https://github.com/GenericMappingTools/gmt/issues/5056#issuecomment-811579882_

PaulWessel:

This PR adds the **-N**_mode_[**+p**_pen_] option to rose so that the best-fitting von Mises distribution can be plotted when -A is used and angular range is 360.  At the present only *mode == 0* is supported, and to preserve backwards compatibility (for now) with the old **-N** option the mode argument is required.  I am leaving other modes available for future expansion.

A new test was added for this.  Top plot uses a synthetic data derived from **gmt math VPDF** as input (so giving an exact fit for **-N**) and the other uses the standard data in that test dir used for other tests. 

sector_vpdf

 Also fixed minor issues in histogram docs related to the default pen for **-N**.  Closes #5064.

This is the last "new feature" PR from me for 6.2.  We will try to wrap up animation 14 and maybe a bug fix or two before RC1?

### DIFF
--- a/src/psrose.c
+++ b/src/psrose.c
@@ -89,9 +89,11 @@ struct PSROSE_CTRL {	/* All control options for this program (except common args
 		bool active;
 		struct GMT_SYMBOL S;
 	} M;
-	//struct PSROSE_N {	/* -N [Deprecated to -S+a] */
-	//	bool active;
-	//} N;
+	struct PSROSE_N {	/* -N[<kind>]+p<pen>, <kind = 0 (for now) OR -N [Deprecated to -S+a] */
+		bool active;
+		bool selected;
+		struct GMT_PEN pen;
+	} N;
 	struct PSROSE_Q {	/* -Q<alpha> */
 		bool active;
 		double value;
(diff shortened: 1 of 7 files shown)


src/psrose.c
```diff
@@ -949,6 +976,26 @@ EXTERN_MSC int GMT_psrose (void *V_API, int mode, void *args) {
 		}
 	}
 
+	if (Ctrl->N.active) {	/* Draw best-fitting probability distribution curve */
+		unsigned int k, NP = 361;
+		double a, mu, kappa, pdf, inc = 360.0 / (NP - 1), scl = 2.0 * area * Ctrl->A.inc * D2R;
+		double *xp = gmt_M_memory (GMT, NULL, NP, double);
+		double *yp = gmt_M_memory (GMT, NULL, NP, double);
```
joa-quim, line 983: It's normally not a good idea to do declaration and assignment in other then last one. VS uses to error on that.
Didn't test it yet though.

PaulWessel: These are all the same though.  What is not OK (but some compilers allow) is a mix of declarations and actions:

```
double x = 55;
if (p > 5.0) x += 9;
int unused = 99;
...
```

joa-quim: It was not that type things that I was referring. Still think we should maintain the code style
```
double *xp, *yp;

xp = gmt_M_memory (GMT, NULL, NP, double);
yp = gmt_M_memory (GMT, NULL, NP, double);
```
But as is, VS2019 accepts it.

joa-quim (approved):

src/psrose.c
PaulWessel: Sure, and *.c is full of these things - has been for years.  Valid syntax in C.

PaulWessel:
@meghanrjones, one more things for the ChangeLog I guess.

meghanrjones:
> @meghanrjones, one more things for the ChangeLog I guess.

I will update it now.

PR #5092, opened 2021-04-05
Ensure geographic datasets passed by memory files are seen as geographic

PaulWessel:
Externals (in this case [PyGMT](https://github.com/GenericMappingTools/pygmt/issues/1172)) may pass data into GMT modules.  These may be created and data added, and the creator may know the data are geographic and may flag that via the bit-flag **GMT_GRID_IS_GEO** in _GMT_Create_Data_.  Unfortunately, this was not honored.  Furthermore, this concept should not be restricted to grids: The user may know that the created dataset, grid, image, matrix, or vector is geographic and would like to pass that information into the data structure.  This PR allows for that while fixing the grid-specific bug.  Now,

1. **GMT_GRID_IS_GEO** is deprecated and we use **GMT_DATA_IS_GEO** instead for all the above data types (the old enum is kept for backwards compatibility)
2. The bit-flag, if set, calls _gmt_set_geographic_ for the chosen direction (input or output source)
3. When _GMT_Create_Data_ builds the various structures we check on that and add a _grdtype_ flag for grids, images, matrices (same value and meaning as in hidden grid header) and a _geographic_ = 0|1 flag for datasets and vectors
4. When _gmt_init_module_ checks if an input file (if found) is a remote data set (and hence sets geographic) we also check if given a memory file to see if it is known to be geographic.
5. When we call _GMT_Read_Data_ and the sources is a memory file we are now able to process the hidden geographic information.

With these changes, a memory file passed that is known to be geographic will trigger an automatic geographic projection if none had been selected explicitly with **-J**.

**Note**: Tested with the original PyGMT problem.  I assume it will work similarly for images and grids, as well as data sets and vectors, but there are no such external tests (yet).  I also updated the API doc to mention the purpose of the **GMT_DATA_IS_DATA** flag under _GMT_Create_Data_.

### DIFF
--- a/doc/rst/source/api.rst
+++ b/doc/rst/source/api.rst
@@ -1301,7 +1301,9 @@ directly (they either expect **GMT_IS_GRID** or **GMT_IS_DATASET**, for instance
 The solution is to specify the container type the GMT module expects but add in the special
 flags **GMT_VIA_MATRIX** or **GMT_VIA*VECTOR**.  This will create the **GMT_IS_MATRIX** or
 **GMT_IS_VECTOR** container the user needs to add the user data, but will also tell GMT how
-they should be considered by the module.
+they should be considered by the module. **Note**: When creating your own geographic data
+(dataset, grid, image, matrix, or vector) you may add ``GMT_DATA_IS_GEO`` to ``mode`` so that
+the structure that is created will retain this information.
 
 For grids and images you may pass ``pad`` to set the padding, or -1 to
 accept the prevailing GMT default. The ``mode`` determines what is actually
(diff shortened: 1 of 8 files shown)


seisman (approved):
I can confirm that it solves the initial PyGMT problem.

> GMT_GRID_IS_GEO is deprecated and we use GMT_DATA_IS_GEO instead for all the above data types (the old enum is kept for backwards compatibility)

Do you mean that PyGMT should pass `GMT_GRID_IS_CARTESIAN` for a Cartesian grid, and `GMT_DATA_IS_GEO` for a geographic grid?

PaulWessel:
No need to pass GMT_GRID_IS_CARTESIAN which is 0 so I dont think that would work - it is sort of the default.  So either GMT_DATA_IS_GEO or nothing.  It probably should not have been added to gmt_resources.h but...

PR #5100, opened 2021-04-05
The gmtlib_validate_id function stepped on another object's family

PaulWessel:
How rude!  Nasty business, such squabbles.  The if-tests in _gmtlib_validate_id_ were incorrect and we ended up changing a grid object's type to dataset because its actual family was matrix.  But that should only happen if the stated family is neither grid nor image.

This PR should fix the PyGMT problem reported [here](https://github.com/GenericMappingTools/pygmt/issues/1177); please check and then approve if fixed.


seisman (approved):
I can confirm that it fixes https://github.com/GenericMappingTools/pygmt/issues/1177

PaulWessel:
A small step for a man, a giant leap for ...

PR #5107, opened 2021-04-06
Use a separate grid index for area grid in grdfilter

PaulWessel:
See this [post](https://github.com/GenericMappingTools/pygmt/issues/859) for details.  The actual problem was that there was an assumption that the input grid and the area-per-node grid could share the same grid _index_, but in this case the input grid was a matrix with no pad while the area grid is created internally and has the regular pad.  Hence different indices are needed.  Now works for me with the pyGMT exampe as well as passing the usual CLI tests.


seisman (approved):
It works.

PR #5111, opened 2021-04-08
Do not use GMT's memory allocation for what GMT_Encode_Options returns

PaulWessel:
Since this function is only called by some externals (Mex and Julia I think) that only include gmt.h, they have no way to properly free this memory and notify GMT's garbage man.  Using calloc, realloc, and free instead.


PaulWessel:
Ping @joa-quim

joa-quim (approved):

PR #5115, opened 2021-04-08
Stronger checks before running build-release.sh

PaulWessel:
Enforce that gcc9-mp, gnutar etc are all available, plus add _meghanj_ as user able to ftp to SOEST pwessel/release directory.

### DIFF
--- a/admin/build-release.sh
+++ b/admin/build-release.sh
@@ -10,6 +10,8 @@
 #	4) sphinx-build 
 #	5) grealpath (package coreutils)
 #	6) GNU tar (package gnutar)
+#	7) gcc-mp-9 must be installed and in path
+#	8) g++-mp-9 must be installed and in path
 
 reset_config() {
 	rm -f ${TOPDIR}/cmake/ConfigUser.cmake
@@ -46,6 +48,36 @@ fi
 # pngquant is need to optimize images
 if ! [ -x "$(command -v gmt)" ]; then
 	echo 'build-release.sh: Error: pngquant is not found in your search PATH.' >&2
+	exit 1
+fi
+
+# sphinx-build is need to build the docs
+if ! [ -x "$(command -v sphinx-build)" ]; then
+	echo 'build-release.sh: Error: sphinx-build is not found in your search PATH.' >&2
+	exit 1
+fi
+
+# grealpath is need to build the dependency list of libs
+if ! [ -x "$(command -v grealpath)" ]; then
+	echo 'build-release.sh: Error: grealpath is not found in your search PATH.' >&2
+	exit 1
+fi
+
+# gnutar is need to build the tarballs
+if ! [ -x "$(command -v gnutar)" ]; then
+	echo 'build-release.sh: Error: gnutar is not found in your search PATH.' >&2
+	exit 1
+fi
+
+# gcc-mp-9 is need to build with OpenMP
+if ! [ -x "$(command -v gcc-mp-9)" ]; then
+	echo 'build-release.sh: Error: gcc-mp-9 is not found in your search PATH.' >&2
+	exit 1
+fi
+
+# gcc-mp-9 is need to build with OpenMP
+if ! [ -x "$(command -v g++-mp-9)" ]; then
+	echo 'build-release.sh: Error: g++-mp-9 is not found in your search PATH.' >&2
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


meghanrjones:
I think that this check should be improved because `cmake/` is also inside the build directories:
```
if [ ! -d cmake ]; then
	echo "build-release.sh: Must be run from top-level gmt directory" >&2
	exit 1
fi
```
An option would be to add a check for `admin/` and `cmake/`.

admin/build-release.sh
```diff
@@ -120,12 +162,19 @@ echo "build-release.sh: Report sha256sum per file" >&2
...
-echo "build-release.sh: Place gmt-${Version}-src.tar.* on the ftp site" >&2
-if [ -f gmt-${Version}-darwin-x86_64.dmg ]; then
-	echo "build-release.sh: Place gmt-${Version}-darwin-x86_64.dmg on the ftp site" >&2
-fi
-if [ "${USER}" = "pwessel" ]; then	# Place file in pwessel SOEST ftp release directory and set permissions
+
+# 10. Paul or Meghan may palce the candidate products on the pwessel/release ftp site
+if [ $do_ftp -eq 1 ]; then	# Place file in pwessel SOEST ftp release directory and set permissions
+	echo "build-release.sh: Placing gmt-${Version}-src.tar.* on the ftp site" >&2
 	scp gmt-${Version}-darwin-x86_64.dmg gmt-${Version}-src.tar.* ftp.soest.hawaii.edu:/export/ftp1/ftp/pub/pwessel/release
```
meghanrjones, line 169:
```suggestion
	scp gmt-${Version}-src.tar.* ftp.soest.hawaii.edu:/export/ftp1/ftp/pub/pwessel/release
```

admin/build-release.sh
```diff
@@ -120,12 +162,19 @@ echo "build-release.sh: Report sha256sum per file" >&2
...
+	if [ -f gmt-${Version}-darwin-x86_64.dmg ]; then
+		echo "build-release.sh: Placing gmt-${Version}-darwin-x86_64.dmg on the ftp site" >&2
+		scp gmt-${Version}-darwin-x86_64.dmg ftp.soest.hawaii.edu:/export/ftp1/ftp/pub/pwessel/release
+	fi
+	ssh ${USER}@ftp.soest.hawaii.edu 'chmod og+r /export/ftp1/ftp/pub/pwessel/release/gmt-*'
+else
+	echo "build-release.sh: Place gmt-${Version}-src.tar.* on the ftp site" >&2
+	if [ -f gmt-${Version}-darwin-x86_64.dmg ]; then
+		echo "build-release.sh: Place gmt-${Version}-darwin-x86_64.dmg on the ftp site" >&2
+	fi
```
meghanrjones, line 179:
```suggestion
```

admin/build-release.sh
```diff
@@ -87,12 +127,14 @@ mkdir build
...
 # 2c. Set CMake cache for MP build:
+COMPC=$(which gcc-mp-9)
+COMPG=$(which g++-mp-9)
 cat << EOF > cache-mp-gcc.cmake
 # Cache settings for building the macOS release with GCC and OpenMP
 # This cache file is set for the binary paths of macports
 #
-SET ( CMAKE_C_COMPILER "/opt/local/bin/gcc-mp-9" CACHE STRING "GNU MP C compiler" )
-SET ( CMAKE_CXX_COMPILER "/opt/local/bin/g++-mp-9" CACHE STRING "GNU MP C++ compiler" )
+SET ( CMAKE_C_COMPILER "${COMP}" CACHE STRING "GNU MP C compiler" )
```
meghanrjones, line 136:
```suggestion
SET ( CMAKE_C_COMPILER "${COMPC}" CACHE STRING "GNU MP C compiler" )
```

PaulWessel:
Sorry about not trying it once first myself...

meghanrjones:
These improvements seem good but the only way that I could get cpack to work was to use an absolute path for `CMAKE_INSTALL_PREFIX` in `admin/ConfigReleaseBuild.cmake`.

I also needed to change `/opt/bin/*` and `/opt/share/*` to `/opt/local/bin` and `/opt/local/share/` in `admin/build-macos-external-list.sh`. Has this been tested with Homebrew or Fink?

seisman:
> These improvements seem good but the only way that I could get cpack to work was to use an absolute path for `CMAKE_INSTALL_PREFIX` in `admin/ConfigReleaseBuild.cmake`.

I have the same problem, but the relative path works for Paul. I don't know why it differs, maybe the default CMake configurations of Homebrew and MacPorts are difference.

> 
> I also needed to change `/opt/bin/*` and `/opt/share/*` to `/opt/local/bin` and `/opt/local/share/` in `admin/build-macos-external-list.sh`. Has this been tested with Homebrew or Fink?

It won't work for Homebrew, because the Ghostscript from Homebrew doesn't allow packaging Ghostscript in bundles.

meghanrjones (approved):
@PaulWessel, I think it would be worth putting a note in the header that CMAKE_INSTALL_PATH, EXEPLUSLIBS, and EXESHARED may need to be changed for different users. Otherwise seems good.

PaulWessel:
I added that comment.

PR #5116, opened 2021-04-08
Failed to detect lack of modifier for -E in GMT_Encode_Options

PaulWessel:
The **grd2cpt -E** option takes an optional **+f**_file_ modifier, and for externals it is possible to just give **+f** and _GMT_Encode_Options_ will fill in the memory reference.  However, in this case **-**E was given with no argument yet the API failed to detect this case and added a memory reference.  This PR is more careful to check that an option that may take modifiers did in fact do so before we consider this hookup.  Works for me wiht a Julia example but need @joa-quim to run the full Julia tests to see if anything else broke.  I will run the mex 2017 paper tests.
PS. I am not sure how this regression came to be since it worked for the 2017 mex paper, but we must have added form sutbleties in dealing with modifiers and introduced this bug then.,

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12990,7 +12990,7 @@ struct GMT_RESOURCE * GMT_Encode_Options (void *V_API, const char *module_name, 
 		else info[ku].pos -= nn[info[ku].direction][K_SECONDARY];	/* Move any primary objects to start of list for this direction */
 	}
 	GMT_Report (API, GMT_MSG_DEBUG, "GMT_Encode_Options: Found %d inputs and %d outputs that need memory hook-up\n", input_pos, output_pos);
-#if DEBUG
+#if 0
 	{	/* Left here for simple debugging with messages on the GMT side */
 		text = GMT_Create_Cmd (API, *head);
 		GMT_Report (API, GMT_MSG_NOTICE, "GMT_Encode_Options: Revised command before memory-substitution: %s\n", text);
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12357,7 +12357,7 @@ GMT_LOCAL int gmtapi_extract_argument (char *optarg, char *argument, char **key,
 			code[1] = key[k][K_MODIFIER];
 			if ((c = strstr (optarg, code))) {	/* Found +<modifier> */
 				strcpy (argument, optarg);
-				if (!c[2]) *takes_mod = 2;	/* Flag that option had what KEY was looking for */
+				if (!c[2] || c[2] == '+') *takes_mod = 2;	/* Flag that option had no argument that KEY was looking for */
 				pos = (unsigned int) (c - optarg + 2);	/* Position of this modifier's argument. E.g., -E+f<file> will have pos = 2 as start of <file> */
 			}
 			else	/* No modifier involved */
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12988,7 +12988,7 @@ struct GMT_RESOURCE * GMT_Encode_Options (void *V_API, const char *module_name, 
 		else info[ku].pos -= nn[info[ku].direction][K_SECONDARY];	/* Move any primary objects to start of list for this direction */
 	}
 	GMT_Report (API, GMT_MSG_DEBUG, "GMT_Encode_Options: Found %d inputs and %d outputs that need memory hook-up\n", input_pos, output_pos);
-#if 0
+#if DEBUG
 	{	/* Left here for simple debugging with messages on the GMT side */
 		text = GMT_Create_Cmd (API, *head);
 		GMT_Report (API, GMT_MSG_NOTICE, "GMT_Encode_Options: Revised command before memory-substitution: %s\n", text);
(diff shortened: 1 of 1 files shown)


joa-quim:
You can run the Julia tests easily too. Just do
```
] test GMT
```
but I'm getting a crash when accessing the pointer returned by ``GMT_Encode_Options()`` so it looks that this is caused by the alloc issue.

joa-quim:
Confirmed, #5111 must be reverted. Don't understand why it causes the crash but it does.

joa-quim:
This branch crashes Julia. Before the dying scream I now see this message
```
GMT [NOTICE]: GMT_Encode_Options: Revised command before memory-substitution: -R0/1/0/1 -JX0.001c -T -O ->>C:\TMP\GMTjl_tmp.ps --PS_MEDIA=11920x16850
```

PaulWessel:
What are you running?

PaulWessel:
I got past WL 2 now but crash in WL 3 so perhaps you ahve that one in JL too?

joa-quim:
```
julia> gmt("psxy -R0/1/0/1 -JX0.001c -T -O")
GMT [NOTICE]: GMT_Encode_Options: Revised command before memory-substitution: -R0/1/0/1 -JX0.001c -T -O ->?
```
and now it even hanged running the command with one core at 100%

PaulWessel:
Well, I failed to merge in your fix from master.  Now done.  That was the problem for me and probably for you.

joa-quim:
Yes, it was. Now tests pass but you have to delete this
```
GMT [NOTICE]: GMT_Encode_Options: Revised command before memory-substitution:
```
it pops out at every command.

PaulWessel:
Yes I will, but note it only pops up because you and I have set -DDEBUG in the compiler switches, yes?

joa-quim:
Have ``MEM_DEBUG`` (switched it on recently)

PaulWessel:
I will comment that out as it was earlier.  I tried to run the Julia tests but

```
┌ Info: GDAL 3.2.2, released 2021/03/05
│ 205 GDAL drivers found
└ 78 OGR drivers found
GDAL Dataset (Driver: MEM/In Memory Raster)
File(s): 

Dataset (width x height): 241 x 181 (pixels)
Number of raster bands: 1
  [1] Band 1 (Undefined): 241 x 181 (Float64)
ERROR 5: Output driver `AAIGrid' not recognised.
ERROR 10: Pointer 'hDS' is NULL in 'GDALGetRasterBand'.

ERROR 10: Pointer 'hBand' is NULL in 'GDALGetRasterDataType'.

ERROR 10: Pointer 'hBand' is NULL in 'GDALGetRasterBandXSize'.

ERROR 10: Pointer 'hBand' is NULL in 'GDALGetRasterBandYSize'.

GDAL: Error During Test at /Users/pwessel/.julia/packages/GMT/GWSnp/test/test_gdal.jl:76
```

PaulWessel:
Let me know if tests works for you and we can accept this one.  Q: How to I tell matlab to find gs? I ran WL_3 to the end but it could not find gs.

joa-quim:
WTF? I get only this GDAL complain that must ivestigate but is not serious enough to make the tests fail. And they pass on the Ubuntu CI machine
```
GDAL Dataset (Driver: GeoJSON/GeoJSON)
File(s):
  point.geojson

Number of feature layers: 1
  Layer 0: point (Point)
ERROR 10: Pointer 'hBand' is NULL in 'GDALGetRasterDataType'.

Geometry: POLYGON ((1198554.34 648493.09,1198553.65476738 64 ... .09))Test Summary: | Pass  Total
GDAL          |   16     16
ERROR 1: Failed to write .vrt file in FlushCache().
```

PaulWessel:
Nevermind on the gs.  I think all the mex paper examples work now.

joa-quim:
and I have
```
┌ Info: GDAL 3.3.0dev, released 2021/99/99
│ 227 GDAL drivers found
└ 90 OGR drivers found
```

PaulWessel:
If you can run your Julia tests then all is well.  our 2017 paper scripts work for mex.  If you find the same then perhaps approve this PR so we can check the gmtmex box?

joa-quim:
I don't have them in the form of tests but my half broken set of gallery tests seem to be fine too.

But when you run this in the shell, don't you see this?
```
gdal_translate --formats | sort
  AAIGrid -raster- (rwv): Arc/Info ASCII Grid
```

joa-quim (approved):

PaulWessel:
No, but remember macports has this idiotic system of flavors and one has to add a bunch of options, such as to get openjpeg, because I assume the macports developers are working on 500 Gb hard drives that they are terrified of filling up.

joa-quim:
But that driver is by default. I don't find anyway of selecting/deselecting it. 
Why you keep using that stupid Macports?

PaulWessel:
Because I am stupid.  Also, so we can include gs and libgs in the bundle.

joa-quim:
Wrapped that gdal test in a try-catch. Update (``] up``) and maybe the tests now run.

PaulWessel:
Later, guests and heading to KoOlina for drinks

PR #5122, opened 2021-04-09
Set rc1

PaulWessel:

Temporarily changes settings to allow for release rc1.

### DIFF
--- a/cmake/ConfigDefault.cmake
+++ b/cmake/ConfigDefault.cmake
@@ -52,7 +52,7 @@ set (GMT_PACKAGE_VERSION_MAJOR 6)
 set (GMT_PACKAGE_VERSION_MINOR 2)
 set (GMT_PACKAGE_VERSION_PATCH 0)
 # If this is a beta version or similar, add a string suffix
-set (GMT_PACKAGE_VERSION_SUFFIX "-rc1")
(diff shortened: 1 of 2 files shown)


cmake/ConfigDefault.cmake
```diff
@@ -52,12 +52,12 @@ set (GMT_PACKAGE_VERSION_MAJOR 6)
 set (GMT_PACKAGE_VERSION_MINOR 2)
 set (GMT_PACKAGE_VERSION_PATCH 0)
 # If this is a beta version or similar, add a string suffix
-# set (GMT_PACKAGE_VERSION_SUFFIX "")
+set (GMT_PACKAGE_VERSION_SUFFIX "-rc1")
```
seisman, line 55: I think we should set it to `rc1`, without the hyphen, so that the package name is `gmt-6.2.0rc1-xxx`, consistent with our old rc releases.
```suggestion
set (GMT_PACKAGE_VERSION_SUFFIX "rc1")
```

meghanrjones (approved):

cmake/ConfigDefault.cmake
PaulWessel: Done

seisman (approved):

PR #5124, opened 2021-04-10
Let -Jp|P optionally set what angles"w/e" really are

### LINKED ISSUE #5120: Failed to specify the suffix for latitudes (i.e. N/S) in the polar coordinate system (-Jp, -JP)
I wanted to plot data that are given in polar coordinates (latitude, radius), so I chose the linear projection `-JP`. But I encountered the following problems:
1. The angle was treated as longitude by default rather than latitude, so the suffix was W or E. 
2. The options **+l**_label_ and  **+u**_unit_ didn't work under the projection with polar coordinates `-JP`.

Here is a minimum working example:
```
gmt begin Jp jpg E300
    gmt set FORMAT_GEO_MAP ddd:mm:ssF
    gmt basemap -JP5c+z+t0 -R-30/30/4371/6371 -Bxa15f+u" N" -Bya+l"Depth (km)" -BWNse
gmt end show
```
I have tried `-JP3c+z+t0+a`, but it didn't work.

**Actual outcome**
Jp

**Expected outcome**

I want to change W (E) to S (N), like this:
image12

**System information**

* Operating system: Ubuntu20.04

PaulWessel:
See #5120 for background.  Since we cannot tell what the angles are we default to great-circle angles (0-360 or -180/80 depending on **FORMAT_GEO_MAP** setting, and append the degree symbol) unless users add **+kx** to select longitude annotations or **+ky** for latitude annotations.  The latter also requires **-R** to be within -90/90 range.

I labelled this WIP for now since we have to decide if this is a bug fix or a feature that should be included in 6.2.0.

Closes #5120.


PaulWessel:
This PR yields these failures:

```
	480 - test/grdinterpolate/slices.sh (Failed)
	481 - test/grdinterpolate/slices_file.sh (Failed)
	645 - test/psbasemap/azim_radial.sh (Failed)
	665 - test/psbasemap/polarapex.sh (Failed)
	666 - test/psbasemap/polarcases.sh (Failed)
	667 - test/psbasemap/polarcircles.sh (Failed)
```
because they previously all were assumed to be longitudinal crossections for a north pole but in this PR we do not make that assumption.  So either (1) add **+kx** to these examples and then they will pass, or (2) update some scripts/plots to allow for **+kx** and **+ky** as well.

meghanrjones:
@PaulWessel, should I review this now or is more work needed first?

PaulWessel:
Not yet, working on the labels.

PaulWessel:
This PR is good to be reviewed.  It contains these changes:

1. Fixes the bug that angles were always assumed to be longitudes.  We do so by offering new modifier **+k**_x_|_y_ to select longitudes, latitudes [Default is great circle angles]
2. Adds some checks based on this knowledge (e.g. if **+k**_y_ then angle range must not exceed ±90).
3. Adds documentation to this effect and updates test plots that falsely used longitude annotations
4. Honors label annotation for the radial direction (depth or radius, for instance) which had been left off earlier since the annotation machinery used is that for geographic maps.
5. Adds three new **psbasemap** tests to ensure the radial labels are placed correctly for various **-JP** modifier settings.

All tests pass for me.

PaulWessel:
Just pinging @joa-quim @meghanrjones @seisman about this finalized PR since it has been awhile.

joa-quim (approved):

PR #5131, opened 2021-04-12
Features.rst Issues in escape sequences examples

Esteban82:
In the escape sequence section there is a list with some example. The last is "Acceleration (ms-2)." I add the backslash so the brackets are shown when the code is used (I try it) and complete the result with the "(ms@+-2@+)" (not sure if I edit it well).

Also, in the previous example (Manoa Straße) it is not clear what the escape sequences does in "Manoa". It was  Månoa?

### DIFF
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -1622,7 +1622,7 @@ of several types. The escape sequences recognized in GMT are listed in
 Table :ref:`escape <tbl-escape>`. Only one level of sub- or superscript is supported.
 Note that under Windows the percent symbol indicates a batch variable,
 hence you must use two percent-signs for each one required in the escape
-sequence for font switching.
+sequence for font switching. In bash scripts the brackets have special meaning, hence you must add double quotes.
 
 .. _tbl-escape:
 
@@ -1697,7 +1697,7 @@ GMT strings using the Standard+ encoding:
 | ``Stresses are @~s@~@+*@+@-xx@- MPa`` = Stresses are :math:`\sigma^{*}_{xx}` MPa
 | ``Se@nor Gar@con`` = Señor Garçon
 | ``M@!\305anoa stra@se`` = Manoa straße
-| ``A@\#cceleration@\# \(ms@+-2@+\)`` = ACCELERATION (ms\ :math:`^{-2}`)
+| ``A@\#cceleration@\# (ms@+-2@+)`` = ACCELERATION (ms\ :math:`^{-2}`)
 
 The option in :doc:`/text` to draw a
 rectangle surrounding the text will not work for strings with escape
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -1697,7 +1697,7 @@ GMT strings using the Standard+ encoding:
 | ``Stresses are @~s@~@+*@+@-xx@- MPa`` = Stresses are :math:`\sigma^{*}_{xx}` MPa
 | ``Se@nor Gar@con`` = Señor Garçon
 | ``M@!\305anoa stra@se`` = Manoa straße
-| ``A@\#cceleration@\# (ms@+-2@+)`` = ACCELERATION (ms\ :math:`^{-2}`)
+| ``A@#cceleration@# (ms@+-2@+)`` = ACCELERATION (ms\ :math:`^{-2}`)
 
 The option in :doc:`/text` to draw a
 rectangle surrounding the text will not work for strings with escape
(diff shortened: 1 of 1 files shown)


seisman:
I think this one is correct. What do you get?
```
gmt basemap -R0/10/0/4 -Jx1c -Baf -BWSen+t"A@#cceleration@# (ms@+-2@+)" -png map
```
image

Esteban82:
Right, with " also work. I try without them.  In fact the first three commands work. The 4th (which is the one of the docs) doesn't. I am not sure which is the most appropriate. 

```
gmt begin Escape png
    gmt basemap -R0/10/0/4 -Jx1c -Ba
    echo "A@#cceleration@# (ms@+-2@+)"    | gmt text -F+cTL+f12p -Gwhite -W1
    echo A@#cceleration@# \(ms@+-2@+\)    | gmt text -F+cTR+f12p -Gwhite -W1
    echo "A@#cceleration@# \(ms@+-2@+\)"  | gmt text -F+cBR+f12p -Gwhite -W1
#    echo A@#cceleration@# (ms@+-2@+)     | gmt text -F+cBL+f12p -Gwhite -W1
gmt end show
```

Escape

PaulWessel:
I think the double-quote first case is the one we should recommend.  There are probably strange situations of scripts calling scripts passing args where one has to do something different, but we should not worry about that.

seisman:
Actually, `echo A@#cceleration@# (ms@+-2@+)` is invalid because `( )` has special meanings.

Esteban82:
Another option is 
`A@#cceleration@# "(ms@+-2@+)" `

Which should I use for the docs? The previous first one?

seisman:
I believe it depends on how the string is passed to GMT. If you save `A@#cceleration@# (ms@+-2@+)` (also x, y coordinates) to a text file, and let `gmt text` read input from this file, it should work.

However, when typing the command in a terminal or in a bash script, `A@#cceleration@# (ms@+-2@+)` won't work, because `( )` has special meanings. I don't use DOS, so I'm not sure what happens on Windows.

Strictly speaking, `A@#cceleration@# (ms@+-2@+)` is the only correct one, and we need double quotes on Linux simply because bash does more things before passing strings to GMT.

Esteban82:
> `I believe it depends on how the string is passed to GMT. If you save A@#cceleration@# (ms@+-2@+) (also x, y coordinates) to a text file, and let gmt text read input from this file, it should work.`

I am using bash in Ubuntu and I try the followings:

Don't work:
```
     echo 5 3 A@#cceleration@# (ms@+-2@+) > temp
     gmt text temp -F+f12p -Gwhite -W1 -R0/10/0/10 -Jx1c -Ba -png 
```
Work: 
```
     echo 5 3 A@#cceleration@# "(ms@+-2@+)" > temp
     gmt text temp -F+f12p -Gwhite -W1 -R0/10/0/10 -Jx1c -Ba -png 
```

> because ( ) has special meanings.

So I think we maybe could add a warning note for bash users.

seisman:
> Don't work:
> 
> ```
>      echo 5 3 A@#cceleration@# (ms@+-2@+) > temp
>      gmt text temp -F+f12p -Gwhite -W1 -R0/10/0/10 -Jx1c -Ba -png 
> ```

For this to work, you have to manually add the text to the temp file.

> So I think we maybe could add a warning note for bash users.

I think so, but I'm not sure if `A@#cceleration@# (ms@+-2@+)` works for DOS on Windows.

Esteban82:
> I'm not sure if A@#cceleration@# (ms@+-2@+) works for DOS on Windows.

I check and it works.

> For this to work, you have to manually add the text to the temp file.

Rigth too.

seisman:
To me, the best solution is to list `A@#cceleration@# (ms@+-2@+)` in the documentation, but add a note that, for bash we have to add double quotes due to the special meaning of `( )`.

Actually, since `@` has special meaning on Windows, we have to use `@@` instead. That's also why we should document the raw strings instead, since each scripting language has its own syntax and may need special handlding.

Esteban82:
> To me, the best solution is to list A@#cceleration@# (ms@+-2@+) in the documentation, but add a note that, for bash we have to add double quotes due to the special meaning of ( ).

I agree.

PaulWessel:
Note looks good, but the backslashes in front of the hash-tags need to be removed.  Remember, those are just examples of the syntax, not strings that can be copied/pasted as is into any particular file or command for some OS.  As it is, the string cannot be used without also escaping the parenthesis.  So just give the clean syntax `A@#cceleration@# (ms@+-2@+)`

Esteban82:
> but the backslashes in front of the hash-tags need to be removed. 

Mmm, ok, I could remove them but I think that they were already there. I just add the backslash before the brackects (and already removed them).

PaulWessel:
Ok, that may be true.  But since that section is "OS-free" it should not have any escapes of any sort.  Since this is your branch, perhaps you can do the edit?

PaulWessel (approved):

PR #5133, opened 2021-04-13
nearneighbor.rst. Set -N as optional.

Esteban82:
From the first example, it can be seen that -N is optional. I reorder the synopsis and description accordingly.


PaulWessel (approved):
Looks god, thanks!

PR #5147, opened 2021-04-16
Add generic GMT_Free function to API for freeing odd GMT memory

PaulWessel:
Such as what is returned by _GMT_Encode_Options_.  It expects the address of the pointer to the memory.  Currently the only case use is to free what _GMT_Encode_Options_ returns to an external environment.  Updated the docs as well returned _GMT_Encode_Options_ to using GMT memory management.  Fixed some odd return codes in _GMT_Destroy_Data_ as well.

_GMT_Free_ will next be used in gmtmex.c to free that memory, but I will add a version check since _GMT_Free_ is only 6.2.0 and onward.


joa-quim (approved):

PaulWessel:
Looks like the only failure in the checks is related to the GMT server which is still down and no word from IT yet.

PR #5179, opened 2021-04-29
Fix recently introduce memory leak

PaulWessel:
I had left some paths out of a function before it freed temporary memory.  This PR plugs the hole.


joa-quim (approved):

meghanrjones:
New compiler warning:
```
[101/275] Building C object src/CMakeFiles/gmtlib.dir/gmt_api.c.o
../src/gmt_api.c:855:10: warning: incompatible pointer to integer conversion assigning to 'int' from 'int (void *, unsigned int, const char *, ...)' [-Wint-conversion]
                        error = GMT_Report;
                              ^ ~~~~~~~~~~
```

PaulWessel:
I am constantly fooled by the mouse cursor to place it one line above where it should go, then copy/paste the wrong item.  I know what you are thinking (seriously, how hard can this be), but I blame Apple cursor design on this...

PR #5198, opened 2021-05-06
Let movie master frame accept a different dpu than the movie frames

### LINKED ISSUE #5195: Movie. Allow select resolution of Master frame
See [1639 in the forum](https://forum.generic-mapping-tools.org/t/movie-resolution-of-master-frame-m/1639) for background.

PaulWessel:
By default, a raster master frame has the same dimensions as the movie, but this modifier allows you to set the master dpu independently.  The default is the same as the movie frames.  Closes #5195.


Esteban82 (approved):

PaulWessel:
We will merge this to master after 6.2.0 release.

PR #5203, opened 2021-05-08
Fix grdinterpolate KEYS for external use

PaulWessel:
The **-G** option sets the output file name for grids, cubes, or datasets.  Hopefully addresses Julia problems?

### DIFF
--- a/src/grdinterpolate.c
+++ b/src/grdinterpolate.c
@@ -35,7 +35,7 @@
 #define THIS_MODULE_MODERN_NAME	"grdinterpolate"
 #define THIS_MODULE_LIB		"core"
 #define THIS_MODULE_PURPOSE	"Interpolate a 3-D cube, 2-D grids or 1-D series from a 3-D data cube or stack of 2-D grids"
-#define THIS_MODULE_KEYS	"<G{+,>?}"
+#define THIS_MODULE_KEYS	"<G{+,G?}"
 #define THIS_MODULE_NEEDS	""
 #define THIS_MODULE_OPTIONS	"->RVfn"
 


PaulWessel:
Ping @joa-quim to see if this worked.

joa-quim:
Nope, it crashes. You can try yourself
```
G = gmt("grdinterpolate ......./cube.nc -T4")
```

joa-quim:
And why **-E** says this?
```
grdinterpolate c:/progs_cygw/GMTdev/gmt5/master/test/grdinterpolate/cube.nc -E5/5/6/6+i0.5
grdinterpolate [ERROR]: Option -E requires either equidistant levels or resampling via -T
```

PaulWessel:
Maybe confusion.  The -E sets the equidistance along-surface spacing while -T sets equidistant down-the-stack sampling.  Since cube.nc is not equidistant you would have to specify -T to resample since we are making a grid.
z = 1, 2, 3, 5 ;

joa-quim:
Yes, confusing. To start with I don't understand why the coordinates of a track (that's what -E does?) need to be equidistant.

> Specify a crossectinonal profile via a file or from specified line coordinates and modifiers. If a file, it must be contain a single segment with either lon lat or lon lat dist records. These must be equidistant.

PaulWessel:
> Yes, confusing. To start with I don't understand why the coordinates of a track (that's what -E does?) need to be equidistant.

Oh yes you do.  The track becomes x and the sampling through the layers become y, and all GMT grids must be equidistant.

joa-quim:
OK, I start to understand that, but that's not all evident from the **-E** description nor the error message.

PaulWessel:
Maybe, but -E does start with this:

`Specify a crossectinonal profile via a file or from specified line coordinates and modifiers. If a file, it must be contain a single segment with either lon lat or lon lat dist records. These must be equidistant.`

joa-quim:
Which is already confusing because it makes think that the equidistant condition applies only to the case when the track is given by a file. And does not say that it must be equidistant in the  *z* coordinate.

PaulWessel:
This can certainly be improved - I agree.  Wonder if I shall venture to try to do this debug via Julia and see if the debug instructions for Julia are accurate and up-to-date.   Since the mex sometimes gives me lots of trouble.  Can you share again the Julia calls that failed?

joa-quim:
> G = gmt("grdinterpolate ......./cube.nc -T4")

Adapt for your path. Should be the same in MEX

joa-quim:
Hmmm, this time I can't catch the crash from Julia (simply does pops up the debuger) but can do it from Matlab. It crashes in the call (the **R** contents are crazy)
image

PaulWessel:
Follow GMT Julia debug instructions.  After "using GMT" I get some errors.  Ideas?

```
julia> using GMT
[ Info: Precompiling GMT [5752ebe1-31b9-557e-87aa-f909b540aa54]
ERROR: LoadError: LoadError: failed process: Process(`grep libproj`, ProcessExited(1)) [1]

Stacktrace:
  [1] pipeline_error(procs::Base.ProcessChain)
    @ Base ./process.jl:538
  [2] (::Base.var"#644#645"{Base.ProcessChain})()
    @ Base ./process.jl:319
  [3] iterate(itr::Base.EachLine{Base.PipeEndpoint}, state::Nothing) (repeats 2 times)
    @ Base ./io.jl:1014
  [4] _collect(cont::UnitRange{Int64}, itr::Base.EachLine{Base.PipeEndpoint}, #unused#::Base.HasEltype, isz::Base.SizeUnknown)
    @ Base ./array.jl:613
  [5] collect
    @ ./array.jl:602 [inlined]
  [6] #readlines#340
    @ ./io.jl:551 [inlined]
  [7] readlines(s::Base.OrCmds)
    @ Base ./io.jl:551
  [8] top-level scope
    @ ~/.julia/packages/GMT/WnkyZ/src/proj_utils.jl:5
  [9] include(mod::Module, _path::String)
    @ Base ./Base.jl:386
 [10] include(x::String)
    @ GMT ~/.julia/packages/GMT/WnkyZ/src/GMT.jl:1
 [11] top-level scope
    @ ~/.julia/packages/GMT/WnkyZ/src/GMT.jl:196
 [12] include
    @ ./Base.jl:386 [inlined]
 [13] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::Nothing)
    @ Base ./loading.jl:1213
 [14] top-level scope
    @ none:1
 [15] eval
    @ ./boot.jl:360 [inlined]
 [16] eval(x::Expr)
    @ Base.MainInclude ./client.jl:446
 [17] top-level scope
    @ none:1
in expression starting at /Users/pwessel/.julia/packages/GMT/WnkyZ/src/proj_utils.jl:5
in expression starting at /Users/pwessel/.julia/packages/GMT/WnkyZ/src/GMT.jl:1
ERROR: Failed to precompile GMT [5752ebe1-31b9-557e-87aa-f909b540aa54] to /Users/pwessel/.julia/compiled/v1.6/GMT/jl_gJ0U2Y.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::Base.TTY, internal_stdout::Base.TTY)
   @ Base ./loading.jl:1360
 [3] compilecache(pkg::Base.PkgId, path::String)
   @ Base ./loading.jl:1306
 [4] _require(pkg::Base.PkgId)
   @ Base ./loading.jl:1021
 [5] require(uuidkey::Base.PkgId)
   @ Base ./loading.jl:914
 [6] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:901
```

PaulWessel:
I completely removed ~/.julia and started over, same thing, something about libproj:

`ERROR: LoadError: LoadError: failed process: Process(`grep libproj`, ProcessExited(1)) [1]`

joa-quim:
Oh dear, why unix fs so much? That works in Linux and same logic works to find the GDAL shared lib but it's failing for you with proj.
What do you get with?
``otool -L **the gmt shared lib** | grep libgdal``

maybe you can go around that error if you comment the line 197 of GMT.jl        (include("proj_utils.jl")

PaulWessel:
I get

	`/opt/local/lib/libgdal.28.dylib (compatibility version 29.0.0, current version 29.2.0)`


BTW, with this branch things run in gmtmex:

```
G = gmt ('grdinterpolate', 'cube.nc -T4');
G = gmt ('grdinterpolate', 'cube.nc -E5/5/6/6+i0.5 -T1/5/1');
G = gmt ('grdinterpolate', 'cube.nc -E5/5/6/6+i0.5 -T1/5/1');
```

Sure you built the branch?

PaulWessel:
Commenting that line makes using GMT complete fine.

joa-quim:
> Sure you built the branch?

This branch only changed (which I did manually). It crashes Matlab too.

#define THIS_MODULE_KEYS	"<G{+,G?}"

joa-quim:
> I get

> `/opt/local/lib/libgdal.28.dylib (compatibility version 29.0.0, current version 29.2.0)`

Bloody MacPorts again? You don't have  a PROJ4 dependency? How's that possible? Or MacPorts named it differently?

PaulWessel:
That is strange since I just built it and tested it, in gmtmex, and it works.  I cannot debug Julia as I get spinning wheel in Xcode when I step into Call_Module.

PaulWessel:
Yes, macports. Its libgdal requires proj7:

```
otool -L /opt/local/lib/libgdal.28.dylib | grep proj

	/opt/local/lib/proj7/lib/libproj.19.dylib (compatibility version 22.0.0, current version 22.1.0)

```

joa-quim:
And does 

```G = gmt('grdinterpolate cube.nc", -S=5/5')```
return a filled Dataset?

**Comment by PaulWessel** [comment]
I assume you mean this:

```
>> D = gmt('grdinterpolate cube.nc -S5/5')

D = 

  struct with fields:

       data: [4x4 double]
       text: {4x1 cell}
     header: 'Location 5,5'
    comment: []
      proj4: []
        wkt: []

```

**Comment by PaulWessel** [comment]
All those commands I listed work just fine for me in matlab.

**Comment by joa-quim** [comment]
The file ``gdal.jl`` has
```
@static Sys.iswindows() ?
	(Sys.WORD_SIZE == 64 ? (const libgdal = "gdal_w64") : (const libgdal = "gdal_w32")) : (
		Sys.isapple() ? (const libgdal = Symbol(split(readlines(pipeline(`otool -L $(GMT.thelib)`, `grep libgdal`))[1])[1])) : (
			Sys.isunix() ? (const libgdal = Symbol(split(readlines(pipeline(`ldd $(GMT.thelib)`, `grep libgdal`))[1])[3])) :
			error("Don't know how to install this package in this OS.")
		)
	)
```
and it finds the gdal shared lib.

The file ``proj_utils.jl`` has 

```
@static Sys.iswindows() ?
	(Sys.WORD_SIZE == 64 ? (const libproj = "proj_w64") : (const libproj = "proj_w32")) : (
		Sys.isapple() ? (const libproj = Symbol(split(readlines(pipeline(`otool -L $(GMT.thelib)`, `grep libproj`))[1])[1])) : (
			Sys.isunix() ? (const libproj = Symbol(split(readlines(pipeline(`ldd $(GMT.thelib)`, `grep libproj`))[1])[3])) :
			error("Don't know how to use PROJ4 in this OS.")
		)
	)
```
and it works in Linux (the CI tests pass). Why is it failing in Mac?

**Comment by PaulWessel** [comment]
Looks like your are looking for a libproj entry in the libgmt.dylib file but there are none since GMT does not link with proj.  You would have to do the otool on libgdal I think.

**Comment by joa-quim** [comment]
> D = gmt('grdinterpolate cube.nc -S5/5')

That actually also works for me (must had screwed something in my previous attempt) but the other one crashes Ml

**Comment by joa-quim** [comment]
But one should get the dependencies of dependencies too, or doesn't ``otool`` do that?

**Comment by PaulWessel** [comment]
I dont really know what otool does; perhaps there is an option to follow into libraries for other libraries?
BTW, I cannot make grdinterpolate break with these cases in matlab.  I think the KEYS correct since -G is used to name the output in all cases.  So I will remove the WIP, but I am not sure what is happening on your end.

**Comment by joa-quim** [comment]
My guess is that VS is catching another error that Xcode is closing the eyes to.

**Comment by PaulWessel** [comment]
Can you step up a bit to see where this is called, etc?

**Comment by PaulWessel** [comment]
None of the row-by-row stuff should be in effect in grdinterpoalte (only used in grdblend I think) so I expect that pointer to be NULL.  Hm, where can it get set to non-NULL I wonder.

**Comment by PaulWessel** [comment]
Is GH->extra not NULL?

**Comment by joa-quim** [comment]
The ``gmtlib_free_grid_ptr()`` is called some 7 times an in all but the last ``GH->extra`` is NULL. However, in the last call, called by ``GMT_Destroy_Data``, comming from I don't know where because the stack-trace does not show ``grdinterpolate`` ascendancy, it has garbage.

Have to stop here today.

**Comment by PaulWessel** [comment]
OK, odd.

**Comment by joa-quim** [comment]
So the problem is that ``grdinterpolate`` calls 
```
GMT_Destroy_Data (API, &C[GMT_OUT]);	/* Done with the output cube */
```
and and the MEX too in
```
	/* 8. Free all GMT containers involved in this module call */

	for (k = 0; k < n_items; k++) {
		void *ppp = X[k].object;
		if (GMT_Close_VirtualFile (API, X[k].name) != GMT_NOERROR)
			mexErrMsgTxt ("GMT: Failed to close virtual file\n");
		if (GMT_Destroy_Data (API, &X[k].object) != GMT_NOERROR)
```
and it crashes there. Why not in Mac?

**Comment by PaulWessel** [comment]
Well, not sure, and since I cannot actually debug this shit anymore... But if the object is freed and NULL-ed I wonder why GMT_Destroy_Data does not simply skip it?

**Comment by PaulWessel** [comment]
is X[k].object NULL? If allready freed, should it not be? The address passed is always non NULL but I thought there is a check somewhere if we pass an address whose *address == NULL?

**Comment by joa-quim** [comment]
No, ``X[k].object`` is not NULL, ``GMT_Destroy_Data()`` frees memories but ``GH->extra`` has garbage and crash follows.

**Comment by PaulWessel** [comment]
The Destroy of the cube in grdinterpolate is normal procedure.  We have written (to file or memory) and in this case it is to a grid memory.

**Comment by PaulWessel** [comment]
In all previous cases like this that I have fixed, it turns out we are destroying an incorrect pointer hence strange things show up in its variables.  I am sure what you are looking at is not the structure but some other memory (of some other kind than a grid).  So it is likely a bug.  I guess I will have to try to rebuild everything with an older Xcode that used to let me debug...

**Comment by joa-quim** [comment]
Commenting out the calls to ``GMT_Destroy_Data()`` makes no difference. But in the call that crashes the data seems to have sense. At least ``G->x`` and ``G->y`` so the pointer is not pointing to so stray memory.

But rebuilding with an old Xcode brings what? The MEX works for you.

**Comment by PaulWessel** [comment]
Yes, it works, but perhaps by stepping into everything I could still learn that a wrong pointer is passed, even if it by accident happens to be NULL or whatever to avoid the crash.
I think you can approve the PR though as the KEYS are correct, they are not the source of this crash, so we can keep #5202 open.

**Review by joa-quim — APPROVED** [review]

**Comment by joa-quim** [comment]
In 
```
GMT_LOCAL int gmtapi_export_grid (struct GMTAPI_CTRL *API, int object_ID, unsigned int mode, struct GMT_GRID *G_obj) {
```
we have
```
S_obj->resource = G_obj;	/* Set resource pointer to the grid */
```
and ``GH`` of ``G_obj`` is already non-NULL so the shit's origin comes from  ``G_obj``

joa-quim:
No need to ``add`` again and again. 
```] update```

PR #5204, opened 2021-05-09
Better defense against filenames with spaces

### LINKED ISSUE #5119: grdimage: Use of grid names with space and -I+d
See example below.
The initial call to ``grdimage`` works fine due to the file name being contained within ``'`` or ``"``. The subsequent call to ``grdgradient`` apparently does not do this, and the file is not found. 

```
gmt grdcut -RNO @earth_relief_10m -G"stupid file name.nc"
gmt grdimage 'stupid file name.nc'
gmt grdimage 'stupid file name.nc' -pdf f -I+d
grdgradient [ERROR]: Cannot find file stupid
```

PaulWessel:

Some users, for whatever reason, end up creating or passing filenames that contain spaces.  This means they need to use quotes when passing these on the command line to GMT.  However, many GMT modules will create commands that involve these input file names and call _GMT_Call_Module_.  However, the filenames now have lost their quotes and the commands fail.  This PR makes an attempt to fix this:

1. Before a module calls _GMT_Call_Module_ with a filename, that filename is run via _gmt_filename_set_ which will replace any space with ASCII 29.
2. The _GMT_File_Path_ function now always does the reverse, replacing any ASCII 29 found with spaces
3. _spotter_init_ does the same thing in the spotter supplement.

Closes #5119 when we merge after 6.2.0 release, hence WIP.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -15145,6 +15145,8 @@ int GMT_Get_FilePath (void *V_API, unsigned int family, unsigned int direction, 
 
 	if ((mode & GMT_FILE_CHECK) == 0) gmt_set_unspecified_remote_registration (API, file_ptr);	/* Complete remote filenames without registration information */
 
+	gmt_filename_get (file);	/* Replace any ASCII 29 with spaces (if filename had spaces they may now be ASCII 29) */
+
 	switch (family) {
 		case GMT_IS_GRID:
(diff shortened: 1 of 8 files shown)


anbj (approved):
Tested, and works!

PaulWessel:
Wonder if we should add this before releasing rc2 since it is in fact a bug fix, no?

seisman:
> Wonder if we should add this before releasing rc2 since it is in fact a bug fix, no?

Yes to me.

Issue #5202, opened 2021-05-08
grdinterpolate not working from externals


joa-quim:
Should have tried this before.
```
julia> G = grdinterpolate("c:/progs_cygw/GMTdev/gmt5/master/test/grdinterpolate/cube.nc", T=4, Vd=1)
        grdinterpolate c:/progs_cygw/GMTdev/gmt5/master/test/grdinterpolate/cube.nc  -n+a -T4
grdinterpolate [ERROR]: Option -G: Must specify output grid file
```
and
```
julia> G = grdinterpolate("c:/progs_cygw/GMTdev/gmt5/master/test/grdinterpolate/cube.nc", T=4, G="lixo.grd", Vd=1)
        grdinterpolate c:/progs_cygw/GMTdev/gmt5/master/test/grdinterpolate/cube.nc  -n+a -Glixo.grd -T4
ERROR: get_grid: programming error, output matrix is empty
```
Changing the KEYs to ``"<G{+,GG}" lets last command work but then the first one (without the G) crashes.

PaulWessel:
Will check after tennis

-- 
Paul Wessel, Professor and Chair
Dept. of Earth Sciences
SOEST, U of Hawaii at Manoa

PaulWessel:
Apparently freeing the wrong memory but does not crash on macOS, making it hard to debug for me.

PaulWessel:
Just to keep something in this issue as well, this apparently crashes Matlab on WIndows:

`G = gmt('grdinterpolate cube.nc -T4');`

with cube.nc found in the test/grdinterpolate folder.

PaulWessel:
Hi @joa-quim, does it also crash if we actually request a subset cube instead of a single grid layer?

`C = gmt('grdinterpolate cube.nc -T2/4/0.5');`

PaulWessel:
I am thinking that we are writing a cube with one layer, so a grid, and perhaps there is confusion in that.  I do not think (not checked yet) that we switch the KEYS from cube to grid if only one layer is found, and I am not sure we can since that is detected much later in the processing.

PaulWessel:
OK, I think I udnerstand the problem, at least for the example above: The grdinterpolate cube.nc -T4 command is just pulling out a grid lice for z = 4, but the module will call 

```
[...]

PR #5220, opened 2021-05-16
Set the correct return type for grdinterpolate for external calls

PaulWessel:

See #5202 for context.  One key problem was the case when we ran commands like `G = grdinterpolate("cube.nc", T=4)` in gmtmex or Julia it would crash because the KEYS in this case was changed to expect a **GMT_GRID** structure (single layer) but in fact we are still returning a **GMT_CUBE** structure.  The mismatch of the very similar structures caused crashes in one OS but not another.  This PR fixes this.  In the process I also fix a memory leak (both CLI and externals see this)

`Matlab (gmt_grdio.c:3499(gmtlib_free_cube_ptr)): Wrongly tries to free item
`

I am separately updating gmtmex to detect when the returned cube has a single layer only and then return a grid instead, but this switcharoo has to take place in the wrappers, not the GMT API.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12444,28 +12444,21 @@ GMT_LOCAL bool gmtapi_operator_takes_dataset (struct GMT_OPTION *opt, int *geome
 }
 
 GMT_LOCAL char gmtapi_grdinterpolate_type (struct GMTAPI_CTRL *API, struct GMT_OPTION *options) {
-	unsigned int n_files = 0;
-	char type = 'D';	/* For -S */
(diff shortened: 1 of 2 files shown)


PaulWessel:
Spoke too soon about

Matlab (gmt_grdio.c:3499(gmtlib_free_cube_ptr)): Wrongly tries to free item

but the type is now set correctly.  Will rename to WIP since not quite ready yet.

joa-quim (approved):

PaulWessel:
OK, the message from _gmtlib_free_cube_ptr_ was due to my error in gmtmex_parser.c and not in GMT API, so took off the WIP and will now merge.

PR #5231, opened 2021-05-18
Fix bug in project with -N and update docs

meghanrjones:

This PR fixes a bug in project reported in [this forum post](https://forum.generic-mapping-tools.org/t/using-project-in-cartesian-coordinate/1680/12). A minimal example to reproduce the bug is:
```
echo "0 0" | gmt project -C-10/0 -E10/0 -N -Fxypq
```
where the expected result is
```
0 0 10 0
```
and the actual result is
```
0	0	6.12323399574e-16	10
```

`test/project/cartesian.sh` fails with this PR, so still need to 1) confirm whether the original test result was in error and 2) update the test or revise bug fix.

Fixes #

**Reminders**

### DIFF
--- a/doc/rst/source/project.rst
+++ b/doc/rst/source/project.rst
@@ -15,11 +15,11 @@ Synopsis
 **gmt project** [ *table* ] |-C|\ *cx*/*cy* [ |-A|\ *azimuth* ]
 [ |-E|\ *bx*/*by* ] [ |-F|\ *flags* ]
 [ |-G|\ *dist*\ [/*colat*][**+c**\|\ **h**] ]
-[ |-L|\ [**w**\|\ *l\_min*/*l\_max*] ]
+[ |-L|\ [**w**\|\ *lmin*/*lmax*] ]
 [ |-N| ] [ |-Q| ] [ |-S| ]
 [ |-T|\ *px*/*py* ]
 [ |SYN_OPT-V| ]
-[ |-W|\ *w\_min*/*w\_max* ]
+[ |-W|\ *wmin*/*wmax* ]
 [ |-Z|\ *major*/*minor*/*azimuth*\ [**+e**\|\ **n**] ]
 [ |SYN_OPT-b| ]
 [ |SYN_OPT-d| ]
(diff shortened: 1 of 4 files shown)


PaulWessel:
Boy, it is hard on the brain to check but what is _poorly documented_ is that the **-A**_azimuth_ sets the direction of the positive p-axis [so _azimuth_ = 90 degrees is _p=x_ and _q=y_].  So 

```
echo 1 0 | gmt project -N -C0/0 -A90   
1	0	1	0	1	0
```

is good because _p_ aligns with _x_ and we get the same output as input.  Doing other multiples of 90 degrees, including 0 (which is a 90-degree CCW rotation then, gives the expected values.  So I think this is the correct fix, and we need to fix the cartesian.sh which places the "P" text at the negative (opposite) end of the _p_-axis and has the _q_-arrow and label on the wrong side.  The _p_-arrow is in the right place though.  The values look correct as far as I can tell.

meghanrjones:
> Boy, it is hard on the brain to check but what is _poorly documented_ is that the **-A**_azimuth_ sets the direction of the positive p-axis [so _azimuth_ = 90 degrees is _p=x_ and _q=y_]. So
> 
> ```
> echo 1 0 | gmt project -N -C0/0 -A90   
> 1	0	1	0	1	0
> ```
> 
> is good because _p_ aligns with _x_ and we get the same output as input. Doing other multiples of 90 degrees, including 0 (which is a 90-degree CCW rotation then, gives the expected values. So I think this is the correct fix, and we need to fix the cartesian.sh which places the "P" text at the negative (opposite) end of the _p_-axis and has the _q_-arrow and label on the wrong side. The _p_-arrow is in the right place though. The values look correct as far as I can tell.

I updated the "P" and "Q" text placement and pushed a new ps file for that test. I agree that some documentation improvements would be helpful, which can be addressed separately.

To be honest, I am having trouble wrapping my head around a cartesian vector projection (-N) using a rotation pole definition (-T and -C), which is currently untested. Any recommendations for this case, @PaulWessel?

PaulWessel:
Kneejerk reaction is that **-T** and **-C** cannot both be set.  There is no "pole" for the Cartesian case other than the origin, I think.  So origin and rotation only.  Thinking more, perhaps **-T** would set a 2nd point that then defines the azimuth from **-C** to **-T** and that is the extent of it.  I can see it may be more convenient to specify that point than to compute that azimuth.

PaulWessel:
Looking at cartesian.ps now.   I am pretty sure that -A takes azimuth for both Cartesian and geographic cases.  What you have done is consistent and it means that azimuth is the amount of rotation (N over E) we rotate the Q axis. I am OK with this but we need to document it.  Other than that it looks good.  Oh, since we are changing cartesian.ps anyway, maybe add a thin -Wfaint outline to the red circles so that the lower-left example is clear that the circle plots on top of the vector head?

meghanrjones:
> Kneejerk reaction is that **-T** and **-C** cannot both be set. There is no "pole" for the Cartesian case other than the origin, I think. So origin and rotation only. Thinking more, perhaps **-T** would set a 2nd point that then defines the azimuth from **-C** to **-T** and that is the extent of it. I can see it may be more convenient to specify that point than to compute that azimuth.

OK, so then **-E** and **-T** are the same for the Cartesian case. I think that should be specified in the docs as well.

PaulWessel:
ACtually, probably better to disallow -T for Cartesian since -C -E or -C -A are really the only viable setups, right?  Makes no point of having -T double as -E I think.

meghanrjones:
> ACtually, probably better to disallow -T for Cartesian since -C -E or -C -A are really the only viable setups, right? Makes no point of having -T double as -E I think.

I agree and will update code and documentation.

meghanrjones:
I cleaned up the docs a bit to make things more clear; it should be available for preview soon. I think this is a module where a couple simple plots could really help the explanation, perhaps we could add that to the list of documentation improvements for after 6.2.0.

PaulWessel:
Yep, agreed.

PaulWessel (approved):

PR #5236, opened 2021-05-19
Update INSTALL.md

rogerfraser:
The path needs to include the bin folder in which all the binaries are installed.


PaulWessel (approved):
Thanks!

PR #5238, opened 2021-05-20
Use default pen if not given in gmt_getpen

### LINKED ISSUE #5237: Broken PS file when setting dashed frame
Found this bug when trying to answer this [forum post](https://forum.generic-mapping-tools.org/t/how-to-change-the-width-and-color-of-the-box-frame-in-psxy-plot/1693/7).

To reproduce the issue:
```
gmt begin test png
gmt basemap -R1960/2000/0/30 -JX5/8 -Bxa10f5 -Bya5f1 --MAP_FRAME_PEN=red,-
gmt end show
```
The command crashes for me with the following errors:
```
Error: /rangecheck in --setdash--
Operand stack:
   --nostringval--   0
Execution stack:
   %interp_exit   .runexec2   --nostringval--   setdash   --nostringval--   2   %stopped_push   --nostringval--   setdash   setdash   false   1   %stopped_push   1974   1   3   %oparray_pop   1973   1   3   %oparray_pop   1961   1   3   %oparray_pop   1817   1   3   %oparray_pop   --nostringval--   %errorexec_pop   .runexec2   --nostringval--   setdash   --nostringval--   2   %stopped_push   --nostringval--   1818   2   4   %oparray_pop
Dictionary stack:
   --dict:736/1123(ro)(G)--   --dict:0/20(G)--   --dict:75/200(L)--   --dict:157/250(L)--
Current allocation mode is local
Last OS error: No such file or directory
Current file position is 21471
psconvert [ERROR]: System call [gs -q -dNOSAFER -dNOPAUSE -dBATCH -sDEVICE=bbox -DPSL_no_pagefill -dMaxBitmap=2147483647 -dUseFastColor=true '/Users/seisman/.gmt/sessions/gmt_session.66120/gmt_0.ps-' 2> '/Users/seisman/.gmt/sessions/gmt_session.66120/psconvert_68850c.bb'] returned error 256.
```

However, if I change `MAP_FRAME_PEN` to `1p,red,-`, it works.

PaulWessel:
Since we are passing in a pointer with the current pen we should let that pen thickness be the default if the user does not specify a new width, just color  and/or style.  Closes #5237.

PS. Labeled WIP until confirmed working by @seisman, @joa-quim  and @meghanrjones.  The reason is that when I ran this fix in Xcode debugger it worked just fine, but when I ran it with my command line version it fails for unclear reasons.  So please check.


PaulWessel:
I should point out that all tests pass so to me it seems like I am picking up an older libgmt.dylib somehow.

meghanrjones:
I get the same failure as posted in #5237 after building this branch.

PaulWessel:
Interesting isn't it? All tests pass (probably because none is doing this setting), and works in Xcode (correct PS) so hard to debug...

PaulWessel:
OK, has to do with NaN.  Just a minute.

PaulWessel:
Problem is that we cannot compute a style without knowing the pen width, and when in modern theme we set width to NaN then we cannot convert things like "-" to actual spacings.  So this fix is nontrivial since it would force us to delay the interpretation of style until after the pen has been computed for us.  That means we must wait until gmt_set_undefined_defaults to finalized the styles for all pens that had NaN as width.  Basically, I will need to let gmt_getpen leave the style as given (e.g., "-" or "dotted") and then finish the convertsion once the width is set.  I think I can do this in 30 minutes. OK?

PaulWessel:
OK, please try again.

PaulWessel:
Works for that case but tons of failures in the tests...

PaulWessel:
OK, some more progress.  This is a bit odd.  We may have had a subtle bug until this fix.  I am still getting 80 failures, all (or mostlly) subplots.  Take sphareacentroid.sh which plots two circular -JG plots with dimension 12 cm. Per the scaling law at line 9958-59 in gmt_init.c a 12 cm plot shall have a scale of ((2.0/15.0) * (12 - 10.0) + 9)/10 = 0.92666. So for a nominal frame pen of 1.5p we should get an actual frame pen of width 1.39p.  In PS the 1.39p is 1.39 / 72 * 1200 = 23 PS units. That is what the PS file has.  However, the original file has a PS pen thickness of 25 which is the original 1.5p pen.  So no shrinking - and hence that is a bug.  If I print out the scale when running this stripped down sphareacentroid.sh plot:

```
gmt begin sphareacentroid ps
  gmt subplot begin 2x1 -Fs12c -Bafg -Rg -M0 -Y1.25c
    gmt basemap  -JG0/10/? -c
    gmt basemap -JG0/-10/?  -c
  gmt subplot end
gmt end show
```
I get these scales:

Scale = 0.992941
Scale = 0.926666
Scale = 0.926666

There are three because subplot first needs to initialize pens even though it does not yet know the panel plot. Note that 1.5p times 0.992941 still gives 25 PS units.  Somehow the two plots end up using the 1.5p in master but correctly uses 1.39p in this branch.

Reporting this now as I need to do some more checking to understand that this is indeed a bug.

PaulWessel:
These scripts "fail" in this branch (80 in total):

```
The following tests FAILED:
	  3 - doc/scripts/GMT_-B_geo_2.sh (Failed)
	  5 - doc/scripts/GMT_-B_log.sh (Failed)
	  6 - doc/scripts/GMT_-B_pow.sh (Failed)
	 20 - doc/scripts/GMT_App_E.sh (Failed)
	 21 - doc/scripts/GMT_App_F_stand+_iso+.sh (Failed)
	 22 - doc/scripts/GMT_App_F_symbol_dingbats.sh (Failed)
	 23 - doc/scripts/GMT_App_G.sh (Failed)
	 32 - doc/scripts/GMT_App_M_1a.sh (Failed)
	 33 - doc/scripts/GMT_App_M_1b.sh (Failed)
	 34 - doc/scripts/GMT_App_M_1c.sh (Failed)
	 35 - doc/scripts/GMT_App_M_1d.sh (Failed)
	 54 - doc/scripts/GMT_agefig.sh (Failed)
	 75 - doc/scripts/GMT_chunking.sh (Failed)
	 87 - doc/scripts/GMT_cycle_6.sh (Failed)
	 99 - doc/scripts/GMT_grid2pix.sh (Failed)
	119 - doc/scripts/GMT_map_frame_type.sh (Failed)
	132 - doc/scripts/GMT_panel.sh (Failed)
	141 - doc/scripts/GMT_seamount_cum_inc.sh (Failed)
	188 - doc/examples/ex03/ex03.sh (Failed)
	191 - doc/examples/ex06/ex06.sh (Failed)
	199 - doc/examples/ex14/ex14.sh (Failed)
	201 - doc/examples/ex16/ex16.sh (Failed)
	229 - doc/examples/ex44/ex44.sh (Failed)
	230 - doc/examples/ex45/ex45.sh (Failed)
	283 - test/filter1d/varfilt.sh (Failed)
	320 - test/gmt_core/testsuite.sh (Failed)
	369 - test/gmtspatial/sphareacentroid.sh (Failed)
	375 - test/grd2cpt/equalarea.sh (Failed)
	443 - test/grdfilter/varfilter.sh (Failed)
	478 - test/grdinterpolate/fake_xyz_cube.sh (Failed)
	480 - test/grdinterpolate/slices.sh (Failed)
	481 - test/grdinterpolate/slices_file.sh (Failed)
	495 - test/grdmath/earth_shade.sh (Failed)
	572 - test/modern/box.sh (Failed)
	575 - test/modern/insetlegend.sh (Failed)
	576 - test/modern/longbasemap.sh (Failed)
	579 - test/modern/polarpanels.sh (Failed)
	583 - test/modern/somepanels.sh (Failed)
	584 - test/modern/sublegend.sh (Failed)
	586 - test/modern/vardims.sh (Failed)
	587 - test/modern/varfracs.sh (Failed)
	636 - test/psbasemap/JOb_pole.sh (Failed)
	637 - test/psbasemap/absrelaxis.sh (Failed)
	642 - test/psbasemap/asymm_grid.sh (Failed)
	643 - test/psbasemap/azim_global.sh (Failed)
	645 - test/psbasemap/azim_radial.sh (Failed)
	660 - test/psbasemap/nav_ticks.sh (Failed)
	665 - test/psbasemap/polarapex.sh (Failed)
	666 - test/psbasemap/polarcases.sh (Failed)
	667 - test/psbasemap/polarcircles.sh (Failed)
	670 - test/psbasemap/radial_labels1.sh (Failed)
	671 - test/psbasemap/radial_labels2.sh (Failed)
	672 - test/psbasemap/radial_labels3.sh (Failed)
	719 - test/pshistogram/altwidth.sh (Failed)
	733 - test/pslegend/autolegendcolor.sh (Failed)
	734 - test/pslegend/autolegendcolorL.sh (Failed)
	746 - test/pslegend/subplotlegend.sh (Failed)
	827 - test/psxy/autocolorpols.sh (Failed)
	830 - test/psxy/bar2dmultimode.sh (Failed)
	831 - test/psxy/bar2dsidebyside.sh (Failed)
	838 - test/psxy/cartellipse.sh (Failed)
	920 - test/psxy/scalesymb.sh (Failed)
	924 - test/psxy/snakes.sh (Failed)
	951 - test/psxyz/bar3dmultimode.sh (Failed)
	952 - test/psxyz/bar3dsidebyside.sh (Failed)
	965 - test/psxyz/scalesymb.sh (Failed)
	974 - test/sample1d/smooth.sh (Failed)
	975 - test/sample1d/smooth_deriv.sh (Failed)
	1018 - test/subplot/autocolorline.sh (Failed)
	1020 - test/subplot/linpanels.sh (Failed)
	1021 - test/subplot/linpanelsnew.sh (Failed)
	1022 - test/subplot/panelsinside.sh (Failed)
	1023 - test/subplot/subplot_1.0.sh (Failed)
	1026 - test/subplot/subplot_1.3.sh (Failed)
	1027 - test/subplot/subplot_2.0.sh (Failed)
	1030 - test/subplot/subplot_2.3.sh (Failed)
	1031 - test/subplot/subplot_3.0.sh (Failed)
	1034 - test/subplot/subplot_3.3.sh (Failed)
	1039 - test/subplot/subplotoverlay.sh (Failed)
	1040 - test/subplot/tictactoe.sh (Failed)
```

A new test subplot/circles.sh plots two -JG circles at 12cm and as indicated above the frame pen should be 1.39p = 23 PS units.  If you run that script (copy int) with master it will draw the frames at 1.5p (25 PS units) and hence demonstrate the bug that I think these other 80 scripts share.

PaulWessel:
I am OK with waiting since the effects are rather tiny and if we decide to merge later is is just seen as  minor fix anyway.  So let's hold it and move ahead with 6.2.0.

meghanrjones:
Most of the changes in the failing tests seem fairly inconsequential, but the updates to colorbar frame thickness (e.g., `doc/scripts/GMT_agefig.sh`, `docs/scripts/GMT_App_M_1*.sh`) looks much worse after this PR in my opinion.

PaulWessel:
I will build and look.

PaulWessel:
So if we use MAP_FRAME_PEN on a small or skinny colorbar then I agree it is too thick.  But isn't this what we get in GMT classic too, going way back?  In general, colorbars are much smaller than the figures they embellish so perhaps it is just a too thick default pen?

meghanrjones:
To me it looks like this PR makes thicker borders than in 6.1:
https://docs.generic-mapping-tools.org/6.1/datasets/remote-data.html#global-earth-seafloor-crustal-age-grids
https://docs.generic-mapping-tools.org/6.1/cookbook/cpts.html

PaulWessel:
Yes, but 6.1 did not have the auto-scaling theme stuff so did we perhaps erroneously change the default pen for the frame? Could you have a look at the 6.1.1 tag source for psscale and see if we set another pen or changed it in any way?

meghanrjones:
`doc/scripts/GMT_agefig.sh` includes `gmt set MAP_TICK_LENGTH 2p MAP_ANNOT_OFFSET 2p MAP_FRAME_PEN 0.5p`, but here the colorbar frame is 1.5p. Same case with `doc/scripts_GMT_APP_M_1*.sh` (`gmt set MAP_FRAME_PEN thinner`). So, gmtset is not being honored properly in this branch.

PaulWessel:
OK, I will strip down GMT_agefig.sh to its bare bits while retaining the bug and see why.

PaulWessel:
Stepping through the commands in the script I see that when gmt colorbar is run the gmt.conf file has the correct 

`MAP_FRAME_PEN                  = thinner,black`

setting.  Yet, colorbar draws an outline with 25 units * 0.06 = 1.5 points... Will debug now.

PaulWessel:
As **colorbar** starts, defaults are set to NaN as per auto-scaling in modern mode sessions.  So although the thinner pen was set in gmt.conf it gets wiped and replaced with NaN.  Then **colorbar** calls _gmt_auto_offsets_for_colorbar_ which runs 

`gmt_set_undefined_defaults (GMT, 0.0, false);	/* Must set undefined to their reference values for now */`

and that is why it is 1.5p again.  I suppose the solution is to **not** set things to NaN if the value was set via gmt.conf, right?

Need to run (drive) to the bank for notary public service so 2 hours will blow away soon.

meghanrjones:
> As **colorbar** starts, defaults are set to NaN as per auto-scaling in modern mode sessions. So although the thinner pen was set in gmt.conf it gets wiped and replaced with NaN. Then **colorbar** calls _gmt_auto_offsets_for_colorbar_ which runs
> 
> `gmt_set_undefined_defaults (GMT, 0.0, false); /* Must set undefined to their reference values for now */`
> 
> and that is why it is 1.5p again. I suppose the solution is to **not** set things to NaN if the value was set via gmt.conf, right?
> 
> Need to run (drive) to the bank for notary public service so 2 hours will blow away soon.

Sounds correct.

PaulWessel:
Found a bug that affected at least some of these cases.  When _gmt_getpen_ got `thinner,black` it failed to turn the bool _set_NaN_ back to false since we got a valid width.  It only did that if we passed a width only.  The _GMT_agefig_ looks good to me now - please have a try.

meghanrjones:
Only three failing tests now, all of which may relate to this error which appears new:
```
5: basemap [ERROR]: Pen attributes not using just - and . for dashes and dots. Offending character --> :
5: basemap [ERROR]: Pen attributes not using just - and . for dashes and dots. Offending character --> :
5: basemap [ERROR]: Pen attributes not using just - and . for dashes and dots. Offending character --> :
```

PaulWessel:
After the gmt set commands the gmt.conf has this

`MAP_GRID_PEN_PRIMARY           = black,.:0p`

This comes from thinnest,. being set to a NaN penwidth by the theme and then it gets written out.  It probably should not write the :0p though which I think trigger parsing related to actual lengths (not known yet).  I will fix this.

PaulWessel:
Looks like parsing :phase was not implement when we use dots and dashes to define the line, only when numbers are used.  E.g., 3_4_1_5:1p is OK but ..-..:2p is not.  Seems like it should be since width is defined here.

PaulWessel:
Progress.  Fixed that part and tested it.  However, found that this now makes circles.sh fail as the pen is back to 25 units.  The problem is this section in subplot.c where we need to define the undefined values for all panels up front:

```
		if (Ctrl->F.mode == SUBPLOT_FIGURE)	/* Got figure dimensions */
			gmean_dim = sqrt (Ctrl->F.dim[GMT_X] * Ctrl->F.dim[GMT_Y]);
		else {	/* Got panel dimension(s), compute total figure dimensions */
			if (Ctrl->F.reset_h) {	/* Update h based on map aspect ratio and width of a constant column */
				for (row = 0; row < Ctrl->N.dim[GMT_Y]; row++) Ctrl->F.h[row] = Ctrl->F.w[0] * (GMT->current.map.height / GMT->current.map.width);
			}
			/* Sum up individual widths or heights and add the fluff space */
			for (col = 0; col < Ctrl->N.dim[GMT_X]; col++) width  += Ctrl->F.w[col];
			for (row = 0; row < Ctrl->N.dim[GMT_Y]; row++) height += Ctrl->F.h[row];
			gmean_dim = sqrt (width * height);
			width = height = 0.0;	/* Reset */
		}
```

As you can see, I am using the entire subplot's area to compute that representative _gmean_dim_.  However, this does not make much sense to me: If I am plotting a 4x4 subplot with lots of small panels then I do not want fat pens and large fonts based on the size of the entire thing but I want what is sensible for those smaller panels.  So I think the above needs to be adjusted this way

`width /= n_columns;  height /= nrows;`

before computing the gmean.  This will give the exact size if **-Fs** is set for a constant panel size and a reasonable compromise in case we have variable panel sizes.  It will be more representative of the typical panel size and hence a good constant scale to use for all the panels.

Trying this now.

PaulWessel:
Hi @meghanrjones, please run the test after updating this branch.  This changes the scale for all subplot based on the average panel size instead of the entire subplot frame size.  I think this makes more sense, but means ~80 failures but for the better I hope.

PaulWessel:
For instance, those polarcircles.sh test cases with 4x3 = 12 panels look nicer I think when the font size is determined based on the size of each panel and not the entire page!

PaulWessel:
FYI, I just went through and looked at all the failing plots.  There are 83 and I think almost all are subplots which are of course affected by fixing the auto-scaling to depend on the average panel size.  I think they look good and things are consistent.  Have a look early next week and see if you agree.  Leaving failures as they are for now.

meghanrjones:
Can you explain whether the 25 pt result for circles.sh was a bug or just a design consequence of scaling based on sqrt(width * height) of the entire figure? 

I prefer the font sizes based on sqrt(width * height) rather than sqrt(width/cols * height/nrows), but will defer to your preference if you want to update the PS files. I looked throught the test results and did not notice any problems.

PaulWessel:
> Can you explain whether the 25 pt result for circles.sh was a bug or just a design consequence of scaling based on sqrt(width * height) of the entire figure?
> 
Given the proposed scaling, the scale is ((2.0/15.0) * (12 - 10.0) + 9)/10 = 0.92666 which when scaled by 1.5 p and converted to PS units (1/0.06) is 23 units.  I think in master this scaling used the geometric mean of the two figures which is almost 17 cm and that gives ((2.0/15.0) * (17 - 10.0) + 9)/10 = 0.9933 and hence 1.5p it is.

> I prefer the font sizes based on sqrt(width * height) rather than sqrt(width/cols * height/nrows), but will defer to your preference if you want to update the PS files. I looked throught the test results and did not notice any problems.

Well, since you like the other scheme it is worth discussing it.  My thinking was that if you have a large number of panels then the proposed scheme will shrink down the fonts so your annotations are in proportions to the panel size, whereas the previous scheme that you prefer would only scale it to the overall subplot size, hence the fonts would be larger. I can see benefits to both of these schemes. Might this be worth an option or modifier? E.g., **-F**...[**+ap**|**s**] , where **+a** sets auto-scaling based on average **p**anel size or **s**ubplot size [subplot].

meghanrjones:
> Might this be worth an option or modifier? E.g., **-F**...[**+ap**|**s**] , where **+a** sets auto-scaling based on average **p**anel size or **s**ubplot size [subplot].

This seems good, although I recommend **-F**...[**+af**|**s**] for scaling on figure or subplot, similar to **-Ff**|**s**, since panel and subplot are often used interchangeably.

PaulWessel:
Yep, that is better, and if we default to **+af** then we can declare victory and have fewer failures.

PaulWessel:
OK, **+af**|**s** have been implemented and documented.  Down to 4 failures, including circles (but not because of the frame pen).  I think this is a good compromise and gives added flexibility.  See what you think.

meghanrjones:
Tests look good, except that **-F+af** and **-F+as** do the same thing. For example, `test/subplot/autocolorline.sh` fails when I add **+af** as a test where it should pass.

PaulWessel:
Yikes, only checked for +a, not looking at the args.  OK, now only switching if s is appended.

meghanrjones (approved):
Looks good after updating the PS files.

PR #5243, opened 2021-05-21
Use correct architecture in bundle name

PaulWessel:
Also use -p to places the products instead of -n to NOT place products.
[Note: Maybe too hard to bet build-release to work under brew since there are several macports-path-specific things going on.  I will see if macports now will install under M1 since some time has passed since last attempt].

I am also not sure if we should wait for this effort.  With the installers working, is the plan to declare victory and rebuild installers as 6.2.0?  What is a suitable wait period for this?

### DIFF
--- a/admin/build-release.sh
+++ b/admin/build-release.sh
@@ -36,26 +36,22 @@ abort_build() {	# Called when we abort this script via Crtl-C
 }
 
 TOPDIR=$(pwd)
-if [ "${USER}" = "pwessel" ] || [ "${USER}" = "meghanj" ]; then	# Set ftp default action
+do_ftp=0
+if [ "X${1}" = "X-p" ]; then
 	do_ftp=1
-else
-	do_ftp=0
-fi
-
-if [ "X${1}" = "X-n" ]; then
-	do_ftp=0
 elif [ "X${1}" = "X-m" ]; then
 	do_ftp=2
 elif [ $# -gt 0 ]; then
 	cat <<- EOF  >&2
-	Usage: build-release.sh [-n]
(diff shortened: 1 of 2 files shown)


meghanrjones:
> I am also not sure if we should wait for this effort. With the installers working, is the plan to declare victory and rebuild installers as 6.2.0? What is a suitable wait period for this?

Not sure I understand this question. Are you suggesting not releasing rc2?

PaulWessel:
No, in my mind we had already released rc2 but of course we have not.  My mind is sometimes ahead but often behind actual events.
So we move forward releasing rc2!

PaulWessel:
Should we leave this as a brach for now and WIP or is this something that can be merged in now?

meghanrjones:
So, for me this line:
```
set (CPACK_PACKAGE_FILE_NAME "gmt-${CPACK_PACKAGE_VERSION}-darwin-${CMAKE_HOST_SYSTEM_PROCESSOR}")
```
will change the package name from `x86_64` to `i386` (I think). Since the bundle is already made and placed, it should not matter for this release. But we will need to remember to update the workflows for the next releases.

meghanrjones (approved):

PaulWessel:
Hm, OK, so uname -m gives i386 for you on the iMac Pro? I get x86_64 on macnut (and arm64 on M1)

meghanrjones:
> Hm, OK, so uname -m gives i386 for you on the iMac Pro? I get x86_64 on macnut (and arm64 on M1)

You are correct. I was thinking of uname -p.

PR #5247, opened 2021-05-22
Implement the documented -N and update the docs

### LINKED ISSUE #5246: Is grdfill -N option available?
The [grdfill documentation](https://docs.generic-mapping-tools.org/dev/grdfill.html#optional-arguments) lists the **-N***nodata* option, but I don't see it in the synopsis or in the source code.

Please verify if **grdfill** has the **-N** option or not.

PaulWessel:
We advertised **-N** but had not code it up; this PR does that.  Also, we stated that **-As** had not been implemented but it has been for a long time. I have now improved the man page to mention all of this.  Closes #5246 but will not be merged until after 6.2.0 has been released, hence WIP.

### DIFF
--- a/doc/rst/source/grdfill.rst
+++ b/doc/rst/source/grdfill.rst
@@ -15,8 +15,9 @@ Synopsis
 **gmt grdfill** *ingrid*
 |-A|\ *mode*\ [*arg*]
 |-G|\ *outgrid*
-[ |SYN_OPT-R| ]
 [ |-L|\ [**p**] ]
+[ |-N|\ *value* ]
+[ |SYN_OPT-R| ]
 [ |SYN_OPT-V| ]
 [ |SYN_OPT-f| ]
 [ |SYN_OPT--| ]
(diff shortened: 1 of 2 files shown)


meghanrjones:
The docs state that **-A** is required but grdfill without **-A** passes without warning (see below). Is -A required? If not, what is the default?
```
gmt grdclip @earth_relief_05m -R199:30/206/18/23 -Sa0/NaN -Gislands.nc
gmt grdfill islands.nc -Gnew.nc
```

PaulWessel:
I think **-A** should be required.  It is not obvious that anyone would expect a default method in this module.  So I think we should check for it and complain if not given.  Again, the docs are clear while the grdfill synopsis prints it as optional.  I will fix this herein.

PaulWessel:
I will do it in a separate PR.

PaulWessel:
Hm. So **either** -A or -L are required.  If you use -L then -A is not required.  I think when that happens in other modules we list both -A and -L as optional but state that one must be selected, right?

meghanrjones:
> Hm. So **either** -A or -L are required. If you use -L then -A is not required. I think when that happens in other modules we list both -A and -L as optional but state that one must be selected, right?

Yes and usually break with error if neither are provided.

PaulWessel:
Should be OK here now, @meghanrjones.

meghanrjones (approved):

PR #5248, opened 2021-05-22
Check status return of build-macos-external-list.sh properly

PaulWessel:
Attempt number 1 was stupid and failed.


meghanrjones (approved):

PaulWessel:
Had to add the OMP library manually to ensure it gets into the bundle

PR #5252, opened 2021-05-24
Clarify grdfill required options

PaulWessel:

Note that **-A** is not requires but one of **-A -L** is, and hence **-G** is optional too.


meghanrjones (approved):
Looks good for after rc2.

meghanrjones (approved):

PR #5254, opened 2021-05-24
Include arm64 bundle in draft release workflow

meghanrjones:

This PR adds the arm64 bundle to the draft release workflow.

Fixes #

**Reminders**


PaulWessel (approved):
I approve, but perhaps you should add (intel) or (x86_64) for the old bundle to match what you do with the arm bundle?

Issue #5260, opened 2021-05-25
Add plot -Ar|t to draw lines following theta or r direction in polar projections


seisman:
**Description of the desired feature**

Originally posted in the [forum](https://forum.generic-mapping-tools.org/t/1719).

For polar projections, there is no simple way to draw lines along the theta or r direction. It would be good to add **-Ar|t** to draw lines (1) **-Ar**: following the r direction then theta direction; (2) **-At**: following the theta direction then r direction.

**A minimal example:**
```
gmt begin map
gmt basemap -R0/60/0/10 -JP10c -Baf
gmt plot -W2p,red -At << EOF
0  5
60 5
EOF
gmt end show
```

**Expected output:**
![image](https://user-images.githubusercontent.com/3974108/119432639-b2016a00-bce2-11eb-8bfd-364b4afea2c9.png)

PaulWessel:
Closed via #5263.

PR #5263, opened 2021-05-26
Add polar resampling via -Ar|t in plot, plot3d, and clip

PaulWessel:

See #5260 for background.  This PR implements the feature request in modules **plot**, **plot3d**, and **clip**.  As an example of what this looks like I added a new test (_polarcurves.sh_) which produces this plot with **-Ar**|**t** vs no **-A**:

polarcurves

I set the theta resampling interval to 1 degree for now - this may need to be revisited.  To be merged after 6.2.0 release.


PaulWessel:
Did you get a chance to check this out @seisman ?

seisman (approved):
Looks great.

PR #5264, opened 2021-05-26
Let grdclip handle polar resampling via -Ar|t

PaulWessel:
Inspired by #5260, do the same when creating mask grids from polygons if we wish to resample the input polygon along straight theta and r path segments.  Merge after 6.2.0, and will have no effect until #5263 have been merged as well.


PaulWessel:
Sorry, I have no test for this but it is the same under-the-hood machinery as used in plot and plot3d, so it is expected to work as is, @seisman.

seisman (approved):

PR #5265, opened 2021-05-26
Use non-scaled pen default for +p

### LINKED ISSUE #4955: Potential issues or breaking changes with the GMT modern theme?

meghanrjones:

This PR changes the default pen for +p to be `thicker,black` rather than scaled based on the size. 

Several tests fail, but for the better. I can push new postscript files after review of the fixed default.

Fixes #4955 and relates to https://github.com/GenericMappingTools/pygmt/pull/1287.

Fixes #

**Reminders**

### DIFF
--- a/doc/examples/ex22/ex22.ps
+++ b/doc/examples/ex22/ex22.ps
@@ -1,11 +1,11 @@
 %!PS-Adobe-3.0
 %%BoundingBox: 0 0 595 842
 %%HiResBoundingBox: 0 0 595.0000 842.0000
-%%Title: GMT v6.2.0_6b09bc8_2021.03.13 [64-bit] Document from coast
+%%Title: GMT v6.2.0_eb9390f_2021.05.26 [64-bit] Document from coast
 %%Creator: GMT6
 %%For: unknown
 %%DocumentNeededResources: font Helvetica Helvetica-Bold
-%%CreationDate: Sat Mar 13 18:31:02 2021
+%%CreationDate: Wed May 26 14:11:47 2021
 %%LanguageLevel: 2
 %%DocumentData: Clean7Bit
 %%Orientation: Landscape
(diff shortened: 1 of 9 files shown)


PaulWessel (approved):
Maybe we should add (void) in front of _gmt_getpen_ since we are discarding the return value sinc we know there will be no error as the arg is a constant.  One less compiler warning probably.

meghanrjones:
> Maybe we should add (void) in front of _gmt_getpen_ since we are discarding the return value sinc we know there will be no error as the arg is a constant. One less compiler warning probably.

OK, done in https://github.com/GenericMappingTools/gmt/pull/5265/commits/eb9390f39b91ba8789d1f258f6afa00974a05789. 

OK to merge after updating the PS files? I think this should be considered a 6.2.0rc2 bug that needs to be fixed for 6.2.0.

PaulWessel:
So I am unclear on how this all work.  We release rc2 waiting for feedback.  GMT bugs probably follows a Poisson distribution and they come in at odd intervals.  if we wait long enough bugs will be reported and we then should fix them and do rc3?  It is an never-ending cycle then.  I have fixed other bugs and labelled them WIP.   I mean. I am OK with merging those bug fixes and get 6.2.0, but if we do an rc and wait a few days then we are back at it again.  Thoughts?

meghanrjones:
> So I am unclear on how this all work. We release rc2 waiting for feedback. GMT bugs probably follows a Poisson distribution and they come in at odd intervals. if we wait long enough bugs will be reported and we then should fix them and do rc3? It is an never-ending cycle then. I have fixed other bugs and labelled them WIP. I mean. I am OK with merging those bug fixes and get 6.2.0, but if we do an rc and wait a few days then we are back at it again. Thoughts?

I would prefer to release 6.2.0 next rather than do another RC. I think it's reasonable to merge bug fixes between the final RC and full release, but I could be missing something about the process since this is the first release cycle that I have been involved in.

PaulWessel:
What say you, @joa-quim and @seisman ? Do we merge the bug fixes or do we stop and just release 6.2.0, and after what waiting period?

joa-quim:
I have raised a couple times before the question on why we were releasing an RC2 instead of 6.2.0 but got no attention.
RCs trail around during about a week and if no regressions or serious bugs are found they become the final release. Leaving them at open during one month leads to the current situation where normal bug fixes start to pile up. I think it's silly to leaving them out of the 6.2 release which should be released almost ASAP.

PaulWessel:
Given that once we release 6.2.0 there won't be a quick release right away, so leaving known bug fixes in the github as WIP does not seem like a good idea to me.  I think we should merge in bug fixes now.

PR #5267, opened 2021-05-27
Let gmt_parse_array look out for memory files

### LINKED ISSUE #5266: saple1d -T crash in Julia
This crashes in ``GMT_Create_Options`` because, I don't know why, the vector in **T** below, which is = 1:332, is converted to text and the **p** variable is declared as "1024", which is not big enough. However, increasing its size only transfer the crash further down when trying to access to a non-initializes *opt*. The main question, I think, is why the -T vector argument is converted to text?
```
t = gmtread("t.dat")
sample1d(gmtread("stack.dat"), T=t[1].data)
```
But this works
```
sample1d(gmtread("stack.dat"), T="1/332/1")
```

[t.zip](https://github.com/GenericMappingTools/gmt/files/6550279/t.zip)

PaulWessel:
In the externals, the argument to **sample1d -T** may well be an in-memory array, but _gmt_parse_array_ had no check for that, just for regular files.  The next step is _gmt_create_array_ which uses _GMT_Read_Data_ so I have good hope this will fix #5266.

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -17023,13 +17023,19 @@ unsigned int gmt_parse_array (struct GMT_CTRL *GMT, char option, char *argument,
 		return (GMT_NOERROR);
 	}
 
-	/* 1b. Check if argument is a local file */
+	/* 1b. Check if argument is a memory file */
+	if (gmt_M_file_is_memory (argument)) {	/* Yep */
+		T->file = strdup (argument);
+		return (GMT_NOERROR);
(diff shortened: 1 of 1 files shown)


PaulWessel:
Perhaps @joa-quim can approve this one then?

PaulWessel:
Suspect @joa-quim is eating again and hence unavailable to approve?

joa-quim (approved):

Issue #5268, opened 2021-05-27
Improve the readability of the animation examples?


seisman:
**Description of the desired feature**

I will take the [anim01.sh](https://github.com/GenericMappingTools/gmt/blob/master/doc/examples/anim01/anim01.sh) as an example. The code to generate the main script is:

```
# 2. Set up the main frame script
cat << EOF > main.sh
gmt begin
#	Plot smooth blue curve and dark red dots at all angle steps so far
	last=\$(gmt math -Q \${MOVIE_FRAME} 10 MUL =)
	gmt convert sin_curve.txt -Z0:\${last} | gmt plot -W1p,blue -R0/360/-1.2/1.6 -JX3.5i/1.65i -X0.35i -Y0.25i
	gmt convert sin_point.txt -Z0:\${MOVIE_FRAME} | gmt plot -Sc0.1i -Gdarkred
#	Plot bright red dot at current angle and annotate
	gmt plot -Sc0.1i -Gred <<< "\${MOVIE_COL0} \${MOVIE_COL1}"
	printf "0 1.6 a = %3.3d" \${MOVIE_COL0} | gmt text -F+f14p,Helvetica-Bold+jTL -N -Dj0.1i/0.05i
gmt end
EOF
```
In the script, we use `\$` a lot to avoid expanding these variables when generating the main script. 

Usually, I feel the "anmi01.sh" difficult to read and sometimes confusing due to the use of `\$`. It's possible to suppress variable expansion by simply using `'EOF'` rather than `EOF` (see https://stackoverflow.com/questions/31645341/how-to-suppress-variable-substitution-in-bash-heredocs for the original post).

Here is the new code using `'EOF'`:
```
# 2. Set up the main frame script
cat << 'EOF' > main.sh
gmt begin
#	Plot smooth blue curve and dark red dots at all angle steps so far
	last=$(gmt math -Q ${MOVIE_FRAME} 10 MUL =)
	gmt convert sin_curve.txt -Z0:${last} | gmt plot -W1p,blue -R0/360/-1.2/1.6 -JX3.5i/1.65i -X0.35i -Y0.25i
	gmt convert sin_point.txt -Z0:${MOVIE_FRAME} | gmt plot -Sc0.1i -Gdarkred
#	Plot bright red dot at current angle and annotate
	gmt plot -Sc0.1i -Gred <<< "${MOVIE_COL0} ${MOVIE_COL1}"
[...]

PR #5271, opened 2021-05-27
Simplify animation scripts by using 'EOF'

PaulWessel:
See #5268 for background.  This PR implements the suggestion.

### DIFF
--- a/doc/examples/anim06/anim06.sh
+++ b/doc/examples/anim06/anim06.sh
@@ -14,7 +14,7 @@
 rate=24			# Frames per seconds
 frames=$(gmt math -Q 60 $rate MUL =)
 # 0. Initial parameters
-cat << 'EOF' > init.sh
+cat << EOF > init.sh
 R=-R-7.5/2.5/-1.5/2	# Fixed plot domain window
 J=-JX22c/11.5c		# Frame size after removing margin space
 f=2			# Frequency of chirp in Hz at end time
--- a/doc/examples/anim06/anim06.sh
+++ b/doc/examples/anim06/anim06.sh
@@ -14,7 +14,7 @@
 rate=24			# Frames per seconds
 frames=$(gmt math -Q 60 $rate MUL =)
 # 0. Initial parameters
-cat << EOF > init.sh
+cat << 'EOF' > init.sh
 R=-R-7.5/2.5/-1.5/2	# Fixed plot domain window
 J=-JX22c/11.5c		# Frame size after removing margin space
 f=2			# Frequency of chirp in Hz at end time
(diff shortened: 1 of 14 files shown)


doc/examples/anim06/anim06.sh
```diff
@@ -14,15 +14,15 @@
 rate=24			# Frames per seconds
 frames=$(gmt math -Q 60 $rate MUL =)
 # 0. Initial parameters
-cat << EOF > init.sh
+cat << 'EOF' > init.sh
```
seisman, line 17: We probably should not change this one, because we want to expand the variables `$rate` and `$frames`.

doc/examples/anim12/anim12.sh
```diff
@@ -21,7 +21,7 @@ FILT=21
 # INC=5.625
 # FILT=10.5
 # 1. Create preflight script to build data files needed in the loop
-cat << EOF > pre.sh
+cat << 'EOF' > pre.sh
 gmt begin
```
seisman, line 25: This one is also problematic, because the `pre.sh` script need the expanded `$SIZE`.

doc/examples/anim14/anim14.sh
```diff
@@ -50,7 +50,7 @@ gmt math -T$Angles T SIND -o0,1,0 = | gmt math STDIN -T -C0 COSD = | gawk '$0=$0
 Ha=$(gmt math -Q $H 0.7c SUB =) # Height of inset with angles
 Wa=$(gmt math -Q $Ha $VE DIV =) # Width of inset with angles (calculated from vertical exaggeration to have same deformation as the profile)
 #       -----------------------------------------------------------------------------------------------------------
-cat << EOF > pre.sh
+cat << 'EOF' > pre.sh
```
seisman, line 53: The same problem here.

doc/examples/anim06/anim06.sh
PaulWessel: OK, done

doc/examples/anim14/anim14.sh
PaulWessel: Yep, removed quote.

doc/examples/anim14/anim14.sh
```diff
@@ -79,12 +79,12 @@ gmt begin
 gmt end
 EOF
 # 2. Set up main script
-cat << EOF > main.sh
+cat << 'EOF' > main.sh
```
seisman, line 82: The main script in anim14 uses both movie-specific variables (e.g., `${MOVIE_COL0}`) and general bash variables (e.g., `$W`). 

I think you have to create a `inc.sh` file to make it work.

seisman (approved):
The current changes look good to me.

PaulWessel:
Would you want to wait for @Esteban82's input on anim14 before merging this one?

seisman:
anim14 is more complicated. It's better to work on it in a separate PR.

PaulWessel:
OK. @federico, we can include any changes you make later.

Esteban82:
> OK. @federico, we can include any changes you make later.

Better. Right now I don't have much time.

PR #5285, opened 2021-05-31
Duplicate into grid with full pad instead of allocating twice

PaulWessel:
When an incoming grid has no pad and **grdimage** requires it we duplicate the incoming grid, then expand it to add the pad.  However, this was done in a wasteful way where we first did an identical duplicate, then allocated the final padded grid and copied over, thus having an extra grid in memory temporarily.  Turns out _GMT_Duplicate_Data_ has a mode for this called **GMT_DUPLICATE_RESET** that does this in one step without adding the temporary memory.  Need @joa-quim and @seisman to see if this branch has any side-effects in GMT.jl and pyGMT since I cannot test that.  Just doing some tests passing in a memory matrix or grid with no pad should do it.


PaulWessel:
Hi @joa-quim and @seisman. Just letting you now that this branch, which avoids some memory duplication, works fine for me, with all tests working, including the one involving the new testapi_matrix_pad.c program to make sure the faster way gives the same result as the slower duplication.  Please review and approve after running tests in Julia and Python just in case.

seisman:
All PyGMT tests pass with this branch.

PaulWessel:
And @joa-quim said he only sends non-padded when padding won't be required by grdimage. How hard would it be to send in a non-added matrix when padding is required, @joa-quim, to make sure this branch is solid for both PyGMT and GMT.jl?

joa-quim:
I guess that's because I'm not sending matrices.
Tests pass with this branch.

```
julia> imshow(I, proj=:guess, Vd=1)
        grdimage  -R-156.000416666667/-155.499583333333/18.9995833333333/19.5004166666667 -JM14c -Baf -BWSen -n+a -D -P -K > C:\TMP\GMTjl_tmp.psgrdimage [ERROR]: gmt_img_project: Input image does not have sufficient (2) padding
```

joa-quim (approved):

PR #5291, opened 2021-06-02
Ensure external vectors are scaled via -i

PaulWessel:
See https://github.com/GenericMappingTools/pygmt/issues/1313 for details.  This PR ensures the record-by-record reading of external vectors have any adjustments set via **-i** applied.  I will add a similar fix for the table-reading later today.


PaulWessel:
I have now added the complement to this for the full dataset ingestion via vectors and the above Mississippi plot works for me.  Please look for any other oddities.

seisman:
Looks good to me. I don't see any side-effects on the PyGMT side.

seisman (approved):

PR #5297, opened 2021-06-04
Add arm64 bundle to INSTALL.md

PaulWessel:
There are now two macos bundles, and this PR updates the language.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -14,7 +14,7 @@ for compiling GMT source package (either stable release or development version).
 
 - [Windows](#windows)
 - [macOS](#macos)
-  * [Application Bundle](#application-bundle)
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -14,7 +14,7 @@ for compiling GMT source package (either stable release or development version).
 
 - [Windows](#windows)
 - [macOS](#macos)
-  * [Application Bundle](#application-bundle)
+  * [Application Bundles](#application-bundle)
```
meghanrjones, line 17:
```suggestion
  * [Application Bundles](#application-bundles)
```

meghanrjones:
Will approve after the current workflow finishes, so we do not need to wait another few hours before checking that box.

meghanrjones (approved):

PR #5302, opened 2021-06-05
Clarify that mapproject -L is spherical only

### LINKED ISSUE #5295: Something's wrong with distances calculation in mapproject -L
If I generate a geodesic circle with 100 km radius with Proj and compute the distance with ``mpapproject -L`` I get variable distances like this and all considerable different from 100 km.
(We also get an WARNING that should be there)
```
julia> Ddist = mapproject(c, L=(line=[0 0], unit=:e), j=:e, par=(PROJ_ELLIPSOID="WGS-84",), Vd=1)
        mapproject  --PROJ_ELLIPSOID=WGS-84 -je -L+ue
mapproject [WARNING]: Your distance mode (Great Circle) differs from your -j option (Geodesic) which takes precedence.
Vector{GMT.GMTdataset} with 1 segments
First segment DATA
36×5 Matrix{Float64}:
  0.0        0.904369   99330.7  0.0  0.0
  0.160413   0.889834   99352.1  0.0  0.0
  0.315668   0.846697   99413.6  0.0  0.0
  0.460773   0.776346   99507.2  0.0  0.0
  0.591064   0.681042   99621.0  0.0  0.0
  0.702354   0.563851   99740.3  0.0  0.0
  0.791067   0.428539   99850.1  0.0  0.0
  0.854355   0.279455   99936.3  0.0  0.0
  0.890187   0.121392   99988.0  0.0  0.0
  0.897411  -0.0405727  99998.7  0.0  0.0
  0.875796  -0.201233   99967.0  0.0  0.0
  0.826036  -0.355427   99896.9  0.0  0.0
```
taking just the first point in my Matlab Vincenty function I get it very close and most of the difference is probably due to truncation in input
```
>> vdist(0.904369,0,0,0)

ans =

          100000.030638943
``` 

If I use ``project -Z`` I also cannot recover the input circle radius with any **-j** option. Here the case is more fuzzy because ``project`` computes only spherical circles and it's not obvious that **-jg** is doing the same sphere, though I would expect so. 
```
project  -o0,1 -C0/0 -G10 -Z200 > circ.dat
mapproject -Lpt.dat+ue circ.dat
5.50104544457e-17       0.89831528412   98776.2018419   0       0
-0.0447809811942        0.897198516113  98778.9802858   0       0
-0.0894505753527        0.893850990596  98787.287538    0       0
...
```

PaulWessel:
See discussion for #5295.  This PR prevents -je from being set with -L.  Closes #5295.  Approve and merge after 6.2 branch is made.

### DIFF
--- a/doc/rst/source/mapproject.rst
+++ b/doc/rst/source/mapproject.rst
@@ -182,6 +182,8 @@ Optional Arguments
     report the line segment id and the fractional point number instead
     of lon/lat of the nearest point.
     See `Output Order`_ for how **-L** affects the output record.
+    **Note**: Calculation mode for geographic data is spherical, hence **-je**
+    cannot be used in combination with **-L**.
 
 .. _-N:
 
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -236,6 +236,7 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Message (API, GMT_TIME_NONE, "\t   Unit C means Cartesian distances after first projecting the input coordinates (-R, -J).\n");
 	GMT_Message (API, GMT_TIME_NONE, "\t   Three columns are added on output: min dist and lon, lat of the closest point on the line.\n");
 	GMT_Message (API, GMT_TIME_NONE, "\t   Append +p to get line segment id and fractional point number instead of lon/lat.\n");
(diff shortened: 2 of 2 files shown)


joa-quim (commented):
I wonder if we should not raise it to the error category. If user is asking for -je it should be pretty clear that it's not possible and not risk it pass unnoticed in some automatic process.

joa-quim (approved):

PR #5313, opened 2021-06-07
Add -R to coast example

meghanrjones:

The coast example in the docs did not work as written. This PR adds -R to the example.

Fixes #

**Reminders**

### DIFF
--- a/doc/rst/source/coast.rst
+++ b/doc/rst/source/coast.rst
@@ -73,7 +73,7 @@ To plot Great Britain, Italy, and France in blue with a red outline and
 Spain, Portugal and Greece in yellow (no outline), and pick up the plot
 domain from the extents of these countries, use::
 
-    gmt coast -JM6i -Baf -EGB,IT,FR+gblue+p0.25p,red -EES,PT,GR+gyellow -pdf map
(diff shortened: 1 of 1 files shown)


seisman:
Isn't it a bug?

meghanrjones:
> Isn't it a bug?

This was not clear to me. I skimmed the docs and did not see any mention of an automatic -R determination happening here except when -M is used as well:

> *If -M is used with -E then -R or the +r modifier to -E are not required as we automatically determine the region given the selected geographic entities.*

This quote from -M mentions a +r modifier to -E, but there's no +r modifier documented for -E.

PaulWessel:
This feature evolved from just -E to being applicable in -R, so I think some man pages may have obsolete language.  That +r modifier is no longer relevant since -R sets region.
However, I think @seisman is right to consider it a bug.  We have the -R available via those codes set with -E so we should not have to split it with -R.  I am flagging it as a bug.

seisman:
Actually, the old command without **-R** works well for me:
```
gmt coast -JM6i -Baf -EGB,IT,FR+gblue+p0.25p,red -EES,PT,GR+gyellow -pdf map
```

meghanrjones:
I'm closing this PR and have opened a bug report (https://github.com/GenericMappingTools/gmt/issues/5314) instead.

joa-quim:
> EDIT: Sorry, it doesn't.

It does for me

PR #5315, opened 2021-06-08
grdblend could leak memory for pixel grids just outside range

PaulWessel:
If a tile is exactly outside the final grid region, then that check is slightly different for pixel and gridline-registered grids: For the same area the w/e/s/n are the same, but at the joint the pixel-grid has no nodes.  Hence pixel grids with the same north boundary as the south boundary in the final grid would be considered inside, memory allocated, but never touched and hence failed to free.


meghanrjones:
Could the grdmath data that you used to debug be used as a test or is it too large?

PaulWessel:
Found the bug - separate PR coming soon.

PaulWessel:
Just so it is clear, this PR has nothing to do with the grdblend problem reported on the Forum - it is just something I realized while debugging that case and wanted to keep it separate.

meghanrjones (approved):

PR #5316, opened 2021-06-08
Wrong count variable used when assigning tile weight in grdblend

PaulWessel:
See forum [post](https://forum.generic-mapping-tools.org/t/grdblend-randomly-skipping-tiles/1736/2) for background. Somewhere in the transcription from old GMT to the API a variable was renamed but not changed in one later line - probably by me.  So the tile weight for the third tile if given via a blend file, would be set to zero...
This PR fixes that problem; I also use a clearer variable name (_n_scanned_) for this part of the code.  All tests passed, meaning none of our tests explored this particular scenario, so I am adding the simple example I used to debug this (_blend_weights.sh_).


PaulWessel:
In case you are wondering, the single problem was on line 256 where the old _n_ (current tile number) was used instead of _nr_ (number if items scanned in the tile record).

meghanrjones (approved):

PR #5319, opened 2021-06-09
Use stable mergesort in -i parsing

### LINKED ISSUE #5311: gmtinfo -i from externals is screwing on Windows
Probably the source of the PyGMT [test failure](https://github.com/GenericMappingTools/pygmt/pull/1322) on Win 
```
julia> gmtinfo(rand(5,3), i="0-2,2", C=true)
Vector{GMT.GMTdataset} with 1 segments
First segment DATA
1×8 Matrix{Float64}:
 0.147981  0.882137  0.366351  0.884463  0.0  2.25401e-315  0.0  2.25401e-315
```
but it works well on Linux
```
julia> gmtinfo(rand(5,3), i="0-2,2", C=true)
Array{GMT.GMTdataset,1} with 1 segments
First segment DATA
1×8 Array{Float64,2}:
 0.0540404  0.662781  0.173727  0.691856  0.0268022  0.900573  0.0268022  0.900573
```
It works well on Win if instead of an array we pass in a file name or if we pass a simpler (and redundant) **-i0-2** option

PaulWessel:
Since _qsort_ is unstable, especially on Windows.  Hopefully works and closes #5311.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -9135,6 +9135,7 @@ int gmt_parse_i_option (struct GMT_CTRL *GMT, char *arg) {
 			}
 		}
 	}
+	/* Use mergesort since qsort is unstable (i.e., unpredictable order) when items are identical */
 	mergesort (GMT->current.io.col[GMT_IN], k, sizeof (struct GMT_COL_INFO), gmtinit_compare_cols);
 	GMT->common.i.n_cols = k;
 	if (k) {	/* Because the user may have repeated some columns we also determine how many unique columns were requested */


joa-quim (approved):

PaulWessel:
So this bug and the grdblend bug yesterday make we wonder if 6.2.1 should be in the card here.

PaulWessel:
In looking at all places we call _qsort_ to see if we may want to user _mergesort_ instead I found this in **psxyz**.  Was this you and what does it mean?

```
GMT_LOCAL int psxyz_dist_compare (const void *a, const void *b) {
	if (((struct PSXYZ_DATA *)a)->dist[0] < ((struct PSXYZ_DATA *)b)->dist[0]) return (-1);
	if (((struct PSXYZ_DATA *)a)->dist[0] > ((struct PSXYZ_DATA *)b)->dist[0]) return (1);
	if (((struct PSXYZ_DATA *)a)->dist[1] < ((struct PSXYZ_DATA *)b)->dist[1]) return (-1);
	if (((struct PSXYZ_DATA *)a)->dist[1] > ((struct PSXYZ_DATA *)b)->dist[1]) return (1);
#if defined(WIN32) || defined(__MINGW32__)
	/* MSVC qsort call a quick sorting function when number of elements to sort is small. e.g.
	 * below a certain size, it is faster to use a O(n^2) sorting method *
		if (size <= CUTOFF) {
			__SHORTSORT(lo, hi, width, comp, context);
		}
		and that function damn looks bugged as it imposes

		if (__COMPARE(context, p, max) > 0) { ...

		as condition to NOT change order (instead of >= 0). Se we force the hand here and
		return 1 to circumvent that bad behavior

		Note: Since it uses MSVCRT.DLL MinGW GCC suffers from the same bug
	*/
	return (1);
#else
	return (0);
#endif
}
```

PaulWessel:
BTW, I looked at all qsort instances and I think all the others are sorting floating point numbers that are very unlikely to be equal, and when they are it does not matter the order.

joa-quim:
Well, reading the description it seems a patch for the exact same bug (which they refuse to admit as such).
I remember that I had a lot of suffering to find out why something was failing and did a lot of debugging. Even in the MSVC code that the debugger could access.

PaulWessel:
Perhaps too much to add this if-def to all compare functions.  I think parse_i_ is certainly one where we have small arrays and equalities, so perfect storm for your qsort.  Also mergesort is perfect solution for that.  So perhaps we should just declare victory and move on to other things.

joa-quim:
Yes. Let's leave as is. If it comes back again we'll be more to remember this.

Issue #5182, opened 2021-04-29
Option to set MAP_LABEL_OFFSET to be relative to axis


evangowan:
**Description of the desired feature**

Right now, the MAP_LABEL_OFFSET option for placing the axis label text is set to be relative to the side of the annotations. For stacked plots (e.g. a stack of time series), this can lead to the problem of the labels not being aligned on all of the plots if the annotations are of difference sizes. Ideally, an option to set the MAP_LABEL_OFFSET to be relative to the axis rather than the annotation would make it easier to line things up.

To illustrate what I mean, I have attached this image. As you can see, the label “Sliding Mechanism” is much further to the left of the other labels. 

![labels](https://user-images.githubusercontent.com/8361371/116530953-e863e900-a8de-11eb-9307-3948c52067f0.png)
 
There is some further discussion on the [GMT forum](https://forum.generic-mapping-tools.org/t/axis-label-offset/1616)

As mentioned in the post, the workaround for now is to use `gmt text -N`.  

**Are you willing to help implement and maintain this feature?** No

PaulWessel:
Might it be possible for you to post the script that made the example above?  It would make a good test when I implement the feature.

evangowan:
Dear Paul,

I apologize for the delay. I have created a test script that you can use.

`#! /bin/bash



font_label="--FONT_LABEL=10p --MAP_LABEL_OFFSET=6p"
font_anno="--FONT_ANNOT_PRIMARY=8p"

loess_line="2p"
dot_size="0.1"

plot="test.ps"


plot_width=8c
plot_height=2c

xmin=6
xmax=17

############## C4

parameter="C4"
ymin=0
ymax=24.9
ytick=10
ysubtick=5
axis_label="C@-4@- (%)"

J_options="-JX${plot_width}/${plot_height}"
R_options="-R${xmin}/${xmax}/${ymin}/${ymax}"



gmt psbasemap  -BWSen -Bxa2f1+l"Age (Ma)" -Bya${ytick}f${ysubtick}+l"${axis_label}"  ${J_options} ${R_options} -P  -K ${font_label} ${font_anno} > ${plot}




############## MC total

parameter=MC_total

ymin=80
ymax=12000
axis_label="MC@-T@-"


J_options="-JX${plot_width}/${plot_height}l"
R_options="-R${xmin}/${xmax}/${ymin}/${ymax}"



[...]

PR #5320, opened 2021-06-10
Allow for a fixed label offset

PaulWessel:

See issue #5182 for background.  This PR lets a negative **MAP_LABEL_OFFSET** be interpreted as the absolute distance from the axis to the base of the label [The default is that a positive offset is seen as a relative addition beyond the end of the annotations].  The use for this is to better align multiple y-axes plots that are stacked so that the labels align even if the annotations have different magnitudes and formats.  TO demonstrate this effect I added a test based on the original poster's issue.  This plot is found below this text.  One remaining issue is whether of not we restrict this mechanism to just the y-axis.  There are two different consequences depending on the answer:

- If applicable to all axis then the user may need to use separate **basemap** calls so that the effect only applies to the label they want.
- If only applicable to the y-axis then one cannot align x-labels on stacks of plots in several columns, and not just this example of one column.

For these reasons I favor not imposing any restriction on which axis it applies to.  Also note: The offset is the exact absolute distance between axis and base of label . That means it is the same on both left and right axis, but since the base is on the right there is a difference.  Another question is then if we need to ensure it works symmetrically on left and right, meaning on the right axis (and bottom) we must deal with the absolute distance to the _top_ of the label.

Comments please. 

fixed_labeldist

### DIFF
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -788,7 +788,10 @@ MAP Parameters
     **MAP_LABEL_OFFSET**
         Distance from base of axis annotations to the top of the axis label
         [default is :doc:`theme dependent <theme-settings>`]. Choose **auto**
-        for :ref:`automatic scaling with plot size <auto-scaling>`.
+        for :ref:`automatic scaling with plot size <auto-scaling>`. **Note**:
+        If a negative offset is given then it is interpreted as the absolute
+        offset from the axis to the base of the label; this is useful when
+        wishing to align several y-axis labels in stacked plots.
 
     **MAP_LINE_STEP**
         Determines the maximum length (> 0) of individual straight
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -795,10 +795,10 @@ MAP Parameters
     **MAP_LABEL_OFFSET**
         Distance from base of axis annotations to the top of the axis label
         [default is :doc:`theme dependent <theme-settings>`]. Choose **auto**
-        for :ref:`automatic scaling with plot size <auto-scaling>`. **Note**:
-        If a negative offset is given then it is interpreted as the absolute
-        offset from the axis to the base of the label; this is useful when
-        wishing to align several y-axis labels in stacked plots.
+        for :ref:`automatic scaling with plot size <auto-scaling>`. To
+        set different offsets for the *x* and *y* axes, separate distances
+        by a slash (e.g., **8p/12p** for 8p offset for the x-axis and 12p
+        offset for the y-axis)
 
     **MAP_LINE_STEP**
         Determines the maximum length (> 0) of individual straight
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -789,8 +789,8 @@ MAP Parameters
         Determines from where the label offset is measured: Choose **annot**
         to mean the distance from the end of the annotation or **axis**
         to mean the distance from the axis.  To set separate modes for the
-        *x* and *y* axes, separate modes by a slash [annot/annot]. Choose **axis**
-        if you need to align multiple axes labels across many rows or columns.
+        *x* and *y* axes, separate modes by a slash [default is **annot/annot**]. Choose
+        **axis** if you need to align multiple axes labels across many rows or columns.
 
     **MAP_LABEL_OFFSET**
         Distance from base of axis annotations to the top of the axis label
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -785,6 +785,13 @@ MAP Parameters
         [default is :doc:`theme dependent <theme-settings>`]. Choose **auto** for
         :ref:`automatic scaling with plot size <auto-scaling>`.
 
+    **MAP_LABEL_MODE**
+        Determines from where the label offset is measured: Choose **annot**
+        to mean the distance from the end of the annotation or **axis**
+        to mean the distance from the axis.  To set separate modes for the
+        *x* and *y* axes, separate modes by a slash [annot/annot]. Choose **axis**
+        if you need to align multiple axes labels across many rows or columns.
+
     **MAP_LABEL_OFFSET**
         Distance from base of axis annotations to the top of the axis label
         [default is :doc:`theme dependent <theme-settings>`]. Choose **auto**
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -10669,11 +10669,19 @@ unsigned int gmtlib_setparameter (struct GMT_CTRL *GMT, const char *keyword, cha
 			break;
 		case GMTCASE_MAP_LABEL_MODE:
 			if ((i = sscanf (value, "%[^/]/%s", txt_a, txt_b)) == 2) {	/* Separate settings for x and y */
-				GMT->current.setting.map_label_mode[GMT_X] = (!strcmp (txt_a, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
-				GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (txt_b, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+				if (strcmp (txt_a, "annot") && strcmp(txt_a,"axis") && strcmp(txt_b,"annot") && strcmp(txt_b,"axis"))
+					error = true;
+				else {
+					GMT->current.setting.map_label_mode[GMT_X] = (!strcmp (txt_a, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+					GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (txt_b, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+				}
 			}
-			else
-				GMT->current.setting.map_label_mode[GMT_X] = GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (value, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+			else {
+				if (strcmp (value, "annot") && strcmp(value,"axis"))
+					error = true;
+				else
+					GMT->current.setting.map_label_mode[GMT_X] = GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (value, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+			}
 			break;
 		case GMTCASE_LABEL_OFFSET:
 			GMT_COMPAT_TRANSLATE ("MAP_LABEL_OFFSET");
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -10018,12 +10018,18 @@ void gmt_set_undefined_defaults (struct GMT_CTRL *GMT, double plot_dim, bool con
 		if (conf_update) GMT_keyword_updated[GMTCASE_MAP_ANNOT_OFFSET_SECONDARY] = true;
 	}
 	if (gmt_M_is_dnan (GMT->current.setting.map_label_offset[GMT_X])) {
-		GMT->current.setting.map_label_offset[GMT_X] = 6 * pt * scale;	/* 6p */
+		if (GMT->current.setting.map_label_mode[GMT_X] == GMT_LABEL_AXIS)
+			GMT->current.setting.map_label_offset[GMT_X] = 32 * pt;
+		else
+			GMT->current.setting.map_label_offset[GMT_X] = 6 * pt * scale;	/* 6p */
 		auto_scale = true;
 		if (conf_update) GMT_keyword_updated[GMTCASE_MAP_LABEL_OFFSET] = true;
 	}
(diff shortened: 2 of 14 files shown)


seisman:
> This PR lets a negative **MAP_LABEL_OFFSET** be interpreted as the absolute distance from the axis to the base of the label

It's backward-incompatible, right?

PaulWessel:
Do you mean we already are using a negative label offset for something else?

seisman:
> Do you mean we already are using a negative label offset for something else?

People may be using a negative label offset to make labels closer to the axis.

PaulWessel:
OK, I can imagine people doing that.  The alternative implementation would then be some sort of flag to interpret the offset the absolute way.  **MAP_LABEL_OFFSET**=**offset**[_unit_][**+a**] with a modifier for absolute. Not sure if this is the first defaults setting with a modifier so perhaps that is a bad precedent.  The final alternative is a separate and new defaults setting like **MAP_LABEL_MODE**=_auto_|_absolute_ [_auto_].

And I guess it really should measure from the axis outward to the start of the label, be it to the base of the top of it, so that means working this into PostScript for exact height, etc.

meghanrjones:
> OK, I can imagine people doing that. The alternative implementation would then be some sort of flag to interpret the offset the absolute way. **MAP_LABEL_OFFSET**=**offset**[_unit_][**+a**] with a modifier for absolute. Not sure if this is the first defaults setting with a modifier so perhaps that is a bad precedent. The final alternative is a separate and new defaults setting like **MAP_LABEL_MODE**=_auto_|_absolute_ [_auto_].
> 
> And I guess it really should measure from the axis outward to the start of the label, be it to the base of the top of it, so that means working this into PostScript for exact height, etc.

Is it possible to make these both customizable for x and y axes independently? For example, --MAP_LABEL_OFFSET 12p/24p --MAP_LABEL_MODE annot/axis for 12p x-axis labels relative to the annotations and 24p y-axis labels relative to the axis.

Either way, I prefer **MAP_LABEL_MODE** to **+a**, but find annot/axis more descriptive than auto/absolute.

PaulWessel:
I have introduced the new GMT defaults **MAP_LABEL_MODE** = _annot_|_axis_ which can be set separately for the x and y axis via a slash.  Similarly, the **MAP_LABEL_OFFSET** can now be given separate offsets for x and y if desired.  With these upgrades we can offer support for having proper alignment of numerous y-axis labels without imposing the same condition on the x-axis.  But we could, if we wish to align several sets of x labels across columns.
All tests still pass.

doc/rst/source/gmt.conf.rst
```diff
@@ -785,10 +785,20 @@ MAP Parameters
...
+        if you need to align multiple axes labels across many rows or columns.
+
     **MAP_LABEL_OFFSET**
         Distance from base of axis annotations to the top of the axis label
         [default is :doc:`theme dependent <theme-settings>`]. Choose **auto**
-        for :ref:`automatic scaling with plot size <auto-scaling>`.
+        for :ref:`automatic scaling with plot size <auto-scaling>`. **Note**:
+        If a negative offset is given then it is interpreted as the absolute
+        offset from the axis to the base of the label; this is useful when
+        wishing to align several y-axis labels in stacked plots.
```
meghanrjones, line 801:
```suggestion
        for :ref:`automatic scaling with plot size <auto-scaling>`. To
        set different offsets for the *x* and *y* axes, separate distances
        by a slash (e.g., **8p/12p** for 8p offset for the x-axis and 12p
        offset for the y-axis)
```

doc/rst/source/gmt.conf.rst
```diff
@@ -785,10 +785,20 @@ MAP Parameters
         [default is :doc:`theme dependent <theme-settings>`]. Choose **auto** for
         :ref:`automatic scaling with plot size <auto-scaling>`.
 
+    **MAP_LABEL_MODE**
+        Determines from where the label offset is measured: Choose **annot**
+        to mean the distance from the end of the annotation or **axis**
+        to mean the distance from the axis.  To set separate modes for the
+        *x* and *y* axes, separate modes by a slash [annot/annot]. Choose **axis**
+        if you need to align multiple axes labels across many rows or columns.
```
meghanrjones, line 793:
```suggestion
        *x* and *y* axes, separate modes by a slash [default is **annot/annot**]. Choose
        **axis** if you need to align multiple axes labels across many rows or columns.
```

meghanrjones:
Do you think it is possible to update the auto settings for MAP_LABEL_OFFSET to work better when MAP_LABEL_MODE is set to axis?
```
gmt begin panels pdf
  gmt subplot begin 2x1 -Fs8c -M20p -A -SCb -SRl
    gmt set MAP_LABEL_MODE axis/axis MAP_LABEL_OFFSET auto/auto
    gmt subplot set
    gmt basemap -R0/8000/0/1000 -BWSne -Bya+l"Axis label"
    gmt subplot set
    gmt basemap -R0/80/0/10 -BWSne -Bya+l"Axis label"
  gmt subplot end
gmt end show
```
panels

PaulWessel:
We could, but no single value will work well I think.  But I agree 8p is not it!  Suggestions?

meghanrjones:
> We could, but no single value will work well I think. But I agree 8p is not it! Suggestions?

32p seems OK. Maybe it would be worth adding a warning as well for MAP_LABEL_OFFSET=auto with MAP_LABEL_MODE=axis, like what is done if annotations are removed due to gmt.conf settings.

PaulWessel:
Yes, that would be good.  So 8p/32p for the two types of offsets plus warning for axis actions.  Would you like to try this?
-p

meghanrjones:
> Yes, that would be good. So 8p/32p for the two types of offsets plus warning for axis actions. Would you like to try this?
> -p

Do you mean implementing the changes or testing? Fine doing either.

PaulWessel:
Sorry, meant implementing etc. Includes doc update I think as well.

meghanrjones:
> Sorry, meant implementing etc. Includes doc update I think as well.

Yes, will do.

src/gmt_init.c
```diff
@@ -10654,11 +10667,24 @@ unsigned int gmtlib_setparameter (struct GMT_CTRL *GMT, const char *keyword, cha
...
 			GMT->current.setting.map_heading_offset = gmt_M_to_inch (GMT, value);
 			break;
+		case GMTCASE_MAP_LABEL_MODE:
+			if ((i = sscanf (value, "%[^/]/%s", txt_a, txt_b)) == 2) {	/* Separate settings for x and y */
+				GMT->current.setting.map_label_mode[GMT_X] = (!strcmp (txt_a, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+				GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (txt_b, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+			}
+			else
+				GMT->current.setting.map_label_mode[GMT_X] = GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (value, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+			break;
```
meghanrjones, line 10677: It would be good to add a check here that the value is "axis" or "auto" and to report an error if not. Let me know if you want me to do this.

PaulWessel: Yes, agreed - I was being lazy.  Although you mean axis or annot, right?

meghanrjones: Yes - clearly I spend too much time thinking about auto settings. axis or annot.

src/gmt_init.c
```diff
@@ -10654,11 +10667,24 @@ unsigned int gmtlib_setparameter (struct GMT_CTRL *GMT, const char *keyword, cha
 		case GMTCASE_MAP_HEADING_OFFSET:
 			GMT->current.setting.map_heading_offset = gmt_M_to_inch (GMT, value);
 			break;
+		case GMTCASE_MAP_LABEL_MODE:
+			if ((i = sscanf (value, "%[^/]/%s", txt_a, txt_b)) == 2) {	/* Separate settings for x and y */
+				GMT->current.setting.map_label_mode[GMT_X] = (!strcmp (txt_a, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+				GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (txt_b, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
+			}
+			else
+				GMT->current.setting.map_label_mode[GMT_X] = GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (value, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
```
meghanrjones, line 10676:
```suggestion
				if (strcmp (txt_a, "annot") && strcmp(txt_a,"axis") && strcmp(txt_b,"annot") && strcmp(txt_b,"axis"))
					error = true;
				else {
					GMT->current.setting.map_label_mode[GMT_X] = (!strcmp (txt_a, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
					GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (txt_b, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
				}
			}
			else {
				if (strcmp (value, "annot") && strcmp(value,"axis"))
					error = true;
				else
					GMT->current.setting.map_label_mode[GMT_X] = GMT->current.setting.map_label_mode[GMT_Y] = (!strcmp (value, "axis") ? GMT_LABEL_AXIS : GMT_LABEL_ANNOT);
			}
```

meghanrjones: Where does the setting get changed if it is in subplot mode? This only works outside subplot mode.

PaulWessel: If it is set via gmt set prior to gmt subplot begin it should be in the gmt.conf file that applies to the whole subplot.  Is that what you mean?

meghanrjones: Nevermind, the test works now - not sure if I was building the wrong branch or what before.

PaulWessel:
Sorry, I completely forgot to access these suggestions...

meghanrjones:
OK, I seem to be having the subplot related problem again. Is something wrong in the second example here?
**Example 1** (works as expected, with MAP_LABEL_OFFSET = 32p)
```
gmt begin panels pdf
  gmt set MAP_LABEL_MODE axis
  gmt subplot begin 2x2 -Fs8c -M25p
    gmt subplot set
    gmt basemap -By+ltest -Bx+ltest -R0/8000/0/1000
    gmt subplot set
    gmt basemap -By+ltest -Bx+ltest -R0/8000/0/1000
    gmt subplot set
    gmt basemap -By+ltest -Bx+ltest -R0/80/0/10
    gmt subplot set
    gmt basemap -By+ltest -Bx+ltest -R0/80/0/10
  gmt subplot end
gmt end show
```
panels
**Example 2** (MAP_LABEL_OFFSET still uses auto scaling):
```
gmt begin panels png
  gmt subplot begin 2x2 -Fs8c -M25p
    gmt set MAP_LABEL_MODE axis
    gmt subplot set
    gmt basemap -By+ltest -Bx+ltest -R0/8000/0/1000
    gmt subplot set
    gmt basemap -By+ltest -Bx+ltest -R0/8000/0/1000
    gmt subplot set
    gmt basemap -By+ltest -Bx+ltest -R0/80/0/10
    gmt subplot set
    gmt basemap -By+ltest -Bx+ltest -R0/80/0/10
  gmt subplot end
gmt end show
```
panels

PaulWessel:
I get the same, so looks like 6p is scaled to ~the same.  It is clearly not scaling 32p to something.  Not obvious anything in the code is wrong.  May need to debug.
Oh, the issue is that you are setting the MAP_LABEL_MODE too late.  subplot has already done the calculations and made a gmt.conf for the subplot based on the 6p then the mode changes...

src/gmt_init.c
```diff
@@ -10016,8 +10017,19 @@ void gmt_set_undefined_defaults (struct GMT_CTRL *GMT, double plot_dim, bool con
 		auto_scale = true;
 		if (conf_update) GMT_keyword_updated[GMTCASE_MAP_ANNOT_OFFSET_SECONDARY] = true;
 	}
-	if (gmt_M_is_dnan (GMT->current.setting.map_label_offset)) {
-		GMT->current.setting.map_label_offset = 6 * pt * scale;	/* 6p */
+	if (gmt_M_is_dnan (GMT->current.setting.map_label_offset[GMT_X])) {
+		if (GMT->current.setting.map_label_mode[GMT_X] == GMT_LABEL_AXIS)
+			GMT->current.setting.map_label_offset[GMT_X] = 32 * pt;
```
meghanrjones, line 10022:
```suggestion
			GMT->current.setting.map_label_offset[GMT_X] = 32 * pt * scale;
```
I set this as constant intentionally, but in retrospect scaled might be better.

meghanrjones:
> I get the same, so looks like 6p is scaled to ~the same. It is clearly not scaling 32p to something. Not obvious anything in the code is wrong. May need to debug.
> Oh, the issue is that you are setting the MAP_LABEL_MODE too late. subplot has already done the calculations and made a gmt.conf for the subplot based on the 6p then the mode changes...

OK, so it should not be expected to work. Thanks for the explanation.

meghanrjones (approved):

src/gmt_init.c
PaulWessel: Do you still want this suggestion to be active?

PR #5327, opened 2021-06-11
Add information to windows install instructions

meghanrjones:

This PR adds details about a bash shell emulator to the windows installation instructions in INSTALL.md.

Fixes #

**Reminders**

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -39,6 +39,10 @@ In addition to the GMT installer, you also need to download and install
 [GraphicsMagick](http://www.graphicsmagick.org/) if you want to create
 animated GIFs.
 
+
+**NOTE:**
+It is recommended that you use a UNIX style terminal with a bash shell for GMT scripting, to leverage GMT's symbiotic relationship with UNIX. One simple option for accessing a UNIX style terminal is *Git for Windows*, which can be downloaded from [their official website](https://gitforwindows.org/).
+
 **NOTE:**
 At the installation step, you may get the warning message:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -41,12 +41,12 @@ animated GIFs.
 
 
 **NOTE:**
-Three possibilities for using [GMT on non-UNIX systems](https://docs.generic-mapping-tools.org/latest/cookbook/non-unix-platforms.html)
-such as Windows are to install GMT under MinGW/MSYS2, install under Cygwin,
-or install using Microsoft C/C++ or other compilers. The last option will not
-provide you with any UNIX tools so you will be limited to what you can do with DOS
-batch files. One simple option for accessing a UNIX style bash terminal is
-*Git for Windows*, which can be downloaded from [their official website](https://gitforwindows.org/).
+There are several options for using [GMT on non-UNIX systems](https://docs.generic-mapping-tools.org/latest/cookbook/non-unix-platforms.html)
+such as Windows, including [Windows Subsytem for Linux](https://docs.microsoft.com/en-us/windows/wsl/),
+MinGW/MSYS2, Cygwin, or DOS batch scripts. The last option will not provide you
+with any UNIX tools so you will be limited to what you can do with DOS batch files.
+One simple option for accessing a UNIX style bash terminal is *Git for Windows*,
+which can be downloaded from [their official website](https://gitforwindows.org/).
 
 **NOTE:**
 At the installation step, you may get the warning message:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -41,10 +41,12 @@ animated GIFs.
 
 
 **NOTE:**
-It is recommended that you use a UNIX style terminal with a bash shell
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Looks good, maybe use the word "bash" somewhere like  UNIX style bash terminal.

joa-quim:
I do not agree with this recommendation (never recommended it). Off course the ``batch`` language is horrible and ``bash`` (or cousins) is a bit better but only for advanced scripting will ``bash`` be needed. One of the few examples that came to my mind is when one need to capture the output of a command into a variable. Others are loops but these are very rare cases and basically only happen in advancing scripting. But most of cases are not advanced and in those cases ``batch`` files are way more convenient (can be executed with a double click) than shell scripts.

I rarely use a ``bash`` script.

PaulWessel:
Maybe we need to be more nuanced here: This issue came up because we get lots of questions about how to enable a UNIX environment on Windows.  It is for that reason that the changes to INSTALL.md were added.  I agree that Windows users in general may not need a bash terminal but when they are sent bash scripts and need to run them they do.
So change the "recommended" to something that addresses this case.

joa-quim:
> but when they are sent bash scripts and need to run them they do.

99.xx% of the cases what they need to do is change the script file extension to ``.bat`` and replace ``#`` by ``REM``

I think we should replace the *recommend* by something like *for a more reach experience when one needs some programming capabilities such as loops and variables assignment*

PaulWessel:
I am fine with that but note our own examples are closer to 0.00 than 99.xx% I think.

joa-quim:
That's because you never consider the other hypothesis. Windows users sending ``.bat`` files to unix colleagues 😃 
And that because, despite being ~2/3 of GMT users (old download numbers told us that) they are conditioned to think themselves as lower ranked users.

PaulWessel:
OK, maybe @meghan can propose a revised paragraph that replaces a recommendation with a more practical statement for when you may want to install Git for Windows.

meghanrjones:
I removed the recommendation and added other options.

@joa-quim, you may want to submit a PR with your preferred changes to this section of the cookbook: https://docs.generic-mapping-tools.org/dev/cookbook/non-unix-platforms.html. The statement that MINGW|MSYS (now msys2) is the recommended way of running GMT on Windows predates the migration to Git. It is reasonable for newer developers (such as me) to assume that opinionated statements in the cookbook are valid to use as a reference for documentation improvements elsewhere.

anbj:
I used GMT on Windows and made batch scripts for many years.

Then I discovered Windows Subsystem for Linux (WSL) in 2016-ish. Since then, I never have - and never will - even consider touching a batch script again. Getting familiar and friendly with ``bash`` and converting my existing GMT scripts written in ``batch`` to ``bash`` made a huge difference for me. It's like it broadened the horizon. This ultimately made me convert from Windows to *nix full time on everything except my work laptop.

I think it's important to make people aware of WSL. I guess most users run Windows 10 now - then they should have WSL readily available. See e.g. https://docs.microsoft.com/en-us/windows/wsl/install-win10. Shouldn't this be mentioned?

meghanrjones:
> I used GMT on Windows and made batch scripts for many years.
> 
> Then I discovered Windows Subsystem for Linux (WSL) in 2016-ish. Since then, I never have - and never will - even consider touching a batch script again. Getting familiar and friendly with `bash` and converting my existing GMT scripts written in `batch` to `bash` made a huge difference for me. It's like it broadened the horizon. This ultimately made me convert from Windows to *nix full time on everything except my work laptop. I'm not interested in a Windows vs. *nix discussion, but I do claim that *nix OS's with it's KISS philosophy (usually), pipes and scripting capabilities suits GMT very very well.
> 
> I think it's important to make people aware of WSL. I guess many Windows users run Windows 10 now - then they should have WSL readily available. See e.g. https://docs.microsoft.com/en-us/windows/wsl/install-win10. Shouldn't this be mentioned?

OK, I have added this option to the new list in INSTALL.md. Would you be willing to open a PR with an additional section on the WSL option in https://github.com/GenericMappingTools/gmt/blob/master/doc/rst/source/cookbook/non-unix-platforms.rst (the source file for https://docs.generic-mapping-tools.org/dev/cookbook/non-unix-platforms.html)?

joa-quim:
Yes, that section of the CookBook is very old and should be rewritten and references to MSYS or Cygwin dropped or only briefly mentioned. Reference to WSL is a must be.

anbj:
> OK, I have added this option to the new list in INSTALL.md. Would you be willing to open a PR with an additional section on the WSL option in https://github.com/GenericMappingTools/gmt/blob/master/doc/rst/source/cookbook/non-unix-platforms.rst (the source file for https://docs.generic-mapping-tools.org/dev/cookbook/non-unix-platforms.html)?

I’d love to. As soon as I have time!

anbj:
This PR should be merged before I start working on the WSL-section, right?

PR #5328, opened 2021-06-12
Let colorbar handle CPTs with transparency

### LINKED ISSUE #5326: Allow psscale to place transparent CPTs
See https://forum.generic-mapping-tools.org/t/transparency-in-plot-and-scale/1763/4?u=andreas for context.

The ability of having transparency in CPT-tables exists, but ``psscale`` is not able to use it. Allowing ``psscale`` to place transparent CPTs would be a fruitful addition.

PaulWessel:
If the CPT has slices with specified transparency then we now honor that.  For discrete CPTs you may have all or individual slices be transparent, while for continuous CPTs you must either have all or no slices be transparent.  The only case we are not handling is continuous CPTs with a section that implies a change in transparency.  I added a test to demonstrate this capability:

transp_scale

Closes #5326.


anbj (approved):
In my specific use case, this seems to be working now.

Simply great!

meghanrjones (approved):

PR #5357, opened 2021-06-19
Improve code readability using named PSL vector, wedge, and matharc dim indices

PaulWessel:
It is hard enough to debug GMT without also having to remember what dim[9] contains, and that it is different for vectors and wedges.  This adds three sets of _enums_ with named indices that are used internally by GMT modules.  All tests still pass.


PaulWessel:
Anyone around to approve this one?

meghanrjones (approved):

Issue #5348, opened 2021-06-17
Bug in gmtmath when using values with units


Esteban82:
When I add units to the values I get some weird results
```
gmt math -Q 1  1  DIV = 1
gmt math -Q 1c 1  DIV = 1
gmt math -Q 1c 1c DIV = 2.54
gmt math -Q 1i 1i DIV = 2.54 
gmt math -Q 1  1i DIV = 2.54
gmt math -Q 1  1c DIV = 6.4156 (2.54 x 2.54)
```

**System information**
* Operating system: Win64 
* GMT 6.2

PaulWessel:
Good one.  So when units are given the numbers are internally converted to inches.  Then on output they are reported according to **PROJ_LENGTH_UNIT** [cm].  Your first example is unrelated to this since no units.  So 1c/1c is dimensionless but gmtmath does not know that, so it has the result as 1 (with internal unit inch) and on output that becomes 2.54 cm. Save for 1i/1i. 1/1i is therefore the same as 1i/1i. 1/1c becomes 1i/1c and 2.54 internally (dimensionless but seen as inches) so out comes 6.4156.  If I set PROJ_LENGTH_UNIT to inch and run your examples I get

```
gmt math -Q 1  1  DIV = 1
gmt math -Q 1c 1  DIV = 0.3937
gmt math -Q 1c 1c DIV = 1
gmt math -Q 1i 1i DIV = 1
gmt math -Q 1  1i DIV = 1
gmt math -Q 1  1c DIV = 2.54
```

which is better since it still converts numbers with units to internal inches, but then does not do any conversion on the output (assuming inches).

Given how **gmtmath** works I do not think I can easily fix this - i.e., keep track of units.  Maybe document that divisions with united numbers will cause confusion.  I will think about this a bit.

PaulWessel:
It would be pretty simple to turn off any unit conversion on output.  This means you would get the same as I get when setting **PROJ_LENGTH_UNIT** to inch.  I think those results are more sensible in most but not all cases.  Or, we could let **-Q** get a small enhancement:

**-Q**[**c**|**i**|**p**|**n**]

[...]

PR #5362, opened 2021-06-19
Improve gmtmath -Q behavior

PaulWessel:
See #5348 for background and explanations.  This PR implements those suggestions to make it simpler to select the output unit or have no unit conversion on output, as well as improved documentation as to what is happening.  Closes $5348.


Esteban82:
With the division it works. however, when I try MUL all the result were 1/2.54 or 1/2.54<sup>2</sup>

```
gmt math -Qi 1c 1c MUL = 0.155
gmt math -Qn 1c 1c MUL = 0.155
gmt math -Qc 1c 1c MUL = 0.3937
gmt math -Q  1c 1c MUL = 0.3937
```

Only when the units are in inches I get 1

`gmt math -Qi 1i 1i MUL = ` 1
`gmt math -Qn 1i 1i MUL = `1

PaulWessel:
Yes, and can't help that.  If you specify units then GMT will convert them to inches.  Now you have 1/2.54 inside.  If you dont want that conversion, dont give units.

Esteban82:
So, should I approve this feature, right? And then avoid using units? Maybe a warning message could be add in the docs.

PaulWessel:
Give me a minute to update the docs and if you think they are clear then you can approve.

PaulWessel:
Should now make it clear that using units on input gives internal INCHES, and if that is not desired then do not use units.  See new section in man page called _Scalar math with units_.

Esteban82 (approved):

PR #5363, opened 2021-06-20
Let grdimage resample 2nd grid used for intensities

### LINKED ISSUE #5347: Allow grdview -G option in grdimage
See [forum](https://forum.generic-mapping-tools.org/t/grdimage-with-two-datasets-on-top-of-each-other/1642/9) for background.

As Paul said: "You would like to derive the intensities from one grid (say topo) and look up the colors via the CPT and another grid, and the two grids may not have the same grid spacing? It would seem this would require a modification to the grdimage -I option to allow you to pass a grid there that is not an intensity grid but a surface from which you wish to derive intensities (instead of from the main grid)."

PaulWessel:
This follows up on issue #5347.  Turns out **-I** already can handle a second datagrid combined with the derive-modifiers, but the two grids have to have the same size and registration.  This PR changes this so that if **-I** specifies a second data grid from which intensities shall be derived (rather than from the input grid), we ensure the secondary grid is resampled to the resolution of the primary grid if it is not already matching.  I added a new test script that shows the use of different resolution grids.  Closes #5347. 

twogrids

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -1425,6 +1425,10 @@ EXTERN_MSC int GMT_grdimage (void *V_API, int mode, void *args) {
 			/* If dimensions don't match the data grid we must resample this secondary z-grid */
 			if (Grid_orig && (I_data->header->n_columns != Grid_orig->header->n_columns || I_data->header->n_rows != Grid_orig->header->n_rows)) {
 				char int_z_grd[GMT_VF_LEN] = {""}, *res = "gp";
+				if (I_data->header->wesn[XLO] > region[XLO] || I_data->header->wesn[XHI] < region[XHI] || I_data->header->wesn[YLO] > region[YLO] || I_data->header->wesn[YHI] < region[YHI]) {
+					GMT_Report (API, GMT_MSG_ERROR, "Your secondary data grid given vie -A does not cover the same area as the primary grid - aborting\n");
+					Return (GMT_GRDIO_DOMAIN_VIOLATION);
+				}
 				if (GMT_Open_VirtualFile (API, GMT_IS_GRID, GMT_IS_SURFACE, GMT_OUT|GMT_IS_REFERENCE, NULL, int_z_grd))
 					Return (API->error);
 				sprintf (cmd, "%s -R%.16g/%.16g/%.16g/%.16g -I%.16g/%.16g -r%c -G%s --GMT_HISTORY=readonly ",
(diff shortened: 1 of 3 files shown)


joa-quim:
I think that it should also check that GRD2 totally encloses GRD1.

PaulWessel:
Yes, that is reasonable, too.  Maybe later today. It is Father's Day in the US and that means forced meals, activities and fewer therapeutic debugging sessions.

Esteban82:
Happy Father's Day!! It is also in Argentina.

It would be possible to use irregular grid like the following link from the forum?

https://forum.generic-mapping-tools.org/uploads/default/original/2X/f/fd6eb02da35af9f2471aa44c9a3ab56bea2e6671.png

PaulWessel:
Likewise @Esteban82 .  I am not sure I understand what you mean by "irregular grids" since all grids in GMT have to be equidistant and not irregular?

Esteban82:
With "irregular" I mean its shape (not the spacing between nodes). Like the previous link where there is no data in the edges.

PaulWessel:
And by shape you mean there are NaNs? The shape of all grids is rectangular but many only have data in some of that rectangle.  I think NaNs are fine - you just are not going to get any shading from that part.

PaulWessel:
OK @joa-quim I added a region check, insisting that the secondary grid covers the same region (or more).

src/grdimage.c
```diff
@@ -1415,12 +1415,36 @@ EXTERN_MSC int GMT_grdimage (void *V_API, int mode, void *args) {
...
 			Return (API->error);
 		if (Ctrl->I.file) {	/* Gave a file to derive from. In case it is a tiled grid we read it in here */
-			if ((I_data = GMT_Read_Data (API, GMT_IS_GRID, GMT_IS_FILE, GMT_IS_SURFACE, GMT_CONTAINER_AND_DATA, wesn, Ctrl->I.file, NULL)) == NULL)	/* Get grid data */
+			if ((I_data = GMT_Read_Data (API, GMT_IS_GRID, GMT_IS_FILE, GMT_IS_SURFACE, GMT_CONTAINER_AND_DATA, wesn, Ctrl->I.file, NULL)) == NULL)	/* Get grid data header*/
+				Return (API->error);
+			/* If dimensions don't match the data grid we must resample this secondary z-grid */
+			if (Grid_orig && (I_data->header->n_columns != Grid_orig->header->n_columns || I_data->header->n_rows != Grid_orig->header->n_rows)) {
+				char int_z_grd[GMT_VF_LEN] = {""}, *res = "gp";
+				if (I_data->header->wesn[XLO] > region[XLO] || I_data->header->wesn[XHI] < region[XHI] || I_data->header->wesn[YLO] > region[YLO] || I_data->header->wesn[YHI] < region[YHI]) {
+					GMT_Report (API, GMT_MSG_ERROR, "Your secondary data grid given vie -A does not cover the same area as the primary grid - aborting\n");
```
joa-quim, line 1429: Typo. via not vie

PaulWessel: Also not via -A... but -I.  Fixed

joa-quim (approved):

Issue #4448, opened 2020-11-12
Add CPT of the geological time scale


Esteban82:
Add the CPT with the colors defined by the ICS.

https://ccgm.org/img/cms/2013%20Chart%20CGMW%20color%20codes.pdf

There are some done by Christian Heine. Maybe only the times must be checked. 
https://tectonicwaters.wordpress.com/2014/09/06/geological-time-scale-color-palettes-for-gmt/
https://bitbucket.org/chhei/gmt-cpts/src/master/
http://soliton.vm.bytemark.co.uk/pub/cpt-city/heine/

**Are you willing to help implement and maintain this feature?** Yes

Esteban82:
Maybe this could be discussed in the next Thursday in the community meeting.

PaulWessel:
Yes, I will add it to the agenda

Esteban82:
I copy the notes from the community meetings

-  Can be tricky to add too many domain specific ones (which ones do we add and which ones we don’t?)
- Can have them managed as remote data files
- CPTCity is great for this already
- Have a `makecpt -C @somecpt`? Point to some cpt-city URL.
-  Need to be careful with license and citations if we’re storing CPTs

In general, think it is OK to add to cache but not to GMT core since it is not generic and not tied to data we provide
- What do we do with the data cache? There are a lot of datasets but they are not listed
- We don’t want to have to maintain a lot of data files
- Could have a clearer naming scheme (like @sample_* for every dataset used in the docs).-

Esteban82:
I shared some plots with the geological CPT. I think that the plots can be useful. Maybe one or two of these plots can be polished and add in the gallery (or in the forum showcase).

Batchs scripts: [Earth_Age_CPT-City.zip](https://github.com/GenericMappingTools/gmt/files/5578893/Earth_Age_CPT-City.zip)

The important commands used were (in a batch file): 

```	
REM Links of the CPT
REM Geological CPT
rem	gmt which -G > %color% "http://soliton.vm.bytemark.co.uk/pub/cpt-city/heine/GTS2012_eons.cpt"
rem	gmt which -G > %color% "http://soliton.vm.bytemark.co.uk/pub/cpt-city/heine/GTS2012_eras.cpt"
rem	gmt which -G > %color% "http://soliton.vm.bytemark.co.uk/pub/cpt-city/heine/GTS2012_periods.cpt"
[...]

Issue #3320, opened 2020-05-12
Different line styles as example in the Gallery or Cookbook


KristofKoch:
**Description of the desired feature**

As the possibilities for styling lines has greatly increased over the last couple of months (thank you, Gentlemen) it might be beneficial to document and showcase what can be achieved with the current feature set. 

I propose to either expand the [Cookbook chapter 3.13 _Specifying pen attributes_](https://docs.generic-mapping-tools.org/latest/cookbook/features.html#specifying-pen-attributes) which is a bit sparse on the possibilities or add another example in the Gallery. 

Your thoughts about this?

**Are you willing to help implement and maintain this feature?** Yes

PaulWessel:
Hi @KristofKoch, yes of course that would be great.  The way things are implemented is that each illustration in the documentation is a script that lives in doc/scripts and it automatically becomes part of the test suite as well.  So apart from text in the relevant cookbook section, there needs to be a script and an original PostScript plot (created with PS_MEDIA = letter) in doc/script.  Then see other scripts and how they are addressed in the docs.

KristofKoch:
Hi @PaulWessel thank you for the primer on how to edit the cookbook. 

I currently think of showing the fancy stuff in the cookbook but explaining it might take up a lot of space. I'm undecided how to get around the problem as at least for me the main reference for experimenting with new things is the cookbook. 

Are the examples considered documentation which I can reference from the cookbook? Basically show the fancy stuff in the cookbook and direct the user to example fifty something if he wants to know how it was done?

PaulWessel:
yes, the examples are part of the documentation and can be referenced from anywhere I believe; @seisman knows how for sure.

seisman:
Yes, examples can be referenced anywhere in the cookbook.

Currently, the "Examples Gallery" lists 52 fancy, complicated examples. It's good to show how powerful GMT is, but users may find it hard to learn these examples because each example may use 5 or more GMT modules. 

Ideally, we'd like to have a new gallery of simple examples. Each example uses 1 or 2 commands, and plot a simple basemap, a time axis or different styling lines as @KristofKoch proposed, just like the matplotlib Gallery (https://matplotlib.org/3.2.1/gallery/index.html).

Actually, GMT already has some simple and useful examples, but they're all hidden in the long cookbook.

joa-quim:
We can this to the Thursday's meeting agenda. I would rather start another document with simpler examples than to continue expanding the Gallery.

[...]

PR #5364, opened 2021-06-20
Add a page for simple examples

seisman:

As mentioned in https://github.com/GenericMappingTools/gmt/issues/4448 and https://github.com/GenericMappingTools/gmt/issues/3320, the current 52 examples in the [Gallery](https://docs.generic-mapping-tools.org/latest/gallery.html) page are a little too complicated. 

Users may find that simple examples are more useful to learn GMT quickly. For example, both the [PyGMT Gallery](https://www.pygmt.org/dev/gallery/index.html) and the [GMT.jl Gallery](https://www.generic-mapping-tools.org/GMT.jl/dev/gallery/tables/) provides many simple examples. Users may also want to contribute simple examples rather than complicated ones (e.g., #4448 and #3320).

In this PR, I create a new gallery page for simple examples. The new page is called "Examples", and the old "Example Gallery" page is renamed to "Gallery" (We can change them to other names after some discussions).

The "Gallery" page contains complicated examples, and the scripts are located in the `doc/examples` directory. 

The new "Examples" page contains simple examples (e.g., plotting lines with different styles). Each example is stored in the `doc/rst/source/examples/exXXX` directory. Each directory should contain the following files:

- an `index.rst` file explaining the example
- an `exXXX.sh` script file
- necessary and small data files [optional]

In this PR, I add a demo example "ex001" to show how the new page works. The generated image (ex001.ps) is not included, because

1. Adding hundreds of PS files increases the repository size quickly
2. These simple examples should always work, and should be already covered by existing tests. I don't think we should add them to the GMT tests.
3. The generated PNG image is uploaded to github (by dragging the image to any issue comment), and the github image link should be valid for a very long time.

**Preview**: https://gmt-git-example-page-gmt.vercel.app/examples/index.html


seisman:
Ping @PaulWessel @joa-quim @meghanrjones @Esteban82 @KristofKoch for comments

PaulWessel:
This is great, @seisman! I have a few comments and suggestions:

I agree this should focus on very simple things to demonstrate them.  Might there be separate categories or groups we want to keep separate?  I am thinking

- Basemaps (various basemap examples)
- Coastlines (various coast examples)
- symbols (various plot -S examples)
- lines/polygons (various plot examples)
- grid images (grdimage, grdview)
- contour maps (grdcontour, pscontour)

Maybe others? I think just having a hodge-podge of examples will be frustrating.

seisman:
> I agree this should focus on very simple things to demonstrate them. Might there be separate categories or groups we want to keep separate? I am thinking
> 
> * Basemaps (various basemap examples)
> * Coastlines (various coast examples)
> * symbols (various plot -S examples)
> * lines/polygons (various plot examples)
> * grid images (grdimage, grdview)
> * contour maps (grdcontour, pscontour)
> 
> Maybe others? I think just having a hodge-podge of examples will be frustrating.

Yes, similar categories are being used in the [PyGMT Gallery](https://www.pygmt.org/dev/gallery/index.html) and we can also do that in the GMT side.

The `index.rst` is very flexible and can include any figures in any categories. We can even include images that are already in the cookbook.

anbj:
This is an excellent idea!

* Any  maximum number of script lines? How simple is simple?
* A script may be catagorized in multiple ways, which take precedence?
* Is it allowed to use other tools, e.g. ``gdal``?
* Should the threshold be low for creating an example?
    * Simply, "look, I made this" or
    * "I bet other people are struggling with this as well, better create an example"?

joa-quim:
We have talked about this many times. I think we should aim to mimic the example types like the Matplotlib example pages. Don't think the main goal here is to have very simple plots, but to have simple scripts (as much as they can be) to produce practical figs. *xy, bars, histograms, images, etc...*

seisman:
> I think we should aim to mimic the example types like the Matplotlib example pages. Don't think the main goal here is to have very simple plots, but to have simple scripts (as much as they can be) to produce practical figs. _xy, bars, histograms, images, etc..._

Yes, I agree.

> * Any maximum number of script lines? How simple is simple?

As mentioned by @joa-quim, the scripts should be as simple as possible. Some simple but practical examples can be found at the [PyGMT Gallery](https://www.pygmt.org/dev/gallery/index.html), the [GMT.jl Gallery](https://www.generic-mapping-tools.org/GMT.jl/dev/gallery/tables/) and [Matplotlib Gallery](https://matplotlib.org/stable/gallery/index.html). We can even create some examples from the PyGMT or GMT.jl gallery, so that users can know how to plot the same plot using GMT CLI, PyGMT or GMT.jl.

> * Is it allowed to use other tools, e.g. `gdal`?

Yes, any command-line tools should be allowed, but a note should be added to make it clear that external commands are needed.

> * Modern mode, classic mode or both?

I would prefer the modern mode only. Mixing modern and classic modes in the examples would make users more confusing.

joa-quim:
Remember that all GMT.jl scripts when using the **Vd=1** option will print the generated GMT command that can be run directly (plus the input data file name).

KristofKoch:
I like it @seisman – but where to draw the line? Currently the docs for each program have that example section at the end with mostly one-liner examples but without the resulting plots. This might be a good starting point? I feel it is a bit difficult to separate the examples properly. We got some examples in the man pages, some in the cookbook and some on the current examples page. 

I think we can all agree that the learning curve for the finer points is still rather steep. So to quote @anbj:

> "look, I made this"

would be the best approach in my eyes. Maybe even multiple examples for different options. `-B` comes to mind.

meghanrjones:
The matplotlib gallery includes a `showcase` section, which seems analogous to the current GMT example gallery section and also an "I made this" section. I think that the other sections should be about simple scripts that demonstrate how to accomplish a single, practical goal using GMT (but may leverage other tools), similar to Joaquim's suggestion.

joa-quim:
Between GMT.jl and PyGMT we already have quite a few of those basic but useful examples. We could gather those that are common, or can be easily implemented in the other wrapper, in a common group. Then each image would have 3 links pointing to the languages in which they can be written.

anbj:
One more question:

Should the examples be fully self contained? That is, should the code always produce a finalized figure?

I can think of a couple of cases where a proof-of-concept is enough/the point of the example. E.g.

```
gmt grdcontour [this is where the magic happens] -K -O >> $ps
```

anbj:
And yet more:

How does this new section relate to the existing https://docs.generic-mapping-tools.org/latest/users-contrib-scripts.html?

Esteban82:
Great idea. It would be possible to add labels (like here or the keyword in a paper) for each example? I think that it would help the users. Also some scripts from the [test folder](https://github.com/GenericMappingTools/gmt/tree/master/test) could be add.

meghanrjones:
> And yet more:
> 
> How does this new section relate to the existing https://docs.generic-mapping-tools.org/latest/users-contrib-scripts.html?

This section seems the same to me as the showcase on the forum. Could we consolidate?

seisman:
Sorry for forgetting this PR for months.

> but where to draw the line? Currently the docs for each program have that example section at the end with mostly one-liner examples but without the resulting plots. This might be a good starting point? I feel it is a bit difficult to separate the examples properly. We got some examples in the man pages, some in the cookbook and some on the current examples page.

The `index.rst` page is quite flexible, so it's even possible to include existing figures in anywhere. For example, we can add histogram plots in the example section of the `histogram` man page, and also directly show them in the new example gallery.

> Should the examples be fully self contained? That is, should the code always produce a (finalized) figure?
> 
> I can think of a couple of cases where a proof-of-concept is enough/the point per se. E.g.
> 
> ```
> gmt grdcontour [this is where the magic happens] -K -O >> $ps
> ```

I think the examples should be fully self contained so that users can copy the script, run it and then modify it. Without a final figure, people may have to read a long paragraph to understand what the example does and what the magic is.
 
> How does this new section relate to the existing [docs.generic-mapping-tools.org/latest/users-contrib-scripts.html](https://docs.generic-mapping-tools.org/latest/users-contrib-scripts.html)?

I think the section is here for some historic reasons, but since there is only one script, we may move it to here.

> It would be possible to add labels (like here or the keyword in a paper) for each example? I think that it would help the users. 

It's definitely possible to add labels or keywords. It would be better if we can filter examples by lables/keywords, but it's challenging for me (need more knowledge about javascript and other web knowledge),

> Also some scripts from the [test folder](https://github.com/GenericMappingTools/gmt/tree/master/test?rgh-link-date=2021-06-23T14%3A28%3A18Z) could be add.

Yes, but we may have to duplicate the scripts because the doc directory may not be able to access the test directory.

> This section seems the same to me as the showcase on the forum. Could we consolidate?

We can also show the examples posted in the forum and add links to them.

Esteban82:
Maybe this could be talk is next community meeting?

meghanrjones:
> Maybe this could be talk is next community meeting?

Sounds good, I will add this to the agenda.

gd-a:
What's the status of this PR ? Do we have a list of examples that need to be done ?

seisman:
Sorry again for forgetting this PR. 

I agree with @meghanrjones's 3 points that a separate repository is a better option for GMT examples. So, As I understand it, what we should do is:

- Create a separate repository (e.g., `gmt-examples`) to host examples
- Create a new site from the new repository. The site domain can be 
  - `gallery.generic-mapping-tools.org`
  - `showcase.generic-mapping-tools.org`
  - `examples.generic-mapping-tools.org`
- After the new site is live, we should add the link to the sidebar of the `docs.generic-mapping-tools.org` site

My question is: do we still want to use dvc to manage these example images as mentioned in #5724 or do we just upload these static images to GitHub?

meghanrjones:
> Sorry again for forgetting this PR.
> 
> I agree with @meghanrjones's 3 points that a separate repository is a better option for GMT examples. So, As I understand it, what we should do is:
> 
>     1. Create a separate repository (e.g., `gmt-examples`) to host examples
> 
>     2. Create a new site from the new repository. The site domain can be
>        
>        * `gallery.generic-mapping-tools.org`
>        * `showcase.generic-mapping-tools.org`
>        * `examples.generic-mapping-tools.org`
> 
>     3. After the new site is live, we should add the link to the sidebar of the `docs.generic-mapping-tools.org` site
> 
> 
> My question is: do we still want to use dvc to manage these example images as mentioned in #5724 or do we just upload these static images to GitHub?

Sorry for leaving this PR without a response for so long. Since we aren't using these examples as tests, I think we can use the sphinx_gmt extension to generate the images when building the docs rather than storing the static images on GitHub or DAGsHub. Do you mind if I create the `gmt-examples` repository and open a PR with a potential structure using some of the executable books extensions and the sphinx_gmt extension?

meghanrjones:
I'll get started on the `gmt-examples` repository today unless there are any objections, since it would be particularly helpful to put in some examples for vector attributes.

PR #5366, opened 2021-06-20
Improve the grdgdal order in the core modules

core-man:

Module `grdgdal` is classified as [**Miscellaneous**](https://docs.generic-mapping-tools.org/latest/modules.html#miscellaneous). Would it be better to near those `grd*` modules in the list of [**Core Modules**](https://docs.generic-mapping-tools.org/latest/modules.html#core-modules)?

BTW, I don't know if it could be better to add `img2google`, which is classified as [**Utility Scripts**](https://docs.generic-mapping-tools.org/latest/modules.html#utility-scripts), to [**IMG**](https://docs.generic-mapping-tools.org/latest/modules.html#img), which only contains `img2grd` now.

Fixes #

**Reminders**

### DIFF
--- a/doc/rst/source/modules.rst
+++ b/doc/rst/source/modules.rst
@@ -212,7 +212,6 @@ Core Modules
     - :doc:`gmtconnect`
     - :doc:`gmtconvert`
     - :doc:`gmtdefaults`
-    - :doc:`grdgdal`
     - :doc:`gmtget`
     - :doc:`gmtinfo`
     - :doc:`gmtlogo`
@@ -234,6 +233,7 @@ Core Modules
     - :doc:`grdconvert`
     - :doc:`grdcut`
     - :doc:`grdedit`
+    - :doc:`grdgdal`
     - :doc:`grdfft`
     - :doc:`grdfill`
     - :doc:`grdfilter`
--- a/doc/rst/source/modules.rst
+++ b/doc/rst/source/modules.rst
@@ -233,10 +233,10 @@ Core Modules
     - :doc:`grdconvert`
     - :doc:`grdcut`
     - :doc:`grdedit`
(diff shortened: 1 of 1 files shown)


joa-quim:
Hmm, not sure. ``grdgdal`` is not a GMT core module. It's a wrapper to GDAL programs offering also the GMT IO interface. That's why it was classified in the *Miscellaneous*.

PaulWessel:
Well, yes and no. **grdgdal** _is_ a GMT core module since it is not a supplement one.  But I think its place under miscellaneous is reasonable.

meghanrjones (approved):
Regardless of the placement in miscellaneous, I think we should fix the alphabetical order.

core-man:
> Regardless of the placement in miscellaneous, I think we should fix the alphabetical order.

Agreed, we could fix the alphabetical order for the core module list.

doc/rst/source/modules.rst
```diff
@@ -234,6 +233,7 @@ Core Modules
     - :doc:`grdconvert`
     - :doc:`grdcut`
     - :doc:`grdedit`
+    - :doc:`grdgdal`
```
meghanrjones, line 236:
```suggestion
```

doc/rst/source/modules.rst
```diff
@@ -234,6 +233,7 @@ Core Modules
...
     - :doc:`grdfft`
     - :doc:`grdfill`
     - :doc:`grdfilter`
```
meghanrjones, line 239:
```suggestion
    - :doc:`grdfilter`
    - :doc:`grdgdal`
```

core-man: thanks, I didn't know my brain told me **g** is before **f** 😄

Issue #5365, opened 2021-06-20
Color rendering in GMT-6.1~6.2 not match the exact value


TristanBlus:
**Rendered color not matching the value.**

Starting from GMT-6.1, I found that the rendered color was incorrect (value - color not matching), which worked well in the older version, see the picture comparision same data and cpt. Note, the cpt file and data were exactlly the same. 

**Full script that generated the error**

```
# PASTE CODE HERE
```

**Full error message**

```
# PASTE ERROR MESSAGE HERE
```

**Actual outcome**

![image](https://user-images.githubusercontent.com/37327637/122660184-dc472a00-d1b1-11eb-938d-7095cf36c522.png)
correct rendered color in GMT-6.0.0

**Expected outcome**

![image](https://user-images.githubusercontent.com/37327637/122660198-0d275f00-d1b2-11eb-9a0a-d026882cd96f.png)
same data and cpt file,  output in GMT-6.2.0

**System information**

* Operating system: Ubuntu-20.04
* GMT version (`gmt --version`): 6.1.0 ~ 6.2.0

PaulWessel:
We won't be able to examine this without a small test data, cpt, script etc.  We have ~1000 test script that produces test plots that we check for any failures so if colors changes we would now.  For starters, perhaps you can post your CPT here?

TristanBlus:
> We won't be able to examine this without a small test data, cpt, script etc. We have ~1000 test script that produces test plots that we check for any failures so if colors changes we would now. For starters, perhaps you can post your CPT here?

OK, the cpt file were attached.(renamed to '.txt)
Besides, i have tested standard cpt in GMT, it also failed to rendered correctly.

[shis.cpt.txt](https://github.com/GenericMappingTools/gmt/files/6681660/shis.cpt.txt)

PaulWessel:
Here is your CPT rendered with ±0.6 intensities in GMT 6.0.0:

![6 0 0](https://user-images.githubusercontent.com/26473567/122660799-fc2e0d80-d11f-11eb-9688-acdaec83e437.png)

Here is your CPT rendered with ±0.6 intensities in GMT 6.2.0:

![6 2 0](https://user-images.githubusercontent.com/26473567/122660800-018b5800-d120-11eb-981f-a3456f9c29d3.png)

[...]

PR #5373, opened 2021-06-21
Cannot trust GDAL metadata without reading the file

PaulWessel:
Addresses the issue in #5365.  Looks like a change in GDAL since no change to this in GMT since 2016. All tests pass as before.


joa-quim (approved):

PaulWessel:
git blame shows no such change by us so maybe an internal GDAL thing?

joa-quim:
Probably. I have a distant recall of the 'true' meaning quick-approximate scan. Maybe at one point they made it trust the metadata. But the docs still says
> bApproxOK – TRUE if an approximate (faster) answer is OK, otherwise FALSE.

PaulWessel:
OK, thanks. Merging in and done.

Issue #5377, opened 2021-06-23
Plot title and forgetting the t in +t


anbj:
This is from the *things devs maybe should not spend time on* department.
Not a bug, just an observation. This is deliberate user error.

If you want a plot title and forget the `t` in `+t` it will still make a plot with a malformed title if the first letter happens to be a `t`:

```
$ gmt psbasemap -R0/10/0/10 -BWSne+"test" -pdf delete
```

If the first letter in the title is an `f`, you get an error:

```
$ gmt psbasemap -R0/10/0/10 -BWSne+"fest" -pdf delete
basemap [WARNING]: Option -B: Cannot use +f for Cartesian axes - modifier ignored
----snip----
```

I guess this means that gmt strips away the `"`?

Use [error] case:
"My plot is titled *gred*, but I'm getting a red fill instead. What's going on?"

```
gmt psbasemap -R0/10/0/10 -BWSne+"gred" -pdf delete
```

Many plot titles is made up of more than one word, causing a need for a pair of "". Would is make sense to always require `""`?

anbj:
And another one:

Calling ``$ gmt`` gives an exit code of ``79``, but ``0``. Is this deliberate?

PaulWessel:
GMT does not see those quotes as the shell strips them away.  So what GMT gets is -BWSne+test which will plot the title "est".  This is why it is hard for us to deal with the title after +t and any random +character sequences that may be part of a legitimate title.  Safest to use the octal code for the plus symbol in those cases. Same for -BWSne+fest which will find **+f**, a valid modifier for geographic axes but not Cartesian, etc.

anbj:
Thanks for the explanation, Paul.

```
$ echo $?
0
```

Perfect.

anbj:
Closing.

PR #5379, opened 2021-06-23
Let gmt return 0 when successfully printing usage message

PaulWessel:
See #5377 for background.  When **gmt** is run without arguments on the command line we print the usage message.  This is not an error situation so the error code returned should be 0.


joa-quim (approved):

anbj:
```
$ echo $?
0
```

Perfect.

Issue #5341, opened 2021-06-16
Improve usage documentation and break synopsis lines for all GMT modules


meghanrjones:
https://github.com/GenericMappingTools/gmt/pull/4796/ added a `GMT_Usage` function to improve the formatting for GMT command line modules and updated the syntax for plot, basemap, and all the common options. Now, it is necessary to update the syntax for all the other modules. Help is greatly appreciated!

The documentation for GMT_Usage is in `src/gmt_api.c` at line 13563. In general, calls to `GMT_Message(API, GMT_TIME_NONE,…` will be replaced with `GMT_Usage(API, indent_level, …`, where `indent_level` is one of the following:

`0`: for synopsis message
`1`: for listing of options (e.g., -R, -J, etc.), which will be wrapped and indented if it continues past one lint.
`-1`: for the same indentation level as one, but without further indentation if it exceeds one line.
`2`: for a paragraph under an option that should be wrapped and indented if it exceeds a line.
`-2`: Same as -1 but for level 2
`3`: for separating modified under an option
`-3`: same as -1 but for level 3
And so on...

Steps for updating the usage message for a module (subject to revision):
1. The checklist below includes all the GMT modules. Before starting working on a module, add your name next to the module in the checklist. If you do not have edit access, post a comment requesting that your name be added next to the module in the checklist. This is an important step to coordinate efforts among different contributors.
2. Find the source code for the module that you will work on. For example the source code for `batch` is `src/batch.c`.
3. Find the section of the source code that creates the usage message (e.g., search for `usage:`)
4. Reformat the usage message from multiple calls to `GMT_Messsage(API, GMT_TIME_NONE,…` to a single call to `GMT_Usage (API, 0…`
5. Reformat the messages that list options from `GMT_Message (API, GMT_TIME_NONE, “\t…` to `GMT_Usage (API, 1 “\n…`
6. Reformat the option descriptions from multiple calls to `GMT_Message (API, GMT_TIME_NONE, “\t…` to one call to `GMT_Usage (API, -2, “…`
7. Reformat the modifiers/directives descriptions from calls to `GMT_Message (API, GMT_TIME_NONE, “\t…` to calls to `GMT_Usage (API, 3, “…`, with one GMT_Usage call per modifier/directive.
8. Update any `gmt_*_syntax` calls.
9. Options that take no arguments whatsoever (e.g., -N) can be set on the same line as the explanation.

To do:
- [x] batch Paul
- [x] begin Kristof
[...]

PR #5380, opened 2021-06-23
Update usage message for batch

PaulWessel:
As directed by issue #5341.

### DIFF
--- a/doc/rst/source/batch.rst
+++ b/doc/rst/source/batch.rst
@@ -14,7 +14,7 @@ Synopsis
 
 **gmt batch** *mainscript*
 |-N|\ *prefix*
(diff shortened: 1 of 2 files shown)


meghanrjones (approved):
GitHub line suggestions is being buggy for me right now, so here are the two comments:
- `Erase input scripts (<mainscript> and any files via -I, -S` is missing a closed parentheses. Could also add that it is incompatible with -Q. 
- -T+w does not mention the option to specify `<str>`.

PaulWessel:
OK done. Usage had **+W** for tabs only as well, which is handy so I implemented that and improved the **+w**|**W** docs, including the **+w*_str_ option.

PR #5392, opened 2021-06-26
Rewrite how -i works internally for external sources

PaulWessel:

There were several conceptual issues with how we had implemented **-i** when reading from matrices and vectors (and datasets) from memory.  This has now been fixed (I hope) and we are now using a procedure more like this:

```
for (col = 0; col < n_columns; col++) {
   col_pos_out = gmtlib_pick_in_col_number (GMT, col, &col_pos_in);
   output[col_pos_out] = gmt_M_convert_col (GMT->current.io.col[GMT_IN][col], input[col_pos_in]);
}
```

Thus, the `col` is just a dummy loop variable and the in/out-cols can be independent from that. `col_pos_out` covers the same values that col does, but not necessarily in any particular order, and `col_pos_in` may go outside that range and even have repeats.  I added a new test program (_testapi_columns.c_) driven by test script _apicolumns_icol.sh_, which passes. Note: Only testing matrices and vectors here as testing datasets revealed some unrelated issues that will be dealt with later.

Keeping this at WIP to see if it fixes recent trouble related to external tables in PyGMT (@seisman) and GMT.jl (@joa-quim).

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -1320,29 +1320,34 @@ GMT_LOCAL double gmtapi_select_record_value (struct GMT_CTRL *GMT, double *recor
 	return (val);
 }
 
-GMT_LOCAL unsigned int gmtapi_pick_in_col_number (struct GMT_CTRL *GMT, unsigned int col) {
-	/* Return the next column to be selected on input */
-	unsigned int col_pos;
-	if (GMT->common.i.select)	/* -i has selected some columns */
-		col_pos = GMT->current.io.col[GMT_IN][col].col; /* Which data column to pick */
-#if 0
-	else if (GMT->current.setting.io_lonlat_toggle[GMT_IN] && col < GMT_Z)	/* Worry about -: for lon,lat */
-		col_pos = 1 - col;	/* Read lat/lon instead of lon/lat */
-#endif
+unsigned int gmtlib_pick_in_col_number (struct GMT_CTRL *GMT, unsigned int col, unsigned int *col_pos_in) {
+    /* If -i is not active then we simply return col_pos_out = col_pos_in = col.  Done!
+     *
+     * With -i, we must return correct value for both the in and out column positions based on the info in the structure.
+     * Here, col is a loop variable that goes from 0 to number of desired output columns.  However, that col number
+     * is likely neither the input(col) we wish to get nor the output(col) where we want to place it in the output array.
+     * Instead, we set these two different columns via col_pos_in and col_pos_out so that output(col_pos_out) = input(col_pos_in).
+     * col_pos_out will uniquely touch the same values as col loops over, but col_pos_in may be outside that range
+     * and even have repeated column values (e.g., due to -i0,1,2,2+s5). */
+	unsigned int col_pos_out;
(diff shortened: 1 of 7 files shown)


joa-quim:
The previous broken cases now work well with this branch.

PaulWessel:
Thanks, taking of WIP.  The old fix was a wack-a-mole fix. Fixed one thing and broke another.  No way to fix both without this rewrite.  Once tested with PyGMT and approved I will merge and then work on similar upgrades for a few other places in the API where this appears, and it may likely lead to improvements such as doing -o2,1,2,2 etc being allowed.

joa-quim:
Thanks. And at the end we'll need another release.

seisman:
The original `contour` test now works, but there is still a failure for `rose`.

The `load_fractures_compilation` loads the `@fractures_06.txt` file and returns a DataArray with two columns, "length" and "azimuth". In the first figure, "length" and "azimuth" is passed, so the figure is correct. In the second figure, "azimuth" and "length" is passed, so `incols=[1, 0]` is used to swap the input columns, but the second figure is wrong:
```python
import pygmt
from pygmt.datasets import load_fractures_compilation

data = load_fractures_compilation()

data1 = data[["length", "azimuth"]]
fig = pygmt.Figure()
fig.rose(
    data=data1,
    region=[0, 1, 0, 360],
    sector=10,
    diameter="10c",
    frame=["x0.2g0.2", "y30g30", "+glightgray"],
    color="red3",
    pen="1p",
    orientation=False,
    norm=True,
    vectors=True,
    no_scale=True,
    shift=False,
)
fig.savefig("rose1.png")

data2 = data[["azimuth", "length"]]
fig = pygmt.Figure()
fig.rose(
    data=data2,
    incols=[1, 0],
    region=[0, 1, 0, 360],
    sector=10,
    diameter="10c",
    frame=["x0.2g0.2", "y30g30", "+glightgray"],
    color="red3",
    pen="1p",
    orientation=False,
    norm=True,
    vectors=True,
    no_scale=True,
    shift=False,
)
fig.savefig("rose2.png")
```

| First figure | Second figure |
|---|---|
|image| image |

PaulWessel:
OK, will look at this later - have some errands that need to be done first...

PaulWessel:
Should be OK now, @seisman.

seisman (approved):

seisman:
Now all PyGMT tests pass.

meghanrjones:
@PaulWessel , you mentioned "Only testing matrices and vectors here as testing datasets revealed some unrelated issues that will be dealt with later". I'm debugging https://github.com/GenericMappingTools/pygmt/issues/1440 and find that the `col_pos` used in:

https://github.com/GenericMappingTools/gmt/blob/b33d6d8e9968eaaa46ee24864489ab0e17f60a6f/src/gmt_api.c#L3932-L3940

is incorrect. Do you know if this is related to the known issues for  for `gmtapi_import_dataset`? If not, I can continue debugging but don't want to repeat efforts.

PaulWessel:
No, please continue, and let me know if you need help.

PR #5394, opened 2021-06-27
docs-cleanup: begin module

KristofKoch:

As per #5341 reformatted the usage message for the `begin` module.
 
One minor issue – something is messing with the line breaks. See line breaks in plot formats list for entries `jpg` and `PNG`:

broken line breaks
You can replicate it with a 45 char wide terminal window. (Easy check with `tput cols`).


PaulWessel (commented):
Looks good!

PaulWessel (approved):

PR #5396, opened 2021-06-27
Clean up gmt math usage

PaulWessel:
This is part of the process in #5341. Because of the long list of aligned operators and args I have left GMT_Message for that part and use fewer whitespace on the output so that it will only wrap (unelegantly) for very narrow terminals.


KristofKoch:
No clever idea for the operator list either. A "very narrow" 110 char minimum terminal width made me smile. 

A layout question: single letter optional arguments have their explanation in the same line? (See `-I` and `-L`) All other arguments have their explanation beneath.

PaulWessel:
I will try another scheme to see if I can get the wrapping to work here as well.

PaulWessel:
Hi @KristofKoch please check this update out.

PaulWessel:
Also pinging @meghanrjones on this update.

KristofKoch:
Neat! Works great! Maybe add another space after the `#args` column for better readability?

Screenshot missing space

I found the `Returns` column hard to read if it starts with a number. (See `INRANGE`, `INV`, `ISFINITE` and `ISNAN` in screenshot above). I think it helps readability with two spaces as in this mockup below:

mockup

PaulWessel:
Can do, but would it be better with something else? let me know.

```
INV        1 1  1 / A
INV        1 1 : 1 / A
INV        1 1 = 1 / A
INV        [1 1] = 1 / A
INV        [1 1]: 1 / A

```

KristofKoch:
> […] would it be better with something else?

Mh, I think I'm happy with the double space. What's your preference?

PaulWessel:
Happy to make you happy on this - will do the 2 spaces.

PaulWessel:
Are you OK with the list of special symbols as they are or should we remove the equal sign and use two spaes there too:

     D_EPS (double eps)  2.2204460492503131e-16
     TMIN                min "time" value set via -T

KristofKoch:
I'm happy with the current design of the special symbols list.

PaulWessel:
OK, if there is nothing else then we need an approval from you or @meghanrjones .

KristofKoch (approved):
Great work!

PR #5399, opened 2021-06-28
Update mapproject usage

PaulWessel:
Part of the efforts described at #5341.

### DIFF
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -293,7 +293,7 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Usage (API, 3, "+f Format the elapsed time using the ISO 8601 convention. "
 		"The FORMAT_CLOCK_OUT setting is used to determine the ss.xxx format.");
 	GMT_Usage (API, 3, "+t Set epoch to get absolute time along track.");
-	GMT_Option (API, "V,bi2,bo,d,e,f,g,h,i,j,o,p,q,s,:,.");
+	GMT_Option (API, "bi2,bo,d,e,f,g,h,i,j,o,p,q,s,:,.");
 
 	return (GMT_MODULE_USAGE);
 }
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -199,14 +199,14 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 
 	if (level == GMT_SYNOPSIS) return (GMT_MODULE_SYNOPSIS);
 
-	GMT_Message (API, GMT_TIME_NONE, "  REQUIRED ARGUMENTS:\n");
+	GMT_Usage (API, -2, "Note: On output, selected items will be pasted at the end of the input record. "
+		"For optional items the order is -A before -G before -L before -Z, if any of these options are used.");
+	GMT_Message (API, GMT_TIME_NONE, "\n  REQUIRED ARGUMENTS:\n");
+	GMT_Option (API, "<");
 	GMT_Option (API, "J,R");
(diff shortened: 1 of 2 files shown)


src/mapproject.c
```diff
@@ -189,84 +189,111 @@ static void Free_Ctrl (struct GMT_CTRL *GMT, struct MAPPROJECT_CTRL *C) {	/* Dea
...
-	GMT_Message (API, GMT_TIME_NONE, "\t   -Qe shows ellipsoid parameters.\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   -Qd shows datum parameters.\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t-S Suppress points outside region.\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t-T Perform coordinate transformation from datum <from> to datum <to>.\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   Prepend h if input data are lon, lat, height [Default sets height = 0].\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   Specify datums using datum ID (see -Qd or man page) or as <ellipsoid>:<dx>,<dy>,<dz>.\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   where <ellipsoid> may be ellipsoid ID (see -Qe or man page) or <semimajor>[,<inv_flattening>].\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   <from> = - means WGS-84.  If /<to> is not given we assume WGS-84.\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   The -T option can be used as pre- or post- (-I) processing for -J -R.\n");
+	GMT_Usage (API, -2, "Note: On output, selected items will be pasted at the end of the input record. "
```
meghanrjones, line 208: this note shows up under table, even though not directly related.

It seems most usage messages list table as optional while most html docs list table as required. Is it worth making this consistent throughout?

src/mapproject.c
```diff
@@ -189,84 +189,111 @@ static void Free_Ctrl (struct GMT_CTRL *GMT, struct MAPPROJECT_CTRL *C) {	/* Dea
...
+	GMT_Usage (API, 1, "\n-N[a|c|g|m]");
+	GMT_Usage (API, -2, "Convert from geodetic to auxiliary latitudes; use -I for inverse conversion. Optionally, append a conversion code:");
+	GMT_Usage (API, 3, "a: Authalic.");
+	GMT_Usage (API, 3, "c: Conformal.");
+	GMT_Usage (API, 3, "g: Geocentric [Default].");
+	GMT_Usage (API, 3, "m: Meridional.");
+	GMT_Usage (API, 1, "\n-Q[e|d]");
+	GMT_Usage (API, -2, "List all projection parameters and stop.  Optionally, select a subset:");
+	GMT_Usage (API, 3, "e: Show ellipsoid parameters only.");
+	GMT_Usage (API, 3, "d: Show datum parameters only.");
```
meghanrjones, line 267:
```suggestion
	GMT_Usage (API, 1, "\n-Q[d|e]");
	GMT_Usage (API, -2, "List all projection parameters and stop.  Optionally, select a subset:")
	GMT_Usage (API, 3, "d: Show datum parameters only.");
	GMT_Usage (API, 3, "e: Show ellipsoid parameters only.");
```

meghanrjones (approved):

src/mapproject.c
PaulWessel: Maybe I can place the Note ahead of REQUIRED ARGUEMNTS instead?
As for table.  I think we discussed this at some point and agreed that input is required but the table name is not since if missing we read stdin.  So the HTML is right and we should fix the usage messages.

meghanrjones: Sounds good.

PR #5403, opened 2021-06-28
gmt.c: add classic option to --new-script

anbj:
I've added the option ``--new-script-classic`` to ``gmt`` to make a script template without the modern mode related output.
You know, for the old-timers.

This was just an experiment to play with C. Consider to merge. I wont be sad if you choose not to.

### DIFF
--- a/src/gmt.c
+++ b/src/gmt.c
@@ -268,17 +268,20 @@ int main (int argc, char *argv[]) {
 					printf ("%s GMT modern mode %s template\n", comment[type], shell[type]);
 				}
 				printf ("%s Date:    %s\n%s User:    %s\n%s Purpose: Purpose of this script\n", comment[type], stamp, comment[type], name, comment[type]);
-				switch (type) {
-					case 0: printf ("export GMT_SESSION_NAME=$$	# Set a unique session name\n"); break;
-					case 1: printf ("setenv GMT_SESSION_NAME $$	# Set a unique session name\n"); break;
-					case 2: printf ("REM Set a unique session name:\n");	/* Can't use $$ so output the PPID of this process */
-						printf ("set GMT_SESSION_NAME=%s\n", api_ctrl->session_name);
-						break;
+				if (strstr (argv[arg_n], "classic") == NULL) { /* Do not include modern mode output if classic */
+					switch (type) {
+						case 0: printf ("export GMT_SESSION_NAME=$$	# Set a unique session name\n"); break;
+						case 1: printf ("setenv GMT_SESSION_NAME $$	# Set a unique session name\n"); break;
+						case 2: printf ("REM Set a unique session name:\n");	/* Can't use $$ so output the PPID of this process */
+							printf ("set GMT_SESSION_NAME=%s\n", api_ctrl->session_name);
+							break;
+					}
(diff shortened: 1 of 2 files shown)


src/gmt.c
```diff
@@ -340,25 +343,25 @@ int main (int argc, char *argv[]) {
...
+			fprintf (stderr, "  --show-library                Show path of the shared GMT library.\n");
+			fprintf (stderr, "  --show-modules                Show all modern module names.\n");
+			fprintf (stderr, "  --show-modules-core           Show all modern module names (core only).\n");
+			fprintf (stderr, "  --show-plugindir              Show directory for plug-ins.\n");
+			fprintf (stderr, "  --show-sharedir               Show directory for shared GMT resources.\n");
+			fprintf (stderr, "  --show-userdir                Show full path of user's ~/.gmt dir\n");
+			fprintf (stderr, "  --version                     Print GMT version number.\n\n");
 			fprintf (stderr, "if <module-options> is \'=\' we call exit (0) if module exist and non-zero otherwise.\n\n");
 			status = GMT_NOERROR;
 		}
```
joa-quim, line 367: Please, do not change the formatting. Those tabs above should not be there. Only spaces guarantee a good alignment in all editor settings (and possible terminals)

anbj: I'm very certain I used spaces when I made the changes, and double checked now. Do they appear as tabs to you?

joa-quim:
I don't know how to load your branch without doing a fork of your fork but when I see this displayed by github it normally means tabs
image

anbj:
You can check out a PR branch with:

```
git fetch origin pull/<ID>/head:<BRANCHNAME>
git checkout BRANCHNAME
```

So in this case it would be:

```
git fetch origin pull/5403/head:add-classic-script-template
git checkout add-classic-script-template
```

But again, I typed <space> when I did this, so it should be good.

joa-quim:
OK, thanks. No big deal either (and easy to fix in case they were there). But can't look more at this now. Need to go out.

joa-quim:
Regarding the PR, I have no particular opinion. Asking classic prints this
```
gmt --new-script-classic
@echo off
REM Date:    2021-06-28T20:48:53
REM User:    j
REM Purpose: Purpose of this script
```
which does not seem a particular useful bit of info.

src/gmt.c
PaulWessel: Hi @anbj, while it may not be used a lot I am OK with accepting your contribution.  However, I'd like to see a few changes:

1. Split the new-script usage message into two separate lines, one for classic and the original one, as is done for other classic versions of options.
2. Use less spaces: Only 1 after the longest of all entries, which after splitting into two lines will remain the show-modules-core I think. As per #5341, we want to keep things tighter.
3. You need to add an entry in the gmt.rst file as well for documentation.

anbj:
I'm closing this. It was fun to do, but with 1-2 people using it, it's not worth the code change.

PaulWessel:
Sorry we could not be more welcoming on this one... But see #5341 if you want to practice more C!

anbj:
No worries! I'll check out #5341!

PR #5404, opened 2021-06-28
Clean up clear usage

KristofKoch:

Changed formatting of usage message as per #5341. Unsure if larger indent jumps are accepted. Here from `2` to `-4` for better readability of tabulated info. Guidance welcome.

### DIFF
--- a/src/clear.c
+++ b/src/clear.c
@@ -44,19 +44,23 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Message (API, GMT_TIME_NONE, "  REQUIRED ARGUMENTS:\n");
 	GMT_Usage (API, 1, "\nDeletes the specified item.  Choose one of these targets:");
 	GMT_Usage (API, 2, "cache");
-	GMT_Usage (API, -4, "Deletes the user\'s cache directory [%s].", API->GMT->session.CACHEDIR);
+	GMT_Usage (API, -4, "Deletes the user\'s cache directory:");
+	GMT_Usage (API, -4, "[%s]", API->GMT->session.CACHEDIR);
 	GMT_Usage (API, 2, "data");
-	GMT_Usage (API, -4, "Deletes the user\'s data download directory [%s/server].", API->GMT->session.USERDIR);
+	GMT_Usage (API, -4, "Deletes the user\'s data download directory:");
+	GMT_Usage (API, -4, "[%s/server]", API->GMT->session.USERDIR);
 	GMT_Usage (API, -4, "Append =<planet> to limit removal to such data for a specific <planet> [all].");
 	GMT_Usage (API, 2, "geography");
-	GMT_Usage (API, -4, "Deletes the user\'s geography directory (gshhg, dcw) [%s/geography].", API->GMT->session.USERDIR);
+	GMT_Usage (API, -4, "Deletes the user\'s geography directory (gshhg, dcw):");
+	GMT_Usage (API, -4, "[%s/geography]", API->GMT->session.USERDIR);
 	GMT_Usage (API, -4, "Append =<name> to limit removal to such data for gshhg or dcw only [all].");
 	GMT_Usage (API, 2, "sessions");
-	GMT_Usage (API, -4, "Deletes the user\'s sessions directory [%s].", API->session_dir);
+	GMT_Usage (API, -4, "Deletes the user\'s sessions directory:");
+	GMT_Usage (API, -4, "[%s]", API->session_dir);
 	GMT_Usage (API, 2, "settings");
 	GMT_Usage (API, -4, "Deletes a modern mode session\'s %s file.", GMT_SETTINGS_FILE);
 	GMT_Usage (API, 2, "all");
-	GMT_Usage (API, -4, "All of the above.\n");
+	GMT_Usage (API, -4, "All of the above.");
 	GMT_Message (API, GMT_TIME_NONE, "\n  OPTIONAL ARGUMENTS:\n");
 	GMT_Option (API, "V,;");
 
--- a/src/clear.c
+++ b/src/clear.c
@@ -37,7 +37,7 @@
 static int usage (struct GMTAPI_CTRL *API, int level) {
 	const char *name = gmt_show_name_and_purpose (API, THIS_MODULE_LIB, THIS_MODULE_CLASSIC_NAME, THIS_MODULE_PURPOSE);
 	if (level == GMT_MODULE_PURPOSE) return (GMT_NOERROR);
-	GMT_Message (API, GMT_TIME_NONE, "usage: %s all|cache|data[=<planet>]|geography[=<name>]|sessions|settings [%s]\n\n", name, GMT_V_OPT);
+	GMT_Usage (API, 0, "usage: %s all|cache|data[=<planet>]|geography[=<name>]|sessions|settings [%s]\n\n", name, GMT_V_OPT);
 
 	if (level == GMT_SYNOPSIS) return (GMT_MODULE_SYNOPSIS);
 
--- a/src/clear.c
+++ b/src/clear.c
@@ -41,15 +41,22 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 
 	if (level == GMT_SYNOPSIS) return (GMT_MODULE_SYNOPSIS);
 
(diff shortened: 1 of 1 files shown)


src/clear.c
```diff
@@ -37,19 +37,26 @@
...
-	GMT_Message (API, GMT_TIME_NONE, "\t             Append =<planet> to limit removal to such data for a specific <planet> [all].\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   geography Deletes the user\'s geography directory (gshhg, dcw) [%s/geography].\n", API->GMT->session.USERDIR);
-	GMT_Message (API, GMT_TIME_NONE, "\t             Append =<name> to limit removal to such data for gshhg or dcw only [all].\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   sessions  Deletes the user\'s sessions directory [%s].\n", API->session_dir);
-	GMT_Message (API, GMT_TIME_NONE, "\t   settings  Deletes a modern mode session\'s %s file.\n", GMT_SETTINGS_FILE);
-	GMT_Message (API, GMT_TIME_NONE, "\t   all       All of the above.\n");
+	GMT_Message (API, GMT_TIME_NONE, "  REQUIRED ARGUMENTS:\n");
+	GMT_Usage (API, 1, "\nDeletes the specified item.  Choose one of these targets:");
+	GMT_Usage (API, 2, "cache");
+	GMT_Usage (API, -4, "Deletes the user\'s cache directory [%s].", API->GMT->session.CACHEDIR);
```
meghanrjones, line 47:
```suggestion
	GMT_Usage (API, -3, "Deletes the user\'s cache directory:");
	GMT_Usage (API, -3, "[%s]", API->GMT->session.CACHEDIR);
```
For these, I think it looks better to have the path on the next line to minimize the chances of the path being split with a '...' which could be confusing. There's a few GMT_Usage calls in clear that print the path where this change could be made.

meghanrjones:
@PaulWessel, the usage message has data|server|... as required whereas the html docs list these as optional. Should the rst file be updated to agree with the usage message?

PaulWessel:
Sorry, I do not understand.  I am looking at usage and docs and cannot see data|server?

PaulWessel:
OK, I see.  Well the code will barf if nothing is given so giving a target is required.

src/clear.c
```diff
@@ -37,19 +37,30 @@
...
-	GMT_Message (API, GMT_TIME_NONE, "\t   geography Deletes the user\'s geography directory (gshhg, dcw) [%s/geography].\n", API->GMT->session.USERDIR);
-	GMT_Message (API, GMT_TIME_NONE, "\t             Append =<name> to limit removal to such data for gshhg or dcw only [all].\n");
-	GMT_Message (API, GMT_TIME_NONE, "\t   sessions  Deletes the user\'s sessions directory [%s].\n", API->session_dir);
-	GMT_Message (API, GMT_TIME_NONE, "\t   settings  Deletes a modern mode session\'s %s file.\n", GMT_SETTINGS_FILE);
-	GMT_Message (API, GMT_TIME_NONE, "\t   all       All of the above.\n");
+	GMT_Message (API, GMT_TIME_NONE, "  REQUIRED ARGUMENTS:\n");
+	GMT_Usage (API, 1, "\nDeletes the specified item.  Choose one of these targets:");
+	GMT_Usage (API, 2, "cache");
+	GMT_Usage (API, -4, "Deletes the user\'s cache directory:");
+	GMT_Usage (API, -4, "[%s]", API->GMT->session.CACHEDIR);
```
meghanrjones, line 48:
```suggestion
	GMT_Usage (API, -3, "Deletes the user\'s cache directory:");
	GMT_Usage (API, -3, "[%s]", API->GMT->session.CACHEDIR);
```

src/clear.c
```diff
@@ -37,19 +37,30 @@
...
+	GMT_Usage (API, 2, "data");
+	GMT_Usage (API, -4, "Deletes the user\'s data download directory:");
+	GMT_Usage (API, -4, "[%s/server]", API->GMT->session.USERDIR);
+	GMT_Usage (API, -4, "Append =<planet> to limit removal to such data for a specific <planet> [all].");
```
meghanrjones, line 52:
```suggestion
	GMT_Usage (API, -3, "Deletes the user\'s data download directory:");
	GMT_Usage (API, -3, "[%s/server]", API->GMT->session.USERDIR);
	GMT_Usage (API, -3, "Append =<planet> to limit removal to such data for a specific <planet> [all].");
```

src/clear.c
```diff
@@ -37,19 +37,30 @@
...
+	GMT_Usage (API, 2, "geography");
+	GMT_Usage (API, -4, "Deletes the user\'s geography directory (gshhg, dcw):");
+	GMT_Usage (API, -4, "[%s/geography]", API->GMT->session.USERDIR);
+	GMT_Usage (API, -4, "Append =<name> to limit removal to such data for gshhg or dcw only [all].");
```
meghanrjones, line 56:
```suggestion
	GMT_Usage (API, -3, "Deletes the user\'s geography directory (gshhg, dcw):");
	GMT_Usage (API, -3, "[%s/geography]", API->GMT->session.USERDIR);
	GMT_Usage (API, -3, "Append =<name> to limit removal to such data for gshhg or dcw only [all].");
```

src/clear.c
```diff
@@ -37,19 +37,30 @@
...
+	GMT_Usage (API, 2, "sessions");
+	GMT_Usage (API, -4, "Deletes the user\'s sessions directory:");
+	GMT_Usage (API, -4, "[%s]", API->session_dir);
```
meghanrjones, line 59:
```suggestion
	GMT_Usage (API, -3, "Deletes the user\'s sessions directory:");
	GMT_Usage (API, -3, "[%s]", API->session_dir);
```

src/clear.c
```diff
@@ -37,19 +37,30 @@
...
+	GMT_Usage (API, 2, "settings");
+	GMT_Usage (API, -4, "Deletes a modern mode session\'s %s file.", GMT_SETTINGS_FILE);
```
meghanrjones, line 61:
```suggestion
	GMT_Usage (API, -3, "Deletes a modern mode session\'s %s file.", GMT_SETTINGS_FILE);
```

src/clear.c
```diff
@@ -37,19 +37,30 @@
...
+	GMT_Usage (API, 2, "all");
+	GMT_Usage (API, -4, "All of the above.");
```
meghanrjones, line 63:
```suggestion
	GMT_Usage (API, -3, "All of the above.");
```

meghanrjones (approved):
Great! I think only one additional indentation level is needed for the descriptions, but am OK with you ignoring these suggestions if you think it is better as is.

KristofKoch:
Hi @meghanrjones I did this for better readability. From a hierarchical point of view you are totally right. One level more would be enough for the logic behind it. That's why I asked if logically correct indentation is mandatory. 

Let me illustrate my point with an example (from `gmt info` but similar problem):

When doing @PaulWessel -style indentations for single character arguments where there is no new line for the description I run into ugly line-breaks when using logical indentation levels. Used here is
```
1 -F
2 i:
2 d:
2 t:
```
Resulting in 
```
  -F Return various counts of tables, segments, headers,
     and records, depending on mode:
     i: One record with the number of tables,
       segments, data records, headers, and overall
       records.
     d: Dataset: One record per segment with tbl_no,
       seg_no, nrows, start_rec, stop_rec.
     t: Tables: Same as D but the counts resets per
       table.
```
When skimming the displayed text it is easy to overlook `i` as it is nestled beneath the text block and the alingment of the wrapped lines appears to be one char off.

When using 
```
1 -F
3 i:
3 d:
3 t:
```
It results (at least for me) in a more pleasing layout for the price of a non-logical indentation. 
```
  -F Return various counts of tables, segments, headers,
     and records, depending on mode:
       i: One record with the number of tables,
          segments, data records, headers, and overall
          records.
       d: Dataset: One record per segment with tbl_no,
          seg_no, nrows, start_rec, stop_rec.
       t: Tables: Same as D but the counts resets per
          table.
```
What do you prefer? How do we tread indentation levels? Do we stick to their logical order or may we deviate from the order for purely cosmetical purposes? Maybe @PaulWessel has a thought on this as well?

PaulWessel:
I am OK with the way you did it @KristofKoch.  If we find things that are too deeply nested then perhaps we need to change how we lay out that explanation.

PaulWessel (approved):

PR #5412, opened 2021-06-29
Improve dimfilter usage messages

PaulWessel:
Per #5341. Also fixed **-I** usage indents.


KristofKoch (approved):
Looks good! Both dimfilter usage message and `-I` formatting.

PR #5427, opened 2021-07-03
Skip NaN-records in seis and geodesy due to unrecognized header records

PaulWessel:
See this [report](https://github.com/GenericMappingTools/pygmt/issues/1367#issuecomment-873455274) for background.  The old read-loop in the **meca** supplement codes (now **seis** with **psvelo** in **geodesy**) did not have any protection against non-GMT header records which will give x and y as NaN.  This PR adds that check to avoid junk plotted at BL corner.

### DIFF
--- a/src/geodesy/psvelo.c
+++ b/src/geodesy/psvelo.c
@@ -1117,6 +1117,9 @@ EXTERN_MSC int GMT_psvelo (void *V_API, int mode, void *args) {
 			Return (API->error);
 		}
 
+		if (gmt_M_is_dnan (In->data[GMT_X]) || gmt_M_is_dnan (In->data[GMT_Y]))	/* Probably a non-recognized header since we got NaNs */
+			continue;
+
 		in = In->data;
 		station_name = In->text;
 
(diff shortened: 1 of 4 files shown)


seisman (approved):

PaulWessel:
Not out of the woods just yet.  I decided that it would be better if GMT i/o could pick up these as headers without the required leading `#` sign and added some code to catch cases where we found text and all "columns" were read as NaNs - that is clearly text.  All tests passed with that, except _pssac_geo.sh_.  The reason has to do with the file station.list which violates the GMT directive of numerical values followed by trailing text, since it has a text in the first column and lon, lat next to it; it should be` lon lat name`.

While I need to do some more work on my checking (since this file should not really say we found 3 Nans only 1), I wonder if the station.list can be changed (and hence the script - take out the -i) or if this is some old convention from elsewhere that we cannot change?

The plan is that once this work generally across GMT we can revert those module-specific fixes in seis and geodesy since those records would be correctly flagged as header records and then normally skipped

PaulWessel:
OK, so I am relearning that the station.list file has a certain format <filename> [<X> <Y> [<pen>]].  Not sure how we wish to handle this for psxy then, since the input file actually starts with trailing text.

PaulWessel:
Solution was to only do the hard check with no **-i**.  A PR will be coming soon.

Issue #5416, opened 2021-06-29
Decorated lines cannot go into legend


joa-quim:
Playing with automatic legend generation in classic and can plot fronts, annotated but not decorated lines.
The legend man says

> Three plot symbols may take special modifiers: front (f), quoted line (q) and vector (v).

but is this lack of support for decorated lines on purpose?

```
julia> lines(x, sin.(x), ls="line&Circ", legend=(label="sin(x)"),show=1, Vd=1)
        psxy  -JX14c/9.5c -Baf -BWSen -S~d0.88:+sc0.11+gwhite+p0.75,black -R-0.2/6.4/-1.1/1.1 -W0.75 -P -K > C:\TMP\GMTjl_tmp.ps
leg = ["S - ~d0.88:+sc0.11+gwhite+p0.75,black 0.75 - 0.75 - sin(x)", #undef, #undef]
psxy [ERROR]: Segment header tries to switch from -Sq to another symbol (~d0.88:+sc0.11+gwhite+p0.75,black) - ignored
```

PaulWessel:
Seems like it says switching from quotes line to decorated line.  Not sure why it thinks you are doing quoted lines first?

joa-quim:
And it also thinks its reading from a segment header. Which is not so dumb since its receiving a dataset

joa-quim:
It seems to be on purpose
```
else if (S.symbol == GMT_SYMBOL_QUOTED_LINE || S.symbol == GMT_SYMBOL_DECORATED_LINE || S.symbol == GMT_SYMBOL_FRONT)
	GMT_Report (API, GMT_MSG_ERROR, "Segment header tries to switch from -S%c to another symbol (%s) - ignored\n", S.symbol, s_args);
```

joa-quim:
The second time it passes at line 1088 of psxy ``S.symbol`` becomes = 107 (``q``). ``options`` have ``S`` pointing to a custom symbol in ``KC:\\TMP\GMT_symbol12068.def``
Can't follow the program's logic any further from this.

PaulWessel:
Sorry, busy all day and maybe tomorrow too.  Is this reproducible from the command line? If so could you please append that data set to go with the command to make it simpler to debug.

joa-quim:
```
[...]

PR #5428, opened 2021-07-04
Let pslegend handle decorated lines too

PaulWessel:
See #5416 for background.  Basically, I lumped quoted and decorated lines together but they need to go into separate jobs.  Now it works as expected.

Note: It is important that users use **-S~n** for this which is fine when running **legend** on a legend file created by hand.  But if we use **-l** then it is entirely possible that both **-Sq** and **-S~** use some different choices that may not be suitable for the legend (for instance, the spacing may be too big for the stated line in the legend).  Perhaps a solution would be to take whatever **-Sq -S~** argument is passed and replace anything between -**S~** and the colon with **n**1?  I.e., we only place one decorated symbol at the center of the line, and the same for quoted lines.  Thoughts on this, @joa-quim ?

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -18585,6 +18585,18 @@ void gmt_add_legend_item (struct GMTAPI_CTRL *API, struct GMT_SYMBOL *S, bool do
 		if (do_line && pen == NULL) pen = &(API->GMT->current.setting.map_default_pen);	/* Must have pen to draw line */
 		if (S->size_x > 0.0 && S->size_y > 0.0 && S->symbol == PSL_RECT)
 			fprintf (fp, "S - %c %gi,%gi %s %s - %s\n", S->symbol, size, S->size_y, (do_fill) ? gmtlib_putfill (API->GMT, fill) : "-", (do_line) ? gmt_putpen (API->GMT, pen) : "-", label);
+		else if (S->symbol == 'q') {	/* Quoted line [Experimental] */
+			char scode[GMT_LEN64] = {""};
+			sprintf (scode, "qn1:+ltext");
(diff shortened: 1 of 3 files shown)


joa-quim:
Yes, forcing **n1** in legends was my idea too. Otherwise the legend most of times will miss the annotation or decoration.

PaulWessel:
Ok , after Van Gogh exhibit

-- 
Dept. of Earth Sciences
SOEST, U of Hawaii at Manoa

PaulWessel:
With the latest updates I can do this:

`gmt math -T0/2pi/100+n T SIN = | gmt plot t.txt -B -S~d1:+sc0.2+gred+p0.75,black -W1.5p -llabel -png t`

There are still some complications I think:

While for decorated lines we place the symbol, what do we do for quoted lines where the labels are often read from data files or computed based on things? We don't have access to that, so not sure what makes sense to plot.

PaulWessel:
I think this PR is just for decorated lines which is a bit simpler than quoted lines since we know the symbol.  See if you are OK with this PR and if so approve.

joa-quim (approved):

joa-quim:
With this branch this plots just a plain line in legend

```
julia> lines(x, sin.(x), ls="Line&Circ", legend="sin(x)", show=1, Vd=1)
        psxy  -JX14c/9.5c -Ba -BWS -S~d0.88:+sc0.11+gwhite+p0.75,black -R-0.2/6.4/-1.1/1.1 -W0.75 -P -K > C:\TMP\GMTjl_tmp.ps
        S - ~d0.88:+sc0.11+gwhite+p0.75,black 0.75 - 0.75 - sin(x)
```
but I guess it's up to me to to catch and change ``~d0.88`` to ``~d1`` in the legend call.

joa-quim (approved):

PaulWessel:
Yes, I think for manually crafted legend files we should not intervene and change things.

PR #5429, opened 2021-07-04
Detect headers

PaulWessel:

Let GMT detect if input record has just text and no **-i** has been set.  In that case we treat this non-GMT header as a regular header instead of badly parsing all columns into NaN.  A follow up to #5427.


seisman (approved):

PaulWessel:
Thanks, I will leave the seis check for NaN as is as they cannot possibly hurt.

PR #5434, opened 2021-07-05
Improve the gmtconvert usage message

PaulWessel:
As per issue #5341.

### DIFF
--- a/src/gmtconvert.c
+++ b/src/gmtconvert.c
@@ -133,64 +133,79 @@ static void Free_Ctrl (struct GMT_CTRL *GMT, struct GMTCONVERT_CTRL *C) {	/* Dea
 static int usage (struct GMTAPI_CTRL *API, int level) {
 	const char *name = gmt_show_name_and_purpose (API, THIS_MODULE_LIB, THIS_MODULE_CLASSIC_NAME, THIS_MODULE_PURPOSE);
 	if (level == GMT_MODULE_PURPOSE) return (GMT_NOERROR);
-	GMT_Message (API, GMT_TIME_NONE, "usage: %s [<table>] [-A] [-C[+l<min>][+u<max>][+i]] [-D[<template>[+o<orig>]]] [-E[f|l|m|M<stride>]] [-F%s] [-I[tsr]]\n", name, GMT_SEGMENTIZE4);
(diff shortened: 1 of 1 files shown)


meghanrjones (approved):

seisman:
**-W** is not correctly formatted.
```
  -V[q|e|w|t|i|c|d]
     Change the verbosity level (currently w). Choose among 7 levels; each level adds more detailed messages:
       q: Quiet, not even fatal error messages.
       e: Error messages only.
       w: Warnings [Default].
       t: Timings (time-intensive operations only).
       i: Informational messages (or just -V).
       c: Compatibility warnings.
       d: Debugging messages.
	-W Convert trailing text to numbers, if possible.  Append +n to suppress NaN columns.
```

seisman:
Actually, there are more indentation problems for other options:
```
  OPTIONAL ARGUMENTS:

  <table> is one or more data files (in ASCII, binary, netCDF). If no files are given, standard input is read.
	-A Paste files horizontally, not concatenate vertically [Default].
	   All files must have the same number of segments and rows,
	   but they may differ in their number of columns.
	-C Only output segments whose number of records matches criteria:
	     +l<min> Segment must have at least <min> records [0].
	     +u<max> Segment must have at most <max> records [inf].
	     +i will invert the test.
	-D Write individual segments to separate files [Default writes one
	   multisegment file to stdout].  Append file name template which MUST
	   contain a C-style format for an integer (e.g., %d) that represents
	   a sequential segment number across all tables (if more than one table).
	   [Default uses gmtconvert_segment_%d.txt (or .bin for binary)].
	   Use +o<orig> to start numbering at <orig> [0].
	   Alternatively, supply a template with two long formats and we will
	   replace them with the table number and table segment numbers.
	   Use +o<t_orig>/<s_orig> to start numbering at <t_orig> for tables and <s_orig> for segments [0/0].
	-E Extract first and last point per segment only [Output all points].
	   Append f for first only or l for last only.
	   Append m<stride> to pass every <stride> records; use -EM to include the last point.
```

meghanrjones:
What OS/shell are you using? Indentation looks OK for me on macOS with bash shell, building the master branch.

seisman:
Just built the latest master branch, now it works well. Sorry for the noise.

Issue #5344, opened 2021-06-17
-Bs is incorrectly interpreted?


seisman:
**Description of the problem**

The following command uses `-Bnesw` to plot 4 axes:
```
gmt basemap -R0/40/0/40 -JX4c/4c -Bnesw -Baf -png map
```
![image](https://user-images.githubusercontent.com/3974108/122440878-fd6a0980-cf6a-11eb-94d6-ee2b060e293c.png)

Use `-Bn` gives:

![image](https://user-images.githubusercontent.com/3974108/122440917-06f37180-cf6b-11eb-977a-6f91e0183f8f.png)

Use `-Bs` gives:

![image](https://user-images.githubusercontent.com/3974108/122441043-27233080-cf6b-11eb-854b-4505cf179c11.png)

GMT may incorrectly interpret `s` as the secondary axis/annotation/label.

**System information**

* Operating system: macOS
* GMT version (`gmt --version`): `6.3.0_17131b2_2021.06.15`

PaulWessel:
Hm, I note it works as expected if you place -Bs as the last -B option.  I think the order is important and I think I am actually sorting these options somewhere to ensure that but perhaps it is failing in this case.  It is true the -Bs is problematic given its multiple meanings...

PaulWessel:
See gmt_parse.c function
`GMT_LOCAL void gmtparse_ensure_b_options_order (struct GMT_CTRL *GMT, struct GMT_OPTION *options);`
The fun thing about that function is that it sets up priorities for the various -B option as to how they should be sorted.
Then it does nothing...

PaulWessel:
I am finally tackling the sorting of the options array - seems to work.  Doing testing now.  This means the user order will no longer matter as we do an internal shuffle.

PR #5441, opened 2021-07-06
Sort -B options under the hood to ensure proper order for parsing

PaulWessel:

The parsing of **-B** works with fewest complications if the intervals are parsed first, then axes settings, then the frame settings.  However, the user should not have to worry about this.  This PR checks the **-B** options and if there are more than one and the order is not the preferred one we sort the options to place -B in the correct order at the end.

With this PR issues such as #5344 works.  I am also fixing a few scripts to use the proper syntax.  Closing #5344.

This is also potentially useful for GMT.jl where the order of the **-B** options generated would no longer need to be handled on the Julia side, @joa-quim.


joa-quim (approved):
Yes, handy but lots of pain already invested on this.

PR #5443, opened 2021-07-06
Update the usage messages for gmt defaults, get, and set

PaulWessel:
Follows the rules of #5341.

### DIFF
--- a/src/gmtset.c
+++ b/src/gmtset.c
@@ -65,7 +65,7 @@ static void Free_Ctrl (struct GMT_CTRL *GMT, struct GMTSET_CTRL *C) {	/* Dealloc
 static int usage (struct GMTAPI_CTRL *API, int level) {
 	const char *name = gmt_show_name_and_purpose (API, THIS_MODULE_LIB, THIS_MODULE_CLASSIC_NAME, THIS_MODULE_PURPOSE);
 	if (level == GMT_MODULE_PURPOSE) return (GMT_NOERROR);
-	GMT_Usage (API, 0, "usage: %s [-C | -D[s|u] | -G<defaultsfile>] [-[" GMT_SHORTHAND_OPTIONS "]<value>] PARAMETER1 value1 PARAMETER2 value2 PARAMETER3 value3 ...\n", name);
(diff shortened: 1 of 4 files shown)


src/gmtset.c
```diff
@@ -65,25 +65,29 @@ static void Free_Ctrl (struct GMT_CTRL *GMT, struct GMTSET_CTRL *C) {	/* Dealloc
...
+	GMT_Usage (API, 3, "s: Use the SI version of defaults.");
+	GMT_Usage (API, 3, "u: Use the US version of defaults.");
+	GMT_Usage (API, 1, "\n-G<defaultsfile>");
+	GMT_Usage (API, -2, "Set name of specific %s file to modify "
+		"[Default looks for file in current directory.  If not found, "
+		"it looks in the home directory, if not found it uses GMT defaults.]", GMT_SETTINGS_FILE);
+	GMT_Usage (API, -2, "Note: Only settings that differ from the GMT SI system defaults are written "
+		"to the file %s in the current directory (under classic mode) "
+		"or in the current session directory (under modern mode).", GMT_SETTINGS_FILE);
+	GMT_Usage (API, 1, "\n-%s", GMT_SHORTHAND_OPTIONS);
```
meghanrjones, line 87: Shorthand options should be -B[JRXYp]<value> rather than -BJRXYp

meghanrjones (approved):

src/gmtset.c
PaulWessel: Why is that?  GMT_SHORTHAND_OPTIONS cannot be used in that way.

meghanrjones: Sorry typo. The syntax and online docs have -[BJRXYp]<value>

PaulWessel: OK, I see.  I am guessing this is a bit lazy and should be -[B|J|R|X|Y|p]<value> ?

meghanrjones: > OK, I see. I am guessing this is a bit lazy and should be -[B|J|R|X|Y|p] ?

Yes

PR #5457, opened 2021-07-07
Update grdconvert usage message

meghanrjones:

Following https://github.com/GenericMappingTools/gmt/issues/5341

**Reminders**


PaulWessel (approved):
The entry for ingrid under REQUIRES ARGUMENTS has a leading hyphen that should be removed. Not part of your edits so must have been there from before.

PR #5458, opened 2021-07-07
Update grdgdal usage message

meghanrjones:

**Reminders**


PaulWessel (approved):
Question for -A: Is the brackets correct here? I.e., both +m and +c must be given or neither?

meghanrjones:
> Question for -A: Is the brackets correct here? I.e., both +m and +c must be given or neither?

From the gdal docs, it seems `+c<cpt_name>` is needed when `+m<method>` is color-relief (unless there's a default cpt, not clear on this). So I suppose it should be `-A<prog>[+m<method>[+c<cpt_name>]]`.

PaulWessel:
Yes, I agree. @joa-quim ?

joa-quim:
Yes, trusting in docs ...

PR #5459, opened 2021-07-07
Update grdinterpolate usage message

meghanrjones:

Follows https://github.com/GenericMappingTools/gmt/issues/5341

**Reminders**


PaulWessel (approved):
For -E: While not spelled out in the guidelines, I prefer both the modifier and directive lists to start a sentence.  So "+g uses gridline ..." should be "+g Use gridline ...".  Also, if the modifier takes an argument I do not list it at the start, so "+a<az> defines a ..." should be "+a Defines a ...".
For -T: This is a near-common option; see -T in gmt math for how I did the modifiers and perhaps do the same layout?

PR #5460, opened 2021-07-07
Update grdlandmask usage message

meghanrjones:

Follows https://github.com/GenericMappingTools/gmt/issues/5341

**Reminders**

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -44,6 +44,7 @@
  *	gmtlib_setparameter		Sets a default value given keyword,value-pair\n
  *	gmtlib_getparameter
  *	gmt_GSHHG_syntax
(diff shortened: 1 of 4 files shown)


PaulWessel (commented):
See gmtselect for how to set -D and -N so it is the same.
Re -D: Surely we have to issue the same info in pscoast, grdlandmask, and grdmath, so I suggest we make a gmt_gshhg_resolution () function instead that is called by all three instead of doing this three times.

PaulWessel:
Would you want me to make the gmt_gshhs-resolution function or will you try that?

meghanrjones:
> Would you want me to make the gmt_gshhs-resolution function or will you try that?

I would like to try this.

PaulWessel (approved):

PR #5467, opened 2021-07-07
Update grdview usage messages

PaulWessel:
As per #5341.

### DIFF
--- a/src/grdview.c
+++ b/src/grdview.c
@@ -418,7 +418,7 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	const char *name = gmt_show_name_and_purpose (API, THIS_MODULE_LIB, THIS_MODULE_CLASSIC_NAME, THIS_MODULE_PURPOSE);
 	if (level == GMT_MODULE_PURPOSE) return (GMT_NOERROR);
 	GMT_Usage (API, 0, "usage: %s <topogrid> %s [%s] [-C[<cpt>]] [-G<drapegrid> | <image> | -G<grd_r> -G<grd_g> -G<grd_b>] "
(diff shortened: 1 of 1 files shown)


src/grdview.c
```diff
@@ -417,65 +417,73 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
...
+	GMT_Usage (API, -2, "To force a monochrome image using the YIQ transformation, append +m.");
 	GMT_Option (API, "R");
-	GMT_Message (API, GMT_TIME_NONE, "\t-S Smooth contours first (see grdview for <smooth> value info) [no smoothing].\n");
-	gmt_pen_syntax (API->GMT, 'T', NULL, "Image the data without interpolation by painting polygonal tiles.\n"
-	                "\t   Append +s to skip tiles for nodes with z = NaN [Default paints all tiles].\n"
-	                "\t   Append +o[<pen>] to draw tile outline [Default uses no outline].", 0);
-	GMT_Message (API, GMT_TIME_NONE, "\t   Cannot be used with -Jz|Z as it produces a flat image.\n");
+	GMT_Usage (API, 1, "\n-S<smooth>");
+	GMT_Usage (API, -2, "Smooth contours first (see grdcontour for <smooth> value info) [no smoothing].");
+	gmt_pen_syntax (API->GMT, 'T', NULL, "Image the data without interpolation by painting polygonal tiles.", 0);
```
meghanrjones, line 472:
```suggestion
	gmt_pen_syntax (API->GMT, 'T', NULL, "Image the data without interpolation by painting polygonal tiles in the form [+o[<pen>]][+s].", 0);
```
 It would be best to include +o and +s options in the first line, rather than just the pen syntax, but this is a simpler solution.

meghanrjones (approved):

PR #5487, opened 2021-07-17
Update seis supplement usage messages

### LINKED ISSUE #5486: Two ways to set annotation attributes in pscoupe
The two modules **meca** and **coupe** share many settings, such as **-S** and **-F**.  However, in pursuing a revision via #5341 I noticed that both documentation and code for **coupe** let users specify the same label attributes (font, justification, angle) via modifiers to either **-S** or **-F**, whereas **meca** only lets this happen via **-S**.  It is messy to have duplication of both parsing and documentation in two separate places, and only in **coupe**.

I propose we make **coupe** similar to **meca** and only let **-S** handle the label attributes.  What says @seisman ?

PaulWessel:
As per #5341.  Also fix that word-breaking at hyphens require something before the hyphen.  Also, per #5486  we remove the font-specifications for the label from **coupe -F** since it is also done via **-S** as in **meca**.  The code is left unchanged for backwards compatibility.  Closes #5486.


seisman:
> Also, per #5486 we remove the font-specifications for the label from **coupe -F** since it is also done via **-S** as in **meca**. The code is left unchanged for backwards compatibility. Closes #5486.

Are you going to make more changes to **coupe** in another PR? Currently, using **-Fs** still says it expects 4 input columns.

PaulWessel:
I combined those tasks.  However, I assume you are referring to this:

**-Fs**\ *symbol*\ [*size*]
    Select a symbol instead of mechanism. Choose from the following:
    (**c**) circle, (**d**) diamond, (**i**) itriangle, (**s**) square,
    (**t**) triangle, (**x**) cross. *size* is the symbol size in
    :term:`PROJ_LENGTH_UNIT` (unless **c**, **i**, or **p** is appended
    to indicate that the size information is in units of cm, inches, meters,
    or points, respectively). If *size* must be read, it must be in column 4
    and the text string will start in column 5.

I am not sure what it should say.   Could you make a suggestion?

PaulWessel:
Sorry, perhaps you mean this as well:

   Parameters are expected to be in the following columns:

    **1**,\ **2**:
      longitude, latitude of event (**-:** option interchanges order)
    **3**:
      depth of event in kilometers
    **4**:
      Text string to appear near the beach ball

seisman:
> Parameters are expected to be in the following columns:
> 
> ```
> **1**,\ **2**:
>   longitude, latitude of event (**-:** option interchanges order)
> **3**:
>   depth of event in kilometers
> **4**:
>   Text string to appear near the beach ball
> ```

I tried the example in #5486, but provided input data with only 4 columns, it still works. I don't understand why it works with 4-column data.

PaulWessel:
Note that I did not change what the code does, just the usage and docs part.  So functionality is the same, even if wrong...

seisman:
> Note that I did not change what the code does, just the usage and docs part. So functionality is the same, even if wrong...

OK. I'll find some time to understand how **coupe** works. The changes to the **coupe -F** makes sense to me.

I didn't follow how the new `GMT_Usage` function works, so perhaps @meghanrjones can review this PR?

PaulWessel:
Yes, @meghanrjones can review the style changes since we have been doing that without bugging you and joaquim.

meghanrjones (approved):

PR #5490, opened 2021-07-17
Update usage message for img2grd

PaulWessel:
As per #5341.  Fix usage order of arguments.


meghanrjones (approved):
Please confirm that -T should be under required, since there is a default listed.

PR #5491, opened 2021-07-17
Update nearneighbor usage messages

PaulWessel:
As per #5341.  Also fixed **-n** to only give the shorter version with modifiers suitable for this module.


meghanrjones (approved):
Two updates needed based on online docs:
1. Move -N to optional
2. Add -g common option

PaulWessel:
Thanks, it was worse: C code listed g and s but neither can be used here, and RST listed g but not s.  No g/s, and -N now optional in usage.

PaulWessel:
Also needed -a (which a few others should also get, like surface...).

PR #5509, opened 2021-07-20
Turn off implicit output in externals if spectrum1d -T is given

PaulWessel:
**spectrum1d** has a long history, initially writing individual files for each cross-spectra component.  Then we added the capability to write a single table with all the cross-spectra in separate columns to _stdout_ by default, but this could be turned _off_ via **-T**.  Because of backwards compatibility we left the default action as it was, hence **-T** is needed to turn off a call to _GMT_Write_Data_.  Thus, _GMT_Encode_Options_ needs to check for **-T** and if set it should deactivate the implicit output argument.

Note: I think it is very rare if any externals uses **spectrum1d** and does _not_ want the result back as normal output, but in case they do and use **-T** we do not want a nasty crash.  In fact, I would recommend that external wrapping of **spectrum1d** by default pass **-N** to _bypass_ the individual-written files (which is outside the control of the API) and receive the output as a single multi-column table like from similar modules writing to _stdout_.

This PR thus adds the missing check in _GMT_Encode_Options_. @joa-quim should review it, and @seisman should be aware of the Note.

### DIFF
--- a/doc/rst/source/sph2grd.rst
+++ b/doc/rst/source/sph2grd.rst
@@ -18,7 +18,7 @@ Synopsis
 [ |-D|\ [**g**\|\ **n**] ]
 [ |-E| ]
 [ |-F|\ [**k**]\ *filter* ]
-[ |-N|\ [*norm*] ]
+[ |-N|\ **g**\|\ **m**\|\ **s** ]
 [ |-Q| ]
 [ |SYN_OPT-V| ]
 [ |SYN_OPT-bi| ]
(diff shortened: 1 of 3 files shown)


PaulWessel:
Sorry, I managed to add sph2grd usage updates to this PR.  @meghanrjones, perhaps you can review the sph2grd part and then @joa-quim can do the gmt_api.c part - or you can do both since there is nothing special here.

meghanrjones (approved):

PR #5510, opened 2021-07-20
Revised subplot usage messages

PaulWessel:

There are there types of changes in this PR, hence the large number of files:

1. Update to the subplot usage message per #5341, plus other improvements.
2. Upgrading options **-SC**|**R** to **-Sc**|**r** to follow GMT convention elsewhere.  This was done in a backwards compatible way.  The old uppercase directives are still honored quietly.
3. Revisions to numerous scripts and 8 PS files that changed due to the **-S** update.


meghanrjones (approved):
-B is missing from subplot begin.
One of the tests should keep the old -SC|R syntax as a regression check.

PR #5515, opened 2021-07-21
Update grdmask usage statement

meghanrjones:

**Reminders**


PaulWessel (approved):
Looks good.  However, elsewhere I have replaced lists like a), b) or 1. 2. with the GMT_LINE_BULLET see pstext for one place.  I think we should do tat, so either director:, +modifier, or bullet in lists.

PR #5525, opened 2021-07-23
Let coupe detect range of depths as default y-values

### LINKED ISSUE #5516: coupe -A fails unless +z is given
While listed as an optional modifier, +z is in fact required as otherwise we get bad values and a crash. We should fix this to determine the min/max depth of the data as default.

PaulWessel:
The **-A+z** modifier is listed as optional but without it we crash.  This rewrite reads in the entire data set first instead of record-by-record, and if no **+z** we determine the z-range from the data.  Closes #5516.

### DIFF
--- a/src/seis/pscoupe.c
+++ b/src/seis/pscoupe.c
@@ -501,7 +501,7 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Usage (API, 3, "+f Set font attributes for the label [%s].", gmt_putfont (API->GMT, &font));
 	GMT_Usage (API, 3, "+j Set the label <justification> [TC].");
 	GMT_Usage (API, 3, "+o Set the label offset <dx>[/<dy>] [0/0].");
-	GMT_Usage (API, -2, "Note: If fontsize < 0 then no label written; offset is from the limit of the beach ball.");
+	GMT_Usage (API, -2, "Note: If fontsize = 0 (+f0) then no label written; offset is from the limit of the beach ball.");
 	GMT_Message (API, GMT_TIME_NONE, "\n  OPTIONAL ARGUMENTS:\n");
 	GMT_Option (API, "B-");
 	GMT_Usage (API, 1, "\n-C<cpt>");
@@ -950,7 +950,7 @@ static int parse (struct GMT_CTRL *GMT, struct PSCOUPE_CTRL *Ctrl, struct GMT_OP
 
 	n_errors += gmt_M_check_condition (GMT, !Ctrl->A.active, "Must specify -A option\n");
 	//n_errors += gmt_M_check_condition (GMT, Ctrl->F.active && Ctrl->S.active, "Cannot specify both -F and -S\n");
-	n_errors += gmt_M_check_condition (GMT, !GMT->common.R.active[RSET], "Must specify -R option\n");
+	n_errors += gmt_M_check_condition (GMT, !GMT->common.R.active[RSET] && !Ctrl->A.frame, "Must specify -R option or -A...+r modifier\n");
 	n_errors += gmt_M_check_condition (GMT, !Ctrl->S.active, "Must specify -S option\n");
 	n_errors += gmt_M_check_condition (GMT, Ctrl->S.active && Ctrl->S.scale < 0.0, "Option -S: must specify scale\n");
 
@@ -977,10 +977,12 @@ EXTERN_MSC int GMT_pscoupe (void *V_API, int mode, void *args) {
 	int i, transparence_old = 0, not_defined = 0;
 	int n_scanned = 0;
 	unsigned int xcol = 0, scol = 0, icol = 0, tcol_f = 0, tcol_s = 0;
+	uint64_t tbl, seg, row, col;
+
 	FILE *pnew = NULL, *pext = NULL;
 
 	double size, xy[2], xynew[2] = {0.0}, plot_x, plot_y, n_dep, distance, fault, depth;
-	double scale, P_x, P_y, T_x, T_y, nominal_size, *in = NULL;
+	double scale, P_x, P_y, T_x, T_y, nominal_size, in[GMT_LEN16];
 
 	char event_title[GMT_BUFSIZ] = {""}, Xstring[GMT_BUFSIZ] = {""}, Ystring[GMT_BUFSIZ] = {""};
 
@@ -989,8 +991,9 @@ EXTERN_MSC int GMT_pscoupe (void *V_API, int mode, void *args) {
 	struct M_TENSOR mt, mtr;
 	struct AXIS T, N, P, Tr, Nr, Pr;
 	struct GMT_PALETTE *CPT = NULL;
-	struct GMT_RECORD *In = NULL;
 	struct GMT_PEN current_pen;
+	struct GMT_DATASET *D = NULL;	/* Pointer to GMT multisegment input tables */
+	struct GMT_DATASEGMENT *S = NULL;
 
 	struct PSCOUPE_CTRL *Ctrl = NULL;
 	struct GMT_CTRL *GMT = NULL, *GMT_cpy = NULL;		/* General GMT internal parameters */
@@ -1017,6 +1020,7 @@ EXTERN_MSC int GMT_pscoupe (void *V_API, int mode, void *args) {
 
 	gmt_M_memset (&meca, 1, meca);
 	gmt_M_memset (&moment, 1, moment);
+	gmt_M_memset (in, GMT_LEN16, double);
 	nominal_size = Ctrl->S.scale;
 
 	if (Ctrl->C.active) {
@@ -1028,26 +1032,6 @@ EXTERN_MSC int GMT_pscoupe (void *V_API, int mode, void *args) {
 		gmt_illuminate (GMT, Ctrl->I.value, Ctrl->G.fill.rgb);
 		Ctrl->I.active = false;	/* So we don't do this again */
 	}
-
-	if (Ctrl->A.frame) {
-		GMT->common.R.wesn[XLO] = 0.0;
-		GMT->common.R.wesn[XHI] = Ctrl->A.p_length;
-		GMT->common.R.wesn[YLO] = Ctrl->A.dmin;
-		GMT->common.R.wesn[YHI] = Ctrl->A.dmax;
-		if (gmt_M_is_zero (Ctrl->A.PREF.dip)) Ctrl->A.PREF.dip = 1.0;
-	}
-
-	if (gmt_map_setup (GMT, GMT->common.R.wesn)) Return (GMT_PROJECTION_ERROR);
-
-	if ((PSL = gmt_plotinit (GMT, options)) == NULL) Return (GMT_RUNTIME_ERROR);
-	gmt_plane_perspective (GMT, GMT->current.proj.z_project.view_plane, GMT->current.proj.z_level);
-	gmt_set_basemap_orders (GMT, Ctrl->N.active ? GMT_BASEMAP_FRAME_BEFORE : GMT_BASEMAP_FRAME_AFTER, GMT_BASEMAP_GRID_BEFORE, GMT_BASEMAP_ANNOT_BEFORE);
-	gmt_plotcanvas (GMT);	/* Fill canvas if requested */
-	gmt_map_basemap (GMT);
-
-	PSL_setfont (PSL, GMT->current.setting.font_annot[GMT_PRIMARY].id);
-
-	if (!Ctrl->N.active) gmt_map_clip_on (GMT, GMT->session.no_rgb, 3);
 
 	if (Ctrl->S.read) {	/* Read symbol size from file */
 		scol = Ctrl->S.n_cols;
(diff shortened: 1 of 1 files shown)


PaulWessel:
Pinging @seisman.

seisman:
I tried this script: https://github.com/GenericMappingTools/2021-unavco-course/blob/main/seismology/beachball-cross-section.sh.

This branch gives me this plot:

image

The shallowest and deepest beachballs are cropped because the automatically determined depth range is too tight. Can it be more clever about the depth?

Esteban82:
In the animation 14 I also get the beachballs cropped. I have to add 10 (km; trough gmt math) to avoid it for the deepest one. The shallowest are cropped.

PaulWessel:
Yes, I noticed that too and thought about remedies.  Perfection would require us to add/subtract magnitude * scale / height to each depth and keep the extremes instead of just getting depth, since you may have an event that is not the shallowest but its size is larger so it would be clipped.  Something like that, no?

seisman:
> Perfection would require us to add/subtract magnitude * scale / height to each depth and keep the extremes instead of just getting depth, since you may have an event that is not the shallowest but its size is larger so it would be clipped.

Sounds reasonable to me.

Esteban82:
For me too. Although the text above of the shallowest beachballs will be cropped, right?

PaulWessel:
Let's reset this discussion a bit - I worry about setting a bad precedence by treating **coupe** in a special way.  In **plot** and other places one can leave **-R** off (as in **coupe**) and we determine an exact fit to the data.  If you are plotting any symbol then half of the most extreme symbols will be clipped.  In **coupe** there is also the label so this would get out of hand. GMT also has **-Re**[exact] and **-Ra**[uto] options as well.  If **-Ra** is used then we round the exact region outward.  Getting involved in determining symbol sizes and using that to change **-R** is a bit too much and cannot easily be doing in **plot**.

So my suggestions is that **+z** could take arguments **e**xact and **a**uto, with perhaps **e** being default.  With **+z**e we do exactly as this PR implements, while with **+za** we would employ the same rounding-outward algorithm that **-Ra** uses (_gmtinit_round_wesn_).  My only question is if **+za** should also affect the distance (x) rounding or not.  I think it should.

seisman:
> So my suggestions is that **+z** could take arguments **e**xact and **a**uto, with perhaps **e** being default. With **+z**e we do exactly as this PR implements, while with **+za** we would employ the same rounding-outward algorithm that **-Ra** uses (_gmtinit_round_wesn_). My only question is if **+za** should also affect the distance (x) rounding or not. I think it should.

I agree, but instead of using **+ze** or **+za**, perhaps it makes more sense to use **+ra** and **+re**.

meghanrjones:
> Let's reset this discussion a bit - I worry about setting a bad precedence by treating **coupe** in a special way. In **plot** and other places one can leave **-R** off (as in **coupe**) and we determine an exact fit to the data. If you are plotting any symbol then half of the most extreme symbols will be clipped. In **coupe** there is also the label so this would get out of hand. GMT also has **-Re**[exact] and **-Ra**[uto] options as well. If **-Ra** is used then we round the exact region outward. Getting involved in determining symbol sizes and using that to change **-R** is a bit too much and cannot easily be doing in **plot**.
> 
> So my suggestions is that **+z** could take arguments **e**xact and **a**uto, with perhaps **e** being default. With **+z**e we do exactly as this PR implements, while with **+za** we would employ the same rounding-outward algorithm that **-Ra** uses (_gmtinit_round_wesn_). My only question is if **+za** should also affect the distance (x) rounding or not. I think it should.

Are **-Ra** and **-Re** documented? I had not heard about these options before.

PaulWessel:
They are well documented in my head, but apparently have not made it into the docs yet...
Modern mode only.  So I guess needs to go into several places.

PaulWessel:
Note: I will add the docs on -Ra|e now in a new PR

PaulWessel:
Update: Those who wish this to be perfect and include symbol size and label: Note that unless **-M** is used it requires the full calculation of moment and magnitude to scale the symbol size properly.  That is just too much work to implement for this minor adjustment.  What we could do is this:

1. I can do the quick/dirty z-adjustment for a Mag 5 symbol size given in **-S** and the optional label.  This gives an adjusted raw z-range that is pretty close but you may have some symbols of larger magnitude that happen to be at the bottom/top that will be only partially dealt with.
2. **+r** could take a few more arguments: **a** for auto will do the auto roundoff, i.e., the same as **-Ra.**, while **+re** would be exact [Default].  We could use **+rA** for the same as **+ra** except we let _zmin_ be forced to zero (surface of Earth) to avoid ranges like 5/400, for instance.
3.  +r could take optional _dx_ which would round the x-distances to a multiple of _dx_ [Default is raw distance range].
4. We could allow **+z**_dz_ which would round the determined range to a multiple of _dz_. This would override the auto for z if **+ra** is given, for instance, but **+rA** would still set _zmin_ to 0.

Thoughts?

PaulWessel:
I have finalized this PR.  I made a few changes to my plan above as I worked through it.  The auto-range handling symbol size and labels work pretty well, and exactly for **-M** (without **-M** I base my calculation on a max magnitude 8 event).  The **-A** option has been upgraded:

-**Aa**|**b**|**c**|**d**_params_[**+d**_dip_][**+r**[**a**|**e**|_dx_]][**+w**_width_][**+z**[**s**]**a**|**e**|_dz_|_min_/_max_]

I think this should cover all the bases.  The default as before sets the exact domain which now includes the size of the symbols.  You can then explore the auto rounding or fixed rounding or clamping.  Hopefully, @seisman can give it a test (see doc for explanations).  Here is the previous example:

beachball-cross-section

Esteban82:
According to the new features, what would happen if I have a maximum value of 300 and I use +za or +z50. If I understood correctly, I would get the same 300, am I right? 

So I wonder. Is there any way to extend the range of values by a certain percentage (e.g. 5%)? That is, if my data had a range from 100 to 200, then the range I get for the graph is 95-205. In this way I would get in the chart a small margin without data on the edges and this would be regardless of the values of my data. I hope this is understood.

PaulWessel:
I think +z50 would give you 300, but +za should extend it further.  Try it out.  Remember if you max depth actually is 300 then the max z I compute will be 300 + symbol-radius * inverse-scale which is > 300.  It would be great if you could play with this a bit and see how it performs.

Esteban82:
I played a bit with the branch (I think, I am not completed sure if I installed well). In the previous example I try +z50 and I get this figure with NO error messages:

beachball-cross-section

It there a bug?

```
gmt begin beachball-cross-section png
	gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+z50+r -Q \
		-JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
gmt end show
```

Esteban82:
With -Za and -Ze I get this error message:

Error: /undefinedresult in --charpath--
Operand stack:
   PSL_LH   (M)   true
Execution stack:
   %interp_exit   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--   false   1   %stopped_push   1990   1   3   %oparray_pop   1989   1   3   %oparray_pop   1977   1   3   %oparray_pop   1833   1   3   %oparray_pop   --nostringval--   %errorexec_pop   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--
Dictionary stack:
   --dict:739/1123(ro)(G)--   --dict:0/20(G)--   --dict:75/200(L)--   --dict:163/250(L)--
Current allocation mode is local
Current file position is 28030
psconvert [ERROR]: System call [gs -q -dNOSAFER -dNOPAUSE -dBATCH -sDEVICE=bbox -DPSL_no_pagefill -dMaxBitmap=2147483647 -dUseFastColor=true '/home/federico/.gmt/sessions/gmt_session.71080/gmt_0.ps-' 2> '/home/federico/.gmt/sessions/gmt_session.71080/psconvert_71083c.bb'] returned error 256.

PaulWessel:
No, something must have happen on your end.  Bad gmt.conf? Nothing in coupe would affect basemap.

PaulWessel:
Try this script: It shows 12 different permutations of **+r+z.**  All look good for me.  I could not make a 4x3 subplot since coupe cannot go in a subplot as first command since there really is no -R for subplot to work with... I will make an issue of this - either we document or we find a solution.

```
#!/usr/bin/env bash
#
# Plot cross-section of beachballs
#
gmt begin
    gmt figure coupe_01 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+r -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+r" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_02 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+ze+r -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+ze+r" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_03 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+za+r -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+za+r" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_04 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+zs50+r -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+zs50+r" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_05 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+ra -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+ra" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_06 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+ze+ra -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+ze+ra" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_07 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+zsa+ra -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+zsa+ra" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_08 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+z50+ra -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+z50+ra" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_09 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+r25 -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+r25" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_10 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+ze+r25 -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+ze+r25" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_11 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+za+r25 -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+za+r25" | gmt text -F+cTR+f12p -Dj0.2c
    gmt figure coupe_12 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+z50+r25 -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -BWSen
    echo "+z50+r25" | gmt text -F+cTR+f12p -Dj0.2c
gmt end show
```

PaulWessel:
Note: I added modifier **+c**[**n**|**t**] which will do no plotting but simply report the region determined via **-A** either as 4 numbers _xmin xmax ymin ymax_ or as trailing text **-R**_xmin/xmax/ymin/ymax_.

Esteban82:
I played a bit with Paul script. With the deepest symbols worked perfectly. For the shallowest, in some instances they were cropped like in the following figure (shallowest symbols range from 5 to 20). 
coupe_03

However, then I try it with real data (japan-beachballs.sh) and a smaller size for the symbols (-Sd0.35c) is often used so I get a good figure: 
```

	gmt coupe japan.dat -Sd0.35c -Aa126/42/146/40+d90+w500+za+r -Q -C \
		-JX15c/-10c -BWSen -Bxaf+l"Distance (km)" -Byaf+l"Depth (km)" -i0-9
```


japan-beachballs-sideview

In conclusion, I think that in practice it would we worked fine.

PaulWessel:
Thanks @Esteban82, could you post the script and data for the first plot with the clipped shallow balls? I am curious as to why +za is not working there.

Esteban82:
Sure Paul. I use this data
```

112 32 5  30  90   0  4  Strike-slip
113.5 34 10  30  60  90  5  Reverse
115 34 15 30  60  90  5  Reverse
117 32 700  30  60 -90  6  Normal
117 32 20  30  60 -90  6  Normal
```
and this script:
```
gmt begin
    gmt figure coupe_03 png
    gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+za+r -Q -JX15c/-8c -Bxaf+l"Distance (km)" -Bya50f10+l"Depth (km)" -BWSen
    echo "+za+r" | gmt text -F+cTR+f12p -Dj0.2c
gmt end
```

PaulWessel:
Thanks, I think it is because of my code doing this:

`GMT->common.R.wesn[YLO] = MAX (0, GMT->common.R.wesn[YLO] - dz);	/* But do not go negative into the air... */`

My thought was that one would not want depth to go negative.  But I can see if you have shallow quakes on a large range then perhaps we must, so I will take that off.  We still have the **+zs**[_arg_] modifier to clamp at zero so we dont need to above limitation.

PaulWessel:
Yes, that was it.  Update and try again and you will see.

Esteban82:
Test it. Now it works perfect. 

coupe_03

PaulWessel:
Pinging @seisman to see if he has time for a review so we can get this one done?

seisman:
I'm trying this script: [https://github.com/GenericMappingTools/2021-unavco-course/blob/main/seismology/beachball-cross-section.sh](https://github.com/GenericMappingTools/2021-unavco-course/blob/main/seismology/beachball-cross-section.sh?rgh-link-date=2021-07-27T03%3A15%3A08Z).

1. `-Aa111/33/119/33` without `-R`: should report an error, but it crashes
2. `-Aa111/33/119/33 -R0/800/0/50`: should work, but crashes
3. `-Aa111/33/119/33` with the modifier `+ra` and `+re` affect both x and y ranges, but modifier `+r20` only affects the X range. I'm OK with the behavior, but it should be better documented if this is the expected behavior.
4. `-Aa111/33/119/33+r+z20`: the plotted depth range is about -5 to 65 km, but I expect it to be 0 to 60 km. Is it because beachball sizes are also considered in this case?

PaulWessel:
I will check and report back.

PaulWessel:
Thanks @seisman I fixed the two issues.

seisman:
Points 1 and 2 are fixed now. What about my point 4?

PaulWessel:
Will address after a zoom with my student...

PaulWessel:
> Points 1 and 2 are fixed now. What about my point 4?

Would it be OK to simply state in the documentation somewhere that when automatic depth range is selected then we adjust the range for the size of the selected symbols so that they are _not clipped?_  I assumed when someone wanted an automatic range they do not want things clipped - i.e., what @Esteban82 brought up initially.  You can always impose a hard depth range with either a specific range or turn on **+zs** to never go above the surface.

seisman (approved):

PR #5529, opened 2021-07-24
Update usage message for xyz2grd

meghanrjones:

**Reminders**


PaulWessel (approved):
Sorry, I guess I prefer the items under -Z to be like this:
```
A: ASCII (multipel...)
a: ASCII (one value)

Also, I think for modules where -J is barely used we should use "J-" in GMT_Option to keep output shorter.
```
since they are all 1-char directives.

PR #5549, opened 2021-07-27
Add sorting of --PAR=value as well

### LINKED ISSUE #5475: PS_PAGE_COLOR in unset by modern theme
This does not set the page color, but removing the ``--GMT_THEME`` does
```
printf "0 0\n10 10" | psxy -R0/10/0/10 -JX14c/9.5c -Bpx+lx -Baf -By+lx@+2@+ -BWSen+gwhite --PS_PAGE_COLOR=#D0DFE6FF -W1 -P -K --GMT_THEME=modern > lixo.ps
```

PaulWessel:
This is to ensure **--GMT_THEME** is processed before any other --PAR=value settings since it sets a bunch of theme and could override user settings if applied later.  Closes #5475.

@joa-quim, please build this branch and try your case to make sure it is solid before approval.

### DIFF
--- a/src/gmt_parse.c
+++ b/src/gmt_parse.c
@@ -472,10 +472,10 @@ GMT_LOCAL int gmtparse_compare_B (const void *p1, const void *p2) {
 }
 
 /*! */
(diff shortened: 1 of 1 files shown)


joa-quim (approved):
Cool, even the Julia example works now.

PR #5562, opened 2021-07-28
Flip order of S and -V in usage and synopsis

PaulWessel:
The -V was listed before -S


meghanrjones (approved):

PaulWessel:
FYI, I also fixed a dozen or more similar and even more minor things throughout, via direct commit to master (...).

Issue #5582, opened 2021-08-02
Adding plot margin via psconvertoptions fails in modern mode


PaulWessel:
This example shows the problem:

```
#!/bin/bash
gmt begin
	gmt set PS_MEDIA 4ix4i
	gmt figure map1 jpg A+m0.2i
	echo 0 0 | gmt plot -R-2/2/-2/2 -JX4i -SC1i -Gred -B0 -X0 -Y0 --MAP_FRAME_PEN=faint
	gmt figure map2 eps
	echo 0 0 | gmt plot -R-2/2/-2/2 -JX4i -SC1i -Gred -B0 -X0 -Y0 --MAP_FRAME_PEN=faint
gmt end
gmt psconvert map2.eps -Tj -A+m0.2i -P
file map1.jpg
file map2.jpg
```

Here, map1.jpg does not have the 0.2i margin added despite giving A+m0.2i to the **figure** command, whereas map2.jpg gets it by running **psconvert** separately.  So the problem must lie somewhere in how the **figure** argument is eventually passed to **psconvert** via gmt end.  It is the same with gmt **begin**.

PS. This continues my CostCo saga for metal prints of large maps: After receiving a flawed print and talking to CostCo support I learned that they have an undocumented "bleed" of 0.125" on all sides, hence if you wish to defeat their auto-cropping you must add the bleed up front, so that a 36x24 inch image at 300 dpi actually should be submitted as 10875x7275 pixels (with 75 pixels, nicely divisible by 2 (not) is cropped).  Hence, I would like to add **A+m**_bleed_ to my script but alas that fails.

PaulWessel:
Perhaps this is a better description of the problem.  The current **-A** has a mix of crop-related and non-crop-related modifiers.  A backwards-compatible change that would clarify matters might introduce a new option **-N** which takes the non-crop-related modifiers:

1. **-A**[**+r**][**+u**]
2. **-N**[**+f**_fade_][**+g**_fill_][**+m**_margins_][**+p**[_pen_]][**+s**[**m**]|**S**_width_[/_height_]]

This solution eliminates the awkward **-A+n** where **+n** negates the crop implied by **-A** in the first place...
This clean separation of cropping in the **-A** means that **-A** controls whether or not we crop to BB or use the media size, and then **-N** controls any expansion of that box, scaling, and fill/outline of box.

[...]

PR #5583, opened 2021-08-03
Let gmt end better handle -A args

PaulWessel:
See #5582 for background.  This minor change allows the margins and other **-A** modifiers such as **+g**, **+f**, etc.

### DIFF
--- a/doc/rst/source/psconvert.rst
+++ b/doc/rst/source/psconvert.rst
@@ -133,14 +133,14 @@ Optional Arguments
     Append **+m** to specify extra margins to extend the bounding box.
     Give either one (uniform), two (x and y) or four (individual sides)
     margins; append unit [Default is set by :term:`PROJ_LENGTH_UNIT`].
-    Use the **-I+s**\ *width* to resize the output image to exactly *width* units.
+    Append **+s**\ *width* to resize the output image to exactly *width* units.
     The default unit is set by :term:`PROJ_LENGTH_UNIT` but you can append a new
     unit and/or impose different width and height (**Note**: This may change the
     image aspect ratio). What happens here is that Ghostscript will do the re-interpolation
-    work and the final image will retain the DPI resolution set by **-E**.  Use **-I+sm**
+    work and the final image will retain the DPI resolution set by **-E**.  Append **+sm**
     to set a maximum size and the new *width* is only imposed if the original figure width
     exceeds it. Append /\ *height* to also impose a maximum height in addition to the width.
-    Alternatively, use **-I+S**\ *scale* to scale the image by a constant factor.
+    Alternatively, append **+S**\ *scale* to scale the image by a constant factor.
 
 .. _-L:
 
@@ -229,7 +229,7 @@ Optional Arguments
     Together with **-V** it prints on screen the gdal_translate
     (gdal_translate is a command line tool from the GDAL package)
     command that reads the raster + world file and creates a true
-    geotiff file. Use **-W+g** to do a system call to gdal_translate
+    geotiff file. Append **+g** to do a system call to gdal_translate
     and create a geoTIFF image right away. The output file will have a
     .tiff extension.
 
@@ -238,7 +238,7 @@ Optional Arguments
     .tfw, for jpeg we have a .jgw and so on. **Note**: This option automatically
     sets **-A** **-P**.  Append **+c** to *not* crop the image.
 
-    Use **-W+k** to create a minimalist KML file that allows loading the
+    Append **+k** to create a minimalist KML file that allows loading the
     image in GoogleEarth. Note that for this option to work it is necessary that the postscript
     image must have been created with **-JX** or **-Jx** Cartesian projection of
     geographical coordinates. If not, a warning is issued but the KML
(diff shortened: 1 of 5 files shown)


PaulWessel:
The key problem is that **-A** was initially created to do cropping, but then I had lots of other features I wanted added, such as controlling paper color, outlines, margins, scaling, etc, and they all got added to **-A**.  However, none of those modifers are really about cropping, so we added **+n** to say no crop as well.  It should be possible to use **+m** to add margins to a cropped image and an uncropped image.  This is what is failing right now.  I wonder if it may be best to let **-A** with no args turn on cropping (finding the BB) while a **-A** with arguments does not turn on cropping, and the two can be combined when we want cropping and margin, for example.  Wonder if @joa-quim is around to opine on this?

PaulWessel:
I have completed the upgrade to **psconvert** (and gmt end as well) to handle the new options:

- Crop settings: **-A**[**+r**][**+u**]
- Scaling/extending settings: **-I**[**+m**_margins_][**+s**[**m**]_width_[/_height_]][**+S**_scale_]
- Color/pen settings: **-N**[**+f**_fade_][**+g**_fill_][**+i**][**+p**[_pen_]]

Tests pass and my map that wants to add 0.125i bleed (margin) to the fixed media now works as well.  I few notes of things done during the updates:

1. Parsing of **-A** anticipates old syntax and sorts modifiers into potential **-A**, **-I**, **-N** sections which are then parsed by the respective parsers (if nonzero args).
2. The old **-I** for using ICC profiles has been reborn as **-N+i** but no-arg **-I** is also parsed under backwards rules.
3. A dozen or so _GMT_Report_ calls had been passed _Ctrl_ instead of _API_ as first argument, now fixed.
4. Modifier order now alphabetical and synopsis and documentations have been updated.
5. The old **-A+n** which was added to turn off cropping when **-W** is used (which automatically turns on **-A -P**) is still honored under backwards rules but has been replaced by a new **+c** modifier to **-W** instead (where it belongs).


The only thing I have not tested is any potential issues when **-W** is used.  Hoping @joa-quim can check for compliance before we remove the WIP.

joa-quim:
Many changes to eye follow. 
**-N+g** <==> to ``--PS_PAGE_COLOR``, no?

Maybe come out with long-options name for **-N** & **-I**

PaulWessel:
> Many changes to eye follow.
> **-N+g** <==> to `--PS_PAGE_COLOR`, no?
> 
> Maybe come out with long-options name for **-N** & **-I**

We have had -A+gfill forever and not sure if pS_PAGE_COLOR=rgb works the same way when you extend the region.  I will try.

PaulWessel:
Yes, you can do --PS_PAGE_COLOR or you can do -N+g.  I think we added it because it was simpler.

meghanrjones:
gmt begin *options* documentation needs updating too, since the new **I** and **N** options should be available.

As a GMT user, I would find it helpful to have these deprecations included in the changelog. What do you think of having an additional deprecation section? From a maintenance perspective, it would just mean having another label 'deprecation' and using the 'add-changelog' label for PRs such as this, as well as using descriptive PR titles as per usual. We could adjust the release-drafter workflow to do the rest.

PaulWessel:
I think that would go useful, too.  I suspect most GMT users are resistant to changing how they use GMT and only notice if there is a deprecation if they get a compatibility warning.  But they wont get those after we changed the verbosity levels to place compatibility further down the list and beyond the default.

PaulWessel:
> Maybe come out with long-options name for **-N** & **-I**

As well as for -A.  Perhaps

-A:  --crop[+timestamp][+round]
-I: --extend[+margins=args][+size=arg][+scale=arg]
-N: --canvas[+fill=arg][+pen=arg][+icc][+fade=arg]

meghanrjones:
Not sure if I am missing something from the documentation but I cannot get -N+f*fade* to do anything:
```
gmt grdimage @AK_gulf_grav.nc -Jl142W/55N/18/24/1.5c -Bafg > test.ps
gmt psconvert -Tj -A -I+m2c -N+f50 test.ps
open test.jpg
```
Also, a new notice shows up from gmt end in this PR.

PaulWessel:
I think it is only meant to be used via movie (fade in fade out). But I will check.

PaulWessel:
I remove the temporary NOTICE message.

meghanrjones (approved):

joa-quim:
> As well as for -A.

Sure, but ``--crop`` was obvious.

PaulWessel:
@joa-quim any issues with -W that you can detect? I dont have any good examples to try.  As mentioned, the only potential issue would be related to -A+n now being -W+c.  You may recall that using -W automatically sets -A -P and we added -A+n to turn -A off, basically, which is an odd way of doing things.  The -W+c is better in that sense.  But not tested.

joa-quim:
Try the Patagonia example in the ``psconvert`` man page.
Some Windows update screw the compiler and can't currently build.

PaulWessel:
Will do.

PaulWessel:
Patagonia example worked, I could import that geotiff in Google Earth Pro and it obliquely laid down and fit the background.  So I will merged shortly.

PR #5592, opened 2021-08-04
Minor revison of long-option items

PaulWessel:
After reviewing the current _gmt_common_longoptions.h_ table, Meghan and I made a few changes:

1. Only use underscores to separate _twowords_ that are hard to read otherwise (e.g., splitting two vowels)
2. Changed the single modifier name starting with uppercase to be lowercase
3. Renamed the long modifiers for **+L**, **+s**, **+S** in **-B** axis
4. Improved the **-j** words. --distance is better but still looking for a "family" name for spherical, ellipsoidal, flatearth.  @WalterHFSmith ?

### DIFF
--- a/src/gmt_common_longoptions.h
+++ b/src/gmt_common_longoptions.h
@@ -8,8 +8,8 @@
 	 * The items below correspond to the named parameters in the GMT_KEYWORD_DICTIONARY structure:
 	 */
 	/* separator, short_option, long_option, short_directives, long_directives, short_modifiers, long_modifiers */
-	{   0, 'B', "frame",         "",        "",                                         "b,g,n,o,t,s",				"box,fill,noframe,obliquepole,title,subtitle" },
-	{   0, 'B', "axis",          "x,y,z",   "x,y,z",                                    "a,f,l,L,p,s,S,u",			"angle,fancy,label,Label,prefix,second_label,Second_label,unit" },
+	{   0, 'B', "frame",         "",        "",                                         "b,g,n,o,t,s",				"box,fill,noframe,pole,title,subtitle" },
+	{   0, 'B', "axis",          "x,y,z",   "x,y,z",                                    "a,f,l,L,p,s,S,u",			"angle,fancy,label,hlabel,prefix,altlabel,alt_hlabel,unit" },
 	{   0, 'J', "projection",    "",        "",                                         "",         				""},
 	{   0, 'R', "region",        "",        "",                                         "r,u",        				"rectangular,unit"},
 	{   0, 'U', "timestamp",     "",        "",                                         "c,j,o",    				"command,justify,offset"},
@@ -17,7 +17,7 @@
 	{   0, 'X', "xshift",        "a,c,f,r", "absolute,center,fixed,relative",           "",         				""},
 	{   0, 'Y', "yshift",        "a,c,f,r", "absolute,center,fixed,relative",           "",         				""},
 	{   0, 'a', "aspatial",      "",        "",                                         "",         				""},
-	{   0, 'b', "binary",        "",        "",                                         "b,l",      				"bigendian,littleendian"},
+	{   0, 'b', "binary",        "",        "",                                         "b,l",      				"bigendian,little_endian"},
 	{   0, 'c', "panel",         "",        "",                                         "",      				    ""},
 	{   0, 'd', "nodata",        "i,o",     "in,out",                                   "",         				""},
 	{   0, 'e', "find",          "",        "",                                         "f",        				"file"},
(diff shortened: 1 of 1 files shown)


src/gmt_common_longoptions.h
```diff
@@ -8,26 +8,26 @@
...
 	{   0, 'g', "gap",           "",        "",                                         "n,p",        				"negative,positive"},
 	{   0, 'h', "header",        "i,o",     "in,out",                                   "c,d,r,t", 					"columns,delete,remark,title"},
 	{ ',', 'i', "incols",        "",        "",                                         "l,o,s",   					"log10,offset,scale"},
-	{   0, 'j', "spherical",     "e,f,g",   "ellipsoidal,flatearth,greatcircle",        "",   			    		""},
-	{   0, 'l', "legend",        "",        "",                                         "D,G,H,L,N,S,V,f,g,j,o,p,s,w",   	"drawline,gap,header,linetext,ncols,size,vertline,font,fill,justify,offset,pen,scale,width"},
-	{   0, 'n', "interpolation", "b,c,l,n", "bspline,bicubic,linear,nearestneighbor",   "a,b,c,t",     				"antialias,bc,clip,threshold"},
+	{   0, 'j', "distance",     "e,f,g",   "ellipsoidal,flatearth,spherical",        "",   			    		""},
+	{   0, 'l', "legend",        "",        "",                                         "D,G,H,L,N,S,V,f,g,j,o,p,s,w",   "horline,gap,header,linetext,ncols,size,vertline,font,fill,justify,offset,pen,scale,width"},
```
meghanrjones, line 29:
```suggestion
	{   0, 'l', "legend",        "",        "",                                         "D,G,H,L,N,S,V,f,g,j,o,p,s,w",   "hline,gap,header,linetext,ncols,size,vertline,font,fill,justify,offset,pen,scale,width"},
```
I think hline is better here.

PaulWessel: OK, but then vertline goes to vline, no?

meghanrjones: Yes, just did not notice vertline. vline is good.

src/gmt_common_longoptions.h
```diff
@@ -8,26 +8,26 @@
 	 * The items below correspond to the named parameters in the GMT_KEYWORD_DICTIONARY structure:
 	 */
 	/* separator, short_option, long_option, short_directives, long_directives, short_modifiers, long_modifiers */
-	{   0, 'B', "frame",         "",        "",                                         "b,g,n,o,t,s",				"box,fill,noframe,obliquepole,title,subtitle" },
-	{   0, 'B', "axis",          "x,y,z",   "x,y,z",                                    "a,f,l,L,p,s,S,u",			"angle,fancy,label,Label,prefix,second_label,Second_label,unit" },
+	{   0, 'B', "frame",         "",        "",                                         "b,g,n,o,t,s",				"box,fill,noframe,pole,title,subtitle" },
+	{   0, 'B', "axis",          "x,y,z",   "x,y,z",                                    "a,f,l,L,p,s,S,u",			"angle,fancy,label,hlabel,prefix,altlabel,alt_hlabel,unit" },
```
seisman, line 12: Perhaps we should use `altlabel` + `althlabel` here, or `alt_label` + `alt_hlabel`?

src/gmt_common_longoptions.h
```diff
@@ -8,26 +8,26 @@
...
 	{   0, 'J', "projection",    "",        "",                                         "",         				""},
 	{   0, 'R', "region",        "",        "",                                         "r,u",        				"rectangular,unit"},
 	{   0, 'U', "timestamp",     "",        "",                                         "c,j,o",    				"command,justify,offset"},
 	{   0, 'V', "verbosity",     "",        "",                                         "",         				""},
 	{   0, 'X', "xshift",        "a,c,f,r", "absolute,center,fixed,relative",           "",         				""},
 	{   0, 'Y', "yshift",        "a,c,f,r", "absolute,center,fixed,relative",           "",         				""},
 	{   0, 'a', "aspatial",      "",        "",                                         "",         				""},
-	{   0, 'b', "binary",        "",        "",                                         "b,l",      				"bigendian,littleendian"},
+	{   0, 'b', "binary",        "",        "",                                         "b,l",      				"bigendian,little_endian"},
```
seisman, line 20: `big_endian` + `little_endian`?

src/gmt_common_longoptions.h
PaulWessel: OK done

PaulWessel:
Wonder if in **-q** the words _perset,perfile,persegment_ would be better as _byset,byfile,bysegment_?

meghanrjones:
> Wonder if in **-q** the words _perset,perfile,persegment_ would be better as _byset,byfile,bysegment_?

Yes to me for byset, etc.

PaulWessel:
Given the general complication with keywords that are composed of two logical words (e.g., "by segment" becoming either _bysegment_ or _by_segment_ but not _by-segment_, for instance [since these may be variables in Julia]), I think we should always try really, really hard to come up with synonyms or other roundabout phrases that reduces to a single logical word.  Surely this must be possible so we can avoid or at least minimize the two-word problem?

Looking at just the common option table, the candidates for hard thinking are
-B+s: currently alt_label.
-b: big_endian, little_endian
-j: flatearth
-q: byset, byfile, bysegment

For instance, while "Flat Earth" may tell us something, is that generally a good term for what -jf does?  Might there be a better term? Big and little endian is very tricky since we are so used to the endian which derives from Gulliver's travels...

meghanrjones (approved):

seisman:
@PaulWessel It seems you also need to update the test `test/modern/longbasemap.sh`

```
Warning: ING]: Long-modifier form WSne+fill=lightgreen+obliquepole=60/20 for option -B not recognized!
/usr/local/bin/gm compare: image difference exceeds limit (0.0506297 > 0.003).
ERROR: gmtest:167
memtrack errors: 0
exit status: 1
test/modern/longbasemap.ps: RMS Error = 0.0506 [FAIL]
memtrack errors: 0
exit status: 1
```

PR #5602, opened 2021-08-07
Consistently describe input grid name and modifiers

PaulWessel:

Of the modules that read a grid, most called it _ingrid_ and a few called in _grid_.  Except for **grdconvert**, none bothered to list the options to specify a specific grid format [_=ID_] or the modifiers to scale/offset the grid or assign a certain value to mean NaN.  This PR adds a **GMT_INGRID** constant in _gmt_synopsis.h_ that is now used in the synopses, as well as a new function _gmt_ingrid_syntax_ which isused to explain the purpose of the input grid and then list the modifier explanations.

Also, to follow the lead in **-i**, I added a new gridfile modifier **+d**_divisor_ for the times it is better to specify a divisor than a factor (e.g. +d3 instead of +s0.333333...).

No calculation changes so all tests are unaffected

We will need similar changes in the RST docs, but this is the module synopsis and usage part.


PaulWessel:
Sorry about the back/forth.  This PR is exclusively about input grids now.  I will do a separate branch for output grids.

doc/rst/source/cookbook/features.rst
```diff
@@ -2402,17 +2402,18 @@ the data may need translation and scaling prior to use. Therefore, all
 GMT programs that read or write grid files will decode the given
 filename as follows:
 
-name[=\ *ID*][**+s**\ *scale*][**+o**\ *offset*][**+n**\ *invalid*]
+name[=\ *ID*][**+d**\ *divisor*][**+n**\ *invalid*][**+o**\ *offset*][**+s**\ *scale*]
```
meghanrjones, line 2405: Also needs to be updated in explain_grd_inout.rst_

PaulWessel: Done, as well as in grdconvert.rst which has more of the same.

PaulWessel:
There is also a modifier **+b**_band_ which applies to images only (right, @joa-quim?).  This is discussed in the [docs](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#modifiers-to-read-and-write-grids-and-images-via-gdal), but not shown in any usage synopsis. However, there is nothing stopping me from doing `gmt grdmath image.png=bd+b2 10 DIV = junk.nc` so perhaps we need to add **+b** to the general ingrid syntax.  However, selecting 3 bands via +b should only be allowed when actually reading an image (limited to grdimage, grdmix, image, ???). Thoughts on that?

joa-quim:
Yeap, **-b**_band_ was thought for images but can be used for any multi-band file (I think). Do not confuse with multi subdatasets.

meghanrjones:
?*varname* is similar to +b*band* and could be considered a modifier. I think it would be good to include both so that people know they can look for more information regarding subdatasets.

PaulWessel:
Currently, I have

`#define GMT_INGRID   "<ingrid>[=<ID>][+d<divisor>][+n<invalid>][+o<offset>][+s<scale>]"`

The questions then are:

- Do we add `[+b<band>] `for grids?
- Do we add `[?<varname>] `and how? I think Remko added this so not sure if it would work with, say, =gd.  Do you know, @joa-quim? If not, then maybe `[=<ID>|?<varname>]`?

Since netCDF is the default format I agree we should add `[?<varname>]`.  I am let certain about various GDAL modifiers since they are relatively rarely used (?) and may be best described with examples in the cookbook.

PaulWessel:
I have added the `?varname `and **+b**_band_ to the ingrid synopsis and explanation - I think this one is good to go?

meghanrjones (approved):

PR #5606, opened 2021-08-08
Add v (vector) to custom symbol kit, plus handle azimuth vs angle

PaulWessel:
It may be useful to draw a simple one-headed vector as part of a custom symbol, but we had no such capability and the only way to combine a simple vector with a custom symbol was to plot things twice. This PR adds **v** as a new known symbol to the table displayed in the appendix (and it takes _x, y, angle, length_).  To keep things simple there are some hard-wired decision being made: 

1. The vector has a head at the end of the vector.
2. The head parameters are derived from the current pen thickness: Head length is 8 x the pen _width_, and the head apex is 30 degrees, while the shape is set to 0.5.
3. The head outline is always drawn, and head may be filled if **-G** is used. 

This PR also clarifies how to ensure angles are considered as Cartesian vs geographic azimuths, and allows a trailing **a** to an angular constant to mean azimuth.  Prior to this PR, it was not possible to draw an arc, wedge, or ellipse using azimuths, for instance.

All previous tests pass, but it would be helpful if @KristofKoch given his branch a try given his numerous custom symbols...


Esteban82 (approved):
I run the GMT_angle-azim.sh and I get this figure. As far as I know it is ok. 

@PaulWessel this could be used in animation02, right?

PD: Don't worry Paul. I just solve it (replace plot -Sv by -SV).

GMT_angle-azim

PaulWessel:
You could, but not sure if it buys you anything,  The annotations perhaps?  Probably worth a try to see if it looks good etc with the font size.

Esteban82:
> The annotations perhaps?

Ok, I will try it.

PR #5607, opened 2021-08-08
Fix failure when using +sa for auto grid-range scaling

PaulWessel:
When one wishes to save disk space by using an integer grid format, it is useful to have GMT automaticallly determine scales and offsets so that the precision of the data can be preserved despite being stored as integers.  However, we must have had a regression here since **+sa** or **+oa** lead to errors.  This PR fixes this problem.  For instance, this sequences creates a grid with two nodes that are -0.5 and 1.25 and we wish to store the grid using bytes (0-255).  GMT determines the optimal scale to use the full range of bytes:

```
(echo 0 0 1.25; echo 1 1 -0.5) | gmt xyz2grd -R-1/1/-1/1 -I1 -Gout.nc=nb+sa
gmt grdinfo out.nc
out.nc: Title: z
out.nc: Command: xyz2grd -R-1/1/-1/1 -I1 -Gout.nc=nb+sa
out.nc: Remark: 
out.nc: Gridline node registration used [Cartesian grid]
out.nc: Grid file format: nb = GMT netCDF format (8-bit integer), CF-1.7
out.nc: x_min: -1 x_max: 1 x_inc: 1 name: x n_columns: 3
out.nc: y_min: -1 y_max: 1 y_inc: 1 name: y n_rows: 3
out.nc: v_min: -0.501968503937 v_max: 1.25 name: z
out.nc: scale_factor: 0.00984251968504 add_offset: 0 packed z-range: [-51,127]
out.nc: format: classic
gmt grd2xyz out.nc -s
1	1	-0.501968502998
0	0	1.25
```

So despite using bytes we are still able to approximately store the two floating points via netCDF scale factors.

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -13832,7 +13832,7 @@ char *gmtlib_last_valid_file_modifier (struct GMTAPI_CTRL *API, char* filename, 
 				allow_a = false;	/* a is not a valid number */
 				switch (modifiers[k]) {	/* The modifier code */
 					case 'o': case 's':	/* +o<offset>|a,  +s<scale>|a */
-						allow_a = true;	/* A for auto is OK for these modifiers */
+						allow_a = true;	/* Argument "a" for auto is OK for these modifiers */
 						/* Fall through on purpose */
 					case 'h': case 'i': case 'n': 	/* +h[<hinge>], +i<inc>, +n<nodata> */
 						k++;	/* Move to start of argument, next modifier, or NULL */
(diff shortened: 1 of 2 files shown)


meghanrjones:
Can we add your example or something similar as a test to prevent future regressions?

PaulWessel:
Added test _xyz2grd/gridrange.sh_ which passes after using **+sa+oa**.

meghanrjones (approved):

PR #5612, opened 2021-08-09
WIP Consistently describe output grid name and modifiers

PaulWessel:
Similar to #5602 but for output grids.  It differs because on output you can use **+sa** and **+oa** as well as specify GDAL drivers and datatypes.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -7729,6 +7729,24 @@ void gmtlib_explain_options (struct GMT_CTRL *GMT, char *options) {
 	}
 }
 
+/*! Output grid specification */
+/*!
+	\param GMT ...
+	\param message ...
+*/
+void gmt_outgrid_syntax (struct GMTAPI_CTRL *API, char option, char *message) {
+	if (option == 0)
+		GMT_Usage (API, 1, "\n%s", GMT_OUTGRID);
+	else
+		GMT_Usage (API, 1, "\n-%c%s", option, GMT_OUTGRID);
+	GMT_Usage (API, -2, "%s. Optionally append =<ID> for writing a specific file format and add any modifiers:", message);
+	GMT_Usage (API, 3, "+d Divide data values by the given <divisor> [0]");
+	GMT_Usage (API, 3, "+n Replace data values matching <invalid> with a NaN.");
+	GMT_Usage (API, 3, "+o Offset data values by the given <offset>, or append a for automatic range offset [0].");
+	GMT_Usage (API, 3, "+s Scale data values by the given <scale>, or append a for automatic range scale [1].");
+	GMT_Usage (API, -2, "Note: Any offset is added after any scaling, and +sa also sets +oa (unless overridden). "
+		"To write specific formats via GDAL, use <ID> = gd and supply <driver> [and optionally <dataType>].");
+}
 /*! GSHHG subset specification */
 /*!
 	\param GMT ...
--- a/src/gmt_synopsis.h
+++ b/src/gmt_synopsis.h
@@ -32,6 +32,9 @@
 #ifndef GMT_SYNOPSIS_H
 #define GMT_SYNOPSIS_H
 
+/* Full syntax of output grids */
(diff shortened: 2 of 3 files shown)


PaulWessel:
This PR currently uses this syntax to be used in options that currently say `-G<outgrid>`:

`#define GMT_OUTGRID  "<outgrid>[=<ID>][+d<divisor>][+n<invalid>][+o<offset>|a][+s<scale>|a][:<driver>[/<dataType>]]"`

Any concerns with this one?  Note, **+b** is not used for output (I think) and `?varname` is not used either (we use **-D** to set a named netcDF variable).

src/gmt_synopsis.h
```diff
@@ -32,6 +32,9 @@
 #ifndef GMT_SYNOPSIS_H
 #define GMT_SYNOPSIS_H
 
+/* Full syntax of output grids */
+#define GMT_OUTGRID  "<outgrid>[=<ID>][+d<divisor>][+n<invalid>][+o<offset>|a][+s<scale>|a][:<driver>[/<dataType>]]"
```
meghanrjones, line 36:
```suggestion
#define GMT_OUTGRID  "<outgrid>[=<ID>][+d<divisor>][+n<invalid>][+o<offset>|a][+s<scale>|a][:<driver>[/<dataType>][+c<options>]]"
```

src/gmt_init.c
```diff
@@ -7729,6 +7729,24 @@ void gmtlib_explain_options (struct GMT_CTRL *GMT, char *options) {
...
+		GMT_Usage (API, 1, "\n%s", GMT_OUTGRID);
+	else
+		GMT_Usage (API, 1, "\n-%c%s", option, GMT_OUTGRID);
+	GMT_Usage (API, -2, "%s. Optionally append =<ID> for writing a specific file format and add any modifiers:", message);
+	GMT_Usage (API, 3, "+d Divide data values by the given <divisor> [0]");
+	GMT_Usage (API, 3, "+n Replace data values matching <invalid> with a NaN.");
+	GMT_Usage (API, 3, "+o Offset data values by the given <offset>, or append a for automatic range offset to preserve precision for integer grids [0].");
+	GMT_Usage (API, 3, "+s Scale data values by the given <scale>, or append a for automatic range scale to preserve precision for integer grids [1].");
+	GMT_Usage (API, -2, "Note: Any offset is added after any scaling, and +sa also sets +oa (unless overridden). "
+		"To write specific formats via GDAL, use <ID> = gd and supply <driver> [and optionally <dataType>].");
```
meghanrjones, line 7748:
```suggestion
		"To write specific formats via GDAL, use <ID> = gd and supply <driver> (and optionally <dataType> and/or one or more concatenated GDAL -co <options> using +c.");
```

meghanrjones (commented):
Two more potential changes to be more comprehensive.

PaulWessel:
Looks odd to not include **+c** in the list of modifiers in the usage, no?  Shall I move it in there and leave most of the explanation as a note, @meghanrjones ?

meghanrjones:
> Looks odd to not include **+c** in the list of modifiers in the usage, no? Shall I move it in there and leave most of the explanation as a note, @meghanrjones ?

The current synopsis implies that `:<driver>` is required to use **+c**. Unless this is false, the entire `:<driver>[/<dataType>][+c<options>]` would need to be included in the list of modifiers.

PaulWessel:
OK, I guess I don't know.  @joa-quim, is the **+c** only valid or meaningful if one specifies a driver first?

joa-quim:
The situation seems a bit confuse. While **+c** is only usable with ``=gd`` I don't think there should be any constrain to use it in all cases but currently we can only use it in ``grdimage -A``. However, I don't see anything that obliges specifying a  ``:<driver>`` to set **+c**.

PaulWessel:
Looks like we can add the +c in this synopsis but the code should check that if +c is given and gd is not auto-detected nor selected then there is an error.

PR #5613, opened 2021-08-09
Consolidate -A into -F for a more unified option

### LINKED ISSUE #5522: Movie. Improve use of -F
From the slack group of 2021 Unavco Course: 

> Federico Esteban: Is it -F (in movie) mandatory? I use movie with -M and without -F and I get all the frames (and they would be deleted if I used -Z).
> Paul Wessel: We may need to improve that.  I would agree that if no -F and a -M then it should be the same as -Fnone -M

PaulWessel:
While **-F** let you choose among video options and add some suitable modifiers, animated GIF was not one of them and it had its own option **-A** with modifier.  Yet, **-A+s** required **-F** to have been set, etc.  Same with **-M**, requiring **-F**_none_ if no animation desired. This PR lets **-F** specify which products to build from the list of gif,png,mp4,webm and offers modifiers.  The option is repeatable.  We thus deprecate **-A** (now **-F**_gif_) and **-F**_none_ (now default) which will be handled via standard backwards compatibility rules.  Closes #5522.

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -16,10 +16,9 @@ Synopsis
 |-C|\ *canvas*
 |-N|\ *prefix*
 |-T|\ *nframes*\|\ *min*/*max*/*inc*\ [**+n**]\|\ *timefile*\ [**+p**\ *width*]\ [**+s**\ *first*]\ [**+w**\ [*str*]\|\ **W**]
-[ |-A|\ [**+l**\ [*n*]]\ [**+s**\ *stride*] ]
 [ |-D|\ *displayrate* ]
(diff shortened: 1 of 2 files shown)


Esteban82 (approved):

Esteban82:
Should the scripts of the animations (1 to 5) be updated?

PaulWessel:
Yes, they should (of course they are working via backwards compatibility). If you have time, please make a PR with the changes.

Esteban82:
I am trying to update the scripts of the animations. Where I can find anim01.txt?? Is this the file of the script, right?

.. literalinclude:: /_verbatim/anim01.txt

PaulWessel:
anim01.txt and others like it are created on the fly by cmake so you only need to worry about the *.sh files.

PR #5617, opened 2021-08-10
Update -g synopsis, syntax, usage and docs

PaulWessel:
The **-g** syntax was a bit awkward (had a separate **-qa** to change a bool value). Now we use a modifier (**+a**) like elsewhere.  Also, we now set an alternative z-coordinate column via **+c**_col_ like elsewhere [2].  Also fixed wording and minor mistakes in the RST and synopsis.


meghanrjones:
I find the new syntax more confusing than before (in particular the relation between **z** directive and **+c** modifier). I think this would be more similar to the syntax in other common options:
**-gx|y|d|X|Y|D|z**[*col*/]*gap*[**+a**][**+n|p**]

PaulWessel:
Let's discuss this.  I see **+c**_col_ is used in several places:

Common options:
```
-qi[~]<rangevalues>,...+c<col>
-qo[~]<rangevalues>,...+c<col> 
-wy|a|w|d|h|m|s|c<period>[/<phase>][+c<col>]
```

gmtinfo:
`[-T<dz>[+c<col>]]`

gmtselect:
 `-Z<min>/<max>[+c<col>][+i][+a]`

In that context, this looks similar to me:

`-gx|y|z|d|X|Y|D<gap>[<unit>][+a][+c<col>][+n|p`

Adding the slash if a zcol is specified means more complications in doing a long-option I think.  PS. I need to read a thesis and attend a defense and then a Dr visit so will be very intermittent here today.

doc/rst/source/std-opts-classic.rst
```diff
@@ -40,7 +40,7 @@ Common Options (Classic Mode)
      - :ref:`Filter data records that match the given pattern <-e_full>`
    * - **-f**\ [**i**\|\ **o**]\ *colinfo*
      - :ref:`Set formatting of ASCII input or output <-f_full>`
-   * - **-g**\ [**a**]\ **x**\|\ **y**\|\ **d**\|\ **X**\|\ **Y**\|\ **D**\|[*col*]\ **z**\ *gap*\ [**+n**\|\ **p**]
+   * - **-g**\ **x**\|\ **y**\|\ **z**\|\ **d**\|\ **X**\|\ **Y**\|\ **D**\|\ *gap*\ [**u**][**+a**][**+c**\ *col*][**+n**\|\ **p**]
```
meghanrjones, line 43:
```suggestion
   * - **-g**\ **x**\|\ **y**\|\ **z**\|\ **d**\|\ **X**\|\ **Y**\|\ **D**\ *gap*\ [**u**][**+a**][**+c**\ *col*][**+n**\|\ **p**]
```

doc/rst/source/std-opts.rst
```diff
@@ -36,7 +36,7 @@ Common Options
      - :ref:`Filter data records that match the given pattern <-e_full>`
    * - **-f**\ [**i**\|\ **o**]\ *colinfo*
      - :ref:`Set formatting of ASCII input or output <-f_full>`
-   * - **-g**\ [**a**]\ **x**\|\ **y**\|\ **d**\|\ **X**\|\ **Y**\|\ **D**\|[*col*]\ **z**\ *gap*\ [**+n**\|\ **p**]
+   * - **-g**\ **x**\|\ **y**\|\ **z**\|\ **d**\|\ **X**\|\ **Y**\|\ **D**\|\ *gap*\ [**u**][**+a**][**+c**\ *col*][**+n**\|\ **p**]
```
meghanrjones, line 39:
```suggestion
   * - **-g**\ **x**\|\ **y**\|\ **z**\|\ **d**\|\ **X**\|\ **Y**\|\ **D**\ *gap*\ [**u**][**+a**][**+c**\ *col*][**+n**\|\ **p**]
```

meghanrjones:
> Let's discuss this. I see **+c**_col_ is used in several places:
> 
> Common options:
> 
> ```
> -qi[~]<rangevalues>,...+c<col>
> -qo[~]<rangevalues>,...+c<col> 
> -wy|a|w|d|h|m|s|c<period>[/<phase>][+c<col>]
> ```
> 
> gmtinfo:
> `[-T<dz>[+c<col>]]`
> 
> gmtselect:
> `-Z<min>/<max>[+c<col>][+i][+a]`
> 
> In that context, this looks similar to me:
> 
> `-gx|y|z|d|X|Y|D<gap>[<unit>][+a][+c<col>][+n|p`
> 
> Adding the slash if a zcol is specified means more complications in doing a long-option I think. PS. I need to read a thesis and attend a defense and then a Dr visit so will be very intermittent here today.

OK, I will instead just make a couple small wording suggestions for the new syntax.

doc/rst/source/explain_-g_full.rst_
```diff
@@ -3,25 +3,24 @@ The **-g** option
...
+A criterion is specified using one of the **x**\|\ **y**\|\ **z**\|\ **d**\|\ **X**\|\ **Y**\|\ **D** directives. The
+upper-case directives specify that the criterion should be applied to the projected coordinates for
 modules that map data to map coordinates.
 
 - **x**\|\ **X** - define a gap when there is a large enough change in the x coordinates (upper case to use
   projected coordinates).
 - **y**\|\ **Y** - define a gap when there is a large enough change in the y coordinates (upper case to use
   projected coordinates).
+- **z** - define a gap when there is a large enough change in the data in z column *col* [default *col* is 2
+  (i.e., 3rd column); use **+c**\ *col* to change this selection].
```
meghanrjones, line 21:
```suggestion
- **z** - define a gap when there is a large enough change in the z data. Use **+c**\ *col* to change
  the z data column [Default *col* is 2 (i.e., 3rd column)].
```

meghanrjones (approved):
Looks good after the two suggestions removing the extra | in the synopsis are resolved.

PR #5621, opened 2021-08-10
Fix typos in gmtselect -Z option

weiji14:

Spotted a few typos while wrapping `gmtselect` for PyGMT at #1429. Specifically a 'has' which should be 'as', some missing '+' signs for the modifiers, and adding a '[Default is ...]'.

Fixes #

**Reminders**


meghanrjones (approved):
Thanks!

PR #5623, opened 2021-08-11
Docs cleanup

PaulWessel:
For **-q**, explain that min/max has defaults if one is not given.  Also, since we return an error if more tha one ~ is given then only one is allowed (not required).
For **-e**, the **+f**_file_ is not a modifier but the way to selectpatterns a file.  I do not like that since it is not consistent with the rest of GMT - will get back to that.  
For **-w**: Replace first colon with period and use (first) rather than (x)

### DIFF
--- a/doc/rst/source/explain_-q_full.rst_
+++ b/doc/rst/source/explain_-q_full.rst_
@@ -3,7 +3,7 @@ The **-q** option
 
 **Syntax**
 
-**-q**\ [**i**\|\ **o**][~]\ *rows*\ [**+c**\ *col*][**+a**\|\ **f**\|\ **s**]
+**-q**\ [**i**\|\ **o**][~]\ *rows*\|\ *limits*\ [**+c**\ *col*][**+a**\|\ **f**\|\ **s**]
     Select specific data rows to be read and/or written.
 
 **Description**
--- a/doc/rst/source/explain_-q_full.rst_
+++ b/doc/rst/source/explain_-q_full.rst_
@@ -23,6 +23,6 @@ Alternatively, use **+c**\ *col* to indicate that the arguments instead are min/
 column *col*. With **+c**\ *col*, only rows whose data for the given column *col* are within the range(s) given by the
 *min*/*max* limits are read (with **-qi**) or written (with **-qo**). **Note**: Because arguments may contain colons or
 be negative, your must specify *start*/*stop* instead of *start*\ [:*inc*]:*stop*. If *min* or *max* is not given we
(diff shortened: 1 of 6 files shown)


doc/rst/source/explain_-q_full.rst_
```diff
@@ -22,6 +22,7 @@ modifiers to control how the rows are counted [Default is **+a**]:
 Alternatively, use **+c**\ *col* to indicate that the arguments instead are min/max *data limits* for the values in
 column *col*. With **+c**\ *col*, only rows whose data for the given column *col* are within the range(s) given by the
 *min*/*max* limits are read (with **-qi**) or written (with **-qo**). **Note**: Because arguments may contain colons or
-be negative, your must specify *start*/*stop* instead of *start*\ [:*inc*]:*stop*.
+be negative, your must specify *start*/*stop* instead of *start*\ [:*inc*]:*stop*. If *min* or *max* is not given we
```
meghanrjones, line 25:
```suggestion
be negative, you must specify *min*/*max* instead of *start*\ [:*inc*]:*stop*. If *min* or *max* is not given we
```
I think this is more descriptive, if I am interpreting the usage correctly.

doc/rst/source/explain_-q_full.rst_
```diff
@@ -22,6 +22,7 @@ modifiers to control how the rows are counted [Default is **+a**]:
...
+default to -infinity and + infinity, respectively (e.g., **-qo**\ 50/**+c**\ 2 will only write records whose z-values is ≥ 50).
```
meghanrjones, line 26:
```suggestion
default to -infinity and + infinity, respectively (e.g., **-qo**\ 50/**+c**\ 2 will only write records whose z-values (in 3rd column) is ≥ 50
while *-qo**\ /50**+c**\ 2 will only write records whose z-values is ≤ 50).
```
Is the inequality for the suggested example correct here (i.e., equal than is included in both directions) or should it be < 50?

meghanrjones (commented):
What about including the *min*/*max* option in the synopsis? Is there a way to use both *row* and *limits*[**+c***col*] at the same time?

**-q**[**i**|**o**][~]*rows*|*limits*[**+c***col*][**+a**|**f**|**s**]

Also, looking at the interval specification, I think the following is more accurate becaause stop is not required and you can use just the colon in addition to colon stop.

*start*[:*inc*][:[*stop*]]

doc/rst/source/explain_-q_full.rst_
PaulWessel: Yes, both ways.

PaulWessel:
OK, I added in the _limits_ part to synopsis and docs.

meghanrjones (approved):

PR #5636, opened 2021-08-12
Do not prevent sample1d from sampling a duplicate knots

PaulWessel:
This PR fixes the concern raised in [this](https://forum.generic-mapping-tools.org/t/sample1d-dont-eliminate-duplicate-values/1976/2) forum post.  Whatever the initial issue was related to sample1d discussing the referred [post](https://forum.generic-mapping-tools.org/t/sample1d-on-gmt6-1-and-gmt6-2/830/5), there should never be a restriction on whether we can sample a spline more than once in the same spot.  That restriction falls on the data defining the spline only: These must be monotonically increasing and cannot have duplicate points.
I will ask Sabin to remind us what problem he was running into - whatever it was the removal of duplicates from the -Tknots is not the solution.  If it comes to it, we can add a modifier to -T to force removal of duplicates even in the knots file.

### DIFF
--- a/src/sample1d.c
+++ b/src/sample1d.c
@@ -325,7 +325,8 @@ static int parse (struct GMT_CTRL *GMT, struct SAMPLE1D_CTRL *Ctrl, struct GMT_O
 		Ctrl->T.active = true;
 	}
 	if (Ctrl->T.active) {	/* Do this one here since we need Ctrl->N.col to be set first, if selected */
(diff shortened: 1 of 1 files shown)


PaulWessel:
So Sabin responded that the removal of duplication is a valuable feature.  Hence I have added a modifier **+u** to **-T** so that when knots are given via a file (**-T**knots.txt) and those knots are not necessarily sorted and may include duplicates, the **+u** modifier will sort the knots and remove those duplicates in the process. Sabin will report back once he has had a chance to test this branch.

PR #5637, opened 2021-08-12
Feautures.rst. Typo in HDF4 example file name?

Esteban82:
In line 2761 the subset name is: HDF4_SDS:UNKNOWN:"A20030012003365.L3m_YR_NSST_9":0 (notice the position of the last ").
I change the position of " in lines 2772 and 2793.

I am not sure that it is a typo. I understand that the name should be exactly the same.

### DIFF
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -2769,7 +2769,7 @@ simple example using :doc:`/grdinfo` would be
 
    ::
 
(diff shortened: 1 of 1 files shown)


PaulWessel:
Pinging @joa-quim since I do not know this syntax.

PaulWessel:
Pinging @joa-quim so we can finish this simple but old issue...

joa-quim:
Not sure but I don't think it matters at all (with or without).

meghanrjones (approved):
The revision looks fine to me.

PR #5641, opened 2021-08-12
Maintenance: use gmt_get|set_column_type

PaulWessel:
We have two functions that gets (_gmt_get_column_type_ ) and sets (_gmt_set_column_type_ ) the column _type_ for input or output and given column, but we also have legacy code that manipulate the hidden array entries directly.  This PR just replaces some of the latter with the cleaner functions that makes the code easier to read.  No change in behavior and all tests still pass.


PaulWessel:
Maybe someone could approve this boring PR?

seisman (approved):

PR #5642, opened 2021-08-13
Update anim01.sh

Esteban82:
Update based on modifications of #5613.

Let me know if it ok so I correct the other animations.


PaulWessel (approved):
Yep, that looks good.

Esteban82:
I think that is all.

PaulWessel (approved):

PR #5657, opened 2021-08-16
Enable gridlines when colormap -B implies it

PaulWessel:
See this forum [post](https://forum.generic-mapping-tools.org/t/more-colorbar-user-errors-or-bugs/1983) for background.  Turns out that when **colormap** used **-B** we did not draw gridlines since the Cartesian axis function used by design skipped gridlines as those were expected to be done later in the _gmt_map_basemap_ function.  Except for **colormap** which does not use _gmt_map_basemap_ and hence there was no call to plot gridlines.  This PR fixes that issue, updates a test original PS (_legcpt.sh_) that now failed (since it should have had gridlines but did not), and did some minor coding improvements to make it easier to understand the algorithms related to custom annotations.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -102,6 +102,13 @@
 #	include <mex.h>
 #endif
 
+/* These are used in gmtinit_init_custom_annot and gmtinit_decode_tinfo onlly */
+#define GMT_ITEM_ANNOT		0
+#define GMT_ITEM_INTVAL		1
+#define GMT_ITEM_TICK		2
+#define GMT_ITEM_GRID		3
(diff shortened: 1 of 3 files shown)


Esteban82:
This is correct, right?

`    gmt colorbar -Cseis_pgv.cpt.txt -DjCM+h -F -Q -Bpxc"annots.txt" -By+l"cm/s"`

sample_colorbar_issues

PaulWessel:
Yes, before we never got gridlines if -B was used, only if -B was not used and -S+s was not used.  I guess most people did not want gridlines so they did not notice until the person who posted the problem tried.

Esteban82 (approved):

PR #5658, opened 2021-08-16
Ensure colorbar uses -S+y when no -B given but -L is given

PaulWessel:
There was some confusion, as documented in this forum [post](https://forum.generic-mapping-tools.org/t/more-colorbar-user-errors-or-bugs/1983). This PR fixes this issue.  All tests still pass.

PR #5659, opened 2021-08-16
Better estimation of background frame for colormap -Q

PaulWessel:
The calculation of the size of the background panel (**-F**) did not consider the differences in annotation widths when **-Q** (logarithmic) is used, as pointed out in this forum [post](https://forum.generic-mapping-tools.org/t/more-colorbar-user-errors-or-bugs/1983).  This PR addresses this omission by estimating the width of 10^value strings as well as the vertical increase in annotation height due to the superscripts.


Esteban82:
Is it ok, right Paul? 

```
    gmt colorbar -Cseis_pgv.cpt.txt -DjCM+h -F -Q
    gmt colorbar -Cseis_pgv.cpt.txt -DjCM+h -F -Q -S+y"cm/s"

```
sample_colorbar_issues

PaulWessel:
Hm, no if the -S is not working for this case.  I will have a look (but may be after dinner).

Esteban82:
I think that -S if another PR (#5658). In this PR I think that tje issue was the frame of subplot B and C. 

imagen

PaulWessel:
yes, b and c are correct here.

Esteban82 (approved):

PR #5678, opened 2021-08-18
Strengthen parsing of events -Ardpi

PaulWessel:
The **-Ar**_dpi_[**c**|**s**] option had parsing that did not check for bad arguments.  This PR fixes that.  All tests still pass.


Esteban82:
Should I approved without testing, right?

PaulWessel:
I think so.  I ran all the test and no change, and it was not in response to a bug (just wanted to fix something as I thought you had passed -Ar1080p but you did not).

Esteban82 (approved):

PR #5681, opened 2021-08-19
Point out that events -Ardpi may need to go into the main script

PaulWessel:
If the region or projection changes as part of the frame changes then the resolution selection for the discretization of the line may become inaccurate.  In that case this step should be added as part of the frame script.  This PR points this out in the documentation.

### DIFF
--- a/doc/rst/source/events.rst
+++ b/doc/rst/source/events.rst
@@ -57,9 +57,7 @@ To show the display of events visible for May 1, 2018 given the catalog of
 large (>5) magnitude earthquakes that year, using a 2-day rise time during
 which we boost symbol size by a factor of 5 and wash out the color, followed
 by a decay over 6 days and then a final shrinking to half size and darken the
-color, we may try
-
-   ::
+color, we may try::
 
     gmt begin layer
       gmt convert "https://earthquake.usgs.gov/fdsnws/event/1/query.csv?starttime=2018-01-01%2000:00:00&endtime=2018-12-31%2000:00:00&minmagnitude=5&orderby=time-asc" \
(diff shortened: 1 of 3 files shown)


Esteban82:
Are you sure that is the way to do it? I have my doubts.

PaulWessel:
I will label this a WIP branch for now.  The thing is that if you wish to plot a variable thickness line then you must use -Ardpi to descretize it and then use -Scsize to plot points that simulate a line.  If you use -Ar as you do then you draw an actual line and the -Ardpi step is not needed.  Sorry, I was fixated on the first case and did not notice that you used -Ar.  So your line works great but you have no way of controlling variable thickness, color, or intensity at the tip, for instance.

PaulWessel:
See anim13 for what I am talking about.

Esteban82:
Yes, I agree that -Ardpu is needed. 

However I am not sure if it is a good idea to include that step within the main script when the maps change projection/region. I have the impression that some problems will arise with lines going trough the antimeridian. And that the solution would be to split the data file it two halves (eastern and western hemisfere) and run -Ardpu for each subdataset. 

But I have to test it first.

Esteban82:
I based my suspicion on this:

https://forum.generic-mapping-tools.org/t/dr-jones-fligth-to-venice-in-his-last-crusade/2004/17?u=esteban82

PaulWessel:
The added documentation is needed and correct. When events **-Ar**_dpi_ is used it must convert input line to plot coordinates which involves **-R -J** and then determine the required point spacing along lines using the desired resolution.  If **-R -J** changers then the spacing between points will change.  How much depends on how much **-R -J** changes, of course, but it is nevertheless true that it does.  Note that most of your test cases with Indiana Jones involved plotting lines and not the dense point cloud as circles - it is that latter case that this applies to.  There is no point using **-Ar**_dpi_ unless you plot the output as circles. So I have removed the WIP.

Esteban82 (approved):
Ok. I approved it because you sound so sure.

PR #5682, opened 2021-08-20
Ensure that valid formats jpeg and tiff result in those extensions

### LINKED ISSUE #5669: Error if using the "jpeg" as the format suffix

```
gmt begin map jpeg
	gmt basemap -R0/10/0/10 -JX10c -Baf
gmt end show
```

`jpeg` is used but the generate plot is "map.jpg".

**Full error message**

```
The file /Users/seisman/workspace/map.jpeg does not exist.
docs [ERROR]: Opening local files via open map.jpeg failed with error 256
docs: No such file or directory
end [ERROR]: Failed to call docs
```

**System information**

* Operating system: macOS
* GMT version (`gmt --version`): 6.3.0_c9da416_2021.08.11

PaulWessel:
While we accepted these names, they would not set the correct file extensions (always got .jpg and .tif regardless) and jpeg failed to open via **gmt end show**.  This PR renames the output from **psconvert** (which is always .jpg and .tif) to the requested names.
Closes #5669.

### DIFF
--- a/src/gmt_gsformats.h
+++ b/src/gmt_gsformats.h
@@ -29,7 +29,7 @@
 #define GMT_GSFORMATS_H
 
 /* List ps at end since it causes a renaming of ps- to ps only.  Also allow alternative jpeg and tiff spellings */
-static char *gmt_session_format[] = {"pdf", "jpg", "jpeg", "png", "PNG", "ppm", "tif", "tiff", "bmp", "eps", "ps", NULL};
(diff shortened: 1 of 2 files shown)


PaulWessel:
Pinging reviewers @meghanrjones and @seisman.

seisman:
It works well. 

One small "bug" is that, if someone uses `jpg` first and generates a "map.jpg" image in the current directory. Changing `jpg` to `jpeg` and running the script again, the old "map.jpg" file will disappear.

seisman (approved):

PaulWessel:
Right.  Small price to pay for making the same format but using two different names.  I have no solution to this but to ignore it.

seisman:
I'm OK with ignorning it.

PR #5683, opened 2021-08-20
Improve animation of trajectories in events

PaulWessel:

_Trajectories_ are lines given by (_x,y,t_) and in animations we wish to plot the portion of this line up to the current time.  Because the discrete times of points in the trajectory will in general not coincide with the frame times, we must interpolate between the last data point older than current time and the next (future) point so that the end of the trajectory for the current time can be estimated.  This PR enables this feature and improves the documentation.  We also add a test script (test/movie/trajectory.sh) to demonstrate the process.  The resulting 24-frame movie shows the original points of the trajectory (circles), the anticipated path (dashed line) and the red line is the advancing trajectory.  Since this is plotted as a line via **events -Ar** the only control over appearance is that via **-W**_pen_.

https://user-images.githubusercontent.com/26473567/130303135-6da223bb-73cb-4c4d-9a5b-5676cadc9711.mp4

This PR improves **-Ar** which did not do any interpolation and was dependent on the lines being finely resampled so that we could simply just plot all the points older older than or equal to the current frame time and not worry about interpolation.  The documentation now points out that if you are plotting geographic trajectories with large spacing between nodes then you may wish to resample such lines with a finer spacing so that the final interpolation in **events** does not affect the shapes of great circles or loxodromes.


PaulWessel:
Adding more reviewers as I need this merged before another PR that depends on it.

seisman (approved):

PR #5687, opened 2021-08-21
Demonstrate events trajectory via resampling the line as points

PaulWessel:
This is the second test case showing how to animate a trajectory via **events**. Unlike the case in #5683, this case uses **events -Ar**_dpu_ to create a dense collection of points that can then be plotted via **events** as circles.  The resulting movie looks like this:

https://user-images.githubusercontent.com/26473567/130308304-89ab1654-6223-4fcb-8cdd-6d8f0452f014.mp4

Some differences with the regular trajectory plotting via **-Ar**:

1. No line is plotted, just dense circles.  This means the line will have always the appearance as being plotted with **PS_LINE_CAP**=_round_.
2. Because points are plotted, the full machinery of events **-M** can be used to create a variable thickness line.
3. The resampling from coarse line to dense points reflects the default spline [Akima].  To avoid excursions from straight lines one need to add **--GMT_INTERPOLANT**=_linear_ to the **events -Ar**_dpu_ call (which was done for this example).

Future PR will demonstrate the variable line thickness case. These cases will eventually be used in a new documentation accessible from the events page that discusses how to animate lines.

### DIFF
--- a/test/movie/trajectory2.ps
+++ b/test/movie/trajectory2.ps
@@ -1,0 +1,1611 @@
+%!PS-Adobe-3.0
+%%BoundingBox: 0 0 346 195
+%%HiResBoundingBox: 0 0 345.6000 194.4000
+%%Title: GMT v6.3.0_2ac3556-dirty_2021.08.20 [64-bit] Document from plot
+%%Creator: GMT6
+%%For: unknown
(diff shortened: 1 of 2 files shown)


Esteban82 (approved):
I get the same movie.

PR #5688, opened 2021-08-21
Let events use GMT_Read_Data instead of GMT_Get_Record

PaulWessel:

This simplifies i/o and avoids some complications as **events** makes calls to **psxy** and **pstext**.  Also properly does the disabling and reenabling of any **-b -g -h -i** options around the calls to other modules.


Esteban82 (approved):
I approved it because I suspect that there is nothing to test. If it not the case, the me know.

PaulWessel:
Yes, true.  For these PRs I always run the full test suite and they pass so it is mostly internal recoding.  Thanks.  I will deal with the conflict which is due to the recoding.

PR #5695, opened 2021-08-22
Let events support temporal color changes

### LINKED ISSUE #5684: Let events also have option to modulate color with time

Presently, **events** offer **-Mi**|**s**|**t** to modulate a symbols intensity, size, and transparency via their individual amplitude-time curves.  This allows us to briefly magnify a symbol, change the intensity of the color, and modify transparency as the time of that event reaches the time of a frame.  The fourth attribute, color, is _invariant_ other than via a CPT and is looked up via the symbols _z_-value.  However, the _z_-value is fixed for the given event time (it is an attribute of the point).  Hence, there is no way to change the color of a specific symbol through time (the intensity option keeps the hue _constant_ and only changes saturation-value).  An application where this may be desirable is when we plot a trajectory as a function of time and wish the leading head of the trajectory (or 'comet' in MATLAB speak) briefly to change its color from when it first arrives and then decay back to its background color when it joins the other points that make up the tail.  The solution to this would be a new directive **-Mc**[_amp_].  The presence of this option would change a point's data value from _z_ (static) to _z + dz(t)_, where _dz(t)_ is another time-function like those the user can define for size, intensity, and transparency. The _dz(t)_ would go from 0 and reach a peak of _amp_ at the time of the event.  Without **-Mc**, _amp = 0_ and _dz(t) = 0_ for all times.  Otherwise, the event's _z_-value will briefly increase and via the CPT lookup it will return different colors.  A simulation is shown here:

t

I think this feature makes most sense for symbols such as dense circles simulating a line (probably generated via **events -Ar**_dpi_), and that all these points share a _z_ of zero (basically, they have no z-value).  Then, **-C**_cpt_ would simply have a color ramp from _z = 0_ to _z = amp_ and colors are thus changing because z(t) of each point is changing. As default, _amp_ would be 1 and the CPT would be expected to return different colors on the _z = 0/1_ interval only.

@Esteban82 is probably best positioned to comment on this, but anything goes.

PaulWessel:

This PR adds one more time-function (via **-Mz**) that adds the relevant fraction of the amplitude [1] to the data set's _z_-values (mostly expected to be zero).  Thus, as _z_ now becomes modulated around the event time the colors obtained by CPT lookup also changes.  I have updated the documentation and added the fourth time-function description, as well as a movie still example (trajectory5.sh) which when run as GIF demonstrates the effect (note the change of line color at the burning end).

trajectory5

Closes #5684.


PaulWessel:
Note: I am adding a modifier **+z**[_z_] to **-Ar** so that it is easy to insert a constant _z_-column on the fly.  This columns is needed if **-Mz** is used and most lines will not have an original _z_-column, so this simplifies making one.

Esteban82 (approved):
Great feature. Thanks Paul!

PR #5699, opened 2021-08-23
Improve events testing via debug feature and test script

PaulWessel:
Like we did for **greenspline**, I have implemented an undocumented way to output just plain time-functions so we can test them, here in test script _psevents_functions.sh_, which have been added.  The main change was to create a sub-function for the evaluation of the four time-functions so that it could easily be called for the debug case.

### DIFF
--- a/src/psevents.c
+++ b/src/psevents.c
@@ -54,7 +54,14 @@ enum Psevent {	/* Misc. named array indices */
 	PSEVENTS_VAR_ENDTIME = 3,
 	PSEVENTS_LINE_REC = 1,
 	PSEVENTS_LINE_SEG = 2,
-	PSEVENTS_LINE_TO_POINTS = 4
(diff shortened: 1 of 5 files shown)


Esteban82 (commented):
For test/movie/trajectory5.sh I get this figure:

trajectory5

But for test/psevents/psevents_functions.sh I get this error message:
```
events [ERROR]: Unrecognized option -/
events [ERROR]: Option -T: -T<now> is a required option.
events [ERROR]: Must specify either -A, -S or -Z unless just plotting labels.
events [ERROR]: Option -Mz: Requires -Cevents [ERROR]: Must specify -R option
```
Am I doing something wrong?

PaulWessel:
No, I had placed that option under #ifdef DEBUG so you dont get it unless you build with that defined.  Try again, I just removed the ifdefs.

Esteban82 (approved):

PR #5700, opened 2021-08-23
Use dynamic allocation in legend to accommodate any number of items

### LINKED ISSUE #5698: Auto-legend crashes if more than 65 entries are added

_The bug was first reported by https://github.com/maihao14 in https://github.com/GenericMappingTools/pygmt/issues/1461._

```
gmt begin map
gmt basemap -R0/10/0/10 -JX10c -Baf -Y20c
for i in $(seq 1 65); do
	echo 5 5 | gmt plot -Sc0.2c -Gred -lPoint$i
done
gmt end show
```

It works for 64 legend entries but crashes for >=65 entries.

If the maximum number of allowed entries is 64, we should:

1. Document the number in the **-l** option explanation
2. Return an error or a warning (truncated to the first 64 entries) if more than 64 entries are given

**System information**

* GMT version (`gmt --version`): 6.3.0_c12a98f_2021.08.21

PaulWessel:
As currently designed, **legend** can only place _N_ items of the same kind (e.g., symbol entries).  This was not documented and _N_ was set to 64.  I have changed this to 256 and added the limitation to the documentation.

Closes #5698.

### DIFF
--- a/doc/rst/source/legend_common.rst_
+++ b/doc/rst/source/legend_common.rst_
@@ -301,3 +301,8 @@ script you need to repeat it (%%); hence the font switching mechanism
 This only applies to text inside a script or that otherwise is processed
 by DOS. Data files that are opened and read by the module do not need
 such duplication.
+
+Limitations
+-----------
+
+Currently, each type of legend entry is limited to a maximum of 256 such entries.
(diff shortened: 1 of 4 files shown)


seisman (commented):
It still crashes if more than 256 entries are given, although so many entries are rarely used. 

Is it possible to raise an error, rather than a crash?

PaulWessel:
Yes, let me see...

PaulWessel:
This one reallocates those tables.  I tested it and should work now regardless of number of items.  Please give it a spin, @seisman.  Note I added some future code to gmt_io.c that I may work on later - it is not being called.

seisman (approved):
Looks good to me. Please update the PR title before merging.

PR #5701, opened 2021-08-25
Improve -R description and fix a few minor docs issues

### LINKED ISSUE #5686: Improve -Rxmin/xmax/ymin/ymax[+uunit] docs
I think that this way of indicating the region should be separated of -Rxmin/xmax/ymin/ymax. It is important to indicate that when this method is used, a longitude and latitude must be indicated in -J (important for mercator at least), as far as I know. Are those distances measured from the given long, lat to the sides? Because it could vary based on the projection (like Mercartor).

meghanrjones:

This PR separates the **-R***xmin*/*xmax*/*ymin*/*ymax* and **-R***xmin*/*xmax*/*ymin*/*ymax* **+u***unit* options and makes a few minor documentation improvements.

Fixes #5686 

**Reminders**


Esteban82:
If I use `-R0/100/0/100+uk -JM0/0/15c` then the distances are relative to the LB the map.

meghanrjones:
> If I use `-R0/100/0/100+uk -JM0/0/15c` then the distances are relative to the LB the map.

Thanks for fixing this.

PR #5710, opened 2021-08-27
Draw clip path before initiating clipping

### LINKED ISSUE #5704: psclip (gmt6.2) bug with -W -N option
I am observing a potential bug with psclip using GMT6.2, when activating the option to plot the outline of the clip path, which seems to override the -N option. 

An example, underneath, where the only difference between the two plots is adding the -W.25p,blue on the psclip line.

#!/bin/sh
RR=-R0/1/0/1
JJ=-JX10
echo "0 0
1 1
0 1" > clip
echo ".5 .25
.5 .5
.5 .75
.25 .5
.75 .5" > data
gmtswitch gmt6

psfile=test-psclip6a.ps
psbasemap $RR $JJ -Bxya.5g0f.1 -BWSen -X4 -Y4 -K > $psfile
psclip $RR $JJ -O -K clip -N >> $psfile
psxy $RR $JJ -O -K -Sc.5c -G128 -W.5p,0 data >> $psfile
psclip $RR $JJ -O -C >> $psfile

psfile=test-psclip6b.ps
psbasemap $RR $JJ -Bxya.5g0f.1 -BWSen -X4 -Y4 -K > $psfile
psclip $RR $JJ -O -K clip -N -W.25p,blue >> $psfile
psxy $RR $JJ -O -K -Sc.5c -G128 -W.5p,0 data >> $psfile
psclip $RR $JJ -O -C >> $psfile

test-psclip6a
test-psclip6b

PaulWessel:
See #5704 for background.  The problem was that the drawing of the path happened in the same loop as we are laying down the clip path and that fails somehow with **-N**.  Now we draw the clip path first (if **-W** is given), and then lay down the clip path.  I have added a test to show the four version of the OP's case.  In fixing this I noticed **clip** would complain if used in subplots since the **-N** restriction on **-B** collides with what **-B** is set via **subplot**.  Thus I needed to let **clip** know when it is being used in a panel.

 Closes #5704.

clip_draw

All tests including the new one now passes.

### DIFF
--- a/doc/rst/source/clip_common.rst_
+++ b/doc/rst/source/clip_common.rst_
@@ -96,7 +96,7 @@ Optional Arguments
 .. _-W:
 
 **-W**\ *pen*
-    Draw outline of clip path using given pen attributes [Default is no outline].
+    Draw outline of clip path using given pen attributes before clipping is initiated [Default is no outline].
 
 .. |Add_-XY| replace:: |Add_-XY_links|
 .. include:: explain_-XY.rst_
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -2862,7 +2862,7 @@ GMT_LOCAL void gmtinit_freeshorthand (struct GMT_CTRL *GMT) {/* Free memory used
 	gmt_M_free (GMT, GMT->session.shorthand);
 }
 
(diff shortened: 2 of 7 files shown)


Esteban82 (approved):
I got the same figure.

mvanneste:
Many thanks!
Maarten

Issue #5663, opened 2021-08-17
Improve anim02


Esteban82:
I see that the [animation 02](https://docs.generic-mapping-tools.org/dev/animations/anim02.html) has some issues that could be improve:

1. Its upper boundary is cropped. Also the right (see the W at the LR).
2. Its has low quality (dpi). I think it could be upgrade.
3. A mp4 could be make with more resolution/size (and post on youtube as other animations). 
4. Anything else?

I offer to make the modifications. Let me know if you agree with the 3 items.

PaulWessel:
Thanks for bringing this up. We may wish to discuss this at a community meeting - or perhaps via github discussions.  Those first 4-5 animations predate movie and was done via scripting.  To show the result without having a very large file we dumbed them down to small gifs mostly.  Of course, this makes them look crude and not very impressive, especially 2-4.  They also differ from some of the later ones which are more complicated and fancier.  So there is some value in having very simple animation examples (spinning earth, etc) to get people started. I see anom02.sh also is printing its own angle label at TR even though movie -L should be used for this. So yes, we should fix this one but probably all of them, and possibly only do high-res YouTube versions and using standard HD format.  For anim02 this may mean extending the region to better fit a HD frame.  Looks like anim04 is failing - I see just a white movie with some line at my end.  Anim 5 only has a small number of frames (one per eigenvalue) so we annot really make that a longer movie unless we use a more elaborate gridding example.

So let's wait with doing anything until we have a group discussion next Thursday on what to do with animations in general, @meghanrjones can you please add that topic to the agenda?

meghanrjones:
I added it to the agenda for next week. Thanks for opening the issue and offering the make the modifications, @Esteban82!

Esteban82:
I am working a bit on the script. I see that the following line is used. 

`	gmt plot -Sc0.8i -Gwhite -Wthin <<< "256.25 35.6"`

Just for curiosity, is it the same than:

`echo 256.25 35.6 | gmt plot -Sc0.8i -Gwhite -Wthin
`

Should I use a pipe instead?

Paul, is it possible to print the rotating arrow with movie -L? I don't know how. I used an inset.

PaulWessel:
Yes, the <<< can be replaced with the pipe you suggested.

[...]

PR #5711, opened 2021-08-28
Anim02 improve

Esteban82:
See #5663 for details

### DIFF
--- a/doc/examples/anim02/anim02.sh
+++ b/doc/examples/anim02/anim02.sh
@@ -1,10 +1,14 @@
 #!/usr/bin/env bash
 #               GMT ANIMATION 02
 #
-# Purpose:      Make simple MP4 of an illuminated DEM grid
+# Purpose:      Make simple MP4 of an illuminated DEM grid of the US Rockies
 # GMT modules   math, makecpt, grdcut, grdimage, inset, plot, text, movie
 # Unix progs:   cat
 # Note:         Run with any argument to build movie; otherwise 1st frame is plotted only.
+#
+# The finished movie is available in our YouTube channel as well:
+# https://youtu.be/WQ6JrtEu_Fk
+# The movie took 1 minute 45 seconds to render on a 24-core MacPro 2013.
 
 if [ $# -eq 0 ]; then	# Just make master PostScript frame 0
 	opt="-Mps"
--- a/doc/examples/anim02/anim02.sh
+++ b/doc/examples/anim02/anim02.sh
@@ -1,8 +1,9 @@
+
 #!/usr/bin/env bash
 #               GMT ANIMATION 02
 #
 # Purpose:      Make simple animated GIF of an illuminated DEM grid
-# GMT modules   math, makecpt, grdimage, plot, movie
+# GMT modules   math, makecpt, grdcut, grdimage, inset, plot, text, movie
 # Unix progs:   cat
 # Note:         Run with any argument to build movie; otherwise 1st frame is plotted only.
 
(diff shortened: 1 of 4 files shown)


Esteban82:
Lacks to upload the video on youtube and add the link. Also the precessing information? It is fast to process.

PaulWessel:
I think the movies that are fast to make have no information about how long.  It is just the longer ones where I added that I think.

Esteban82:
Yes, I think that no one will get anxious.

doc/examples/anim02/anim02.sh
```diff
@@ -1,8 +1,9 @@
+
```
seisman, line 1:
```suggestion
```
Remove the empty line so that the first line is `#!/usr/bin/env bash`

doc/examples/anim02/anim02.sh
```diff
@@ -14,20 +15,22 @@ fi
 # 1. Create files needed in the loop
 cat << 'EOF' > pre.sh
 gmt begin
-	gmt math -T0/360/10 T 180 ADD = angles.txt
-	gmt makecpt -Crainbow -T500/4500 -H > main.cpt
-	gmt grdcut @earth_relief_30s -R-108/-103/35/40 -Gtopo.nc
+	gmt math -T0/360/1 T -o0 = angles.txt
+	gmt makecpt -Coleron -T500/4500 -H > main.cpt
+    gmt grdcut @earth_relief_15s -R-114/-103/35/40 -Gtopo.nc
```
seisman, line 20: Use tabs or whitespaces?

doc/rst/source/animations/anim02.rst
```diff
@@ -3,20 +3,17 @@
 (2) Examining DEMs using variable illumination
 ----------------------------------------------
```
seisman, line 4: The PS file is too large. We should not add it to the repository.

doc/examples/anim02/anim02.sh
Esteban82: Maybe both. Which should I use?

seisman: I think most other examples use tabs.

PaulWessel (approved):

seisman:
I think we need to:

1. Upload the high-resolution video to YouTube 
2. Add the YouTube link to the `doc/rst/source/animations.rst` file and `doc/rst/source/animations/anim02.rst`
3. Reomve the PS file

right?

PaulWessel:
Yes and no. This short 360 frame animation is saved as an animated GIF.  Cannot upload to YouTube I think? Shall we change it to a MP4 so we can do that instead? I guess we dont what HD GIFs included in the repo.

PaulWessel:
Two comments.  I tested making a mp4 instead and noticed this:

1. The angles.txt file should be renamed azimuths.txt to reflect what they are used for.
2. The MP4 last frame says 359 degrees even though the PNGs goes to 360. Is this just me in QuickTime that is missing that frame or is it a FFmpeg issue? The last PNG frame does shows 360.  Perhaps it is because we have 361 frames and 36 frames per second and you get 10 seconds plus 1 frame? Changing it to standard 24/s works fine, so perhaps we should do that (36/s is a bit fast anyway)?

Agree, @Esteban82 ?

Esteban82:
Fine by me. Since 24 fps is the default value, then that argument can be deleted.

PaulWessel:
Hi @seisman and @Esteban82, in animations.rst we have 6 initial GIFs then the youtubes, but now anim02 will be part of the youtubes. How should we arrange this?

1. Split up to show anim01 gif, then 02 youtube, then 3-6 git, then the rest youtube
2. Renumber things so 3-6 becomes 2-5, our 2 becomes 6 and we do gif 1-5 and youtube rest
3. Replace all the others gif anims 1-6 with mp4 HD versions and place all on youtube.

The key with doing this in the first place is that a HD animation is a better selling point than a tiny crude animated gif made that way to avoid storing large files in github.  So I guess I am leaning on 3 even though the work has not been done yet.

seisman:
I agree with option 3.

Esteban82:
I also agree with option 3.

PaulWessel:
OK, so we will replace 1,3-5 as well.  I am working on a 5 replacement now.  Let me know if you are interested in upgrading any of the others so we dont both do the same ones...  I can do any of them (and will do any you dont care about).  Goals is HD Mp4 version and do not have to be the exact same but perhaps "inspired" by the theme of those animations.

Esteban82:
Ok. I have already started with 3 and I could do also 1 and 4 (which in a first impression it seems similar to anim09). I let you anim05.

PaulWessel:
Thanks, and I think despite making them longer and higher resolution we should try to keep the first few relatively simple (i.e., relatively short script) and not spend too much perfecting them with enhancements.  They are meant to get people started on animation and then the complexity can increase for the later examples.

Esteban82:
Ok. 

This PR can be squash and merge? There is an error in Vercel.

PaulWessel:
Not sure yet what to do with animations.rst etc.

seisman:
Need to remove the PS file, remove the anim02 from the tests and also add it to animations.rst. 

I can help with it by committing changes directly to this branch.

PaulWessel:
OK, that would be great @seisman

Esteban82:
I view the [vercel deployment ](https://gmt-c0vcmfyap-gmt.vercel.app/animations.html)and doesn't look good. Maybe we should update all the animations within this PR?

seisman:
@Esteban82 I fixed a typo in animations.rst. Please view the website again.

PR #5714, opened 2021-08-30
Update greenspline -C for writing intermediate grids

PaulWessel:
Currently, the modifiers **+m** and **+M** are _mutually exclusive_ and thus it is only possible to output intermediary grids for the cumulative _or_ incremental solution.  However, since the main calculation time is in determining the solution, it is helpful to be able to write both types of grids at the same time.  This PR replaces the **+m** and **+M** with two new modifiers that both can be set: **+c** for cumulative grids and **+i** for incremental grids.  We now use the regular **-G**_grdfile_ setting and automatically insert` _inc_###` or `_cum_###` before the file extension (the number of ### is also determined), and we start numbering with eigenvalue 0 rather than calling it 1 (everything else in GMT starts at 0 so starting at 1 causes grief in **batch** and **movie**). The old **+m**|**M** modifiers are honored in a backwards compatible way with the original file name template and numbering scheme.  All tests pass and a new replacement _anim05.sh_ using the new system works as well.


meghanrjones:
The example from the greenspline docs crashes with a seg fault:
`gmt greenspline @Table_5_11.txt -R0/6.5/-0.2/6.5 -I0.1 -Gcontribution.nc -Sc -Z1 -C+c`

PaulWessel:
Yes it does! Of course I did not try that simple example, just my more complicate one.  Will see why.

PaulWessel:
Fixed.  My case worked because I used +i also.  Now it works in all cases.

meghanrjones (approved):
Two comments:
1. The results from **-C** **+f***file* are written with 1 as the first eigenvalue. Should this be updated to zero similar to the grids?
2. From the examples, *value* is optional. The docs should be updated accordingly with the default behavior specified.

Also, the **-C**[**n**] has an impact on which eigenvalues are eliminated for the fit, but not for what is written in the **+f** *file* output or the intermediate grids. Is this intentional?

PaulWessel:
Good points, likely answers:

1. We should be consistent using the 0-based counting, so first eigenvalue should be number 0.
2. I will update the docs regarding the optional value
3. Unintentional, and I think if someone gives a value for n then they probably do now want to run this all the way but stop at n.  I will make it that way.

PaulWessel:
Yes, I forgot to make a PS and check for height.  I will do that and possibly add a -Y or change the plot heights a bit.  Thanks!

PaulWessel:
I have made those changes and clarified some of the logic which is due to the fact that the eigenvalues/vectors we receive are not sorted so we set those we dont want to zero, but those zeros are distributed across the full vector, hence we must loop over the length of the eigen value vector and skip if entry is zero.

PaulWessel:
Also fixed the compile failure for Windows due to unsigned into loop variable in OpenMP...

Issue #5718, opened 2021-08-30
grdtrack: Add oblique crossline option


chhei:
``grdtrack`` allows the erection of orthogonal crosslines from input line segments using the ``-C`` option, however, constructing a set of  crosslines at oblique angle to input line segments is currently not possible.

It would be great if the functionality of ``grdtrack -C`` could be extended to include a way to erect a set of crosslines which can be oriented at an **o**blique angle - potentially using ``-C+o<azimuth>`` (the ``-C+a`` flag is already taken).

Ref [forum post](https://forum.generic-mapping-tools.org/t/grdtrack-erect-cross-profiles-at-oblique-angles/2041)

PR #5719, opened 2021-08-31
Allow for angled cross-tracks in grdtrack -C

PaulWessel:
See #5718 for background.  This PR adds **-C** modifier **+d** to set a _deviation_ of tracks from being strictly orthogonal.  As the documentation explains, the sense of the deviation is such that when you look in the direction of the input track, a positive deviation will rotate the cross-tracks by that amount in the clockwise direction.  I have added to tests to ensure this all works for both Cartesian and geographic geometries.
I also found that the **+l**|**r** modifiers for Cartesian (only do the left or right half of the profiles) had things backwards - now fixed.
Hopefully @chhei can take this branch for a spin to ensure it works as he expects.


Esteban82 (approved):
When I run the tests I got the same figures than the tests but with the title. It is cropped in the ps. I wonder if this  could caused some issue. 

PS: I approve it because the important part is the same. 

crosstrack_cart

chhei:
Confirming that the added ``-C+d`` option works with a real-world example

PR #5727, opened 2021-09-05
Report residuals, not deviation from mean

PaulWessel:
The **-E** output table listed as residuals the difference between data and mean data.  Now we correctly list the difference between data and model.


Esteban82 (approved):
I don't see any test file. I approve. I assume that you are sure.

PaulWessel:
Yes, sorry, no test for this (too big) but it is a simple bug fix so I am sure.  Thanks.
-p

PR #5732, opened 2021-09-06
Reset PSL internals at end of building each completion procedure

PaulWessel:
The way **subplot** tags and **movie** labels and progress indicators work is that the _PostScript_ commands needed to place these items are stored as a "completion" _PostScript_ procedure and then the procedure is called at the end of the **subplot** or the whole plot. One tricky thing is that if you for instance change a color in a completion procedure, the next PSL function requesting the same color in the main module will think it was already set, but the procedure has not yet run, hence we fail to set the color.  I had a partial solution to this by resetting key internal parameters, but it was incomplete and failed in **movie** when several **-L** labels with the same color or pen outline were selected (the first label was correct, the next missed the same fill/pen).  This PR gathers all the resetting statements in a new internal function (_gmtplot_reset_PSL_) that is now called at the end of building a completion procedure.  Tests still pass and my problem case also works.

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -8580,6 +8580,18 @@ GMT_LOCAL void gmtplot_prog_indicator_F (struct GMT_CTRL *GMT, double x, double 
 	PSL_plotsymbol (GMT->PSL, xt, 1.2*s*dy2, &w, symbol);
 }
 
+GMT_LOCAL void gmtplot_reset_PSL (struct GMT_CTRL *GMT, struct PSL_CTRL *PSL) {
+	/* Because PSL_*_completion procedures are called at the end of the module but crafted at the start, we must forget we used any fonts, fills, pens in creating them,
+	 * otherwise the next PSL call requesting a fill or pen will not be properly set since PSL may think it is already the current choice. */
+	PSL->internal.font[PSL->current.font_no].encoded = 0;	/* Forget about font encoding */
+	PSL->current.font_no = -1;					/* Forget what the current font is */
+	PSL->current.linewidth = -1.0;              /* Forget we ever set a line width */
+	PSL->current.outline = -1;					/* Forget we requested polygon outline */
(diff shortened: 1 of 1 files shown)


Esteban82 (approved):

Esteban82:
Is this WIP? Because I am getting this message when I run gmt movie. However I don't see anything wrong in the mp4 or master frame.

```
free(): invalid pointer
movie_frame.sh: línea 16: 59025 Abortado                (`core' generado) gmt grdview above.nc -Rabove.nc -Ciceland.cpt -Qc${MOVIE_DPU} -Iintens.nc -JM19c -JZ0.4c -p-210/35+v4.5c/1.7c -X7.6c -Y5c
movie_frame.sh: línea 16: 59028 Abortado                (`core' generado) gmt grdview above.nc -Rabove.nc -Ciceland.cpt -Qc${MOVIE_DPU} -Iintens.nc -JM19c -JZ0.4c -p-210/35+v4.5c/1.7c -X7.6c -Y5c
free(): invalid pointer
movie_frame.sh: línea 16: 59043 Abortado                (`core' generado) gmt grdview above.nc -Rabove.nc -Ciceland.cpt -Qc${MOVIE_DPU} -Iintens.nc -JM19c -JZ0.4c -p-210/35+v4.5c/1.7c -X7.6c -Y5c

```

PaulWessel:
No, it was not WIP and I ran all the tests.  However, I did not run all the animations (takes too long) except my new versions of 5 and an addition one with no problems.  I will have to test your animation I think.

PaulWessel:
As usual, no complains on macOS for your script.  Do you get the error with no -F, i.e., just making a single PNG? RUn that with -Vd to see if you can learn when it tries to free, and ideally try to figure out which line that causes it.  Could try to add `fprintf (stderr, "Got this far\n");` in a few places in grdview.c to see when it happens.

Esteban82:
Ok. Thanks. I will keep it in mind. The messages doesn't appear now.

PR #5734, opened 2021-09-07
Improve greenspline and gpsgridder histories, update and add animations

PaulWessel:
This PR further improves **greenspline** and **gpsgridder** via common constants and settings and options.  New here is

1. Introduction of three directives to **-C**: **n**, **r**, **c**.  We had **n** (control number of eigenvalues [all]) from before and **r** was default (limit based on ratio of eigenvalue to largest, but no **r** was allowed [0]).  This PR also adds **v** which limits eigenvalues based on the desired model variance fraction [1].  Both **n** and **v** can take trailing % for percent instead of values.
2. Enhanced **-E** so that when **-C +i** and/or **+c** are used we write a history misfit file with summaries of rms misfit for each eigenvalue solution (and add chi2 sums if **-W** is used).  For **gpsgridder** we also do this as total and east and north components separately.
3. Demonstration of the revised code is done via updated animation 05 (**greenspline**) and new animation 15 (**gpsgridder**).  Working on these lead to other bug fixes (e.g., #5732).
4. Documentation has been updated accordingly and have new sections on explaining the selection of eigenvalues and refer to the animations [links will need to be added and updated later].  I also added the basic solution equations (Green's functions and solution) to **gpsgridder** as we already do in **greenspline**.

Both animations show the cumulative and incremental solution as more eigenvalues are added to the solution. The purpose of both is to illustrate how selecting the number of eigenvalues is crucial for the final result and you are making trade-offs between misfit and undesirable short-wavelength noise. The _anim05_ example grids a ship-track data set whose points are shown on earth_relief_01m on the left panel, with the gridded solution on the right panel, and incremental additions in the center panel.  The rms misfit in meters are shown on the right.  Here is just a short snippet of the full movie (~2000 frames):

https://user-images.githubusercontent.com/26473567/132409677-60ab2932-17b7-44d7-b2a8-eb9f4ffb33e1.mp4

The _anim15_ animation derives from test _gpsgridder1.sh_ but has been converted to a movie example.  It shows the GPS data and model prediction on the left panel and the incremental vector components (x10 so we can see them) on the right.  Above the panels is the history of the rms for the combined (c) as well as for east (u) and north (v) components.  Again, this is just a short snippet of the full movie but you get the idea.

https://user-images.githubusercontent.com/26473567/132409880-662df14f-715b-4471-9912-d72a1bb87daf.mp4

All tests pass as before. **Note**: For no eigenvalues we only see the least-squares trend solution upon which the Green's function adds deviations.


meghanrjones:
Some comments about the new animation 5:

1. The current animation is actually quite nice in my opinion, since it is so simple. Though, I think it would be better as a mp4 rather than gif. Would you be open to keeping the old example in addition to the new example (both could be included as part of animation 5)?
2. I think it would look better to have the earth_relief on the left, gridded result in the center, and incremental changes on the right.
3. The colorbar should be at the top or to the side, so that it can be viewed while the playback GUI is up on youtube.
4. What is the motivation for having the colorbar range for the incremental values from -100 to 100? It seems that many values are cut off by this range.
5. The gridded data should have a colorbar as well.
6. I find it difficult to get any value from the center panel at the current video speed. Could you slow the animation down?
7. Would it be possible to color the data points by their z-values?

PaulWessel:
> Some comments about the new animation 5:
> 
> 1. The current animation is actually quite nice in my opinion, since it is so simple. Though, I think it would be better as a mp4 rather than gif. Would you be open to keeping the old example in addition to the new example (both could be included as part of animation 5)? Even if we fixed the layout to HD, how would you include it as part of the same animation?

Hm, not sure.  The whole point of this exercise was to beef up the animations to HD and MP4.  The format for the old anim05 is not HD given its region etc.
> 2. I think it would look better to have the earth_relief on the left, gridded result in the center, and incremental changes on the right.

That I can do.
> 3. The colorbar should be at the top or to the side, so that it can be viewed while the playback GUI is up on youtube.

You mean inside, like now, but at top or left/right? There is no outside here.
> 4. What is the motivation for having the colorbar range for the incremental values from -100 to 100? It seems that many values are cut off by this range.

Magnitudes drop off with eigenvalues.  We either have to show some normalized plot which does not tell you much about how large the contribution is, or use a variable CPT taht decays over time to lower magnitues, or we pick something like what I have which forces the first few frames to clip but it sort of works for the rest.
> 5. The gridded data should have a colorbar as well.

Yes, was lazy.
> 6. I find it difficult to get any value from the center panel at the current video speed. Could you slow the animation down?

I could but would not people just pause and advance one frame etc? Or perhaps not possible in YouTube viewer?
> 7. Would it be possible to color the data points by their z-values?

Yes, for which panel?  I could color the ones in the left panel with the actual predicted values (so slowly should converge on the background color?

PaulWessel:
I have updated anim05 as suggested and both animations are now up on YouTube and the GMT docs has been updated with the URLs.:

[anim05](https://youtu.be/7NQa4TORA3E)
[anim15](https://youtu.be/Pvvc4vb8G4Y)

doc/examples/anim05/anim05.sh
```diff
@@ -1,32 +1,48 @@
...
+# the left panel and the gridded result in the center panel as a function of the
+# number of eigenvalues included.  The right panel shows the incremental values
+# added as new eigenvalues are included.  Note the data distribution is such that
+# we have a good coverage in the north with larger gaps in the south - the solution
+# quality reflects this disparity.  We also report the RMS misfit between the model
+# and the data as the solution builds incrementally. Note that the ship data have
+# a few bad tracks which leads to artifacts in the gridded solution.
+#
+# The finished movie is available in our YouTube channel as well:
+# https://youtu.be/xxxxxxxxxxxx
```
meghanrjones, line 17:
```suggestion
# https://youtu.be/7NQa4TORA3E
```

meghanrjones (approved):

PR #5735, opened 2021-09-07
gmtsplit ended i/o before writing

### LINKED ISSUE #5729: gmtsplit doesn't work with externals
This failing test has escaped me for a long. It doesn't matter if input is an array or a file name. Result is the same.
```
julia> gmtsplit([-14.0708 35.0730 0; -13.7546 35.5223 0; -13.7546 35.5223 0; -13.2886 35.7720 0; -13.2886 35.7720 0; -12.9391 36.3711 0], C=45, A="45/15", f="g")
[Session GMT (9)]: Error returned from GMT API: GMT_WRITTEN_ONCE (85)
[Session GMT (9)]: Error returned from GMT API: GMT_WRITTEN_ONCE (85)
[Session GMT (9)]: Error returned from GMT API: GMT_WRITTEN_ONCE (85)
Vector{GMTdataset} with 0 segments
```

PaulWessel:
Closes #5729.  There was a superfluous _GMT_Begin_IO_ call (only needed for rec-by-rec operations, which a previous version of **gmtsplit** used) but worse it called _GMT_End_IO_ before writing the output...

### DIFF
--- a/src/gmtsplit.c
+++ b/src/gmtsplit.c
@@ -436,9 +436,6 @@ EXTERN_MSC int GMT_gmtsplit (void *V_API, int mode, void *args) {
 	if (GMT_Init_IO (API, GMT_IS_DATASET, GMT_IS_PLP, GMT_OUT, GMT_ADD_DEFAULT, 0, options) != GMT_NOERROR) {
 		Return (API->error);
 	}
-	if (GMT_Begin_IO (API, GMT_IS_DATASET, GMT_OUT, GMT_HEADER_ON) != GMT_NOERROR) {
-		Return (API->error);	/* Enables data output and sets access mode */
-	}
 
 	if (!Ctrl->S.active) {	/* Must extend table with 2 cols to hold d and az */
 		n_columns = D[GMT_IN]->n_columns + 2;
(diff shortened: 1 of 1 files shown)


joa-quim:
Hmm, worse. It now crashes Julia and trying to catch the crashing point does not even show any GMT call in the stacktrace.

joa-quim:
It crashes already in the ``get_dataset()`` julia function in the call to ``unsafe_copyto!`` at
```
https://github.com/GenericMappingTools/GMT.jl/blob/master/src/gmt_main.jl#L734
```
but this is used by all modules and debugging showed that the numbers (rows, cols, segments) are right so I can only think that the pointer points to corrupted memory.

Matlab crashes as well.

PaulWessel:
OK, will dig deeper and test in mex

PaulWessel:
Is your _get_dataset_ at the step where it is trying to obtain the result back from gmtsplit?

joa-quim:
It's like in MEX. It gets the data from a GMT_DATSET that comes from GMT, regardless of what module produced it.

joa-quim:
Now it works, thanks.

PaulWessel:
Great, please approve.

joa-quim (approved):

PR #5741, opened 2021-09-09
Fix grammatical error in triangulate docs

willschlitzer:
This pull request fixes a minor grammatical error in the doc string of the function `triangulate`.

Fixes #

**Reminders**

### DIFF
--- a/doc/rst/source/triangulate.rst
+++ b/doc/rst/source/triangulate.rst
@@ -47,7 +47,7 @@ Description
 
 **triangulate** reads one or more ASCII [or binary] files (or standard
 input) containing x,y[,z] and performs Delaunay triangulation, i.e., it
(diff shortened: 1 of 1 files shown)


willschlitzer:
I noticed this typo when I reviewed the [pygmt triangulate PR](https://github.com/GenericMappingTools/pygmt/pull/731). I proposed changing 

`performs Delaunay triangulation, i.e., it find how the points should be connected to give the most equilateral triangulation possible` 
to 
`performs Delaunay triangulation to find how the points should be connected to create the most equilateral triangulation possible`

I think the latter version reads better, but the change I have pushed so far in this PR is just fixing the grammar on "find".

PaulWessel (approved):
thanks!

PR #5760, opened 2021-09-14
Double-check after file lock if file magically is there

### LINKED ISSUE #5757: Need better handling when simultaneous jobs are trying to download the same file

In experimenting with a movie that uses @earth_relief and where the resolution depends on the settings (#5753 ), **movie** is launching many jobs to make the frames, and these all run concurrently.  If many or all of them decide they need to use @earth_relief_01m (which is a tiled dataset) then it results in a flurry of calls via curl to download the individual tiles.  Hence several concurrent jobs may seek to download the exact same file, e.g., N60E120.earth_relief_01m_p.nc.  We prevent this chaos by creating a _lock file_ with the name /tmp/_file_.download and then apply standard file lock protocols (_fcntl_ for the *nix and _LockFileEx_ on Windows).  What this does is to deny the other processes access - they will hang until the lock is released.  However, when it is released the file of interest has been downloaded so the whole need to download again has gone away, but it is too late - we are too committed at this point.  Other jobs may successfully be reading the downloaded file when another process starts re-downloading and ruining it - leading to failed plots [there are some guesses on my part here - hard to debug this scenario...].

It seems the solution here would be that once the lock is released (which means the file has been downloaded or possibly failed for other reasons) we should _check again_ if the file is there before we continue to download.  If the file miraculously has arrived (via one of the other processes) then we should abort the download and get back to treating it as a file on the system.  It seems if we simply return from gmt_download_file with no error at that point it should continue as if it was successful.

Any thoughts or experiences on this, @GenericMappingTools/core ?

PaulWessel:
See #5757 for background.  This PR addresses the solution suggested by checking after we pass the file-lock step if the file magically has been created by another process before we start downloading the file as we set out to do.  Closes #5757.

### DIFF
--- a/src/gmt_remote.c
+++ b/src/gmt_remote.c
@@ -616,6 +616,12 @@ GMT_LOCAL int gmtremote_get_url (struct GMT_CTRL *GMT, char *url, char *file, ch
 	if ((LF = gmtremote_lock_on (GMT, file)) == NULL)
 		return 1;
 
+	/* If file locking held us up as another process was downloading the same file,
+	 * then that file should now be available.  So we check again if it is before proceeding */
+
+	if (!access (file, F_OK))
+		return GMT_NOERROR;	/* Yes it was! */
+
 	/* Initialize the curl session */
 	if ((Curl = gmtremote_setup_curl (API, url, file, &urlfile, GMT_HASH_TIME_OUT)) == NULL)
 		goto unlocking1;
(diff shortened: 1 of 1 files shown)


PaulWessel:
I know maybe hard to check this since no change really for most things.  I think you have to trust me on this one.  The added code cannot hurt since for most cases it will immediately return with "file not there" and continue on to download, but in these cases where we run simultaneous jobs (**batch** and **movie**) it will really improve things to not download the same files again and again.  I did this test (I had some temporary fprintf lines to see progress):

1. Ran one gmt grdcut job in debugger and have it stop at the set lock step
2. Stepped through the lock setting as the first processes to get there.
3. Then started the same job in the terminal.  It got to that place and hang/waited.
4. In debugger, I continued the process to completion and downloaded the file
5. The waiting process then continued and printed that it found the file and aborted the download

So I think all is well.  Please approve so I can merge in and test more via #5753.

Esteban82 (approved):
Approved as requested. I couldn't try it

Issue #5763, opened 2021-09-15
batch and movie calling grdblend can lead to instability


PaulWessel:
**Description of the problem**

This is another problem specific to **batch** and **movie**: When working with tiled remote datasets the collection of tiles is sent to **grdblend** via a _GMT_Call_Module_ call.  As the blending proceeds there are places where temporary grids are created, such as 

`/tmp/grdblend_resampled_53553_5.nc`

where 53553 was the process PID and 5 is a running number (starting at 0).  The problem seems to be (again, hard to know for sure and to debug these issues) that some of the other concurrent processes launched by **movie** or **batch** may be trying to create the same files, and suddenly we see these types of messages:

`grdblend [ERROR]: File /tmp/grdblend_resampled_53553_5.nc was not found`

Clearly, one process thought it was done with its temporary file but another process had the same name and now there is chaos.  Running **movie** or **batch** with **-x**2 means we only launch one job at the time and no such error appears.  The more cores, the more likely the instability.

A relative simple solution would be not to just use the PID to make the temporary file but use the standard _mktemp_ function.  That way we get unique files on the file system and not just in some process.

meghanrjones:
Is there more to do here or should we close the issue?

PR #5764, opened 2021-09-15
Centralize creation of templates and opening of temporary files

PaulWessel:
This PR simplifies the handling of temporary files.  Before, various parts of GMT would do this manually: Set up a template that contains the XXXXX pattern, call _mktemp_, then maybe append an extension. If a temp dir exists then that was prepended.  Then we sometimes create the file, in which case there is a choice between how this is done on Windows and Linux.  Now, all this is cerntralized into two new functions:

`EXTERN_MSC int gmt_get_tempname (struct GMTAPI_CTRL *API, char *stem, char *extension, char path[]);`

This function takes a desired filename stem (or NULL, which means use "gmttemp"), possible file extension (e.g., ".nc" or NULL) and makes a unique system file name.  This file could then be used for making a temp directory, for instance.

`EXTERN_MSC FILE *gmt_create_tempfile (struct GMTAPI_CTRL *API, char *stem, char *extension, char path[]);`

This function also creates a unique name and then opens that file for writing.  This step depends on the OS as Windows uses _mktemp_ and then we must create the file with _fopen_, while Linux uses _mkstemp_ which does both in the same system function.  The difference is related to a potential race condition: Between creating the file name and creatling the file another process could have done the same and the file already exist.  This is extremely unlikely for us so I am not to concerned; this is probably more for servers with endless file i/o.  Anyway, this function abstracts us away from those detail in where it is used.

All tests pass, and this will hopefully address #5763, but this PR is good regardless by eliminating duplicated codes.

PS. Apologies for lots of nonsense changes in gmt_plot.c - it has to do with my editor first replacing tabs with four spaces and then me undoing it, which also removed four spaces in places where they are not indentations...

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -13960,34 +13960,15 @@ GMT_LOCAL int gmtinit_get_region_from_data (struct GMTAPI_CTRL *API, int family,
 				char *file = NULL;
 				void *content = NULL;
 				size_t n_read = 0;
-#ifndef _WIN32
-				int fd = 0;
-#endif
 
 				GMT_Report (API, GMT_MSG_DEBUG, "gmtinit_get_region_from_data: Must save stdin to a temporary file.\n");
-				if (API->tmp_dir)			/* Have a recognized temp directory */
-					snprintf (tmpfile, PATH_MAX, "%s/", API->tmp_dir);
-				strcat (tmpfile, "gmt_saved_stdin.XXXXXX");	/* The XXXXXX will be replaced by mktemp */
-#ifdef _WIN32
-				if ((file = mktemp (tmpfile)) == NULL) {
-					GMT_Report (API, GMT_MSG_ERROR, "gmtinit_get_region_from_data: Could not create temporary file name.\n");
-					return GMT_RUNTIME_ERROR;
-				}
-				if ((fp = fopen (file, API->GMT->current.io.w_mode)) == NULL) {
-					GMT_Report (API, GMT_MSG_ERROR, "gmtinit_get_region_from_data: Could not open temporary file %s.\n", file);
-					return GMT_RUNTIME_ERROR;
-				}
-#else
-				if ((fd = mkstemp (tmpfile)) == -1) {
-					GMT_Report (API, GMT_MSG_ERROR, "gmtinit_get_region_from_data: Could not create and open temporary file %s.\n", tmpfile);
+
+				if ((fp = gmt_create_tempfile (API, "gmt_saved_stdin", NULL, tmpfile)) == NULL) {	/* Not good... */
+					GMT_Report (API, GMT_MSG_ERROR, "gmtinit_get_region_from_data: Could not create and open temporary file name %s.\n", tmpfile);
 					return GMT_RUNTIME_ERROR;
 				}
 				file = tmpfile;
-				if ((fp = fdopen (fd, API->GMT->current.io.w_mode)) == NULL) {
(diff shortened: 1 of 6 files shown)


PaulWessel:
I think because all tests still passes that I did this correctly.  If I made an error then all **grdblend**-related and some other scripts would fail badly.  So please have a check of what was done in gmt_support.c and hopefully approve.  I am not sure how I can test this PR in another PR (#5757) without first merging this general one.

meghanrjones:
Do you think all the places where there were four spaces should be tabs? If not, I think we should find a way to roll back the massive diff in gmt_plot.c. To me, it looks like some of the changes make the code less readable (example below). 

**master branch**

image

**tempfile-fix branch**

image

PaulWessel:
OK, after zooms, including with you this morning...

PaulWessel:
OK, I think I got most of it - does this look more reasonable?

meghanrjones:
> OK, I think I got most of it - does this look more reasonable?

Looks much better, thanks!

I tried testing this branch with the example from the batch docs and received a segfault. Does it work for you?

```
cat << EOF > pre.sh
gmt begin
    gmt coast -E=EU+l > countries.txt
gmt end
EOF
cat << EOF > main.sh
gmt begin \${BATCH_NAME} pdf
    gmt coast -R\${BATCH_WORD0}+r2 -JQ10c -Glightgray -Slightblue -B -B+t"\${BATCH_WORD1}" -E\${BATCH_WORD0}+gred+p0.5p
    echo \${BATCH_WORD0} | gmt text -F+f16p+jTL+cTL -Gwhite -W1p
gmt end
EOF
cat << EOF > post.sh
gs -dQUIET -dNOPAUSE -sDEVICE=pdfwrite -sOUTPUTFILE=\${BATCH_PREFIX}.pdf -dBATCH \${BATCH_PREFIX}_*.pdf
rm -f \${BATCH_PREFIX}_*.pdf
EOF
gmt batch main.sh -Sbpre.sh -Sfpost.sh -Tcountries.txt+w"\t" -Ncountries -V -W -Zs
```

Esteban82:
> I tried testing this branch with the example from the batch docs and received a segfault. Does it work for you?

Mmmm the other day I test this same example and also got a segfault and I wans't in this branch. I will confirm later.

Esteban82:
I got this error in `remote-auto-resolutuion` and `master` branchs.

```
/usr/local/lib/libgmt.so.6(GMT_batch+0x219f)[0x7fb4b21d887f]
[0x0]
Stack backtrace:
/usr/local/lib/libgmt.so.6(sig_handler+0x2ad)[0x7fb4b1ff663d]
/lib/x86_64-linux-gnu/libc.so.6(+0x46210)[0x7fb4b1de8210]
/usr/local/lib/libgmt.so.6(GMT_batch+0x219f)[0x7fb4b21d887f]
/usr/local/lib/libgmt.so.6(GMT_Call_Module+0x3bf)[0x7fb4b201700f]
gmt(main+0x26a)[0x5570bfec35ea]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7fb4b1dc90b3]
gmt(_start+0x2e)[0x5570bfec46be]
```

PaulWessel:
Working on it.

PaulWessel:
I believe this branch is OK now.  I just ran anim08.sh as one that uses lots of earth_relief and grdblend.  Also, the batch example works fine.  Let me know if there are other concerns or if it can be approved.

meghanrjones (approved):

PR #5765, opened 2021-09-15
Consistently describe output grid name and modifiers in rst docs

meghanrjones:

Equivalent of https://github.com/GenericMappingTools/gmt/pull/5612 for html docs build from ReST.

One point to note - in `gmt_outgrid_syntax` and the updated syntax for explain_grd_inout.rst, I updated "Note: Any offset is added after any scaling" to "Note: Any offset is added before any scaling". Previously, the order for both input and output was documented as the same which would not preserve values round-trip. I think probably just copy/paste issue.

Fixes #

**Reminders**


PaulWessel (approved):
Nice, this really cleans up a lot!

PR #5768, opened 2021-09-16
Simplify checking to determine if certain modifiers are given

PaulWessel:
Lots of places in GMT, especially where we have modernized parsing, still need to check if we get old deprecated or new syntax.  We do this by looking if any of the modern valid modifiers are present in the argument.  This can lead to long lines of repetitive code like

`if (strstr (opt->arg, "+a") || strstr (opt->arg, "+f") || strstr (opt->arg, "+j") || strstr (opt->arg, "+l") || strstr (opt->arg, "+m") || strstr (opt->arg, "+o") || strstr (opt->arg, "+s")) {`

which is ugly and subject to mistakes (in fact I fixed a few where new modifiers had been added but the checks had not been updated, such as **pscoast -E +c**|**C**). This PR introduces a new function that does the same thing:

`if (gmt_found_modifier (GMT, opt->arg, "afjlmos")) {`

This PR searched for such instances and replaced these chains with similar function calls to simplify the code.  All tests still pass.


meghanrjones (approved):
Very nice!

PR #5772, opened 2021-09-16
Raise error if dims, range, and inc are all not given properly

meghanrjones:

This PR adds an error if an equidistant vector arrangement is not given and dim is also not provided. While I am confident this is the right place to raise an error for the case from https://github.com/GenericMappingTools/pygmt/issues/1483, it would be good to get feedback about whether this is the proper error to raise out of the options in GMT_api_error_code.

Fixes https://github.com/GenericMappingTools/pygmt/issues/1483

**Reminders**


PaulWessel (approved):
Yep, that should do it.

PR #5773, opened 2021-09-16
GMT need -fs to read just text

PaulWessel:
See the problems with the batch example at the end of #5764. I missed an extra check before I declare no content.  But user needs to give **-fs** to make this work since GMT in general will not process just text records.  The places where we may end up reading text only are in **movie** and **batch**.

### DIFF
--- a/doc/rst/source/batch.rst
+++ b/doc/rst/source/batch.rst
@@ -299,7 +299,7 @@ we combine all the individual PDFs into a single PDF file and delete the individ
     gs -dQUIET -dNOPAUSE -sDEVICE=pdfwrite -sOUTPUTFILE=\${BATCH_PREFIX}.pdf -dBATCH \${BATCH_PREFIX}_*.pdf
     rm -f \${BATCH_PREFIX}_*.pdf
     EOF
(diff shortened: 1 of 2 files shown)


PaulWessel:
An alternative would be to let _gmtio_examine_current_record_ know that if module is batch or movie to look harder at that record and if it looks like text then be OK with it and implicitly set **-fs**.  That way we do not have to break/update the batch example.

meghanrjones:
Updated script works, but is there a way to give an error rather than a segfault if -fs is forgotten?

meghanrjones:
> An alternative would be to let _gmtio_examine_current_record_ know that if module is batch or movie to look harder at that record and if it looks like text then be OK with it and implicitly set **-fs**. That way we do not have to break/update the batch example.

Didn't see this comment. This would work too instead of an error.

Esteban82 (approved):

Esteban82:
I add -fs in the script and I got the pdf well.

joa-quim:
> Updated script works, but is there a way to give an error rather than a segfault if -fs is forgotten?

Programs should never crash. A crash is a kind of bug by itself.

PaulWessel:
Restored old behavior for batch and movie only since those are the only ones reading text.

Esteban82:
Now it works without -fs.

PR #5775, opened 2021-09-17
Allow colorbar length and/or width to be given in percentages

### LINKED ISSUE #5771: colorbar length as a percentage relative to a map side

Hello dear PyGMT community,

I would like to request a feature that will allow to set a colorbar's length to any desired percentage relative to a map side. Currently, it is set to 80 % by default, and a user needs to provide a specific value with `+w` modifier to change the colorbar's length. It would be very nice though if one could set any arbitrary number in percents of a map side, like 100 % as sometimes is asked by reviewers during the publication process.

This request is the follow-up to the [discussion in the GMT community forums](https://forum.generic-mapping-tools.org/t/pygmt-colorbar-length-settings/2071).

Thank you.

**Are you willing to help implement and maintain this feature?** No, I don't think I have enough skills so far

PaulWessel:
If either of these are given in percentages then their values replace the default percentage values of 80% and 5% for length and width, respectively.  I have added this information to the usage and documentation, and added a test script that tries various combination of percentages and actual dimensions:

```
gmt begin scale_percent ps
        gmt makecpt -Cturbo -T0/10
        gmt basemap -R0/10/0/10 -JX15c -B
        gmt colorbar -DJTC+w40%/15% -B
        gmt colorbar -DJML+w9c/5% -B
        gmt colorbar -DJMR+w80%/0.6c -B
        gmt colorbar -DJBC+w70% -B
gmt end show
```

which yields

scale_percent

Closes #5771.

### DIFF
--- a/doc/rst/source/colorbar.rst
+++ b/doc/rst/source/colorbar.rst
@@ -115,7 +115,9 @@ Optional Arguments
     away from the map frame.
     However, you can override any of these with these modifiers:
     Append **+w** followed by the *length* and *width* of the color bar.  If *width* is not
-    specified then it is set to 4% of the given *length*.
+    specified then it is set to 4% of the given *length*. If *length* is not given then it defaults
+    to 80% of the corresponding map side dimension.  If either *length* or *width* end with % then
+    those percentages are used instead to set the dimensions.
     Give a negative *length* to reverse the scale bar, or append **+r**. Append **+h** to get a
     horizontal scale [Default is vertical (**+v**)].
     By default, the anchor point on the scale is assumed to be the bottom left corner (BL), but this
--- a/doc/rst/source/colorbar.rst
+++ b/doc/rst/source/colorbar.rst
@@ -117,7 +117,8 @@ Optional Arguments
     Append **+w** followed by the *length* and *width* of the color bar.  If *width* is not
     specified then it is set to 4% of the given *length*. If *length* is not given then it defaults
     to 80% of the corresponding map side dimension.  If either *length* or *width* end with % then
-    those percentages are used instead to set the dimensions.
+    those percentages are used instead to set the dimensions, where *width* is defined as a
+    percentage of the colorbar *length*.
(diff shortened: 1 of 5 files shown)


Esteban82:
I have a problem with `gmt colorbar -DJML+w9c/5% -B`


scale_percent

Esteban82:
I get the same width with 1% or 100%.

```
gmt begin scale_percent png
        gmt makecpt -Cturbo -T0/10
        gmt basemap -R0/10/0/10 -JX2/10c
        gmt colorbar -DjML+w9c/1% -B
        gmt colorbar -DjMR+w9c/100% -B
gmt end show
```


scale_percent

PaulWessel:
Strange that I dont get that.  So it happens when the first (length) argument is a dimension and the second (width) is a percentage? Is that the only combination that fails?

Esteban82:
Yes, only that combination.

PaulWessel:
I am testing anim08 on the tempfile_fix branch so it will take an hour before I can switch...

PaulWessel:
OK, should be better now - please try again.

Esteban82 (approved):
Now it works.

doc/rst/source/colorbar.rst
```diff
@@ -115,7 +115,9 @@ Optional Arguments
     away from the map frame.
     However, you can override any of these with these modifiers:
     Append **+w** followed by the *length* and *width* of the color bar.  If *width* is not
-    specified then it is set to 4% of the given *length*.
+    specified then it is set to 4% of the given *length*. If *length* is not given then it defaults
+    to 80% of the corresponding map side dimension.  If either *length* or *width* end with % then
+    those percentages are used instead to set the dimensions.
```
meghanrjones, line 120:
```suggestion
    those percentages are used instead to set the dimensions, where *width* is defined as a
    percentage of the colorbar *length*.
```

meghanrjones (approved):
Very nice!

PR #5779, opened 2021-09-17
Clarify that -R is not always required in modern mode

meghanrjones:

The **-R** parameter is listed as required for most GMT modules even if it can be automatically set by GMT. This PR adds some clarifying information to the docs.

So far, I have only added the explanation to `plot` for feedback. Here is a capture of the updated description:

updated -R description for plot

Fixes #

**Reminders**


PaulWessel (commented):
Looks good.  As for style I think we usually add a **Note**: before the clarifying sentence, i.e., your first "If ..."

PaulWessel (approved):

PR #5782, opened 2021-09-19
Let mapproject report rectangular region that covers an obliquely defined region

PaulWessel:
When we specify an oblique region via an azimuthal perspective projection or the region via **-R**_w/s/e/n_**+r**, it is hard to know what the corresponding rectangular region is, which can be very useful piece of information for advanced scripting.  Here we add modifier **r**  to **-W** that will report this region; use uppercase **R** to get a "-Rw/e/s/n" string as trailing text instead.  The man page has an example of this.  Here are other examples:

```
gmt mapproject -Rg -JG10/20/1900/0/0/0/170/180/3i -WR
-R-31.2072906633/51.2072895162/-13.2762303233/49.3230415777
-R-31.2072906633/51.2072895162/-13.2762303233/49.3230415777

gmt mapproject -R270/20/305/25+r -JOc280/25.5/22/69/2c -Wr
270	305	9.48892505833	35.6071567812
```

I do not think we need to add a specific test script for this feature since it is only reflecting internal knowledge.

### DIFF
--- a/doc/rst/source/mapproject.rst
+++ b/doc/rst/source/mapproject.rst
@@ -27,7 +27,7 @@ Synopsis
 [ |-S| ]
 [ |-T|\ [**h**]\ *from*\ [/*to*] ]
 [ |SYN_OPT-V| ]
-[ |-W|\ [**g**\|\ **h**\|\ **j**\|\ **n**\|\ **w**\|\ **x**] ]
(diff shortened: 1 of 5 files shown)


meghanrjones (approved):
Looks good. Note that this is available now regardless of the projection specified, do you want to limit to only oblique projections?

PaulWessel:
No, I think it is OK that it works for any projection.  Even as -R0/30/0/30 -JM6i will just return 0/30/0/30.

PR #5786, opened 2021-09-20
Avoid setting an oblique mercator flag if projection is not oblique Mercator

PaulWessel:
There was a check after parsing **+du**|**h**|**w** that only would apply to **-JO** but there was no check to make sure it was that projection.


Esteban82 (approved):

Esteban82:
I didn't squash and merge due to the Vercel error.

meghanrjones:
> I didn't squash and merge due to the Vercel error.

I think this was a dependency conflict that is now resolved, so it shouldn't be a problem moving forward.

PR #5789, opened 2021-09-20
Skip gdal scan images

PaulWessel:

This PR skips the scanning for z min/max when these are not known and the dataset appears to be an image (byte storage model).  This will avoid such scans when we only want the grid header.


PaulWessel:
OK, forgot to remove the check on nbands in the code, now done.

joa-quim (approved):

Issue #5778, opened 2021-09-17
Upgrade Anim04


Esteban82:
I am upgrading anim04. Below is the animation based on the earth_relief grid. I should try a higher resolution but I had some issues (that I think it will be solved with Paul PR). 

I have two doubts/questions:

1. I think it would be better to use a satellite image. Maybe the black Marble (see the image below). I think we could use it a bit more. 
2. I don't understand well the General Prospective projection. In the black marble image, I reduce it and I see that I get a rectangular image. In the animation the lower part was outside the canvas. Is it ok that way (like in the script)? Or what parameters should I use for -JG?

https://user-images.githubusercontent.com/17908951/133854154-881ec849-72d8-4d74-9822-4e195b661f47.mp4

![anim04_sat](https://user-images.githubusercontent.com/17908951/133855084-1297715a-b827-486a-aef2-73acd4ecd492.png)

Full script:

```
if [ $# -eq 1 ]; then	# Just make master PostScript frame 27 (az = 206)
	opt="-M27,png -Fmp4"
else	# Make animated GIF, infinitely looping
	opt="-M27,png"
fi

# 1. Create files needed in the loop
cat << 'EOF' > pre.sh
gmt begin
	gmt project -C-73.8333/40.75 -E-80.133/25.75 -G10 -Q > flight_path.txt
	gmt grdcut @earth_relief_30s -R-100/-70/10/50 -GUSEast_Coast.nc
	gmt grdgradient USEast_Coast.nc -A90 -Nt1 -Gint_US.nc
	gmt makecpt -Cglobe -H > globe_US.cpt
gmt end
EOF
# 2. Set up the main frame script
cat << 'EOF' > main.sh
gmt begin
	gmt grdimage -JG${MOVIE_COL0}/${MOVIE_COL1}/160/210/55/0/36/34/${MOVIE_WIDTH}+du -Rg USEast_Coast.nc -Iint_US.nc -Cglobe_US.cpt -Xc -Y-3c
	gmt plot -W1p flight_path.txt
gmt end
EOF
# 3. Run the movie
gmt movie main.sh -Chd -Nanim04 -Tflight_path.txt -Sbpre.sh -Zs -H2 -Lf+o0.25c+f14p,Helvetica-Bold $opt
```

PaulWessel:
Thanks @Esteban82.  I am working on a PR to replace the horrible -JG syntax with

[...]

PR #5794, opened 2021-09-21
Update anim04.sh

Esteban82:
Upgrade anim04. See #5778 for discussion.

### DIFF
--- a/doc/examples/anim04/anim04.sh
+++ b/doc/examples/anim04/anim04.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 #               GMT ANIMATION 04
 #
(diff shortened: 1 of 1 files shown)


PaulWessel:
We still have an issue with images, so please do not merge yet.  While we correctly determine the subregion w/e/s/n for a frame, there is still a difference:

1. If @earth_relief then we build the list of tiles needed for that region.  It is read in by **grdimage** and the grid has that w/e/s/n
2. If @earth_night then there are no tiles, just a giant image, and the **grdimage** code has not been updated yet to pass the subregion values determined inside the belly of GMT.  I am working on how to make it available so we don't melt my computer reading 24 copies of earth_night_30s.tiff...

PR #5795, opened 2021-09-21
Complete documentation of genper and fix special cases

PaulWessel:
The **-JG** general perspective projection had some old special cases using poorly documented syntax and some not documented at all.  This PR fixes this.  The changes pertains to how _altitude_ can be set.  There are now four documented ways

- Give altitude in km above sea level [infinity]
- Give radius in km from center of the Earth instead (append **r**)
- Give radius in Earth radii from center of the Earth instead (append **R**)
- Give altitude as **g** for the altitude of the geosynchronous orbit (~35700 km; now documented as to what the code does)

The old deprecated ways (in backwards compatibility mode) involved giving a negative altitude less that -10 to mean radius in km and values less than 10 to mean Earth radii.  The latter restriction would mean one could not select an actual altitude of only 10 km, for instance.

I also removed the unused variable _g_auto_twist_ which was set with an undocumented letter but there is no code that actually uses this variable.


doc/rst/source/cookbook/map-projections.rst
```diff
@@ -404,7 +404,7 @@ Perspective projection (**-Jg** **-JG**)
 
 **Syntax**
 
-    **-Jg**\|\ **G**\ *lon0/lat0*\ */*\ *scale*\|\ *width*\ [**+a**\ *azimuth*][**+t**\ *tilt*][**+v**\ *vwidth/vheight*][**+w**\ *twist*][**+z**\ *altitude*]
+    **-Jg**\|\ **G**\ *lon0/lat0*\ */*\ *scale*\|\ *width*\ [**+a**\ *azimuth*][**+t**\ *tilt*][**+v**\ *vwidth/vheight*][**+w**\ *twist*][**+z**\ *altitude*\ [**r**\|\ **R**]\|\ **g**]
```
meghanrjones, line 407:
```suggestion
    **-Jg**\|\ **G**\ *lon0/lat0*\ */*\ *scale*\|\ *width*\ [**+a**\ *azimuth*][**+t**\ *tilt*][**+v**\ *vwidth/vheight*][**+w**\ *twist*][**+z**\ *altitude*\ [**r**\|\ **R**\|\ **g**]]
```
Is *altitude* **Rg** allowed? My interpretation is that it should be **r** or **R** or **g**.

PaulWessel: No, the four ways are _altitude_, _altitude_**r**, _altitude_**R**, and **g**.  I thought that is how one would interepret that syntax, no?

meghanrjones: OK, I misread.

meghanrjones (approved):

PR #5796, opened 2021-09-22
Ensure the auto-resolution selects a good resolution

PaulWessel:
This PR improves some aspects of the remote grid auto-resolution settings.  These are the improvements:

1. We now generate a list of the available resolutions and registrations for the selected data set family (e.g., earth_age) instead of using a generic list that may contain resolutions not actually available.  This fixes a bug in the selection.
2. Instead of finding the one closest to the desired resolution (which might be one of a lower resolution that requested), this PR finds the one with the closest, but better resolution (if available).  For instance, if the estimates nodes per degree is 72 then we are in between 60 (01m) and 120 (30s).  The initial algorithm would pick 01m but now we would pick 30s.
3. We now check if **--GMT_GRAPHICS_DPU**=dpu is given on the same command since the calculations take place before _GMT_Parse_Common_ has been called so we must parse it separately before computing the region.
4. For the text-format selection (**-D+t**) we now format the **-R** and **-I** strings using the ddd:mm:ss format if geographic and if the increment is in seconds, minutes, or degrees. This gives more compact strings for those cases.

The autores.sh script is likely to be a bit unstable until we finalize the algorithm, but has been updated based on the new scheme.  All other tests still passed since this only affect the -D and any use of non-resolution remote grid (not used in tests yet but being checked in new animations which is driving this).


Esteban82:
When I run autores.sh I got. Is it a problem with GMT or something else?

```
grdcut [NOTICE]: Remote data courtesy of GMT data server oceania [http://oceania.generic-mapping-tools.org]
grdcut [NOTICE]: Earth Seafloor Crustal Age (GTS12 timescale) at 1x1 arc minutes from EarthByte [Seton et al., 2020].
grdcut [NOTICE]:   -> Download grid file [188M]: earth_age_01m_g/
grdcut [ERROR]: Libcurl Error: Failed writing received data to disk/application
grdcut [WARNING]: You can turn remote file download off by setting GMT_DATA_UPDATE_INTERVAL to "off"
```

PaulWessel:
Something else, the file is there and download works for me.  Space problems?

joa-quim:
This had happened to me, in a non-reproducible way (I think), when I was not connected to internet but file(s) was already downloaded.

anbj:
Technically not related to this PR, but I sometime get similar issues. Hard to reproduce.

E.g. 90% of the time the script I'm using will run fine, but suddenly, I get:

```
grdimage [ERROR]: Unable to obtain remote file @earth_relief_02m
```

When using the auto-resolution-branch, I also got an error _some of the time_ (I think), in pseudoerror:

```
libcurl did not find earth_relief
```

Esteban82:
I had more than 136 Gb free. And when I make the animation, I have no problem. 

https://user-images.githubusercontent.com/17908951/134670685-15b98620-d17e-4170-abb3-a4474249d5fe.mp4

anbj:
Shouldn't this work?

```
$ gmt grdimage @earth_relief -RNO -JM15c -Atest.tif=gdal:GTiff
gmt [ERROR]: Except for grdcut, cannot request automatic remote grid resolution if not plotting [@earth_relief]
```

PaulWessel:
> Shouldn't this work?
> 
> ```
> $ gmt grdimage @earth_relief -RNO -JM15c -Atest.tif=gdal:GTiff
> gmt [ERROR]: Except for grdcut, cannot request automatic remote grid resolution if not plotting [@earth_relief]
> ```

Probably, will have to relax that check.

PaulWessel:
> gmt grdimage @earth_relief -RNO -JM15c -Atest.tif=gdal:GTiff

Works now, thanks.

PaulWessel:
I need someone to approve this one otherwise I must do it myself.  I need this capability to work further on the **grdcut** for images and **grdimage** without reading the global huge files.  Remember, this PR _improves_ things, even if it may not be perfect yet.

Esteban82 (approved):
Approve as requested.

meghanrjones:
> I need someone to approve this one otherwise I must do it myself. I need this capability to work further on the **grdcut** for images and **grdimage** without reading the global huge files. Remember, this PR _improves_ things, even if it may not be perfect yet. All tests pass for me.

Sorry, this PR slipped the radar. Thanks @Esteban82 for reviewing.

PR #5798, opened 2021-09-22
Implement image subsets within GMT and via grdcut

PaulWessel:
As a way to debug and implement subsets of images, I am letting **gmt grdcut -I** get the ability to cut a subset out of an _image_.  This also makes it easier to debug this feature which is still not working in GMT (and affects **grdimage**, for instance, where the entire image must be read in, leading to excessive run times and memory requirements for high-resolution global grids).  Derives bits and pieces from stalled PR #4004. Note: I also updated the **KEYS** and _GMT_Encode_Options_ so it should be possible to receive the image back from the externals as well.

Trying this command:

`gmt grdcut @earth_day_01m -R0/10/0/10 -I -Gcut.png`

Yields an image with the correct region, increment meta data but likely bad info on the order or RGB, yet the pixels are in the right layout (since Africa outline etc is correct) and the region matches that of a pscoast plot:

cut

Hoping @joa-quim can advise where to make changes in this PR.

Note: This PR allocates memory only for the subset and pad, and currently grdimage commands like the one below will crash in the PR because it assumes a full image.  This will be dealt with once we correctly can read subsets of images.  However, in master, this comand

`gmt grdimage @earth_day_01m -R0/10/0/10  -JQ5.07614213198c -png out`

yields this (same-size) image (for comparison with above):

out

Clearly, this is an incorrect region despite having the RGB in the right order, so this seems to be a bug in master.  I suggest we work on getting the subset to work correctly via **grdcut** and then the rest will be simpler.


meghanrjones:
Not sure that I understand your last post - are you saying to ignore the problems with the colors?

PaulWessel:
> Not sure that I understand your last post - are you saying to ignore the problems with the colors?

No, there are two things:

1. In this PR, looks like **grdimage** with an image crashes due to the changes we are trying in the i/o for images.  This will go away when we are finished.
2. I did not expect master to give the wrong image but I think @federico pointed out some problems with grdimage using the earth_day stuff earlier so maybe the same.  I don't really want to stop this PR to fix that in master since the real solution is the one we will get from this PR.

PaulWessel:
Hi @joa-quim, towards the end of gmtlib_read_image in gmt_grdio.c there is

```
if (to_gdalread->O.mem_layout[0]) 	/* If a different mem_layout request was applied in gmt_gdalread than we must update */
		gmt_strncpy(I->header->mem_layout, to_gdalread->O.mem_layout, 4);
```

In our case, the image I has TRBa which the to_gdalread version says BRPa, so we copy it over.  This gives the above wrong coloring.  Should it be another string?

PaulWessel:
OK, this is an indexed image so not wrong R/G/B order, but upon comparing the image produced with grdcut to the equivalent derived via 

`gdal_translate -srcwin 10800 4800 600 600 ~/.gmt/server/earth/earth_day/earth_day_01m_p.tif gdal.tif`

I find that the two colortables are just very different.  How can we get that wrong?

PaulWessel:
Stepping through to see what happens to the indexed colors.

Line 539 gmt_gdalread.c: We are getting the colors one by one from GDAL and storing in the Ctrl->ColorMap 4x256 array.  For each color i = 0, 1, ...you do this

```
GDALGetColorEntryAsRGB(hTable, i, &sEntry);
Ctrl->ColorMap[j++] = sEntry.c1;
Ctrl->ColorMap[j++] = sEntry.c2;
Ctrl->ColorMap[j++] = sEntry.c3;
Ctrl->ColorMap[j++] = sEntry.c4;
```
That means the RGBA for the first entry in the colormap is the first 4 values in Ctrl->ColorMap.  For my example that is 224,164,84,255.

Then in Line 177 (gmt_gdalwrite.c) you place these values into another array, this time 4x256 float:

```
for (k = 0; k < n_colors; k++) {
	to_GDALW->C.cpt[k] = I->colormap[k] / 255.0;
	to_GDALW->C.cpt[k + n_colors]   = gmt_M_is255(I->colormap[k + n_colors]);
	to_GDALW->C.cpt[k + n_colors*2] = gmt_M_is255(I->colormap[k + n_colors*2]);
}
```

So here, let k = 0.  Then you get the red (colormap[0]) but not the green from the first (since that would be colormap[1], divide by 255 and store it in the same place in the cpt array.  I guess this is OK but not sure why you do it this way since it does not follow the structure of colormap.

Finally, just before we call GDAL to write (line 375) you do this:

```
for (i = 0; i < nColors; i++) {
	sEntry.c1 = (short)(gmt_M_s255(ptr[i]));
	sEntry.c2 = (short)(gmt_M_s255(ptr[i+nColors]));
	sEntry.c3 = (short)(gmt_M_s255(ptr[i+2*nColors]));
	sEntry.c4 = (short)255;
	GDALSetColorEntry(hColorTable, i, &sEntry);
}
```

Again, try i = 0.  This picks supposedly an r, g, b from the ptr (which is the cpt above) and then loads up the sEntry structure. For i = 0 these become 224,72,72,255. So no longer 224,164,84,255.  I think that is why the image looks junk.  Please let me know what you were trying to do here, @joa-quim

joa-quim:
I think there must be missing some step in the middle. While the first case stores the color matrix as row-major the other two are accessing it as column major. The column major ordering is quite likely because when that part was written the main idea was to interface with the MEX.

PaulWessel:
OK, I can ensure the ColorMap is always stored internally in column order throughout and see how that goes.

PaulWessel:
There are other places in GMT where the assumtion is that colormap is rows, e.g. in gmtapi_expand_index_image we do

```
		for (node = k = 0; node < h->size; node++) {	/* For all pixels, including the pad */
			index = I->data[node];	/* Pixel index into color table */
			start_c = index * 4;	/* Start of the r,g,b entry in the color table for this index */
			for (c = 0; c < 3; c++, k++) data[k] = I->colormap[start_c+c];	/* Place r,g,b in separate bands */
		}
```

I guess I will need to go through all of GMT where ->colormap is referenced and ensure it is set and used as columns.  It is not that many places I think.

PaulWessel:
OK, I have gotten **grdcut -I** to work.  Not tested extensively on different images though, just earth_day, but output matches _gdal_translate_. This is reading and returning an image with the pad (as expected for instance in grdimage) but when we write we strip off the pad.

` gmt grdcut @earth_day_01m -R-5/10/0/10 -I -Gcut.tif`

Of course, it cannot handle periodic boundary at ±180:

```
gmt grdcut @earth_day_01m -R170/190/0/10 -I -Gcut.tif
grdcut [ERROR]: Computed -srcwin falls outside raster size of 21600x10800.
grdcut [ERROR]: ERROR reading image with gdalread
```.

That may require the use of 

`gmt_M_err_pass (GMT, gmt_grd_prep_io (GMT, header, wesn, &width_in, &height_in, &first_col, &last_col, &first_row, &last_row, &actual_row), HH->name);`

and reading GDAL twice.  Not sure what to do about this.

**Comment by joa-quim** [comment]
The read sub-regions part originated long ago in Mirone and either the number of options were more limited, or more likely my knowledge. It seems we should be using **-a_ullr** instead of **-srcwin** since former let us specify the coordinates in geogs while latter uses only row/col. GDAL can also deal with round Earth but it often needs to be told to do so.

**Comment by meghanrjones** [comment]
Would it be possible to get this to work without the **-I** argument being required, similar to how grdimage accepts *grid* | *image* without any extra user input?

**Comment by meghanrjones** [comment]
> The read sub-regions part originated long ago in Mirone and either the number of options were more limited, or more likely my knowledge. It seems we should be using **-a_ullr** instead of **-srcwin** since former let us specify the coordinates in geogs while latter uses only row/col. GDAL can also deal with round Earth but it often needs to be told to do so.

This branch also has a lot of test failures caused by problems similar to 'grdimage [ERROR]: Computed -srcwin falls outside raster size of ...', which could be caused by this.

**Comment by PaulWessel** [comment]
Yes, it looks like we need to upgrade the gdalread to specify geographic coordinates instead when the image is global.  Can @joa-quim get us started on that, i.e., how to pass different arguments with coordinates instead?

**Comment by PaulWessel** [comment]
A cheap and ugly fix that would avoid the periodic boundary failures would be to revert to read the whole image in longitude (cut cut in latitude) when geographic and periodic, else set the longitude limits as well.  I.e., until @joa-quim and I can figure out how to use GDAL to do the thing for us.

**Comment by joa-quim** [comment]
You talk as if I new all that stuff. The **-a_ullr** doesn't work after all. It only sets new limits but image is the same. i.e. this does nothing good
```
gdal_translate -a_ullr 170 45 190 0 earth_day_01d_p.tif lixo.tiff
```
There is a ``-co CENTER_LONG=180`` but apparently only for OGR and ``gdalwarp``. Can't find nothing for ``gdal_translate``.

**Comment by PaulWessel** [comment]
Maybe my temporary plan is not so bad after all.  It will be a big improvement in speed for almost all cases except very large areas and hight resolutions.

**Comment by PaulWessel** [comment]
Ping @meghanrjones, @Esteban82, @joa-quim, @seisman: This PR is now ready for serious review.  It has grown a bit from the initial attempt but should now be acceptable.  Here are the highlights:

1. gmt_gdalread.c: Implemented image subsets so that we no longer have to read the entire image to plot it (which gets excessive for high-resolution global grids).  This involved passing the region in, setting the padding (similar to grids), ensuring we do not use the region if the image has no geographic region information, and handling the situation when the desired subset straddles the periodic boundary, resulting in two separate image chunks being read into the same buffer. Because pad is now added I am able to set data BCs when we cut a subset out of a larger image (as we do for grids).
2. gmt_gdalwrite.c: Now strips off the image pad before writing an image. The non-file output options (i.e., to memory) may need further test cases and debugging since those have not been exercised by GMT I think.
3. grdcut.c: The temporary **-I** is now gone and **grdcut** detects if given a grid or image.  It can apply the same operators to both grids and images, except for **-Z** which requires a grid.
4. gmt_support.c: The boundary condition settings for images (_gmtlib_image_BC_set_) was based on the corresponding one for grids (_gmt_grd_BC_set_) got a bunch of updates, many of them cosmetic (call the image variable I not G, shorten expressions by using img = I->data).  The biggest change was to examine the results of the various boundary conditions.  Because these are essentially derivatives forced to be zero along the borders, they can amplify noise in the data, as also happens with grids.  For instance, to have d2/dx2 = 0 along w or e, we do things like `out - 2*on + in = 0` and solve for `out = 2*on - in`.  This allows us to set the image values in the pad. Ah, but the pad is part of the image and is an unsigned char, yet the results of the BCs often fall outside 0-255.  In fact, here is a plot of the values of computed outside image nodes from the tests involving images that exceed 0-255:

BC

So before, _gmtlib_image_BC_set_ would happily compute 357 and try to store that in an unsigned char, resulting in truncation to 101.  Later, if the image needs resampling by the bicubic/bilinear operator we are using incorrect values in the pads, resulting in incorrect  resampled image values near the borders.

Because _gmtapi_import_image_ now uses **-R** I had to make a few changes where we read in an image but the **-R** is _not_ meant as a subset.  This happens in **grdimage** (e.g., when we read some dumb PNG and want it to take on the **-R** bounds), **psimage** (where **-R** and **-J** are used for projection setup purposes, never for image subset), and when reading CPTs with patterns (again, any **-R** has to do with projection, not subset).  With those implementations I get three failures: marbles.sh, png_image.sh, and mixcompose.sh.  These look fine to me but instead of updating them I thought it would be good for you to run the test and see for yourself.  Note the differences are mostly due to the fixed BC scheme as in the past this would not clip the values outside 0-255, just throw out the higher bits.

Once you agree that these three are fine, please approve and I will update the PS before merging. Or let me know if you find something odd.

**Comment by Esteban82** [comment]
Is this suppose to solve #4003 ? I try the following script. For some resolutions, it worked. But failed for other. 
Are you wokring on this @PaulWessel ? I try it before the "update grdcut.c" commit. 


Failed: 01d, 20m. 06m, 05m, 03m, 02m
Ok: 30m, 15m, 10m, 04m, 01m


```
RES=01m
gmt begin BlueMarble_Intens png
	gmt grdgradient @earth_relief_$RES -Nt1.00 -A45 -GIntens.nc -R-85/-54/9/26
	gmt grdimage @earth_day_$RES -IIntens.nc -Baf -JM10c 
gmt end show

```

BlueMarble_Intens

**Comment by PaulWessel** [comment]
I see the 01d crashes.  Will have a look - I did not test specifically with intensity.  We may turn your example into a test once it is working.

**Comment by Esteban82** [comment]
Now I try it after the commit.

These are the errors (indicated by resolution of the grid/image):

02m, 03m and 01d: 
munmap_chunk(): invalid pointer
image.sh: línea 6: 37164 Abortado                (`core' generado) gmt grdimage @earth_day_$RES -IIntens.nc -Baf -JM10c


05m and 06m: 
free(): invalid next size (normal)
image.sh: línea 6: 37194 Abortado                (`core' generado) gmt grdimage @earth_day_$RES -IIntens.nc -Baf -JM10c


20m:
/lib/x86_64-linux-gnu/libc.so.6(+0x9b941)[0x7f746cc9b941]
[0x0]
Stack backtrace:
/usr/local/lib/libgmt.so.6(sig_handler+0x2ad)[0x7f746ce5475d]
/lib/x86_64-linux-gnu/libc.so.6(+0x46210)[0x7f746cc46210]
/lib/x86_64-linux-gnu/libc.so.6(+0x9b941)[0x7f746cc9b941]
...

**Comment by PaulWessel** [comment]
This may drag into tomorrow as I have been busy moving my UH home office back to UH today.  Setting up my computer rig in the office for the first time since March 2020...

**Comment by PaulWessel** [comment]
I think I just fixed that @Esteban82 - update and give it a check.

**Comment by Esteban82** [comment]
> I think I just fixed that @Esteban82 - update and give it a check.

Ok. I will check them as soon as posible.

**Comment by meghanrjones** [comment]
I am getting a segfault in a 6.2 script that I have used to plot images created by grdmix. I'll send an email with more details so that I can share the data since it's not using a remote file.

**Comment by Esteban82** [comment]
> I think I just fixed that @Esteban82 - update and give it a check.

It works with all the resolution, even 30s. 

Is there any other any test besides Meghan issues.

**Comment by PaulWessel** [comment]
Thanks, I have added a test based on @Esteban82 example above (same plus one that straddles the Dateline) and it passes.  Also fixed the things @meghanrjones mentioned related to her non-global large TIFF.  Hopefully we are ready to approve.

**Comment by joa-quim** [comment]
I'm getting a crash when trying to visualize and indexed image whose that is supposed to have transparency set to one of the rows of the cpt. That is done a bit under the hood by setting
```
img.n_colors = n_colors * 1000 + Int32(alpha_ind)
```
However, it also crashes in master so the breaking must have occurred after I implemented it but before the work in this branch. So it should not be a merge stopper. I'll look to what's happening when I have time.

meghanrjones (approved):

PR #5799, opened 2021-09-23
Apply scalings in api_get_record_matrix

meghanrjones:

Similar to https://github.com/GenericMappingTools/gmt/pull/5291, but for matrices. Unlike that PR, I did not do anything about checking whether GMT_IS_REFERENCE | GMT_IS_DUPLICATE is needed.

All GMT and PyGMT tests pass, except one which is wrong (xref https://github.com/GenericMappingTools/pygmt/issues/1539). I did not test GMT.jl.

Addresses https://github.com/GenericMappingTools/pygmt/issues/1313

**Reminders**


joa-quim:
This is the type of things I keep saying we should avoid. If there is something to convert we should have a different ``for loop`` to deal with it, instead of adding more and more ``if`` conditions inside the same loop.

Not particular to this case, though.

PaulWessel (approved):

meghanrjones:
> This is the type of things I keep saying we should avoid. If there is something to convert we should have a different `for loop` to deal with it, instead of adding more and more `if` conditions inside the same loop.
> 
> Not particular to this case, though.

The code could be refactored separately, but in this PR I am simply fixing a bug using the current structure.

PR #5803, opened 2021-09-26
CI: Install optional gdal executables from vcpkg

seisman:

In this PR (https://github.com/microsoft/vcpkg/pull/19243), vcpkg no longer builds GDAL's executables by default. We need to add the `[tools]` parameter to re-enable building the executables.

**Reminders**

### DIFF
--- a/.github/workflows/ci-caches.yml
+++ b/.github/workflows/ci-caches.yml
@@ -10,7 +10,7 @@ on:
   # Uncomment the "push:" line to manually re-cache data artifacts in pushes
   #push:
   # Uncomment the 'pull_request:' line to manually re-cache data artifacts in PRs
-  #pull_request:
+  pull_request:
   # Schedule runs on 12 noon every Sunday
   schedule:
     - cron: '0 12 * * 0'
--- a/.github/workflows/ci-caches.yml
+++ b/.github/workflows/ci-caches.yml
@@ -10,7 +10,7 @@ on:
   # Uncomment the "push:" line to manually re-cache data artifacts in pushes
   #push:
   # Uncomment the 'pull_request:' line to manually re-cache data artifacts in PRs
-  pull_request:
+  #pull_request:
   # Schedule runs on 12 noon every Sunday
   schedule:
     - cron: '0 12 * * 0'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-caches.yml
```diff
@@ -10,7 +10,7 @@ on:
   # Uncomment the "push:" line to manually re-cache data artifacts in pushes
   #push:
   # Uncomment the 'pull_request:' line to manually re-cache data artifacts in PRs
-  #pull_request:
+  pull_request:
```
seisman, line 13:
```suggestion
  #pull_request:
```

.github/workflows/ci-caches.yml
```diff
@@ -10,7 +10,7 @@ on:
   # Uncomment the "push:" line to manually re-cache data artifacts in pushes
   #push:
   # Uncomment the 'pull_request:' line to manually re-cache data artifacts in PRs
```
seisman, line 12: Need to comment Line 13 before merging!

ci/install-dependencies-windows.sh
```diff
@@ -19,7 +19,7 @@ PACKAGE="${PACKAGE:-false}"
 WIN_PLATFORM=x64-windows
 
 # install libraries
-vcpkg install netcdf-c gdal pcre2 fftw3[core,threads] clapack openblas --triplet ${WIN_PLATFORM}
+vcpkg install netcdf-c gdal[tools] pcre2 fftw3[core,threads] clapack openblas --triplet ${WIN_PLATFORM}
```
meghanrjones, line 22: Is only `gdal[tools]` needed here rather than `gdal[core,tools]`?

seisman: `core` is always built, but it does no harm to use `gdal[core,tools]`.

seisman:
```suggestion
vcpkg install netcdf-c gdal[core,tools] pcre2 fftw3[core,threads] clapack openblas --triplet ${WIN_PLATFORM}
```

meghanrjones (approved):
Looks good after re-commenting the cache trigger.

meghanrjones:
@seisman, should I add [core,tools] to the wiki page or is this only needed for the CI?

seisman:
Yes, need to update the wiki page, too.

PR #5814, opened 2021-09-28
use zlo/zhi instead of z_low/z_high for user messages

anbj:
Documentation use ``zlo/zhi``, so this should be used in user message.


PaulWessel (approved):
Looks good, thanks!

Issue #5802, opened 2021-09-26
Bug with Polar and horizont (-JG/-90/horizon) map.


Esteban82:
When I make a map center in south pole, the horizon parameter don't work.

` gmt coast -Rd -W -png map -Bfg -JG0/-90/50/10c -Sdodgerblue2  -N1`

![map](https://user-images.githubusercontent.com/17908951/134790635-35150421-8842-4de7-a490-e8edb1dcb7eb.png)

Here I change the latitude to -89.9.

` gmt coast -Rd -W -png map -Bfg -JG0/-89.9/50/10c -Sdodgerblue2  -N1`

![map](https://user-images.githubusercontent.com/17908951/134790656-e372af19-225a-45be-9fb6-7fb6d1bcd0e9.png)

PaulWessel:
They are actually two different projection codes in GMT for polar stereographic and just general stereographic.  I will have a look to see why horizon is not honored.

PaulWessel:
Sorry, I confused -JS with -JG....hold on.

PaulWessel:
So there are several things that ought to be improved here:

- In your S pole case, we take the **-Rd**, complain about north being +90, and change it to 0.
- For **-JA** and **-JS** the same case instead gives a hard error and exits.
- For **-JA** it is allowed to have a horizon up to and including 180.  If it is 180 then we may run into polygon fill issues similar to **-JE** of course, but there should not be a restriction to 90 here.
- For **-JS** it is also allowed to have a longer horizon, except exactly 180 which blows up.

I think I will:

1. Add a test on the horizon so it is not outside accepted ranges.
2. For the polar aspects of **-JG**, **-JS**, **-JA**, I will use the horizon to crop the N or S boundary accordingly if they are beyond the horizon and issue the warning and not have a mix of errors and warnings.  I think this is a reasonable solution since **-Rg** and **-Rd** are what is used for non-polar and it makes it odd to have to be specific for those special cases.

Issue #3409, opened 2020-05-29
Color the corners of fancy borders not just plain white


KristofKoch:
**Description of the desired feature**

Currently the corners of `fancy` and `fancy+` borders are simply white regardless of the color of the adjacent border. See the images in the upper row:

![corners](https://user-images.githubusercontent.com/45796239/82306273-1f3b3a80-99bf-11ea-9ff9-b0d74ff5e14c.jpg)

I would prefer the behavior in the lower row with the adjacent border color filling the corner up to its bisectrix. In my eyes this makes for a more eye-pleasing corner.

This issue is made from my post in issue #3344. I feel it should be separated form the discussion over there.

KristofKoch:
As my feature request hasn't seen any activity in a year – is it worth keeping it around or should I let it silently die?

PaulWessel:
Keep it there.  I did look briefly when first posted but it was not an easy thing to add so it did not get a high mental priority...

PR #5821, opened 2021-09-29
Clip boundaries to horizon for polar aspect azimuthal projections

PaulWessel:
See #5802 for background and discussion.  This PR treats **-JA**, **-JG**, and **-JS** the same regarding how **-Rg** and **-Rd** are adjusted to honor the horizon, and relegates the tedious warning message to information instead.

The WIP aspect of this PR is because I noticed that when running @Esteban82's example from #5802 I got this:

map

which seems to have way too many latitudinal gridlines.  In comparison, if we move the center point off the S pole it changes quite a bit:

map

The reason for that is that for the latter case, the region remains at **-Rd** (map boundaries are not parallels and meridians) while for the polar aspect the north boundary is set to -40.  So longitude range is 360 and latitude range is 50 and we get this disparity. Those decisions are done by _gmt_auto_frame_interval_ which assumes we have a Cartesian geometry.  I am experimenting with a revised scheme where for polar azimuthal projections I treat the map width as a diameter and scale by PI (for longitude) and divide it by 2 (for latitudes) to better compare degree range to length scale.  Invariable, whatever one chooses here there will be failures against the originals.  For my current values (note I only multiply by PI/2 for longitudinal length scale instead of PI) I get something like this:

map

which is more even, but perhaps too many (plot is only 10 cm wide).  Adding another divisor of 2 gives this (which does not change the latitude spacing in this case):

map

What do you think of this improvement and do you have better values? @meghanrjones, you looked at a lot of this during the themes buildup - see what you think (you can change those constants in _gmt_auto_frame_interval_).


meghanrjones:
Nice, definitely an improvement. My preference would be for even fewer latitudinal gridlines for this case, though that would involve more adjustments than the simple scaling of `f` due to `d *= MAX (0.05, MIN (5.0 * GMT->current.setting.font_annot[item].size / f, 0.20));` (L6754).

As a general question, why is there guessing for what the user might want for -Rg|d for azimuthal projections but not cyclindical projections? For example, in my opinion interpreting -Rd for -JM as -R0/360/-70/70 seems about as reasonable as interpreting -Rd for -JS0/90 to be -R0/360/0/90. Currently, this is disallowed whereas for using -Rd for -JT is allowed with some weird results.

PaulWessel:
Fewer latitude lines: Perhaps you can think of a better geometric argument than I did.  Maybe linear is not a good scale for determining circle spacing?

> As a general question, why is there guessing for what the user might want for -Rg|d for azimuthal projections but not cyclindical projections?

I think for azimuthal there really is no guessing: We simply apply the horizon clip.  If the user does not change the horizon value (90 for these) then a polar view will stop at Equator since it must.  For Mercator there is no equivalent.  Why would -R0/360/-70/70 be more logical than -R0/360/-80/80? Since there is no hard limit here I do not think we should do a bad guess - better to slap the error.

meghanrjones:
> Fewer latitude lines: Perhaps you can think of a better geometric argument than I did. Maybe linear is not a good scale for determining circle spacing?
> 

The scaling might be fine, it's the bounds of the interval that may need adjusting. The largest that the interval between circles can be now is 0.2 * *horizon*, subject to rounding to even intervals, which might be too small.

> > As a general question, why is there guessing for what the user might want for -Rg|d for azimuthal projections but not cyclindical projections?
> 
> I think for azimuthal there really is no guessing: We simply apply the horizon clip. If the user does not change the horizon value (90 for these) then a polar view will stop at Equator since it must. For Mercator there is no equivalent. Why would -R0/360/-70/70 be more logical than -R0/360/-80/80? Since there is no hard limit here I do not think we should do a bad guess - better to slap the error.

OK, so I think these same restrictions should apply to -JO and -JT then.

PaulWessel:
> OK, so I think these same restrictions should apply to -JO and -JT then.

You mean for -R without +r?  For oblique, we almost always use +r to pick two points, and for -JT as well.  But if no +r I guess -JT has a limit on < 90 from central meridian in longitude, and for -JO I am not sure...

meghanrjones:
> > OK, so I think these same restrictions should apply to -JO and -JT then.
> 
> You mean for -R without +r? For oblique, we almost always use +r to pick two points, and for -JT as well. But if no +r I guess -JT has a limit on < 90 from central meridian in longitude, and for -JO I am not sure...

It should have a limit, but I am not sure that it actually does:
`gmt coast -Rg -JT330/-45/10c -Ba30g -BWSne -Dc -A2000 -Slightblue -G0 --MAP_ANNOT_OBLIQUE=lon_horizontal -pdf GMT_TM`

PaulWessel:
Looking at TM.  We have documentation scripts like GMT_TM.sh that does this:

`gmt coast -R0/360/-80/80 -JT330/-45/10c -Ba30g -BWSne -Dc -A2000 -Slightblue -G0 --MAP_ANNOT_OBLIQUE=lon_horizontal --GMT_THEME=cookbook -ps GMT_TM`

Per gmtmap_init_tm, we say that if longitude range is 360 (like here), this happens:

```
	GMT->current.map.is_world = gmt_M_360_range (GMT->common.R.wesn[XLO], GMT->common.R.wesn[XHI]);
	if (GMT->current.map.is_world) {	/* Gave oblique degrees relative to central meridian */
		double w, e, dummy = 0.0;
		w = GMT->current.proj.central_meridian + GMT->common.R.wesn[YLO];
		e = GMT->current.proj.central_meridian + GMT->common.R.wesn[YHI];
		GMT->common.R.wesn[YLO] = -90;
		GMT->common.R.wesn[YHI] = 90;
		GMT->common.R.wesn[XHI] = e;
		GMT->common.R.wesn[XLO] = w;
```

I am not entirely clear on this but it is definitely a special case since the latitudes are reset to the full range when e-w = 360.  Notice that when you give -Rw/e/s/n to these projections we assume these are oblique w/e/s/n in the projection's coordinate system, not really longitude/latitudes as we normally expect.  So basically I think the -90/+90 are essentially the real longitude limit and we can go 360 in the other direction.

What I can do is in the case where 360 is not specified.  Here we can add a check:

```
		/* Ensure that oblique longitudes are < 90 away from central meridian */
		if (GMT->common.R.wesn[XLO] <= -90 || GMT->common.R.wesn[XHI] >= 90) {
			GMT_Report (GMT->parent, GMT_MSG_ERROR, "Transverse Mercator: Bounds on oblique longitudes must be withing +-90\n");
			return (GMT_NOT_A_VALID_DOMAIN);
		}
```

That means this passes fine:

`gmt coast -R-60/40/-50/50 -JT330/-45/10c -Ba30g -BWSne -Dc -A2000 -Slightblue -G0 --MAP_ANNOT_OBLIQUE=lon_horizontal -png GMT_TM`

GMT_TM

while this one fails:

```
gmt coast -R-60/100/-50/50 -JT330/-45/10c -Ba30g -BWSne -Dc -A2000 -Slightblue -G0 --MAP_ANNOT_OBLIQUE=lon_horizontal -png GMT_TM
coast [ERROR]: Transverse Mercator: Bounds on oblique longitudes must be withing +-90
coast [ERROR]: General map projection error
```

I do not think most people make these types of plots and I notice for valid ranges (like -R-60/70/-50/50) there are wrapping problems that I do not want to track down at the moment...You can also see the longitude annotations are not really anticipating this shape.

So once tests pass here I will submit a PR that adds that check only for TM.

PaulWessel:
This is not working well. There is no limit on oblique longitudes (360 range is fine for oblique longitudes), there is of course a limit on < 90 for the lats.  I see both -JM and -JO does not check that N or S might be exactly 90/-90 so adding that test only.

PaulWessel:
Adding a check that oblique north cannot be +90 exactly causes other failures where 90 is given and the projection works.  So perhaps best to let sleeping bears lie.

meghanrjones:
> Adding a check that oblique north cannot be +90 exactly causes other failures where 90 is given and the projection works. So perhaps best to let sleeping bears lie.

OK, sorry for disturbing the sleeping bear.

src/gmt_map.c
```diff
@@ -6718,10 +6728,12 @@ void gmt_auto_frame_interval (struct GMT_CTRL *GMT, unsigned int axis, unsigned
 	if (axis == GMT_X) {
 		f = sxy * fabs (GMT->current.proj.rect[XHI] - GMT->current.proj.rect[XLO]);
 		d = fabs (GMT->common.R.wesn[XHI] - GMT->common.R.wesn[XLO]);
+		if (gmtmap_polar_az (GMT)) f *= M_PI_2;	/* Circle perimeter is the right comparison to circular longitudes */
```
meghanrjones, line 6731:
```suggestion
		if (gmtmap_polar_az (GMT) && d > 350) f *= M_PI_2;	/* Circle perimeter is the right comparison to circular longitudes */
```
In my opinion, `doc/scripts/GMT_stereographic_polar.sh` looked better before the changes. What do you think about limiting these adjustments to only maps that span most longitudes?

PaulWessel: Yes, that is probably OK, but is 350 the best limit? Did 180 not work?

meghanrjones: I'll do more tests to determine a reasonable limit. Is it OK for me to push some recommendations to this branch? It would be easy enough to revert them if needed and would work better than GitHub suggestions which are limited to certain lines.

PaulWessel: Yes, go ahead.

meghanrjones:
@PaulWessel, I updated the scaling after some additional tests for your consideration.

While testing this branch, I noticed that for polar aspect azimuthal projections that form a wedge, the frame at r=0 does not touch. Do you think this is something that we could improve?

```
gmt coast -R0/120/45/90 -W -Bfg -JS0/90/50/10c -Sdodgerblue2 -N1 -png JSR2
```

JSR2

PaulWessel:
Yes, we need to fix that.  Have not noticed that before - a regression?  Brings #3409 to mind as well.

One solution would be to draw an arc with the fat pen at that polar point given the angles.  I.e., if n == 90 or s == -90 and e-w is not 180 or 360 (nt sure about 90 270) then draw than arc.

meghanrjones:
> Yes, we need to fix that. Have not noticed that before - a regression? Brings #3409 to mind as well.
> 

It looks different, but not better in 6.1.1 (below). Not sure about releases prior to that.

JSR2

PaulWessel:
I am committing soon a section of code to gmtplot_polar_map_boundary that connets the two straight sides with an arc.  It seems fine for your case.  If you agree, could you do the copy/paste/revise to also make it do the S pole and test that as well?

meghanrjones:
> I am committing soon a section of code to gmtplot_polar_map_boundary that connets the two straight sides with an arc. It seems fine for your case. If you agree, could you do the copy/paste/revise to also make it do the S pole and test that as well?

Yes, will do.

meghanrjones:
I think the arc looks odd if the origin is on the white part of the pajama frame. What is your opinion on this?

```gmt basemap -R0/120/45/90 -Bf50g -JS0/90/10c -png test```
test

PaulWessel:
Yes, needs more work

Paul Wessel, Professor
Dept. of Earth Sciences
SOEST, U of Hawaii at Manoa

PaulWessel:
Maybe just add 90 to the start and subtract 90 from stop and only draw if stop > start?

meghanrjones:
> Maybe just add 90 to the start and subtract 90 from stop and only draw if stop > start?

Looks better now. One other issue is that the tick marks go inwards in the poles.  

```gmt basemap -R0/120/-90/-45 -Bafg -JS0/-90/10c -png test_s --MAP_TICK_LENGTH=20p --MAP_TICK_PEN=2p,red```

test_s

PaulWessel:
Hm, this must also be true in master then, no? If so, lets deal with that in a separate PR.

meghanrjones:
> Hm, this must also be true in master then, no? If so, lets deal with that in a separate PR.

Yes true in master since at least 6.1.1. Sounds good. I think the new PS tests are good. Do you want to take a look at them or should I go ahead and update now?

PaulWessel:
Let me run tests first

Paul Wessel, Professor
Dept. of Earth Sciences
SOEST, U of Hawaii at Manoa

PaulWessel:
OK, looks fine to me, please update the PS changes.

meghanrjones (approved):
No more comments here.

PR #5824, opened 2021-09-29
Simplify conversions between km and degrees

PaulWessel:
Users (like myself...) who have data columns representing distances on the Earth (or other planets) may need to convert between arc and distance units (mostly degrees and km).  This is an awkward thing to do via proxies like +s111.195.  In **grdmath** we have operators **DEG2KM** and **KM2DEG** for this.  For tables I have now implemented special arguments **d** and **k** to **-i**_col_**+s**.  So, `-i0+sk` will convert spherical degrees in column one to km, using the prevailing mean ellipse radius.
No effect on any existing tests, and I checked that both codes work correctly.

### DIFF
--- a/doc/rst/source/explain_-icols_full.rst_
+++ b/doc/rst/source/explain_-icols_full.rst_
@@ -3,7 +3,7 @@ The **-i** option
 
 **Syntax**
 
-**-i**\ *cols*\ [**+l**][**+d**\ *divisor*][**+s**\ *scale*][**+o**\ *offset*][,\ *...*][,\ **t**\ [*word*]]
+**-i**\ *cols*\ [**+l**][**+d**\ *divisor*][**+s**\ *scale*\|\ **d**\|\ **k**][**+o**\ *offset*][,\ *...*][,\ **t**\ [*word*]]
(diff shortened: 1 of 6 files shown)


doc/rst/source/explain_-icols_full.rst_
```diff
@@ -16,7 +16,8 @@ one of the following modifiers to any column or column range to transform the in
 
 - **+l** to take the **log10** of the input values.
 - **+d** to divide the input values by the factor *divisor* [default is 1].
-- **+s** to multiply the input values by the factor *scale* [default is 1].
+- **+s** to multiply the input values by the factor *scale* [default is 1]. Alternatively, give **d** to convert
+  km to degrees or **k** to convert degrees to km.
```
meghanrjones, line 20:
```suggestion
  km to degrees or **k** to convert degrees to km using :term:`PROJ_MEAN_RADIUS`.
```

meghanrjones (approved):
Nice!

PaulWessel:
Yikes, I forgot to accept @meghanrjones suggestion... OK, I added the missing text directly to master...

Issue #5823, opened 2021-09-29
auto resolution broken


anbj:
Since yesterday, I've had problems with the new auto resolution feature.

With git head:

```
$ gmt grdimage @earth_relief -JM10c -RNO -pdf auto-res-problem
grdimage [ERROR]: Passing zmax or zmin == NaN prevents automatic CPT generation!
grdimage [ERROR]: Failed to read CPT (null).
```

Specify the resolution of the grid, and it works:
```
$ gmt grdimage @earth_relief_10m -JM10c -RNO -pdf works
```

So it appears to be a problem with the auto-resolution-thing.

After a session of `git bisect`, it turns out that this commit b6c470536bb2b49bc09d8a5677c46554a90265f5 is the culprit.

Let me know if you need more info.

meghanrjones:
Closed by #5825

PR #5825, opened 2021-09-30
Forgot check on dry-run or not

PaulWessel:
See #5823 for the failure.  The problem was I had a goto line that should only be reached if we are during **grdcut -D,** not **grdimage** and other modules.  So it skipped assembling the grid...


anbj (approved):

anbj:
Script runs fine again - looks good, Paul!

meghanrjones (approved):

PR #5840, opened 2021-10-03
Prevent crashes when I->n_indexed_colors have encoded info about alpha layer

joa-quim:
It is possible to pass in a colormap with 4 columns, where 4rth contains info about transparency. It is even possible to pass info to GDAL about which color in the colormap is to be taken as the transparent color (``GDALSetRasterNoDataValue``). But since what we pass to GMT is a pointer to the colormap we must tell it also if the colormap has 3 or 4 columns. That is done by multiplying the number of colors in colormap by 1000, and later check if that number is > 2000 to know if we have 3 or 4 columns.

Further encoding of the transparent color in colormap is coded by adding the color number entry in colormap to the number of colors. For example if we have a cololrmap with 256 colors and want to the second color as the transparent one, we set ``ncolors = 256 *1000 +2``.

Now, this used to work but doesn't anymore because when ``gmtapi_expand_index_image`` comes in, it converts the indexed image to RGB but looses the information about the transparency. Saving image with gdalwrite encodes this info but currently when I open that image I cannot see it working, but this is a different issue.

This PR, prevents crashings that would otherwise result from accessing the image data with a wrong number of colors.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -4363,12 +4363,16 @@ GMT_LOCAL bool gmtapi_expand_index_image (struct GMT_CTRL *GMT, struct GMT_IMAGE
 	h = I->header;
 	data = gmt_M_memory_aligned (GMT, NULL, h->size * 3, unsigned char);	/* The new r,g,b image */
 
+	size_t n_colors = I->n_indexed_colors;
+	if (n_colors > 2000)			/* If colormap is Mx4 or has encoded the alpha color */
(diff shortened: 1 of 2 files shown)


PaulWessel (approved):
All tests pass, so seems OK.

joa-quim:
Update. Saving image with a transparent color from the colormap actually works. This (rare) time Mirone failed me in comparison with QGis, that sows the transparency where due.

PR #5843, opened 2021-10-05
Handle polar caps for TM maps as well

PaulWessel:
And possibly UTM: Like oblique Mercator, transverse Mercator maps may show a pole as well, and then we should honor the polar cap defaults.  This PR addresses that and updates 4 TM plots that now have caps.


PaulWessel:
> In TMcontours.ps, why do the latitude gridlines inside the polar cap not align with the gridlines outside the cap?

I think the spacing outside is controlled by auto -Bg while inside it is set via MAP_POLAR_CAP.

PR #5846, opened 2021-10-06
Clarify docs for grdvolume -C

meghanrjones:

This PR cleans up the documentation for grdvolume -C. A couple questions/comments regarding this:

- the old version says "find area, volume and mean height (volume/area)". What does mean height (volume/area) mean?
- the old version says "The **Cr** form on the other hand computes the volume below the grid surface and above the planes defined by *low* and *high*". I think this describes the same situation as **-C**. I have adjusted to refer to above the grid surface and below/between the plans for **-Cr**.

**Reminders**

### DIFF
--- a/doc/rst/source/grdvolume.rst
+++ b/doc/rst/source/grdvolume.rst
@@ -51,16 +51,26 @@ Optional Arguments
 
 .. _-C:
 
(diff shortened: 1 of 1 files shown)


PaulWessel:
> * he old version says "find area, volume and mean height (volume/area)". What does mean height (volume/area) mean?

It means mean height = volume / area, so surface of constant height of the same volume and area.

PaulWessel (approved):

Issue #5857, opened 2021-10-07
Add a new bit flag to GMT_Create_Session() for externals


joa-quim:
The problem: Both GMT and the externals may compete to interact with the GDAL shared lib in the use of ``GDALDestroyDriverManager``.

While multiple calls to ``GDALAllRegister()`` seem to have no visible adverse consequences, a call to ``GDALDestroyDriverManager`` inevitably *finishes the GAL session* and GMT does it in many places. Now, in Julia other than calling GMT modules that do this, I also use (many) other functions that access the GDAL lib directly and expect the driver manager to be alive all the time. I have somehow worked around this problem by trying to detect the situation and relaunching ``GDALAllRegister()``. However, this works only some times and the other comes Julia the garbage collector trying to free the memory holding the result of the first ``GDALAllRegister()`` call doing and crashes Julia because it's no longer a valid memory.

I've tried all I can imagine but still have random crashes when using GMT modules and other GMT.jl functions that call GDAL. The only way of securely prevent this would be to have a *mode* in which GMT would never call ``GDALDestroyDriverManager`` when operated from externals. And to make this eve safer it would operate only if ``GMT_Create_Session`` were passed a new mode. Then we could perhaps let ``API->external = 2`` transmit this info into a new function that would call  ``GDALDestroyDriverManager`` only if ``API->external < 2``. This to try to avoid adding yet another member to the already long ``GMTAPI_CTRL`` struct.

PR #5858, opened 2021-10-07
Move the GDALDestroyDriverManager() function into a wrapper function to allow conditional call.

joa-quim:
Experimental implementation of #5857. Tentative name and location.
Still misses adding a new bitflag to *mode* but apparently works for intended purpose.

### DIFF
--- a/doc/rst/source/devdocs/api.rst
+++ b/doc/rst/source/devdocs/api.rst
@@ -947,8 +947,10 @@ is a sum of bit flags and the various bits control the following settings:
 #. Bit 5 (16 or GMT_SESSION_RUNMODE): If set, the we enable GMT's modern run-mode (where -O -K are
    not allowed and PostScript is written to hidden temp file).  Default
    is the GMT classic run-mode.
-#. Bit 6 (32 or GMT_SESSION_NOHISTORY): If set, the we disable GMT's command shorthand via gmt.history files.
+#. Bit 6 (32 or GMT_SESSION_NOHISTORY): If set, then we disable GMT's command shorthand via gmt.history files.
    The default is to allow this communication between GMT modules.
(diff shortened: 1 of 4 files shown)


joa-quim:
Sorry, I don't see what I would ask to Even. If the drivers are not registered one cannot call functions that deal with data dependent on knowing about the drivers. It's like us having to call ``GMT_Create_Session`` before calling any GMT function because some 3rd party program had called ``GMT_Destroy_Session``

PaulWessel:
The question is how are separate libraries expected to open/close GDAL if used by an application (e.g. GMT.jl).  You could tell him what GMT.jl does (initialize GDAL then call many times etc) and that GMT.jl calls the GMT library which also talks to GDAL, and the problem is when GMT closes the connection it affects GMT.jl's own GDAL connection.  This seems like an odd design.  Are there GDAL functions to check if there is an active GDAL connection that one could use?

joa-quim:
This PR is not ready. It needs a new _flag_ such that

```
API->external = (mode & GMT_SESSION_EXTERNAL) ? 1 : 0;
API->external = (mode & GMT_SESSION_NODESTROY_GDALDRIVERS) ? 2 : API->external;
```
and in Julia or **MEX** one can call 

```
GMT_Create_Session("GMT", pad, GMT_SESSION_NOEXIT + GMT_SESSION_EXTERNAL)

or

GMT_Create_Session("GMT", pad, GMT_SESSION_NOEXIT + GMT_SESSION_EXTERNAL + GMT_SESSION_NODESTROY_GDALDRIVERS)
```

So we need a ``GMT_SESSION_NODESTROY_GDALDRIVERS`` or  some shorter name

joa-quim:
Yes, please. That's why it was a WIP PR

PR #5870, opened 2021-10-14
Enable sharing of SRTM tiles among multiple DEMs

PaulWessel:
With the introduction of `@earth_synbath` there will be more than one data set (`@earth_relief` is the other) that for the highest resolutions will switch to SRTM 1 or 3 arc second tiles.  To avoid having to duplicate 41 Gb on the server as well as in the user's .gmt tree, I am adding a few minor code modifications to extend the current switch from earth_relief_01|3s to SRTMGL1|3 to also extend to other DEMs with similar needs.  Currently that is only `@earth_synbath`.  I do not anticipate more DEMs that need to share SRTM, but if that should happen it will require a minor GMT version update.  This also means that only GMT 6.3 users can use `@earth_synbath` for regions where SRTM tiles may be involved (they can use it just fine away from land with GMT 6.1 etc).

All the other new remote data will work with existing GMT versions that can handle remote data.

I ran all the tests using **GMT_DATA_SERVER** as _test_ instead of _oceania_ and we can anticipate ~40 test failures due to the `@earth_relief` refresh.  Of course, this is similar to what we expect when the tests remove the resolution specification and go with the auto-resolution setting.

All tests works as before with the regular data server.

### DIFF
--- a/doc/rst/source/explain_-B.rst_
+++ b/doc/rst/source/explain_-B.rst_
@@ -95,7 +95,7 @@ The Axes settings are specified by
 
 but you may also split this into two separate invocations for clarity, i.e.,
 
-   | **-B**\ [**p**\|\ **s**][**x**\|\ **y**\|\ **z**][**+a**\ *angle*\|\ **n**\|\ **p**][**+f**]\
+   | **-B**\ [**p**\|\ **s**][**x**\|\ **y**\|\ **z**][**+a**\ *angle*\|\ **n**\|\ **p**][**+e**\ [**l**\|\ **u**]][**+f**]\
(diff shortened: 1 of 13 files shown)


meghanrjones:
> I had lots of these too until I realized I needed to rename the server and server2 stuff so I am back at the default data settings... Could that be it?

I renamed the server directory before the tests. I am double-checking now but it will take a while because I cannot run the tests in parallel when the data are not pre-downloaded.

PaulWessel:
Sure, and I just rebuilt this branch after merging in master and ran all tests and all passes (well ex49 always fails a bit on my M1 chip Mac mini but not on Intel...)

meghanrjones:
> Sure, and I just rebuilt this branch after merging in master and ran all tests and all passes (well ex49 always fails a bit on my M1 chip Mac mini but not on Intel...)

ex49 always fails on DarthMac (intel) too.

PR #5877, opened 2021-10-19
INSTALL.md: add --verbose and --ask to gentoo

anbj:


anbj:
Much obliged.

anbj:
Semirelated: when OS's have package managers that install old (or ancient) versions of GMT; should we care about this? Should we do an effort to try to get them up-to-date?

PR #5883, opened 2021-10-20
Fix anim02 -V

maxrjones:

Fix -V1 to -Vi

Fixes #

**Reminders**


meghanrjones:
> No, I don't think so. I was mostly concerned with -V1 being wrong. Hi @meghanrjones did you intend to mean that all animations should have -V so the user can see things moving along? I think for the hour-long movies that may be a good idea, so perhaps doing it for all is OK?

I didn't mean to imply that it's required. But I agree that it is useful for long movies, so would be OK to do it for all.

PR #5887, opened 2021-10-21
Better auto-resolution selection for grids and images

PaulWessel:

I have been tinkering with how to determine a suitable remote grid image resolution that would work well with all projections.  My initial attempts (in master) examined both longitude and diagonal degree range for a map.  However, longitudes are difficult as the longitude range can suddenly jump from a small range to 360 if a pole enters the region.  So this PR only considers the latitude range (_Y_) relative to the map height (_H_).  Then, the required number of grid nodes per degree is computed via `N_f = L * dpi / D,` where _dpi_ is the desired dots-per-inch needed.  We then use the next higher actual _N_r_ of the nearest grid resolution.  I am testing this with a zoom from 100,000 km altitude to 10 km altitude over a fixed point (Vienna).  The movie script is

```
#!/usr/bin/env bash
lat=47.8095
lon=13.0550
dpc=100
cat << EOF > pre.sh
gmt begin
	gmt math -T1/5/720+n 10 T POW -o1 -I = altitude.txt
	gmt makecpt -Cgeo -H > topo.cpt
gmt end
EOF
cat << EOF > main.sh
gmt begin
	gmt grdimage @earth_relief -Rd -JG$lon/$lat/14.9c+z\${MOVIE_COL0}+v60 -Yc -Xc -B0 -Ctopo.cpt -I+d --GMT_GRAPHICS_DPU=${dpc}c --MAP_FRAME_PEN=faint
	echo $lon $lat | gmt plot -SE-10 -Gred@50
	L=\$(gmt grdcut @earth_relief -Rd -JG$lon/$lat/14.9c+z\${MOVIE_COL0}+v60 -D+t --GMT_GRAPHICS_DPU=${dpc}c)
	echo \${L} | gmt text -F+cTR+jTR+f8p
	echo "altitude = \${MOVIE_COL0} L = \${L}"
gmt end
EOF
gmt movie main.sh -Sbpre.sh -NZoom_Relief_${dpc}c -Taltitude.txt -C15cx15cx100 -D24 -H8 -M0,png -Fmp4 -Lc0+f12p+t"Altitude = %6.1lf km" -V
```

The movie can be seen on [YouTube](https://youtu.be/JJ0PDYBXzpo).  Via debug dump I have captured both _N_f_ and _N_r_ and for this wide range of altitudes it looks like this:

summary

As you can see, the selected grid resolution (blue) is always equal or higher than the required, until we run out of even higher resolutions after 1 arc second around 25 km altitude.  Please view the movie and see how it transitions between resolutions.  Since the data sets are not always exactly the same you will notice the jumps to the better data but it is pretty good I think. Remember that making a movie with variable resolution like this is the most extreme case of this.  Static plots will not experience this of course.

Given the stability that using Y brings to computing the required nodes per degree, I recommend you approve this PR.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -14824,25 +14824,15 @@ GMT_LOCAL int gmtinit_compare_resolutions (const void *point_1, const void *poin
 	return (0);
 }
 
-GMT_LOCAL double gmtinit_map_diagonal_degree (struct GMT_CTRL *GMT) {
-	/* Return a diagonal or representative distance in degrees across the map */
-	double X = GMT->common.R.wesn[XHI] - GMT->common.R.wesn[XLO];
+GMT_LOCAL double gmtinit_map_vertical_degree (struct GMT_CTRL *GMT) {
+	/* Return the max latitude separation in degrees across the map */
 	double Y = GMT->common.R.wesn[YHI] - GMT->common.R.wesn[YLO];
-	double D = gmtlib_great_circle_dist_degree (GMT, GMT->common.R.wesn[XLO], GMT->common.R.wesn[YLO], GMT->common.R.wesn[XHI], GMT->common.R.wesn[YHI]);	/* "Diagonal" great circle distance in degrees */
-	/* For some global projections the D here may be 180 even though we see a full 360 degrees in longitude.  Hence these checks */
-	if (Y > D) D = Y;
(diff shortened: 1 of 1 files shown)


PaulWessel:
Forgot to remote the printf of course, will do in next PR.

Issue #5894, opened 2021-10-23
grdinterpolate does not define many documdented common options (lower case)


joa-quim:
It only defines
```
#define THIS_MODULE_OPTIONS "->RVfn"
```
but if I replace it with
```
define THIS_MODULE_OPTIONS "-:>RVbdefghinoqs"
```
then it crashes when I use it with **-S -o**

PR #5901, opened 2021-10-24
Add missing KEYs and options

joa-quim:
**-E** and **-S** KEYs were missing 
Partially addresses #5894 but not the fact that when **-o** is used the program crashes.

### DIFF
--- a/src/grdinterpolate.c
+++ b/src/grdinterpolate.c
@@ -35,7 +35,7 @@
 #define THIS_MODULE_MODERN_NAME	"grdinterpolate"
 #define THIS_MODULE_LIB		"core"
 #define THIS_MODULE_PURPOSE	"Interpolate a 3-D cube, 2-D grids or 1-D series from a 3-D data cube or stack of 2-D grids"
-#define THIS_MODULE_KEYS	"<G{+,G?},ED),SD)"
+#define THIS_MODULE_KEYS	"<G{+,G?},ED(,SD("
 #define THIS_MODULE_NEEDS	""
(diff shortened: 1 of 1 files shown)


PaulWessel:
Maybe we are stuck until I can do gmtmex on Intel at work tomorrow.  All I can see is when GMT_Read_Data on line 543 in grdinterpolate.c returns, In is non-zero but In->table is NULL.  And all GMT_Read_Data does is to find the right entry in API->objects and return that resource.  So if you can, you should check that the resource you hook up on your side is not only non-NULL but that its table is non-Null.  If not then we must wait.

joa-quim:
SHIT, the KEYs are wrong. Must be ``(``, not ``)``
```
#define THIS_MODULE_KEYS	"<G{+,G?},ED(,SD("
```

joa-quim:
Updated the PR

joa-quim:
Yes, you can check it yourself too.

joa-quim:
You are probably on master or some other branch. Here
```
julia> grdinterpolate("cube.nc", S=[4.5 4;5.5 5], o="0-3", Vd=1)
        grdinterpolate cube.nc  -n+a -o0-3 -S
Vector{GMTdataset} with 2 segments
Header1:        Location 4.5,4
First segment DATA
4×4 Matrix{Float64}:
 4.5  4.0  1.0  18.0
 4.5  4.0  2.0  19.8
 4.5  4.0  3.0  21.6
 4.5  4.0  5.0  25.2
```

Issue #5875, opened 2021-10-18
Let inset begin accept -R and -J to automatically set width or height


maxrjones:
**Description of the desired feature**

Currently, the user must use mapproject to determine the appropriate size for a inset to match a given region and projection. It would be useful to allow `inset begin` to accept the region and projection of the inset, similar to subplot begin, to avoid this extra step. Here is an example of the anticipated usage:

```
gmt begin test png
    gmt grdimage @earth_relief -R-51/-48/-24/-23 -JM15c -B -Cworld
    gmt inset begin -DjBL+o0.1c -F+p1p,black -R-80/-28/-43/10 -JM2c
        gmt coast -Wthin -Swhite -Ggray 
    gmt inset end
gmt end show
```

![expected output](https://user-images.githubusercontent.com/26473567/137820144-aa6a3b06-d1b4-41a7-a052-0451c766e213.png)

**Are you willing to help implement and maintain this feature?** Yes

PaulWessel:
OK, to close this issue, @meghanrjones ?

PR #5903, opened 2021-10-25
Let inset begin optionally take -R -J to determine inset size instead of via -D

PaulWessel:
See #5875 for background.  Normally, **gmt inset begin** is called without an explicit **-R -J** and instead it picks those up from the main figure history file that it is setting up an inset for.  In that case, users must specify the inset size via **-D...+w**. However, there are times when the user wishes to specify the inset size indirectly via the **-R -J** settings that will be used _within the inset_ itself.  This PR implements that suggestion and adds a new test _inset2.sh_ to demonstrate its use.

### DIFF
--- a/doc/rst/source/inset.rst
+++ b/doc/rst/source/inset.rst
@@ -89,7 +89,8 @@ Optional Arguments (begin mode)
     This is clearance that is added around the inside of the inset.  Plotting will take place
     within the inner region only. The margins can be a single value, a pair of values separated by slashes
     (for setting separate horizontal and vertical margins), or the full set of four margins (for setting
-    separate left, right, bottom, and top margins) [no margins].
+    separate left, right, bottom, and top margins) [no margins]. Append units as desired [Default is set
+    by :term:`PROJ_LENGTH_UNIT`].
 
 .. _-N:
 
--- a/doc/rst/source/inset.rst
+++ b/doc/rst/source/inset.rst
@@ -23,6 +23,8 @@ Synopsis (begin mode)
 [ |-F|\ *box* ]
 [ |-M|\ *margins* ]
 [ |-N| ]
+[ |SYN_OPT-R| ]
+[ |SYN_OPT-J| ]
 [ |SYN_OPT-V| ]
 [ |SYN_OPT--| ]
 
@@ -36,7 +38,9 @@ records the current region and projection so that we may return to the initial
 plot environment when  the inset is completed.  The user may select any plot region
 and projection once plotting in the inset, but if the first command uses a projection
 that leaves off the scale or width then we supply a scale or width to fill the inset as best
-as possible, given the inset size and margins (if selected).
+as possible, given the inset size and margins (if selected). **Note**: If you wish to let
+the inset dimensions be determined by the region and projection that will be used to draw in
+the inset, then give these arguments on the **gmt inset begin** command.
 
 
 Required Arguments (begin mode)
@@ -74,6 +78,11 @@ Optional Arguments (begin mode)
 
     .. include:: explain_-F_box.rst_
 
+.. |Add_-J| replace:: |Add_-J_links|
+.. include:: explain_-J.rst_
+    :start-after: **Syntax**
+    :end-before: **Description**
+
 .. _-M:
 
 **-M**\ *margins*
@@ -86,6 +95,11 @@ Optional Arguments (begin mode)
 
 **-N**
     Do NOT clip features extruding outside map inset boundaries [Default will clip].
+
+.. |Add_-R| replace:: This is useful when you want the inset **-R -J** to also determine the inset size. |Add_-R_links|
+.. include:: explain_-R.rst_
+    :start-after: **Syntax**
+    :end-before: **Description**
 
 .. |Add_-V| replace:: |Add_-V_links|
 .. include:: explain_-V.rst_
--- a/doc/rst/source/subplot.rst
+++ b/doc/rst/source/subplot.rst
@@ -148,7 +148,8 @@ Optional Arguments (begin mode)
     to the automatic space added for tick marks, annotations, and labels.  The margins can be specified as
     a single value (for same margin on all sides), a pair of values separated by slashes
     (for setting separate horizontal and vertical margins), or the full set of four slash-separated margins
-    (for setting separate left, right, bottom, and top margins).  The actual gap created is always a sum of
+    (for setting separate left, right, bottom, and top margins).  Append units as desired [Default is set
+    by :term:`PROJ_LENGTH_UNIT`]. The actual gap created is always a sum of
     the margins for the two opposing sides (e.g., east plus west or south plus north margins) [Default is
     half the primary annotation font size, giving the full annotation font size as the default gap].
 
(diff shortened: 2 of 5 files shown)


meghanrjones (approved):
Nice!

Is it possible to show annotations for inset maps? For example, below the -Ba does not impact the inset map:
```
gmt begin
    gmt figure test-inset png 
    gmt grdimage @earth_relief -R-51/-48/-24/-23 -JM15c -B -Cworld
    gmt inset begin -DjBL+o0.1c -F+p1p,black -R-80/-28/-43/10 -JM2c
        gmt coast -Wthin -Swhite -Ggray -Bafg -BwsNE
    gmt inset end
    gmt figure test png
        gmt coast -Wthin -Swhite -Ggray -Bafg -R-80/-28/-43/10 -JM2c -BwsNE
gmt end show
```

Esteban82:
> Is it possible to show annotations for inset maps? For example, below the -Ba does not impact the inset map:

I think that annotations fall outside of the inset. The issue is that the width of the inset is the same of the map. So the inset should take into account the extra space.

Esteban82:
I test it with ex 44 but with GMT 6.2 (not dev version) where I use -Baf

ex44


```
gmt begin ex44 png 
	gmt coast -R10W/5E/35N/44N -JM15c -EES+gbisque -Gbrown -Wfaint -N1/1p -Sazure1 -Dh --FORMAT_GEO_MAP=dddF
	gmt inset begin -DjTR+w5c/2.35c+o0.4c/0.25c -F+gwhite+p1p+c0.1c+s
		gmt coast -R15W/35E/30N/48N -JM? -Da -Gbrown -Bafg -EES+gbisque --MAP_FRAME_TYPE=plain
	gmt inset end
gmt end show
```

PaulWessel:
When -R -J is used with gmt inset you get exactly that size and that is the projection you have in the inset.  So the only hope would be to add a margin to the inset dimensions I think.

PaulWessel:
This all depends on annotation size etc but you can slap a  -M2p/14p/2p/14p to that inset call and play with the numbers.

Esteban82:
> This all depends on annotation size etc but you can slap a -M2p/14p/2p/14p to that inset call and play with the numbers.

Ok. 

BTW, what is the default units for -M? inches? I noticed that it is not clarified in the docs.

PaulWessel:
Should be cm like any other dimensions (only exception I think is font sizes which are assumed to be in points).

PaulWessel:
It seems like a lot of work to get a nice tiny inset with annotations, e.g.:

```
gmt begin test-inset png 
    gmt coast -R-51/-48/-24/-23 -JM15c -B -Gpink
    gmt inset begin -DjBL+o0.1c -F+p1p,black -R-80/-28/-43/10 -JM2c -M3p/9p/3p/7p
        gmt coast -Wthin -Swhite -Ggray -Bafg -BwsNE --FONT_ANNOT_PRIMARY=4p --MAP_FRAME_TYPE=plain --MAP_FRAME_PEN=0.5p --MAP_TICK_LENGTH_PRIMARY=2p --MAP_ANNOT_OBLIQUE=lat_parallel
    gmt inset end
gmt end show
```

I guess I think that insets showing the entire world or a large chunk of it may not really benefit much from having separate annotations.  Do we really need to know that South America has these coordinates, when the main plot has the coordinates for the data we actually are presenting?  Perhaps a personal choice but I would not bother to add tiny tick marks, tiny fonts etc and worry about it all fitting.  But, as you see, you can if you want to.

meghanrjones:
> It seems like a lot of work to get a nice tiny inset with annotations, e.g.:
> 
> ```
> gmt begin test-inset png 
>     gmt coast -R-51/-48/-24/-23 -JM15c -B -Gpink
>     gmt inset begin -DjBL+o0.1c -F+p1p,black -R-80/-28/-43/10 -JM2c -M3p/9p/3p/7p
>         gmt coast -Wthin -Swhite -Ggray -Bafg -BwsNE --FONT_ANNOT_PRIMARY=4p --MAP_FRAME_TYPE=plain --MAP_FRAME_PEN=0.5p --MAP_TICK_LENGTH_PRIMARY=2p --MAP_ANNOT_OBLIQUE=lat_parallel
>     gmt inset end
> gmt end show
> ```
> 
> I guess I think that insets showing the entire world or a large chunk of it may not really benefit much from having separate annotations. Do we really need to know that South America has these coordinates, when the main plot has the coordinates for the data we actually are presenting? Perhaps a personal choice but I would not bother to add tiny tick marks, tiny fonts etc and worry about it all fitting. But, as you see, you can if you want to.

Good to know that it's possible, thanks.

PaulWessel:
Any other concerns before we merge this one?

meghanrjones:
> Any other concerns before we merge this one?

Looks good to me.

PaulWessel:
I just addressed @Esteban82 comment on units for margins in both inset and subplot.

Esteban82 (approved):

Esteban82:
Looks good for me too.

Esteban82:
> It seems like a lot of work to get a nice tiny inset with annotations,

I think it would be good to have an easy way to do this. I agree that is has no much sense for insets. But I think it could be good to set the dimensions of canvas in movies. In some occasions, I just want to get the animation and I don't care about the final dimension of the canvas.

PaulWessel:
Could you explain this differently? I do not understand what you mean related to canvas and movies. You mean not specifying **-C** and let it be found from main.sh as the dimensions of the cropped frame? You would have to specify dpi though, right?  So if you gave -C100c you want a 100 dots per cm movie frame of whatever size that is given your script?  This would mean (1) need to run main once to call **psconvert** and determine the bounding box size, then (2) have to assume all frames are the same size.

Esteban82:
Yes. Like with any figure. I define the map width, I then add some stuffs like annotations and colorbar outside the map and finally psconvert gives me the bounding box size (which is slighter bigger than the original width).

PaulWessel:
And what about dpi? Via -C, or default to the same as psconvert [300i]? Probably would give 4k-size movies...

Esteban82:
Mmm, I haven't think about dpi. The default 300 dpi sounds good. Or maybe with -C?/?/dpi??

PaulWessel:
I think -Cdpi is enough for that.  If you compose on a typical piece of paper (say make a plot around  24x13cm which is the HD format setup now) then 300 is pretty large.  The current scheme is 80dpc or 200dpi:

`       1080p:  1920 x 1080    24 x 13.5 cm    80.0000    9.6 x 5.4 inch   200.0000`

But, I think it is still the best default since it is what **GMT_GRAPHICS_DPU** currently is set to.

So to get this to work, post 6.3, would be to

1. Detect that user only give a dpi, or none at all (no **-C** meaning -C300)
2. If so, give a warning that we will determine the canvas size from the main script
3. Before we get to the main loop, run frame 1 once and get the bounding box from that one plot
4. Set the **-C**_width_/_height_ parameters accordingly and continue the run

One issue is the pre.sh script which may need to run to create data to even run the movie, and it may also create a background PostScript plot.  That plot needs to know the canvas size too, which we don't know yet.  SO it gets a bit messy in having to run it twice.

Esteban82:
Ok. Sounds good. I think that might be an issue with for the bounding box for the [animation 03](https://docs.generic-mapping-tools.org/dev/animations/anim03.html).

PaulWessel:
Right, so it cannot work for those types of animations, only the static frame one.  Which begs the question if all the above work is really worth the effort...

Esteban82:
I think so. I will definitely use it.

PR #5934, opened 2021-11-01
Migrate animation 01 to YouTube mp4

meghanrjones:

This PR migrates animation 01 to produce an mp4 that will be hosted on YouTube. I made some small adjustments to the animation in the process. Note that animation 02 is used as a placeholder for the YouTube link until the script is finalized.

This is the result of the current script:

https://user-images.githubusercontent.com/14077947/139735696-2781eafa-5079-4f69-b722-c6db9e33e8e6.mp4

Fixes #

**Reminders**


PaulWessel (approved):
I think those things you can commit directly to master.  Not really needing a review from a non-english speaker anyway :-)

meghanrjones:
> I think those things you can commit directly to master. Not really needing a review from a non-english speaker anyway :-)

I assume you're referring to the last commit with the typo fix? Just to confirm, you also approve of the animation update? If so, I will upload to YouTube and update the link.

PaulWessel:
Oh, sorry, I click on the commit I think which only had the spelling fix.  Not the entire PR which I am seeing now.  Give me a minute to have a look.

PaulWessel:
Looks fine.  However, given the larger animation I think we can add more frames without making it long and slow.  Instead of 20 degrees and 2 degree intervals, how about 10 and 1?
Also, I think most recent anime??.sh no longer has the shell-stuff up front that sets -Mps or -Fmp4 anymore.  Should we just remove all that?

meghanrjones:
> Looks fine. However, given the larger animation I think we can add more frames without making it long and slow. Instead of 20 degrees and 2 degree intervals, how about 10 and 1? Also, I think most recent anime??.sh no longer has the shell-stuff up front that sets -Mps or -Fmp4 anymore. Should we just remove all that?

Sounds good. Here is the result with 10 and 1:

https://user-images.githubusercontent.com/14077947/139745532-259b1b48-3373-44d3-b6d2-53c720250b74.mp4

PR #5936, opened 2021-11-02
API: Switch to method from GMT_IS_REFERENCE to GMT_IS_DUPLICATE as needed

PaulWessel:
When a user matrix is used on input to represent a grid or a cube, and the input method is set to **GMT_IS_REFERENCE**, then there are some restrictions that apply, including one cannot specify a subset, the matrix must be of the right type (float [or double depending on compilation]), must be row-oriented, and not have complex components.  If any of those are not true then we switch method to **GMT_IS_DUPLICATE** and try again.  Same for user matrices passed as data cubes.

There are no tests to check this PR, but PyGMT may one day pass a matrix by reference but want a subset.  At the very least, this PR will now handle the "wrong data type" error and turn it into a change of method.

Notes:

1. If the input matrix represents the full grid and the user passes a subset request, then it should be possible to honor that by increasing the pads so that the pad eats away the data outside the subset w/e/s/n and we still are doing read-only on the matrix.  I will need to experiment with this a bit, but that is what we do when we receive a GMT grid via **GMT_IS_REFERENCE** and want a subset.  One exception would be for periodic grids when a rotation in longitude is required (which we cannot do in read-only mode). E.g., input grid is 0/360 and subset request is -30/30 will require a change of method to **GMT_IS_DUPLICATE**.
2. The user of complex components in GMT is very limited (**grdfft** only I think, but the _GMT_FFT_* functions in the API are broader than that.  So, it is not clear yet to me if an input matrix has complex components why it could not be accepted, provided the values are arranged in the same order as internally for GMT grids.  So more experimentation here.

### DIFF
--- a/doc/rst/source/devdocs/api.rst
+++ b/doc/rst/source/devdocs/api.rst
@@ -3825,11 +3825,11 @@ allocated will be *size * n_bands*, where the latter is one of the parameters in
 .. code-block:: c
 
   struct GMT_CUBE {
-       struct GMT_GRID_HEADER *header;      /* The full GMT header for the grid */
+       struct GMT_GRID_HEADER *header;      /* The full GMT header for the cube */
        float                  *data;        /* Pointer to the float 3-D array */
        unsigned int           mode;         /* Indicates data originated as a list of 2-D grids rather than a cube */
-       double                 z_range[2];   /* Minimum/max z values (complements header->wesn) */
-       double                 z_inc;        /* z increment (complements header->inc) (0 if variable z spacing) */
+       double                 z_range[2];   /* Minimum/maximum z-dimension values (complements header->wesn) */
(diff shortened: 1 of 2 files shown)


meghanrjones (approved):
Looks good. Why repeat these lines rather than making them a function?
```
				method -= GMT_IS_REFERENCE;
				method += GMT_IS_DUPLICATE;
				start_over_method = GMT_IS_DUPLICATE;
				GMT_Report (API, GMT_MSG_INFORMATION, <message>);
				goto start_over_import_cube;
```

PaulWessel:
I don't think the goto should be hidden inside a function, so is it worth it to replace 4 lines with one? Yes, OK, will do that.

PaulWessel:
Added a short function.

Issue #5294, opened 2021-06-03
CPT range auto-stretching works differently with NetCDF and virtual reliefgrid inputs


weiji14:
**Description of the problem**

When comparing a `grdimage` plot from a regular remote file and virtual file input, while setting a specific region (e.g. Greenland/GL in the case below), the CPT range is stretched differently. Specifically, the CPT range is -3893 to 3195.5 for the remote file case, but -8592.5 to 5559 for the virtual file case. See https://github.com/GenericMappingTools/pygmt/pull/750#discussion_r644475505 for context.

| expected (correct) | actual (wrong) | diff |
|--|--|--|
| ![test_grdimage_grid_and_shading_with_xarray png -expected](https://user-images.githubusercontent.com/23487320/120717600-f0392d80-c51b-11eb-9311-2bfa279682ed.png) | ![test_grdimage_grid_and_shading_with_xarray png](https://user-images.githubusercontent.com/23487320/120717597-ee6f6a00-c51b-11eb-937a-7d458e23ff1b.png) | ![test_grdimage_grid_and_shading_with_xarray png -failed-diff](https://user-images.githubusercontent.com/23487320/120717598-efa09700-c51b-11eb-8bcd-710a5dfd5f58.png) |

~~Need to install PyGMT branch from https://github.com/GenericMappingTools/pygmt/pull/750 first using `pip install https://github.com/GenericMappingTools/pygmt/archive/grdimage-xarray-shading.zip` first before running the below:~~ Edit: just do `pip install pygmt=0.5.0`.

**Full script that generated the error**

```python
import numpy as np
import pygmt
import xarray as xr

# xarray.DataArray grid used as input for shading (-I) parameter
longitude = np.arange(0, 360, 1)
latitude = np.arange(-89, 90, 1)
x = np.sin(np.deg2rad(longitude))
y = np.linspace(start=0, stop=1, num=179)
data = y[:, np.newaxis] * x

xrgrid = xr.DataArray(
    data,
    coords=[
        ("latitude", latitude, {"units": "degrees_north"}),
        ("longitude", longitude, {"units": "degrees_east"}),
    ],
    attrs={"actual_range": [-1, 1]},
)

# Baseline (correct) case using @earth_relief_01d_g reliefgrid
fig = pygmt.Figure()
fig.grdimage(
    grid="@earth_relief_01d_g", region="GL", cmap="geo", shading=xrgrid, verbose="i"
)
fig.colorbar()
fig.show()


[...]

PR #5940, opened 2021-11-03
Reset header zmin and zmax before getting grid subregion

meghanrjones:

This is a partial fix for the bug reported in https://github.com/GenericMappingTools/gmt/issues/5294. Previously the header zmin and zmax were not reset prior to comparing to data values in the grid subregion, such that the min/max for the entire grid was always chosen.

There's still an additional issue where the region extracted for a virtual file might be offset by one row relative to the region extracted for an actual file, leading to a slight difference in the z-range:
```
grdimage [INFORMATION]: Auto-stretching CPT file geo to fit data range -3893 to 3195.5

grdimage [INFORMATION]: Auto-stretching CPT file geo to fit data range -3824.5 to 3195.5
```

 

Addresses #5294 

**Reminders**


PaulWessel (approved):
Yep, good one.

PR #5947, opened 2021-11-04
Fix wrap-around error in indexing for matrix-to-grid input

PaulWessel:
See #5294 for background.  When desired subregion **-R**_w/e/s/n_ and a grid's internal region are off by 360 degrees relative to each other, we need to reconcile that shift before we compute column numbers since gridline vs pixel registration throws a wrench otherwise (we are off by one column).  This PR implements the following:

1. Adjusts the internal grid header _wesn_ to be compatible with the desired sub region _wesn_ before computing subset indices. The same fix also applies to _gmtapi_import_cube_.
2. Updates the object's _method_ variable when we switch from reference to duplicate (I don't think this caused any issues elsewhere but it is the right thing to).

CLI tests pass, and the pyGMT example returns the correct z-range. Please check that this PR works well for PyGMT, @meghanrjones.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -4874,10 +4874,12 @@ unsigned int gmt_whole_earth (struct GMT_CTRL *GMT, double we_in[], double we_ou
 	return 1;	/* Different central meridians */
 }
 
-GMT_LOCAL unsigned int gmtapi_switch_method (struct GMTAPI_CTRL *API, unsigned int *method, char *message) {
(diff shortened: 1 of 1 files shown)


meghanrjones (approved):
Thanks, this fixes the result for ``pygmt/tests/test_grdimage.py::test_grdimage_shading_xarray``. The only remaining issue here is this example: https://github.com/GenericMappingTools/gmt/issues/5294#issuecomment-959859425, but that's unrelated to this PR.

PaulWessel:
Sorry, not quite sure what you mean (after reading that issue).  What is not quite working?

meghanrjones:
> Sorry, not quite sure what you mean (after reading that issue). What is not quite working?

I'll post an explanation in that issue. This PR is good to merge.

PR #5949, opened 2021-11-05
Look for the spatial_ref attribute among all variables

### LINKED ISSUE #5928: GMT can't fetch referencing system when the "spatial_ref" is an attribute of other than the global "grid_mapping"
```
julia> G = gmtread("HLS.S30.T10TEK.2021089T184941.v1.5_EVI.grd")
julia> G.proj4
""
julia> G.wkt
""
```

```
gdalinfo HLS.S30.T10TEK.2021089T184941.v1.5_EVI.grd
...
transverse_mercator#spatial_ref=PROJCS["unknown",GEOGCS[...
...
```

As an example of one that works
```
gdalinfo evi.grd
...
Metadata:
  grid_mapping#spatial_ref=PROJCS["unknown",
```

[HLS.S30.T10TEK.2021089T184941.v1.5_EVI.zip](https://github.com/GenericMappingTools/gmt/files/7437758/HLS.S30.T10TEK.2021089T184941.v1.5_EVI.zip)

PaulWessel:
Hedge against other people's dumb ideas, part LXXIV. .. See L557-561 (the rest are just replacing -1 with **GMT_NOTSET**).

Closes #5928.

### DIFF
--- a/src/gmt_nc.c
+++ b/src/gmt_nc.c
@@ -426,6 +426,7 @@ GMT_LOCAL int gmtnc_grd_info (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *head
 	int j, err, has_vector, has_range, registration, var_spacing = 0;
 	int old_fill_mode, status;
 	unsigned int d_off = (cube) ? 1 : 0;
+	size_t len;
 	double dummy[2], min, max;
 	char dimname[GMT_GRID_UNIT_LEN80], coord[GMT_LEN8];
 	nc_type z_type;
@@ -433,7 +434,7 @@ GMT_LOCAL int gmtnc_grd_info (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *head
 	struct GMT_GRID_HEADER_HIDDEN *HH = gmt_get_H_hidden (header);
 
 	/* Dimension ids, variable ids, etc.. */
-	int i, ncid, z_id = -1, ids[5] = {-1,-1,-1,-1,-1}, gm_id = -1, dims[5], nvars, ndims = 0;
+	int i, ncid, z_id = GMT_NOTSET, ids[5] = {-1,-1,-1,-1,-1}, gm_id = GMT_NOTSET, dims[5], nvars, ndims = 0;
 	size_t lens[5], item[2];
 
 	/* If not yet determined, attempt to get the layer IDs from the variable name */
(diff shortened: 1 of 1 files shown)


joa-quim:
This works but it can give a wrong result if the file has more than one 2D variable with different ``spatial_ref`` attribs. The one we are interested in, if it exists, is that of the specific 2D variable.

PaulWessel:
OK, will check that it is a 2-D variable.  But cannot data cubes have 3 and we are pulling out one layer via grid?var[0] or something?

joa-quim:
Yes, and one more reason to get the ``spatial_ref`` attrib of the variable that we are actually reading. Not of any other variable.

joa-quim:
Nope, is more dumb. The variable being read has (when it does) an attribute called ``grid_mapping`` whose value is the name of a global attribute that has an attribute called ``spatial_ref`` whose value is the CRS

image

PaulWessel:
So, maybe this recipe?

1. First we find (as we always do) the layer/grid to read.  It has an netcdf ID _z_id_.
2. Check if _z_id_ has an attribute called _spatial_ref_
3. If not, check if _z_id_ has an attribute _grid_mapping_, and if it does, get that as name of _variable_ whose attribute is _spatial_ref_
4. If this fails, check if the file has a variable called _grid_mapping_, and if so get its _spatial_ref_ attribute

I think we only did (4) before this PR.  Did I miss any steps and order?

joa-quim:
I think 2. never happens. What we have, again I think, is that is that the *z_id* has an attribute ``grid_mapping`` that may have or not a value different from its name, but it's always the attribute value that is the name of the global attribute containing the CRS. Next fig is the situation that we currently deal with.

image

PaulWessel:
OK, so do 1, 3, 4 only then.  About to commit.

PaulWessel:
To be clear, we then never loop over all variables.  We only do 1,3,4 as described, yes?

joa-quim:
Yes but I think *4* is included in *3*

PaulWessel:
Will debug first and see, but 4 is a global attribute not tied to a variable, so separate I think

joa-quim:
But they are always global attributes. Se my images.

PaulWessel:
OK, check the latest commit.

PaulWessel:
On gmt_nc.c line 648, after we have found _gm_id_, we do this

`gmt_M_err_trap (nc_inq_attlen (ncid, gm_id, "spatial_ref", &len));	/* Get attrib length */`

Might it ever be the case when the variable gm_id would _not_ have an attribute called spatial_ref? If so, we probably should skip and not give error, like here?  You can see I actually check earlier on line 562, so could do this here instead to avoid an error.

joa-quim:
It's in a separate case probably because at 562 we are under the condition *read or update* (if (job == 'r' || job == 'u')), whilst later is for the *read* case only.

PaulWessel:
I updated it to check that spatial_ref is found before copying into the header.  Just one extra layer of safety.

joa-quim (approved):
Gona sleep.

PR #5964, opened 2021-11-09
Ensure external use of blockm can set many fields

### LINKED ISSUE #5963: Multi grids otput for the block* are broken in Julia/MEX
Tests like this now fail (for some time but haven't pay attention to it before)
```
G,L = blockmode(d, region=[0 2 0 2], inc=1, fields="z,l", reg=true);
```
The problem is that ``gmt_api.c``, L#12940 have this
```
	if (opt->arg[0]) {	/* Gave -A: Determine how many output grids are requested */
		for (k = 1, len = 0; len < strlen (opt->arg); len++) if (opt->arg[len] == ',') k++;
	}
	if ((opt = GMT_Find_Option (API, 'G', *head))) {	/* This is a problem unless -G actually sent in a file name */
		if (opt->arg[0] == '\0') {	/* Cannot just give -G here */
			GMT_Report (API, GMT_MSG_ERROR, "GMT_Encode_Options: %s cannot set -G when called externally\n", module);
			return_null (NULL, GMT_NOT_A_VALID_OPTION);
		}
		else	/* Gave a (presumably) file argument, no need to add -G? */
			k = 0;
	}
	else	/* No -A or -G; default is to just add the z grid via -G?  */
		k = 1;
```
So **k** is reset such that it never > 1

PaulWessel:
See #5963 for background.  This PR should fix it, pending review by @joa-quim to close #5963.

### DIFF
--- a/src/blockmean.c
+++ b/src/blockmean.c
@@ -182,7 +182,7 @@ static int parse (struct GMT_CTRL *GMT, struct BLOCKMEAN_CTRL *Ctrl, struct GMT_
 					Ctrl->E.mode = BLK_MODE_SIMPLE;
 				break;
 			case 'G':	/* Write output grid(s) */
-				n_errors += gmt_M_repeated_module_option (API, Ctrl->G.active);
+				if (!API->external) n_errors += gmt_M_repeated_module_option (API, Ctrl->G.active);
 				Ctrl->G.active = true;
 				if (!GMT->parent->external && Ctrl->G.n) {	/* Command line interface */
 					GMT_Report (API, GMT_MSG_ERROR, "-G can only be set once!\n");
--- a/src/blockmedian.c
+++ b/src/blockmedian.c
@@ -201,7 +201,7 @@ static int parse (struct GMT_CTRL *GMT, struct BLOCKMEDIAN_CTRL *Ctrl, struct GM
 					n_errors++;
 				break;
 			case 'G':	/* Write output grid(s) */
-				n_errors += gmt_M_repeated_module_option (API, Ctrl->G.active);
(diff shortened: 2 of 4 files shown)


joa-quim:
New error now
```
julia> G,L = blockmean(d, region=[0 2 0 2], inc=1, fields="z,l", reg=true);
blockmean -Az,l -I1 -rp -R0/2/0/2 -G? -G? ?
blockmean [ERROR]: Option -G: Given more than once (offending option is -G@GMTAPI@-S-O-G-G-G-N-000015)
ERROR: Something went wrong when calling the module. GMT error number = 72
```
Looks like the ``blocks``  became overzealous on the number of **-G's**?

PaulWessel:
Well the two -G? Are needed but som check gets in the way

Paul Wessel, Professor
Dept. of Earth Sciences
SOEST, U of Hawaii at Manoa

joa-quim:
> Well the two -G?

Yes. One for each -A option

```
while (k) {	/* Add -G? option k times */
```

PaulWessel:
Well, related to the much more recent

`gmt_M_repeated_module_option (API, Ctrl->G.active);`

See if this commit helps:

`if (!API->external) n_errors += gmt_M_repeated_module_option (API, Ctrl->G.active);`

joa-quim:
Yes, that fixes it but needs to be applied to the other blocks as well.

joa-quim:
> Well, related to the much more recent

> gmt_M_repeated_module_option (API, Ctrl->G.active);

I told you that sooner or later that would bite me.

PaulWessel:
> Yes, that fixes it but needs to be applied to the other blocks as well.

OK, coming.

PaulWessel:
OK, done.  Please check and approve if OK.

joa-quim (approved):

Issue #5630, opened 2021-08-11
histogram -di does not set values to NaN


meghanrjones:
**Description of the problem**

The **-di***nodata* value does not replace values equal to *nodata* with *nan*, at least when used in the histogram module.

**Full script that generated the error**

```
#!/bin/bash

echo "0" > data.txt
echo "1" >> data.txt
echo "2" >> data.txt
gmt histogram data.txt -Bafg -R-0.5/2.5/0/2 -T-0.5/2.5/1 -W1p -Gred3 -di0 -png test
```

**Full error message**

No error message
**Actual outcome**
![test](https://user-images.githubusercontent.com/14077947/129090492-dab4e6aa-8312-498a-8b4f-e21fa4edc373.png)

**Expected outcome**

No counts at zero

**System information**

* Operating system: macOS
* GMT version (`gmt --version`): 6.3.0

PaulWessel:
While certainly not documented well, I believe **-d** only affects the z column, which is not used here.  It originated in tools like **xyz2grd** and similar.  Perhaps it will need a **+c**_col_ modifier... [2].

meghanrjones:
> While certainly not documented well, I believe **-d** only affects the z column, which is not used here. It originated in tools like **xyz2grd** and similar. Perhaps it will need a **+c**_col_ modifier... [2].

Yes, a **+c***col* modifier seems helpful and the simplest solution.

meghanrjones:
Actually, it would be nice for this to work for multiple or all columns. So either [**+c***col*][**+a**] or [*cols*] with the usual column range notation.

PaulWessel:
A bit of looking reveals the following implementation truths.

1. On **input**, any column after the _first two_ (x, y) may be affected by **-di**:
	`if (col_no > 1 && gmt_input_is_nan_proxy (GMT, GMT->current.io.curr_rec[col_no]))	/* Input matched no-data setting, so change to NaN */`
2. On **output**, all columns that have a NaN will be changed by **-do**:
	`if (GMT->common.d.active[GMT_OUT] && gmt_M_is_dnan (val)) val = GMT->common.d.nan_proxy[GMT_OUT];	/* Write this value instead of NaNs */`

There is no trace of this being discussed in the docs, so we can choose to make this a feature or a bug :-).

[...]

PR #5966, opened 2021-11-09
Let -d set starting column via modifier

PaulWessel:
See #5630 for background discussion.  The best solution is to add an optional modifier **+c**_col_ to override the default settings for the starting column, which for backwards compatible reasons shall remain 2 for input and 0 for output.

All tests pass for me since the default remains the same.

Note: I don't understand how to get the modifier to show up in the options.rst file - it works in the explain*.rst_ files.


meghanrjones:
The remaining files that need to be updated are doc/rst/source/std-opts-classic.rst and doc/rst/source/std-opts.rst.

PaulWessel:
Seems pretty good now, no?

meghanrjones (approved):

meghanrjones:
Ran into a problem building mbsystem due to the update to ``gmt_input_is_nan_proxy`` in this PR:
```
mbgrd2obj.c:360:46: error: too few arguments to function call, expected 3, have 2
                if (gmt_input_is_nan_proxy(GMT, G->data[ij])) {
                    ~~~~~~~~~~~~~~~~~~~~~~                 ^
/Users/meghanj/Sandbox/gmt-dev/include/gmt/gmt_prototypes.h:335:17: note: 'gmt_input_is_nan_proxy' declared here
EXTERN_MSC bool gmt_input_is_nan_proxy (struct GMT_CTRL *GMT, double value, unsigned int col);
```

PaulWessel:
Thanks, will fix after breakfast

Paul Wessel, Professor
Dept. of Earth Sciences
SOEST, U of Hawaii at Manoa

PR #5967, opened 2021-11-09
Add GEOS to the list of optional dependencies

meghanrjones:

This PR adds GEOS to the list of dependencies so that it's documented in case of problems such as https://github.com/GenericMappingTools/pygmt/issues/1600.

Fixes #

**Reminders**


PaulWessel:
I guess we should also update the wiki page for dependency installs.

PR #5969, opened 2021-11-09
Add geos to install-dependencies-macos.sh

meghanrjones:

The macOS docs build is failing due to a problem with GEOS (https://github.com/GenericMappingTools/gmt/runs/4157819081?check_suite_focus=true). This PR adds GEOS as an explicit dependency to try to resolve the issue.

Probably should to the same for install-dependencies-linux.sh and install-dependencies-windows.sh, but I am not sure which specific package should be added based on https://trac.osgeo.org/geos/#Binaries.

Fixes #

**Reminders**

### DIFF
--- a/ci/install-dependencies-macos.sh
+++ b/ci/install-dependencies-macos.sh
@@ -17,7 +17,7 @@ PACKAGE="${PACKAGE:-false}"
 
 # packages for compiling GMT
 # cmake is pre-installed on GitHub Actions
-packages="ninja curl pcre2 netcdf gdal fftw ghostscript"
+packages="ninja curl pcre2 netcdf gdal geos fftw ghostscript"
 
 # packages for build documentation
 if [ "$BUILD_DOCS" = "true" ]; then


PaulWessel (approved):

meghanrjones:
Unfortunately, this did not fix the packaging error in the macOS CI - https://github.com/GenericMappingTools/gmt/runs/4158509171?check_suite_focus=true. The last passing check for the docs workflow was prior to GEOS 3.10 release.

PaulWessel:
Hm, not sure what that means. How is the setup in the cloud different from a local setup? My libgmt.dylib says /opt/local/lib/libgeos_c.1.dylib so no @rpath there.

meghanrjones:
> Hm, not sure what that means. How is the setup in the cloud different from a local setup? My libgmt.dylib says /opt/local/lib/libgeos_c.1.dylib so no @rpath there.

I'm not sure either. The CI config uses brew, here's from the logs:
```
*  GEOS library               : /usr/local/Cellar/geos/3.10.0/lib/libgeos_c.dylib
*  GEOS include dir           : /usr/local/include
```

PR #5970, opened 2021-11-09
Let ex49 use mean for crustal ages

PaulWessel:
The data distribution of crustal ages within an age range is almost uniform, making a mode estimator a bad choice and very sensitive to noise (the Intel and arm hardware differ by over 1 Myr).  We fix this by asking for the mean crustal age instead which is a unique estimator.  This requires an update the PostScript plot so I also fixed some inconsistencies with the label and colorbar (now smaller, both have same shade box, etc.


doc/examples/ex49/ex49.sh
```diff
@@ -26,7 +26,7 @@ gmt begin ex49
 	# WHy do we need the -R below? otherwise it fails to work
```
meghanrjones, line 26: Is there an answer to this question?

PaulWessel: Hm, let me look.

PaulWessel: I think the answer is this: Note that this is not a subplot but old-fashioned vertical shifting, yet it is modern mode.

We maintain history for two sorts of -R: plot  (RP) and data processing (RG).  Here, we use both and have already plotted a grid in line 16, so RP is set in the history.  Then we set RG by running the xyz2grd command. The next command is another grdimage call with a 12c upwards shift (so you and I know we are starting a new plot, probably).  However, without repeating the desired new plot region, GMT will search and find the current plot region, which is not what we want.  So the only way here is to tell GMT to explicitly use the new region.  Yes, a bit clumsy, but all because I did not make a subset out of this.

So perhaps I should rewrite this using subset then to avoid this dumb situation?

meghanrjones: Yes, using subplots would be better. Do you know if the RP and RG distinction is documented anywhere? If not, we should add it to https://docs.generic-mapping-tools.org/dev/cookbook/features.html#gmt-modern-mode-hierarchical-levels.

PaulWessel: I will look at the docs (this is under the hood stuff). Finally updated the scripts to use subplot and made a bunch of minor changes to simplify.

meghanrjones:
I think the -Q got left out of the latest commit. 

This statement is outdated: "We use module grdsample to convert the age grid to the same pixel registration used by the depth grid."

PaulWessel:
Thanks, I was fighting some stashed changes that caused a conflict...

meghanrjones:
Just need a final .PS version and then we can put this to rest.

meghanrjones (approved):
Such a wondrous thing:

image

PR #5971, opened 2021-11-10
Document the two histories of domains

PaulWessel:
Without getting too technical, there is a different treatment for plot region and data regions.  See if this makes sense.  I put it under the command line history section.

### DIFF
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -446,7 +446,7 @@ which often are the same.  When a plot is specified without providing
 a region then we look for a previous plot region in the history first, and
 if it is not found then we look for the processing domain to use instead.  However,
 if a data-processing module is not given a region then we only look
-for a previous data domain; we never substitute a plot domain in that case.
+for a previous processing domain; we never substitute a plot domain in that case.
 
 Usage messages, syntax- and general error messages
 --------------------------------------------------
--- a/doc/rst/source/cookbook/features.rst
+++ b/doc/rst/source/cookbook/features.rst
@@ -441,7 +441,7 @@ do not need to repeat the region and projection information, as shown here::
 
 Thus, the chosen options remain in effect until you provide new option
 arguments on the command line.  **Note**: We keep track of two types of regions,
-One is the domain used for a map and one is the domain used for a grid.
+One is the domain used for a map and one is the domain used for a grid,
 which often are the same.  When a plot is specified without providing
 a region then we look for a previous plot region in the history first, and
 if it is not found then we look for a grid domain to use instead.  However,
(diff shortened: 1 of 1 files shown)


doc/rst/source/cookbook/features.rst
```diff
@@ -440,7 +440,13 @@ do not need to repeat the region and projection information, as shown here::
      gmt end show
 
 Thus, the chosen options remain in effect until you provide new option
-arguments on the command line.
+arguments on the command line.  **Note**: We keep track of two types of regions,
+One is the domain used for a map and one is the domain used for a grid.
```
meghanrjones, line 444:
```suggestion
One is the domain used for a map and one is the domain used for a grid,
```

doc/rst/source/cookbook/features.rst
```diff
@@ -440,7 +440,13 @@ do not need to repeat the region and projection information, as shown here::
...
+which often are the same.  When a plot is specified without providing
+a region then we look for a previous plot region in the history first, and
+if it is not found then we look for a grid domain to use instead.  However,
```
meghanrjones, line 447: So based on the grid domain description, if a plot region is not found a table region will not work (e.g., blockmean)?

PaulWessel: What I mean to say is that RG will be used if there is no RP, but RP will never be used if there is no RG.

doc/rst/source/cookbook/features.rst
```diff
@@ -440,7 +440,13 @@ do not need to repeat the region and projection information, as shown here::
      gmt end show
 
 Thus, the chosen options remain in effect until you provide new option
-arguments on the command line.
+arguments on the command line.  **Note**: We keep track of two types of regions,
+One is the domain used for a map and one is the domain used for a grid,
```
meghanrjones, line 444:
```suggestion
One is the domain used for a map and one is the domain used for processing,
```

doc/rst/source/cookbook/features.rst
```diff
@@ -440,7 +440,13 @@ do not need to repeat the region and projection information, as shown here::
...
+which often are the same.  When a plot is specified without providing
+a region then we look for a previous plot region in the history first, and
+if it is not found then we look for a grid domain to use instead.  However,
```
meghanrjones, line 447:
```suggestion
if it is not found then we look for the processing domain to use instead.  However,
```

doc/rst/source/cookbook/features.rst
```diff
@@ -440,7 +440,13 @@ do not need to repeat the region and projection information, as shown here::
...
+if a data-processing module is not given a region then we only look
+for a previous data domain; we never substitute a plot domain in that case.
```
meghanrjones, line 449:
```suggestion
for a previous processing domain; we never substitute a plot domain in that case.
```

meghanrjones (approved):
A couple suggestions based on my interpretation that RG can be set in the history regardless of whether the region was set on a table or grid.

PR #5972, opened 2021-11-10
WIP Improve grdcall.c string breaking

PaulWessel:

I changed the quote checking to look for single quotes instead.  This keeps the quoted args to `-a_srs `as one item, but something is still not quite right.  Perhaps you know what to do next so you can add to this branch:

```
gmt grdgdal t.jpg -Atranslate -Gnew.tiff -F"-mo TIFFTAG_XRESOLUTION=300 -mo TIFFTAG_YRESOLUTION=300 -a_srs '+proj=stere +lat_0=90 +lon_0=35 +k=0.9996 +x_0=0 +y_0=0 +units=m +a=6378137.000 +b=6356752.314 +ellps=WGS84 +datum=WGS84 +units=m +no_defs' -co COMPRESS=LZW -co TILED=YES"  -M
ERROR 1: PROJ: proj_create: unrecognized format / unknown name
ERROR 1: Failed to process SRS definition: '+proj=stere +lat_0=90 +lon_0=35 +k=0.9996 +x_0=0 +y_0=0 +units=m +a=6378137.000 +b=6356752.314 +ellps=WGS84 +datum=WGS84 +units=m +no_defs'
```

### DIFF
--- a/src/gmt_gdalcall.c
+++ b/src/gmt_gdalcall.c
@@ -108,7 +108,7 @@ GMT_LOCAL GDALDatasetH gdal_vector (struct GMT_CTRL *GMT, char *fname) {
 
 /* ------------------------------------------------------------------------------------------------------------ */
 GMT_LOCAL char ** breakMe(struct GMT_CTRL *GMT, char *in) {
-	/* Breake a string "-aa -bb -cc dd" into tokens "-aa" "-bb" "-cc" "dd" */
+	/* Break a string "-mo TIFFTAG_XRESOLUTION=300  -a_srs '+proj=stere +lat_0=90'" into tokens */
 	/* Based on GMT_Create_Options() */
 	unsigned int pos = 0, k, n_args = 0;
 	bool quoted;
@@ -122,11 +122,11 @@ GMT_LOCAL char ** breakMe(struct GMT_CTRL *GMT, char *in) {
 	txt_in = strdup (in);
 	args = gmt_M_memory (GMT, NULL, n_alloc, char *);
 
-	/* txt_in can contain options that take multi-word text strings, e.g., -B+t"My title".  We avoid the problem of splitting
-	 * these items by temporarily replacing spaces inside quoted strings with ASCII 31 US (Unit Separator), do the strtok on
+	/* txt_in can contain options that take multi-word text strings, e.g., '+proj=stere +lat_0=90'.  We avoid the problem of splitting
+	 * these items by temporarily replacing spaces inside single quoted strings with ASCII 31 US (Unit Separator), do the strtok on
 	 * space, and then replace all ASCII 31 with space at the end (we do the same for tab using ASCII 29 GS (group separator) */
 	for (k = 0, quoted = false; txt_in[k]; k++) {
-		if (txt_in[k] == '\"') quoted = !quoted;	/* Initially false, becomes true at start of quote, then false when exit quote */
+		if (txt_in[k] == '\'') quoted = !quoted;	/* Initially false, becomes true at start of quote, then false when exit quote */
 		else if (quoted && txt_in[k] == '\t') txt_in[k] = GMT_ASCII_GS;
 		else if (quoted && txt_in[k] == ' ')  txt_in[k] = GMT_ASCII_US;
 	}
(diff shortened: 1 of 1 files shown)


PaulWessel:
Any thoughts on this, @joa-quim ? We also call gdal_translate in grdcut to write a multi-band TIFF and there we don't need --a_srs so perhaps it might be easier to get grdgdal to do that job?  I like the idea of avoiding system calls.

joa-quim:
In the single quotes case when ``'+proj=stere +lat_0=90'`` becomes ``'+proj=stere_+lat_0=90'`` shouldn't we get rid of the ``'`` too? Otherwise they are passed verbatim to GDAL encapsulating the proj string.

PaulWessel:
That may be the problem.  I could try that.  meanwhile see #5983 .

PaulWessel:
See if this works for you.

joa-quim:
Looks like the single quote recipe is for *nix only. The Patagonia example fails

```
gmt coast -JS-55/-60/15c -R-77/-55/-57.5/-48r -Di -Gred -Bg2 --MAP_FRAME_TYPE=inside -ps patagonia
psconvert patagonia.ps -Tt -W+g -V

...
psconvert [INFORMATION]: Proj4 definition: +proj=stere +lat_0=-60 +lon_0=-55 +k=0.9996 +x_0=0 +y_0=0 +units=m +a=6378137.000 +b=6356752.314 +ellps=WGS84 +datum=WGS84 +units=m +no_defs
psconvert [INFORMATION]: The grdgdal command:
"patagonia.tif" -Atranslate -M -G"patagonia.tiff" -F"-mo TIFFTAG_XRESOLUTION=300 -mo TIFFTAG_YRESOLUTION=300 -a_srs "+proj=stere +lat_0=-60 +lon_0=-55 +k=0.9996 +x_0=0 +y_0=0 +units=m +a=6378137.000 +b=6356752.314 +ellps=WGS84 +datum=WGS84 +units=m +no_defs" -co COMPRESS=LZW -co TILED=YES"
grdgdal [ERROR]: Cannot find file +lat_0
grdgdal [ERROR]: Cannot find file +lat_0
grdgdal [ERROR]: Cannot find file +lon_0
grdgdal [ERROR]: Cannot find file +lon_0
grdgdal [ERROR]: Cannot find file +k
...
```

PaulWessel:
Maybe you could add  a WIN33 ifdef to pick the quote needed? I am off for a
few hours.

Paul Wessel, Professor
Dept. of Earth Sciences
SOEST, U of Hawaii at Manoa

joa-quim:
This doesn't work for you either. Shit happens earlier at create options. The proj string was already broken into pieces.

image

PaulWessel:
Create_Options are concerned with stuff like -BWesn+t"My title" -Bxa5+l"Some label" I think.  However, when grdgdal.c was added, it actually expects a -Margs option where the args are in double quotes and may contains sub-options in single quotes.  Is this a correct description of life?

If so, yes, then Create_Option needs to be smarter or grdgdal needs to intervene somehow.  But first, let me know if the above is correct.

joa-quim:
It's not **-M**, is **-F**

> List of GDAL options for the selected program in -A wrapped in double quotes.

From this, I assume yes. But no single quotes, only double. Or if *nix insists all singles on *nix and all doubles on Win.

PaulWessel:
So on Windows, if you want to pass a series of options via **-F**, you need to quote the whole string, but if there are the -a_srs suboptions, you cannot also use double quotes for those, are you?  Don't you do

`-F"-mo TIFFTAG_XRESOLUTION=300 -a_srs '+proj=stere +lat_0=-60 +lon_0=-55 +k=0.9996'"`

 ?

joa-quim:
I think this is a not anticipated case. We need to be able to break up the **-F** string in tokens and the ``+proj=...`` is one of them. In **-J** we let it be "+proj=stere +lat_0=-60 +lon_0=-55 +k=0.9996" or "+proj=stere+lat_0=-60+lon_0=-55+k=0.9996" (no spaces). One solution would be to do the same, that is, no spaces and then we introduce them. A simpler one seems to be to wrap the +proj4 in single quotes. But this is OS dependent. Would work in present case because the command is generated in ``psconvert`` but not in the general case where the shells sneak in to f things.

PaulWessel:
Thanks, forgot about how the +proj stuff was handed: We expect no spaces and then they are added in the J internal strings passed out.  Seems like we just need to remove them before passing to the next gmt module.

PaulWessel:
We should revisit this after 6.3 has been released to see if we can eliminate gdal_translate calls in psconvert and grdcut.

PR #5974, opened 2021-11-10
Handle deprecation of function for MB-system

PaulWessel:
See #5966 for background.  Solution is to rename the function and add a backwards compatible macro for the old function in gmt_mb.h to handle this change. We just pass col = 0 sa the missing third argument since any column checking was dong outside the function before 6.3.0.

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -48,7 +48,7 @@
  *  gmt_check_z_io:     Fill in implied missing row/column
  * gmt_set_geographic
  * gmt_set_cartesian
- * gmt_input_is_nan_proxy
+ * gmt_input_col_is_nan_proxy
  * gmtlib_write_tableheader
  * gmt_fgets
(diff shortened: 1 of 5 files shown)


meghanrjones:
Build error:
```
mbgrd2obj.c:360:7: error: use of undeclared identifier 'value'
                if (gmt_input_is_nan_proxy(GMT, G->data[ij])) {
                    ^
/Users/meghanj/Sandbox/gmt-dev/include/gmt/gmt_mb.h:35:37: note: expanded from macro 'gmt_input_is_nan_proxy'
#define gmt_input_is_nan_proxy (GMT,value) gmt_input_col_is_nan_proxy (GMT,value,0)    /* Deprecated in GMT 6.3.0 */
```

PaulWessel:
Now builds for me (!)

meghanrjones (approved):

PR #5980, opened 2021-11-11
Add -o to histogram and rose

### LINKED ISSUE #5979: Add -o option for rose and histogram to retrieve information.
If a run this code, I get this info

```
gmt pshistogram  @v3206_06.txt -F -T250 -I
-5932   -24     0.0920104278485 20.3036344119
```

I think it would be good to have the option to only output one value (like the maximum value of Y) so I can create a variable. Something like this: 

`max=$(gmt pshistogram  @v3206_06.txt -F -T250 -I -o3)`

The same case happens for gmt rose -I (which gives all these values  n, mean az, mean r, mean resultant length, max bin sum, scaled mean, and linear length sum).

I manage to get the variable saving the info into a file and then using `gmt info -C -o`. So I don't really need this feature but I think that maybe it would be good to have it.

PaulWessel:
Both of these plotting modules have an inquiry mode (**-I**) were they report a record of information, so it may be helpful if -o is available for selecting just a subset of this information.  This PR adds **-o** to both modules.  Closes #5979.

### DIFF
--- a/doc/rst/source/histogram.rst
+++ b/doc/rst/source/histogram.rst
@@ -139,7 +139,7 @@ Optional Arguments
     Inquire about min/max x and y after binning. The *xmin xmax ymin
     ymax* is output; no plotting is done. Append **o** to output an
     ASCII table of the resulting x,y data instead. Upper case **O** will
-    output all x,y bin data even when y == 0. **Note**: You may use **-i**
+    output all x,y bin data even when y == 0. **Note**: You may use **-o**
     to select a subset from this record.
 
 .. _-Jz:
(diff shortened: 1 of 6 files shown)


doc/rst/source/histogram.rst
```diff
@@ -138,7 +139,8 @@ Optional Arguments
     Inquire about min/max x and y after binning. The *xmin xmax ymin
     ymax* is output; no plotting is done. Append **o** to output an
     ASCII table of the resulting x,y data instead. Upper case **O** will
-    output all x,y bin data even when y == 0.
+    output all x,y bin data even when y == 0. **Note**: You may use **-i**
```
Esteban82, line 142: Is it correct here -i? Or should be -o?

PaulWessel: Sorry, -o.  Fixed.

Esteban82 (approved):

PR #6001, opened 2021-11-13
Fix inaccuracy in -Gvalue+n in project

PaulWessel:
The ability to specify the desired number of points was recently added to **project**.  I had initially used _gmt_great_circle_dist_meter_ to compute the length between C and E but this value depends on the chosen _ellipsoid_ whereas the rest of **project** is strictly spherical.  This then lead to the wrong number of points on output because the increment was not set correctly.  This PR uses the length computed within  **project** to obtain the increment when a number is given.  No tests were affected but my intended user for this feature caught the error...

As an example, this command gave 719 points in master but the correct 720 with this fix:

`gmt project -C157.8583W/21.3069N -E13.0550/47.8095 -G720+n -Q | wc`


Esteban82 (approved):
```
gmt project -C157.8583W/21.3069N -E13.0550/47.8095 -G720+n -Q | wc
    720    2160   30530
```

PR #6003, opened 2021-11-14
Typo. Delete extra "See"

Esteban82:


joa-quim:
We should not waste electricity with simple PRs like this. Just commit it directly.

PaulWessel (approved):

Esteban82:
> We should not waste electricity with simple PRs like this. Just commit it directly.

I agree but I think I can't do that. I will keep in mind.

Issue #6006, opened 2021-11-15
Bug with blockmean -Sn -I10k and -G


Esteban82:
I want to create a grid with the amounts of earthquakes in a region. These works:

```
gmt blockmean @usgs_quakes_22.txt -Sn -I10k -R0/40/0/40 -C -i2,1 -h | gmt xyz2grd -I10k -R0/40/0/40 -Gcount.nc
gmt blockmean @usgs_quakes_22.txt -Sn -I10k -R0/40/0/40 -C -i2,1 -h

```
When I add the -G option it fails:
```
gmt blockmean @usgs_quakes_22.txt -Sn -I10k -R0/40/0/40 -C -i2,1 -h -Gcount.nc
Segmentation fault
```

If I use -I10m it works.

anbj:
Just out of curiosity; is the `blockmean` step necessary? Wouldn't `xyz2grd -An [..]` do the same?

Esteban82:
Mmm I am not sure. I will check it out.

meghanrjones:
I'm having some difficulty reproducing the issue. I have tried your code with the main branch on macOS and Ubuntu and do not get a segfault. What is your `gmt --version` and using what shell and OS?

I am also wondering if it makes sense for **-C** to be allowed with **-G**. To me, it seems that the coordinates should be set by **-R**, **-I**, and **-r** for the **-G** case. What do you think?

PaulWessel:
I also cannot reproduce a problem.

Yet, agree that **-C** makes no sense when **-G** is used.  I can add a "ignored" message for this.  The same pertains to **-Q** in median and mode.  In fact, I will check if **blockmedian** **-G** still does the harder work of computing useless median x,y when as @meghanrjones says these are now grid nodes regardless.

anbj:
I'm with @Esteban82:

```
$ gmt blockmean @usgs_quakes_22.txt -Sn -I10k -R0/40/0/40 -C -i2,1 -h -Gcount.nc
ERROR: Caught signal number 11 (Segmentation fault) at
/usr/local/lib/libgmt.so.6(GMT_blockmean+0x2275)[0x7f9eb590ab75]
[0x55a92aeec558]
Stack backtrace:
/usr/local/lib/libgmt.so.6(sig_handler_unix+0xc0)[0x7f9eb583cd00]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x14140)[0x7f9eb56c6140]
[...]

PR #6015, opened 2021-11-15
Add debug information about Xcode diagnostics

PaulWessel:
As #6006 reminded us, we must often set the optional diagnostics settings in Xcode for it to be able to reproduce SEGV crashes.  This PR adds text and a picture of the Diagnostic tab to the debug documentation.


meghanrjones (approved):
Great, thanks!

PR #6039, opened 2021-11-19
Improve the argument for GMT_TRACK_MEMORY

PaulWessel:
This is an environmental parameter only used by developers to debug memory issues.  It used to be 0, 1, 2 or an integer but this prevented passing the integers 0-2 for tracking.  This PR uses ON, OFF, LOG and integers to handle this case.

No effect on anything but debugging.  Approve anytime but we will wait to merge after 6.3.0 release.


PaulWessel:
This can now be approved anytime.

PaulWessel:
Second thought, only I care about or use this so I will force the merge.

Issue #6043, opened 2021-11-20
GMT 6.3.0 fails to build on big endian architectures (undefined reference to `bswap32')


sebastic:
GMT 6.3.0 fails to build on big endian architectures:
```
cd /<<PKGBUILDDIR>>/debian/build/src && /usr/bin/cmake -E cmake_link_script CMakeFiles/gmt.dir/link.txt --verbose=1
/usr/bin/cc -std=gnu99 -fstrict-aliasing -g -O2 -ffile-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -O2 -g -DNDEBUG -Wl,-z,relro -Wl,-z,now -rdynamic CMakeFiles/gmt.dir/gmt.c.o -o gmt  -Wl,-rpath,/<<PKGBUILDDIR>>/debian/build/src: libgmt.so.6.3.0 -lnetcdf -lcurl -lgdal -lgeos_c -lpcre -lfftw3f -lfftw3f_threads -llapack -lblas libpostscriptlight.so.6.3.0 -lz -lm -ldl 
/usr/bin/ld: libgmt.so.6.3.0: undefined reference to `bswap32'
```
[Full buildlog](https://buildd.debian.org/status/fetch.php?pkg=gmt&arch=s390x&ver=6.3.0%2Bdfsg-1&stamp=1637390684&file=log)

PR #6044, opened 2021-11-20
Include gmt_common_byteswap.h for bswap32.

sebastic:
`src/grd2xyz.c` uses `bswap32` in 6.3.0 but failed to include the header which defines it.

This patch was tested on the Debian s390x porterbox.

Fixes: #6043


PaulWessel (approved):
Thanks, new feature and no bigendian machines in sight...

PR #6045, opened 2021-11-20
Internal index for custom gridinterval was wrong

PaulWessel:
See [forum post](https://forum.generic-mapping-tools.org/t/custom-axes-gridline-will-now-draw/2316).  The problem was that a wrong index (the annotation index) was used to determine if the grid lines were selected when they were custom.


PaulWessel:
Windows CI test is having trouble installing gs so it fails, but unrelated to the PR itself.

anbj (approved):
Looks good; fixed!

PR #6056, opened 2021-11-24
Flip frame default for south pole aspect

### LINKED ISSUE #6055: modern theme MAP_FRAME_AXES setting needs improvement for polar plots

Polar azimuthal projections should have WesN annotations by default.

```
gmt basemap -R0/360/-90/-60 -JS0/-90/50/5i -Ba -B+t'South Pole' -png antarctica
```

**Actual outcome**
antarctica_wrong

**Expected outcome**

antarctica_correct

**System information**

* Operating system: Ubuntu 20.04
* GMT version (`gmt --version`): 6.3

PaulWessel:
See #6055 for background.  The issue was that the default of _WeSn_ only makes sense in the northern hemisphere or when S is not the S pole.  This PR checks if we are doing a South polar azimuthal axpect with the S pole as the latitude parameter, in which csae we flip S to N in the axes defaults.  This also fixes the missing annotations in example 42 which then needed to be 1 cm taller.

Other tests were not affected. Closes #6055.

### DIFF
--- a/doc/examples/ex42/ex42.bat
+++ b/doc/examples/ex42/ex42.bat
@@ -28,7 +28,7 @@ gmt begin ex42
 	echo 6.5 14		>> lines.txt
 	echo 13	11.5	>> lines.txt
 	echo 19	11.5	>> lines.txt
-	gmt plot lines.txt -R0/19/0/25 -Jx1c -B0 -W2p -X-6c -Y-13.5c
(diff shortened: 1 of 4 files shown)


meghanrjones (approved):

PaulWessel:
Any idea why the Windows check fails?

PR #6058, opened 2021-11-25
Improve movie check and make it less intrusive

PaulWessel:
See [forum](https://forum.generic-mapping-tools.org/t/supporting-scripts-and-use-with-movie/2378/3) for background.  This PR tries to limit a specific check for sub-shells to lines that do not have double-quotes (which could hide plain text that looks like a sub-shell incantation inside it).  It also demotes the message from warning to information.  Also fixes a typo.  No change to tests.

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -18344,8 +18344,8 @@ int gmt_token_check (struct GMT_CTRL *GMT, FILE *fp, char *prefix, unsigned int 
 					GMT_Report (GMT->parent, GMT_MSG_WARNING, "Main script appears to have a deprecated sub-shell call `...`, please use $(...) instead: %s", start);
 				else if (strchr (line, ')') && (p = strchr (line, '('))) {	/* sub-shell call without leading $ */
 					prev = p - 1;	/* Get previous character */
-					if (prev < start || prev[0] != '$')
-						GMT_Report (GMT->parent, GMT_MSG_WARNING, "Main script appears to have a sub-shell call $(...) without the leading $: %s", start);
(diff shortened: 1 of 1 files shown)


anbj:
When I run the script:

```
cat <<eof > globe.sh
gmt begin
   gmt coast -Rg -JG\${MOVIE_FRAME}/20/10c -Gmaroon -Sturquoise -Bg -X0 -Y0
   bash $(pwd)/output.sh 0 60 | gmt plot -Sc.2c -Gred -W -l"well hello (there)"
gmt end
eof

cat <<eof > output.sh
echo \$1 \$2
eof

gmt movie globe.sh -Nglobe -T30 -Fmp4 -Chd -Lf -Pb+ap -Z

rm globe.sh output.sh
```
I don't get information about suspicious `()` even when I include ` -Vi`.

I think it's a good idea to give a heads up about this.

anbj:
But maybe this falls in the *(2) see if I can catch more cases like this from even being flagged in the first place* category..?

PaulWessel:
yes, given the double quotes I do not complain about it anymore.

anbj:
Then it looks fine to me.

But again, it's a healthy check; I would rather get one unnecessary warning instead of not getting a necessary warning.

PaulWessel:
> Then it looks fine to me.
> 
> But again, it's a healthy check; I would rather get one unnecessary warning instead of not getting a necessary warning.

Agree, but this fix will only remove unnecessary warnings so I think it is an improvement.

meghanrjones (approved):

PR #6060, opened 2021-11-27
gmtspatial -Q ignored user selection of unit

PaulWessel:
When we upgraded the area calculation of spherical polygons using the functionality provided by the _gmt_sph.c_ library (earlier this year?), we inadvertently hardwired geographic areas to always report km^2, even though we state that the user may append any desired unit [with k being the default].  Well, that unit was parsed, used to initialize internal scales, but was ultimately not applied to the area.  The usage message also missed the optional `<unit>` selection (but the RST documentation was correct, yet did not make it clear that the unit was optional and what the default was).  This PR fixes the documentation and now reports areas in the desired length unit squared, which is now also listed in the segment header.  I also replaced some numerical constants with the named constants instead to make the code clearer.

No tests are affected. segment length calculations were not affected and uses the desired unit.

### DIFF
--- a/doc/rst/source/gmtspatial.rst
+++ b/doc/rst/source/gmtspatial.rst
@@ -163,9 +163,9 @@ Optional Arguments
     **-Q+h** to append the area to each polygons segment header [Default
     simply writes the area to stdout]. For polygons we also compute the
     centroid location while for line data we compute the mid-point
-    (half-length) position. Append a distance unit to select the unit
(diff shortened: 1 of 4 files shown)


Esteban82 (approved):
I test it with:
```
gmt spatial @GSHHS_h_Australia.txt -fg -Qm
134.208785723	-25.6616364739	2.73069728844e+16

```

PR #6062, opened 2021-11-27
Let gmt usage message spell out standard input and output

PaulWessel:
We shall only use stdin and stdout in specific technical descriptions but otherwise use standard input and standard output, to match the user documentation (#6061).


seisman (approved):

PaulWessel:
The Windows check keeps failing lately, @seisman.  Anything we should be concerned about?

seisman:
> The Windows check keeps failing lately, @seisman. Anything we should be concerned about?

I believe we need to rebuild the cache for vcpkg packages. I will do it later. This PR should be good to merge.

PR #6069, opened 2021-11-29
Increase string length for command

PaulWessel:
Foolishly, the **grdinterpolate** module used a very short string for a _GMT_Call_Module_ call but the file path was so long we exceeded the memory set aside.  This ups that to deal with such cases.

### DIFF
--- a/src/grdinterpolate.c
+++ b/src/grdinterpolate.c
@@ -580,7 +580,7 @@ EXTERN_MSC int GMT_grdinterpolate (void *V_API, int mode, void *args) {
 		unsigned int io_mode = GMT_WRITE_NORMAL;
 		uint64_t seg, row, rec, col;
 		uint64_t dim[4] = {1, 1, 1, 2};	/* Dataset dimension for one point */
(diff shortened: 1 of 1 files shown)


meghanrjones (approved):

src/grdinterpolate.c
```diff
@@ -580,7 +580,7 @@ EXTERN_MSC int GMT_grdinterpolate (void *V_API, int mode, void *args) {
 		unsigned int io_mode = GMT_WRITE_NORMAL;
 		uint64_t seg, row, rec, col;
 		uint64_t dim[4] = {1, 1, 1, 2};	/* Dataset dimension for one point */
-		char i_file[GMT_VF_LEN] = {""}, o_file[GMT_VF_LEN] = {""}, grid[GMT_LEN128] = {""}, header[GMT_LEN256] = {""}, cmd[GMT_LEN128] = {""};
+		char i_file[GMT_VF_LEN] = {""}, o_file[GMT_VF_LEN] = {""}, grid[GMT_LEN128] = {""}, header[GMT_LEN256] = {""}, cmd[GMT_BUFSIZ] = {""};
```
joa-quim, line 583: ``GMT_BUFSIZE`` is 4 k chars. Don't need that much either.

PaulWessel: OK, true.  I can set it to PATH_MAX+64 or something like that.  One of the args can be a very long path but the others I have control over.  I will change this in another PR that fixes another bug shortly...

PR #6079, opened 2021-12-02
grdselect.rst: fix example polygon name

anbj:
Please check.

Text refers to `my_data_area.txt,`, but example shows `wesn_polygon.txt`.


meghanrjones (approved):
You are correct, thanks!

Issue #6080, opened 2021-12-02
Bug in grdcountour -T+a


Esteban82:
### BUG: 
There is a contour line (that goes trough the Andes) with the tickmark in the opposite direction. 

### Doubt: 
Should all the contour line have tickmarks? 

### Full script:

```
gmt begin grdcontour png
	gmt grdmath -Rg -I1 40 21 SDIST = tmp_dist.nc
	
	gmt coast -Rd -JX10/5c -G200 -Sdodgerblue2
	gmt grdcontour tmp_dist.nc -C1000 -Wthinnest -T+a -A
	
	gmt coast -Rd -JW40/10c -G200 -Sdodgerblue2 	   -Y5c
	gmt grdcontour tmp_dist.nc -C1000 -Wthinnest -T+a
gmt end show
```
![grdcontour](https://user-images.githubusercontent.com/17908951/144441076-5af1c457-913a-4436-b637-2d426b82ee65.png)

PaulWessel:
> ### BUG:
> There is a contour line (that goes trough the Andes) with the tickmark in the opposite direction.
> 
> ### Doubt:
> Should all the contour line have tickmarks?
> 
> ### Full script:
> ```
> gmt begin grdcontour png
> 	gmt grdmath -Rg -I1 40 21 SDIST = tmp_dist.nc
> 	
> 	gmt coast -Rd -JX10/5c -G200 -Sdodgerblue2
> 	gmt grdcontour tmp_dist.nc -C1000 -Wthinnest -T+a -A
> 	
> 	gmt coast -Rd -JW40/10c -G200 -Sdodgerblue2 	   -Y5c
> 	gmt grdcontour tmp_dist.nc -C1000 -Wthinnest -T+a
> gmt end show
> ```
> 
> ![grdcontour](https://user-images.githubusercontent.com/17908951/144441076-5af1c457-913a-4436-b637-2d426b82ee65.png)

No, they are only applied to closed polygons.

PaulWessel:
This is the 13000 km line.  Interesting that it is determined to be a closed polygon.  I will need to debug to see why it is selected in the first place.

PR #6083, opened 2021-12-03
WIP Improve detection of closed but split contours

PaulWessel:

See #6080 for background.  After much forensics I learned that the failing contour reported a clockwise (CW) polygon whereas all others are counter-clockwise (CCW).  Due to how the contour-trace algorithm and later clipping and projection works, I believe they _are_ all counter-clockwise or should _be_ CCW.  Adding a sanity check to converts CW to CCW and prints a warning was added to both **grdcontour** and **pscontour**.  The correction fixes the problem reported in #6080 and all tests pass in master, except _grdcontour/closed.sh_ gave a slightly different plot (as the change affected what is considered beginning and end.  Instead of updating the PS I just upped the RMS limit to 0.02 so the test passes.

Here are all the changes in this PR:

1. Change the internal named constants for CCW and CW to be -1 and +1 which better match the uses of handedness in GMT.
2. Modify a place in _gmt_plot.c_ where an unset handedness was indicated by GMT_NOTSET.  Now we must use 0 instead.
3. Add the check, message, and switch if CW is found.

While I am pretty sure this fix is robust (the combined tests found something like ~80 CCW situations for closed contours that were to be ticked and just 1 (in _grdcontour/closed.sh_) plus the case in #6080 were CW, I am flagging this PR as WIP until we run some more contours with **-T+a** on more grids.

### DIFF
--- a/src/gmt_constants.h
+++ b/src/gmt_constants.h
@@ -374,8 +374,8 @@ enum GMT_enum_inside {
 
 /*! Return codes from gmt_polygon_orientation */
 enum GMT_enum_polorient {
-	GMT_POL_IS_CCW = 0,
-	GMT_POL_IS_CW  = 1};
+	GMT_POL_IS_CCW = -1,
+	GMT_POL_IS_CW  = +1};
 
 /*! Codes for -q selections */
 enum GMT_enum_skiprows {
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -9421,7 +9421,7 @@ void gmt_geo_polygons (struct GMT_CTRL *GMT, struct GMT_DATASEGMENT *S) {
 	struct GMT_DATASEGMENT *S2 = NULL;
 	struct GMT_DATASEGMENT_HIDDEN *SH = gmt_get_DS_hidden (S);
 	bool add_pole, separate;
-	int outline = 0, P_handedness = GMT_NOTSET, H_handedness;
+	int outline = 0, P_handedness = 0, H_handedness;
 	int geo = gmt_M_is_geographic (GMT, GMT_IN);
 	uint64_t used = 0;
 	char *type[2] = {"Perimeter", "Polar cap perimeter"};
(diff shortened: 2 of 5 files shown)


Esteban82 (approved):
Now it works.

PaulWessel:
Not out of the woods yet.  Feel free to try this script and run it repeatedly for random locations of the center of the grid:

```
gmt begin grdcontour png
	lon=$(gmt math -Q -180 180 RAND =)
	lat=$(gmt math -Q -80 80 RAND =)
	gmt grdmath -Rg -I1 $lon $lat SDIST = tmp_dist.nc
	gmt coast -Rd -JW$lon/10c -G200 -Sdodgerblue2 -B+t"($lon, $lat)"
	gmt grdcontour tmp_dist.nc -C1000 -Wthinnest -T+a
gmt end show
```

In general, it looks better but once in a while we have contours (very close to a pole?)  and it comes out the wrong way.

PaulWessel:
I added annotations so we can report bad cases and know which contour it was.  Some observations so far:

1. Some contours that enclose small areas and wrapped on the E/W sides gets a crazy number of ticks on the East part of the polygon.  I think it is always on the east (or _xmax_). My theory is that somehow the length from the west to east (not drawn) are included in the length estimate which is sued to determined how many ticks.
2. We find consistently many contours that are clockwise and I am forcing them to be counter-clockwise (this was the above fix).
3. The failing contours are not "closed" on the map (entry and exit at different sides) while the ones that form a closed loop on the map are correctly ticked.


Hi @Esteban82 and @meghanrjones please post if you find other systematic issues.

Esteban82:
I have just got this. Is it ok? I don't like the tick in the bottom left and in the upper right just in the borders. 


grdcontour_1

And here the 5000 km contour has ticks in the opposite side.

grdcontour

PaulWessel:
That one is OK but if you run the script with random lon/lat you will quickly get cases that still fail and give the wrong direction.

joa-quim:
Ping

PaulWessel:
Will try to look tomorrow...

PR #6092, opened 2021-12-04
Improved handling of output grid history in grdedit and grdconvert

PaulWessel:

These two modules are special in that they only make minor (or no) changes to the grid, and hence one may wish to control what the output grid's history should be.  This was previously not very clear: **grdedit -C** would wipe the history, while **grdconvert** would write the old history if both input and output grids were netCDF and no **-R** was changed.  This PR standardizes the behavior of **grdedit** and **grdconvert** via a common **-C** option (revised for **grdedit** but backwards compatible and new to **grdconvert**):

**-C**[**b**|**c**|**n**|**p**]

**b** will write both command histories separated by a semi-colon
**c** will only write the current module history
**n** will write no history whatsoever [Default if -C is used with no argument]
**p** will write the previous module history

This means the kind of grid no longer matters. Of course, combined histories may be longer than the standard 320 characters and only the netCDF format will handle longer histories, remarks, and titles.

No test are affected that I am aware of.

### DIFF
--- a/doc/rst/source/grdconvert.rst
+++ b/doc/rst/source/grdconvert.rst
@@ -13,7 +13,7 @@ Synopsis
 .. include:: common_SYN_OPTs.rst_
 
 **gmt grdconvert** *ingrid* |-G|\ *outgrid*
-[ |-C|\ **b**\|\ **n**\|\ **o** ]
+[ |-C|\ **b**\|\ **c**\|\ **n**\|\ **p** ]
 [ |-N| ]
 [ |SYN_OPT-R| ]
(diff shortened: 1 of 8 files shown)


PaulWessel:
OK with this standardization, @joa-quim ?

joa-quim (approved):

anbj:
Great, but this does not affect a workflow of say, `xyz2grd` -> `grdfilter` -> `grdconvert`, right? Only the last `grdconvert` will be present in the command history. I my head, this should apply to all grid generating/manipulating commands.

PR #6096, opened 2021-12-06
movie.rst: create table for preset video formats

anbj:
Create table of preset video formats. Easier to read. Consider removing the descriptions in the paragraph.

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -66,7 +66,7 @@ Required Arguments
 
 **-C**\ *canvassize*
     Specify the canvas size used when composing the movie frames. You can choose from a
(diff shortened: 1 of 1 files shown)


Esteban82:
Great idea @anbj. 

I have my own table where I add columns with pixel density (px/cm and px/inch). I wonder if it is worth including it.

anbj:
Please do if you think it’s useful. 

My git expertise is not good enough; can you do commits to this branch in any way? Or maybe you can add it later.

Esteban82:
I think I can commit but I am not really sure if it is useful. We can always add it later.

meghanrjones (approved):
Great, thanks!

I agree with removing the descriptions from the paragraph above, do you mind if I commit that change to this branch before merging?

anbj:
Not at all - please do. Thanks!

PR #6098, opened 2021-12-07
Better handling of trailing text in sample1d output

PaulWessel:
See forum [post](https://forum.generic-mapping-tools.org/t/best-way-to-make-time-file-for-movie-fill-inn-date-gaps/2421/5) for discussion.  If input records have trailing text then if there are output records at the same times then these should inherit any input trailing text:

**in.txt:**

```
1967-01-01 6 A B
1967-01-08 5
1967-01-12 7 C D
1967-01-15 1 E
1967-01-20 2 F G
```

```
gmt sample1d -fT --TIME_UNIT=d -I1 -Fa in.txt
1967-01-01T00:00:00	6	A B
1967-01-02T00:00:00	5.60489000795	
1967-01-03T00:00:00	5.33262655712	
1967-01-04T00:00:00	5.15995229261	
1967-01-05T00:00:00	5.06360985953	
1967-01-06T00:00:00	5.02034190299	
1967-01-07T00:00:00	5.00689106812	
1967-01-08T00:00:00	5	
1967-01-09T00:00:00	5.31835669826	
1967-01-10T00:00:00	6.0269814984	
1967-01-11T00:00:00	6.72211554934	
1967-01-12T00:00:00	7	C D
1967-01-13T00:00:00	5.59980754838	
1967-01-14T00:00:00	2.9098328522	
1967-01-15T00:00:00	1	E
1967-01-16T00:00:00	0.364936170213	
1967-01-17T00:00:00	0.130553191489	
1967-01-18T00:00:00	0.31370212766	
1967-01-19T00:00:00	0.931234042553	
1967-01-20T00:00:00	2	F G
```

### DIFF
--- a/doc/rst/source/sample1d.rst
+++ b/doc/rst/source/sample1d.rst
@@ -58,7 +58,7 @@ Required Arguments
     This is one or more ASCII [of binary, see
     **-bi**] files with one column containing the
     independent *time* variable (which must be monotonically in/de-creasing)
-    and the remaining columns holding other data values. If no file is
+    and any number of optional columns holding other data values. If no file is
     provided, **sample1d** reads from standard input.
 
 Optional Arguments
--- a/doc/rst/source/sample1d.rst
+++ b/doc/rst/source/sample1d.rst
@@ -14,6 +14,7 @@ Synopsis
 
 **gmt sample1d** [ *table* ]
 [ |-A|\ [**f**\|\ **p**\|\ **m**\|\ **r**\|\ **R**][**+d**][**+l**] ]
+[ |-E| ]
 [ |-F|\ **l**\|\ **a**\|\ **c**\|\ **n**\|\ **s**\ *p*\ [**+d1**\|\ **2**] ]
 [ |-N|\ *col* ]
 [ |-T|\ [*min/max*\ /]\ *inc*\ [**+a**][**+i**\|\ **n**][**+u**] ]
@@ -80,6 +81,14 @@ Optional Arguments
     **+l** if distances should be measured along rhumb lines (loxodromes).
     **Note**: Calculation mode for loxodromes is spherical, hence **-je**
     cannot be used in combination with **+l**.
+
+.. _-E:
+
+**-E**
+    If the input dataset contains records with trailing text then we will attempt
+    to add these to output records that exactly match the input times.  Output records
+    that have no matching input record times will have no trailing text appended [Default
+    ignores trailing text].
 
 .. _-F:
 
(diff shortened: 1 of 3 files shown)


meghanrjones:
My concern is that this PR could break workflows that were written with an assumption that sample1d does not output trailing text. If this is considered an enhancement, I think it would be better to set a new modifier to trigger the behavior.

PaulWessel:
I did not think any of the options were suitable for a modifier for this so I added an option **-E** to attempt to deal with trailing text.  Default thus remains the same as before.

anbj:
Looks good. Extra information is now preserved.

Two other things:

* I may have one _or more_ rows of the same date. This cause `sample1d` to stop:

```
$ cat <<eof | gmt sample1d -fT --TIME_UNIT=d -I1 -Fa
2021-01-01 1
2021-01-02 2
2021-01-02 3
eof
sample1d [ERROR]: x-values are not monotonically increasing/decreasing (at zero-based record 2)!
sample1d [ERROR]: Failure in gmt_intpol near row 3!
```

Why is not duplicate rows (on the 0th collumn) not allowed? Could it be?

* In my case, I don't want to interpolates anything else but the 0th column (dates in my case).  But `sample1d` requires a value that it can interpolate (2nd column). Any way to make it only look at the 0th column?

PaulWessel:
I will look toes if we can get away with no data set to resample.  Seems like that should be easy.  Right now we allow n columns to be resampled and I think 0 should be allowed.
As for duplicates. Remember we are building a spline f(t) and it cannot handle two values at the same time.  Just like with surface or greenspline, one would have to average those points before the interpolation.  There is -A+d to skip duplicates but that may be limited to spatial curves (cannot remember).

PaulWessel:
Now, sample1d is OK with interpolating just the time column if there are no other data columns, so @anbj can test this out.  I also updated the documentation.  I agree this is a useful feature (resample a set of time nodes without having to give a data column when none is needed.

anbj:
Super!

```
cat <<eof | gmt sample1d -fT --TIME_UNIT=d -I1 -Fa -E
2021-01-01 hello
2021-01-10 there
eof
2021-01-01T00:00:00	hello
2021-01-02T00:00:00
2021-01-03T00:00:00
2021-01-04T00:00:00
2021-01-05T00:00:00
2021-01-06T00:00:00
2021-01-07T00:00:00
2021-01-08T00:00:00
2021-01-09T00:00:00
2021-01-10T00:00:00	there
```

Thank you Paul!

anbj (approved):

PR #6102, opened 2021-12-09
Correctly compute contour length for wrapped contours

PaulWessel:
Contours split across periodic boundaries will have jumps across the map.  These are not drawn, but the code used them in computing the length which affects how many ticks are set.  This corrects the contour length calculations.  Here is an example of before (see the lower right side):

before

This is after the fix:

after

No tests were affected since we rarely place lots of ticks on contours.


PaulWessel:
Note this fix is not related to the problems in #6083 hence I made a separate PR.  This one is benign in that it fixes a long-standing issue not revealed by our tests.  Once this is merged and #6083 is fixed we will add a test that exercise these two issues.

Esteban82:
@PaulWessel  Whtt script do you use for the figure?

Esteban82 (approved):
I try it with the script in #6080 (comparing master from contour-lenghts-fix).

PR #6103, opened 2021-12-10
Add a tutorial version in Julia

joa-quim:
As it says, this PR adds a version of the Tutorial translated to Julia.

One reason to submit it here is that converting the RST to Markdown (used in the Julia docs) would be a lot of work.

### DIFF
--- a/doc/rst/source/index.rst
+++ b/doc/rst/source/index.rst
@@ -45,6 +45,7 @@ Quick links
         gallery
         animations
         tutorial
+        tutorial_jl
 
     ---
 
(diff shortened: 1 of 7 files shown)


PaulWessel (approved):
Looks OK, but I think you should remove all our names from the Julia first page and just cite via link to the CLI tutorial since none of us have contributed to the Julia docs or code, no?

PaulWessel:
Maybe you could also add a link to https://julialang.org/learning/getting-started/ in the first page for those who need to explore this a bit first.

joa-quim:
Added some links and reduced the authors list. Still left you since you since you actually wrote the original tutorial.

PaulWessel:
OK, I guess that is fine for now.  I think I was thinking of just you writing GMT.jl and thus only you need to be listed there.

Issue #6108, opened 2021-12-11
Better presentation of wind field


leeyupeng119:
**Description of the desired feature**
I use **grdvector** function to plot the wind field, I adjusted -Q and -Si parameters repeatedly, but I couldn't get satisfactory results.  

Here is my final code (turn parameters again and again) and output figure,
```
gmt begin wind2_gmt.png
gmt coast  -JM5i -R75/140/15/55 -Baf -A4000  -W0.25p,black
gmt grdvector u.nc v.nc -Ix5 -Q4p+e+h0.1+n1k -Gblack -W0.5p -Si30k
gmt end
```
![GMT plot](https://user-images.githubusercontent.com/55687784/145677155-f8e8ebc8-bd3d-4852-b793-faf222479381.png)

And I did the same thing with NCL (NCAR Command Language), which is a very commonly used software in atmospheric science research. And here's what NCL output, 
![NCL plot](https://user-images.githubusercontent.com/55687784/145677111-cec694bc-7481-451d-89ff-b8f2240f4251.png)
By comparing the two plots, it can be found that GMT still has some limitations in drawing wind fields, the main limitations are as follows：

1.  Wind direction is not easily discernible in the weak wind speed zone, so you can't see the circulation pattern very well.
2.  There is no feature for automatically generating the legend of the reference vector.

**Are you willing to help implement and maintain this feature?** Yes

In my opinion, when the wind speed is very small, the stem may disappear, but there should be a certain lower limit for the reduction of the arrow, and the overall shape of the arrow should not be destroyed, so that the wind direction of each grid point can be clearly identified, which is very important in circulation analysis. In the research of meteorology,  we pay attention not only to the value of the wind speed but also to the wind direction. Therefore, would you consider adding the features I mentioned above? Thanks in advance.

Maybe I am not good at using **grdvector**, so I have uploaded the data, welcome you to test and turning.
[data_uv.zip](https://github.com/GenericMappingTools/gmt/files/7697118/data_uv.zip)

PaulWessel:
Thanks for this example and the comparison. Having **+n** be subjected to a minimum shrinking is a good idea that can be implemented.  I will play with your data example to see how we can improve the plot. And adding -l for auto-legend is on my mental agenda anyway.

PaulWessel:
[...]

PR #6115, opened 2021-12-12
Add optional fraction for terminal shrinking value

PaulWessel:
The **+n**_norm_ modifier that shrinks down vector attributes for shorter vectors would shrink to 0 by default.  This PR adds an optional terminal shrink value [0] that can be set by appending _/frac_.  The shrink_lim.sh test explores how this works. The +n is used to shrink attributes when the vector length is less than 1 inch (**-R** values are in inches and the gridlines indicates the cutoff):

shrink_lim

Comments:

1. For very short vectors and relatively large heads, we seem to plot a head that implies a starting point below the vector start (y = 0).  See top-left plot for length = 0.1" = 7.2p is less that head length of 12 p
2. @joa-quim suggested that perhaps default _frac_ should be nonzero and we would claim this is a bug fix.  The argument is that when _frac_ goes to zero we can no longer tell the azimuth of the vector as it approaches a point.  Perhaps 0.25?

Thoughts?

### DIFF
--- a/doc/rst/source/explain_vectors.rst_
+++ b/doc/rst/source/explain_vectors.rst_
@@ -51,7 +51,7 @@ end point (**e**) of a line segment:
 
     **+n**\ *norm*\ [/*min*] scales down vector attributes (pen thickness, head size)
     with decreasing length, where vector plot lengths shorter than *norm* will have
-    their attributes scaled by length/\ *norm* [other arrow attributes remains
+    their attributes scaled by length/\ *norm* [other arrow attributes remain
     invariant to length]. For Cartesian vectors specify a length in plot units, while
     for geovectors specify a length in km. Optionally, append /*min* for the minimum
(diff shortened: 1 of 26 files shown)


meghanrjones:
I agree that the current behavior could be considered a bug; 0.25 minimum produces a more reasonable plot.

joa-quim:
Yes, 0.25 seems a good dfalt.

PaulWessel:
OK, I will update the docs that way.  I think I will fix any tests that use **+**n to have an explicit /0 so that we don't get useless failures. Might as well since this example will exercise the new defaults, and I am working on a wind-vector plot based on #6108 that will do a good job but requires review of #6113.

PaulWessel:
OK, this PR has been updated, scripts revised to avoid new PS originals, plus the new test case.  Ready for a review.

PR #6134, opened 2021-12-20
Fix bad KEYs in psxy|z introduced in 88fbed1588df5e0ec693760e04b3d5b3bac57225

joa-quim:


PaulWessel (approved):
OK, I will add comments in GMT_Encode_Options later after reviewing all the KEYS with magic =+etc

PR #6143, opened 2021-12-22
Fix subset of global virtual grids

### LINKED ISSUE #6140: Plotting grid sub-regions in linear projections is bugged from Julia
This came out in the [forum post](https://forum.generic-mapping-tools.org/t/a-weird-issue-related-to-subplot/2460) and it doesn't look like a Julia bug. See, it screws with linear projections (including Plate Carre) but works well for *true* projections (Mercator in this example). 
Note that plotting directly from the file (i.e. without loading it first) also works well.
```
julia> G = gmtread("@earth_relief_05m");

julia> grdimage(G, region=(-180,-40,15,80), color=:geo, figsize=10, Vd=1, show=1)
        grdimage  -R-180/-40/15/80 -JX10/0 -Baf -BWSen -n+a -Cgeo -P -K > C:\TEMP\GMTjl_tmp.ps

julia> grdimage(G, region=(-180,-40,15,80), color=:geo, figsize=10, proj=:merc, Vd=1, show=1)
        grdimage  -R-180/-40/15/80 -JM10 -Baf -BWSen -n+a -Cgeo -P -K > C:\TEMP\GMTjl_tmp.ps
```

GMTjl_tmp-fs8

GMTjl_tmp-fs8

PaulWessel:
When virtual grids are passed and we need a subset, we were stymied by the first test for the rotation of a global grid.  In reality, we may need to do both rotation and cropping but the first test was chained with an else-if and hence the subset was skipped.  Closes #6140.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -3032,7 +3032,7 @@ GMT_LOCAL void * gmtapi_pass_object (struct GMTAPI_CTRL *API, struct GMTAPI_DATA
 						gmt_grd_shift (API->GMT, G, shift_amount);	/* In-memory rotation */
 					}
 				}
-				else if (object->region) {	/* Possibly adjust the pad so inner region matches requested wesn */
+				if (object->region) {	/* Possibly adjust the pad so inner region matches requested wesn */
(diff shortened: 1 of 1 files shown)


joa-quim (approved):
Yes, that fixed it. Thanks.

PR #6161, opened 2021-12-28
Address a series of issues related to vectors

### LINKED ISSUE #6158: Vectors with heads exceeding vector length
We have a **+n** modifier to deal with short vectors and can shrink the head and vector attributes.  However, if **+n** is not used and the (fixed) head size exceeds the vector length, what should happen? Here is an example that compares what happens for geographic and Cartesian vectors (plain and regular):

full_heads

There is consistency between Cartesian and plain geovector, but the filled-head geovector just vanishes when we hit the critical length = head-length.   I think we want consistent behavior, and here are two possibilities:

1. Do exactly as the Cartesian cases do (and then fix the geovector case).
2. Refuse to draw any head if head > length (and fix the 3 other cases).

The second solution would at least draw a short line that

1. Indicates the true vector length (no longer hidden by large head).
2. Indicates the azimuth (which _is_ clearer with the large head).

I suspect there is no obvious answer here so perhaps this warrants a modifier, e.g., **+n** with no args means stop drawing head if it is longer than the vector length [Default does as in the top 3 plots]. I guess I do not like the current default since the magnitude of the vector is basically lost behind the vector head, but rather than breaking backwards compatibility I would be OK with the argument-less **+n** solution.

Comments, @GenericMappingTools/core ?

PaulWessel:
There have been several bugs and omissions identified in GMT vectors lately (e.g., #6108) and we have had some partial fixes.  This PR reconciles several problems and completes incomplete enhancements:

1. The new minimum vector shrink in **+n**_norm/limit_ was documented to default to 0.25, but in fact it was never set so it was still zero.  This is now fixed.
2. The check for geovectors to skip heads if they exceed the vector length was flawed so some small vectors ended up with bad heads (e.g., #6108).  Now the test is fixed and heads exceeding size are not plotted.
3. Cartesian vectors never bothered to check and hence always plotted the heads even though exceeding the vector lengths.  This has now been addressed.
4. Finally, there are uses of vectors where the vector length should _not_ be considered when plotting heads (e.g., adding a head to an existing line). Many of these are internal (e.g., adding a vector head at the end of the line via **plot -W** or adding one at the end of a math x-axis for **--MAP_FRAME_TYPE** = _graph_), but placing flow vectors on a graph may also fall prey to the limitation.  Using **+n** without any argument now means we always plot the requested heads regardless of vector length.

These corrections meant that 8-10 PostScript plots needed to be updated. For some I chose to use **+n** to avoid having to change the  plot, while others would benefit from a refresh (e.g., properly giving scaling in km for geographic vectors in _sample.sh_).

Closes #6158 and #6108; the latter is also the inspiration for new test _met_uv.sh_


joa-quim (approved):

meghanrjones:
The new met_uv test is failing on linux and windows due to rounding error when the vector length is close to the vector head size (see diff below). @PaulWessel, would you be able to modify the test to avoid this?

met_uv

PaulWessel:
Interesting.  The decision to draw head or not is just

`if (dr[0] > h_length_limit)`

so there is nothing I can do code-wise to beat any roundoff.  As for the test, I guess we would have to hunt for a slightly different scale or initial head size that gives the same plot?

Another take: I am about to improve the documentation on the geovector and Cartesian vector as the docs are not clear enough.  Basically, the geovector means you specify a length in map units (km, for example) and then we plot that length on the map.  This means the actual plot length on the paper depends on the map projection.  A 50 km-long vector at Equator is half that of one at 60N.  This is not what people who have vectors in other units (plate velocity, wind speed, etc) want, since it is biasing what the user sees.  For those applications they need to plot Cartesian vectors, which do have the same length regardless of location and projection.  To make matters worse, we have some tests that mixes these cases up, such as met_uv.sh which you can see is using geovector due to the -Si20k setting instead of, say, a -Si5c argument (random number, not sure what it needs to be to be visible).  But for some perspective projections (-JG for instance), a geovector is suitable for those applications since the length is affected by the perspective.  Documentation with figures on the grdvector page is needed for this.

So I think I will fix met_uv.sh to use Cartesian vectors first, and then we can see if that causes different images for different OS.  OK?

joa-quim:
I don't understand the roundoff story. There are rules for rounding and all compilers should produce the same result. Unless some are using *fastmath* compiler options.

PaulWessel:
Hm, yes.  Perhaps the cloud are doing things lie -O3 or -O? for some other setting that affects accuracy at the benefit of speed? Do we know the compilation flags for GMT as they are used in the cloud, @meghanrjones ?

meghanrjones:
I got the failure locally on linux using the same compilation flags as on macOS. If it's worthwhile, I could run through the debugger to check exactly what is different between the two.

joa-quim:
If the examples can be reduced to a very simple case where the results are different in MacOS vs Others you can print the dr[0]  & h_length_limit of the ``if (dr[0] > h_length_limit)`` and at least have an idea of the different values involved.

meghanrjones:
dr[1] and h_length_limit are the same on macOS and linux, but for this particular vector it is only drawn on macOS:
```
grdvector [INFORMATION]: dr[1] = 0.23587; h_length_limit = 0.23587 
```

PaulWessel:
Probably needs %.16lg to see the difference in values.

meghanrjones:
Sorry, here is the comparison (I'll edit the post to add the linux case):
| macOS | linux |
|--------|------|
| dr[1] = 0.2358698342821208; h_length_limit = 0.2358698342821207 | |

joa-quim:
So, it's the last bit that differs. This case can be solved by setting ``>=`` instead of ``=`` but there will be others.

PaulWessel:
Would be OK to use >= instead of > in general here, since that means the length equals or is larger than the limit.

PR #6162, opened 2021-12-28
More exhaustive tests of vectors

PaulWessel:
Exercises begin, end, and both vector head skipping and shrinking for geo and Cartesian vectors.  All new tests, so they all pass.

### DIFF
--- a/test/psxy/full_heads_b.ps
+++ b/test/psxy/full_heads_b.ps
@@ -1,0 +1,2897 @@
+%!PS-Adobe-3.0
+%%BoundingBox: 0 0 612 792
+%%HiResBoundingBox: 0 0 612.0000 792.0000
+%%Title: GMT v6.4.0_c923ecc_2021.12.27 Document from plot
+%%Creator: GMT6
+%%For: unknown
+%%DocumentNeededResources: font Helvetica
+%%CreationDate: Mon Dec 27 17:08:30 2021
+%%LanguageLevel: 2
+%%DocumentData: Clean7Bit
+%%Orientation: Portrait
+%%Pages: 1
+%%EndComments
+%%BeginProlog
+250 dict begin
+/! {bind def} bind def
+/# {load def}!
+/A /setgray #
+/B /setdash #
+/C /setrgbcolor #
+/D /rlineto #
+/E {dup stringwidth pop}!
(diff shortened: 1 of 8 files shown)


meghanrjones:
All four new tests yield SEGV. There also seems to be some issues for the latest vector updates on other OS's (https://github.com/GenericMappingTools/gmt/runs/4646750010?check_suite_focus=true). Let me know if you need me to run the tests on Linux to inspect.

PaulWessel:
Hm, I see no SEGV, please try to get more info.  These tests are not that different from other tests so I am surprised about this.

meghanrjones:
I'll work more on it, but as a start the SEGV occurs inside `gmt_cart_to_geo` called from `gmtplot_circle_pen_poly` (https://github.com/GenericMappingTools/gmt/blob/c923eccf1b72cdfdd842a660d6f9fb1591eb3b0b/src/gmt_plot.c#L4683) where data is NaN.

meghanrjones:
The problem occurs with this input line in t.txt:
```
0	0	0	0
```
For this case, the code below returns `nan` since `hypot(x1 - x0), (y1 - y0)` = 0.

https://github.com/GenericMappingTools/gmt/blob/c923eccf1b72cdfdd842a660d6f9fb1591eb3b0b/src/gmt_plot.c#L4642-L4652

Which sets `scl[0]` to nan in the line below https://github.com/GenericMappingTools/gmt/blob/c923eccf1b72cdfdd842a660d6f9fb1591eb3b0b/src/gmt_plot.c#L5245, which leads to an unrestrained loop later on.

PaulWessel:
Ok, I guess we are missing the test that says if the length is zero get out of here.

Could you please try to add a line at the top of the _gmt_geo_vector_ function that just does

`if (gmt_M_is_zero (length)) return (GMT_NOERROR); `

meghanrjones:
That fixes the new tests. Now I get a failure for the test `test/psxy/smallcircvec.sh`. I do not see where the vector length is set in that test script, so I think it is a test script issue rather than a code issue.

PaulWessel:
Hm, OK, please make a PR with the above fix and then we can see what is happening with small circles.  Please post what the failure is.

PR #6164, opened 2021-12-29
Add check for zero length vectors

meghanrjones:

This PR adds a check for zero length vectors to prevent failures with the new tests in https://github.com/GenericMappingTools/gmt/pull/6162. 

With this change, [smallcircvec.sh](https://github.com/GenericMappingTools/gmt/blob/vector-length-check/test/psxy/smallcirvec.sh) fails with the following diff (vectors in magenta are plotted in master but not this branch):

smallcirvec

I think the issue is that an azimuth is not given in the test script, so the length is used as the azimuth:
`echo $lon $lat $length | gmt psxy -R -J -W1p,$color -S=0.4i+j${just}+o${plon}/${plat}+h1 -G$color -O -K >> $ps`

Fixes #

**Reminders**

### DIFF
--- a/test/psxy/smallcirvec.sh
+++ b/test/psxy/smallcirvec.sh
@@ -13,7 +13,7 @@ cat << EOF > t.txt
 EOF
 gmt psbasemap -Rg -JG0/0/4.5i -Bag -P -K -Xc > $ps
 while read lon lat length color plon plat just; do
-	echo $lon $lat $length 0 | gmt psxy -R -J -W1p,$color -S=0.4i+j${just}+o${plon}/${plat}+h1 -G$color -O -K >> $ps
+	echo $lon $lat $length | gmt psxy -R -J -W1p,$color -S=0.4i+j${just}+o${plon}/${plat}+h1 -G$color -O -K >> $ps
 	echo $lon $lat   | gmt psxy -R -J -O -K -Sc0.1i -G$color -W0.25p >> $ps
 	echo $plon $plat | gmt psxy -R -J -O -K -St0.1i -G$color -W0.25p >> $ps
 done < t.txt
--- a/test/psxy/smallcirvec.sh
+++ b/test/psxy/smallcirvec.sh
@@ -13,7 +13,7 @@ cat << EOF > t.txt
 EOF
 gmt psbasemap -Rg -JG0/0/4.5i -Bag -P -K -Xc > $ps
 while read lon lat length color plon plat just; do
-	echo $lon $lat $length | gmt psxy -R -J -W1p,$color -S=0.4i+j${just}+o${plon}/${plat}+h1 -G$color -O -K >> $ps
+	echo $lon $lat $length 0 | gmt psxy -R -J -W1p,$color -S=0.4i+j${just}+o${plon}/${plat}+h1 -G$color -O -K >> $ps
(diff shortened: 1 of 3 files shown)


PaulWessel:
Hm, that is possible, but I note we are drawing small circles around a specified pole.  I cannot recall what the input is in that case since you cannot change the azimuth of that line from the starting point, just adjust the length.  Will have to look at this - have  not plotted small circles in a while...

PaulWessel:
OK, for small circles the two args are angle_1 and angle_2.  So you would think we need 

`echo $lon $lat 0 $length | gmt psxy -R -J -W1p,$color -S=0.4i+j${just}+o${plon}/${plat}+h1 -G$color -O -K >> $ps`

In addition the 

GMT_LOCAL unsigned int gmtplot_geo_vector_smallcircle (struct GMT_CTRL *GMT, double lon0, double lat0, double azimuth, double length, struct GMT_PEN *ppen, struct GMT_SYMBOL *S) {

function should probably replace azimuth, length with angle_1 and angle_2 to avoid confusion.  I will do that.

PaulWessel:
The test I suggested earlier only applies for the great circle path.  For the small circle there is another test.  Will update soon.

test/psxy/smallcirvec.sh
```diff
@@ -13,7 +13,7 @@ cat << EOF > t.txt
 EOF
 gmt psbasemap -Rg -JG0/0/4.5i -Bag -P -K -Xc > $ps
 while read lon lat length color plon plat just; do
-	echo $lon $lat $length | gmt psxy -R -J -W1p,$color -S=0.4i+j${just}+o${plon}/${plat}+h1 -G$color -O -K >> $ps
+	echo $lon $lat $length 0 | gmt psxy -R -J -W1p,$color -S=0.4i+j${just}+o${plon}/${plat}+h1 -G$color -O -K >> $ps
```
meghanrjones, line 16: The order of length and 0 is switched from your comment on the PR. This relates to your comment about improving the docs (https://github.com/GenericMappingTools/gmt/pull/6161#issuecomment-1002303114); from what I can tell documentation about how to set small circle characteristics is missing.

PaulWessel: Yes, I switched because that is the only thing that worked to make the test pass.

PaulWessel:
Looking at this case. Logic would dictate that the small circle vector needs two measures: start and stop, and these are either in map length (e.g., km) or angles (if +q is set).  However, when we pass 0 $length in the script I can tell the 4th column is not read or set to 0, so we get 0 0 and hence nothing.  Looking at why now.

PaulWessel:
The implementation of the small vectors allows for two cases:

1. A given length of the arc in map units, hence input is lon, lat, length
2. The start and stop angle in degree of the arc, hence input is lon, lat, angle1, angle2

There is no case for length1 and length2, hence the script was correct passing just $length.

PaulWessel:
All tests pass for me (well, on M1 gspline_5 fails but that is a separate story).  OK to approve and merge?

meghanrjones:
> All tests pass for me (well, on M1 gspline_5 fails but that is a separate story). OK to approve and merge?

Yes. Even though all the changes are yours, I cannot approve a PR that I opened.

PaulWessel (approved):

Issue #6163, opened 2021-12-28
Better auto-legend horizontal spacing


PaulWessel:
**Description of the desired feature**

For historical reasons, the input to **legend** for a symbol has this syntax:

`S [dx1 symbol size fill pen [ dx2 text ]]`

where _dx1_ is the distance from the left margin to the center of the symbol and _dx2_ is the distance from the left margin to the start of the optional text.  If _dx1_ is given as "-" then we auto-compute _dx1 = 0.5 * size_, while if _dx2_ is given as "-" then we compute _dx2 = 1.5 * size_. For typical geometric symbols (e.g., circle) this works great: We get a gap of half the symbol size between the symbol and the text.  For lines or vectors, it works not so well.  Consider placing a contour or vector that is 1 cm via `-l"My vector"`.  Then, the label will start 0.5 cm after the vector and the gap is starting to seem too big.  Go to 2 cm and now you have a big gap.  This is essentially one of the challenges for adding **-l** to **grdvector** and **velo** (there are others). Note: If you placed the label with a tight frame (no offset, i.e. **+o**0) then the symbol would be tangent to the left frame.  In reality we have a default offset of 0.2 cm so the left gap in all the auto-legend examples is basically 0.2 cm.  This also does not scale well for large posters.

I think there are two elements of scale here:

1. The symbol size
2. The annotation text

Seems to me it would be better if _dx2_ involves a mix of symbol size and font size.  E.g., instead of the above definition for an auto _dx2_, perhaps _dx2 = size + 2*annot_font_size_ would be better.  Consider a large plot on a poster: You are not going to plot a 4 cm vector and use a 10p font label.  You'll be doing 24 to 32 points and you want that gap to relate to the font size. Since characters have different widths (except in Courier and similar fonts), I will consider the width of the letter M as one unit below, but it could simply be the actual font size (e.g., 12p)

My proposal is therefore to make some changes when _dx1_ and/or _dx2_ is given as "-":

```
dx1 = one unit + 0.5 * size
dx2 = one unit + size + two unit = 3 units + size
```

This means if _dx1 = dx2 =_ "-" (which is what we do when **-l** is used), we get a reasonable left-space after the margin and a getter alignment of the label text to the right of the symbol.

I will make a branch and experiment a bit with this (perhaps "one" and "two" are too little/too much and need to be adjusted).  Let me know if @GenericMappingTools/core have any suggestions.

PaulWessel:
[...]

PR #6165, opened 2021-12-29
New default legend symbol dimensions when no details are provided

PaulWessel:

While **legend** will plot exactly what you tell it to, when we now start to rely on the auto legend feature in **-l** there were several shortcomings.  Once was the lack of a sense of the annotation font in placing the text (#6163) but also the lack of default parameters for most symbol when none is provided.  And the few defaults that were there needed some adjustments as well.  Before we accept these changes as is we should make sure they are adequate.  Here is a plot of all the symbols that have auto-settings (historically, we had one for front but there really is no good way to know what kind of front (here squares) so it should probably be removed as auto, just like we cannot guess what quoted and decorated lines might look like:

autosize

What I mean by "no details" is simulated here by passing record like this (for circle):

`S - A - red 0.5p - Star`

[In **-l,** pen, colors, symbol type and label comes from the calling module, while _size, dz1,_ and _dx2_ may be unset]

Here are issues to tackle or make decisions on:

1. For ellipse, the rectangles, the wedge and vector sthere are named parameters at the top of _pslegend.c_ that can be tinkered with (and recompiled).  Are these the finalized settings, you think?
2. Given the new equation for _dx2 = size + W_ (symbol size plus letter width), most tests using **-l** will have slight failures
3. Two old examples (03 and 07) had very small symbols relative to the label so they have been updated (and then fail)
4. Any legend-related test script showing many types of symbols will fail due to the adjustments of parameters.

I think the proposed PR is vastly superior to what is in master, but some tinkering can happen, and I think for consistency we should remove the auto-square setting from front so it is like the other line-embellishment symbols that need more than "-" to do anything useful.

Once this PR can be merged I can start tackling the job of adding **-l** to **grdvector** and maybe **velo**.

**Note**: No _PostScript_ updates are presented here so you can run the tests and compare the results to see the changes.

### DIFF
--- a/doc/examples/ex07/ex07.bat
+++ b/doc/examples/ex07/ex07.bat
@@ -6,7 +6,7 @@ REM
 gmt begin ex07
 	gmt coast -R-50/0/-10/20 -JM24c -Slightblue -GP26+r300+ftan+bdarkbrown -Dl -Wthinnest -B --FORMAT_GEO_MAP=dddF
 	gmt plot @fz_07.txt -Wthinner,-
-	gmt plot @quakes_07.txt -h1 -Scc -i0,1,2+s0.025 -Gred -Wthinnest -l"ISC Earthquakes"+S0.2c
+	gmt plot @quakes_07.txt -h1 -Scc -i0,1,2+s0.025 -Gred -Wthinnest -l"ISC Earthquakes"+S0.5c
 	gmt plot @isochron_07.txt -Wthin,blue
 	gmt plot @ridge_07.txt -Wthicker,orange
 	gmt legend -DjTR+o0.5c -F+pthick+ithinner+gwhite --FONT_ANNOT_PRIMARY=18p,Times-Italic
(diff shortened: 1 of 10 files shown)


meghanrjones:
Generally looks good, just a couple issues: the symbols do not appear in the `region_u_i.sh` or `autolegendcolor.sh` tests using this branch. I don't think that the rotated ellipses look better in `leg3d.sh`, `legend3d.sh` and `legend.sh`.

PaulWessel:
Darn it, I must had buggered something, the region_u_i.sh and autolegendcolor.sh was fine yesterday but now I see the problems.  I will fix whatever that is but forced to take a break and go swimming.  Grrrr.

PaulWessel:
Fixed.

meghanrjones:
Thanks. 

I think the ellipses could be improved. Consider the example below, where the ellipse conditions set with -S differ from the legend result:
```
gmt begin test png
    gmt plot @DSDP.txt -R100/160/-20/30 -JM5c -Se280/0.2/0.1 -Ggreen -Bafg -l"ellipse"
    gmt legend -DJCB+w2c+o0c/0.5c
gmt end show
```
test

PaulWessel:
Hm, it should have passed those args in - will have to debug.

PaulWessel:
Thanks, the latest revisions returns the ellipse to zero angle and does a better job parsing incoming parameters for the multi-arg symbols.  Your ellipse case now works better, for instance.  I think this is pretty good?

meghanrjones (approved):
Looks good. My preference would be to leave the failures as is for now in case there are adjustments made in the next legend PR(s).

PaulWessel:
OK, we will leave the failures for now.

PR #6178, opened 2022-01-02
Save any default CPT for remote datasets in hidden grid header

### LINKED ISSUE #6167: Retain information about default colormaps for remote files
It would be helpful if GMT had a way to retain information about the default colormaps for remote files during processing. For example, if the following code produced two images using the `geo` colormap. This feature would be helpful for PyGMT, where the remote data files are often read into memory before being passed to GMT, resulting in the loss of the default cpt information (xref https://github.com/GenericMappingTools/pygmt/issues/499).

```bash
gmt grdimage @earth_relief -R0/10/0/10 -B -JM5c -png ex1

gmt begin ex2 png
    gmt grdcut @earth_relief -R0/10/0/10 -JM5c -Gtest.nc
    gmt grdimage test.nc -B
gmt end show
```

ex1:
ex1

ex2:
ex2

PaulWessel:
This addresses #6167 when we cut a subset from, say, `@earth_relief_10m` and then no longer know what the CPT should be.  it is now added as a netCDF z-variable attribute and can thus be found by GMT as well as other environments.  A new test (defect.sh) have been added that shows it is now working.
Two older tests (polycut.sh and highpass.sh) were changed to explicitly use turbo so that they would not fail due to this improvement.  Closes #6167.


src/gmt_nc.c
```diff
@@ -1029,6 +1032,8 @@ GMT_LOCAL int gmtnc_grd_info (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *head
 
 		/* Define z variable. Attempt to remove "scale_factor" or "add_offset" when no longer needed */
 		gmtnc_put_units (ncid, z_id, header->z_units);
+		if (GMT->parent->remote_info && GMT->parent->remote_id != GMT_NOTSET && GMT->parent->remote_info[GMT->parent->remote_id].CPT[0] != '-')	/* Subset of remote grid with default CPT, save name as an attribute */
+			gmt_M_err_trap (nc_put_att_text (ncid, z_id, "cpt", strlen (GMT->parent->remote_info[GMT->parent->remote_id].CPT), GMT->parent->remote_info[GMT->parent->remote_id].CPT));
```
joa-quim, line 1036: "cpt" is a very GMT parlance. I would prefer "palette" like those grids from OceanColor

test/grdimage/defcpt.sh
```diff
@@ -0,0 +1,11 @@
...
+# Ensure that grdimage picks up the CPT stored in the netCDF z:cpt attribute
+# https://github.com/GenericMappingTools/gmt/issues/6167
+# Fixed via 
+gmt begin defcpt
+	gmt grdcut @earth_relief_01d_g -R0/10/0/10 -Gtest.nc
+	gmt subplot begin 2x1 -R0/10/0/10 -JQ10c -Fs10c
+		gmt grdimage @earth_relief_01d_g -R0/10/0/10 -B -c
+		gmt grdimage test.nc -B -c
+	gmt subplot end
+gmt end show
```
joa-quim, line 11: Here we are again creating a new PS. Preferably the test should detect the presence of the cpt but if that's too complicated, at least make the test a tinny figure.

src/gmt_nc.c
PaulWessel: I don't mind calling it palette, but since it is just a string with a GMT CPT name it is still very GMT centric.  We are not sticking a 3xn array into the grid, just a string.  Still want to rename it?

test/grdimage/defcpt.sh
PaulWessel: I just updated to remove the boundary.  It is a 1010 grid - that is pretty small, no? The size of the PS is 22K, og which 20K is the PSL header.

src/gmt_nc.c
joa-quim: Ok, sorry. I thought that it was the cpt that was added to the nc file and that would be an interesting addition. If it's just a *cpt name suggestion for GMT scripts* then I'm not even sure we should be doing this.

PaulWessel: Nothing to do with GMT scripts though.  This is for the module and hence for wrappers as well if interested.  Since I think you pass wrapped GMT grids in Julia then this will work for you off the bat.  PyGMT will need to do some other work since they read in a netcdf grid via Xarrays, so they have access to the attributes that way.

We could of course stick a CPT into the netCDF but it has little value in GMT (we would do it because the outside world wants it) and it is limiting in that 3xn matrices do not tell us continuous or discrete and now we have to add more stuff to explain that thing.

test/grdimage/defcpt.sh
PaulWessel: I do not see how I can make a non-plot test to detect that GMT correctly identifies the attribute and stretches the CPT, etc.

src/gmt_nc.c
joa-quim: I am not exploring the hidden struct because there is no guaranties that it won't change and then each change would mean a Julia breakage. I've pulled out some hairs in the past trying to inventing a system where the structs could have some self-referencing system that could be used to find the struct members of interest regardless of structs changes. It would involve the use of ``offsetof`` C macro. But never came to get any workable prototype.

PaulWessel: OK.  Here is the issue for users in GMT.jl:

```
G = grdcut("@earth_relief_01d_g", region=[0 10 0 10]);
grdimage(G, proj=:Winkel, colorbar=true, show=true)
```

You are getting the turbo colors instead of geo.  Same in GMT master.  But with this PR it has that information available in the subset grid and can be acted upon.  The modules can do it since they either read the grid into the full structure or they pass the full structure as a virtual file to another module.

Another thing we could do, if it made sense for Julia and Python and Matlab, would be to set the current CPT when we **grdcut** from a remote grid.  Then, the above Julia lines would also work as is, and if you really wanted turbo then you would specify it.

PaulWessel: I.e., we set the default CPT even though we are not plotting, because **grdcut** is often used to select a smaller chunk before plotting.  No harm done if then used for **grdmath**...

test/grdimage/defcpt.sh
PaulWessel: I like my idea though: It complements the attribute idea since that solution will survive a session whereas the current CPT has a limited scope in where it is found (figure, subplot, inset), etc.

src/gmt_nc.c
joa-quim: GMT.jl tries expand the classic mode to work like *modern* and only misses that goal for the modules that require *modern* (*i.e.* ``subplot``). So, for example the, implementing the current cpt concept is achieved via global variable. It's not completely idiotic to imagine that each grid could have a *suggested cpt* (or *none*) but this requires changing the ``GMTgrid`` type (structure). However, I return to the same point. I can't know about it from GMT because the info is stored in the hidden struct. Don't know how PyGMT will find this information using a external tool. Ofc, we can scan fir it from all of the nc attributes, but this seems overkilling.

PaulWessel: Sorry I see I posted a few comments to the wrong conversation (below).  But if you are not creating a modern session then there is no place to write the current CPT to.  And I am not sure this will help for PyGMT either @meghanrjones since each GMT module call is wrapped in a Create/Destroy session.  So how is the concept of current CPT maintained in PyGMT?

joa-quim: BTW, will this work in **classic**?
```
grdcut("@earth_relief_01d_g", region=[0 10 0 10], G="cut.nc");
grdimage("cut.nc", proj=:Winkel, colorbar=true, show=true)
```

PaulWessel: For CLI and C interface it works in classic since the module can use the full grid header.
For externals needing the current CPT it only works for modern since there is no session in classic and hence no place to write the CPT...

PaulWessel: Since you are doing stuff under the hood. presumably you could do something when **grdcut** is called? In your C-wrap do you have access to the full header under the hood?  Even if there is a void *hidden pointer? We could imagine adding an API function that could report something based on that void *.

joa-quim: If it will work with CLI then it should probably work in Julia (the disk file version) but possibly not because GMT.jl classic sticks in a default CPT in that grdimage command when not CPT is used.

PaulWessel: Yes, if you do not pass any default CPT but let GMT control the default CPT then it would work.  Are you supplying a different default CPT than GMT would normally (turbo?).  if not then just don't supply one.

meghanrjones: I think this is a worthwhile improvement for the CLI, regardless of usability for the wrappers. While PyGMT can easily access the cpt attr, I'm not sure how it would communicate it back through the API during the `GMT_Create_Data`, `GMT_Put_Matrix`, `GMT_Open_VirtualFile` process.

[GMT_Set_Comment](https://docs.generic-mapping-tools.org/dev/devdocs/api.html#gmt-set-comment) is only for the 'top-level' grid header (those listed [here](https://docs.generic-mapping-tools.org/dev/devdocs/api.html#gmt-grids)), correct? Could anything be added to the API for setting the items that are mostly used internally (i.e., full list [here](https://docs.generic-mapping-tools.org/dev/devdocs/api.html#gmt-grid))?

PaulWessel: I don't think that is the step you will want to intervene.  If you learn there is a cpt attribute then you add "-C<cpt>" to the string you prepare for the module, no?

meghanrjones: Yes, we can do that. If that is the case, we may want to eventually document that users can add a `cpt` attr to their netcdf grid in order to associate a default colormap with their own grids. In which case Joaquim's comment about using ``palette`` over ``cpt`` becomes relevant again.

PaulWessel: I suppose **grdedit** is the module where it makes sense for us to add an attribute to the grid file. Something like

`grdedit my.grd -Qelevation`

This brings up the issue of "I want my specially made cpt file to be embedded in the grid", to which I suggest someone read the gmtmex_parser.c on the gymnastics on converting a very rich CPT standard into a very poor Matlab color table scheme.  It would be the same issue - lots of work to accommodate all the special things in a CPT, vs just adding the name of the cpt.

meghanrjones (approved):
The current implementation is fine to me.

PaulWessel:
A slight issue has arisen: We have test and doc scripts that sets a default CPT via makecpt, then cuts a subset via **grdcut** afterwards.  Now, that is changing the current CPT.  In some places I can switch the order of those to commands, but i others we set a current cpt before a **subplot** and then we have **grdcut** commands inside those panels and now the CPT takes on the one in the grid since the **grdcut** command forces the resetting of the current CPT.

Two options:

1. Undo this feature in **grdcut**.  After all, it was added in haste yesterday to help with externals by setting the current CPT but it remains unclear if this even affects GMT.jl and it seems PyGMT can easily detect a default CPT via array attributes.
2. Add a modifier to set any detected attribute cpt as the current CPT

Any thoughts on this, @joa-quim and @meghanrjones? Note that GMT modules etc has no need for the **grdcut** feature so I am OK with option 1, but willing to do option 2 if there is any desire to keep this feature for the externals.

meghanrjones:
Option 1 would still leave the cpt attr and use that as the default if there isn't already a colormap set in the session, correct? If so, I think that's fine.

PaulWessel:
Yes, the attribute in the written grid remains, it is just the whole section

```
if (GMT->current.setting.run_mode == GMT_MODERN && HH->cpt) {	/* Save the CPT as the current CPT for a modern mode session */
	struct GMT_PALETTE *P = gmt_get_palette (GMT, HH->cpt, GMT_CPT_OPTIONAL, G->header->z_min, G->header->z_max, 0.0);
	if (GMT_Destroy_Data (API, &P) != GMT_NOERROR)
		Return (API->error);
}
```
that will be rescinded due to unintended consequences...

meghanrjones:
Sounds fine.

joa-quim:
Yes ls agree to let it there to be used if expressly stated and not a setting that happens by default.

PaulWessel:
> Yes ls agree to let it there to be used if expressly stated and not a setting that happens by default.

Could you clarify this: Are you voting to add an option to **grdcut** or do you agree with Meghan?

joa-quim:
Agree with Meghan.

Issue #6104, opened 2021-12-10
Add -l to event


anbj:
I dont' understand why the `-l` option (https://docs.generic-mapping-tools.org/dev/gmt.html#l-full) is not available for `events`.
`events` is in many ways similar to `plot`, and it's useful to have this option to create an easy and quick legend.

PaulWessel:
It can be added.  The thinking was that you would not want to waste CPU time redrawing that legend again and again and that you instead would use -Sf to have a script that makes an overlay.

anbj:
That’s reasonable. Thanks Paul.

anbj:
-Sf to make an overlay; which command does this refer to?

anbj:
Maybe having the possibility to use -l in movie would be nice after all. It should be up to the user to decide what he/she wants to spend electricity on. And maybe add a note on the man page with a suggestion on how to donit the smrt way?

PaulWessel:
In **movie**, -**Sf**_file_ means you wish to either (a) run the script _file_ which will produce an PS overlay of the same dimension as the movie frames and it will be overlain on all frames as a constant foreground, or (b) _file_ is a PS overlay of same dimensions and will be overlain on all frames as is.  This is usually a good solution if you wish to place some constant legend, labels, and other things that will not change during the movie.

I think **movie -l** makes no sense since **movie** does not plot anything - it just runs scripts.  So that is not possible.  I assume you still meant **events -l**.  But even in that case you should note that **events** does not plot any data itself - it makes calls to **plot**, **text**, **meca**, **coupe** or **velo** depending on other option.  So not clear to me how to handle this (yet).

anbj:
Sorry, of course I meant `events -l`. Thanks for explaining Paul. The point anyway, from a pragmatic point of view, would be to have a similar `-l` capability when using `events`.

PR #6182, opened 2022-01-03
Allow -l for auto-legends in events

PaulWessel:
See #6104 for some background.  This PR allows parsing of **-l** in **events** which is then passed along to the actual module that does the work. Note that not all modules may support **-l** yet (e.g., **meca**) so use with caution.


anbj (approved):
Perfect.

PR #6184, opened 2022-01-03
movie.rst: add default -L

anbj:
I can't see that the movie man page mentions the default choice of `-L` (which is `f`). Added.

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -200,8 +200,7 @@ Optional Arguments
 .. _-L:
 
 **-L**\ *labelinfo*\ [*modifiers*]
-    Automatic labeling of individual frames.
-    Default is running frame number (f).
+    Automatic labeling of individual frames [Default is running frame number (f)].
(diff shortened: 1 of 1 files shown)


doc/rst/source/movie.rst
```diff
@@ -200,7 +200,9 @@ Optional Arguments
 .. _-L:
 
 **-L**\ *labelinfo*\ [*modifiers*]
-    Automatic labeling of individual frames.  Repeatable up to 32 labels.  Places the chosen label at the frame perimeter:
+    Automatic labeling of individual frames.
+    Default is running frame number (f).
```
meghanrjones, line 204:
```suggestion
    Automatic labeling of individual frames [Default is running frame number (f)].
```
The default is conventionally listed in brackets.

meghanrjones (commented):
Thanks! Just one comment below.

anbj:
Super. Thanks Meghan!

PaulWessel (approved):

Issue #6190, opened 2022-01-04
Proposal for more consistency for scaling vectors and shrinking attributes


PaulWessel:
**Description of the desired feature**

We still seem to have some inconsistencies when it comes to vector scaling (converting user vector magnitude to plot length) and attribute shrinking (reduce head size and stem pen width when vectors are smaller than a limit)

### Scaling issues:

- Cartesian vectors plotted by **plot** are expected with direction and length (in plot units, say cm) in columns 3 and 4. However, if you prefer to give components _dx, dy_ instead then these are scaled by **+z**_scale_ and combined with _hypot_.  So while I can then give _dx_ and _dy_ in some random user unit (say mm/yr) and use _scale_ to convert that to reasonable component lengths in inches, I cannot give the polar form in anything but final plot units.  This is inconsistent (same is true in **plot3**).  In contrast, for **grdvector** the vectors are of course expected to be in user units in both polar and component forms, and they are scaled via the information in **-S** to get them into plot lengths.  This makes sense, and the same is true in **velo**.
- Geovectors are special great-circle vectors whose lengths are given in map distances (e.g., km or spherical degrees).  In **plot** and **plot3** the input is expected to be azimuth and length (in any map distance unit).  The **+z** modifier does not apply to geovectors.  In **grdvector**, one can create geovectors from user data by selecting a suitable scale in **-S.**  This converts user unit magnitudes to map distances, then plots the geovector.  This cannot be done in **plot** or **plot3**, but in **velo**, so a bit inconsistent.

### Shrinking issues:

- Cartesian vectors can shrink if their lengths are below a threshold.  That threshold is currently in plot units (e.g., 0.5i, 1c, 24p) and if the plot length is shorter we compute a shrink scale as `s = length/limit`, but also prevent _s_ from being smaller than a user-selectable threshold [0.25].  If my input data are in mm/yr then I am out of luck to specify a limit in mm/yr and I would have to work out via the plot scale what such a limit would be in cm or points.
- In contrast, geovectors have their attributes shrunk in a similar way but the length here is the geographic map length (e.g., in km).  Since the plot length of these vectors vary with projection and view point it makes no sense to use a plot-length limit, hence the limit is in map lengths. Yet, if my data in **grdvector** are in mm/yr and I wanted a shrink limit in those units I am out of luck again.

Proposals:

[...]

PR #6194, opened 2022-01-05
Address items 3, 4, 7 in vector proposal

PaulWessel:
This PR addresses these three items in #6190:

(3) Allow **+z** in **plot** and **plot3** to accept components for a geovector and have them scaled to map distances. Properly compute what azimuth and length this corresponds to and plot the geovector accordingly.
(4) Disable (or give error) **+z** in grdvector since here we control polar vs component via **-A**.
(7) Accept a default scale of 1 if +z has no argument (currently this would give junk).

No tests are affected.

Some examples of geovector components calls:

```
echo 0 60 100 100 | gmt plot -R-2/3/59/62 -JM10c -B -S=12p+e+z -W1p -Gred -pdf map
echo 0 60 100k 100M | gmt plot -R-2/3/59/62 -JM10c -B -S=12p+e+z -W1p -Gred -pdf map
echo 0 60 1 2 | gmt plot -R-2/3/59/62 -JM10c -B -S=12p+e+z100k -W1p -Gred -pdf map

```

### DIFF
--- a/doc/rst/source/explain_vectors.rst_
+++ b/doc/rst/source/explain_vectors.rst_
@@ -86,7 +86,7 @@ In addition, all but circular vectors may take these modifiers:
     **+s** means the input *angle*, *length* are instead the :math:`x_e, y_e`
     coordinates of the vector end point.
 
-Finally, Cartesian vectors may take these modifiers:
+Finally, Cartesian vectors and geovectors may take these modifiers (except in :doc:`grdvector`):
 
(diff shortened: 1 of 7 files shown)


Esteban82:
Are these ok? 

map1
map2
map3

Esteban82 (approved):
I approve because my tests looks fine.

PaulWessel:
I think so: 100 km is a bit shorter than 2 degrees longitude at 60 N

PR #6197, opened 2022-01-07
Further vector scale improvements and enhancements

PaulWessel:
See #6190 for background.  This PR addresses the sub-tasks 1 , 2 and 6 (last task 5 will be handled after this PR is merged):

(**Task 1)**: Offer a way, like **+z**_scale_, to accept polar form direction, length input, with length in data units to be scaled to plot units. I chose **+v**_scale_ for "scale vector data length to plot length". Append a unit from c|i|p to indicate the unit of the product.
(**Task 2**): Allow the shrinking limit to be given in data units or plot units. Follow the lead of **plot** **-Sb** and use unit **q** to indicate data units (since no unit would mean cm and hence a plot unit for Cartesian vectors and km as map unit for geovectors).
(**Task 6**): Ensure that any use of vector modifiers that cannot apply in a particular module result in errors.

Along the way, a few other things had to happen:

1. I selected **q** (for user quantity) to use as a unit signifying the user's data unit.  We used **u** in plot **-Sb**, **-So**, and **-Su** but these have now been changed to **q** as well (backwards compatible of course).  This prevents the conflict with unit **u** (survey foot).
2. Both **+v** and **+z** can take unit **q** as well to indicate that input data are given in user quantities and the scale results in either plot units (Cartesian vector) or map units (geovector).
3. New vector attribute modifier **+c** was introduced to let us use the data magnitude instead of a separate input column with **-C** to look up vector head colors.
4. Fixed a bug for plotting vectors with fill and pen color taken from a CPT. The bug prevented the head outline pen from being changed.  We had no such test example so no failure, but I found this when plotting the top two rows.
5. Documentation updates and clarifications.

Here is one of two new tests (for **psxy**, the **psxyz** is identical since I did not use any perspective view point):

vector2d_mods

This test (psxy/vector2_mods.sh) plots a bunch of 1 inch = 2.54 cm vectors (left) or 100 km geovectors (right) but it does so using a wide range of input, units and vector attributes; these are plotted as text.  Top two row use the new **+c** modifier with a CPT and **-W+c** modifications. Left column is Cartesian, right column are geovectors.  View them from the bottom to the top to see what is going on with the input.  We test combination of unit-less and unit input points, vector components in plot or map units, and magnitudes and components in quantity units (**q**) scaled to suitable plot and map lengths. Note result for left side y = 3: Input length is given as 1i (with unit) so specifying **PROJ_LENGHT_UNIT** as cm has no effect since the data unit takes precedence. We experiment with the new **+v** and the improved **+z** modifier.

### DIFF
--- a/test/psxy/vector2d_mods.sh
+++ b/test/psxy/vector2d_mods.sh
@@ -39,14 +39,14 @@ gmt begin vector2d_mods
 	echo 0 5 ${dx} ${dy} | gmt plot -Sv20p+e+z1q -W2p -Gred
 	echo "-0.1 5.9 Input: ${dx} ${dy} Vector: -Sv20p+e+z1q" | gmt text $textopt
 	# dx, dy in specified plot units with scale, and color based on magintude
-	echo 0 6 ${dx} ${dy} | gmt plot -Sv20p+e+z${scl}q -W2p -C --PROJ_LENGTH_UNIT=inch
-	echo "-0.1 6.9 Input: ${dx} ${dy} Vector: -Sv20p+e+z${scl}q --PROJ_LENGTH_UNIT=inch" | gmt text $textopt
+	echo 0 6 ${dx} ${dy} | gmt plot -Sv20p+e+z${scl}q+c -W2p -C --PROJ_LENGTH_UNIT=inch
+	echo "-0.1 6.9 Input: ${dx} ${dy} Vector: -Sv20p+e+z${scl}q+c --PROJ_LENGTH_UNIT=inch" | gmt text $textopt
 	# dx, dy in specified data units with scale, and color based on magintude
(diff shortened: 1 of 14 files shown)


test/psxy/vector2d_mods.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
meghanrjones, line 1:
```suggestion
#!/usr/bin/env bash
```
To make the code validator happy

test/psxyz/vector3d_mods.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
meghanrjones, line 1:
```suggestion
#!/usr/bin/env bash
```

meghanrjones (approved):

PR #6192, opened 2022-01-04
file-formats.rst: minor formatting

anbj:
Do the man pages/cookbook etc. have a convention about formatting of utility names (e.g. `ogr2ogr`)?
There are some inconsistent ways of formatting them.

This would be a nice addition to the reStructuredText Cheatsheet (https://docs.generic-mapping-tools.org/dev/devdocs/rst-cheatsheet.html).

PR #6198, opened 2022-01-07
fix formatting for consistency

anbj:
This is work in progess.
Please review.

See #6192 for background.

### DIFF
--- a/doc/rst/source/cookbook/include-figures.rst
+++ b/doc/rst/source/cookbook/include-figures.rst
@@ -63,16 +63,17 @@ accomplished.
    Corel Draw will allow you to edit the BoundingBox graphically.
 
 -  A command-line alternative is to use freely-available program
-   **epstool** from the makers of Aladdin ghostscript. Running
+   *epstool* from the makers of `Aladdin ghostscript <https://www.ghostscript.com/>`_.
+   Running
 
       ::
 
        epstool -c -b myplot.ps
 
-   should give a tight BoundingBox; **epstool** assumes the plot is
+   should give a tight BoundingBox; *epstool* assumes the plot is
    page size and not a huge poster.
 
--  Another option is to use **ps2epsi** which also comes with the
+-  Another option is to use *ps2epsi* which also comes with the
    ghostscript package. Running
 
       ::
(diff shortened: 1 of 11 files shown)


doc/rst/source/kml2gmt.rst
```diff
@@ -34,8 +34,8 @@ This is a bare-bones operation that aims to extract coordinates and
 possibly the name and description tags of each feature. The main use
 intended is to capture coordinates modified in Google Earth and then
 reinsert the modified data into the original GMT data file. For a more
-complete reformatting, consider using **ogr2ogr -f** "GMT" somefile.gmt
-somefile.kml.
+complete reformatting, consider using
+*ogr2ogr -f "GMT" somefile.gmt somefile.kml*.
```
meghanrjones, line 38:
```suggestion
``ogr2ogr -f "GMT" somefile.gmt somefile.kml``.
```
I think since this is a full line of code rather than a reference to the tool it would be helpful to format it as code.

doc/rst/source/cookbook/include-figures.rst
```diff
@@ -63,16 +63,17 @@ accomplished.
    Corel Draw will allow you to edit the BoundingBox graphically.
 
 -  A command-line alternative is to use freely-available program
-   **epstool** from the makers of Aladdin ghostscript. Running
+   *epstool* from the makers of `Aladdin ghostscript <https://www.ghostscript.com/>`_.
```
meghanrjones, line 66: @PaulWessel, can you provide some guidance as the appropriate way to refer to the owner of these tools? From the wikipedia page, it seems Aladdin was forked and then merged again before being bought by Artifex. I cannot find any current web presence for Aladdin Enterprise.

I would actually prefer that we only document the GMT way to do these tasks, so that we don't need to maintain the lists of other tools. An example of the downside for listing alternative programs is that Aldus Freehand is listed as an option to edit the BoundingBox, which had it’s final release in 2003.

meghanrjones (commented):
Great work, thanks @anbj!

I made one in-line suggestion and a discussion comment about the 'Including GMT Graphics into your Documents' cookbook section, which won't require any additional changes in this PR.

doc/rst/source/cookbook/include-figures.rst
PaulWessel: That section is ancient and reminisces of a simpler time...not.  Yes, I agree we should prune things like eps tool etc.  Our way with **psconvert** is superior given our settings etc.

meghanrjones: Sounds good, I can submit an update to that section as a separate PR afterwards.

anbj:
Thanks Meghan.
I agree on your suggestion.

I feel this PR got kind of messy, with changes all over the place, but hopefully it's a start.

meghanrjones:
Looks good!

Would you rather continue working on this PR or approve/merge this and make more changes separately?

anbj:
The best I think is if this can be squashed and merged and I will continue separately. I’m not experienced enough with git (and PR’s on github) and I’m afraid I’ll mess something up at some point. So, I prefer small increments!

meghanrjones (approved):

PR #6199, opened 2022-01-08
Add -l to grdvector for legend entry

PaulWessel:
This PR adds **-l** to the **grdvector** repertoire.  The magnitude of the legend vector (in data units) can be set via **-S** modifier **+s**_refval_.  Note that **-l** requires Cartesian vectors as geovectors do not have a constant representation on a map w.r.t. a reference length but can vary with position and projection. A simple new test (_veclegend.sh_) demonstrates the effect:

```
gmt grdmath -R0/30/30/60 -I5 -fg X = x.grd
gmt grdmath -R0/30/30/60 -I5 -fg Y 30 SUB = y.grd
# Set the LL grid values to the same as the UR values
echo 0 30 30 > t.txt
gmt grdedit -Nt.txt x.grd
gmt grdedit -Nt.txt y.grd
# Use reference length as 25 mm/year in the legend
gmt begin veclegend
	gmt grdvector x.grd y.grd -R-2/38/29/65 -JM15c -B -Q14p+e+n25q/0 -Gred -W1p -S15c+s25 -l"Velocity (25 mm/yr)"
	echo 13 47 | gmt plot -Sc0.1i -Gblue -l"Arbitrary point"
gmt end show
```

veclegend

A few minor issues had to be fixed, such as an update to the plate.ps original, and a special check to honor **grdvector** settings for shrinking which are assumed to be in data units unless an actual unit is added (outside **grdvector** we expect to see the unit **q** for this).

Pinging @gd-a as well for information.


gd-a:
> vectors as geovectors do not have a constant representation on a map

Hi @PaulWessel,

Indeed, but we could still use a reference like
"10 m/s at 25ºN"

25ºN being defined manually through `-l+m25` or `-l+m+25` (m for manual)

PaulWessel:
Since the latitude 25N has no bearing on the plotted length of the arrow, you would just use -l"10 m/s".  The only arrows that vary with latitude would be geovectors and right now grdvector does not allow -l with those due to their variable length.  I mean, you could att "@ 25N" to your label but there is no need to pass +m25 since it has no effect.

gd-a:
> and right now grdvector does not allow -l with those due to their variable length

That's my point, you could allow it as long as "@25N" is required (or middle latitude by default)

PaulWessel:
Well, it would need to be like in the map scale syntax **-L**.  So perhaps we require **+c** for geovectors:

Use **+c**[[_slon_/]_slat_]] to control where on a geographic map the reference geovector length set via **+s** applies. Legend vector length is calculated for latitude _slat_ (optionally supply longitude _slon_ for oblique projections [default is central meridian]). If **+c** is given with no arguments then we select the length to be true at the middle of the map.

gd-a:
👍  I like that (if not too much work)

I would greatly clean some of my codes by getting rid of `gmt legend` multi-lines :)

PaulWessel:
Now, **grdvector** supports legends for geovector also, by requiring the **+c** setting.  Also note that while the length of the vector is computed as discussed, the vector that is actually plotted in the legend is the Cartesian vector since there is no simple way to plot a slightly curved great circle stem there anyway.

meghanrjones:
I think t.txt, t.pdf, x.grd, and y.grd in test/grdvector may have been committed by mistake. We could also remove the two postscript files because I can add them using dvc when updating the other failing tests.

PaulWessel:
OK, got rid of those plus gmt.history and a PDF file. yikes...

meghanrjones (approved):

gd-a:
> Now, **grdvector** supports legends for geovector also, by requiring the **+c** setting.  Also note that while the length of the vector is computed as discussed, the vector that is actually plotted in the legend is the Cartesian vector since there is no simple way to plot a slightly curved great circle stem there anyway.

I'll update some of my scripts next month to try this
Thanks Pål !

PR #6212, opened 2022-01-11
Let map roses have a default size of 10% of map width

### LINKED ISSUE #6211: Set width for map roses (-Td) with % values and make it optional
For colorbar it is possible to define its [length](https://docs.generic-mapping-tools.org/dev/colorbar.html#d) as % value of the map side. 

1. I think it would be nice to have the same option for map roses (magnetic and directional).  Something like `-TdJTR+w10%`
2. Also maybe we could define a default width value. I think 10% is fine. This would convert +w and optional parameter. So the users could have a directional rose with just `-TdjTR`.

PaulWessel:
This PR makes **+w**_width_ optional for magnetic and directional roses.  if not given we default to 10% of the map width. Alternatively, give **+w** with % as unit to change this value.  Closes #6211.

@Esteban82, please test this a bit, also for magnetic rose (see GMT_App_K_3.sh or GMT_mag_rose.sh for examples).  if you think 10% is too small for the magnetic rose let me know - we can always have a different default there.


PaulWessel:
Note this bug/feature which is separate from this PR:

`gmt coast -RGB+r2 -JM20c -Gred -B -pdf map -TdjTR+l`

We do not adjust the offset when the N label is requested.  I will fix this in another PR I think.

PaulWessel:
Same/worse if I add **+f** for fancy.

Esteban82:
Here I test the sizes of roses. 
Directional roses goes from 5% to 10% (with 1% inc).
Magnetic roses: 10%, 12.5%, 15%, 17.5% and 20% (2.5 % inc). 

Now I think we could use 8% for directional and 15% for magnetic. 

test_autoN

Esteban82 (approved):
It works. We should define the default values.

PaulWessel:
OK, done.  I hope you based your suggestion on the symbols and not the text since another PR will fix the too-large font sizes.

Esteban82:
I do it based on the spacing of the tickmarks.

BTW what do you think in similar feature for the scale bar ? I.e to have a default length (and unit).

PaulWessel:
I think we will wait merging this branch until the auto-font size/offset is done.  Then we can revisit those percentages.

Sure, you can open a feature request for that.

PaulWessel:
Sorry, could you share your script above so I can use it to test font sizes etc?

Esteban82:
> Sorry, could you share your script above so I can use it to test font sizes etc?
```
gmt begin test_autoN png
	gmt coast -Df -N1/thinnest -N2/thinnest,- -W1/thinnest -R-72/-51/-62/-53.5 -JM15c -Bxaf -Byaf 

#	Directional Rose
	gmt basemap -TdjTR+w10%+f
	gmt basemap -TdjTR+w9%+o2c/0+f
	gmt basemap -TdjTR+w8%+o4c/0+f
	gmt basemap -TdjTR+w7%+o6c/0+f
	gmt basemap -TdjTR+w6%+o8c/0+f
	gmt basemap -TdjTR+w5%+o10c/0+f
	
#	Margnetic Rose	
	gmt basemap -TmjBR+w20%
	gmt basemap -TmjBR+w17.5%+o3c/0
	gmt basemap -TmjBR+w15%+o6c/0
	gmt basemap -TmjBR+w12.5%+o9c/0
	gmt basemap -TmjBR+w10%+o12c/0
gmt end
```

PaulWessel:
FYI @Esteban82 I decided to merge this branch into the one doing the justification since I kind of want to see how both work together.

Issue #6213, opened 2022-01-11
Justification of map roses ignore space for label


PaulWessel:
We can place a directional rose anywhere on the map via **-T**, and most easily via justification codes for the 12 points on the map.  However, if  labels are requested then two issues are apparent:

1. The default font used need to scale with the symbol and not be a fixed separate setting. Turns out this is **FONT_TITLE**                     which defaults to 24p,Helvetica,black.  This is way too big for all but poster-size maps I think, so I am sure any users of **-T** also sets a smaller font that is suitable.  In fact, our own example in the cookbook (_GMT_obl_merc.sh_) uses  **--FONT_TITLE**=8p.
2. The space needed for the labels needs to be considered in the justification so they are not falling outside the map

Here is the what is happening now:

![roses](https://user-images.githubusercontent.com/26473567/149019002-1fb9b798-51d7-4a9e-a3cd-67758a37eb59.png)

At least in modern mode that font should scale with the rose symbol size. For both classic and modern the offset should take into account the labels.

seisman:
Maybe related to this issue:

```
gmt basemap -R-10/10/-10/10 -JM15c -TdjMC+w5c -Bafg -png map
```
The direction rose is not plotted at the middler center (MC) of the map:
![map](https://user-images.githubusercontent.com/3974108/156929712-95d8f362-2716-45c8-9e9b-b0f3598d5b15.png)

In the following command, I mistyped `+w5c` as `+5c`:
```
gmt basemap -R-10/10/-10/10 -JM15c -TdjMC+5c -Bafg -png map
```
and I got the below figure. Perhaps GMT can do better at checking the syntax?
![map](https://user-images.githubusercontent.com/3974108/156929795-62e4bff0-e524-405d-89af-ee4b9dcfe15f.png)

seisman:
Similarly, please also verify the placements (especially the offset between the rose and the frames) are correct:

![rose](https://user-images.githubusercontent.com/3974108/165125094-c36d94b7-f4f6-4358-93fd-9940f6bcfc3f.png)
![rose](https://user-images.githubusercontent.com/3974108/165125143-b03a5596-528f-41e4-8647-80bcb02d5ea2.png)

PaulWessel:
[...]

PR #6215, opened 2022-01-11
Proper positional adjustment of directional rose due to labels

PaulWessel:
See #6213 for background.  This PR does a few things and raises a question (at the end):

1. We adjust the center of the rose coordinate by the label offset amount that have been selected.
2. We use PSL to determine the exact width and height of the label (if specified) to transform the origin by the amount.

This addresses the main concern in #6213.  The PR does a few more things for modern mode:

1. It sets the label font size to 25% of the rose diameter.
2. It sets the label offset (gap between rose thorns and text) to be 50% of the label size.

Both of those fractions are up for discussion and refinement.  Also, the classic defaults for font size (**FONT_TITLE**) and offset (**MAP_TITLE_OFFSET** for fancy and (curiously) **MAP_ANNOT_OFFSET** for plain) are much too large and furthermore inconsistent.  We could either declare a bug here or just keep the auto-scaling to modern mode in line with other auto-scalings. 

Happy to take comments on this.  Below is a new test.  There is one tiny issue remaining: The S label is still clipped and I cannot yet see why this can happen...

roses


meghanrjones:
Keeping the auto-scaling to modern mode in line with other auto-scalings sounds good to me.

Esteban82 (approved):

PaulWessel:
Ping @Esteban82, @joa-quim  and @meghanrjones. I have implemented auto-scaling for the magnetic rose as well.   Please build from this branch and use this script to explore and give feedback on the scalings I chose. FYI, I set the scales so that for size 2.5i (the original script) the roses looks similar.  For smaller sizes things scale down accordingly:

```
#!/usr/bin/env bash
# Change W for smaller compass whose attributes scale with size
W=2.5i
gmt begin mag
	gmt set GMT_THEME cookbook
# Magnetic rose with a specified declination
	gmt basemap -R-10/-2/12.8812380332/0.661018975345+r -JOc0/0/50/60/7i -B0 -X1.25i
	gmt basemap -Tmg-2/0.5+w${W}+d-14.5+t45/10/5+i0.15p,blue+p0.15p,red+l+jCM \
		--FONT_ANNOT_PRIMARY=9p,Helvetica,blue --FONT_ANNOT_SECONDARY=12p,Helvetica,red --FONT_LABEL=14p,Times-Italic,darkgreen \
		--MAP_FRAME_WIDTH=10p --COLOR_BACKGROUND=green --MAP_DEFAULT_PEN=2p,darkgreen --COLOR_BACKGROUND=darkgreen \
		--MAP_VECTOR_SHAPE=0.5 --MAP_TICK_PEN_SECONDARY=thinner,red --MAP_TICK_PEN_PRIMARY=thinner,blue
EOF
gmt end show
```

See _gmt_plot.c_ line 3308 for where the scales are - feel free to experiment changing those numbers if you think they can be improved.

Note: I have not yet dealt with positions and adjustments due to text size as I have done for directional rose - that comes later.

meghanrjones:
Would auto-scaling for the intervals typically set using +t be possible? Or an option to apply MAP_ANNOT_MIN_SPACING to those?

PaulWessel:
I've added a variable to tell me if +t was set.  if not set then my plan was to change the settings for smaller circles.

PaulWessel:
For oblique projections, the N arrow is not parallel to the meridians.... This messes up the justification and offset of the oblique N arrow.  I have tried to tackle this.  For the fancy star it is not an issue since it is sort of like a circle.  The plain N arrow with cross and circle is more of a rectangle, with the width = 0.5 height.  This was not even adjusted for in master.  I  have fixed that but also tried to tackle the justification for arbitrary angle.  Because of the letter N at the end of the arrow it is nontrivial to do it perfectly.  This comes close enough for government work I hope.  See what you think as the movie below loops through all 360 degrees (a fakery).  The red lines indicate the requested offset from the map frame.For the common case of N straight up I can do exact in PostScript.  Let me know if @joa-quim, @Esteban82 and @meghanrjones are happy with this improvement before I merge my changes into this branch:

https://user-images.githubusercontent.com/26473567/150436735-6da85514-7808-4fa3-ac24-959661c4f4a7.mp4

Other questions:

1. The fancy rose: Shall we flip the text angles when things go upside down? (Fortunately not an issue for N in the plain rose!)
2. DO you think the label offset between arrow and N is too big? It ought to be the same as in the fancy but it looks bigger.
3. ???

meghanrjones:
Looks nice! The offset doesn't seem too bad.

> The fancy rose: Shall we flip the text angles when things go upside down? (Fortunately not an issue for N in the plain rose!)

Could we make this an option? For animations, I think it would look too jumpy to always flip the text angles when upside down, but it would be nice to have the choice for static maps.

joa-quim:
I think it’s reasonable that Antarcticlanders don’t see the cardinal letter upside down. But it’s also valid that on movies should not show jumpy texts. An option may solve it but I think static cases should take the default behavior.

PaulWessel:
Well, the approximate look violates the GMT prime directive (perfection first if possible), and being smarter than the average bear (just barely) I decided to do it exactly in _PostScript_.  I also added **+F** which is like **+f** (fancy) but adjusts text angles for readability from south.  Here you can see both results (although we would not use +F in a movie usually).

https://user-images.githubusercontent.com/26473567/150464842-62bb17b7-5bf6-477a-a2e0-5ce9da9b3a01.mp4

PaulWessel:
There is much more work to do with automatic scales as well, and there are **bugs** in master (and this branch) related to painting or outlining a background panel behind roses (via **-F** in **basemap** and **coast**, for instance).  I prefer not to make this branch more complicated that it already is so those items can be tackled after this one has been approved.  Here is a summary what this branch does:

1. Sets default values for the size of map roses: 10% for directional rose and 15% for magnetic rose of max map dimension.
2. Allows sizes of roses to be specified in percent of max map dimension (in addition to regular plot units).
3. Adds **MAP_EMBELLISHMENT_MODE** = auto|manual to the mix.  FYI @joa-quim, the decision on how this affects classic will be _deferred_ until all the embellishment updates and bug fixes are completed so I know the full picture.  For now, classic has this set to manual so that users of master and classic wont have to deal with changes back and forth until we finalize it.
4. Add **+F** to complement **+f** for fancy directional rose, which ensures label baselines are in the ±90 range [Default rotates with projection]
5. Auto-compute map rose offsets for **MAP_EMBELLISHMENT_MODE** = auto when **+o** not set, to 10% of size.
6. For smaller magnetic roses (size < 2.5 cm) the auto-default for **+t** is changed from 30/5/1/30/5/1 to 90/30/3/45/15/3.
7. Justification for the rectangular-shaped plain N-arrow symbol is not perfectly justified and fit using _PostScript_ computations regardless of map projection rotations of North [see movie above].

Two new tests highlights the improved justification where labels no longer exceeds the bounds (_dir_roses.sh_ and _mag_roses.sh_).  I have not uploaded PS files for these or the examples and scripts that now "fail" in a good way.

I know @Esteban82 just approved as I was typing up this comment, so please have a look and I think I would also like to hear from @meghanrjones and @joa-quim on this before merging anything.  I do need this eventually merged so I can tackle the bugs in **-F**.

meghanrjones:
Looking through the failing tests, I think most results look better now but the N label on GMT_obl_nz is too small to read on my modestly sized computer screen.

Is the font for the labels in the new tests the same as the map annotations? It looks much bolder.

PaulWessel:
NZ: These are tiny maps (3i) so default symol is 0.3 inches or about 0.75 cm.  It is small - the original had a fixed 0.5i symbol.  I can restore to 0.5i but will still fail due to (now) correct placement.  This figure is included in the cylindrical doc section so we do want it to be readable. OK, have set it back to 0.5i since the script is not shown anyway.

There have been no change to which font is used, only size of that font.

meghanrjones:
> NZ: These are tiny maps (3i) so default symol is 0.3 inches or about 0.75 cm. It is small - the original had a fixed 0.5i symbol. I can restore to 0.5i but will still fail due to (now) correct placement. This figure is included in the cylindrical doc section so we do want it to be readable. OK, have set it back to 0.5i since the script is not shown anyway.

The rose on the right panel looks to still be shrunk:
image



> There have been no change to which font is used, only size of that font.

Ah, as you said it is set by FONT_TITLE which is Helvetica-Bold.

My main concern is about MAP_EMBELLISHMENT_MODE taking precedence over FONT_TITLE in this PR. I think that a user should not need to set both MAP_EMBELLISHMENT_MODE=manual and FONT_TITLE=9p for their font preference to be applied, since they may prefer auto offset with a specific font size or already have scripts in modern mode with a specific font size. For example, the relevant line in `GMT_obl_nz.sh` sets ``--FONT_TITLE=9p`` but this does not affect the plot in this PR.

PaulWessel:
TYpo in script, now same size.
As for FONT: yes the 9p has no effect.  I see the auto affecting lots of parts and it would make little sense to micro-manage font sizes from the user.  THis is an older script so I should just remove that setting (predates this PR).

meghanrjones:
> As for FONT: yes the 9p has no effect. I see the auto affecting lots of parts and it would make little sense to micro-manage font sizes from the user. THis is an older script so I should just remove that setting (predates this PR).

Do you mean that you prefer the current behavior in this PR of MAP_EMBELLISHMENT_MODE overriding FONT_TITLE settings?

PaulWessel:
Yes, because I do not know how to track what was set manually.  Remember, this is after the auto-scaling of fonts etc have already been set for the map.  So without a set of book-keeping I would not know if I should not change the font-size because the user wanted 24p.  They can change the font, but not manually set a size and still benefit from all the other auto-things.  I think the vast amount of users will be pleased they dont have to think about these settings, and a few may wish to do things entirely manually.

meghanrjones:
OK, I think it's good to go after removing the now irrelevant settings from the scripts.

PaulWessel:
FYI: This PR got delayed a bit because the alignment of the N-arrow in example 4 was off.  The solution is to temporarily reset the PostScript matrix projection, compute the offsets, then reinstate the projection (the -p60/30 as implemented in PostScript) as the offsets need to be measured without perspective.  Now, ex04 has a good default placement and size I think.  About 8 tests related to map roses, including some new tests with no PS files, fail.  I think we will add these to dvc soon.

PR #6220, opened 2022-01-12
Let grdinfo report default CPT for a grid

PaulWessel:
Like remark, title, command etc, we now also list Default CPT and give blank if no such thing.


meghanrjones:
I think it would be best to put the new item at the end, in case anyone has scripts relying on e.g., the grid file format being on the 5th line.

meghanrjones (approved):

PR #6230, opened 2022-01-17
Improve openMP variables and macros to avoid unnecessary warnings

PaulWessel:
The _gmt_M_grd_loop_ and similar macros result in tons of warnings due to the problem with Windows not supporting OpenMP > 2.0 and therefore forces us to have signed loop variables.  Because GMT internals, including published API structures, uses unsigned variables for quantities that cannot be negative, this means we either have to add many casts or risk not noticing legitimate warnings in a sea of false positives.

This PR addresses this quandary by introducing a typedef _openmp_int_ that is set to `int` under Windows and `unsigned int` elsewhere, and we use casts to keep things quiet regardless of what _openmp_int_ is. I use openmp_int wherever those macro-loops are used, wherever we use OpenMP, and in places were we may want to add OpenMP in the future.

I have tested this branch for both signed and unsigned typedef and all warnings have gone away.  All tests that pass in master still pass in both compilations.  Before approving and merging we should see if anything funny is discovered by @joa-quim and/or the CI tests outside macOS.


meghanrjones:
Do you want to run the full tests on this PR before merging or just go based off the few simple tests run on all PRs? Running the full tests can be done by uncommenting L7 in the tests workflow (https://github.com/GenericMappingTools/gmt/blob/no-warnings/.github/workflows/tests.yml). We would need to recomment that line before merging. The CI tests take a few hours to run.

src/gmt_grd.h
```diff
@@ -34,6 +34,23 @@
...
+ * in OpenMP 2.  Because we do not want a 2008 implementation to force changes to GMT structure I have
+ * now introduced a typedef for use with such loop variables.  This was particularly needed for our
+ * 4 macros, like gmt_M_grd_loop, which prior to this change reuslted in hundreds of warnings about
+ * sign/unsigned mismatches due to the forced signed int loop variables.  I have updated all places
+ * where these macros are used and where OpenMP is used and where it may be used in the near future,
+ * so that things are properly cast to openmp_int, whatever that may be set to be.  With this, the
+ * false warnings go aaway and we can better notice actual warnings due to new implementation changes.
+ * P. Wessel, Jan-17-2022. */
+
+#if _WIN32
```
joa-quim, line 48: This does not test for VisualStudio. MinGW has it too (no idea on what OMP it uses). To detect VisualStudio use
```
#ifdef _MSC_VER
```

PaulWessel: OK, since MinGW uses gcc (right?) I can make this change.

joa-quim: And the ``#if _WIN32`` would probably block the Intel compiler

PaulWessel: OK, done

meghanrjones:
The tests workflow had two new failures on linux and windows, though I was not able to reproduce these failures locally on linux:
499 - test/grdlandmask/greenwich.sh (Failed)
501 - test/grdlandmask/mask.sh (Failed)

The full test results are at https://github.com/GenericMappingTools/gmt/actions/runs/1720120943. I'll revert the tests workflow.

src/grdlandmask.c
```diff
@@ -384,7 +386,7 @@ EXTERN_MSC int GMT_grdlandmask (void *V_API, int mode, void *args) {
 	if (Ctrl->E.linetrace)
 		X = gmt_M_memory (GMT, X, nx_alloc, struct GRDLANDMASK_BINCROSS);
 
-	nx1 = Grid->header->n_columns - 1;	ny1 = Grid->header->n_rows - 1;
+	nx1 = (int)Grid->header->n_columns - 1;	ny1 = (int)Grid->header->n_rows - 1;
```
meghanrjones, line 389: Why are the changes for grdlandmask `int` rather than `openmp_int`?

PaulWessel: I think I tried and got failed tests. It seems col_min could be negative and needs to be set:

```
col_min = irint (ceil (lon_w * HH->r_inc[GMT_X] - Grid->header->xy_off));
if (col_min < 0) col_min = 0;
```

and that would fail as _openmp_int_ on macs and linux where it would be unsigned and a huge positive value.

PaulWessel: But I see elsewhere in that module I use MAX (0,....) to avoid this, so I will try first.

meghanrjones: OK, I'll rerun the tests on windows to see if this fixes the test failures.

PaulWessel: Thanks, I do see there is potential for negative values and hence I think it is safest to leave as ints.

meghanrjones: After the latest commit I get the same failures in master and this branch on windows.

PaulWessel:
OK, I will work some more on grdlandmask then.

meghanrjones:
> OK, I will work some more on grdlandmask then.

Sorry, I wasn't clear. I do not think you need to do more work on grdlandmask. There are still failures on windows, but they are identical on master and this branch so unrelated to openMP implementation. I'm checking linux now and then will approve.

PaulWessel:
OK, I misunderstood - too much multitasking and zooms at the same time.

meghanrjones:
I ran into issues with the tests hanging - I think on [firmoviscous.sh](https://github.com/GenericMappingTools/gmt/blob/no-warnings/test/potential/firmoviscous.sh) so I am holding off on approving until I look into it more.

PaulWessel:
I fixed a few things in **grdmask** related to OpenMP.  Now, when I compare the test results from the MP build with the non-OpenMP build I get the same 12 failures in master (related to lines).  So I think nothing is broken in the PR.

However, I learned that the **grdmask** with OpenMP looping over the few nodes inside a small search radius is much slower than non-OpenMP due to the overhead of setting up the extra threads just to compute a few points.  So either we need to decide during run time that the size of the radius is large enough to warrant doing it or we skip the openMP part.  That would mean an if-test and near-duplicate coding somewhere.  Given the speed up due to #6253 I think we should just yank the OpenMP part from the **-S** section.

PaulWessel:
Let me know if this latest version is acceptable - no warnings on my side and I removed one OpenMP loop from grdmask (kept the one involving polygons).

meghanrjones (approved):

PR #6242, opened 2022-01-19
Set default -di column to 0 in xyz2grd -Z

PaulWessel:
Since **-Z** implies a single data column of z-values, the default GMT_Z column setting is changed to GMT_X,
Fixes the confusion in this forum [post](https://forum.generic-mapping-tools.org/t/command-string-exceeds-upper-length-of-320-characters/2529/3).


meghanrjones (approved):

PaulWessel:
@meghanrjones, how do we rerun these checks (they fail for same reason that Joaquim's example failed).

PaulWessel:
I figured it out.

Issue #6241, opened 2022-01-19
Bug with dcw-collections local file. 


Esteban82:
I am testing dcw-collections with my local file (located in `.gmt/`). I try some examples and it works fine.

The problem was when I wanted to use codes from the official list (located in `.gmt/geography/dcw/dcw-collections.txt`) that I are not included in my local file (like IHO33). If I delete my local file, then GMT finds the code in the official list and I get the map.

```
gmt coast -RIHO33 -W -B -png mapa
coast [WARNING]: No country code matching IHO33 (skipped)
coast [ERROR]: South is outside -90 to +90 degree range
coast [ERROR]: General map projection error
```

PaulWessel:
Just to make sure: Your local file in .gmt is called dcw.conf, yes?

Esteban82:
No, it wasn't. Now I don't have that problem. 

Now I have this issue. Is it suppose that in `dcw.conf` I should be able to redefine the lists and regions?

For example, I redefine the bounding box of South America (in `dcw.conf`), and when I make I still get the official region of SAM (not the one of `dcw.conf`).


```
gmt coast -RSAM -W -B -png mapa

tag: SAM South America
region: -81.3288/-34.7938/-55.9805/0.4645
```

PR #6244, opened 2022-01-19
Read dcw.conf first so it takes precedence over dcw-collections.txt

PaulWessel:
We want the user to have the ability to add their own collections and even redefine what any particular tag also present in the system file should mean.  If I want Europe to include the Bouvet Island in the S Atlantic then I am allowed to give a crazy -R setting for that name.  This requires we read the user dcw.conf _first_ since the search through the array is linear.

@Esteban82 please check if this fixes your issue in #6241.


Esteban82 (approved):
Yes, it works. Test it for regions and lists.

Issue #199, opened 2018-12-03
Paper size for a map


frspin:
In my proc for getting my map I use:

gmt set PS_PAGE_ORIENTATION=portrait
gmt set PS_MEDIA 700cx450c

this, with defining a region
region=-R7.27/8.27/43.5/44.0
and a projection
projection=-JM65c

give me a file with size 690x480 mm

After recent commit:

Reduce chance of PostScript clipping if exceeding canvas in modern mo…

my file size jumped to 1478x480 mm, with a left zone completely void.

Reversing 
#define GMT_PAPER_MARGIN	40	
to
#define GMT_PAPER_MARGIN	5

solve the problem.

Is this a bug or I am missing some other initialization?

Regards

Franco Spinelli

frspin:
Il 03/12/18 09:23, Paul Wessel ha scritto:
> Hm, I ran the 800+ tests before committing that PR and there were no
> changes, so could you please post the script you ran to get the problem?

It is attached to this mail

Regards

Franco Spinelli

PaulWessel:
Sorry, not seeing anything here or in the email?

frspin:
My fault. I have used email client for my answer.
Now I am writing from Web interface on Github

[mappa_100k.zip](https://github.com/GenericMappingTools/gmt/files/2640425/mappa_100k.zip)

Regards

Franco Spinelli

seisman:
@PaulWessel You can reproduce the bug using following simple code:
```
gmt begin map pdf
gmt basemap -JH65c -R7.27/8.27/43.5/44.0 -Bafg
gmt end
```

seisman:
OS: macOS 10.14.1
gmt version: 6.0.0_2a0f78d
gs version: 9.26

Full debug information with -Vd:
```
gmt [DEBUG]: Obtained the ppid from parent: 35488
gmt [DEBUG]: Enter: gmtinit_new_GMT_ctrl
gmt [DEBUG]: GMT->session.SHAREDIR = /usr/local/Cellar/gmt/HEAD-2a0f78d/share/gmt
gmt [DEBUG]: GMT->session.HOMEDIR = /Users/seisman
gmt [DEBUG]: GMT->session.USERDIR = /Users/seisman/.gmt [created]
gmt [DEBUG]: GMT->session.CACHEDIR = /Users/seisman/.gmt/cache [created]
[...]

PR #6247, opened 2022-01-21
Turn on ~1m margin for auto-size plots

### LINKED ISSUE #195: Incorrect cropping if the origin of second panel is >12 cm below the first panel

PaulWessel:
We started with 5 inches but as some people prefer to lay out plots going down they can quickly run out of space.  This moves the starting point to (40i, 40i) which gives us about  1 meter marings for autoplots to move into.  This seems big enough (poster boards are usually < 1m tall?) but perhaps worrying about people making 5+ meter maps is misplaced and we should set it to 100 inches to be absolutely sure this issue is put to rest? Let me know.

Closes #195.

### DIFF
--- a/src/gmt_constants.h
+++ b/src/gmt_constants.h
@@ -88,7 +88,7 @@
 #define GMT_TOP_MODULE	1	/* func_level of top-level module being called */
 
 #define GMT_PAPER_DIM		32767	/* Upper limit on PostScript paper size under modern mode, in points (~11.6 meters) */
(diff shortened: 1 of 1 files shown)


meghanrjones (approved):
Tests look good and neither of the issues from https://github.com/GenericMappingTools/gmt/issues/195 and https://github.com/GenericMappingTools/gmt/issues/199 occur.

PR #6249, opened 2022-01-22
DCW collection documentation

PaulWessel:

Add cross-links from the **-R** common option documentation as well as coast **-E** documentation to the DCW Collections section towards the end of **coast**.  This section discusses the collections and the ability for users to add their own and explains the required format.


Esteban82 (approved):
Looks fine!

Esteban82:
```
To specify a whole continent, prepend **=** to any of the continent codes **AF** (Africa), 
**AN** (Antarctica), **AS** (Asia), **EU** (Europe),   **OC** (Oceania), **NA** (North America), 
 or **SA** (South America), or spell out the full continent name.
```

In the docs are text like the previous one. I think that option was deprecated so it should be deleted, right? 
Also, there now the user will get a different region if uses the code (=EU) or the full continent name. 

So I think that text like this should be deleted from the docs.

PR #6253, opened 2022-01-22
Speed up grdmask -S

PaulWessel:
When we use **grdmask -S** we basically search all nodes within the specified radius of a data point and set them to _inside_.  However, for denser data and/or larger radii, this results in the same nodes being visited multiple times and each time we need to compute a radial distance to see if we truly are inside or not, which can be expensive of a sphere.  This PR simply adds a char array _node_is_set_ to keep track if we have flagged a node as inside and if we visit it again we immediately skip to the next node. This also speeds up **surface** when masking is selected with **-M**.

When **-S** is not used we look for nodes inside polygons.  While not as helpful, the _node_is_set_ scheme is used there too in case there are overlapping polygons.


Esteban82 (approved):
Ok. I test it a bit and looks fine.

PR #6254, opened 2022-01-22
Speed up psmask -S

PaulWessel:
Similar to **grdmask -S** improvement in #6253: When we loop over all the nodes within a radius from a point we check if we have been here before prior to calling the distance calculator.  This will speed up execution for dense points or larger radii.


Esteban82:
How can I test it?

PaulWessel:
No new tests but I think it (a) breaks no existing tests and (b) checking a value is faster than calling spherical trig.  Savings depends on data set density and chosen radius.  For **grdmask** I got factors or 3-4 faster for some combination so I think as long as nothing breaks this is a win.

Esteban82 (approved):

PR #6258, opened 2022-01-23
It is time to make GDAL a required prerequisite

PaulWessel:
Lots of **GMT** functionality now depends on **GDAL**, from handling image i/o, including multilevel geotiff data files, making geotiffs from plots, reading a variety of grid formats as well as handling shapefiles (there are plans to handle more table formats via the GDAL/OGR interface). We also plan to start using more **proj** features as available via **GDAL** (we will eventually also make **PROJ** a requirement but since **PROJ** is required by **GDAL** it does not really add anything extra).  This PR makes the practicality of depending on **GDAL** (already included in all our installers) _formal_.

I have left it at WIP for now to make sure we first all agree that

1. Yes, this needs to be done.  Having > 60 `#ifdef HAVE_GDAL` around the code is complicating things and those 13 people who manage to use GMT but cannot handle the requirement of **GDAL** simply need to get onboard for the sake of the rest of us.
2. We have updated any documentation in this repository that mention **GDAL** as an optional prerequisite.
3. We have let the CI and the various developers had a change to test this branch and find it satisfactory.
4. We have ensured our CMake settings are correct so that Linux distributions will pick up on the change once 6.4.0 gets released later this spring.

Once we approve and merge I believe there are other repos (website, others?) that will need some updates.  Let us know by adding comments herein.


leouieda (approved):
Completely sensible thing to do 👍

If I remember correctly, we already have it as a required dependency on the conda-forge package from the start.

anbj:
This is good. **gdal** is good. That said, is this primarily because of **gdal** libs (i.e. code) being used in **gmt**, or the drive to *make things easier for the user*? Having several tools in a toolbox to do different things is by definition (almost) always a good thing. Please do not create a *black box*.

joa-quim (approved):

joa-quim:
> That said, is this primarily because of gdal libs (i.e. code) being used in gmt,

Yes, and also because PROJ is mandatory in GDAL and we may want to use more things of PROJ directly (or indirectly).

anbj:
> > That said, is this primarily because of gdal libs (i.e. code) being used in gmt,
> 
> Yes, and also because PROJ is mandatory in GDAL and we may want to use more things of PROJ directly (or indirectly).

Baking in PROJ with gmt (i.e. being able to use PROJ projections) sounds very attractive.

meghanrjones:
I agree.

I think making GDAL a required dependency should be included in the changelog for anyone who's been building GMT without gdal.

We'll need to move `libgdal-dev` from the optional packages to the required packages in `ci/install-dependencies-linux.sh`, update the job name "Linux (without GDAL et al.)" in `.github/workflows/build.yml`, add gdal to the install steps in `ci/vercel-docs.sh`, and update all the GitHub wiki pages for gdal to be in the required rather than optional section. There's also a few places throughout the gmt docs where statements such as 'provided GMT was compiled
with GDAL support' need updating.

PaulWessel:
I have searched for GDAL in all GMT files and gone through manually and removed any mention of GDAL being optional or requiring special options.  It should be all good to go as far as GMT is concerned.  Once approved and merged we need to tend to the other issues that @meghanrjones outlined.

meghanrjones (approved):

PaulWessel:
When should we change the information on the wiki? The GDAL being required is not true for the official releases yet.

PaulWessel:
Perhaps @meghanrjones can fix the CI script that does **not** use GDAL?

meghanrjones:
> When should we change the information on the wiki? The GDAL being required is not true for the official releases yet.

I think at release-time makes sense.

PR #6260, opened 2022-01-24
No longer document the =AF code for continents

PaulWessel:
While backwards compatible for old scripts that may use **-R**=AF etc., going forward we only document full names like **-R**Africa.  This PR fixes that and also adds collections as an option to the -R usage description.


Esteban82 (approved):

joa-quim:
Let me say again that I disagree that we ship an *Europe* that includes the entire Siberia. I don't care that UN says it's fine. For me, and for many users I believe, it isn't.

Esteban82:
> Let me say again that I disagree that we ship an _Europe_ that includes the entire Siberia. I don't care that UN says it's fine. For me, and for many users I believe, it isn't.

Why do you think so? I just run` gmt coast -REurope -W -png mapa` and I got map from Ireland to the Urals and from Crete to Norway.

joa-quim:
I'm talking from that pdf with a list of maps that you prepared (don't find it right now). There, Europe went to Bering Strait.

Esteban82:
It was an old version. Then I modified it. Now if you want to get the UN definition you must use `-R"UN Europe"`.

joa-quim:
I'm talking of
```
gmt coast -R"UN150" -Ba -W -png mapa
```
and, BTW, shouldn't at least one of these work?
```
gmt coast -R"UN150 UN Europe" -W -png mapa
coast [WARNING]: No country code matching UN150 UN Europe (skipped)

gmt coast -R"UN Europe" -Ba -W -png mapa
coast [WARNING]: No country code matching UN Europe (skipped)
```
mapa

Esteban82:
> gmt coast -R"UN150 UN Europe" -W -png mapa
> coast [WARNING]: No country code matching UN150 UN Europe (skipped)

This shouldn't work, 


> gmt coast -R"UN Europe" -Ba -W -png mapa

This works for me.

joa-quim:
Hmm, my ``dcw-collections.txt`` is the one that comes in the ``dcw-gmt-2.1.0.zip`` file.

Esteban82:
> I'm talking of
> 
> ```
> gmt coast -R"UN150" -Ba -W -png mapa
> ```

I don't see any problem here. Probably it won't be one the most use lists but I don't think that anyone will confuse "Europe" with `UN150`or with "UN Europe". That is why I include "UN" in the name and leave "Europe" for a bounding box region.

joa-quim:
Adding ``UN`` does not remove ``Europe`` from the name.

Esteban82:
So, basically you don't like the UN M49?

And what do you propose?
1. Delete them UN M49 from the dcw-collections?
2. Only delete or rename "UN Europe"?
3. Something else?

joa-quim:
Since it makes no sense, delete the ``UN Europe`` tag all together.
... and do not include Greenland in North America. We have discussed that before.

Esteban82:
If someone wants to have Europe following UN convention? 

Greenland is NOT included in `-R"North America"`. It is included in `-R"UN North America"`.

joa-quim:
> If someone wants to have Europe following UN convention?

Ask UN to make one for them? :joy:

Current Secretary General is Portuguese. I would be very curious on his opinion on this subject.
I know one person, who knows another person, who knows him but I still hesitant to send this question down the chain.

PaulWessel:
Great, so @joa-quim will get to the bottom (top) of this! Ha!

I think while we are allowed to disagree with the official boundaries, and even ridicule them, we are not allowed to change them.  So people like me and Joaquim may wish to override Europe with a

```
tag: EU Europe
region: w/e/s/n
```
record in our ~/.gmt/dcw.conf

joa-quim:
No need that. It doesn't even works for me. 
But that is obviously not the point.

anbj:
I would like to point out the NO code (Norway) should include the Bouvet Island, located around 60S. This means Europe should extend almost down to Antarctica.

Esteban82:
There is a special isocode for the Bouvet Island. 
BTW the UN says that that island is part of South America.

joa-quim:
And not forget French Guiana, Martinique, Saint Pierre and Miquelon (goes into Canada), Kergelen, New Caledonia and ... the Falklans.
Yea, Europe spans >> 1/2 Earth. I was write all the time that UN Europe is very wrong.

anbj:
I’m sorry Esteban. It was meant as a joke. But Bouvet Island does indeed belong to Norway. As Joaquim says; Europe is a very mixed bag. 

Speaking of Norway (code: NO); before Svalbard and Jan Mayen was included. It’s not anymore - only mainland Norway is now included when using ‘NO’. Have the (original) codes been changed, as well as the inclusion of all the new codes?

PaulWessel:
> Speaking of Norway (code: NO); before Svalbard and Jan Mayen was included. It’s not anymore - only mainland Norway is now included when using ‘NO’. Have the (original) codes been changed, as well as the inclusion of all the new codes?

I think I am to "blame" for this.  Both Svalbard and Jan Mayen has specific codes in DCW and we decided that "most" people wishing to make a map of Norway are not really planning to include Svalbard.  Svalbard may be under Norwegian jurisdiction but it is not part of sovereign Norway - i.e, see Russian coalmines etc.

anbj:
> > Speaking of Norway (code: NO); before Svalbard and Jan Mayen was included. It’s not anymore - only mainland Norway is now included when using ‘NO’. Have the (original) codes been changed, as well as the inclusion of all the new codes?
> 
> I think I am to "blame" for this. Both Svalbard and Jan Mayen has specific codes in DCW and we decided that "most" people wishing to make a map of Norway are not really planning to include Svalbard. Svalbard may be under Norwegian jurisdiction but it is not part of sovereign Norway - i.e, see Russian coalmines etc.

Thanks for the explanatuon Paul. And I agree - this is a better way.

joa-quim:
There is subtle situation in these jokes. The need to have limits to the mainlands only. GMT.jl has now a WebMapServer function that can downloads images given a specified limit. Given the potential image size, I'm not interested in saying ``R="PT", resolution=100`` and try to download an image with 100 m resolution that goes from the middle Atlantic till Spain. As Spanish, French and Norwegians (Bouvet triple Junction included) would not be interested in doing the same.

Have not solution for this though other than creating a new tag that means *mainland only*.

Esteban82:
For countries with some territories very far away from a mainland (Portugal, Spain, Ecuador) I think that there are 3 options:

1. Add the internal sub-divisions of those countries. In the case of Portugal we could add PT-20 and PT-30 for the Autonomous regions of Azores and Madeira). So then the mainland Portugal could be call from the rest of districts (a list in dcw-collections could be created). 
2. Create a bounding box region in dcw-collections. 
3. Within DCW separate the mainland from the islands like if they were an special territories (like the cases of French Guiana,  Svalbard, Puerto Rica, Aruba, etc). So we will have PT (for only the European part of Portugal), PT-20 and PT-30. Maybe the more practical solution but we will modifying the ISO 3166. 

BTW, Chile should be in the same list but the Easter island are not included in CL.txt.

Esteban82:
4. Is It possible to do this `gmt coast -EPT -M | gmt select -REurope` ?

joa-quim:
The ``gmtselect`` option is not a solution because what we are seeking is a boundingbox.

Option 1. is interesting by itself because it allows plotting those (and the other similar cases) subdivisions individually but it doesn't solve the issue because we cannot change what, for instances, ``PT`` means in DCW. So, the only practical option is your option 2.

Esteban82:
> The `gmtselect` option is not a solution because what we are seeking is a boundingbox.

And this: `gmt coast -EPT -M | gmt select -REurope | gmt info -Ie` ?

I got this result: 
`-R-9.496162607/-6.183120789/36.9611787549/42.1454781369`

And how do you suggest to name the tag for option 2?

joa-quim:
Any solution based on ``coast -E -M`` is not an option. _I want_ efficiency.

> And how do you suggest to name the tag for option 2?

Add a ``c`` for continental. *e.g.* ``PTc``, ``ESc``, ``CLc`, ``ECc``

Esteban82:
Ok. 

I will add the tags. BTW they are case-insensitive.

PaulWessel:
How many countries are there that have these colonial appendices?

Esteban82:
> How many countries are there that have these colonial appendices?

I think that the mains cases are:
Portugal
Spain
Costa Rica (Isla del Coco)
Ecuador 


Here I created a pdf with all the countries with exact bounding box ([All2.pdf](https://github.com/GenericMappingTools/gmt/files/7938377/All2.pdf)). There are some minor cases.

Esteban82:
So, this PR can be merge, right?

Paul, should I make a PR to add the regions of the last comments?

PaulWessel:
Sorry, yes I think so - this got left for too long.

Issue #6218, opened 2022-01-12
Padding issue with gridline registered datasets


meghanrjones:
For the gridline registered grids, there is often empty space between the map border and grid edge:
```
gmt grdimage @earth_relief_01d_g -R110/140/20/35 -JB125/20/25/45/12c -B -png test_relief_g
```
![test_synbath_g](https://user-images.githubusercontent.com/14077947/149202966-aa0de428-d21b-4daf-b15d-c561ab375c63.png)

_Originally posted by @meghanrjones in https://github.com/GenericMappingTools/remote-datasets/issues/9#issuecomment-1011348913_

PaulWessel:
The real reason seems to be a combination of this section in grdimage's parse function:

```
	if (!GMT->common.n.active && (!Ctrl->C.active || gmt_is_cpt_master (GMT, Ctrl->C.file)))
		/* Unless user selected -n we want the default not to exceed data range on projection when we are auto-scaling a master table */
		n_errors += gmtinit_parse_n_option (GMT, "b+c");
```

and the changes in gmt_grdio.c. We clearly need the extra row/column and perhaps I can trick the min/max calculation by manipulating the pad temporarily.

meghanrjones:
Fixed by #6272

PR #6272, opened 2022-01-27
Exclude grdimage padding when setting CPT limits

PaulWessel:
See #6218 for background.  This problem had its origin in me trying to limit the extra rows/columns used in order to allow spline resampling near the boundaries.  I had added a check that turned bicubic sampling off (unless **-n** was set) to avoid extending the region and getting z-values that _could_ bias the min/max range and hence the CPT stretching.  This PR removes that check (we _do_ want bicubic sampling as default) but we now recompute the grid zmin/zmax to only apply to the region specified on the command line.  Lots of "failures" because

1. We are undoing the previous decision related to the spline (hence tiny mismatches throughout most images
2. The z-limits for the inside region may differ from those of the extended grid region, affecting the CPT lookup.

However, I believe this is a better fix. We should not update any doc *.ps as these will soon go into DVC anyway.


meghanrjones:
There's a problem in this branch where the maximum and minimum values are not colored despite being within the range of the colormap made by makecpt. Here is a comparison of a section of ex29:

| old | new |
|-----|-----|
| image | image |

PaulWessel:
OK, separate issue I think.  Before I did this if no -n:

		n_errors += gmtinit_parse_n_option (GMT, "b+c");

Perhaps now we should do this if no -n given:

		`n_errors += gmtinit_parse_n_option (GMT, "c+c");`

That would be in line with bicubic interpolation being the default and for avoiding splining beyond the min/max we need +c.

PaulWessel:
I have added in the modified setting of **-nc+c** as the default in **grdimage** if **-n** is not set by the user, and I also document that as well.  This fixes the issue you highlighted above.  I am pretty sure the previous **-nb+c** was a mistake since the docs says bicubic is the default interpolator.

meghanrjones (approved):

PR #6279, opened 2022-01-28
Allow -I0/dy and -Idx/0 in gmt info

PaulWessel:
This is useful when you want an exact range in one dimension but a rounded range in the other.  See the forum [post](https://forum.generic-mapping-tools.org/t/domain-of-profile-with-gmt-info/2566/2) for background. In effect, **-Ie** is a shorthand for **-I**0 or **-I**0/0.


Esteban82 (approved):
Thanks Paul!

PR #6289, opened 2022-01-31
Enhanced node determination in grdmask for polygon input

PaulWessel:
This PR adds a new **-C**_mode_ option to **grdmask**, more or less the same **-C** option offered in **grdblend**.  There, it controls which grids contribute to the final output node (instead of blending, a single grid's node will prevail) while here it selects the single polygon whose z-value sets the node.  The **-C** option determines which polygon:

1. **-Cf**: The first polygon to cover a node sets the node value.
2. **-Co** The last polygon to cover a node sets the node value. [Default]
3. **-Cl:** The polygon with lowest z-valueto cover a node sets the node value.
4. -**Cu:** The polygon with highest z-valueto cover a node sets the node value.

The actual z-value is controlled by **-N**. With default **-Co** we retain backwards compatibility while other modes offer new capabilities.  A new script (_maskmode.sh_ based on _ogrtest.sh_) was added which shows the output for the 4 different modes. I used awk to shuffle some polygon z-values so that the effect of all modes can be seen:

maskmode


meghanrjones:
I think this sentence in the module description should be updated to acknowledge the option to have more values in the output grid with **-Nz**:

> The nodes defined by the specified region and lattice spacing will be set equal to one of three possible values depending on whether the node is outside, on the polygon perimeter, or inside the polygon.

PaulWessel:
OK, made a change.

meghanrjones (approved):

src/grdmask.c
```diff
@@ -616,11 +651,11 @@ EXTERN_MSC int GMT_grdmask (void *V_API, int mode, void *args) {
 
 					/* Here we will have to consider the x coordinates as well (or known_side is set) */
 #ifdef _OPENMP
-#pragma omp parallel for private(col,xx,side,ij) shared(n_columns,GMT,Grid,do_test,yy,S,known_side,row,Ctrl,z_value,mask_val)
+#pragma omp parallel for private(col,ij,xx,side,z_to_set) shared(n_columns,Grid,row,Ctrl,node_is_set,GMT,do_test,yy,S,known_side,z_value,mask_val)
```
meghanrjones, line 654: The new test maskmode.sh fails on Linux when OpenMP support enabled. I have not checked if it's the same on macOS, but I would assume so. There's also a compiler warning that z_to_set may be uninitialized. @PaulWessel do you notice anything obvious wrong with the OpenMP implementation here? Otherwise I can debug.

PR #6294, opened 2022-02-02
Improve error message when parsing default options

### LINKED ISSUE #6293: Improve the message for unrecognized options

One of the more mysterious error messages a newbie can get is this:

`basemap [ERROR]: Unrecognized option -<`

In this case (and same with a few other modules) it is because the user gave an argument that was seen as an input file and no file was excepted.  But inferring that from the message is not obvious.

The solution is to let _gmt_default_error_ report the argument as well so it is clearer.  However, it only takes the option letter as the argument.  if I change the argument to the option itself then we have problems with existing use of _gmt_default_error_ in MB-system, for instance, and MB-system cannot change this immediately as they need to work with recent GMT versions.

My proposal is to add a new function

`int gmt_default_option_error (struct GMT_CTRL *GMT, struct GMT_OPTION *opt);`

that will be used instead of gmt_option_error which will just remain buried in the library and available for programs that expects it (I will ping Caress about the change if he wants to do anything about it). It can also be smart enough to realize that the option in question (that failed) is` -<` which means the input file and give a clearer message.  This simple change will affect 149 files.

Any comments?

PaulWessel:
See #6293 which is very likely what happened in this [forum](https://forum.generic-mapping-tools.org/t/running-gmt-example-as-powershell-cmdlet-fault/2583/2) case.  This PR introduces _gmt_default_option_error_ which now calls _gmt_default_error_ and adds extra messages in case of i/o confusion.  Wile updating all modules, the real changes are in gmt_init.c.  Closes #6293.

As an example of the type of thing that happens when you miss a leading option hyphen:

```
(gmt basemap -RFR -Gred -B Lx1i/1i+c0+w5000k -pdf t
basemap [ERROR]: Unrecognized option -<
basemap [ERROR]: Module does not expect an input file [Lx1i/1i+c0+w5000k]
```

### DIFF
--- a/src/batch.c
+++ b/src/batch.c
@@ -345,7 +345,7 @@ static int parse (struct GMT_CTRL *GMT, struct BATCH_CTRL *Ctrl, struct GMT_OPTI
 				break;
 
 			default:	/* Report bad options */
-				n_errors += gmt_default_error (GMT, opt->option);
+				n_errors += gmt_default_option_error (GMT, opt);
 				break;
 		}
 	}
--- a/src/blockmean.c
+++ b/src/blockmean.c
@@ -247,7 +247,7 @@ static int parse (struct GMT_CTRL *GMT, struct BLOCKMEAN_CTRL *Ctrl, struct GMT_
 				break;
 
 			default:	/* Report bad options */
-				n_errors += gmt_default_error (GMT, opt->option);
+				n_errors += gmt_default_option_error (GMT, opt);
 				break;
 		}
 	}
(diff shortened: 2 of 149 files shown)


PaulWessel:
OK, some minor improvements:

```
gmt basemap -RFR -Gred -B Aasd/bb -pdf t
basemap [ERROR]: Unrecognized option -<
basemap [ERROR]: Aasd/bb was seen as an input file which is not expected by this module

gmt basemap -RFR -Gred -B Lx1i/1i+c0+w5000k -pdf t
basemap [ERROR]: Unrecognized option -<
basemap [ERROR]: Lx1i/1i+c0+w5000k was seen as an input file but looks like an option with modifiers; did you forget a leading hyphen?
```

meghanrjones (approved):

PaulWessel:
I made it a bit clearer for files now:

```
gmt  basemap -RFR -Gred -B Aasd/bb -pdf t
basemap [ERROR]: Module does not expect input files
basemap [ERROR]: Aasd/bb was seen as an input file which is not expected by this module
```

Issue #6288, opened 2022-01-31
inset -R -J -X -Y history


gd-a:
Not sure it is a bug but it seems that `inset` doesn't retain history within its block. Is it on purpose ?

Guillaume

gd-a:
... nevermind ... sorry

gd-a:
Ok ... I've retried ... it does conserve -X -Y but neither -R or -J

```
 gmt inset begin -DjBR+w3.5c+o0.25c/0.25c
		gmt coast -Rg -JG0/0/3c+w45 -Dc -Ggray -A5000 -Bg -Wfaint -X0.25 -Y0.25
	        gmt plot  <<- EOF
> delta-line -Wthin,black -Am
0 -89
0 89
EOF
    gmt inset end
```

PaulWessel:
It should retain -R -J once set inside the inset box.  Could you help by supplying a full test script?

gd-a:
```
gmt basemap -R0/1/0/1 -JX10c -B
gmt inset begin -DjBR+w3.5c+o0.25c/0.25c
                gmt coast -Rg -JG0/0/3c+w45 -Dc -Ggray -A5000 -Bg -Wfaint -X0.25 -Y0.25
                gmt plot  <<- EOF
> delta-line -Wthin,black -Am
0 -89
0 89
EOF
    gmt inset end
gmt end show
```
If I interpret correctly the black line placement, it seems to mean the -JG0/0/**3c** is not respected (it takes 3.5c)

PaulWessel:
I think for **gmt inset** you have to stick with the rules for the inset:

1. The projection should be given with ? as the scale or width
2. Use **-M** if you want to remove some of the inset space for margin padding

A successful script based on what you have above does not use **-X -Y** at all and looks like this:

```
gmt begin test png
	gmt basemap -R0/1/0/1 -JX10c -B
	gmt inset begin -DjBR+w3.5c+o0.25c/0.25c -F+pfaint -M0.25c
		gmt coast -Rg -JG0/0/?+w45 -Dc -Ggray -A5000 -Bg -Wfaint
		gmt plot  <<- EOF
		> delta-line -Wthin,black -Am
		0 -89
		0 89
		EOF
	gmt inset end
gmt end show
```

Yet there are clearly documentation issues here, so I am not quite sure what is the best way forward here.  Here are a bunch of concerns or issues.

[...]

PR #6297, opened 2022-02-04
Let inset take -C instead of -M

PaulWessel:

As discussed in #6288, I unfortunately selected **-M** as the option for adding interior margins for a map inset, even though **-C** as used in **subplot** is the most comparable situation.  This PR rectifies this by switching **inset** to use **-C** but of course retaining backwards compatibility with any **-M** use.  Some related points:

1. I ran all tests as they were first to ensure the **-M** options were correctly processed
2. I then switched the 4 scripts that used **-M** to **-C** and ran the tests again
3. Since the new **-C** has to also handle the syntax of **-M** it can now handle the old **-C** syntax plus **-C**_xmargin_/_ymargin_ and **-C**_wmargin/emargin/smargin/nmargin_  which is even more flexible than either previous option.
4. It also turned out that inset had been left out of the synopsis reformatting (to use terminal width) so I fixed that in the process.

All tests still pass as before, the documentation has been updated.  Once this PR is approved and merged I will enhance the **-C** option in **subplot** to be equally flexible.


seisman (approved):
Looks good to me.

PR #6302, opened 2022-02-04
Improve sample1d docs for spatial resampling

PaulWessel:
See this forum [post](https://forum.generic-mapping-tools.org/t/plot-a-and-output/2591/13) for background. There was no discussion in the sample1d documentation of using **-T**_inc_ to perform spatial resampling of geographic or Cartesian paths.


anbj (approved):
Looks good. I would expect a `-G` option to be used for this, like in `project`.

Issue #6306, opened 2022-02-05
triangulate -Qn kills Julia


joa-quim:
It looks that the warning is true after all
```
julia> triangulate(rand(5,2), R="0/1/0/1", voronoi=:pol, Vd=1)
        triangulate  -R0/1/0/1 -Qn -M -JX14c/9.5c
triangulate [WARNING]: -Qn is experimental and unstable.
```
Will try to debug to debug it latter.

joa-quim:
Found one, but not enough to fix all shits
```
/* Need temp array to hold the coordinates of a single polygon [I hope this is large enough - don't know yet] */
		xcoord = gmt_M_memory (GMT, NULL, n_int_edges/2, double);
		ycoord = gmt_M_memory (GMT, NULL, n_int_edges/2, double);
```
When the killing happened ``n_int_edges = 6`` but later down ``np = 5`` so ``x|ycoord`` overflowed. Fixing this by not dividing by 2, solved the kill. But then another one followed. The command just fall in an apparent infinite loop and I had to kill the process. In fact, sometimes it worked fine but most of them the hang happened. 
Can't debug this since each ``step`` takes > 20 secs.

PaulWessel:
OK, please make a short CLI test case for me so I can debug when I can find some time (doing taxes today).

joa-quim:
No CLI. It's from externals
```
triangulate(rand(5,2), R="0/1/0/1", voronoi=:pol, Vd=1)
```

Can we tax it?

PaulWessel:
It is too taxing.

PaulWessel:
Well, using rand(5,2) is not helpful since each time one starts up again we get a new data set, so the particular path through the mess in _gmtsupport_voronoi_shewchuk_ changes... Can you make this into a script that sets a fixed 5x2 data set that still uses the crash so it is repeatable exactly the same each time, please?

joa-quim:
This one crashes on ``cmd`` as well
```
triangulate xy.dat  -R-1/2/-1/2 -Qn -M -JX14c/9.5c
```
where xy.dat is
0.972426499088	0.947716871068
0.755367493002	0.798499728751
0.730237556614	0.910008572993
0.833957090001	0.951488813143
0.827406268506	0.949740549484

The problem is with the sizes of ``x|ycoord`` in gmt_support.c #L3803 that are too short.

PaulWessel:
[...]

PR #6310, opened 2022-02-07
Increase the memory allocation for voronoi polygon coordinates

PaulWessel:
See #6306 for background.  It is hard to follow what I wrote some years ago regarding building these polygons, but hopefully this change will prevent us from running out of memory - after all these are small polygons.  After all, it is not entirely reassuring when I wrote the comment

`/* Need temp array to hold the coordinates of a single polygon [I hope this is large enough - don't know yet] */`

With this the example runs.  I also added an assert statement so that we are notified if this is still not enough.

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -3865,6 +3865,7 @@ GMT_LOCAL struct GMT_DATASET * gmtsupport_voronoi_shewchuk (struct GMT_CTRL *GMT
 						prev_sign = next_sign;
 						i2 = j2;
 						/* Add next_point to our polygon */
+						assert (np < n_int_edges);	/* Otherwise we want to know */
 						xcoord[np] = x_pos[2*next_point];
 						ycoord[np] = y_pos[2*next_point];
 						if (edge_use[j])	/* Now used twice */
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -3800,8 +3800,8 @@ GMT_LOCAL struct GMT_DATASET * gmtsupport_voronoi_shewchuk (struct GMT_CTRL *GMT
 			if (i >= n_int_edges) edge_use[i] = 1;	/* These edges will only be used once (sign is not important here) */
 		}
 		/* Need temp array to hold the coordinates of a single polygon [I hope this is large enough - don't know yet] */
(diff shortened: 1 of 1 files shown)


joa-quim (commented):
This change is not enough. The crash I got with this issue's example happened with this change that I already tried.

PaulWessel:
Could you then please try 2* n_int_edges?  I mean the most complicated polygon cannot have too many vertices...

joa-quim:
With the *2 it works fine. But one strange thing is that when I opened the issue the most common cause of error was that it hanged with the processor at 100%, not a crash. Crashes are easy to spot, those hangings not. But yesterday and today I was not able to make it happen again.

PaulWessel:
Hm, yes I had to let it run a while in debug before it got to a stop point (yesterday).  But it was just slow in debug - it did get there.  But it is possible that the memory runover could have caused other issues related to the hanging.  let me know if you are OK to approved this - I rather not try to understand this part of the code.

joa-quim (approved):

joa-quim:
I'll make a (richer) Matlab copy of triplot for you to play with.

src/gmt_support.c
```diff
@@ -3800,8 +3800,8 @@ GMT_LOCAL struct GMT_DATASET * gmtsupport_voronoi_shewchuk (struct GMT_CTRL *GMT
 			if (i >= n_int_edges) edge_use[i] = 1;	/* These edges will only be used once (sign is not important here) */
 		}
 		/* Need temp array to hold the coordinates of a single polygon [I hope this is large enough - don't know yet] */
-		xcoord = gmt_M_memory (GMT, NULL, n_int_edges/2, double);
-		ycoord = gmt_M_memory (GMT, NULL, n_int_edges/2, double);
+		xcoord = gmt_M_memory (GMT, NULL, n_int_edges, double);
```
meghanrjones, line 3803: Doesn't this need to be 2*n_int_edges or some other value >n_int_edges for the crash to be resolved?

PaulWessel:
Oops, sorry I thought @joa-quim had added 2 but I guess not committed.  OK will do this sneakily in master.

PaulWessel:
Done

Issue #6308, opened 2022-02-07
pslegend crashes when legend width is not provided.


joa-quim:
See https://github.com/GenericMappingTools/GMT.jl/issues/827

In fact this woks
```
legend(region=(0,10,0,8), (symbol1=(dx_left=0.25, marker="-", size=0.33, pen=2, fill=:black, dx_right=0.75, text="1s delay time"), symbol2=(dx_left=0.25, marker=:star, size=0.5, pen=:false, fill=:yellow, dx_right=0.75, text="Epicentre")), position=(inside=:BL, anchor=:BL, width=5), box=(; pen=0.5), Vd=1, show=1)
        pslegend  -R0/10/0/8 -JX12c/0 -Baf -BWSen -F+p0.5 -DjBL+w5 -P -K > C:\TEMP\GMTjl_tmp.ps
```
but crashes if I remove the ``width=5``. On an attentive look ``pslegend's`` man page says ``width`` is mandatory but the code should check that because as is
```
if (Ctrl->D.dim[GMT_X] == 0.0) {	/* Compute legend width */
		legend_item = gmt_M_memory (GMT, NULL, n_records, struct PSLEGEND_TXT);	/* Array to hold all labels */
		do_width = true;
	}
```
but later the ``legend_item`` is accessed without checking.

Funny, with a  debug build the above working code errors with
```
legend(region=(0,10,0,8), (symbol1=(dx_left=0.25, marker="-", size=0.33, pen=2, fill=:black, dx_right=0.75, text="1s delay time"), symbol2=(dx_left=0.25, marker=:star, size=0.5, pen=:false, fill=:yellow, dx_right=0.75, text="Epicentre")), position=(inside=:BL, anchor=:BL, width=10), box=(; pen=0.5), Vd=1)
        pslegend  -R0/10/0/8 -JX12c/0 -Baf -BWSen -F+p0.5 -DjBL+w10 -P -K > C:\TEMP\GMTjl_tmp.ps
pslegend [ERROR]: Option -B parsing failure. Correct syntax:

...
pslegend [ERROR]: Offending option -BWSen
```

PaulWessel:
The man page probably needs some fixing.  +w is optional in modern mode as we compute the exact width in PostScript.  It may possibly also work in classic - not sure.  It certainly happens when -l is used.

joa-quim:
Then, it must be a externals issue because no +w ==> crash

PaulWessel:
[...]

PR #6311, opened 2022-02-07
Clarify when legend width must be specified

PaulWessel:
See #6308 for some background. Turns out there are a few requirements and special cases, now documented by this PR, but for simple legends the width is optional for both classic and modern mode.


joa-quim (approved):
Right. But not clear if the Julia crashes fall outside these usability options. And anyway, crashes should never happen.

PR #6313, opened 2022-02-08
Let GMT verify counters in external datasets

PaulWessel:
As a sanity check, this PR lets GMT validate the counters found in a dataset passed to a module via reference or duplicate.  Mismatches between stated and actual records, segments and columns will be noted and an error is returned.

In the process of testing I found **pslegend** did not initialize _n_segments_ to zero in the dataset and thus got caught by the verifier, so I guess it works.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -10080,6 +10080,7 @@ GMT_LOCAL void gmtapi_get_record_init (struct GMTAPI_CTRL *API) {
 		case GMT_IS_DUPLICATE:	/* Only for datasets */
 		case GMT_IS_REFERENCE:	/* Only for datasets */
 			API->current_get_D_set = gmtapi_get_dataset_data (S->resource);	/* Get the right dataset */
+			gmt_set_dataset_verify (GMT, API->current_get_D_set);    /* Basic sanity checking of incoming dataset */
 			API->current_get_n_columns = (GMT->common.i.select) ? GMT->common.i.n_cols : API->current_get_D_set->n_columns;
 			API->api_get_record = gmtapi_get_record_dataset;
 			if (!(API->current_get_D_set->type & GMT_READ_TEXT)) GMT->current.io.record.text = NULL;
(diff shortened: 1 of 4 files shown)


joa-quim:
Turns out that I was setting the n_records but apparently no pushing it back to the pointer where it belongs. Also, due to heritage from the old GMT_TEXTSEGMENT I was allowing GMTdatasets with no coordinates (empty matrix) and that might have caused some shits as well. Think I've fixed both of these but since this sanity check is quite implacable it would perhaps be better to turn it into warnings. We need also to see its effect on the MEX.

PaulWessel:
What is the argument for warnings when there are errors in the input data set? Surely the developer would just want to fix that upstream first?  This is not user issues, right?

joa-quim:
The argument is that the lack setting those is not fatal as is proved by the fact that Julia has worked up to now without the ``n_records``. It was the rare situation of passing in a dataset with no coordinates, or better an empty matrix that was translated don't know how, that triggered this issue. So, if it still works and the API didn't start yesterday we better be cautious in not breaking existing codes.

PaulWessel:
OK, or we could simply update those n_records and n_segments to the correct information - no possible reason why setting these to the correct values in external memory is wrong, no?  We are not reallocating anything, just setting variables to correct numbers.

joa-quim:
Yes, fixing when not set is even better.

PaulWessel:
Done, give it a spin - I think this should be good enough.

joa-quim (approved):
There is not much for me to test since I, presumably, closed the door for in in Julia.
However, and I forgot to mention it before, when I tested it earlier running ``gmtinfo(GMTdataset)`` did not land in this function.

PaulWessel:
Thanks, yes when the rec-by-rec modules are passed a memory object they assign a pointer to it internally and now I intervene here too with the verifier.

joa-quim:
This makes me think. From externals there is never a need to rec-by-rec.

PaulWessel:
Not likely, but some gmt modules are still reading that way, hence the complication of GMT_Get_Record (and GMT_Put_Record) when dealing with externals.

PaulWessel:
@meghanrjones is the verbal failure anything to worry about?

meghanrjones:
> @meghanrjones is the verbal failure anything to worry about?

No, it is unrelated and can be safely ignored.

PR #6322, opened 2022-02-09
Only applied fix in one of two places...

PaulWessel:
Forgot there were two near-identical functions...
This applies the same fix in _gmt_bcr_get_z_ as was done in _gmt_bcr_get_z_fast_ earlier...


PaulWessel:
Now all three places treats the node range similarly.

meghanrjones (approved):

PR #6324, opened 2022-02-09
Separate out the supplement tests

meghanrjones:

The supplement tests are on average slower than the core module tests and take up ~15% of the total test time. This PR makes the supplement tests optional based on a new CMake setting DO_SUPPLEMENT_TESTS that is documented in ConfigUserAdvancedTemplate.cmake and the contributing guide.

The default after this PR is that the supplement tests will not be run, which is the same as the default for the API tests. However, I have temporarily added DO_SUPPLEMENT_TESTS to the CI configuration pending further discussion about how frequently these tests should be run.

Of course, there are other steps that we can take to reduce the time cost of tests which can be discussed in separate issues.

Currently WIP in order to research whether there are better ways to accomplish this.

Relates to https://github.com/GenericMappingTools/gmt/issues/53

**Reminders**


PaulWessel (approved):
I think this is a good idea, as long as we do run them from time to time (I may turn it on locally of course)

PR #6327, opened 2022-02-09
Improve option checks and docs for grdgradient

meghanrjones:

This PR fixes a docs mixup in which the direction (-D) and magnitude (-S) were switched.

### DIFF
--- a/doc/rst/source/grdgradient.rst
+++ b/doc/rst/source/grdgradient.rst
@@ -64,7 +64,7 @@ Optional Arguments
     when the slope of :math:`z(x,y)` is downhill in the :math:`a` direction, the
     correct sense for shading the illumination of an image (see
     :doc:`grdimage` and :doc:`grdview`) by a light source above the x,y plane
-    shining from the:math:`a` direction. Optionally, supply two azimuths,
+    shining from the :math:`a` direction. Optionally, supply two azimuths,
     **-A**\ *azim*/*azim2*, in which case the gradients in each of these
     directions are calculated and the one larger in magnitude is
     retained; this is useful for illuminating data with two directions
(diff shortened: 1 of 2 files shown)


meghanrjones:
I'm opening this as a PR to have a discussion related to this section - the docs say that grdgradient computes the directional derivative or the magnitude/direction of the vector gradient. But there doesn't seem to be anything in the code that makes these options (-A versus -D|-S) mutually exclusive (e.g., https://github.com/GenericMappingTools/pygmt/pull/1679). Is this intentional?

PaulWessel (approved):

meghanrjones:
> I'm opening this as a PR to have a discussion related to this section - the docs say that grdgradient computes the directional derivative or the magnitude/direction of the vector gradient. But there doesn't seem to be anything in the code that makes these options (-A versus -D|-S) mutually exclusive (e.g., https://github.com/GenericMappingTools/pygmt/pull/1679). Is this intentional?

Sorry, I forgot to ping @PaulWessel on this comment for feedback.

PaulWessel:
I think **-A** and **-D** are mutually exclusive since both write to the **-G** output grid.  So you cannot get both the directional derivatives in the chosen direction (via **-A**) _and_ output gradient directions by asking for **-D** as well.  Likewise, **-S** can _only_ be used with **-D** to save the magnitude to the separate file.  So I think these changes are needed:

1. Parse needs to give error if both **-A** and **-D** are set
2. Parse needs to give error if **-S** is set but not **-D**
3. Documentation for **-D** should start with "Instead of finding the gradient in a given direction (-A), find ..."

Note that if **-S** is given and **-G** is not then no gradient direction grid is written.

PaulWessel:
Also, while at it.  I think the equations with words look bad.  I suggest you change this to use variable names.  Just introduce variables o (offset) and \sigma (sigma) [which you already have], amplitude (a), and then it will look more like real equations.

meghanrjones:
Sounds good. I'm going to convert this to a WIP and add the changes here rather than opening a different PR, since this already contains the relevant discussion.

PaulWessel (approved):
There is a :math: run-in in the -A section that is not set as math.
-Nt last equation, is the parenthesis around a needed?
Otherwise looks great!

PR #6334, opened 2022-02-13
Addressing Voronoi polygons in triangulate -Qn

PaulWessel:
This follows up on problem #6306.  Some context: We use Schewchuck's **triangle** library to solve for Delaunay triangulation.  One option is to get Voronoi polygons.  The library returns two sets of information:

1. List of point coordinates for known nodes on the Voronoi boundaries
2. List of edges given as point _i_ to point _j_ in the point coordinate list

In many cases, _j_ is -1 meaning an infinite ray starting at point _i_.  Module **triangulate** uses that information and the given **-R** to determine intersections with the boundary, thus replacing point _j_ with a new computed point on the boundary we add to the point list, and we also add the coordinates of the four corners of w/e/s/n to the list.  Finally, we assemble polygons via the box outline.  This seems all well and good in principle.

However, the problem that lead to some crashes (and why **-Qn** is labeled as experimental [or temperamental...]) seems to be any specification of w/e/s/n that are _smaller_ than the domain defined by the Voronoi points returned. Note that these may lay outside the range of the input data.  Having a start point for an infinite ray being _outside_ the w/e/s/n is no good.  In the end I found two bugs:

- Bug 1 was to simply skip Voronoi points outside the region but that completely misses numerous polygons.
- Bug 2 had to do with the ray intersection going wrong since it was assumed the starting point was inside w/e/s/n.

This PR hopefully fixes both bugs by (1) not skipping any of the Voronoi points and (2) adjusting the w/e/s/n outward as needed to contain all the critical points.  This seems to work well.  A future PR may come back and impose the original w/e/s/n and clip the polygons since some of their outer boundary will change.  Also, I plan to add some checks so that **triangulate** is not used on global data sets and regions (there is **sphtriangulate** for spherical Delaney and Voronoi construction).

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -5,7 +5,7 @@
  *
  *	This program is free software; you can redistribute it and/or modify
  *	it under the terms of the GNU Lesser General Public License as published by
- *	the Free Software Foundation; version 3 or any later version.f
+ *	the Free Software Foundation; version 3 or any later version.
  *
  *	This program is distributed in the hope that it will be useful,
  *	but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


joa-quim (approved):
Uf, much more complicated than I thought.
Tried several times with ``rand()`` and never crashed. But maybe remove or move it to a higher level of verbosity
```
triangulate [WARNING]: -Qn is experimental and unstable.
```

PaulWessel:
Do you think I should spend time adding the "clip polygons to original -R"? I can see doing this via functions we have for clipping polygons in gmt_map.c for plotting purposes...

PaulWessel:
I already added doc update in master that says triangulate is meant for Cartesian or small-area geographic that does not involve periodic data or poles.

joa-quim:
Don't know. I really don't know what are the use cases for these Voronoi polygons and if someone would really need them to be closed by the exact specified -R.

PR #6339, opened 2022-02-15
Fix the Visual Stuido path for Windows CI workflows

seisman:

GitHub Actions now uses Windows Server 2022 as the default runner, so the path for Visual Studio should also be updated.


meghanrjones (approved):
Thanks!

PR #6340, opened 2022-02-15
CI: graphicsmagick and ffmpeg are no longer needed for building the GMT documentation

seisman:

I think graphicsmagick and ffmpeg are no longer needed for building the GMT documentation after we upload all GMT animations to YouTube. 

This PR removes graphicsmagick and ffmpeg from the package list for building the GMT documentation.


PaulWessel (approved):
I agree, building the docs just need to convert PS to PNG.

Issue #6025, opened 2021-11-18
Bundle step fails on the macOS docs workflow


meghanrjones:
**Description of the problem**

The macOS docs workflow is consistently failing on the Package GMT step (e.g. https://github.com/GenericMappingTools/gmt/runs/4244772423?check_suite_focus=true) with the error:
```
  /Applications/Xcode_13.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool-classic:
  can't open file: @rpath/libgeos.3.10.1.dylib (No such file or directory)
```

So far, I have tried adding geos to [install-dependencies-macos.sh](https://github.com/GenericMappingTools/gmt/blob/master/ci/install-dependencies-macos.sh) and re-adding the `brew update` step before `brew install`, with no success.

@seisman, do you have any recommendations for fixing the workflow? Otherwise, I can switch from MacPorts to homebrew after the 6.3 release to test the setup locally.

seisman:
Sorry, I have no idea why it fails.

PR #1968, opened 2019-11-03
Avoid hardwired version for GM in startup script and bundle dir

PaulWessel:
GraphicsMagick has a support dir with version number in it (GraphicsMagick-13.33) which obviously will change in the future.  This PR avoids that by placing the files (from whatever version) into a version-free directory name.
Also let CMake set variables *_VERSION for gm, gs, and ffmpeg as these may come in handy.

PR #6344, opened 2022-02-16
Update build-release script for macports or homebrew

meghanrjones:

This WIP PR updates `admin/build-release.sh` to work for either macports or homebrew installations. Here are the updates:

- Separate the steps for creating `cache-mp-clang.cmake` into a new script `build-macos-cmake-cache.sh`
- Determine whether the package manager is homebrew or macports in `build-release.sh`, exit if neither are detected, and pass the distro to `build-macos-external-list.sh` and `build-macos-cmake-cache.sh`.
- Add another safeguard against accidentally deleting CMake config files by exiting if the backup files already exist.
- Add the option to test `build-release.sh` using `-t`, which skips the check for the `GMT_PUBLIC_RELEASE` setting.
- Use distribution dependent paths in `build-macos-cmake-cache.sh` and `build-macos-cmake-cache.sh`.
- Update `build-macos-external-list.sh` to not run `which` on the shared libraries.

A few issues:

1. When pushing this branch, I discovered an older attempt to accomplish this task - [build-release-brew](https://github.com/GenericMappingTools/gmt/tree/build-release-brew) which also seems to include some updates for the M1 chip. @PaulWessel, is there anything from there that we should cherry-pick into this branch?
2. `build-macos-external-list.sh` adds the graphicsmagick file `log.mgk` to `add_macOS_cpack.txt`, which only exists in macports distributions (xref https://github.com/GenericMappingTools/gmt/pull/1968). For now, the script only adds this file if using macports, but there may be other files required for the homebrew installation.
3. I am stuck on resolving  https://github.com/GenericMappingTools/gmt/issues/6025. In addition to homebrew updates, I tried adding the two geos libraries to `EXEPLUSLIBS` in `build-macos-external-list.sh` (didn't help) and building the bundle without GEOS using `set (GMT_EXCLUDE_GEOS TRUE)` (didn't help). Based on the fact that explicitly excluding GEOS as a GMT dependency did not change the problem, I wonder if it could be failing to find the shared library required by gdal and, if so, whether gdal's upcoming transition to CMake could help the problem. I'd appreciate any suggestions for other fixes to try.
4. I do not have a way to check whether this breaks anything for building the bundle with macports and will need @PaulWessel's help to test that.


PaulWessel (approved):
Seems good - did not complete due to signing the bundle on my laptop not set up.  I could not see anything g useful in the old branch so I will delete it.  We can rediscover anything related to M1 but I think those were early days.

meghanrjones:
> Seems good - did not complete due to signing the bundle on my laptop not set up. I could not see anything g useful in the old branch so I will delete it. We can rediscover anything related to M1 but I think those were early days.

I think we can do `if (($ENV{USER} STREQUAL "pwessel") OR ($ENV{USER} STREQUAL "meghanj") AND GMT_PUBLIC_RELEASE)` in admin/ConfigReleaseBuild.cmake so that when `build-release.sh -t` is used the code-signing is not attempted.

PaulWessel:
Sure, that might work.

meghanrjones:
> But I think in my case I just done have installed the certificates on my laptop. I should do that since I will be on the road for 6 month and we may need me to run that command...

OK, we can do both. While I have yet to successfully build the bundle with homebrew distributions, I found code-signing for testing to be a bit annoying so it would be good to have the option to skip that step.

PR #6360, opened 2022-02-21
Let colorbar plot rectangles if possible

### LINKED ISSUE #6350: Add an informational message to grdimage and colorbar regarding Apple Preview

Lots of users are falling into the trap that they make a PDF map of a coarse image, or request a discrete color bar, and when viewing the result on macOS with Preview the image looks either blurry (coarse **grdimage** plots) or the CPT looks continuous (for discrete CPTs).  These are decisions Apple make in Preview and we cannot fight that.  However, here are some things we can do:

1. The **grdimage** man page could have a Note: paragraph where it explains that coarse images viewed in PDF may look blurry depending on your PDF viewer and OS - this is a known issue on macOS.
2. The **colorbar** man page could do the same. But, note that **colorbar -Np** would draw a discrete color bar using _rectangles_ which would not be subject to the Apple defect.  I suggest we make **-Np** for discrete colormaps the default.  This adds just a tiny amount of more PS (draw rectangle versus a tiny image) but will avoid the most common complaint for this problems (discrete color bars).

Let me know what @GenericMappingTools/core think about this.

PaulWessel:
See #6350 for background.  Because Apple Preview (and possibly others) will resample coarse PDF images, any colormap based on a discrete CPT will look very blurred and continuous.  This PR lets -Np be the default if it is possible.
Closes #6350.  Note: A handfull of PostScript plot originals will need updating due to hairline differences between the approximate images and the exact polygon boundaries.


meghanrjones:
@PaulWessel, should I push the test image updates to this branch?

PaulWessel:
yes, please.  I looked at CONTRIBUTING and I probably missed it but I did not find a section on how to revise plots already in dvc?

meghanrjones:
>  I looked at CONTRIBUTING and I probably missed it but I did not find a section on how to revise plots already in dvc?

I will take a look and revise as needed in a separate PR.

meghanrjones (approved):

PR #6363, opened 2022-02-22
Ensure that the raster has x and y increasing normally

PaulWessel:
See discussion on the [forum](https://forum.generic-mapping-tools.org/t/malloc-error-with-grdview-using-a-draped-tiff-image/2644/10) for background.  Unless we put in efforts to handle odd cases, we now return with error if x or y are not increasing in the expected directions (left to right, bottom to top),

### DIFF
--- a/src/gmt_gdalread.c
+++ b/src/gmt_gdalread.c
@@ -426,6 +426,16 @@ GMT_LOCAL int populate_metadata (struct GMT_CTRL *GMT, struct GMT_GDALREAD_OUT_C
 		anSrcWin[2] = irint ((dfLRX - dfULX) / adfGeoTransform[1]);
 		anSrcWin[3] = irint ((dfLRY - dfULY) / adfGeoTransform[5]);
 
+		/* First check if we have a compliant raster (right handed coordinates) */
+		if (anSrcWin[2] < 0) {
+			GMT_Report (GMT->parent, GMT_MSG_ERROR, "Your raster has x-coordinates increasing to the left - exiting\n");
+			return(GMT_NOTSET);
+		}
+		if (anSrcWin[3] < 0) {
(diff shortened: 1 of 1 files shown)


joa-quim (commented):
Hmm, I don't like this. For example these tests are done only when a -R is set. The other case, read it all, will also be upside down but we'll let it go unnoticed.

PaulWessel:
Hm, yes, makes sense.  Will see what to do.

PaulWessel:
What should I check for? Does the adfGeoTransform reflect the upside down settings? Any other parameters we need to consult instead?

joa-quim:
Wait, I'll work on it.

joa-quim:
Managed to avoid the crashes with logic changes, but the UD flipping is resisting to the presumably smart attempts. Will continue after that sport that we practice by dinner time.

Wat happens at the base is that gdalread expects the y-axes to be **always  positive down** but in this case GDAL returns it a positive up ... and shit follows.

PaulWessel:
OK, will leave in your capable hands, especially after filling up with nutrients.

joa-quim:
Many debugs latter, both of this work
```
gmt grdimage drape.tif  -JX14c/0 -Baf -BWSen -n+a -R0/390.2/-49.825549/49.824651 -png lixo

gmt grdimage drape.tif  -JX14c/0 -Baf -BWSen -n+a -png lixo
```

PaulWessel:
Great, I can run all the tests in a bit but have a zoom happening first.

PaulWessel:
All the tests still pass so if you are happy with this then you can approve, etc.  Going exercises for an hour or so.

joa-quim:
Yes, I,m finished here till next trouble.

joa-quim (approved):

PR #6388, opened 2022-02-25
Fix the cmake packaging issue on macOS

### LINKED ISSUE #6025: Bundle step fails on the macOS docs workflow

seisman:

The workflow fails to make the macOS bundle, with the following error message:
```
/Applications/Xcode_13.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool-classic:
  can't open file: @rpath/libgeos.3.10.1.dylib (No such file or directory)
```
It's unclear why other libraries (like gdal) have abosulte paths, but the geos library has a relative path `@rpath/libgeos.3.10.1.dylib`. So the root cause of the error is still unclear to me.

A temporary workaround is passing extra RPATH (`/usr/local/lib` for homebrew) as the last parameter of the `fixup_bundle` function. 

Fixes #6025.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,7 +4,7 @@
 name: Docs
 
 on:
-  # pull_request:
+  pull_request:
   push:
     branches:
       - master
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,7 +4,7 @@
 name: Docs
 
 on:
-  pull_request:
+  # pull_request:
   push:
     branches:
       - master
--- a/cmake/dist/CMakeLists.txt
+++ b/cmake/dist/CMakeLists.txt
@@ -195,7 +195,11 @@ if (APPLE)
 		endfunction(gp_item_default_embedded_path_override)
 
 		include (BundleUtilities)
-		fixup_bundle(\"\${CMAKE_INSTALL_PREFIX}/${GMT_BINDIR}/gmt${CMAKE_EXECUTABLE_SUFFIX}\" \"\${CMAKE_INSTALL_PREFIX}/${GMT_LIBDIR}/gmt${CMAKE_EXECUTABLE_SUFFIX}/plugins/${GMT_SUPPL_LIB_NAME}.so\" \"\")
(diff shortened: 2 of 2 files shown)


seisman:
With the workaround in this PR, the workflow now can create macOS bundle correctly.

@meghanrjones Please check if it works for you (I assume that you're a Homebrew user)
@PaulWessel Please check if it breaks your building with macports.

meghanrjones:
Thanks for working on this. `admin/build-release.sh -t` ran without errors for me using this PR.

joa-quim:
Will the bundle with this build be usable for pygmt?

PaulWessel:
I think that is a separate PR that we need to work on for both Julia and PyGMT.

seisman:
@PaulWessel does this PR affects your bundle building with macports?

PaulWessel:
Not sure, will try soon

PaulWessel:
No errors on Mac ports. I have not actually tested the bundle yet, maybe on upcoming flight

PaulWessel:
Build works, and running gmt coast with the installed bundle works for me on M1 macOS, macports..

.github/workflows/docs.yml
```diff
@@ -4,7 +4,7 @@
 name: Docs
 
 on:
-  # pull_request:
+  pull_request:
```
```suggestion
  # pull_request:
```

Issue #6336, opened 2022-02-14
[WARNING]: gmt_set_dataset_verify: Data set has ...


joa-quim:
I have a couple of tests that still scream inside ``gmt_set_dataset_verify``, like

```
julia> T = text_record(["P", "T d = [0 0; 1 1; 2 1; 3 0.5; 2 0.25]"]);

julia> legend(T, R="-3/3/-3/3", J=:X12,  D="g-1.8/2.6+w12c+jTL", F="+p+ggray", Vd=1)
        pslegend  -R-3/3/-3/3 -JX12 -Baf -BWSen -F+p+ggray -Dg-1.8/2.6+w12c+jTL -P -K > C:\TEMP\GMTjl_tmp.ps
pstext [WARNING]: gmt_set_dataset_verify: Data set has 1 segments but header says it only has 0
```
This is not easy to track and seem to happen when one module calls other (have another one coming from ``psxy``).
Given that this bothers only me (and Julia users by drag) I think the least effort path for this is to move this warning to an upper level of warning.

gd-a:
Got this too : 
```
plot [WARNING]: gmt_set_dataset_verify: Data set has 2 data records but header says it only has 0
text [WARNING]: gmt_set_dataset_verify: Data set has 1 segments but header says it only has 0
```

PaulWessel:
Closing this one as fixed - reopen should you see anything odd.

joa-quim:
Not only it's not fixed as I'm getting a Julia crash

```
julia> T = text_record(["P", "T d = [0 0; 1 1; 2 1; 3 0.5; 2 0.25]"]);

julia> legend(T, R="-3/3/-3/3", J=:X12,  D="g-1.8/2.6+w12c+jTL", F="+p+ggray", Vd=1);
        pslegend  -R-3/3/-3/3 -JX12 -Baf -BWSen -F+p+ggray -Dg-1.8/2.6+w12c+jTL -P -K > C:\TEMP\GMTjl_tmp.ps
pstext [WARNING]: gmt_set_dataset_verify: Data set has 0 data records but header says it only has 1

Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x7ff836359b52 -- gmt_whole_earth at C:\programs\GMTdir\gmt_w64.dll (unknown line)
in expression starting at REPL[3]:1
[...]

PR #6403, opened 2022-03-01
Add test for passing text dataset to legend

PaulWessel:
Basically replicates the situation in #6336.

@joa-quim, might you be able to try this branch and build all and then run the` testapi_text_legend > junk.ps `command? It should crash as well since it is basically doing your case - but perhaps things are easier to see.

### DIFF
--- a/src/pslegend.c
+++ b/src/pslegend.c
@@ -2016,7 +2016,7 @@ EXTERN_MSC int GMT_pslegend (void *V_API, int mode, void *args) {
 			D[PAR]->table[0]->n_segments = D[PAR]->n_segments = n_para;
 			D[PAR]->table[0]->n_records += S[PAR]->n_rows;
 			D[PAR]->n_records = D[PAR]->table[0]->n_records;
-			S[PAR] = D[PAR]->table[0]->segment[n_para] = GMT_Alloc_Segment (GMT->parent, GMT_WITH_STRINGS, krow[PAR], 0U, NULL, S[PAR]);
+			S[PAR] = D[PAR]->table[0]->segment[n_para-1] = GMT_Alloc_Segment (GMT->parent, GMT_WITH_STRINGS, krow[PAR], 0U, NULL, S[PAR]);
 		}
 
 		/* Create option list, register D[PAR] as input source */
--- a/src/testapi_text_legend.c
+++ b/src/testapi_text_legend.c
@@ -1,0 +1,33 @@
+/* Testing the case of passing a virtual dataset for paragraph text to pslegend */
+
+#include "gmt.h"
+#include <string.h>
+
+int main () {
+	void *API = NULL;                 /* The API control structure */
+	struct GMT_DATASET *D = NULL;	  /* Structure to hold input dataset as vectors */
+	struct GMT_DATASEGMENT *S = NULL;	  /* Structure to hold input dataset as vectors */
+	char input[GMT_VF_LEN] = {""};     			/* String to hold virtual input filename */
+	char args[128] = {""};            			/* String to hold module command arguments */
+
+	uint64_t dim[4] = {1, 1, 2, 0};
+
+	/* Initialize the GMT session */
+	API = GMT_Create_Session ("test", 2U, GMT_SESSION_EXTERNAL, NULL);
+	/* Create a dataset */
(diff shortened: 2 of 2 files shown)


joa-quim:
Can't right now. Poor computer is melting with all cores working.

meghanrjones:
Can you provide more guidance about how I can help as a reviewer on this PR? On windows, I added -Vd to the module arguments, added `testapi_text_legend` to gmtest.in, and added a shell script in `test/api` to confirm that it crashes. The only suspicious message in the logs is `GMT_Destroy_data: Skipped due to wrong level for object 1`. The last two messages are:
```
VirtualFile name created: <virtualfile>
RUNNING: PAR: gmt pstext -R0/12/0/12 -Jx1i -O -K -N -M -F+a+f+j <virtualfile> --GMT_HISTORY=readonly
```
I could try setting up debugging for this case on Windows, but I am not sure if I would get further than Joaquim did.

PaulWessel:
Maybe we could learn where it crashes?

meghanrjones:
I am still working on getting my windows setup working again, but will let you know when I get more information about the crash.

joa-quim:
But I said where it crashes. When freeing a corrupted memory.

PaulWessel:
Yes you did, sorry.  So looking at what that commit did that is related to the dataset D[PAR], it is this only:

`+                       D[PAR]->table[0]->n_segments = n_para;`

So before n_segments (probably) was not set but now it is, and that is why 

`for (seg = 0; seg < table->n_segments; seg++) gmt_free_segment (GMT, &(table->segment[seg]));`

is now actually calling that statement while before the loop would not execute.  I will see if I can learn anything in Xcode even though it does not crash.

PaulWessel:
It gets to that point twice:

1. First it cleans up the input to **pslegend**.  That is the dataset you create in Julia via your _text_record_, maybe using _GMT_Create_Data_? I cannot see that part.  But it says all the memory is allocated internally by GMT hence it frees things as we do with all GMT-created memory.  Now, if this memory was _not_ created by GMT then I can understand the crash, but then you need to set the alloc_modes to **GMT_ALLOC_EXTERNALLY**, etc.
2. The second time it is to free the dataset created in **pslegend** that is passed to **pstext -M**.  I see nothing unusual there (and since nothing crashes I cannot tell anyway).

Perhaps you are able to tell me which of these two places is crashing, i.e., trying to free the **pslegend** input or the **pstext** input?

joa-quim:
1. Could not be. It's the same function that reads all datasets

The problem is that when it goes to line 8704 it tries to free a NULL memory.
image

joa-quim:
But even if I add

if (table->segment != NULL) gmt_M_free (GMT, table->segment);

it still crashes with the corrupted heap message so, some bad allocation happened before.

PaulWessel:
If I understand you there are other places where you create such program text strings and that works fine (for pstext I assume but not in pslegend).  So you think it is the internally generated D[PAR] dataset that fails upon freeing, yes?

PaulWessel:
Yes, no need to check if something is non-NULL or NULL since gmt_M_free checks and does not try to free a NULL pointer.

joa-quim:
This was with the testapi_text_legend.exe. No Julia.

PaulWessel:
OK. No julia, but still need to pin you down that when you see the crashing free you are witnesses the destruction of the dataset that was created in pslegend (D[PAR]) and passed into pstext? Yes or no.

joa-quim:
I think I found it. In ``pslegend``
```
	if (D[PAR]) {
		n_para++;
		if (n_para >= 0) {	/* End of last paragraph for sure */
			S[PAR]->n_rows = krow[PAR];
			D[PAR]->table[0]->n_segments = D[PAR]->n_segments = n_para;
			D[PAR]->table[0]->n_records += S[PAR]->n_rows;
			D[PAR]->n_records = D[PAR]->table[0]->n_records;
			S[PAR] = D[PAR]->table[0]->segment[n_para] = GMT_Alloc_Segment (GMT->parent, GMT_WITH_STRINGS, krow[PAR], 0U, NULL, S[PAR]);
		}
```
The last line allocates ``segment[n_para]`` which is already ``segment[1]`` so ``segment[0]`` is never allocated. Changing it to ``segment[n_para-1]`` and the tests runs fine.

PaulWessel:
Yes, that must be it.  I will make a PR (will reorder a few things).

PaulWessel:
Actually, might as well do exactly as you suggested.  Give this a spin and approve if no crash under Windows, @joa-quim

joa-quim (approved):
No more crashes in Julia and the silly test works again.

PR #6410, opened 2022-03-05
Allow grdfill to output unchanged grid if no holes

### LINKED ISSUE #6408: Should grdfill produce an output grid even if no holes were filled?

Currently, grdfill only produces an output grid with -G if holes were found in the grid. If no holes were found, a warning is instead provided - `[WARNING]: No holes detected in grid - grid was not updated`. I think this is inconsistent with the behavior of other gmt modules. For example, `grdclip -Sa<high>/<above> <ingrid> -G<outgrid>` will still produce `<outgrid>` even if `<high>` is greater than the maximum z-value in the grid. Is there a reason that grdfill is different in this regards?

PaulWessel:
It was a mistake to refuse this since workflows may not know if there are holes or not, in which case the unchanged grid should be written.  Closes #6408.

### DIFF
--- a/doc/rst/source/grdfill.rst
+++ b/doc/rst/source/grdfill.rst
@@ -30,7 +30,8 @@ Description
 **grdfill** reads a grid that presumably has unfilled holes that the user
 wants to fill in some fashion.  Holes are identified by NaN values but
 this criteria can be changed via **-N**.  There are several different algorithms that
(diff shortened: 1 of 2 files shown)


meghanrjones (approved):

meghanrjones:
I merged this right after approval because it will help with another PR. Thanks for the quick fix.

PR #6418, opened 2022-03-07
Pass GMT_GRID_NEEDS_PAD1 flag to GMT_Read_Data in grdfill

meghanrjones:

This PR fixes the bug reported by https://github.com/GenericMappingTools/pygmt/issues/1745. This crash occurred in cases where `Grid->header->pad[YHI]`> `Grid->header->pad[YLO]`.

Fixes #

**Reminders**


src/grdfill.c
```diff
@@ -599,8 +599,8 @@ EXTERN_MSC int GMT_grdfill (void *V_API, int mode, void *args) {
 	ID = gmt_M_memory_aligned (GMT, NULL, Grid->header->size, char);
 	/* Set the top and bottom boundary rows to UINT_MAX */
 	offset = (uint64_t)(Grid->header->pad[YHI] + Grid->header->n_rows) * Grid->header->mx;
-	for (node = 0; node < (uint64_t)Grid->header->pad[YHI]*Grid->header->mx; node++) ID[node] = ID[node+offset] = 1;
```
meghanrjones, line 602: @PaulWessel, is it necessary to set all of the boundary rows to 1, rather than just the outside most rows/columns?

src/testapi_matrix_grdfill.c
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+int main () {
+	unsigned int mode = GMT_SESSION_EXTERNAL;
+	struct GMT_MATRIX *M_g = NULL;
+	char input_g[GMT_VF_LEN] = {""};
+	char args_g[128]= {""};
+	struct GMTAPI_CTRL *API = NULL;
+
+	API = GMT_Create_Session ("test", 2U, mode, NULL);
```
meghanrjones, line 13: This regression test doesn't actually reproduce the original crash because the pad for this example is still `[2, 2, 2, 2]`. The pad for the failing PyGMT example is `[1, 3, 1, 6]`. @PaulWessel I know `GMT_Create_Session` can be used to change the default pad from `2`; is there a way to make the pad asymmetric?

PaulWessel: Not via Create_Session but you could run
`
`	gmt_M_grd_setpad (GMT, header, pad);	/* Copy the pad to the header */

where pad is a 4-array and those numbers can be anything.  Then, the header->pad is set and then you must do 
`	gmt_set_grddim (GMT, header);		/* Update all dimensions */`
to have all dimensions updated.  Once that is done you the ->size is set and you can allocate array.
Those calls require gmt_dev.h though.

src/grdfill.c
PaulWessel: Not sure, will look to see why.

PaulWessel: Remember to change the **UINT_MAX** to 1 since I clearly changed my mind mid-coding and now use a char array.

Since we are only expanding (in grdfill_trace_the_hole) in the 4 cardinal directions it is hard to see why we would need more than one boundary of ones.  So you could change it only do the pad[?]-1 row and col inner pad and check.

meghanrjones: OK, I will update this for both the top/bottom and left/right boundaries.

Would using 1 be an issue in the odd case that someone uses grdfill -N1?

PaulWessel: No, the ID array is just used for keeping track of the NaNs, The -N affects what we look for, but you will see the code then simply replace grid values equal what -N says wiht NaNs since that is how the algo works,

src/testapi_matrix_grdfill.c
meghanrjones: I tried that workflow, but the pad gets reset again before reaching the problematic section of grdfill. It may be necessary to add a regression test in the PyGMT repository instead.

meghanrjones: I pushed the changes to the test in https://github.com/GenericMappingTools/gmt/pull/6418/commits/271b051a64d258eb7ec24ab42db6f7aefa911f51. Do you mind taking a look to see if I am missing anything?

PaulWessel:
I am not sure this will work.  If I understand, you wish to pass in a grid that then should have these odd pads. Once you read it in the array is allocated.  I think you have to doctor this by

1. set the pad as you do
2. Modify the header->wesn to reflect that (ie., extend region if pad < 2 and shrink if > 2.  Then call those two functions.
3. Once that is done you call the virtual file
You should be able to verify by looking at the header and sizes etc that the grid/matrix has the properties you want before doing the module call.

Sorry, bedtime for Bonzo here now.

meghanrjones:
I revised this PR to use a simpler fix by requiring a pad of at least 1 for grdfill. I also removed the C API test because the pad kept getting set as 2 regardless of adjustments to the order of setting the pad, adjusting the header, and calling the virtual file. I will add a PyGMT regression test instead.

src/grdfill.c
```diff
@@ -597,10 +597,10 @@ EXTERN_MSC int GMT_grdfill (void *V_API, int mode, void *args) {
 	 * the boundary row/cols in the ID grid to 1. */
 
 	ID = gmt_M_memory_aligned (GMT, NULL, Grid->header->size, char);
-	/* Set the top and bottom boundary rows to UINT_MAX */
+	/* Set the top and bottom boundary rows to 1 */
 	offset = (uint64_t)(Grid->header->pad[YHI] + Grid->header->n_rows) * Grid->header->mx;
-	for (node = 0; node < (openmp_int)Grid->header->mx; node++) ID[node] = ID[node+offset] = 1;
-	/* Set the left and right boundary columns to UINT_MAX */
+	for (node = 0; node < (uint64_t)Grid->header->pad[YHI]*Grid->header->mx; node++) ID[node] = ID[node+offset] = 1;
```
joa-quim, line 602: This bit makes me some tickles. Just wondering. What if the grid received was written top-down (netCDF grids can be either bottom-up or top-down). Shouldn't it then use ``pad[YLO]``?
Ofc, this may have no sense but seeing something depending on the axis direction when we can have both directions makes my bug bell to ring.

PaulWessel: I think the top/bottom order is dealt with in gmt_nc when the file is read.  By the time GMT modules get grids they are in the expected orientation.

meghanrjones: For clarification (just in case), this commit is reverting my changes back to the original code. I could go back to modifying the code to only set one boundary column/row on each side to 1.

PR #6427, opened 2022-03-10
Let one-liners with CPTs add colorbar

### LINKED ISSUE #6426: Let one-liners involving a CPT display colorbar

The main use of [one-liners](https://docs.generic-mapping-tools.org/latest/cookbook/one-liner.html) is to quickly make a plot of a data set, such as a grid.  It is not meant for production of final figures (unless this very simple plot is good enough).  However, plotting a grid via **grdimage** with a master CPT or default will scale the colors to the range of the grid, hence it is unclear what the colors represent unless one

1. Creates a know-range CPT via **makecpt**/**grd2cpt** and use that in **grdimage**
2. Abandons the one-liner in order to place a **colorbar** in a modern mode script.

I suggest that when a one-liner involves a current CPT (either implicitly or explicitly) we add a

`gmt colorbar -DJBC -B
`
at the end of the one-liner call so that it is possible to make sense of the plot.  Anyone object to this backwards-incompatible upgrade/"bug fix"? @GenericMappingTools/core ?

PaulWessel:
Without the colorbar, a colored image has little meaning if we cannot tell what values they represent.
This PR checks if the plotting module involves a CPT, and if so we place it on top via a `colorbar -DJTC -Baf` call.

Example: `gmt grdimage @earth_relief_02m -RFR -B -png map`

map

Closes #6426.

### DIFF
--- a/doc/rst/source/cookbook/one-liner.rst
+++ b/doc/rst/source/cookbook/one-liner.rst
@@ -25,6 +25,9 @@ The syntax for this special mechanism involves a few simple rules:
    leading hyphen for the format option.
 #. The one-liner command cannot itself be part of a longer modern mode session.
 
+**Note**: If the single module involves a CPT (either explicitly or implicitly) then a
+color bar will be plotted above the map.
+
 Examples
 --------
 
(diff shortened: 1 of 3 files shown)


joa-quim (approved):

PaulWessel:
I think because the default -B is -BWsne, but I am open to discussion.

Esteban82:
Ok. The issue is that if I run something like this I get a colorbar below. So, I think we should use the same default position of `gmt colorbar`.

```
gmt begin ...
	gmt grdimage $grd ...
	gmt colorbar 
gmt end
```

Issue #6421, opened 2022-03-08
Perspective projection (-Jg -JG) with +zr[n]r crashes


anbj:
```
$ gmt grdimage @earth_relief_10m -RNO -png w60 -JG10/60/15c+z10r
ERROR: Caught signal number 11 (Segmentation fault) at
/usr/local/lib/libgmt.so.6(GMT_grdimage+0x5a02)[0x7efcdb4e4f22]
[0x0]
Stack backtrace:
/usr/local/lib/libgmt.so.6(sig_handler_unix+0xc0)[0x7efcdb360c60]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x14140)[0x7efcdb1d8140]
/usr/local/lib/libgmt.so.6(GMT_grdimage+0x5a02)[0x7efcdb4e4f22]
/usr/local/lib/libgmt.so.6(GMT_Call_Module+0x38e)[0x7efcdb27024e]
gmt(main+0x546)[0x557d5492c736]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xea)[0x7efcdb025d0a]
gmt(_start+0x2a)[0x557d5492d5aa]
```

anbj:
Does not crash anymore. Fixed in #6429

PR #6429, opened 2022-03-10
Check that altitude is not inside Earth

PaulWessel:
See #6421 for background.  This PR cleans up a few things related to the general perspective projection, such as using our unit macros.  For instance, users cannot give **+z**_dist_**r** with _dist_ less than the radius of Earth, for instance, so that is now checked for rather than a crash.


joa-quim (approved):
Not completely clear why we don't allow *cave projections* but OK, maybe later.

PaulWessel:
Caves are scary.

meghanrjones (approved):

PR #6432, opened 2022-03-11
remote-data.rst: move section about available remote data sets to top

anbj:
The section about available remote data sets should be near the top. It's usually one of the first things a user will investigate.


meghanrjones (approved):
Good idea, thanks!

PR #6438, opened 2022-03-12
Document default Delaunay Triangulation algorithm as Shewchuk

### LINKED ISSUE #6436: Clarifying algorithm used for triangulate - Watson or Shewchuk?

Two things that popped up while trying to wrap `triangulate` in PyGMT, see https://github.com/GenericMappingTools/pygmt/pull/731#discussion_r825325759 for context:

1) The documentation at https://github.com/GenericMappingTools/gmt/blame/6.3.0/doc/rst/source/triangulate.rst#L60-L61 says that the triangulation algorithm used can be determined using `gmt triangulate -`, but that only prints the short help docs. @meghanrjones found that it should be `gmt get GMT_TRIANGULATE` instead
2) Just want to confirm that the Watson [1982] algorithm is indeed the default, and not the newer Shewchuk [1996] algorithm? The conda-forge `gmt` package seems to be using Shewchuk [1996], but I don't see any special flag set in the build recipe at https://github.com/conda-forge/gmt-feedstock/blob/def03480bc8c569e58d94359f8a835016f292883/recipe/build.sh#L1-L51. This may have licensing implications, since the Shewchuk [1996] algorithm is not for commercial use according to https://github.com/GenericMappingTools/gmt/blob/6.3.0/src/TRIANGLE.HOWTO.

```
# PASTE CODE HERE
```

**Full error message**

```
# PASTE ERROR MESSAGE HERE
```

**Actual outcome**

**Expected outcome**

**System information**

* Operating system:
* GMT version (`gmt --version`):

weiji14:

Current default should be the Shewchuk [1996] algorithm, and not Watson [1982]. Modified documentation in contour.rst, triangulate.rst, gmt.conf.rst and TRIANGLE.HOWTO to be clear on this. Also mentioned using `gmt get GMT_TRIANGULATE` as the way to know which one is being used.

Fixes #6436

**Reminders**

### DIFF
--- a/src/TRIANGLE.HOWTO
+++ b/src/TRIANGLE.HOWTO
@@ -1,9 +1,9 @@
 
 GMT contains two programs (triangulate and pscontour) that
 directly or indirectly will do a Delaunay optimal triangulation
-of arbitrarily spaced (x,y,z) data.  By default, GMT uses
+of arbitrarily spaced (x,y,z) data.  In the early days, GMT used
 a public domain (published) algorithm by Dave Watson [1982] to
-to this.  Now, a new and much faster algorithm has been released
+to this.  The default now though, is a new and much faster algorithm
 by Jonathan Shewchuk [1996].  Because Shewchuk holds the copyright
 (i.e., it is not in the public domain nor distributed under the GNU
 Public License or the equivalent) you are advised to read it (see
@@ -15,12 +15,12 @@ If you later need to change the behavior w.r.t. triangulation,
 follow these guidelines:
 
 In makegmt.macro there are two parameters that can be defined:
-TRIANGLE_D and TRIANGLE_O.  By default they are commented out.
+TRIANGLE_D and TRIANGLE_O.  By default these lines are commented out.
 To use Shewchuk's [1996] routine, remove the # from the
 beginning of the lines and then 'make clean' and 'make install'.
-To revert to the default GMT usage of Watson's [1982] algorithm,
+To revert to using Watson's [1982] algorithm,
 simply place a # in the first column at these two lines.
-Then, say 'make clean' and 'make install' again.
+Then, run 'make clean' and 'make install' again.
 
 Paul Wessel, 16-AUG-1999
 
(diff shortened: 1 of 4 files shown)


src/TRIANGLE.HOWTO
```diff
@@ -15,12 +15,12 @@ If you later need to change the behavior w.r.t. triangulation,
...
 In makegmt.macro there are two parameters that can be defined:
-TRIANGLE_D and TRIANGLE_O.  By default they are commented out.
+TRIANGLE_D and TRIANGLE_O.  By default these lines are commented out.
 To use Shewchuk's [1996] routine, remove the # from the
 beginning of the lines and then 'make clean' and 'make install'.
-To revert to the default GMT usage of Watson's [1982] algorithm,
+To revert to using Watson's [1982] algorithm,
 simply place a # in the first column at these two lines.
-Then, say 'make clean' and 'make install' again.
+Then, run 'make clean' and 'make install' again.
```
weiji14, line 23: I need a bit of help rewording these lines. I don't see `makegmt.macro` anywhere (the `git blame` on this file dates back to Dec 2000!!). I do see a `TRIANGLE_D` here:

https://github.com/GenericMappingTools/gmt/blob/434ed1848a1e8886931694b2de4c352e4c6cddae/src/gmt_init.c#L6580-L6585

but no mention of `TRIANGLE_O`.

meghanrjones: My best guess is that these macros are in reference to the pre-CMake days. I think that this paragraph and the preceding to sentences should instead make these points:
1. Use `gmt set GMT_TRIANGULATE` to control which algorithm is used on the fly.
2. Use `set (LICENSE_RESTRICTED GPL)` in `cmake/ConfigUserAdvanced.cmake` to disable Shewchuk's routine entirely.

PaulWessel: Yes, pre-CMake language.

PaulWessel:
OK, but not sure how to check out your branch...

meghanrjones:
Here are the instructions for checking out a PR - https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally. 

That said, I am not sure whether you'd be able to push the Wei Ji's branch (I haven't ever tried to push to a fork from a local checkout).

This is an improvement, so IMO you could merge this as a part 1 and make any additional edits as a separate PR if it's easier.

PR #6439, opened 2022-03-13
cmake: Do not call try_run() if in cross compiling mode

seisman:

The conda-forge is doing the ARM OSX migration in https://github.com/conda-forge/gmt-feedstock/pull/199, but the GMT source code fails to build with the [following error messages](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=473276&view=logs&jobId=ced5d8de-8227-5f3f-33a1-45cf1592c45a&j=ced5d8de-8227-5f3f-33a1-45cf1592c45a&t=f7bcb349-6dca-53ba-d88d-1d9963e119ba):
```
CMake Error: TRY_RUN() invoked in cross-compiling mode, please set the following cache variables appropriately:
   _EXIT_GSHHG_VERSION (advanced)
   _EXIT_GSHHG_VERSION__TRYRUN_OUTPUT (advanced)
For details see /Users/runner/miniforge3/conda-bld/gmt_1647064808556/work/build/TryRunResults.cmake
CMake Error at cmake/modules/FindGSHHG.cmake:77 (string):
  string sub-command STRIP requires two arguments.
Call Stack (most recent call first):
  src/CMakeLists.txt:315 (find_package)

-- Could NOT find GSHHG (missing: GSHHG_VERSION)
```

This is because conda-forge is doing cross-compiling, i.e., building an ARM OSX binary on an x64 OSX host machine. After reading some Google search results about "cmake cross-compiling try_run", I think the failure is because, the x64 OSX host machine cannot run the ARM OSX binray. So every `try_run()` call fails.

The [cmake documentation](https://cmake.org/cmake/help/latest/command/try_run.html#behavior-when-cross-compiling) recommends 

> In order to make cross compiling your project easier, use try_run only if really required. If you use try_run, use the RUN_OUTPUT_VARIABLE or OUTPUT_VARIABLE options only if really required. 

The solution here is, in cross-compiling mode (cmake variable `CMAKE_CROSSCOMPILING` is true), do not call `try_run()`. Here are some `try_run`s that are skipped in cross-compiling mode:

- No check on `HAVE_QSORT_R_GLIBC`
- No check on builtin byteswap (i.e., `HAVE___BUILTIN_BSWAP16`, `HAVE___BUILTIN_BSWAP32`, `HAVE___BUILTIN_BSWAP64` are all not set)
- Can still finds GSHHG and DCW file, but cannot determine GSHHG or DCW version, which requires running the `gshhg_vesion` binary.

I think these changes in cross-compiling mode are acceptable.


weiji14 (approved):
The patch works on https://github.com/conda-forge/gmt-feedstock/pull/199 so I'm happy with this, but 1) I have no clue about Cmake compiling and 2) I don't have a new Mac computer to test this. Read a bit about https://conda-forge.org/blog/posts/2020-10-29-macos-arm64 but it all seems very complicated :laughing:

Just curious though, does the [`macos-11` GitHub Actions runner ](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) currently build gmt on OSX Arm?

seisman:
> Just curious though, does the [`macos-11` GitHub Actions runner ](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) currently build gmt on OSX Arm?

I think it still doesn't, https://github.com/actions/virtual-environments/issues/2187.

PR #6441, opened 2022-03-14
More fixing and improvements in talwani?d

PaulWessel:
This PR improves the two talwani modules in a few ways:

1. It allows densities to be given either in g/cm^3 or kg/m^3
2. **talwani3d** now checks if its input is plain **grdcontour -D** slices and then properly reads the _depth_ from the segment header [no density will be available so needs to set it via **-D**].
3. **talwani3d** failed to scale the supplied z-level to km as done for other vertical distances.
4. Both modules now properly checks the **-F** option and defaults to gravity if no argument, else checks if **f** (one of them wrongly checked for **g** instead)

### DIFF
--- a/doc/rst/source/supplements/potential/talwani2d.rst
+++ b/doc/rst/source/supplements/potential/talwani2d.rst
@@ -52,8 +52,8 @@ Required Arguments
     One or more ASCII files describing cross-sectional polygons of one or more bodies.  Polygons
     will be automatically closed if not already closed, and repeated vertices will
     be eliminated.  The segment header for each body will be examined for a density
-    parameter in kg/m^3; see **-D** for overriding this value.  If no *table* is given then we
-    read standard input.
+    parameter in kg/m^3 or g/cm^3; see **-D** for overriding this value.  If no *table* is
+    given then we read standard input.
 
 Optional Arguments
 ------------------
@@ -66,7 +66,7 @@ Optional Arguments
 .. _-D:
 
 **-D**\ *density*
-    Sets a fixed density contrast that overrides any per-body settings in the model file, in kg/m^3.
+    Sets a fixed density contrast that overrides any per-body settings in the model file, in kg/m^3 or g/cm^3.
 
 .. _-F:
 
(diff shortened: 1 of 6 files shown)


meghanrjones (commented):
I am biased towards 'explicit is better than implicit' and would prefer appending a unit over automatically determining kg/m^3 vesus g/cm^3, but will defer to your preference.

More importantly, there are a couple failures on this branch:
```
647: Test command: /usr/local/bin/bash "gmtest" "test/potential/cyltest.sh"
647: Test timeout computed to be: 10000000
647: Set GMT_SESSION_NAME = 63105
647: talwani3d [WARNING]: Mismatch between actual (2) and expected (3) fields near line 1 in file /Users/meghanj/.gmt/cache/cylinder_mod.txt
647: talwani3d [ERROR]: Mismatch between actual (2) and expected (3) fields near line 2 in file trk
647: talwani3d [ERROR]: Input file trk has 0 column(s) but 3 are needed
647: ERROR: gmtest:90
647: memtrack errors: 0
647: exit status: 1
1/2 Test #647: test/potential/cyltest.sh ........***Failed    0.39 sec
test 669
    Start 669: test/potential/sphtest.sh

669: Test command: /usr/local/bin/bash "gmtest" "test/potential/sphtest.sh"
669: Test timeout computed to be: 10000000
669: Set GMT_SESSION_NAME = 63132
669: talwani3d [WARNING]: Mismatch between actual (2) and expected (3) fields near line 1 in file /Users/meghanj/.gmt/cache/sphere_mod.txt
669: talwani3d [ERROR]: Mismatch between actual (2) and expected (3) fields near line 2 in file trk
669: talwani3d [ERROR]: Input file trk has 0 column(s) but 3 are needed
669: ERROR: gmtest:90
669: memtrack errors: 0
669: exit status: 1
2/2 Test #669: test/potential/sphtest.sh ........***Failed    0.28 sec
```

PaulWessel:
Yikes.  Combination of things.  So the test is sloppy and do not specify any observation level.  it defaults to zero but that is not spelled out.  I am adding a warning.  The stricter checking on file format now backfires on me.  I will fix this but not until tomorrow so just let this sit for now as I am changing it to WIP for now.

PaulWessel:
I think I will declare a bug here: It is expected that the input profile have x,y,z unless z is a fixed quantity given by -Zlevel.  THe reason the test passed is that  the code set to expect 2 columns (for the model file, which has 2 coordinates) but when we then start to read the profile it was never changed to 3, so we never actually would have read a z-column anyway and later on there is a check to see if n_cols == 3 before we touch that array.  So if the test had used z-values in the profiles they would have been ignored (hence a bug).
I will fix the tests to pass -Z0.

meghanrjones (approved):

PR #6445, opened 2022-03-14
Add potential/gravprisms as new module for computing geopotential anomalies

PaulWessel:

This PR adds **gravprisms** which complements recent improvements to **grdseamount** related to a variable density structure:

1. It can compute the faa, geoid, or vgg over a set of vertical prisms - these may have different density values
2. Instead of reading prisms it can create prisms from a seamount density model given via grids.

The documentation to **gravprisms** includes a new figure created with it and **plot3d**, I show it here:

GMT_seamount_prisms

I have not added any test scripts yet (will try to add numerical scripts) but I have compared its output to other GMT modules for the cases we can compare and it is solid:

demo_1a_faa

For constant density between two surfaces it compares well with the more exact output from **grdgravmag3d** but is much faster.  For variable rho(r) is is more accurate than **gravfft**, and for rho(r,z) it is our only option.

I have not yet tested the vgg and geoid output hence WIP.

### DIFF
--- a/doc/rst/source/supplements/potential/gravprisms.rst
+++ b/doc/rst/source/supplements/potential/gravprisms.rst
@@ -259,6 +259,15 @@ for the same model (at 30S) is written to n_crossing.txt by
     gmt gravprisms -Ncrossing.txt -Mh prisms.txt -D1700 -Fn-30 -Gn_crossing.txt
 
 
+Note
+----
+
+The analytical expression for the geoid over a vertical prism (Nagy et al., 2000) is
+fairly involved and contains 48 terms.  Due to various cancellations the end result
+is more unstable than similar expressions for gravity and VGG.  Be aware that the
+result may have less significant digits that you may expect.
+
+
 References
 ----------
 


PaulWessel:
This module is now ready for review.  The single numerical test passes for gravity, VGG, and geoid and the output has been vetted with other modules.  I have added checks for the somewhat unstable geoid expressions (48 terms per prism) to catch exceptions (where a component should be zero rather than calling _atan_ and _log_ with 1/0 and 0 as arguments, respectively).  Documentation mentions this regarding the geoid.  I have done lots of comparisons with **gravfft**, **talwani3d** and **grdgravmag3d** for the cases that can be compared and all looks very good - exact matches for constant density cases.

joa-quim:
Sorry for bringing this up so late but maybe rename this module to ``gravmagprism``. I know It doesn't (yet) do mag case but it should only be a matter of replacing the ``grav`` prism function for the ``mag`` one, which can be done later.

PaulWessel:
ANother thought: We still have just gravfft even though one can do something very similar for magnetics, no?  Given the lack of options available and the possibility of needing specific magnetic options, perhaps gravprisms is not so bad - and when you get energized you can duplicate it to build magprisms.c?

joa-quim (approved):

PaulWessel:
> For the moment `grdgravmag3d` is resisting like hell to give a convincing result with variable density.

You have to be strict and show it who's the boss.

joa-quim:
There is a previous problem in the two layers case and that's why it fails, I think.

meghanrjones:
@PaulWessel do you have the prisms.txt file used in the examples for this module? If so, could we make this a remote file so that the examples are reproducible?

PaulWessel:
You mean for the plots at the top of this issue?  Or the stuff under EXAMPLES in the docs?

meghanrjones:
The stuff under EXAMPLES in the docs.

PR #6466, opened 2022-03-21
Fix the lack of precision in -Ib

PaulWessel:
The **-Ib** option outputs the bounding box of the grid.  But it failed to do so at the increment resolution and instead only output the 5 corner points.  That meant that projecting this polygon would fail to outline the polygon in an oblique projection. This PR corrects this bug by sampling the perimeter at the grid spacing.  Because pixel and gridline-registered grids have the save w/e/s/n but different node coordinates we make sure both types give the correct outer box and that the polygon is closed.

### DIFF
--- a/src/gmt_grdio.c
+++ b/src/gmt_grdio.c
@@ -3851,37 +3851,37 @@ unsigned int gmt_grid_perimeter (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *h
 	if (h->registration == GMT_GRID_PIXEL_REG) {	/* Must add the LL corner for pixel grids */
 		xx[k] = h->wesn[XLO];	yy[k] = h->wesn[YLO];	k++;
 	}
-	gmt_M_memcpy (&xx[k], gx, h->n_columns, double);
+	gmt_M_memcpy (&xx[k], gx, h->n_columns, double);	/* Can always use the x-coordinates along south or north */
 	for (col = 0; col < h->n_columns; col++, k++)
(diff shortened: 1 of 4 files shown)


PaulWessel:
## Basic checks

_Pixel grid:_
```
gmt grdmath -R-3000/2600/-800/800 -I100 -r 0 = p.grd
gmt grdinfo -Ib p.grd > pix_perim.txt
```

_Gridline grid_
```
gmt grdmath -R-3000/2600/-800/800 -I100  0 = g.grd
gmt grdinfo -Ib g.grd > grid_perim.txt

```

meghanrjones:
There are a few grdblend failures on this branch:
```
The following tests FAILED:
	392 - test/grdblend/center_bf_p.sh (Failed)
	394 - test/grdblend/center_cf_p.sh (Failed)
	396 - test/grdblend/center_nf_p.sh (Failed)
```

PaulWessel:
I had those before I made my update using the new gmt_grid_perimeter function and then I got no failures... I suppose they are tiny differences between the 5-point polygon and the detailed outline?

meghanrjones (approved):

meghanrjones:
> I had those before I made my update using the new gmt_grid_perimeter function and then I got no failures... I suppose they are tiny differences between the 5-point polygon and the detailed outline?

I was on a commit from before the update. Works fine with the latest changes.

PaulWessel:
Puh!

PaulWessel:
Let me add a non-plot test just in case.  Will take me a few minutes.

PR #6483, opened 2022-03-25
Get correct relative time for progress bars

PaulWessel:
Fixing a dumb bug in computing the relative time (0-1) from frame number so that it actually starts at 0 and ends at 1...

@meghanrjones, this affects on doc plot and the tests with progress bars:

```
doc/scripts/GMT_movie_progress.sh
test/movie/movie_indicator_de.sh (Failed)
test/movie/movie_indicator_f_hor.sh (Failed)
test/movie/movie_indicator_f_ver.sh (Failed)
```

Could you please update those in dvc?

### DIFF
--- a/src/movie.c
+++ b/src/movie.c
@@ -2069,7 +2069,7 @@ EXTERN_MSC int GMT_movie (void *V_API, int mode, void *args) {
 				/* Note: All dimensions are written in inches and read as inches in gmt_plotinit */
 
 				for (T = 0; T < Ctrl->n_items[k]; T++) {
-					t = (frame + 1.0) / n_frames;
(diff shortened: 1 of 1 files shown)


joa-quim (approved):

meghanrjones:
The new results for the failing tests do not look right to me. I am looking at the percent bar at the bottom of movie_indicator_de.ps and the 33 line does not look to be at 1/3 for the results after this PR, but looked fine before the PR.

PaulWessel:
Hm, well 33 is wrong as it is 31.25%.  I've checked all the others and I think the position is correct now (for 31.25) but clearly the %s value reported is off.  Will check.

PR #6486, opened 2022-03-26
Check if mode is suitable for grid in grdfilter

PaulWessel:
We have a set of distance modes in **grdfilter**.  Some are for Cartesian grids and some are for geographic grids (including complete spherical distances between nodes).  However, there were absolutely no checks that the user selected a sensible mode with their grid.  I was trying to filter a Cartesian grid and from memory thought `-D1` was the thing (but it should be `-D0`).  So I was a bit surprised seeing this on output and was about to file a bug report about strange repeating 360-ghosts:

shit

But of course, I cannot select `-D1 `with a Cartesian grid:

`       1: grid x,y in degrees, <width> in km, Cartesian distances.`

Thus, this PR checks and gives an error if you select a bad distance mode given the type of your grid.

While there might be special cases where the user want to treat their grid as something different that what it is, they can always use **-fg** or **-fc** to pretend otherwise.

### DIFF
--- a/src/grdfilter.c
+++ b/src/grdfilter.c
@@ -975,13 +975,13 @@ EXTERN_MSC int GMT_grdfilter (void *V_API, int mode, void *args) {
 		Return (API->error);
 	}
 	/* Check that -D option is compatible with the type of grid */
-	if (gmt_M_is_geographic (GMT, GMT_IN)) {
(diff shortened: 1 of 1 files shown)


Esteban82:
Is there a way to test it?

joa-quim (approved):

PaulWessel:
> Is there a way to test it?

Sure, take a geo grid and try -D0 or -Dp.  Or a cartesian grid and try -D2.  That is all the PR does.

PR #6492, opened 2022-03-28
Fix event -E parsing and add animation to documentation

PaulWessel:
This PR fixes some issues in **psevents** and adds a simple animated GIF to the man page:

1. If a symbol is selected with **-S** but no **-E** is set then we turn on **-Es** as default.
2. For readability I use the named constants in the parsing of **-M** instead of 0-3.
3. Fixed a bug where **-E** would not set **-Et** unless it had arguments.
4. Add _psevents_action.sh_ which makes an animated GIF.

The animation can be seen here and shows the difference between just plotting symbols during their lifetime (blue) versus enhancing it to have an arrival as well as a code (red).

one_event

@meghanrjones, I am not sure what to do with dvc here. The script above makes an animated gif so there is no PS file....
The gif is 2.8 Mb which is not bad for an animation.

### DIFF
--- a/doc/scripts/psevents_action.sh
+++ b/doc/scripts/psevents_action.sh
@@ -57,4 +57,4 @@ gmt begin
 	gmt sample1d normal.txt -T${MOVIE_COL0}, -Fl | gmt plot -Sc2p -Gblue -N
 gmt end
 EOF
-gmt movie -C22cx12cx100 main.sh -Sbpre.sh -Npsevents_action -T-0.5/1.5/0.01 -D24 -Fmp4 -Lc0 -Lf+jTR -Pf+jBC+o0/1.5c+ac -M150,png -Z
+gmt movie -C22cx12cx100 main.sh -Sbpre.sh -Npsevents_action -T-0.5/1.5/0.01 -D24 -Fnone -Lc0 -Lf+jTR -Pf+jBC+o0/1.5c+ac -M150,ps -Z
--- a/doc/scripts/psevents_action.sh
+++ b/doc/scripts/psevents_action.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 # Create a short MP4 showing the life of two simple events
+# By default, this script only generated the master frame
+# Modify -Fnone to -Fmp4 in L58 to generate the MP4
 # Both events starts at 0 and ends at 1, but we add rise
 # and fade and coda to one of them (red) and let the other be plain (blue).
 echo 20 6 0 1 Red Label > red.txt
--- a/doc/scripts/psevents_action.sh
+++ b/doc/scripts/psevents_action.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+# Create a short animated GIF showing the life of two simple events
+# Both events starts at 0 and ends at 1, but we add rise
+# and fade and coda to one of them (red) and let the other be plain (blue).
+echo 20 6 0 1 Red Label > red.txt
+echo  0 6 0 1 Blue Label > blue.txt
+cat <<- EOF > labels.txt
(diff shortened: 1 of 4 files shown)


joa-quim (approved):

meghanrjones:
Could we make the animation an .mp4 rather than gif and include it on the animations page and reference it from `events` rather than including it with the static images in `doc/scripts`? We could then host the animation on YouTube.

meghanrjones:
> Not sure we want to list it on the animation page but other that that I agree.

That's fine. We need to take steps to prevent building the movie during testing because 13s is too long for a single test. I think we should use `-Fnone -M150,ps` in the script with a note that the YouTube mp4 can be generated using `-Fmp4`. This way, the master frame generation is included in the tests without generating the full mp4.

The alternative would be to exclude the scripts from the tests entirely by modifying doc/scripts/CMakeLists.txt.

Also, I think that the animation would look better in the docs after the description. Preview is at https://gmt-ckq72z30r-gmt.vercel.app/events.html.

PaulWessel:
Looks like I cannot set a movie caption the same way we set figure captions.  So here it is just another paragraph.  We can move it below the description.

doc/scripts/psevents_action.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+# Create a short MP4 showing the life of two simple events
```
meghanrjones, line 2:
```suggestion
# Create a short MP4 showing the life of two simple events
# By default, this script only generated the master frame
# Modify -Fnone to -Fmp4 in L58 to generate the MP4
```

doc/scripts/psevents_action.sh
```diff
@@ -0,0 +1,58 @@
...
+EOF
+cat << 'EOF' > main.sh
+gmt begin
+	gmt events red.txt -T${MOVIE_COL0} -R-20/40/1/9 -JX20c/10c -B -Es+r0.25+p0.25+d0.25+f0.25 -Et+o0.25 -Sc2c -Gred -W1p -Ms2+c0.25 -Mi1+c-0.5 -Mt100+c50 -F+f18p+jBC -Dj3c -L -X1c -Y1c
+	gmt events blue.txt -T${MOVIE_COL0} -Sc2c -Gblue -W1p -F+f18p+jBC -Dj3c -L -E
+	gmt sample1d size_vs_time.txt -T${MOVIE_COL0}, -Fl | gmt plot -Sc4p -Gred -R-0.5/1.5/-0.1/2.1 -JX13.2c/2c -N -X3.4c -Y1.25c
+	gmt sample1d normal.txt -T${MOVIE_COL0}, -Fl | gmt plot -Sc2p -Gblue -N
+gmt end
+EOF
+gmt movie -C22cx12cx100 main.sh -Sbpre.sh -Npsevents_action -T-0.5/1.5/0.01 -D24 -Fmp4 -Lc0 -Lf+jTR -Pf+jBC+o0/1.5c+ac -M150,png -Z
```
meghanrjones, line 58:
```suggestion
gmt movie -C22cx12cx100 main.sh -Sbpre.sh -Npsevents_action -T-0.5/1.5/0.01 -D24 -Fnone -Lc0 -Lf+jTR -Pf+jBC+o0/1.5c+ac -M150,ps -Z
```

meghanrjones (commented):
Two suggested changes so that we can add the master frame to dvc in doc/scripts/images and test the output in the CI.

PaulWessel:
@meghanrjones, is this good to go you think?

meghanrjones:
> @meghanrjones, is this good to go you think?

We just need the result from the script in dvc. Do you want me to push that change to this branch?

meghanrjones (approved):

PR #6501, opened 2022-03-30
WIP address concerns related to 3-D plots

PaulWessel:
There are two issues addressed in this PR:

1. There is a check to parse txt_c for z-values given via **+w** on line 2359 but the count _k_ is not based on the number of arguments to **+w** but the main `az/el[/z] `argument.
2. There is a default setting for view point that is based on paper size, which may not play well in modern mode or make sense in classic mode either.

While I think (1) above is a relatively harmless bug, but a bug nonetheless, (2) is less clear.  Changing it as indicated (I just commented out the original commands for now) did not negatively affect the tests.  However, I had noticed that if I tried to use **+w** instead of **+v** in a movie I got strange offsets on the plot (using **+v** worked fine by comparison).  E.g., compare

```
gmt basemap -R-600/3200/-600/600 -Jx0.004c -Baf -p130/25+w2000/0 -pdf t
gmt psbasemap -R-600/3200/-600/600 -Jx0.004c -Baf -p130/25+w2000/0 -P > t.ps
```

The modern mode plot is fine since we have a 10x10 meter canvas and we crop to the final plot. No worries. The classic mode plot implicitly sets viewpoint **+v** based on the current paper size (if **+v** not given).  Here is the complete page:

t

Quite a bit of vertical offset since the positioning (in master) is based on the page size.  My revision is to pick the mid-point of the current plot region (2-D thinking) instead of the page size (also 2-D thinking).

Hoping that perhaps @remkos or @joa-quim might recall why we used the page center, and if not, perhaps picking a point closer to the actual plotting is a better solution (leaving it as is makes it impossible to use **+w** in laying out several plots for a movie since I have no control really on those vertical shifts).  Just for reference, the same plot using this PR looks like this, which is closer to what you get if you use no **+v** or **+w**.:

new

I have left this PR as WIP but unless I hear lots of opposition I think I will remove the WIP.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -2305,7 +2305,7 @@ int gmtinit_parse_n_option (struct GMT_CTRL *GMT, char *item) {
 
 /*! . */
 GMT_LOCAL int gmtinit_parse_p_option (struct GMT_CTRL *GMT, char *item) {
-	unsigned int k, l = 0, pos = 0, error = 0;
+	unsigned int k, l = 0, n, pos = 0, error = 0;
 	double az, el = 0.0, z = 0.0;
 	char txt_a[GMT_LEN256] = {""}, txt_b[GMT_LEN256] = {""}, txt_c[GMT_LEN256] = {""};
 	char p[GMT_LEN256] = {""}, *c = NULL;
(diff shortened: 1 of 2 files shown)


PaulWessel:
Also, while the documentation says one of **+v** or **+w** can be set, the code is happy to accept both.  From the logic it seems acceptable to set both but not sure if we want that or should add a check to prevent it.

PaulWessel:
For now, I let the PR only fix the parsing bug.  The other issue will have to wait.

PR #6503, opened 2022-03-30
Let GMT timestamp via -U be placed at the end of the plot

### LINKED ISSUE #6499: Time stamp (-U) is hiden under images
If I do this, I see the Time stamp at it's default position
```
julia> imshow(rand(256,256), U="", Vd=1)
        grdimage  -R0.5/256.5/0.5/256.5 -JX14c/0 -Baf -BWSen -U -n+a -C -P -K > C:\TEMP\GMTjl_tmp.ps
```
however
```
julia> imshow(rand(256,256), U="+o0/0", Vd=1)
        grdimage  -R0.5/256.5/0.5/256.5 -JX14c/0 -Baf -BWSen -U+o0/0 -n+a -C -P -K > C:\TEMP\GMTjl_tmp.ps
```
(or any other position inside the axes) shows no Time stamp.

PaulWessel:
Since the Bush 1 administration, the timestamp set via **-U** was plotted first.  Thus, if you placed it inside the plot canvas it would get overwritten.  This PR simply moves it from the _gmt_plotinit_ to the _gmt_plotend_ call sequence so that it is the last item plotted.

Closes #6499.

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -9114,8 +9114,8 @@ struct PSL_CTRL *gmt_plotinit (struct GMT_CTRL *GMT, struct GMT_OPTION *options)
 		}
 		GMT->current.ps.logo_cmd = false;	/* Mission accomplished */
 	}
-	if (GMT->current.setting.map_logo)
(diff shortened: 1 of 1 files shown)


joa-quim (approved):
Thanks. This is useful in animations where the default position may cause extra white space in movies.

PR #6518, opened 2022-04-04
Update scripts for docs to use modern mode

meghanrjones:

This PR modifies scripts in docs/scripts to use modern rather than classic mode and updates a few baseline images for testing due to formatting differences between the two modes.

Fixes #

**Reminders**


PaulWessel (approved):
Looks good, but not that many scripts have a last line in the script that has no liinefeed after it (red stop sign).  Probably need to add those.

PR #5953, opened 2021-11-05
Make sure create_table sets allocation mode

PaulWessel:
I noticed that **gmtmath** were unable to free datasets it had allocated because the tables never had their _alloc_mode_ set to **GMT_ALLOC_INTERNALLY**.  No error related to this, just a memory leak I am trying to track down in gmtmath.c and this is part of it.

PR #6521, opened 2022-04-04
Fix API case for external data passed as vectors

meghanrjones:

This PR adds a version of the bug report from https://github.com/GenericMappingTools/pygmt/issues/1691 to the C API tests. Currently crashes.

Fixes #

**Reminders**

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -3980,9 +3980,11 @@ GMT_LOCAL struct GMT_DATASET * gmtapi_import_dataset (struct GMTAPI_CTRL *API, i
 				if ((V_obj = S_obj->resource) == NULL) {
 					gmt_M_free (GMT, D_obj);	return_null (API, GMT_PTR_IS_NULL);
 				}
-				if (V_obj->type[0] != GMT_DOUBLE) {
-					GMT_Report (API, GMT_MSG_ERROR, "Only double-precision vectors can be passed via reference to datasets\n");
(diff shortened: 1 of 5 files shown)


meghanrjones:
@PaulWessel, I found that this test uses GMT_IS_REFERENCE and crashes as-is but passes if I modify `y` to be `GMT_UINT` (similar to https://github.com/GenericMappingTools/pygmt/issues/1691). Do you think you could use this to debug the issues related to https://github.com/GenericMappingTools/gmt/pull/5953, to circumvent the issues with conda on M1?

PaulWessel:
Yes possibly, I will give it a check tomorrow morning.

meghanrjones:
> Yes possibly, I will give it a check tomorrow morning.

Thanks

PaulWessel:
I had a conceptual error in my thinking when it came to data sets being passed in via vectors and GMT_IS_REFERENCE:  Since we are creating a DATASET container for the input and we set its pointers to the external arrays, we do not need to consult which columns were allocated externally or internally since all of them shall be treated as external so that when the module (here psxy) finishes it does not attempt to free any of them (even one that was allocated by GMT in main) since it will be freed at that level instead.

meghanrjones:
> See if this also fixes the original pyGMT issue, @meghanrjones . That crash is gone for me.

Yes, the PyGMT issue is fixes. Thanks. I will need your approval on the PR before merging.

PaulWessel (approved):
Oops

Issue #559, opened 2019-03-22
grdcontour - Plotting of open/closed contours and dumping of closed contours


anbj:
This is a copy of a feature request from the old website (http://gmt.soest.hawaii.edu/issues/914). I'm particularly interested in 2.

------------

Some suggestions for an *even* better grdcontour:

1) Add an option for plotting only open or closed contours (sometimes, you want to focus only on the closed contours or only on the open contours)
2) (Related to the existing ability of dumping contours based on whether they are closed or open, -D%c) Add an option to distinguish between closed contours based on whether they are closed in a downward or upward direction (not talking about an algorithm for finding locate valleys and hilltops) - e.g. dump closed contours as data line segments into two separate files; for example **C_upwards.gmt** and **C_downwards.gmt**. GMT is obviously capable of this - ref. the -T option in grdcontour.

anbj:
Lack of interest. This is an old one. Closing.

PR #6522, opened 2022-04-05
Allow contour files to only list contour level

PaulWessel:
Since it is **awk**ward to create a list of desired contours and add the character C to the end, this PR allows a plain single-column file of contour levels to be accepted in **-C** by **grdcontour** and **pscontour** and the type defaults to plain contour (C) only. Currently we just get an ugly read error. This is useful if you have some custom contours, e.g., want to contour close to zero but not exactly, then every 1:

```
echo 0.001 > cont.txt
gmt math -T1/10/1 -o0 T = >> cont.txt
```

This avoids having to awk to add C or to suffer a crash,


anbj:
Very nice. Reminds me that #559 is still buried somewhere :)

joa-quim (approved):

PR #6525, opened 2022-04-05
grdcontour.rst: reorder description and change order of default value…

anbj:
Not sure about this. Please review.

*Shift* is mentioned before *factor*, but the default values are indicated with *[1/0]*. Should it not be *[0/1]*; i.e. shift = 0 and factor = 1?


PaulWessel (commented):
The stuff in brackets is GMT-4 style so that should be chagned.  Usually, I now try to list teh modifiers in alphabetical order.  So I think what we have is OK, except we want [+o0+s1] in the brackets.

PaulWessel (approved):
Thanks for noticing!

PR #6520, opened 2022-04-04
Allow use of GMT_GRAPHICS_FORMAT for one-liners

meghanrjones:
**Description of proposed changes**

This PR allows specifying through the prefix `view` that the output format should be set by `GMT_GRAPHICS_FORMAT`, e.g.,:

```
gmt basemap -R2000/2020/35/45 -JX12c -Bxa2f+a-30 -BS --GMT_THEME=cookbook -view GMT_-B_slanted
```

Closes #6508 

**Reminders**

- [ ] Make sure that your code follows our style. Use the other functions/files as a basis.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Describe changes to function behavior and arguments in a comment below the function declaration.
- [ ] If adding new functionality, add a detailed description to the documentation and/or an example.

PR #6526, opened 2022-04-05
Support using GMT_GRAPHICS_FORMAT through view format in movie -M

meghanrjones:

Follow up to https://github.com/GenericMappingTools/gmt/pull/6520. This PR allows using `view` as a format for movie's -M option to use GMT_GRAPHICS_FORMAT.

Fixes #

**Reminders**

### DIFF
--- a/src/movie.c
+++ b/src/movie.c
@@ -95,6 +95,7 @@
 #ifdef WIN32
 #include <windows.h>
 #endif
+#include "gmt_gsformats.h"
 
 #define THIS_MODULE_CLASSIC_NAME	"movie"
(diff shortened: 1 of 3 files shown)


src/movie.c
```diff
@@ -95,6 +95,7 @@
 #ifdef WIN32
 #include <windows.h>
 #endif
+#include "gmt_gsformats.h"
```
meghanrjones, line 98: I couldn't really figure out based on the other modules if there is a convention for ordering the header files relative to each other and the macros, so let me know if this isn't the right place for this line.

PaulWessel (approved):
Looks good

PR #6473, opened 2022-03-22
WIP: Generate images when building docs

meghanrjones:
**Description of proposed changes**

This PR revises the CMake directives to generate the images used in the documentation when building the docs, rather than relying on the dvc-tracked images, when `set (NO_DVC TRUE)` is used in the CMake configuration.

Fixes #6356

**Reminders**

- [ ] Make sure that your code follows our style. Use the other functions/files as a basis.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Describe changes to function behavior and arguments in a comment below the function declaration.
- [ ] If adding new functionality, add a detailed description to the documentation and/or an example.

Issue #6356, opened 2022-02-19
Add CMake target to generate images for the docs


meghanrjones:
**Description of the desired feature**

As a follow-up to https://github.com/GenericMappingTools/gmt/pull/6267, we should add a CMake target to generate the images required for building the documentation, so that those without dvc can still build the documentation.

**Are you willing to help implement and maintain this feature?** Yes

meghanrjones:
I'm running into a problem with [example 51](https://docs.generic-mapping-tools.org/dev/gallery/ex51.html) because the script cannot be run without downloading a ~630 MB file. Workarounds include:
1) modify the script to generate an image without needing that file
2) use a GitHub hosted image similar to [example 53](https://docs.generic-mapping-tools.org/dev/gallery/ex53.html)
3) just require dvc for building the docs and abandon the effort for an alternate option

The downside to the GitHub hosted images is that they are not available in bundled or offline documentation (https://github.com/GenericMappingTools/gmt/issues/6032). But I still prefer option 2 because the example is useful and providing an alternative to dvc is important for allowing people to build the docs from GMT distributions. Any thoughts @GenericMappingTools/gmt-maintainers?

meghanrjones:
Superseded by https://github.com/GenericMappingTools/gmt/pull/6533

Issue #5724, opened 2021-09-01
Proposal for managing test baseline images using data version control (dvc)


maxrjones:
# Proposal for managing test baseline images using data version control (dvc)

This issue proposes a solution to #3470 and a partial solution to #2681 by using [data version control](https://dvc.org/) to manage the baseline images for testing. @weiji14 led an effort to move PyGMT's tests from git version control to data version control with remotes stored on [DAGsHub](https://dagshub.com/) in https://github.com/GenericMappingTools/pygmt/pull/1036; most of the information here is from Wei Ji's posts for PyGMT (thanks! 🙏 🎉 ).

## Motivation for migrating baseline images to dvc

Here's the current breakdown for the GMT repository:

- `.git`: ~1.1 GB (up from ~720 MB on Feb. 06 2020)
- `test`: ~115 MB (101 MB from PS files) (up from ~113 MB on Feb 06. 2020)
- `doc`: ~68 MB (51 MB from PS files; 33 MB from PS in `doc/examples` ; 18 MB from PS in `doc/scripts`) (down from ~70 MB on Feb. 06 2020)
- `share`: ~13.5 MB
- `src`: ~16 MB

The fact that the overall repository size increased by 50% over the past 1.5 years while individual directories have remained the same size supports past developer comments that the repository growth rate due to rewriting PS files is unsustainable.

## What is data version control

Data version control (dvc) is an open source tool for managing and versioning datasets and models. It is built on Git with very similar syntax. Rather than storing bulky images in the repository, small .dvc files are stored that contain metadata, including the md5 hash for the data file. This allows versioning of data files that are stored in a remote location. Options for remote storage include S3, Google cloud, Azure, SSH server and DAGsHub (PyGMT uses DAGsHub).

## Steps required
(Based on PyGMT, may need some updating)

- [x] Add DVC as a dependency for developing GMT
- [x] Initialize dvc in the repository (e.g., https://github.com/GenericMappingTools/pygmt/pull/1036)
- [x] Setup DVC remote
- [x] Add instructions for using DVC for image-based testing to the contributing guide
- [x] Setup dvc for CI tests
- [ ] Add a workflow to support side-by-side comparison of modified images (e.g., https://github.com/GenericMappingTools/pygmt/pull/1219)
- [x] Migrate existing test baseline images to DVC (e.g., https://github.com/GenericMappingTools/pygmt/issues/1131)
[...]

PR #6533, opened 2022-04-07
Include PS images for docs in releases

meghanrjones:

In past GMT releases, users could build the documentation from the source tarballs because the .PS files were included in the `doc/examples` and `doc/scripts` directories. This PR maintains that option for future releases by copying the `doc/examples/images` and `doc/scripts/images` directories to the release directory.

Supersedes https://github.com/GenericMappingTools/gmt/pull/6473, which tried to solve the problem by generating the images directly. I think this PR is better because it is simpler and therefore easier to maintain. While https://github.com/GenericMappingTools/gmt/pull/6473 worked for me locally, it failed to work in the CI due to issues finding dcw and the supplement modules.

Whereas https://github.com/GenericMappingTools/gmt/pull/6473 would have provided an option for building the docs either from a git clone or the releases without DVC, this PR only applies to the releases. To build the docs from a git clone, users will still need to install dvc and run dvc pull first.

Relates to https://github.com/GenericMappingTools/gmt/issues/5724, https://github.com/GenericMappingTools/gmt/issues/6356

### DIFF
--- a/doc/examples/CMakeLists.txt
+++ b/doc/examples/CMakeLists.txt
@@ -15,6 +15,16 @@
 #-------------------------------------------------------------------------------
 
 if (SPHINX_FOUND)
+	# Copy PS sources for gmt release
+	if (GIT_FOUND AND HAVE_GIT_VERSION)
+		add_custom_target (_examples_images_release
+			COMMAND ${CMAKE_COMMAND} -E copy_directory
+			${CMAKE_CURRENT_SOURCE_DIR}/images/
+			${GMT_RELEASE_PREFIX}/doc/examples/images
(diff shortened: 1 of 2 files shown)


PaulWessel:
Seems to work.  I notice the tar ball has both PS and PNG files, e.g.

```
./gmt-6.4.0/doc_release/html/_images/GMT_seamount_flux.png
./gmt-6.4.0/doc/scripts/images/GMT_seamount_flux.ps
```

I guess this is intentional since the png supports to html tree while the PS file is there for building it.

PaulWessel (approved):

Esteban82:
This can be merge, right?

PR #6547, opened 2022-04-10
Fix error in setting max half-width in node search

PaulWessel:
When we need to access all nodes inside a given _search radius_ (as we do in **grdfilter**, **nearneighbor**, **grdmask** etc), we first determine limits on the number of rows and columns away from the search center.  For geographic data this also means we consider the change in longitudinal distance with latitude.  All this works fine.  However, because for high latitudes (even ±90) we may divide by cos(lat) and hence we had a "safety valve" that clipped the upper limit of how many columns we may want to search to either side of a node to `n_columns/2`.  This is an error: If you are at the far left column and have a search radius that equals or exceeds the longitude range then you want _all_ the columns to be inside, hence the upper limit must be `n_columns`.

No tests affected but I ran into this with a **grdseamount**  research case where I only used a small region to view a portion of a seamount whose radius exceeded half the width and things got oddly clipped in longitude but not in latitude...

uff

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -15626,7 +15626,7 @@ openmp_int * gmt_prep_nodesearch (struct GMT_CTRL *GMT, struct GMT_GRID *G, doub
 
 	dist_y = gmt_distance (GMT, G->header->wesn[XLO], G->header->wesn[YLO], G->header->wesn[XLO], G->header->wesn[YLO] + G->header->inc[GMT_Y]);
 	if (mode) {	/* Input data is geographical, so circle widens with latitude due to cos(lat) effect */
-		max_d_col = urint (ceil (G->header->n_columns / 2.0) + 0.1);	/* Upper limit on +- halfwidth */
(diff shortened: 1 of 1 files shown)


PaulWessel:
Also a good example of a problem that does take a bit of testing and debugging to figure out, and in the end a single line of code is at fault (and have been so for years).

PaulWessel:
Just note after that I wrote the summary on top I realized we can improve matters further:

- The upper limit of `n_columns` also applies to the Cartesian case
- A similar limit of `n_rows` should also apply to the y-direction

Without these we might occasionally request too much memory for a filter, so might as well impose these limits.  Again, no change to any tests.

PaulWessel:
I guess it is Easter Sunday and only godless Norwegians are working.  I will review this myself...

joa-quim:
> I guess it is Easter Sunday

That's next Sunday. Today is the day when we go chase for wild orchids.

 
image

PR #6523, opened 2022-04-05
Improve clip path for images in -Qi

PaulWessel:
The algorithm used for building the staircase clip path for the image in **grdview -Qi** was somewhat smart but not smart _enough_, leaving a path when there is no valid image to the far left or right.  While the clipping worked fine, the fact that the path extends too far means **ghostscript** senses these excursions and hence the bounding box is too large.

This PR eliminates that problem.  I also added more comments to explain the otherwise cryptic section.  The plot below show an actual 3-D image clip path.  The BB of the full projected image is 0/8.37/0/3. The heavy line was the previous version and the new version is the thin red line.

No tests were affected.

mask

PR #6548, opened 2022-04-10
Add landslide option -S for grdseamount

PaulWessel:
GMT_seamount_slide

This feature makes it possible to recreate some of the ancient landslide modeling that was used in [J. Smith and Wessel, 2000](https://dx.doi.org/10.1007/s000240050019) on the isostatic consequence of large landslides in oceanic islands.  This feature required a lot of work on the input, processing of options and modifiers, and the basic math of handling the many tunable functions and parameters that defines a landslide, including time duration.  The man page has been revamped with 5 new figures to explain the purpose and function of the radial slide shape, azimuthal variation within a sector, and the temporal volume fraction curve during the event.  I do not expect anyone to really review this PR since it is very extensive, but I don't think it is necessary.  I am actively using it for research and thus any issues will quickly be found as I am exploring various situations.  As is, it passes all the existing tests, that is, after I fixed a few problems and had to update some of the plots.  The current version is thus better in every way.  I would want it merged for 6.4 since a publication we are working on will use this and reviewers may wish to have access.

Here are some of the things (some has nothing to do with **grdseamount** but got fixed along the way):

1. A few references got reformatted to keep everything the same
2. Five new documentation scripts to make it possible to navigate the settings
3. Extensive improvements to **grdseamount** documentation
4. Fixed bug where the backwards compatible code **l** (for linear flux rate) was in fact not checked for but instead mixed up in checking the build mode (which is either **c** or **i**).
5. Normalized the error function volumetric curve so it goes exactly from 0-1; otherwise the first (and last) volume increments are way too large and very jarring during a movie.
6. Changed a few of the existing scripts to accommodate the above bug fix so we don't need to update too many figures
7. Fixed error in the flux curve plot - it had a factor of 6/2 instead of 3/sqrt(2) inside the error function
8. Fix the seamount types script so that it handles the singularity in the polynomial equation (no way **gmtmath** can know this).  **grdseamount** computes that same shape via a function that was fixed to handle the singularity (#6476).
9. Fix error in **-W** parsing where it used **GMT_IS_CUBE** instead of **GMT_IS_GRID**.  Harmless but wrong.
10. Fixed (via separate PRs) several GMT issues that was discovered during the **grdseamount** work, such as #6547, #6536, #6535, #6523, and #6496; there may be others I forgot.

The **-S** option lets uses specify one or more slides per seamount, each with a different set of parameters and time history.  In crossection it looks something like this (but the shape of the slide is tunable):

GMT_seamount_specs


joa-quim (approved):
One day I'll have to find time to see if I can use this as source for tsunami modeling.

PaulWessel:
Yes, that is a good one.  I will now try to grow an entire seamount chain in the Pacific.

PR #6553, opened 2022-04-14
Three things on a lazy Thursday

### LINKED ISSUE #6551: Trying to plot decorated lines with arrows crashes Julia
I first thought it was because no angle and length columns in input data, but no. It's more *fundamental*

This KaBooms Julia
```
julia> plot([0.0 0 20 1; 0.4 0.6 20 1; 1 1 20 1], S="~d2c:+sv0.3", show=1, Vd=1)
        psxy  -JX14c/9.5c -Baf -BWSen -R-0.04/1.04/-0.04/1.04 -S~d2c:+sv0.3 -P -K > C:\TEMP\GMTjl_tmp.ps
```
So, we can't make decorated lines with arrows.

PaulWessel:
This PR uncortunately does three very different things; I was too lazy to break the commits up...

1. Fix usage of _gmt_access_. This general function returns nonzero if there are problems finding the given file.  However, it already does a few other things:

- If file is a memory file it returns 0 since memory is there
- If file is a cache file then it will first check if we should download it

Hence, there are no points checking both for access and if cache or memory file in various places since _gmt_access_ handles this.  

2. Let _gmtplot_flush_symbol_piece_ be more specific and just set flush to false and return if there are no lines. 

3. Fix _gmtapi_extract_argument_ for the colon options.  Needed a special section to handle the quoted, decorated, or contoured line settings where (after stripping of any colon args) we know exactly what to do. With this the -S~f works for me in Julia.

4. Remove an unused variable.

Closes #6551.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12948,9 +12948,9 @@ struct GMT_RESOURCE * GMT_Encode_Options (void *V_API, const char *module_name, 
 	else if ((!strncmp (module, "psxy", 4U) || !strncmp (module, "psxyz", 5U)) && (opt = GMT_Find_Option (API, 'S', *head))) {
 		/* Found the -S option, check if we requested quoted or decorated lines via fixed or crossing lines */
 		/* If not f|x then we don't want this at all and set type = ! */
-		type = (!strchr ("~q", opt->arg[0]) || !strchr ("fx", opt->arg[1])) ? '!' : 'D';
-		strip_colon = (type && strchr (opt->arg, ':'));
-		strip_colon_opt = opt->option;
+		type = (!strchr ("~q", opt->arg[0]) || !strchr ("fx", opt->arg[1])) ? '!' : 'D';  /* Only -S[~q][fx] will yield D */
+		strip_colon = (strchr (opt->arg, ':' != NULL)); /* true if optional arguments beginning with colon, otherwise false */
+		strip_colon_opt = opt->option;    /* The option with (a possible) colon argument */
 		if (strip_colon)
 			GMT_Report (API, GMT_MSG_DEBUG, "GMT_Encode_Options: Got quoted or decorate line and must strip argument %s from colon to end\n", opt->arg);
 	}
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12685,48 +12685,61 @@ GMT_LOCAL const char *gmtapi_retrieve_module_keys (void *V_API, char *module) {
 	return_null (V_API, GMT_NOT_A_VALID_MODULE);
 }
 
-GMT_LOCAL int gmtapi_extract_argument (char *optarg, char *argument, char **key, int k, bool colon, int *n_pre, unsigned int *takes_mod) {
-	/* Two separate actions:
+GMT_LOCAL int gmtapi_extract_argument (char *optarg, char *argument, char **key, int k, bool colon, char colon_opt, int *n_pre, unsigned int *takes_mod) {
+	/* A few separate actions:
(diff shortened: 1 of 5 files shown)


PaulWessel:
Since the ffile[/slop] was never documented I am considering just removing that part of the code.  WOuld that be OK with you?

joa-quim (approved):

joa-quim:
The ``slop`` value is currently so small (10-8) that only integers would pass that test. Not sure what the coordinates are there (they seem to be inches). Catching points on a neighborhood of ~0.5 mm would be reasonable, I think.

joa-quim:
> With this the -S~f works for me in Julia.

You mean you passed an array as argument to *f*?

PaulWessel:
Yes, I ran the example you had posted:

`julia> gmt("psxy -S~f:+skarrow/0.25+gblack -P", [0.0 0.0; 0.4 0.6; 1 1], [0.2 0.3])
`

except I replaced the arrow with some symbol and added a -R -J for it to run.

joa-quim:
Good. It would be nice to be able to pass kustom symbols  too as def files, but don't know in what format. A dataset with a text column?

PaulWessel:
Yes, I guess we can relax the slop a bit.  It will still search for the closest point so should have no side-effects for existing cases with exact fits.  Yes, internally the world is in inches so we could set a constant

`#define GMT_FIXED_POINT_SLOP (0.05 / 2.54) /* Close point threshold in -Gf and -S[~q]f is 0.5 mm */`

and use that.  I will then remove the checking for two args [well ,after I look at all our tests to verify that we never actually used this]

PaulWessel:
Hm, reading custom symbols and passing via datasets...
Well, we do have the ability to read a file just composed of text.  Given the highly variable column counts and some commands starting with letters in first column I think that would be the only option.

I dont think it would be too much work but do you really want to do this?  We have

```
	if ((fp = fopen (path, "r")) == NULL) {
		GMT_Report (GMT->parent, GMT_MSG_ERROR, "Could not find custom symbol %s\n", &name[pos]);
		return GMT_ERROR_ON_FOPEN;
	}

	while (fgets (buffer, GMT_BUFSIZ, fp)) {

```
and that would need to be

```
if ((D = GMT_Read_Data (.....)

and (assuming all custom symbols are a single segment in a single file)
S = D->table[0]->segment[0];
for (row = 0; row < S->n_rows; rows++) {
   strcpy (buffer, S->text[row]);
....

```

PaulWessel:
Oj well, the slop is fully documented in Appendix O, with example of slop etc.  We even say "The default slop is zero, meaning only exact coordinate matches will do."

So then back to plan B of documenting it better I think.

joa-quim:
The annoying thing with the kustoms are that we need to know where they are and pass the full name. For simple ones like my arrow head (which, BTW, would a ice adding to our .def collection) we can create them programmatically. Ofc, this would be usable only from Julia and MEX.

PaulWessel:
While changing the reading to use _GMT_Read_Data_ is simple, there are other issues:

- **KEYS**. I think custom symbols can be passed either as the symbol (**-Sk**_name_/size) or as the symbol in decorated lines (**-S~**_code_:**sk**_name_/_size_...  We may have to finesse what the KEYS are for -S under the hood since we want somehow to turn on optional reading.
- Then there is what GMT_Encode_Option should do.  In may cases we simply leave off the name when providing data from the external.  So how would -Sk/size work in the parser?  This needs to be -Sk?/size at some point.  Likewise,` -S~f:sk/0.2c` will need to become` -S~f?:sk?/0.2c` which is something we have not had before...
- 

I think this may be one of those cases where it is just simpler to write whatever memory symbol you have to a temp file...

PR #6554, opened 2022-04-14
Document the slop for fixed points in quoted or decorated lines

PaulWessel:
We use this mechanism for both contours, decorated lines, and quoted lines so the syntax is the same and this PR makes sure the man pages matches what is shown in Appendix O and contour programs.


joa-quim:
I think it needs a reference to the slop units (forced to go away for about ~1 hour)

PaulWessel:
Well, it spells that out explicitly, so that is done.  You can use units like you do anywhere ese, e.g.  .../0.05c

joa-quim (approved):

PR #6558, opened 2022-04-14
Streamline parsing of required arguments

PaulWessel:
The module parsers are full of situations where we must obtain a required argument after the option or a modifier.  We thus have lots of repeated checks to see if there is an argument or not, then parse it.  E.g.,

```
if (opt->arg)
	Ctrl->D.value = atof (opt->arg);
else {
	GMT_Report (API, GMT_MSG_ERROR, "Option -D: No argument given.\n");
	n_errors++;
}
break;
```

For file names, we also have preliminary checks for existence of input files.  This goes on in 150 modules.  In many places we have failed to perform a check if an argument was given before converting the string, and in other places where no argument is expected we fail to check if one was erroneously given.

This PR starts a new path forward where we place most of the checking and reporting of such errors in small sub-functions that returns an error code or 0, and passes any value obtained via a pointer.  This drastically reduces the number of lines in the parsers and consolidates the checks in just a few places.  The above line reduces to

`n_errors += gmt_get_required_double (GMT, opt->arg, opt->option, 0, &Ctrl->D.value);`

This PR just implements this for **batch**, **dimfilter**, **filter1d** and the blockers, but will wash across the others in groups of PRs as time permits.

### DIFF
--- a/src/batch.c
+++ b/src/batch.c
@@ -240,31 +240,28 @@ static int parse (struct GMT_CTRL *GMT, struct BATCH_CTRL *Ctrl, struct GMT_OPTI
 		switch (opt->option) {
 
 			case '<':	/* Input file */
-				if (n_files++ > 0) break;
-				if (opt->arg[0]) Ctrl->In.file = strdup (opt->arg);
-				if (GMT_Get_FilePath (API, GMT_IS_DATASET, GMT_IN, GMT_FILE_REMOTE, &(Ctrl->In.file)))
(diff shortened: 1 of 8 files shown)


Esteban82 (approved):
I approve without understanding it very well.

PaulWessel:
> I approve without understanding it very well.

That is how it is for most of us...
I think as long as I run all tests and they pass we are on the safe side.

PR #6563, opened 2022-04-15
Improve macro checking if option given more than once

PaulWessel:
We currently have this check for all options that can only be set once:

```
case 'A':
	n_errors += gmt_M_repeated_module_option (API, Ctrl->A.active);
	Ctrl->A.active = true;
	break;
```

However, it is easy to forget to set that variable to true when we add new module options. This PR combines those statements so that only

```
case 'A':
	n_errors += gmt_M_repeated_module_option (API, Ctrl->A.active);
	break;
```

will be needed, allowing us to

1. Remove a few hundred lines no longer needed, and
2. Be sure that _active_ will always be set to true in each case.

This PR updates the macro but does not yet remove the (now) superfluous `Ctrl->?.active = true` lines in the modules - that will come next.

Oh, and testing this found a bug in gmtflexure.c that I am fixing here as well.

### DIFF
--- a/src/gmt_error.h
+++ b/src/gmt_error.h
@@ -92,10 +92,18 @@ EXTERN_MSC const char * GMT_strerror (int err);
 
 #define gmt_M_is_verbose(C,level) (MAX(C->parent->verbose, C->current.setting.verbose) >= level)
 
-/* Check condition and report error if true */
+/* Check condition and report error if true - passing GMT as arg */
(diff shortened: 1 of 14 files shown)


seisman (approved):
Looks great!

PaulWessel:
No merging yet as we have some double checks that now fails, e.g. grdcontour:

```
				n_errors += gmt_M_repeated_module_option (API, Ctrl->C.active);
				if (Ctrl->C.active) {
					GMT_Report (API, GMT_MSG_ERROR, "Option -C: Given more than once\n");
					n_errors++;
					break;
				}
				Ctrl->C.active = true;

```

Amazing...
I will fix these first.

PR #6584, opened 2022-04-17
Let psconvert -C be repeatable

### LINKED ISSUE #6583: psconvert -C may be repeate but current parsing does not allows it
Don't know how to fix cases like this with the new macrology.

joa-quim:
Fixes #6583


PaulWessel (approved):
Fine as is but for perfection use `if (Ctrl->C.arg[0]) strcat (Ctrl->C.arg, " ");`

PR #6590, opened 2022-04-18
Release the no args option police for -I

joa-quim:
This police actions cause pain for no clear gain.
```
julia> rose(data, A=20, I=true, Vd=1);
         -I -A20
psrose [ERROR]: Option -I: No argument expected
```
In this case it happens because ``opt->arg`` apparently already holds the output in-memory name.

### DIFF
--- a/src/psrose.c
+++ b/src/psrose.c
@@ -324,7 +324,7 @@ static int parse (struct GMT_CTRL *GMT, struct PSROSE_CTRL *Ctrl, struct GMT_OPT
 				break;
 			case 'I':	/* Compute statistics only - no plot */
 				n_errors += gmt_M_repeated_module_option (API, Ctrl->I.active);
-				n_errors += gmt_get_no_argument (GMT, opt->arg, opt->option, 0);
+				if (!API->external) n_errors += gmt_get_no_argument (GMT, opt->arg, opt->option, 0);	/* Externals already have output name in opt->arg */
 				break;
 			case 'L':	/* Override default labeling */
 				n_errors += gmt_M_repeated_module_option (API, Ctrl->L.active);


joa-quim:
I'm not using nothing different here. All modules use the same generic functions (as I've said in other occasion). When I debugged I saw that ``opt->arg`` is not empty and hence the test fail.

The checks broke a lot of julia tests when options used something like ``D=1`` instead of ``D=true`` and ``D=1`` was converted to ``-D1`` which now causes an error while during (...)years it didn't.

PaulWessel:
Can you tell me where D=true or D=1 becomes -D1? This is in the Julia wrapper?  Then you should be able to fix that on the Julia side I think.  We just had this chat about Kurt's crazy scheme of testing random input and instead I decided to be stricter with all the parsing we run into this instead.  Funny, no?

joa-quim:
> Can you tell me where D=true or D=1 becomes -D1? This is in the Julia wrapper? 

Yes.

> Then you should be able to fix that on the Julia side I think.

A couple of hours later they are (I think).

> We just had this chat about Kurt's crazy scheme of testing random input and instead I decided to be stricter with all the parsing we run into this instead. Funny, no?

Well, was not so fun. The *fuzzy* testing tests if things that were supposed to be given were not, and later being accessed. Or if they of wrong *nature*, etc. I don't think it can test things that are ignored in first place.

PaulWessel:
We can close this one, right?

joa-quim:
After merging yes. Why you think it's solved?

PaulWessel:
I thought you fixed the bug in Julia that added a 1 after some options or arguments?

joa-quim:
There is no bug in the Julia side. I made the diagnostic (not the true cause) on why it fails in C.

PaulWessel:
> There is no bug in the Julia side. I made the diagnostic (not the true cause) on why it fails in C. Note that there is no **1** here (that was elsewhere)
> 
> ```
> julia> rose(data, A=20, I=true, Vd=1);
>          -I -A20
> ```

So earlier you said that opt->arg is not empty for this case.  Yet psrose.c does nothing with opt->arg as far as I can tell.  Are you sure GMT is doing something with that argument?

joa-quim:
Don't know what ``opt->arg`` contents are used for but I know that because it's not empty the new check breaks it for julia, while the command works if the check is knocked out (as it used to be).

PaulWessel:
So I debugged this in Julia:

```
rose([1 2;2 3], A=20, I=true, Vd=1);
	 -I -A20
psrose [ERROR]: Option -I: No argument expected
ERROR: Something went wrong when calling the module. GMT error number = 72
```

Examining options I find for **-I** this:

debug

I can see that option **-<** (input file) is correctly hooked up to a virtual file (number 4), and the **->** option (output file) is set to receive the statistics to an output virtual file (number 5).  This is correct.

But clearly there is also another output file (number 3) hooked to the **-I** option.  Since it is already there when I get to this call it is happening in your wrapper somewhere.  Looking at the **KEYS** I do see that we have 

`#define THIS_MODULE_KEYS	"<D{,CC(,ED(,>X},>D),>DI,ID)"`

The `>DI` is good: It says that if **-I** is used then we switch output type form the default (X  = PostScript) to D (for dataset).  But then we also have `ID)` which says that -I is an option that specifies a secondary output.  Yet our code does not do this, it already has a primary output that has been changed to DATASET and that is all we need.

So I will make a PR that removes that erroneous KEY - who knows who or when it was added.

PaulWessel:
We would not have found this bug without the new police, so I think it was worth it.

PR #6600, opened 2022-04-20
Add kustom symbol name check

### LINKED ISSUE #6594: S="~d4c:+skarrow-0.4+gblack" crashes Julia
Tried to find why but the debugger landed on ``gmtapi_extract_argument()`` and the argument of ``isdigit`` in
```
*n_pre = (k >= 0 && key[k][K_MODIFIER] && isdigit (key[k][K_MODIFIER])) ? (int)(key[k][K_MODIFIER]-'0') : 0;
```
key[k][K_MODIFIER] = -3

Ofc, the root of the error is somewhere before but can't reach with with debugging from Julia runs taking 20 sec per step. Anyway, it must be at the parsing of *kustom* symbol

PaulWessel:
Two new checks have been added:

1. When a custom symbol is used with decorated lines we require a symbol size to be present.
2. Custom symbol names need to have names that follow standard variable names (alphanumerical characters plus underscore) plus period (symbol file extension) and @ (remote file). For backwards compatibility we also allow -.

Closes #6594 (I hope - please check).

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -15550,6 +15550,8 @@ int gmt_flip_justify (struct GMT_CTRL *GMT, unsigned int justify) {
 bool gmtlib_invalid_symbolname (struct GMT_CTRL *GMT, char *name) {
 	/* Check that a symbol name only contains valid characters,
 	 * which are the alphanumerics plus /, _, @, - and . */
+	char *text = strrchr (name ,'/');	/* Only check name of symbol, not leading directory */
+	if (text == NULL) text = name;	/* No directory slash found - use entire name */
 	for (unsigned int k = 0; k < strlen (name); k++) {
 		if (!(isalnum (name[k]) || strchr ("@_-/.", name[k]))) {
 			GMT_Report (GMT->parent, GMT_MSG_ERROR, "Symbol name %s contains invalid character %c\n", name, name[k]);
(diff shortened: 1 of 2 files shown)


src/gmt_support.c
```diff
@@ -10485,10 +10485,18 @@ int gmtlib_decorate_specs (struct GMT_CTRL *GMT, char *txt, struct GMT_DECORATE
...
-						strncpy (G->size, &s[1], GMT_LEN64-1);
-						s[0] = '\0';	/* Truncate size */
-						strncpy (G->symbol_code, &p[1], GMT_LEN64-1);
-						s[0] = '/';	/* Restore size */
+						if (s == NULL) {	/* No slash, no size */
+							GMT_Report (GMT->parent, GMT_MSG_ERROR, "Option -S~: Custom symbol %s not given any size\n", &p[2]);
+							bad++;
+						}
+						else {
+							strncpy (G->size, &s[1], GMT_LEN64-1);
```
joa-quim, line 10493: If a unix full file name was passed and no size, this will pick the file name as if it was the size, no?

PaulWessel: Yes, looks like it.  I need to add a check if that string is numeric (there is a function for this).

PaulWessel: Check added.

PaulWessel:
FYI, bedtime here.

joa-quim (approved):

PR #6604, opened 2022-04-21
Fix KEYS for rose -I

PaulWessel:
See #6590 for background.  This updates the KEYS which hopefully fixes this; closes #6590.


joa-quim (approved):
I had looked into that KEY but looked fine to me.

Issue #6598, opened 2022-04-20
Failing tests due to recent changes


seisman:
The following three tests fail recently, likely due to recent changes in argument parsing:
```
1/3 Test #358: test/gmtmath/lsfit.sh ............***Failed    0.19 sec
Set GMT_SESSION_NAME = 68820
Error:  [ERROR]: Option -A: Cannot read file !
ERROR: gmtest:90
memtrack errors: 0
exit status: 1

2/3 Test #102: doc/scripts/GMT_hinge.sh .........***Failed    0.31 sec
Set GMT_SESSION_NAME = 68818
Error: r [ERROR]: Unrecognized option -W
ERROR: /home/runner/work/gmt/gmt/build/test/gmtest:90
memtrack errors: 0
exit status: 1

3/3 Test  #47: doc/scripts/GMT_CPTscale.sh ......***Failed    0.33 sec
Set GMT_SESSION_NAME = 68817
Error: r [ERROR]: Unrecognized option -W
ERROR: /home/runner/work/gmt/gmt/build/test/gmtest:90
memtrack errors: 0
exit status: 1
```

seisman:
The two colorbar tests will be fixed by PR #6602.

So the only remaining failing test is `test/gmtmath/lsfit.sh`

```
1/3 Test #358: test/gmtmath/lsfit.sh ............***Failed    0.19 sec
Set GMT_SESSION_NAME = 68820
Error:  [ERROR]: Option -A: Cannot read file !
ERROR: gmtest:90
memtrack errors: 0
exit status: 1
```

PaulWessel:
I am unable to reproduce that error.  Is this the line ?

`gmt math -A@stepdata.txt+r -N4/1 -C0 1 ADD -C1 T ADD -C2 0.5 STEPT ADD -Ca LSQFIT = solution.txt`

Runs fine for me...

seisman:
> I am unable to reproduce that error. Is this the line ?
> 
> `gmt math -A@stepdata.txt+r -N4/1 -C0 1 ADD -C1 T ADD -C2 0.5 STEPT ADD -Ca LSQFIT = solution.txt`
> 
> Runs fine for me...

This command with `-Vd` gives me the error:
```
gmtmath [DEBUG]: Look for file /home/user/gmtmath.macros
gmtmath [DEBUG]: Look for file /home/user/.gmt/gmtmath.macros
[...]

PR #6606, opened 2022-04-21
Must allow colon in custom symbol paths on Windowa

PaulWessel:
This checker added yesterday forgot about C:\path ...
Partly addresses #6598 for Windows I hope (please let me know @seisman).

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -15555,11 +15555,17 @@ int gmt_flip_justify (struct GMT_CTRL *GMT, unsigned int justify) {
 
 bool gmtlib_invalid_symbolname (struct GMT_CTRL *GMT, char *name) {
 	/* Check that a symbol name only contains valid characters,
-	 * which are the alphanumerics plus /, _, @, - and . */
+	 * which are the alphanumerics plus /, _, @, - and . (and : for Windows) */
+#ifdef WIN32
+#define CUSTOM_EXTRA_CHAR "@_-/.:"
+#else
+#define CUSTOM_EXTRA_CHAR "@_-/."
+#endif
+
 	char *text = strrchr (name ,'/');	/* Only check name of symbol, not leading directory */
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -4,7 +4,7 @@
 name: Tests
 
 on:
-  # pull_request:
+  pull_request:
```
seisman, line 7: Let's wait for 30 minutes to see if this PR fixes the bug. 
Need to apply the suggested change below before merging.
```suggestion
  # pull_request:
```

seisman:
@PaulWessel Unfortunately, I still see the same errors on Windows CI.

PaulWessel:
Hm, OK.  if it still complains about a colon after I added

```
#ifdef WIN32
#define CUSTOM_EXTRA_CHAR "@_-/.:"
#else
#define CUSTOM_EXTRA_CHAR "@_-/."
#endif
```

then why? Of is it complaining about the backslash now?

seisman:
> Hm, OK. if it still complains about a colon after I added
> 
> ```
> #ifdef WIN32
> #define CUSTOM_EXTRA_CHAR "@_-/.:"
> #else
> #define CUSTOM_EXTRA_CHAR "@_-/."
> #endif
> ```
> 
> then why? Of is it complaining about the backslash now?

Actually the errors are different. Now it complains about `\`:
```
Symbol name C:\Users\RUNNER~1\AppData\Local\Temp/GMT_symbol6296 contains invalid character \
```

PaulWessel:
OK, that makes sense.  New fix.

seisman:
Also need to deal with `~`:
```
Symbol name C:\Users\RUNNER~1\AppData\Local\Temp/GMT_symbol2552 contains invalid character ~
```

seisman (approved):

PR #6608, opened 2022-04-21
Use static instead of GMT_LOCAL

PaulWessel:
In some places we must use static to avoid linker troubles.


seisman (approved):

joa-quim:
How come? GMT_LOCAL is an alias of static!!!!

PR #6609, opened 2022-04-21
Simplify parser checks for only one output file

PaulWessel:
We had lots of cases where we used an _n_files_ counter to count how many times an output file was specified but we now have the new macros that can do this for us and give errors as needed.  This PR saves many lines on this and eliminates unnecessary _n_files_ variables.

Also fixed a few glitches were I forgot to use the macro in the first place.

### DIFF
--- a/src/gmt2kml.c
+++ b/src/gmt2kml.c
@@ -313,7 +313,7 @@ static int parse (struct GMT_CTRL *GMT, struct GMT2KML_CTRL *Ctrl, struct GMT_OP
 
 		switch (opt->option) {
 
-			case '<':	/* Input files */
+			case '<':	/* Input files (we only check one) */
 				if (n_files++ > 0) break;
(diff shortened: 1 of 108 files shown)


seisman (approved):

src/gmt2kml.c
```diff
@@ -313,7 +313,7 @@ static int parse (struct GMT_CTRL *GMT, struct GMT2KML_CTRL *Ctrl, struct GMT_OP
 
 		switch (opt->option) {
 
-			case '<':	/* Input files */
+			case '<':	/* Input files (we only check one) */
 				if (n_files++ > 0) break;
```
seisman, line 317: Should this line be updated?

PaulWessel: Thanks, yes, now fixed.
Also found lots of ;; when there should just be one.

PR #6616, opened 2022-04-23
Better check when a column is used for both coordinates and dimension

PaulWessel:
See https://github.com/GenericMappingTools/pygmt/issues/1892#issuecomment-1107375548 fro background.

The issues was that a physical column was used both for an x-coordinates and a dimension via repetition in **-i.**  Because x came first this prevented the text from being converted to internal inches and hence the rectangles major axis was not converted from cm to inches and was too long.

This PR recognizes this situation and keeps track if the initial decode was not in the context of dimension and then adjust the result when the same column is used in other context where dimension is expected.

Seems to work, and other tests pass.

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -3815,7 +3815,7 @@ void gmt_set_column_types (struct GMT_CTRL *GMT, unsigned int n_cols_start, bool
 	 * to affect. */
 	unsigned int j;
 
-	if (GMT->common.i.active) {	/* Must figure out correct pre-shuffle columns for the non-dimensional ones */
+	if (GMT->common.i.select) {	/* Must figure out correct pre-shuffle columns for the non-dimensional ones */
 		unsigned int i_col, pre_col;
 		/* First set them all to be dimensional */
 		for (j = n_cols_start; j < GMT->common.i.n_cols; j++) {	/* All columns beyond x,y and possibly z */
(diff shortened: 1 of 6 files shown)


seisman (approved):

seisman:
@PaulWessel do you think we should add a test?

PaulWessel:
Can you create a CLI version  from her full case?

PaulWessel:
Hi @seisman, is this something you think you can do or is it too much to translate to CLI?

seisman:
This is a script to reproduce the first example in the original issue report. I think this one should be enough to catch the bug:
```
gmt begin
	gmt set FONT_TITLE 12p
	gmt makecpt -Croma -T0/10/1
	gmt subplot begin 2x3 -Fs10c -R0/10/0/10 -Ba1f1g1 -Sc -Sr -M0.5c/0.5c
		for i in 0 1 2 3 4 5; do
			echo 4 5 2 0 3 1 | gmt plot -Sj -C -W0.5p -i${i},1,2,3,4,5 -B+t"-i${i},1,2,3,4,5" -c
		done
	gmt subplot end
gmt end

```

PaulWessel:
OK, I have modified and added your script.

PaulWessel:
And added plot to dvc.

seisman (approved):

PR #6625, opened 2022-04-26
gmt --show-library gave NULL in Xcode

PaulWessel:
The library path for public consumption was never set in Xcode.  Did not seem to have any consequence except when I tried to run `gmt --show-library` inside Xcode.

This PR builds the library path if nothing is set at the end of the checks.


seisman (approved):

PaulWessel:
@joa-quim I don't think this will help us find _earthtide_ etc but just in case.

PR #6626, opened 2022-04-26
Must check for label content before offsetting

PaulWessel:
In the magnetic map rose case, we examine if there are labels to plot (hence we need to do more offsets) but that test was flawed.  It checked if the pointer existed (it always does) instead of its content (may be empty).

This PR fixes that bug.  It is related to the issues in #6213 but probably not the last word.  That may be a mix of this bug (now fixed) and design decisions since the default offset still looks a wee bit too large?


PaulWessel:
The same disease afflicted the directional rose offset calculation.  I have also adjusted the amount of offset to be more suitable.  This makes the mag_roses.sh file but I can fix that once you agree this is a better design.

seisman (approved):

PR #6629, opened 2022-04-26
Prevent both -D and -> at the same time in gmtconvert

PaulWessel:
These are mutually exclusive options.  Once checked we do not need the freeing of Out.file later since it is unset.


src/gmtconvert.c
```diff
@@ -451,6 +451,8 @@ static int parse (struct GMT_CTRL *GMT, struct GMTCONVERT_CTRL *Ctrl, struct GMT
 	                                 "Only one of -Q and -S can be used simultaneously\n");
 	n_errors += gmt_M_check_condition (GMT, Ctrl->N.active && Ctrl->F.active,
 	                                 "The -N option cannot be used with -F\n");
+	n_errors += gmt_M_check_condition (GMT, Ctrl->Out.active && Ctrl->D.active,
+	                                 "The -D option cannot be used with ->\n");
```
seisman, line 455: Users may have no idea what `->` means.

PaulWessel: True, but only bugs in external wrappers should get us there.  What should I say instead? Option -D: Cannot also specify output on command line ??

PR #6644, opened 2022-04-29
Enhance legend -M scenarios

PaulWessel:
This PR adds optional directives to **-M** that allows us to indicate the order and selection of the two possible input sources (hidden and explicit) in modern mode.  Not used nor allowed in classic mode.


joa-quim (approved):

seisman (approved):
Looks good, but I think you forgot to add the PS file to DVC.

PaulWessel:
Yes, will do next.

Issue #6639, opened 2022-04-28
surface -Ll explodes


PaulWessel:
If specifying a lower level constant when gridding with **surface**, via option **-Ll,** the solution honors the lower level but grows in a crazy way.  If the upper level is set instead via **-Lu** it is honored as expected and nothing crazy happens at the lower end. E.g., using the table data we use for many tests in GMT (with data ranging from 690 60 960):

`gmt surface @Table_5_11.txt -R0/6.3/0/6.3 -I0.1 -GLl_limits2.grd -Ll750`

![6](https://user-images.githubusercontent.com/26473567/165837342-3e78775c-a512-49e4-99fa-9622d35eafb4.png)

If I use gmt4 with the same (local) file and options I get the expected plot:

![4](https://user-images.githubusercontent.com/26473567/165837486-391e1bfb-0dca-4e8a-9649-c80980ffc727.png)

Because we (a) have very few tests for surface in general and none for -L, and (b) I have not used that option myself much and (c) neither have others, it seem since we have not received any bug report, this issue could have been there for some time - most likely during the major GMT5 porting.  I will have to compare and see what happens and why only for the lower bound.

PaulWessel:
Must be a new bot. Hate it when I have to show my ID to the new bot so they know who they are talking to...

joa-quim:
You must somehow be violating de CC. 😈

PaulWessel:
It is the norwegian cursing picked up by my microphone.

joa-quim:
See? A clear violation.

PaulWessel:
Just a note that this is not limited to **-Ll**.  If I change the sign of the data and level and use `-Lu-750` as upper limit the same thing happens (and not in GMT 4).
Furthermore, the node used to access the lower boundary grid was not correct and after (trying to) fix that the blow-up was less but still blow -up.  So the theory so far is that there are still a bug in finding the right bound node.  Why a grid you say when I am using a constant in -L? Well, part of the processing involves removing the best-fitting plane from the data so we must also do that to the constraints, hence the 750 constant becomes a plane as well and hence a grid.

PR #6645, opened 2022-04-29
Fix surface -L bug

PaulWessel:

See #6639 for background.  The issue had to do with some of the trickiness of the porting from GMT 4 to 5.  **surface** v4 was unusual in that its innards were so old the the internal grid was column based, like in fortran (which the ancestor to surface was coded in), whereas all other modules plus grid files are scanline oriented.  The section that checks if we need to impose a limit (upper or lower) had to deal with this mess as the limit grids were row-oriented while the internal surface grid was column based.  So we had a fun little node-number calculation to transpose from the data node to the limit node.  When I ported that to GMT 5 I rewrote **surface** to internally use row-orientation and thus I must have assumed I did not need to do this gymnastics anymore.  But I forgot there were two reasons for the node calculation.  One was the transpose (which no longer is needed).  However, the other is that we are doing multi-grid so early on we are stepping in increments larger than 1 over a smaller virtual grid.  Yet, the lower and upper bound grids only exist in final dimensional form, so we just scale up the row and column indices by the increment spacing to get the equivalent node for those grids.  That was the part that got lost, and hence we looked up the wrong nodes, and in this case some of those nodes were outside the grid and not set, and drove the surface crazy (and me, a bit).

The fix was simple once the brain understood the problem.  I have added a new test (_limits.sh_) that exercises the upper and lower constraint for both a constant and a grid and all looks fine.

### DIFF
--- a/src/surface.c
+++ b/src/surface.c
@@ -216,7 +216,7 @@ struct SURFACE_INFO {	/* Control structure for surface setup and execution */
 	struct SURFACE_BRIGGS *Briggs;	/* Array with Briggs 6-coefficients per nearest active data constraint */
 	struct GMT_GRID *Grid;		/* The final grid */
 	struct GMT_GRID *Bound[2];	/* Optional grids for lower and upper limits on the solution */
-	struct GMT_GRID_HEADER *Bh;		/* Grid header for one of the limit grids [or NULL] */
+	struct GMT_GRID_HEADER *Bh;	/* Grid header for one of the limit grids [or NULL] */
 	struct SURFACE_SEARCH info;	/* Information needed by the compare function passed to qsort_r */
 	unsigned int n_factors;		/* Number of factors in common for the dimensions (n_rows-1, n_columns-1) */
 	unsigned int factors[32];	/* Array of these ommon factors */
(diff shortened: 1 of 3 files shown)


joa-quim (approved):

seisman:
This PR makes the following test crash
```
$ gmt surface @Table_5_11_mean.xyz -I1 -R0/4/0/8 -Gabc.nc
surface [WARNING]: 1 unusable points were supplied; these will be ignored.
surface [WARNING]: You should have pre-processed the data with block-mean, -median, or -mode.
surface [WARNING]: Check that previous processing steps write results with enough decimals.
surface [WARNING]: Possibly some data were half-way between nodes and subject to IEEE 754 rounding.
ERROR: Caught signal number 11 (Segmentation fault: 11) at
0   libgmt.6.4.0.dylib                  0x000000010ab8ad0a surface_iterate + 362
1   ???                                 0x00000000000003b4 0x0 + 948
Stack backtrace:
0   libgmt.6.4.0.dylib                  0x000000010a8aed59 sig_handler_unix + 217
1   libsystem_platform.dylib            0x00007ff81de48e2d _sigtramp + 29
2   ???                                 0x0000000300580000 0x0 + 12890669056
3   libgmt.6.4.0.dylib                  0x000000010ab87403 GMT_surface + 22019
4   libgmt.6.4.0.dylib                  0x000000010a7b3534 GMT_Call_Module + 2180
5   gmt                                 0x000000010a1503c8 main + 1816
6   dyld                                0x000000011348b4fe start + 462
```

Issue #6587, opened 2022-04-18
[v6.4.0] grd2kml fails


seungsep:
**Description of the problem**
grd2kml fails to produce KML files with the error message associated with grdfilter. 
But the same command line with grd2kml v6.3.0 works out with no issues.

**Full script that generated the error**

```
gmt grd2kml t.nc -Ntest -T"test" -I+d
```

**Full error message**

```
grd2kml [NOTICE]: Extended grid size is by 1024 by 1024, tile size is 512 x 512
grdfilter [ERROR]: Option -D: Input grid is geographic but your distance mode is Cartesian
```

**System information**

* Operating system: OS Big Sur 11.6.5
* GMT version (`gmt --version`): 6.4.0_2a0ffbc_2022.04.18

PaulWessel:
Could you share the output of grdinfo t.nc?

seungsep:
Here it is~

t.nc: Title: z
t.nc: Command: grdsample /tmp/74153.nc -I0.2s+e -Gt.nc
t.nc: Remark: 
t.nc: Pixel node registration used [Geographic grid]
t.nc: Grid file format: nf = GMT netCDF format (32-bit float), CF-1.7
t.nc: x_min: 184.590444225 x_max: 184.62838867 x_inc: 5.55555555555e-05 name: longitude n_columns: 683
t.nc: y_min: -20.5563741781 y_max: -20.5315408448 y_inc: 5.55555555556e-05 name: latitude n_rows: 447
t.nc: v_min: -0.30519798398 v_max: 154.343490601 name: z
t.nc: scale_factor: 1 add_offset: 0
t.nc: format: netCDF-4 chunk_size: 137,149 shuffle: on deflation_level: 3

PaulWessel:
Thanks, was able to reproduce and learned a lot, see related PR.

seungsep:
Thanks for checking out this issue!
However, I still have the same issue...

Pulled version: 6.4.0_81d3933_2022.05.02

Followings are the error message after running `gmt grd2kml t.nc -Ntest2 -I+d -Vl`

```
grd2kml [NOTICE]: Extended grid size is by 1024 by 1024, tile size is 512 x 512
grd2kml [INFORMATION]: Level 0: Factor = 2 Dim = 1024 x 1024 -> 512 x 512
grd2kml [INFORMATION]: Level 0: Low-pass-filtering the grid(s)
[...]

PR #6646, opened 2022-04-30
Fix some issues related to geo vs cart

PaulWessel:
This PR started as a fix to the problem reported via #6587, which was due to a recently imposed check in **grdfilter** that one should not supply Cartesian distance computations with geographic grids.  Because in this particular module (**grd2kml**) I actually need Cartesian treatment I thought I could just add **-fc** to fix this. But no, then we had some trickle-down issues.

Basically, users have always had the option of using **-f** to override what a grid file says it is.  Typically, one may read a Cartesian grid that perhaps should be seen as geographic, and then using **-fg** makes it so (this is for modules where there is no mapping i.e., **-J** is not used).  However, the opposite turned out not to be true: If you have a geographic grid but wanted it to be treated as a Cartesian grid then **-fc** did _not_ do that since **-fc** is parsed first and then when the geographic grid is read we recognize it as such and basically **-fg** is set.

The solution is to _remember_ that we set **-fg** or **-fc** and then apply it.  This PR checks at the end of _gmtapi_import_grid_ to see if we need to impose one of them, as well as the start of _gmtapi_export_grid_ to see if a change should apply to the output.  Great, simple as that.

Now I got failures from all the **grdflexure** tests! In debugging one of them I learned that _GMT_Get_Values_, which is this clever API function that understands units, temporarily may set **-fg** but then leave it as is when it returns.  So in **grdflexure**, after parsing `-E5k`, the **k** for km triggered **-fg** (even if a 5 km plate has nothing to do with geographic data) and then the 0/600/0/600 region was seen as too big for geographic...

WIth these fixes:

1. Add memory to **-fc** and **-fg**
2. Apply that memory just before writing or just after reading grids
3. Reset col types and **-f** settings to what they were when _GMT_Get_Values_ is used
4. Pass the **-fc** flag to **grdfilter** from **grd2kml**

 all seems OK again and the original problem is fixed.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -5532,6 +5532,13 @@ start_over_import_grid:		/* We may get here if we cannot honor a GMT_IS_REFERENC
 
 	if (done) S_obj->status = GMT_IS_USED;	/* Mark as read (unless we just got the header) */
 
+	/* Some grids are geographic and others are Cartesian, but the user may wish to flip this fact
+	 * via -f. Here we catch such deviousness and set the desired input geo/cart mode imposed by -f. */
+	if (GMT->common.f.is_geo[GMT_IN])
+		gmt_set_geographic (GMT, GMT_IN);
+	else if (GMT->common.f.is_cart[GMT_IN])
+		gmt_set_cartesian (GMT, GMT_IN);
+
 	return (G_obj);	/* Pass back out what we have so far */
 }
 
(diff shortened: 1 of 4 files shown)


PaulWessel:
Any concerns with this, @joa-quim ?

joa-quim (approved):
See why I keep warning you about the useless of blind approvals?

PR #6652, opened 2022-05-01
Let movie set an environmental hint we are building frames

PaulWessel:
As I was making an animation I got annoying messages and wrong decisions being made by GMT: 

> Changing to PostScript landscape orientation based on your plot and paper dimensions, but we cannot be 100%% sure.
> Use PS_MEDIA and/or PS_PAGE_ORIENTATION to specify correct paper dimensions and/or orientation if our guesses are inadequate.

This should never apply when building a movie since we fix the canvas.  Because movie scripts are independent the only solution is to set an environmental parameter (GMT_MOVIE=YES) and check that in gmt_plot_init before we rotate the plot.  Since it is the movie module that does this once (and removes it at the end) it should be relatively stable.

### DIFF
--- a/src/movie.c
+++ b/src/movie.c
@@ -1595,6 +1595,15 @@ EXTERN_MSC int GMT_movie (void *V_API, int mode, void *args) {
 	}
 	fclose (fp);	/* Done writing the init script */
 
+	/* Because when a movie is plotted we do not want gmt_plot_init to contemplate orientation changes
+	 * we help that decision by setting this environmental parameter.  We unset it at the end. */
+#ifdef WIN32
+	if (_putenv ("GMT_MOVIE=YES"))
+#else
+	if (setenv ("GMT_MOVIE", "YES", 1))
+#endif
+		GMT_Report (API, GMT_MSG_WARNING, "Unable to set GMT_MOVIE=YES to inform GMT that movie scripts will be building a movie\n");
+
 	if (Ctrl->S[MOVIE_PREFLIGHT].active) {	/* Create the preflight script from the user's background script */
 		/* The background script must be modern mode */
 		unsigned int rec = 0;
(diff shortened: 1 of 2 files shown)


src/movie.c
```diff
@@ -2630,20 +2678,31 @@ EXTERN_MSC int GMT_movie (void *V_API, int mode, void *args) {
...
 		GMT_Report (API, GMT_MSG_ERROR, "Unable to delete the cleanup script %s.\n", cleanup_file);
-		Return (GMT_RUNTIME_ERROR);
+		error = GMT_RUNTIME_ERROR;
+		goto out_of_here;
 	}
 
-	Return (GMT_NOERROR);
+out_of_here:
+	/* Remove the environmental reminder we are building a movie */
+#ifdef WIN32
```
joa-quim, line 2700: I never know but think it's safer to use _WIN32 then WIN32. Latter may also be used by Cygwin.

PaulWessel: Well, I copy/pasted this from somewhere else so if you would like to go through all of those and advice on which ones should be _WIN32 then be my guest.  I think in 2022 the burden lies on those 7 Cygwin uses to show problems.  As long as it is Windows compliant I think we are OK for now.

joa-quim: The shit is that I don't know. I know that ``_WIN32`` is VisualStudio for sure and that MinGW also defines it. Also seem to understand that MinGW is a mix of VS and GCC libs ... and here starts the *bordel*

PaulWessel: It is pretty harmless.  I realize the reason I got that error was that I used the wrong -R in a place so what I was plotting was indeed exiting the frame. But even so we never want the movie frame to rotate...

joa-quim (approved):

PR #6658, opened 2022-05-02
Tolerate Windows paths when looking for colon separator

PaulWessel:
The syntax for setting quoted and decorated lines both use a colon to separate the directives from optional modifiers.  Since the colon under Windows may be part of a hard path (e.g., C:\my\path\file) we must consider this when looking for the right colon.  This PR will, under WIN32, only recognize the colon in position j if there is at least one more character and that one is not a slash or backslash, and that the previous character is not a letter.  This should skip d:\path is a contender.

This section happens twice in gmt_init.c with a slight variation of initialization so this PR also delegates this search to a local function that is now called in both places.

Will required @joa-quim to test this under Windows.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -16406,7 +16406,7 @@ GMT_LOCAL int gmtinit_colon_pos (struct GMT_CTRL *GMT, char *text) {
 #else
 		if (text[j] == ':') colon = j;
 #endif
-		return (colon);
+	return (colon);
 }
 
 #define GMT_VECTOR_CODES "mMvV="	/* The vector symbol codes */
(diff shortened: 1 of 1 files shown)


joa-quim (approved):
Yeap, now it works. Thanks.
```
julia> plot(r, decorated=(locations=a, symbol=(custom="arrow", size=0.3), fill=:black, dec2=true), show=1, Vd=1)
        psxy  -JX14c/0 -Baf -BWSen -R-10.4/10/-11/11 -S~f:+skC:\Users\joaqu\.julia\dev\GMT\src\..\share\custom\arrow/0.3+gblack -P -K > C:\TEMP\GMTjl_tmp.ps
```

PaulWessel:
Thank Bill for introducing the colon to the world - nothing but pain in the butt.

PR #6661, opened 2022-05-03
Simplify colonoscopy on Windows

PaulWessel:
The solution implemented in #6658 was not correct and too obtuse.  This PR breaks down the checks to dispense with the easy checks up front and save the more involved one for when it is needed.
Hopefully fixes the issue - let the CI begin.

Note: I have only compiled this in Windows in my head...

Also, the second reference to got_slash is not needed (since we only get past the || if got_slash is true anyway) but I left it there for clarity.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,7 @@
 name: Tests
 
 on:
-  # pull_request:
+  pull_request:
   push:
     branches:
       - master
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -16409,7 +16409,7 @@ int gmtlib_colon_pos (struct GMT_CTRL *GMT, char *text) {
 		bool got_slash;
 		if (text[j] != ':') continue;	/* Not a colon, moving on */
 		if (text[j+1] == '\0') continue;	/* End of argument, moving on */
-		got_slash = !strchr ("/\\", text[j+1]);	/* True if next letter is a slash */
+		got_slash = (strchr ("/\\", text[j+1]) != NULL); /* True if next letter is a slash */
 		if (!got_slash || (got_slash && !(isupper (text[j-1]) || islower (text[j-1])))) colon = j;
 #else
 		if (text[j] == ':') colon = j;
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -4,7 +4,7 @@
 name: Tests
 
 on:
-  # pull_request:
+  pull_request:
```
seisman, line 7: I've enabled the tests in this PR temporarily. Remember to revert this change before merging.
```suggestion
  # pull_request:
```

PaulWessel: Looks like those failures are gone.  The other failures are unrelated to this PR.

PaulWessel:
This PR fixes the failures reported on WIndows.

seisman:
> This PR fixes the failures reported on WIndows.

https://github.com/GenericMappingTools/gmt/runs/6271377114?check_suite_focus=true

I think the tests still fail on Windows.

PaulWessel:
How very odd - I just clicked on the link above and found just 3-4 other errors...

PaulWessel:
So before I added your "turn off the full test" commit I clicked on the failing Windows link and found none of the previous errors.  That is why I added in your suggestion and of course that rebuilt the limited tests wich passed.  How is this possible?

seisman:
> How very odd - I just clicked on the link above and found just 3-4 other errors...

I think what you saw are the tests on macOS.

PaulWessel:
OK, now would be a good time for you to run one of these in the debug and tell me why it fails since it seems to me it should not...

PaulWessel:
Pretty sure I clicked the Windows failing link but hard to prove now.  Let's see if @joa-quim can give some insights here first before we reset to full testing.

PaulWessel:
@joa-quim, put your rubber gloves on and get in that colon!

joa-quim:
``got_slash`` was failing. Make it
``got_slash = (strchr ("/\\", text[j+1]) != NULL);	/* True if next letter is a slash */``

PaulWessel:
OK, updated, please run tests

joa-quim:
Tests pass now.

joa-quim (approved):

PR #6663, opened 2022-05-03
Make sure key[k][K_MODIFIER] is > 0 before sending it to isdigit()

joa-quim:
When in debug mode I get crashes on this line because ``key[k][K_MODIFIER] = -3`` and an ``assert`` in VS's ``isdigit`` make it crash. This is strange because it seems there are times where ``key[k][K_MODIFIER]`` is not yet defined and we are checking it's value but in release mode it gets a value of 0 and the test passes.


PaulWessel (approved):
I think this should be fine.

PR #6664, opened 2022-05-03
Demux output grid from grdfft before returning in memory

### LINKED ISSUE #6660: grdfft doesn't work on externals
Result grid is interspersed with columns of zeros
```
G = grdfft("@earth_relief_10m", F="3000000/2000000/500000/300000", N="a+a", f="g");
imshow(G, shade=true)
```
GMTjl_tmp-fs8

PaulWessel:
See #6660 for background. Normally, the demux takes place in the writing section but here we simply want to return the result to the calling environment as a grid.  Thus we must demux the grid first.

WIth this PR the example in #6660 yields

GMTjl_tmp

Closes #6660.


PaulWessel:
One remaining issue is that this internal grid is twice the size since it was allocated in grdfft to hold both the real and imaginary components.  I think after duoing the demux we should also (if memory is owned by GMT):

1. Realloc the grid back to regular size
2. Remove the complex_mode bits from the header

We should also try this example with passing in a grid from Julia as GMT_IS_REFERENCE since internally I think (hope) there will be a change to DUPLICATE since we must double the grid and we should not have permission to do that on Julia memory.  So, @joa-quim, please

1. CHeck that the current PR works for your example
2. SUpply a Julia example that passes the grid into grdfft instead of via file.
3.

joa-quim:
This makes a similir plot to yours and answers both of your questions.

```
G = gmtread("@earth_relief_10m");
Gf = grdfft(G, F="3000000/2000000/500000/300000", N="a+a", f="g");
imshow(Gf, shade=true)
```

joa-quim (approved):

PaulWessel:
I am thinking perhaps grdfft (and probably gravfft and grdseamount) should do this and not pass this into the GMT_Write_Data function.

PaulWessel:
I am working on two changes but today I am headed to Cambridge so more on this tomorrow I hope.

- 100% of GMT usage and probably most other use one can think of expects the final inverse FFT to yield a real grid (no more imaginary values stored anymore).  Turns out GMT returned the real grid interspersed with the (now zero) imaginary components, and then the belly of _gmtapi_export_grid_ took care of the demuxing.  I have now modified _GMT_FFT_ so that unless a new flag is passed (**GMT_FFT_NO_DEMUX**) it will take care of the demux and eliminate the space used for the imaginary values.  This means any export (i.e., via externals) should immediately work like for any other grid.
- In the event some developer or a future module would need to pass **GMT_FFT_NO_DEMUX** then we need a way to eventually do the demux.  Thus, I am introducing _GMT_FFT_Reset_ which will do that step separately and can then be called when needed.  It does nothing unless the _complex_mode_ flag in the grid header says it is a complex grid.

In this PR, **grdfft**, **gravfft** and **grdflexure** all get real grids back and I have removed the special flag on output via _GMT_Write_Grid_ that says to do the demux (since that is not supported in the external exporters).  You can try this branch - should be OK - I think I ran it with the Julia example.

PaulWessel:
I updated the API to mention GMT_FFT_Reset but I think all is well as is.  It is just there in case anyone ever wants to not get a real grid back from the inverse FFT>  All tests pass, and @joa-quim has already approved.  So I will wait a few hours and then merge this one.

doc/rst/source/devdocs/api.rst
```diff
@@ -3323,6 +3325,10 @@ version of your data. The FFT is fully normalized so that calling
 forward followed by inverse yields the original data set. The information
 passed via ``K`` determines if a 1-D or 2-D transform takes place; the
 key work is done via ``GMT_FFT_1D`` or ``GMT_FFT_2D``, as explained below.
+**Note**: When ``direction``is ``GMT_FFT_INV`` we will remove the space for
```
seisman, line 3328:
```suggestion
**Note**: When ``direction`` is ``GMT_FFT_INV`` we will remove the space for
```

seisman (approved):
Just a typo in the docs.

PR #6667, opened 2022-05-04
docs: Make all options clickable

seisman:

In this PR, I plan to replace option flags like `**-E**` with `|-E|`, so that these option flags are clickable and are linked to the corresponding option description.

The replacements are done automatically using the following command:
```
sed -E -i"" 's/ \*\*(-[A-Z])\*\*([ ,.])/ |\1|\2/g' grdtrack.rst
```
The command doesn't cover all cases but should work for most cases.

The new documentation looks like:
image


joa-quim:
Having all options clickable is obviously a good thing (after all the |...| macro was created exactly for that). If that is achievable by a batch or manually seems secondary to me (there aren't many missing, I hope).

PaulWessel (approved):

anbj:
You've probably seen it, but;

I think the **-A** is missed since there is no (white)space after it.
doc-snip
The **-G** is missed since it ends with an parenthesis.
doc-snip2
;

seisman:
> You've probably seen it, but;
> 
> I think the **-A** is missed since there is no (white)space after it.

Yes, I'll open a separate PR working on these edge cases.

PR #6678, opened 2022-05-06
Add grid sampling as new algorithm to grdfill

PaulWessel:
This PR introduces **-Ag**_grid_ as a new algorithm, which will sample the given _grid_ at the nodes with NaN in the main input grid.  I have added a test that demonstrates it works well.  I realized this could be useful in working with David Sandwell on trying to speed up his 5760 x 6912 **surface** commands in GMTSAR - this algorithm is a partial answer for filling holes.  Test output looks like this - also exercises Roman numerals in **subplot**:

gridfill

i) original data, ii) mask,iii) masked data, iv) same with image-masking, v) filled grid, vi) differences

### DIFF
--- a/src/grdfill.c
+++ b/src/grdfill.c
@@ -110,7 +110,7 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 		"No grid fill takes place and -G is ignored. "
 		"Optionally, append p to write polygons corresponding to these regions.");
 	GMT_Usage (API, 1, "\n-N<value>");
-	GMT_Usage (API, -2, "Set alternate node <value> to indicate a on input hole [Default looks for NaN-nodes].");
+	GMT_Usage (API, -2, "Set alternate node <value> to indicate a hole [Default looks for NaN-nodes].");
 	GMT_Option (API, "R,V,f,.");
 
 	return (GMT_MODULE_USAGE);
--- a/src/grdfill.c
+++ b/src/grdfill.c
@@ -491,13 +491,78 @@ GMT_LOCAL void grdfill_nearest_interp (struct GMT_CTRL *GMT, struct GMT_GRID *In
 	gmt_M_free (GMT, ys);
 }
 
+GMT_LOCAL int grdfill_sample (struct GMT_CTRL *GMT, struct GMT_GRID *In, struct GMT_GRID *Out, char *file) {
+	/* 1. Create x,y arrays of NaN locations in the grid and an empty z array
+	 * 2. Call grdtrack with virtual input and virtual output
+	 * 3. Fill in the holes in the grid with these values
+	 */
+	openmp_int row, col, k;
+	uint64_t dim[4] = {0, 0, 0, 0}, node, n_NaNs = 0, n_alloc = 0;
+	char input[GMT_VF_LEN] = {""}, output[GMT_VF_LEN] = {""}, args[GMT_LEN256] = {""};
+	double *data[3] = {NULL, NULL, NULL};
+	struct GMT_VECTOR *V_in = NULL, *V_out = NULL;
+	struct GMTAPI_CTRL *API = GMT->parent;	/* Shorthand */
+
+	/* Build array list of node coordinates where we have NaNs */
+	gmt_M_grd_loop (GMT, In, row, col, node) {
+		if (!gmt_M_is_fnan (In->data[node])) continue;	/* Not part of a hole */
+		if (n_NaNs >= n_alloc) {	/* Allocate more memory (including the first time) */
+			if (n_alloc == 0) n_alloc = GMT_INITIAL_MEM_ROW_ALLOC; else n_alloc *= 2;
+			for (k = 0; k < 2; k++) data[k] = gmt_M_memory (GMT, data[k], n_alloc, double);
+		}
+		/* Keep the coordinates of the NaN node */
+		data[GMT_X][n_NaNs] = In->x[col];
+		data[GMT_Y][n_NaNs] = In->y[row];
+		n_NaNs++;	/* Number of NaNs so far */
+	}
+	if (n_NaNs == 0) {	/* Well, that was a waste of our time */
+		GMT_Report (API, GMT_MSG_WARNING, "There are no NaNs in your grid - returning the input grid as output\n");
+		for (k = 0; k < 2; k++) gmt_M_free (GMT, data[k]);
+		return (GMT_NOERROR);
+	}
+	/* Finalize the memory allocation (and z for the first time) */
+	for (k = 0; k < 3; k++) data[k] = gmt_M_memory (GMT, data[k], n_NaNs, double);
+	/* Allocate two vector containers for input and output separately */
+	dim[0] = 2;	/* Want two input columns but let length be 0 - this signals that no vector allocations should take place */
+	if ((V_in = GMT_Create_Data (API, GMT_IS_VECTOR, GMT_IS_POINT, 0, dim, NULL, NULL, 0, 0, NULL)) == NULL) return (API->error);
+	dim[0] = 3;	/* Want three output columns [we will share the first two with input] */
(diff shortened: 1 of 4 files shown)


test/grdfill/gridfill.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# Test the -Ag algorithm in grdfill
+# Pull out a small piece of 20m DEM
+gmt grdcut -R0/10/0/10 @earth_relief_20m -Gdata.grd
+# Make a mask for two holes
+gmt grdmath -Rdata.grd 4 6 SDIST 250 GE 0 NAN 7 1.5 SDIST 100 GE 0 NAN ADD 2 DIV = mask.grd
+# Combine to get data with the holes
+gmt grdmath data.grd mask.grd MUL = holes.grd
```
meghanrjones, line 8: @willschlitzer asked about a sample dataset for demonstrating grdfill usage in pygmt earlier today. If it's small enough, I think we should put `holes.grd` with a slightly more descriptive name (e.g., `earth_relief_20m_holes.nc`) in the cache for broader use in examples/tests, rather than generating it each time inside this test. This would be helpful for our efforts to make the gmt examples fully reproducible.

PaulWessel: I have no problem with this.  I can comment out the commands that created that file we put in the cache. OK?

meghanrjones: Sounds good

PaulWessel: So the script is still showing the mask and the raw data so there is no savings in the shell script... It just means we almost repeat all the calls except the grdmath that multiplies data and mask.  But I understand grdmath is a tricky thing in pyGMT.

meghanrjones: While grdmath is not yet in pygmt, creating holes in grids is easy using other methods. The motivation is more for simplicity for the people reading the examples, so that they don't need to interpret a couple unrelated commands.

PaulWessel: Perhaps i'll just go back to an earlier version that just shows the holed grid and the result - that avoids the other 4 frames that are just, ehm, fillers.

PaulWessel: Done

PaulWessel:
OK, we scaled by the test to make it simpler and stored the holed data set in cache, per @meghanrjones suggestion.  The server has now synced up the cache so the tests can run again. The plot now looks like this:

gridfill

DVC has been updated and tests pass.  I think this is good to go now.

src/grdfill.c
```diff
@@ -106,7 +110,7 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 		"No grid fill takes place and -G is ignored. "
 		"Optionally, append p to write polygons corresponding to these regions.");
 	GMT_Usage (API, 1, "\n-N<value>");
-	GMT_Usage (API, -2, "Set alternate node <value> to indicate a hole [Default looks for NaN-nodes].");
+	GMT_Usage (API, -2, "Set alternate node <value> to indicate a on input hole [Default looks for NaN-nodes].");
```
meghanrjones, line 113:
```suggestion
	GMT_Usage (API, -2, "Set alternate node <value> to indicate a hole [Default looks for NaN-nodes].");
```
not sure why 'on input' was added here

meghanrjones (approved):

seisman:
The grdfill.test fails on Windows:
```
C:\Program Files\GraphicsMagick-1.3.32-Q8\gm.exe compare: image difference exceeds limit (0.0355743 > 0.003).
test/grdfill/gridfill.ps: RMS Error = 0.0356 [FAIL]
memtrack errors: 0
exit status: 1
```
Here is the diff image:
gridfill_diff

Here is the generated PDF file on Windows: [gridfill.pdf](https://github.com/GenericMappingTools/gmt/files/8657624/gridfill.pdf)

seisman:
If it helps you, here is the grid file generated by the CI [new.grd.zip](https://github.com/GenericMappingTools/gmt/files/8659866/new.grd.zip)

PaulWessel:
Certainly an interesting one:

```
gmt grdcut @earth_relief_01d -R0/10/0/10 -Gb.grd
echo 7.5 1.5 | gmt grdtrack -Gb.grd -nn
7.5	1.5	-1169.5
echo 7.5 1.5 | gmt grdtrack -G@earth_relief_01d -R0/10/0/10 -nn
7.5	1.5	-2347.5

```
Since the script is using the last scheme (passing the remote file to **grdfill**) there must be something funny here. BTW, the Windows produced grid gives:

```
echo 7.5 1.5 | gmt grdtrack -Gnew.grd -nn
7.5	1.5	-2701.41601562
```

while my output form the script gives -1169.5.  There is an Island really close to this point so shallow makes sense.  I think the **grdcut** grid is correct so it should be shallow here and for some reason on Windows we are getting another node.

If you or @joa-quim could modify the script to do

```
gmt grdcut @earth_relief_01d -R0/10/0/10 -Gb.grd
gmt grdfill @earth_relief_20m_holes.grd -Agb.grd -Gnew.grd
```

then my guess is it will give the same result as me (and pass the test) and then we know that the passing of remote files to **grdfill** (which passes those args to **grdtrack**) is suspect.

PaulWessel:
Hi @joa-quim, might you be able to replace that **grdfill** command with the two at the end of the previous message and see if that works well for Windows? That way I know what to look for when I debug `gmt grdtrack -G@earth_relief_01d -R0/10/0/10 -nn`

joa-quim:
Not sure I got it. Is this what you asked?
```
gmt grdcut @earth_relief_01d -R0/10/0/10 -Gb.grd
gmt grdfill @earth_relief_20m_holes.grd -Agb.grd -Gnew.grd
echo 7.5 1.5 | gmt grdtrack -Gnew.grd -nn
7.5     1.5     -1169.5
```

joa-quim:
And, BTW, the tests pass here.
```
ctest -R grdfill
Test project C:/v/build
    Start 436: test/grdfill/constfill.sh
1/4 Test #436: test/grdfill/constfill.sh ........   Passed    3.26 sec
    Start 437: test/grdfill/nnfill.sh
2/4 Test #437: test/grdfill/nnfill.sh ...........   Passed    2.33 sec
    Start 438: test/grdfill/showregions.sh
3/4 Test #438: test/grdfill/showregions.sh ......   Passed    2.32 sec
    Start 439: test/grdfill/splinefill.sh
4/4 Test #439: test/grdfill/splinefill.sh .......   Passed    2.64 sec
```

PaulWessel:
And what does this give

`echo 7.5 1.5 | gmt grdtrack -G@earth_relief_01d -R0/10/0/10 -nn`

PaulWessel:
Funny, so maybe no problem? Of maybe random since I think the things I showed earlier are problematic.

joa-quim:
... but now I run the test script with GIT bash, no errors reported and the grid is different, correct I assume because
```
echo 7.5 1.5 | gmt grdtrack -Gnew.grd -nn
7.5     1.5     -1169.5
```
The error with another shell was 
```
gmt [ERROR]: Not available in classic mode
grdimage [ERROR]: Option -c is not a recognized common option
grdimage [ERROR]: Option -c is not a recognized common option
gmt [ERROR]: Not available in classic mode
gmt [ERROR]: Shared GMT module not found: colorbar
```

PaulWessel:
Thanks, so I need to look at that **grdtrack** with remote file.

PaulWessel:
Specifically, we have this in the gmt_init_module function:

`if (!strcmp (mod_name, "grdtrack")) API->use_gridline_registration = true;	/* Override API default since grdtrack is a data processor */`

joa-quim:
I have nothing against processors use grid reg, but at least for nc grid when they were not produced by us we guess registration from x,y coordinates and from GDAL we apply the rule *if not explicit in file grids are grid and images are pix registered*. This together with the native binary formats pretty much cover everything so we should have nothing with a missing registration info.

PaulWessel:
Well, here we know exactly that things are since we produced these grids. But still need to decide, no?

PR #6683, opened 2022-05-07
Improve documentation and usage for movie

PaulWessel:
The various fading durations in **movie** all have default values but the usage and documentation did not indicate that.  That is now fixed.  Also, I have tried to make the  documentation easier to follow by breaking options with multiple modifiers into a list of modifiers and their actions instead of an endless paragraph.  I have also used bold codes where this improves readability.

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -69,7 +69,7 @@ Required Arguments
     set of preset formats (see Table :ref:`Presets <tbl-presets>`) or specify a custom layout.
     **Note**: Your :term:`PROJ_LENGTH_UNIT` setting determines if **movie** sets
     you up to work with the SI or US canvas dimensions.  Instead of a named format you can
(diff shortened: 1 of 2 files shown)


seisman (approved):

anbj:
This is really nice and makes it alot easier to read. Maybe this could/should be applied to all of the docs.

PaulWessel:
Well, yes.  Spare time is all that is needed!

seisman:
I agree. Perhaps we should have a new issue list all ReST files so that people know they can help and we also know which files need to be re-formatted.

PaulWessel:
I would think it is still a long list, no? I know we (and especially @meghanrjones) did many like this - perhaps mostly the common options, but most modules support some option with modifiers so it is pretty extensive.

I think a similar cross-cutting issue is which modules could really benefit from an explanatory figure (like the one I recently added to **mapproject** to explain once and for all the types of oblique boxes..)

PR #6685, opened 2022-05-07
Use superscripts in math mode for km/m3

PaulWessel:
As it says.

### DIFF
--- a/doc/rst/source/supplements/potential/gravprisms.rst
+++ b/doc/rst/source/supplements/potential/gravprisms.rst
@@ -76,7 +76,7 @@ Required Arguments
 *table*
     The file describing the prisms with record format *x y z_lo z_hi* [ *dx dy* ] [ *rho* ],
     where the optional items are controlled by options **-E** and **-D**, respectively.
-    Density contrasts can be given in kg/m^3 of g/cm^3. **Note**: If **-C** is used then
+    Density contrasts can be given in :math:`\mbox{kg/m}^3` of :math:`\mbox{g/cm}^3`. **Note**: If **-C** is used then
     no *table* will be read.
 
 .. _-I:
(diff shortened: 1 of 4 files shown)


doc/rst/source/supplements/potential/gravprisms.rst
```diff
@@ -76,7 +76,7 @@ Required Arguments
 *table*
     The file describing the prisms with record format *x y z_lo z_hi* [ *dx dy* ] [ *rho* ],
     where the optional items are controlled by options **-E** and **-D**, respectively.
-    Density contrasts can be given in kg/m^3 of g/cm^3. **Note**: If **-C** is used then
+    Density contrasts can be given in :math:`\mbox{kg/m}^3` of :math:`\mbox{g/cm}^3`. **Note**: If **-C** is used then
```
seisman, line 79: Should it be 
```
:math:`\mbox{kg/m^3}`
```
?

PaulWessel: I think mbox is just for plain text in the middle of math, so no?

seisman: > I think mbox is just for plain text in the middle of math, so no?

You're right. The following expressions:
```
:math:`\mbox{kg/m}^3`

:math:`\mbox{kg/m^3}`

:math:`kg/m^3`
```
are rendered like:
image


```
:math:`\mbox{kg/m}^3`
```
just gives me the wrong impression of (kg/m)^3, not kg/(m^3).

PaulWessel: OK, so keep as is then.

seisman (approved):

PR #6688, opened 2022-05-08
Fix wrong pen for contour legends

PaulWessel:
The auto-legend code for **grdcontour** and **pscontour** made an assumption that if only one label was given then it was for the annotated contour.  But if the user only sets` -C10 -Wc0.5p,red `we want to see a red pen since **-A** was not specified.  This PR fixes this problem.  No tests were affected.


seisman (approved):

seisman:
The Windows failure is unrelated.

PR #6689, opened 2022-05-08
Suppress some warnings when building manpage

seisman:

Following the comment in https://forum.generic-mapping-tools.org/t/building-the-docs-error/2793/6?u=seisman.


PaulWessel (approved):
Great!

PR #6698, opened 2022-05-09
Add a function for reading triangle indices

PaulWessel:
With **contour -E**_file_ we can read previously generated Delaunay triangle indices tables.  As we plan to add that option to other modules I have moved the relevant code section from _pscontour.c_ to create a new function _gmt_read_triangulation_ which is now called in **contour**.  Tests pass, and all is well.  Future PRs will use this functions in other contexts.


PaulWessel:
Ping @joa-quim to approve this.

joa-quim (approved):
Remember that you don't have to wait for us to around.

PaulWessel:
Yes, true.  Also note I added a check for negative indices which I also think would be bad.

PR #6701, opened 2022-05-09
Let triangulate also have ability to read prior triangulation

PaulWessel:
We add the new option **-L**_indexfile_ to allow us to bypass the Delauney calculation and reuse these precomputed nodes instead, say, for gridding via **-R -I -G**.  The node file may come from a prior run of triangulate or from an external program (e.g., finite elements or similar).

Also this option, for both **triangulate** and **contour** acquires a new modifier **+b** that allows these modules to read the _indexfile_ via binary i/o in the same format as the main table input.  This is necessary to speed up processing for very large triangulation networks.

### DIFF
--- a/doc/rst/source/contour.rst
+++ b/doc/rst/source/contour.rst
@@ -20,7 +20,7 @@ Synopsis
 [ |SYN_OPT-B| ]
 [ |-C|\ *contours* ]
 [ |-D|\ [*template*] ]
-[ |-E|\ *indexfile* ]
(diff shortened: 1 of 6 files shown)


PaulWessel:
Note: I am aware @joa-quim will tell me to self-approve but I am going old-school here.

PaulWessel:
> Not old-school (there you'd do it). Middle-school.

OK, kindergarten I guess.

PR #6716, opened 2022-05-13
Avoid creating local intermediate PDF file (#6711)

### LINKED ISSUE #6711: GMT cannot do parallel jobs in same directory
When running two gmt (new style) scripts in parallel in the same directory, I errors that look like ghostscript errors that indicate conflicts between the two jobs writing in the same temporary PS files. I suspect that some intermediate file has a fixed name, so that the two jobs clash.

E.g.:
 **** Error: Unknown operator: '168.676114.918191269982' looks like a malformed number, replacing with 0.
                Output may be incorrect.
   **** Error: Unknown operator: '.673.48874.1953.3' looks like a malformed number, replacing with 0.
                Output may be incorrect.

I do not think that used to be the case before.

**System information**

* Operating system: macOS 12.3.1 21E258 x86_64
* GMT version (`gmt --version`): 6.4.0_620a645_2022.03.14

remkos:

This lets psconvert not create a local intermediate PDF file under certain circumstances.
Fixes #6711

**Reminders**

- [x] If adding new functionality, add a detailed description to the documentation and/or an example.

### DIFF
--- a/src/psconvert.c
+++ b/src/psconvert.c
@@ -2486,7 +2486,7 @@ EXTERN_MSC int GMT_psconvert (void *V_API, int mode, void *args) {
 				Ctrl->T.device = GS_DEV_PDF;
 				/* After conversion, we convert the tmp PDF file to desired format via a 2nd gs call */
 				GMT_Report (API, GMT_MSG_INFORMATION, "Convert to intermediate PDF...\n");
-				strncat (out_file, &ps_file[pos_file], (size_t)(pos_ext - pos_file));
+				strncat (out_file, ps_file, (size_t)pos_ext);
 				strcat (out_file, "_intermediate");
 			}
 			else {	/* Output is the final result */


PaulWessel (approved):
I wonder: Should be instead create a unique filename for this instead of temp file using the standard temp file machinery?  I am approving for now but maybe I will do some thing even more foolproof later.

remkos:
> I wonder: Should be instead create a unique filename for this instead of temp file using the standard temp file machinery? I am approving for now but maybe I will do some thing even more foolproof later.

The thing is that the file `gmt_0_intermediate.pdf` should be created in the temporary directory `$HOME/.gmt/sessions/gmt-session.<jobid>`. Instead it was created in the working directory in case (and only in case) the output file was in a specified directory (e.g. `img/image.png`). This case caused `pos_file` to be non-zero. The point of this patch is the `pos_file` should not have been used to create this file. It was clearly a copy-paste from elsewhere.

remkos:
I'll wait a bit to see if the MacOS build succeeds at some point. This is clearly caused by something else related to `openmp`. If it keeps on hanging, I may just merge it.

seisman:
> I'll wait a bit to see if the MacOS build succeeds at some point. This is clearly caused by something else related to `openmp`. If it keeps on hanging, I may just merge it.

The macOS failure is unrelated and will be tracked in issue https://github.com/GenericMappingTools/gmt/issues/6718, so it's OK to merge this now.

PR #6721, opened 2022-05-15
Do not try to access key[k][K_MODIFIER] without checking the key[k] size

joa-quim:
I have trace a mysterious Julia failure that occurs randomly to the test at line 12795 in ``gmtapi_extract_argument()``
```
*n_pre = (k >= 0 && key[k][K_MODIFIER] > 0 && isdigit (key[k][K_MODIFIER])) ? (int)(key[k][K_MODIFIER]-'0') : 0;
```
The problem is that most of KEYs have less than 5 characters (K_MODIFIER = 4) and so the above test most of times is accessing uninitialized memory since ``key[k]`` is set via a ``strdup`` call and hence it will not have 5 chars almost never. Setting a test on the key size as in
```
*n_pre = (k >= 0 && strlen (key[k]) >= 5 && key[k][K_MODIFIER] > 0 && isdigit (key[k][K_MODIFIER])) ? (int)(key[k][K_MODIFIER]-'0') : 0;
```
seems to make the random failure to go. I've added a couple of those test in ``gmtapi_extract_argument()`` except in one place (``if ((*n_pre || key[k][K_MODIFIER] == 0) ...`` because doing so caused test failures namely in one that complained that
```
mapproject [ERROR]: No filename provided
```
but none of the ``mapproject`` KEYs have 5 characters, so that ``if`` test seems to be passing by chance but the size root problem remains there as well.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -12792,20 +12792,20 @@ GMT_LOCAL int gmtapi_extract_argument (char *optarg, char *argument, char **key,
 		strcpy (argument, inarg);	/* Any colon-string will be added back in GMT_Encode_Options */
 		gmt_M_str_free (inarg);
 		/* Also return 1 or 2, depending on -G vs -S */
-		*n_pre = (k >= 0 && key[k][K_MODIFIER] > 0 && isdigit (key[k][K_MODIFIER])) ? (int)(key[k][K_MODIFIER]-'0') : 0;
+		*n_pre = (k >= 0 && strlen (key[k]) >= 5 && key[k][K_MODIFIER] > 0 && isdigit (key[k][K_MODIFIER])) ? (int)(key[k][K_MODIFIER]-'0') : 0;
 		return (0);
 	}
 
 	if (k >= 0 && key[k][K_EQUAL] == '=') {	/* Special handling */
 		*takes_mod = 1;	/* Flag that KEY was special */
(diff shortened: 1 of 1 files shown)


PaulWessel:
Would it be better to add

```
size_t L = strlen (key(k);
if (... && L > K_MODIFIER && ...

```
I don't like "5".  

Of course, the logic assumes that it is not possible to get here unless there is at least 5 characters so it would be even better to find out how the colon-check is triggered and perhaps I have forgotten about a completely unrelated case where  colon is present in the argument but has nothing to do with -Sq or -S~ ?

joa-quim:
OK, no 5's now.
It's not only the colon case. There is the other one further bellow that I referred and which will break a test that calls ``mapproject`` if one do what should be done (test the length).

PaulWessel (approved):

PaulWessel:
I did change to using an L variable so easier to check values later.  ON the road right now so dont have time but want to check this more later and think of a way to catch troubles.

joa-quim:
Ok, some more shits fixed. First it seems I forgot to push my second change so you ended up changing the original PR but did not do it in the two need cases. Then I had strange crashes that tuned out to be because 
``strlen (key[k])`` was being computed without checking that ``k >= 0`` and, at least in debug build, the crash was happening because ``k = -1`` so this is another clue for the root problem: this function can be called with a bogus ``k``.

I'll merge it as is because I need to fix the more and more often Julia failures that its absence was causing.

PR #6728, opened 2022-05-19
Remove earth relief dataset information from grdimage documentation

seisman:

The `doc/rst/source/earth_relief.rst_` file is no longer maintained. I think we can remove it.


PaulWessel (approved):
Agreed.

PR #6731, opened 2022-05-20
Update gmt.conf.rst

Esteban82:
Fix typos and replace pi for π.

### DIFF
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -313,15 +313,15 @@ FORMAT Parameters
         [default is **%.12g**]. To give some columns a separate format,
         supply one or more comma-separated *cols*:*format* specifications,
         where *cols* can be specific columns (e.g., 5 for 6th since 0 is the first)
-        or a range of columns (e.g.,3-7).
+        or a range of columns (e.g., 3-7).
         The last specification without column information will override the format for all other columns.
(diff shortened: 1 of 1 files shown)


Esteban82:
Should `[default is %.12g].` be deleted in the descripition of FORMAT_FLOAT_MAP and FORMAT_FLOAT_OUT?

PaulWessel (approved):
Thanks.  I guess pi works like that - great.  I thought we needed :math: etc but apparently not.

PaulWessel:
Hm, I think not.  The see FORMAT_GEO_OUT is for ddd:mm etc and not floating points.  However, for FORMAT_FLOAT_OUT it lists the default twice - perhaps the first one should be removed.

Esteban82:
I have a doubt. 

For example in `FORMAT_FLOAT_MAP` when it says `see FORMAT_GEO_OUT. [default is %.12g].` 

The default value is for FORMAT_FLOAT_MAP or FORMAT_GEO_OUT?

The first time I read it I thought that was for GEO_OUT but now I think it is for FLOAT_MAP.

PaulWessel:
Yes, it is for float_out - sorry it is bit unclear.

Esteban82:
> Yes, it is for float_out - sorry it is bit unclear.

Ok, so I will edit the text to make it clear.

PaulWessel (approved):

Esteban82:
I can't merge it (because there is a failing check).

Issue #6739, opened 2022-05-24
xyz2grd segfaults when not enough memory


anbj:
This crash in on version 6.3.0. I'm not able to install a newer version on the system. Suspected cause is not enough memory. I installed 6.3.0 on a system with sufficient memory, and I do not experience any crash. My input file is several hundred megabytes so not sure how I can share it, and if it is necessary.

Equivalent smaller input files that do fit memory is gridded without problems.

Command:
```
gmt xyz2grd -I1 60062_L05_32631_DTM_18_S02_20180814_1M.XYZ $(gmt info -I1 60062_L05_32631_DTM_18_S02_20180814_1M.XYZ) -G60062_L05_32631_DTM_18_S02
_20180814_1M.XYZ.nc -Vd
```

Result (with `-Vd`):
```
gmt [DEBUG]: GMT_Create_Session: Terminal width = 100
gmt [DEBUG]: Obtained the ppid from parent: 364586
gmt [DEBUG]: Enter: gmtinit_new_GMT_ctrl
gmt [DEBUG]: GMT->session.SHAREDIR = /usr/share/gmt
gmt [DEBUG]: GMT->session.HOMEDIR = /home/anbj@od.local
gmt [DEBUG]: GMT->session.USERDIR = /home/anbj@od.local/.gmt [created]
gmt [DEBUG]: GMT->session.CACHEDIR = /home/anbj@od.local/.gmt/cache [created]
gmt [DEBUG]: GMT: 0. Will try to find subdir=postscriptlight stem = PSL_custom_fonts suffix=.txt
gmt [DEBUG]: GMT: 1. gmt_getsharepath trying current dir
gmt [DEBUG]: GMT: 2. gmt_getsharepath trying USERDIR /home/anbj@od.local/.gmt
gmt [DEBUG]: GMT: 3. gmt_getsharepath trying USERDIR subdir /home/anbj@od.local/.gmt/postscriptlight
gmt [DEBUG]: GMT: 4. gmt_getsharepath trying SHAREDIR subdir /usr/share/gmt/postscriptlight
gmt [DEBUG]: GMT: 5. gmt_getsharepath trying SHAREDIR /usr/share/gmt
gmt [DEBUG]: GMT: 6. gmt_getsharepath failed
gmt [DEBUG]: Map distance calculation will be Cartesian
gmt [DEBUG]: Exit:  gmtinit_new_GMT_ctrl
gmt [DEBUG]: Enter: New_PSL_Ctrl
[...]

PR #6745, opened 2022-05-25
Add alloc return checks in xyz2grd

PaulWessel:
Might be related to #6739 or not, but fixing a few memory allocations that did not check the return values.


anbj:
Difficult to review since I'm not able to install on affected system. What should I do?

anbj:
Ok!

joa-quim (approved):
Easy rev (between two glasses of wine). Traditional check-if-not-enough-mem.

PaulWessel:
I am thinking this might be @anbj problem: It happily allocates the large grid but when it later try to allocate the flag it fails but it does not check that and chaos ensues.

anbj:
Yes Sir; fixed.

PR #6746, opened 2022-05-25
Check that files in the share-path are nonzero if in current directory

PaulWessel:
This situation can happen when users run things like

`gmt makecpt -Cturbo > turbo.cpt`

because the output is created as an empty file first by the system, and then **makecpt** reads that instead of the master file.  The simple solution is to insist that when we search for GMT files expected in the _share path_, while we will allow such a file to live in the current directory it must pass the extra test that it is not empty.

All tests run and the old problem above is gone.

Recently updated docs (#6744 ) have been edited again to remove the mention of the problem.

If anyone runs into documentation that discusses the old problem we need to update that as well - I did not find any beyond the two modules.

### DIFF
--- a/doc/rst/source/grd2cpt.rst
+++ b/doc/rst/source/grd2cpt.rst
@@ -309,26 +309,6 @@ Here, cdf.txt would be the cumulative hypsometric curve for the Earth.
 
 .. include:: cpt_notes.rst_
 
-Restriction on CPT Output Names
--------------------------------
-
-Since **grd2cpt** will also interpolate from any existing CPT you
-may have in your directory, you should never use one of the master CPT names
-as an output filename; hence the my_gebco.cpt in the example.  If you
-do create a CPT of such a name, e.g., rainbow.cpt, then **grd2cpt** will
-read that file first and not look for the master CPT in the shared GMT
-directory. For instance, the command::
-
-    gmt grd2cpt my_grid.grd -Crainbow > rainbow.cpt
-
-will return error messages like this::
-
-    grd2cpt [ERROR]: Color palette table rainbow.cpt is empty
-
-since the redirection will first create an empty file rainbow.cpt before
-**grd2cpt** even has started and it will then try to read it and it is all
-downhill from there.
-
 See Also
 --------
 
--- a/doc/rst/source/makecpt.rst
+++ b/doc/rst/source/makecpt.rst
@@ -345,26 +345,6 @@ To make a categorical CPT with string keys instead of numerical lookup values, t
 
 .. include:: cpt_notes.rst_
 
-Restriction on CPT Output Names
--------------------------------
-
-Since **makecpt** will also interpolate from any existing CPT you
-may have in your directory, you should never use one of the master CPT names
-as an output filename; hence the my_gebco.cpt in the example.  If you
-do create a CPT of such a name, e.g., rainbow.cpt, then **makecpt** will
-read that file first and not look for the master CPT in the shared GMT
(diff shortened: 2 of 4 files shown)


joa-quim (approved):

PaulWessel:
Before finishing this one: Maybe the even better solution is to never allow share-path files to be read from current dir.  Why should we allow a file called turbo.cpt, which we dont know what contains, be read instead of the share/cpt/turbo.cpt?  While

`gmt makecpt -Cturbo -T-2/6 > turbo.cpt`

will now work fine, any future calls on turbo not sending output to that same name will be using this turbo.cpt.  So if I do

`gmt makecpt -Cturbo -T-2/6  -G0.4/1 > turbo.cpt`

and create a truncated version of turbo, the next time I run

`gmt makecpt -Cturbo -T-2/6 > test.cpt
`
I am not going to get the turbo master but the cropped version.  Same with custom symbols and whatever.  Is there a reason we should allow a user to have flash.def in the current directory and that is the one -Sk0.1i/flash should return?

**Proposal**: Never allow a share-path type file to be read from the current directory in order to protect the integrity of the GMT shared file system.

Comments, @GenericMappingTools/core ?

joa-quim:
Hmm, not sure. We can see that both ways. If for some whatever-good-or-stupid reasons I do
```
gmt makecpt -Cturbo -T-2/6 -G0.4/1 > turbo.cpt
```
and try to use it, the result will not be what is expected and we get a bug report that the program is not working. And also, to implement your proposal the file names (cpts, custom symbols) would have to checked against **all** names in the shared-path. Code complication and no-need slowdown.

PaulWessel:
There is no such check of list of names  Just _gmt_getsharepath_ that refuses to return the path from the current directory.

PaulWessel:
In fact, such a PR would be much simpler than what I have in this one, which is really simple anyway.

remkos:
I don't think you should limit based on file type as @PaulWessel suggested. As @joa-quim already said, that would be extensive to implement, and a waste of time. But secondly, I regularly do something like this:
```
gmt makecpt -Cturbo -T-2/6 > test.cpt
gmt grdimgae -Ctest.cpt ....
```
Note that I would use the full name in `grdimage -C`. If you would block the `cpt` file type from looking in the local directory a lot of people's scripts are going to crash.

If you *really* still want to do something, I would suggest that the automatic extension with `.cpt` is *only* done when looking in the share directory. So `-Cturbo` would look for `turbo.cpt` in the share directory but not in the local directory. The search in the local directory should maybe only be done with the full file name.

joa-quim:
I like the ``-Cturbo.cpt`` restriction. The problem is that then we are back to beginning. How to create that ``turbo.cpt`` in current dir?

remkos:
That would work with my suggestion that `-Cturbo` would only look for `turbo` (literally) in the current directory and `turbo.cpt` in the shared directory. Then `turbo.cpt` in the current is safe.

joa-quim:
> -Cturbo would only look for turbo (literally) in the current directory

That would be a breaking change. By not specifying the ``.cpt`` extension has always meant to use the GMT master CPT.

PaulWessel:
I believe GMT treats -Cturbo differently than -Cturbo.cpt.  Note we are not talking about grdimage -C here, just makecpt and grd2cpt which **creates** CPT files.

joa-quim:
> just makecpt and grd2cpt which creates CPT files.

Anyway. We cannot forbid to create new cpts from other ones that live in current directory. That would be a serious breaking.

Issue #6415, opened 2022-03-06
Inset within subplot panel fails


PaulWessel:
As reported on the [forum](https://forum.generic-mapping-tools.org/t/the-inset-picture-doesnt-displayed/2677), **inset** used in subplot panels fails in terms of placement.  Clearly the -Dj setting is computing strange placement coordinates.  In the original post, the BR box was plotted way to the right and later covered up by other plots.

A simple example contrasts plain inset versus inset inside a panel:

```
#!/bin/bash
# Shows problem with inset within subplot panel
gmt begin
    gmt figure inset_in_panel png
    gmt subplot begin 1x1 -Fs5c/4c -A
        gmt subplot set 0
        gmt coast -R70/138/13/56 -Glightgray -B
        gmt inset begin -DjTL+w0.5c/0.8c+o0.1c -F+p1p,red
        gmt inset end
    gmt subplot end

    gmt figure plain_inset png
    gmt coast -R70/138/13/56 -Glightgray -B -JQ5c
    gmt inset begin -DjTL+w0.5c/0.8c+o0.1c -F+p1p,red
    gmt inset end

gmt end show
```

Only showing the failing inset_in_panel plot here (the other is fine):

![inset_in_panel](https://user-images.githubusercontent.com/26473567/156923887-0753e3bf-72c1-4c9e-9146-60f435234c7e.png)

gd-a:
Related to [#6288](https://github.com/GenericMappingTools/gmt/issues/6288) no?

seisman:
I had the same problem today. Ping @PaulWessel just in case you forgot it.

PaulWessel:
Not truly forgotten, but on the road...

seisman:
Here is another example that doens't work as expected:
```
gmt begin map
	gmt subplot begin 1x2 -Fs15c/15c
	gmt subplot set 0
		gmt coast -R-74/-69.5/41/43 -JM? -Baf -Wthin -Clightblue -Glightyellow
		gmt inset begin -DjBL+o0.5c/0.2c -F+pblack+gwhite+c0.2c/0.2c -R-80/-65/35/50 -JM3c
			gmt coast -Ggray -Cwhite -EUS.MA+gred -W1/thin -N1 -N2
	    gmt inset end
	gmt subplot set 1
		gmt coast -R-74/-69.5/41/43 -JM? -Baf -Wthin -Clightblue -Glightyellow
[...]

PR #6751, opened 2022-05-26
Let modules called by modules in inset know of shrinking

PaulWessel:
See this [post](https://github.com/GenericMappingTools/pygmt/issues/1930) for background. The problem was that the **coast** module called another module (**plot**) while inside the **inset** section but the called module was unaware that because of the **-R -J** and inset size setting there was a scaling of 0.73 applied.  That meant **plot** was off by that factor (unless we countered that by appending **+du** to **-J**, which the user would not know was needed).  The same would happen with any other module called inside the **inset** section that might call another module (e.g., **plot** calls **plot** when dealing with decorated lines, for instance).

The solution is to remember the computed inset scale in the API so that when the secondary module is initialized it can apply the same scaling as the mother module; this scaling happens in the belly of _gmt_proj_setup_.

With those changes, this works correctly:

```
gmt begin map
    gmt coast -R-74/-69.5/41/43 -JM15c -Baf -Wthin -Clightblue -Glightyellow
    gmt inset begin -DjBL+w3c+o0.5c/0.2c -F+pblack+gwhite
        gmt coast -R-80/-65/35/50 -JM? -Ggray -Cwhite -EUS.MA+gred -W1/thin -N1 -N2
    gmt inset end
gmt end show
```

### DIFF
--- a/src/inset.c
+++ b/src/inset.c
@@ -447,6 +447,9 @@ EXTERN_MSC int GMT_inset (void *V_API, int mode, void *args) {
 		GMT_Report (API, GMT_MSG_DEBUG, "inset: Removed inset file\n");
 		gmt_reload_history (API->GMT);
 		gmt_reload_settings (API->GMT);
+		/* Undo any shrink scaling memory */
+		API->inset_shrink_scale = 1.0;
+		API->inset_shrink = false;
 	}
 
 	gmt_plotend (GMT);
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -13753,6 +13753,9 @@ void gmt_end (struct GMT_CTRL *GMT) {
 	gmt_M_free (GMT, GMT->parent->remote_info);
 	/* Free snapshot of GMT common option structure */
 	gmt_M_free (GMT, GMT->parent->common_snapshot);	/* Free snapshot */
+	/* Undo any inset shrink scaling memory */
+	GMT->parent->inset_shrink_scale = 1.0;
+	GMT->parent->inset_shrink = false;
 
 #ifdef MEMDEBUG
 	gmt_memtrack_report (GMT);
--- a/src/gmt_map.c
+++ b/src/gmt_map.c
@@ -2195,6 +2195,10 @@ GMT_LOCAL void gmtmap_setxy (struct GMT_CTRL *GMT, double xmin, double xmax, dou
 		}
 		update_parameters = true;
 		GMT_Report (GMT->parent, GMT_MSG_DEBUG, "Rescaling map for inset by factors fx = %g fy = %g dx = %g dy = %g\n", fx, fy, I->dx, I->dy);
+		if (fx < 1.0) {	/* Need this for modules called within the inset */
+			GMT->parent->inset_shrink = true;
+			GMT->parent->inset_shrink_scale = fx;
+		}
 	}
 	else if (P->active && no_scaling == 0)	{	/* Must rescale to fit inside subplot dimensions and set dx,dy for centering */
 		gmtmap_adjust_panel_for_gaps (GMT, P);	/* Deal with any gaps requested via subplot -C: shrink w/h and adjust origin */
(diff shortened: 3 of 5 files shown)


src/inset.c
```diff
@@ -447,6 +447,9 @@ EXTERN_MSC int GMT_inset (void *V_API, int mode, void *args) {
 		GMT_Report (API, GMT_MSG_DEBUG, "inset: Removed inset file\n");
 		gmt_reload_history (API->GMT);
 		gmt_reload_settings (API->GMT);
+		/* Undo any shrink scaling memory */
+		API->inset_shrink_scale = 1.0;
```
joa-quim, line 451: I'm afraid this precaution might not be enough as it only happens inside ``inset.c``
Because whatever is set in API carries to other modules calls in Julia/MEX this undoing should be done in ``end_session`` as well, I think.

PaulWessel: Hm, it is only set when gmt inset is called and shrinking is required, and it is guaranteed there will be a gmt inset end call to follow.  So it only lives as long as we are in an inset.  It wont hurt to add it to gmt end but not sure if it would make a difference.  Can you demonstrate the need?

joa-quim: Imagine there is an error before the ``gmt inset end`` call? 
What I know is that there still some of these shits that manage to survive and there are cases where I'm forced to call a ``gmt("destroy")`` to workaround the unknown trouble.

PaulWessel: OK, I dont mind.  So you think gmt_end (which is called by GMT_Destroy_Session) is the place then?

joa-quim: Yes, probably ``gmt_end``. It's the place where all this kind of stuff is undone.

joa-quim (approved):

PaulWessel:
As I look at this a bit more, note that this is also a better way to determine the inset: Let **gmt inset** decide from the **-R -J**:

```
gmt begin map
    gmt coast -R-74/-69.5/41/43 -JM15c -Baf -Wthin -Clightblue -Glightyellow
    gmt inset begin -DjBL+o0.5c/0.2c -F+pblack+gwhite -R-80/-65/35/50 -JM3c
        gmt coast -Ggray -Cwhite -EUS.MA+gred -W1/thin -N1 -N2
    gmt inset end
gmt end show
```

Here, **-D** does _not_ specify the size but we give the desired **-R -J** and now the inset is computed correctly with a tight fit.  If you dont want a width of 3c but a height of 3c then you append **+dh** to **-J**, for instance.

In the original user's case, this is the preferred way of doing things.  But, this PR is still useful since it fixes a bug.

PaulWessel:
We have one failure (insetlegend.sh) so I will check that one before this can be merged or revised.

PaulWessel:
I will let @seisman give this branch a test tonight/tomorrow before merging.

seisman:
> ```
> gmt begin map
>     gmt coast -R-74/-69.5/41/43 -JM15c -Baf -Wthin -Clightblue -Glightyellow
>     gmt inset begin -DjBL+o0.5c/0.2c -F+pblack+gwhite -R-80/-65/35/50 -JM3c
>         gmt coast -Ggray -Cwhite -EUS.MA+gred -W1/thin -N1 -N2
>     gmt inset end
> gmt end show
> ```

Great! It's definitely a better solution!

---

Just want to let you know that inset may not works well in subplot, as shown by the script below. But I think it can be addressed in a separate PR instead.

```
gmt begin map
	gmt subplot begin 1x2 -Fs15c/15c
	gmt subplot set 0
		gmt coast -R-74/-69.5/41/43 -JM? -Baf -Wthin -Clightblue -Glightyellow
		gmt inset begin -DjBL+o0.5c/0.2c -F+pblack+gwhite+c0.2c/0.2c -R-80/-65/35/50 -JM3c
			gmt coast -Ggray -Cwhite -EUS.MA+gred -W1/thin -N1 -N2
	    gmt inset end
	gmt subplot set 1
		gmt coast -R-74/-69.5/41/43 -JM? -Baf -Wthin -Clightblue -Glightyellow
	    gmt inset begin -DjBL+o0.5c/0.2c -F+pblack+gwhite+c0.2c/0.2c -R-80/-65/35/50 -JM3c
		    gmt coast -Ggray -Cwhite -EUS.MA+gred -W1/thin -N1 -N2
	    gmt inset end
    gmt subplot end
gmt end show
```

map

seisman:
Just realized that the inset in subplot issue was already reported in https://github.com/GenericMappingTools/gmt/issues/6415. I'll repost the example there.

PaulWessel:
Thanks @seisman - we will look at the old sublot + inset issues separately.  OK to merge this PR though?

seisman (approved):

PR #6764, opened 2022-05-31
Fix typos in seamount modules

ZMAlt:

Fixes #

**Reminders**


PaulWessel (commented):
The gravprisms changes are correct but the grdseamount is not: The -V is used and then we report back the mean density per seamount.  That is separate from what -W does.  Please undo that change and then I can approve the restl

ZMAlt:
Done

PaulWessel (approved):
Thanks for your help!

PR #6768, opened 2022-06-06
Ensure column-wrap does not exceed periodicity

PaulWessel:
See this[ forum post ](https://forum.generic-mapping-tools.org/t/grdmask-core-dump/2988/7)for background.  Some months ago I fixed a bug in how we set the max width of searches per latitude row and as we reach the pole the answer should be the full grid width, not the half.  However, because we use the periodicity count to add to any temporarily negative column values it is important that the upper limit on the search width (in nodes) at most equals that periodicity count.  This count is the same as the number of columns for pixel grids but is _one less_ for gridline grids.  The problem reported on the forum resulted from using _n_columns_ and not the periodicity wrap count (here _n_columns-1_) for the search width.

All tests pass.  However, since the crash reported on the forum did not crash for me on macOS we need someone to test this branch on Linux to verify it has been fixed.


Esteban82 (approved):
I test with Becker's data.

PR #6779, opened 2022-06-08
Fix what seems to have been a logical mistake with ||

joa-quim:
I'm getting this very worrying warning and I think the compiler is right.

```
C:\progs_cygw\GMTdev\gmt5\master\src\gmt_support.c(4008): warning C4554: '|': check operator precedence for possible error; use parentheses to clarify precedence
```


PaulWessel (approved):
Typo

PR #6782, opened 2022-06-10
Add workflow for uploading baseline images as a release asset

meghanrjones:

Based on https://github.com/GenericMappingTools/pygmt/pull/1317

Closes https://github.com/GenericMappingTools/gmt/issues/5724

**Reminders**

### DIFF
--- a/.github/workflows/release-baseline-images.yml
+++ b/.github/workflows/release-baseline-images.yml
@@ -1,0 +1,40 @@
+# Upload the ZIP archive of baseline images as a release asset
+
+name: Upload baseline images
+
+# Only run for releases
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  upload-baseline:
+    name: Upload baseline images
+    runs-on: ubuntu-latest
+    if: github.repository == 'GenericMappingTools/gmt'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3.0.2
+
+    - name: Setup data version control (DVC)
+      uses: iterative/setup-dvc@v1.0.3
+
+    - name: Pull baseline image data from dvc remote
+      run: |
+        dvc pull
+        ls -lhR test/baseline/
+
+    - name: Create the baseline image asset in zip format
+      run: |
+        mkdir baseline-images
+        cp -r test/baseline/* baseline-images/
+        zip -r baseline-images baseline-images/ -x "*.dvc"
+        shasum -a 256 baseline-images.zip
+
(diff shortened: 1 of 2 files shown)


.github/workflows/release-baseline-images.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Checkout
+      uses: actions/checkout@v3.0.2
+
+    - name: Setup data version control (DVC)
+      uses: iterative/setup-dvc@v1.0.3
+
+    - name: Pull baseline image data from dvc remote
+      run: |
+        dvc pull
+        ls -lhR test/baseline/
```
seisman, line 27: I think we also need the images in `doc/examples` and `doc/scripts`

meghanrjones: These are included in the release tarball after https://github.com/GenericMappingTools/gmt/pull/6533, but I added them to the baseline-images asset as well in https://github.com/GenericMappingTools/gmt/pull/6782/commits/aa93c0693d8decf6ae0b41fd3a63663e7ec12d69. This change increases the size from ~40 MB to ~70 MB.

meghanrjones:
@seisman, do you think having this as a separate workflow is better than zipping up the files as part of the [build-release script](https://github.com/GenericMappingTools/gmt/blob/master/admin/build-release.sh) and adding the upload step to the [draft-release](https://github.com/GenericMappingTools/gmt/blob/master/.github/workflows/draft-release.yml) workflow? I am undecided about which would be preferable.

meghanrjones:
Right, it would need to be run by Paul and placed on the GMT FTP server with the other files. @PaulWessel, any opinions on whether to bundle and upload the baseline images manually as part of `build-release.sh` along with the other files or keep it separate as currently written in this PR?

PaulWessel:
I think it is probably fine as you have it in the CI - maybe one day the rest will follow.

meghanrjones:
> In terms of release timing: I have all Saturday and maybe a short slice on Sunday but then it is in the car with the Watts and off to some serious hiking in the Lake District. Probably not going to be productive at some mountain cabin. Then on Tuesday I will be dropped off in Newcastle and take the train to Edinburgh for a Uni visit and talk. Then a 6 hour train trip back on Wednesday. So let's see if we can get lots of stuff done before then - otherwise the release may be end of next week.

Sounds fun!
I can update the changelog, citation info, and test with PyGMT tomorrow AM EST. I think this is ready if I can get an approval on it.

seisman (approved):

Issue #6799, opened 2022-06-17
Can't change font and offset of colorbar's Y label


seisman:
**Description of the problem**

As shown below, `MAP_LABLE_OFFSET` and `FONT_LABEL` can control the X label, but have no effects on the Y label.

```
gmt colorbar -Cseis -Djx0c/0c+h -R0/10/0/10 -JX10c -Bxaf+l"Xaxis" -By+l"Depth" --MAP_LABEL_OFFSET=20p --FONT_LABEL=10p,1 -png map
```

![map](https://user-images.githubusercontent.com/3974108/174211518-13eab8f9-0ebb-405c-b364-19ea37143caa.png)

**System information**

* GMT version (`gmt --version`): 6.4.0

PaulWessel:
True.  The code shows this:

```
		if (label[0]) {	/* Add label */
			form = gmt_setfont (GMT, &GMT->current.setting.font_label);
			gmt_map_text (GMT, xleft + 0.5 * length, y_label, &GMT->current.setting.font_label, label, 0.0, Label_justify, form);
		}
		if (unit[0]) {	/* Add unit label */
			form = gmt_setfont (GMT, &GMT->current.setting.font_annot[GMT_PRIMARY]);

```

So the y-label uses the annotation font as you show.  I cannot see anything in the documentation that says this is how it is supposed to work, so I think we can declare a bug and change this to **FONT_LABEL**.  Agreed? It is possible this will require updated PS files for those cases with y-labels on color bars.

PaulWessel:
For reference, the proposed PR makes this plot:

![map](https://user-images.githubusercontent.com/26473567/174442637-7874c89a-68d6-4a96-b4a0-7c8d22c24d4c.png)

Esteban82:
I think this was already fixed so I will close it. 

For the record, `-By+l"..."` is the colorbar unit, not a label — use `FONT_ANNOT_SECONDARY / MAP_ANNOT_OFFSET_SECONDARY` for it, not `FONT_LABEL / MAP_LABEL_OFFSET` (that only affects `-Bxaf+l"..."`).

```
gmt begin test_6799 png
    gmt set FONT_LABEL 10p,1 MAP_LABEL_OFFSET 10p
    gmt set FONT_ANNOT_SECONDARY 12p,1 MAP_ANNOT_OFFSET_SECONDARY 20p
[...]

PR #6802, opened 2022-06-17
Fix incorrect font for colorbar unit

PaulWessel:
See #6799 for background.  The label on the color bar for units was set using annotation font and offset instead of the correct label font and offset.  @seisman, please build and run tests and see if you agree with me that the ~35 failing tests are better with the larger font for label and in line with other use of labels.  We can then address this after 6.4 as a bug fix.


PaulWessel:
Just pinging @GenericMappingTools/core as well for now.

PaulWessel:
Have not heard back from anybody about this so pinging @seisman, @maxrjones, @joa-quim and @Esteban82.  Perhaps it is worth more discussion.  Here is what master currently does, using the **FONT_ANNOT_PRIMARY** [12p] for the y-axis label:

master

This PR would declare a bug and state that as for all basemaps the **FONT_LABEL** [16p] should be used instead, giving

wip

Once could argue that colorbars are very unusual basemaps in that one axis is much longer than the other (factor 20 by default) and that the short axis is never annotated or ticked.  Hence, using the label font for a short unit label seems too large.  This is of course why I switched to the annotation font decades ago.

However, as it is we are unable to change it independently of the axis annotation, and this WIP solution would affect both axes labels.  An alternative would be to declare the bug differently and say it was meant to be the secondary annotation font **FONT_ANNOT_SECONDARY** [14p] instead of the primary, thus allowing for

- A smaller default font size than label font [14p vs 16p]
- A way to change it without affecting the annotation or main axis label settings

I must say this type of bug-fix offers a better solution for this type of plot. Thoughts?

maxrjones:
> I must say this type of bug-fix offers a better solution for this type of plot

agreed, for this case it is important for the unit label to be controlled independently from the colorbar label

joa-quim:
I don't like the bottom example. Font is way too big. But agree that a different control for the two labels (with a nice default, auto-scaled?) is useful.

PaulWessel:
Yes true.  The whole concept of treating the tiny colormap as a basemap with the same defaults as an actual basemap is wrong.  It would seem that the font sizes, ticks etc ought to scale with the color bar length, independent of the auto-scaling for the map it is attached to.  Otherwise one has to do a lot of --PAR=value to make it look OK.

PaulWessel:
Well, it was just a dumb example to show font size.  Normally all my research into hair-length uses proper SI units.

PaulWessel:
I want to wrap this up.  Sounds like we are in agreement that this is a good fix and that perhaps further work on default sizes is needed, but that is separate from this fix.  Unless I hear fussing I will update those 35 PS files given this fix and complete it.

PaulWessel:
I am resurrecting this stale PR from last year.  I decided (as I think we agreed) to use the SECONDARY annotation parameters for the little unit on the side of the bar instead of LABEL which is too big.  This way, the original request in #6799 is addressed.

This will cause a bunch of failures but with gs 10 and Xcode update I have plenty already.  Hoping @seisman has some clear thinking on what to do next.

PaulWessel:
@GenericMappingTools/core, shall we bite the bullet here and go with **FONT_ANNOT_SECONDARY** for the y-annotations instead of **FONT_LABEL**, and then deal with the tens of PostSCript files that need to be updated given the change? Otherwise this PR will sit here until nobody will remember what to do.

joa-quim:
I guess there is no escape from that (the ps updates). But, given the current ghost/gm partnership screwing of good PSs, how to tell the new failures due to this change from the others?

PaulWessel:
These are easy since they only affect colorers with y-label.  The font size change so they all fail.

PaulWessel:
So @seisman , I am testing with homebrew and gs 10.00.0.  I think you were saying earlier that the CI uses older stuff and if we update the PS to fit 10.00.0 we will get lots of failures? How can we not be held hostage to old installs?

joa-quim:
gs 10 is clearly f. See https://discourse.julialang.org/t/plotting-countries-with-geomakie-reduce-size-and-more/97966/2?u=joa-quim

remkos:
Regarding the mess that gs 10 makes of PNG, in the framework of macports I was offered a patch to try for ghostscript that may work. See https://github.com/macports/macports-ports/pull/18139

Given the description of the patch, I'm not so sure this will resolve the striping, unless the case linked by @joa-quim uses alphabits. In any case, are you willing to try a patched version of ghostscript 10.01.1?

PaulWessel:
I cannot use macports since they f...ed the GDAL port up and we have been waiting for months for them to convert from auto tools to CMake.  But looks like they finally are doing something with it

joa-quim:
Here is a PS that fs with ghost 10. Damn it, they are screwing all the time.

[GMTjl_j.ps.zip](https://github.com/GenericMappingTools/gmt/files/11416521/GMTjl_j.ps.zip)

PaulWessel:
Update: Macports has fixed their GDAL problem an 3.6.4 now installs.  Now, admin/build-release.sh once again works so we can plan the 6.5 release.

joa-quim:
Yes, I think I was running 10.01.1, the latest they have but remove it and installed 9.56

seisman:
> shall we bite the bullet here and go with **FONT_ANNOT_SECONDARY** for the y-annotations instead of **FONT_LABEL**, and then deal with the tens of PostSCript files that need to be updated given the change?

I agree that **FONT_ANNOT_SECONDARY** sounds more reasonable. We also need to update the documentation of `colorbar`.

PaulWessel:
I have now updated the docs for colorbar to discuss **FONT_ANNOT_SECONDARY** instead of **FONT_LABEL** for the unit annotation.  If you run this branch and tests you will get those extra failures that we need to update the PS for.  I cannot do that (yet) because I have 100+ failures due to gs and different data versions....
When you run, do you wife server so that it downloads fresh, or how?

joa-quim:
Not sure I understood what you are proposing. Leaving the 3 cm example as is by default? I think it should be scaled. No reason to produce an ugly fig when we can do better.

PaulWessel:
I mean we should auto-scale colorbars annotations like we do for a map, so that a 3 cm colobar will have smaller annotations than the20x20 cm map it may be inserted inti.  The 3c is just what we get now and it is ugly.

PaulWessel:
Update this colorbar-unit branch, build, and run this script. It simply uses a 15 cm bar for reference and then scales all parameters down by length/15.  Looks OK without having played much with "15" as the reference,

If you sant to play and optimise it, see line 1006 in psscale.c and mess with that.

two

PaulWessel:
@Esteban82 and @anbj might also have opinions on this scaling.  Basically, while gmt modern mode with modern theme automatically scales the map fonts and thickness, but we never did this for color bars.  Clearly a short bar cannot just use the default primary font size (12pm) so it does need its own separate scale.

Because this will look much bette without messing with **FONT_ANNOT_PRIMARY** etc I think we declare a bug.

PaulWessel:
Maybe @seisman who initiated this issue has a comment on whether my sqrt of ratio to 15 cm looks good?

anbj:
Do not have very strong opinions about this. All plots in https://github.com/GenericMappingTools/gmt/pull/6802#issuecomment-1697399020 looks very nice to me. So go for it?

PaulWessel:
@joa-quim is this an ok scaling?

joa-quim:
Yes, looks perfect.

PaulWessel:
> Yes, looks perfect.

OK, maybe an approval?

joa-quim (approved):

src/psscale.c
```diff
@@ -1002,6 +1002,21 @@ GMT_LOCAL void psscale_draw_colorbar (struct GMT_CTRL *GMT, struct PSSCALE_CTRL
...
+	GMT->current.setting.map_frame_pen.width *= scale_down;
+	GMT->current.setting.map_tick_pen[GMT_PRIMARY].width *= scale_down;
+	GMT->current.setting.map_tick_pen[GMT_SECONDARY].width *= scale_down;
+	GMT->current.setting.map_tick_length[GMT_ANNOT_UPPER] *= scale_down;
+	GMT->current.setting.map_tick_length[GMT_TICK_UPPER] *= scale_down;
+	GMT->current.setting.map_annot_offset[GMT_PRIMARY] *= scale_down;
+	GMT->current.setting.map_annot_offset[GMT_SECONDARY] *= scale_down;
+	GMT->current.setting.map_label_offset[GMT_X] *= scale_down;
+	GMT->current.setting.map_label_offset[GMT_Y] *= scale_down;
+	fprintf (stderr, "scale-down = %lg\n", scale_down);
```
seisman, line 1019: Should remove this `fprintf` or change it to GMT DEBUG MESSAGE:
```suggestion
	fprintf (stderr, "scale-down = %lg\n", scale_down);
```

joa-quim:
DVC is giving more troubles than good. I wonder if we shouldn't simply create a separate git reportory with the PS's. And when it grows too much we just delete it and restart. 

```
C:\progs_cygw\GMTdev\gmt5\master>dvc pull
  0% Querying remote cache|                                                     |0/128 [00:00<?,    ?files/s]Exception in thread fsspecIO:
  File "threading.py", line 1016, in _bootstrap_inner
  File "threading.py", line 953, in run
  File "asyncio\base_events.py", line 603, in run_forever
  File "asyncio\base_events.py", line 1871, in _run_once
  File "selectors.py", line 324, in select
  File "selectors.py", line 315, in _select
ValueError: too many file descriptors in select()
```

joa-quim:
I'm having several failures like this

equalarea_diff

PR #6814, opened 2022-06-22
gmt.conf.rst: MAP_ANNOT_OBLIQUE, put keywords in table

anbj:
Put keywords and meaning in table for ``MAP_ANNOT_OBLIQUE``.

I was hoping to be able to break lines in the table (both in the unrendered and rendered gmt.conf.rst) to make it narrower (i.e. see all text without scrolling), but this turned out to be more difficult than I thought. Any tips?

snip


anbj:
Super, thanks.

Issue #6820, opened 2022-06-23
MAP_FRAME_TYPE=graph centered on {0,0}


gd-a:
Would it be possible to add an option to have this : 

```
gmt begin test png
gmt basemap -R-1/1/-1/1 -JX10c -Bafg --MAP_FRAME_PEN=0p
gmt basemap -Bb --MAP_FRAME_TYPE=graph -Y5c
gmt basemap -Bl --MAP_FRAME_TYPE=graph -Y-5c -X5c
gmt end show
```
![test](https://user-images.githubusercontent.com/74073266/175363553-32767e8c-3187-4f42-ad5c-9665615d4aa6.png)

By parsing this simple parameter : 
```
gmt basemap -R-1/1/-1/1 -JX10c -Bafg --MAP_FRAME_TYPE=graph,centered -png result
```

Best,
Guillaume

cf [forum](https://forum.generic-mapping-tools.org/t/shading-between-2-curves/3044)

PaulWessel:
I've had this need myself so it is very likely this will happen in one form or the other.

Issue #6292, opened 2022-02-02
-B+e (to skip first or last annotation) not working with dates


gd-a:
Hi there, never 2 without 3 (not sure this translate well in English)... anyway,

I'm trying to finish [a code](https://forum.generic-mapping-tools.org/t/plot-w-flag-bug/2556/6?u=planetgus) and I'm almost done apart from the first x annotation that bothers me.

I found in the doc the `-B+e` option, and ... it didn't work. After some trial the problem seems to be only for "month" annotation.

MWE : 
```
gmt begin test png
gmt basemap -R2000-01-01T/2001-01-01T/-1/1 -JX5c -Bafg+e --FORMAT_DATE_MAP=o --FORMAT_TIME_PRIMARY_MAP=a 
gmt end show 
```

gd-a:
I has been quite some time... I don't clearly remember but I suppose the +e option works with floats but not with dates ?

Here's where I failed to implement it (see the graph at the BR corner): the **S**eptember overlap with the vertical axis
https://forum.generic-mapping-tools.org/t/animation-sea-ice-v3/2786

PR #6824, opened 2022-06-25
Allow MAP_FRAME_TYPE=graph-origin

PaulWessel:
See #6820 for background (@gd-a).  This is a test branch for exploring this possibility.  There are some remaining decisions to be made (plus adding documentation), hence the WIP.

This matches the request for this command:

`gmt basemap -R-20/20/-20/20 -JX5i -Bafg --MAP_FRAME_TYPE=graph-origin -png 1`

yielding

However, if you don't select grid lines (leaving off the **g** or just use default **-B**, you get

where the annotations look a bit naked.  If you don't want any annotations or ticks you may try **-Bws** (lower case turns off annotation) but ticks persist even with -Ba2f0:

Some possibilities:

1. If gridlines are not specified, use grid pen to draw the axis lines where they would normally have gone (e.g., **WS** here). This may be preferred since depending on **-R** you may not be able to select a grid interval that does that.
2. If annotations are not requested (i.e., **-Bws**) then honor that but also turn off tick marks
3. Here, _origin_ meant draw at _x_ = 0 and _y_ = 0, but I can imagine some joker wanting to specify the desired _x_ and _y_ placements.  So are we talking  **MAP_FRAME_TYPE**=_graph-origin[_,_x_/_y_] then, with default 0?

Finally, I can imagine some users wish to have centered annotated axes and leave out the (otherwise clobbered) zero annotation:

as shown in this mock-up.  It would not be too hard to implement that. Do we call that **MAP_FRAME_TYPE**=_graph-origin[_,_x_/_y_]  ?

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -5786,7 +5786,7 @@ void gmt_xy_axis (struct GMT_CTRL *GMT, double x0, double y0, double length, dou
 				dim[1] = g_scale_begin * length - g_ext;
 			PSL_plotsymbol (PSL, x_axis_pos, y, dim, PSL_VECTOR);
 			PSL_setorigin (PSL, x_axis_pos, 0.0, 0.0, PSL_FWD);
-			GMT->current.setting.map_graph_shift = fabs (dim[1] - length);
+			if (GMT->current.setting.map_graph_centered) GMT->current.setting.map_graph_shift = fabs (dim[1] - length);
 		}
 		if (axis == GMT_X)
 			skip_val = GMT->current.setting.map_graph_origin[GMT_X];
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -5801,6 +5801,12 @@ void gmt_xy_axis (struct GMT_CTRL *GMT, double x0, double y0, double length, dou
 	}
 
 	if (side == GMT_AXIS_DRAW) {	/* Just drawing the axis for this one */
+		if (center_reset) {	/* Undo temporary origin shifts needed to move axes */
+			if (horizontal) 
+				PSL_setorigin (PSL, 0.0, -y_axis_pos, 0.0, PSL_INV);
+			else
+				PSL_setorigin (PSL, -x_axis_pos, 0.0, 0.0, PSL_INV);
+		}
 		PSL_setorigin (PSL, -x0, -y0, 0.0, PSL_INV);
 		return;
 	}
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -466,6 +466,8 @@ GMT_LOCAL void gmtplot_linear_map_boundary (struct GMT_CTRL *GMT, struct PSL_CTR
 				GMT->current.setting.map_graph_origin_txt);
 			gmt_M_memset (GMT->current.setting.map_graph_origin, 2, double);
 		}
+		GMT->current.setting.map_graph_shift = 0.0;
+
 	}
 
 	PSL_command (PSL, "/PSL_slant_y 0 def /PSL_slant_x 0 def\n");	/* Unless x-annotations are slanted there is no adjustment. PSL_slant_y may be revised in gmt_xy_axis */
@@ -488,6 +490,8 @@ GMT_LOCAL void gmtplot_linear_map_boundary (struct GMT_CTRL *GMT, struct PSL_CTR
 	if (!GMT->current.map.frame.header[0] || GMT->current.map.frame.plotted_header) return;	/* No title (and optional subtitle) today */
 
 	PSL_comment (PSL, "Placing plot title\n");
+
+	y_length += GMT->current.setting.map_graph_shift;  /* Extra shift (set in gmt_xy_axis) for map title when we have a centered y-axis with vector */
 
 	if (!GMT->current.map.frame.draw || GMT->current.map.frame.side[N_SIDE] <= GMT_AXIS_DRAW || GMT->current.setting.map_frame_type == GMT_IS_INSIDE)
 		PSL_defunits (PSL, "PSL_H_y", GMT->current.setting.map_title_offset);	/* No ticks or annotations, offset by map_title_offset only */
@@ -5782,6 +5786,7 @@ void gmt_xy_axis (struct GMT_CTRL *GMT, double x0, double y0, double length, dou
 				dim[1] = g_scale_begin * length - g_ext;
 			PSL_plotsymbol (PSL, x_axis_pos, y, dim, PSL_VECTOR);
 			PSL_setorigin (PSL, x_axis_pos, 0.0, 0.0, PSL_FWD);
+			GMT->current.setting.map_graph_shift = fabs (dim[1] - length);
 		}
 		if (axis == GMT_X)
 			skip_val = GMT->current.setting.map_graph_origin[GMT_X];
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -5714,7 +5714,7 @@ void gmt_xy_axis (struct GMT_CTRL *GMT, double x0, double y0, double length, dou
 		center_reset = true;
         skip_center_annot = (GMT->current.map.frame.side[W_SIDE] && GMT->current.map.frame.side[S_SIDE]);
 		if (!below) { /* Warn that these are skipped */
-            GMT_Report (GMT->parent, GMT_MSG_WARNING, "gmt_xy_axis: MAP_FRAME_TYPE=graph-centered only expects -B frame selections W, S, w, s, l, b\n");
+            GMT_Report (GMT->parent, GMT_MSG_WARNING, "gmt_xy_axis: MAP_FRAME_TYPE=graph-origin only expects -B frame selections W, S, w, s, l, b\n");
            return;
         }
 	}
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -451,23 +451,29 @@ GMT_LOCAL void gmtplot_linear_map_boundary (struct GMT_CTRL *GMT, struct PSL_CTR
 	y_length = GMT->current.proj.rect[YHI] - GMT->current.proj.rect[YLO];
 
 	if (GMT->current.setting.map_frame_type == GMT_IS_GRAPH && GMT->current.setting.map_graph_origin_txt[0]) {
-		/* Convert the graph origin string coordinates to internals */
-		char txt_a[GMT_LEN128] = {""}, txt_b[GMT_LEN128] = {""};
-		unsigned int n_errors = 0;
-		sscanf (&GMT->current.setting.map_graph_origin_txt[2], "%[^/]/%s", txt_a, txt_b);
-		n_errors += gmt_verify_expectations (GMT, gmt_M_type (GMT, GMT_IN, GMT_X),
-			gmt_scanf_arg (GMT, txt_a, gmt_M_type (GMT, GMT_IN, GMT_X), false,
-			&GMT->current.setting.map_graph_origin[GMT_X]), txt_a);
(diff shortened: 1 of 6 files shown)


joa-quim:
I think only the last example, centered annotations, is of interest (that's what **everybody** would manually draw). If someone really wants the outside annotations, make a second call to ``basemap``. Grid-lines is probably an exception. I can perfectly imagine the wish of centered annotations plus grid-lines

gd-a:
The last panel is definitely the best looking one (IMO). 
> 

But I'd keep the 0, especially foreseeing this remark : 
> but I can imagine some joker wanting to specify the desired x and y placements

After all, the "first" annotation is supposed to be easy to remove thanks to `B+e` option (it reminds me [this](https://github.com/GenericMappingTools/gmt/issues/6292) BTW.

PaulWessel:
Annotating the center is asking for trouble.  If one used a non 0/0 value then it double always overprint.  Here is 0/0:



I think from all the other annotations there is no need to do something complicated at the center since those values, whatever they are, could be spelled out in a figure caption or text etc., or just inferred from the other annotations.

PaulWessel:
So, based on this limited feedback, the leading candidate is **--MAP_FRAME_TYPE**=_graph-centered_[,x/y] and it would only result in the last map as is.

gd-a:
What about slightly shifting the annotation ? That would solve the grid overlap too

PaulWessel:
> What about slightly shifting the annotation ? That would solve the grid overlap too

This would need to involve font size, absolute values, spacing etc.  It is not a fun thing to do.  I think leaving it out is a good start.  There is always text if people cannot figure out what those coordinates are.

gd-a:
> There is always text if people cannot figure out what those coordinates are.
True. What about the `B+el` option ?

PaulWessel:
Well, -B+e has no knowledge of any center coordinate to skip.  One can image a -B+ec but the point is that nobody would want an annotation that prints on top of the axis and the other annotation, which brings us back to the "can we just shift it a bit".  There is real agony in determine "a bit".

gd-a:
> agony

Alright alright. You're the one doing the work here 🥇

gd-a:
I'm curious about the result with this showcase : 

```
#!/bin/bash

gmt begin test png

	vecsta=$(gmt math -Q -3 PI MUL =)
	vecsto=$(gmt math -Q 3 PI MUL =)
	vecste=$(gmt math -Q 1 32 DIV =)

        # x
	gmt math -T$vecsta/$vecsto/$vecste T -o0 = x.txt
       # y = sin(x)
	gmt math -Q x.txt SIN = y.txt 
        #y2 = A cos(x+B)
	gmt math -Q x.txt PI 3 DIV ADD COS 4 PI MUL 10 DIV MUL = y2.txt

        # Shading for red<blue
	gmt math -Q y.txt y2.txt LE 0 NAN y.txt MUL = y_c.txt
        # Shading for red>blue
	gmt math -Q y.txt y2.txt GE 0 NAN y2.txt MUL = y2_c.txt

        # All in one file (for -Sb+base option)
	paste x.txt y.txt y2.txt y_c.txt y2_c.txt > file.txt

        # Graph-paper
	gmt basemap -R-10/10/-2/2 -JX10c -Bxafpigpi -Byafg -BW+t"Relative anomaly" --MAP_FRAME_TYPE=graph-centered
 
        # Plot sines
	gmt plot file.txt -i0,1 -Wfat,blue -l"@[\sin(x)@["
	gmt plot file.txt -i0,2 -Wfat,red -l"@[\frac{4\pi}{10}\cos(x+\frac{\pi}{3})@["

        # Plot shades
	gmt plot file.txt -i0,3,2 -Gred@90 -Sb${vecste}q+b -l"red > blue"
	gmt plot file.txt -i0,4,1 -Gblue@90 -Sb${vecste}q+b -l"red < blue"

gmt end show
```

PaulWessel:
Give it a try.  I have updated the docs.  For various reasons the syntax is

--**MAP_FRAME_TYPE**=_graph-centered_[**+o**_x/y_] 

My complaints about unwanted ticks were silly since one can just select axis **l** and **b** to just get the vectors and **w** and **s** to add ticks.

gd-a:
> Give it a try. I have updated the docs.

I don't see it in [dev doc](https://docs.generic-mapping-tools.org/dev/gmt.conf.html#term-MAP_FRAME_TYPE).

```
> gmt --version
6.5.0_8bbc44d_2022.06.25

> bash Untitled-1.sh
basemap [WARNING]: Bad fraction, error = 0.00128068

> > 
#!/bin/bash

gmt begin test png

	vecsta=$(gmt math -Q -3 PI MUL =)
	vecsto=$(gmt math -Q 3 PI MUL =)
	vecste=$(gmt math -Q 1 32 DIV =)

        # x
	gmt math -T$vecsta/$vecsto/$vecste T -o0 = x.txt
       # y = sin(x)
	gmt math -Q x.txt SIN = y.txt 
        #y2 = A cos(x+B)
	gmt math -Q x.txt PI 3 DIV ADD COS 4 PI MUL 10 DIV MUL = y2.txt

        # Shading for red<blue
	gmt math -Q y.txt y2.txt LE 0 NAN y.txt MUL = y_c.txt
        # Shading for red>blue
	gmt math -Q y.txt y2.txt GE 0 NAN y2.txt MUL = y2_c.txt

        # All in one file (for -Sb+base option)
	paste x.txt y.txt y2.txt y_c.txt y2_c.txt > file.txt

        # Graph-paper
	gmt basemap -R-10/10/-2/2 -JX10c -Bxafpigpi -Byafg -BWS+t"Relative anomaly" --MAP_FRAME_TYPE=graph-centered+o0/0 --MAP_GRID_PEN=lightgray
 
        # Plot sines
	gmt plot file.txt -i0,1 -Wthick,blue -l"@[\sin(x)@["
	gmt plot file.txt -i0,2 -Wthick,red -l"@[\frac{4\pi}{10}\cos(x+\frac{\pi}{3})@["

        # Plot shades
	gmt plot file.txt -i0,3,2 -Gred@90 -Sb${vecste}q+b -l"red > blue"
	gmt plot file.txt -i0,4,1 -Gblue@90 -Sb${vecste}q+b -l"red < blue"

gmt end show
```

PaulWessel:
Not merged - this is WIP branch only. - need to build that branch.

gd-a:
> need to build that branch

I'm a simple man.

PaulWessel:
Your example looks like this:

test

PaulWessel:
Which tells me that if only one axis is selected we do not have to block the labeling of the origin since there is no overprint.

gd-a:
> if only one axis is selected

Indeed. Do you throw an error if Joker asks for `-BWSEN` or similar ?

PaulWessel:
Not yet, but silently ignore anything but WwlSsb

PaulWessel:
Still need to update the example plot in gmt.conf of the various basemaps to include the graph-centered.  @gd-a example with both axis looks like this:

test

I am unsure if we should automatically add to the title offset (the length of the extension) or if the user should use --**MAP_TITLE_OFFSET**.

gd-a:
> silently ignore anything but WwlSsb

The secret to peace of mind.


It looks neat ! (Well, except for the ugly plot lines)
I'd add the offset, else users would have to declare it every time, which have a tendency to clutter the code.

PaulWessel:
Now with auto shift:
test

Esteban82:
I wonder if we should add the arrows to the negative side of the axis?

image

PaulWessel:
The arrow means positive direction so I dont think you want that.

Esteban82:
> The arrow means positive direction so I dont think you want that.

Ok.

gd-a:
Does ˋ-JX-5` gives an arrow toward negative values or just flips the axis??

PaulWessel:
I am thinking **graph-origin** might be better than **graph-centered**.  Agree?

PaulWessel:
> Does ˋ-JX-5` gives an arrow toward negative values or just flips the axis??

Arrow always points in the positive axis direction, so now the arrow is on the left.

gd-a:
> Agree?

It's more accurate indeed.

PaulWessel:
Documentation and code has been updated to use graph-origin and the figure and script in gmt.conf has been expanded to include the graph-origin basemap.  I think this PR is now ready for final review:

GMT_map_frame_type

joa-quim:
Agree that **graph-origin** is better (a little shorter).

Another thing. Your example figure shows that the figures are small the fonts are still too big.

PaulWessel:
Agree the fonts are too big here.  Maybe it is a subplot thing? Anyway, separate issue from graph-origin.

joa-quim (approved):
Blind aproval.

PaulWessel:
Given Joaquim's blind approval, it would be good if @seisman @Esteban82 and @meghanrjones give it a try first before I merge.

seisman:
Do you think it makes more sense to have the default origin at the center of the x- and y-range. For example, for `-R10/30/-20/-10`, the default origin is `+o20/-15`.
```
gmt basemap -JX5c -R10/30/-20/-10 -Baf -BWS --MAP_FRAME_TYPE=graph-origin -png map
```
map

src/gmt_plot.c
```diff
@@ -5676,6 +5702,23 @@ void gmt_xy_axis (struct GMT_CTRL *GMT, double x0, double y0, double length, dou
...
+		else if (x_axis_pos > GMT->current.map.width) x_axis_pos = GMT->current.map.width;
+		y_axis_pos = gmt_y_to_yy (GMT, GMT->current.setting.map_graph_origin[GMT_Y]);  /* Convert graph origin y to inches on the page */
+		if (y_axis_pos < 0.0) y_axis_pos = 0.0;	/* Never move past south or north */
+		else if (y_axis_pos > GMT->current.map.height) y_axis_pos = GMT->current.map.height;
+		x_axis_pos -= x0;	/* Since we do PSL_setorigin below we must counter act that for these values */
+		y_axis_pos -= y0;
+		center_reset = true;
+        skip_center_annot = (GMT->current.map.frame.side[W_SIDE] && GMT->current.map.frame.side[S_SIDE]);
+		if (!below) { /* Warn that these are skipped */
+            GMT_Report (GMT->parent, GMT_MSG_WARNING, "gmt_xy_axis: MAP_FRAME_TYPE=graph-centered only expects -B frame selections W, S, w, s, l, b\n");
```
seisman, line 5717:
```suggestion
            GMT_Report (GMT->parent, GMT_MSG_WARNING, "gmt_xy_axis: MAP_FRAME_TYPE=graph-origin only expects -B frame selections W, S, w, s, l, b\n");
```

PaulWessel:
> Do you think it makes more sense to have the default origin at the center of the x- and y-range. For example, for `-R10/30/-20/-10`, the default origin is `+o20/-15`

No, I think that would be a problem.  Lets say I use -R-10/20/-20/10.  I would want the axes at the origin, but your default would place them at (5,-5) which is useless.  So now I would have to add +o0/0 to make sure every time.  I think it is better the other way, that you should have to do +o20/-15 instead.

Perhaps a getter solution would be an optional **+oc** for "Center the axes on the given domain", with the default still being the origin at (0,0)?

seisman:
> > Do you think it makes more sense to have the default origin at the center of the x- and y-range. For example, for `-R10/30/-20/-10`, the default origin is `+o20/-15`
> 
> No, I think that would be a problem. Lets say I use -R-10/20/-20/10. I would want the axes at the origin, but your default would place them at (5,-5) which is useless. So now I would have to add +o0/0 to make sure every time. I think it is better the other way, that you should have to do +o20/-15 instead.
> 
> Perhaps a getter solution would be an optional **+oc** for "Center the axes on the given domain", with the default still being the origin at (0,0)?

Sounds good to me.

PaulWessel:
Please check again, @seisman.  The +oc has been implemented and I fixed the lack of reset for unannotated axes.

seisman (approved):

PR #6826, opened 2022-06-26
Fix sign error in FAA slab correction

PaulWessel:
The -Ff+s option in **gravfft** adds back in the slab that was removed when we find the mid-level of our topography and remove it.  However, The sign of this level matters: Contrast these two cases, both with a water depth W = 5000 m:

1. Input z is a positive bump on the seafloor, varying from 0 to 1000 m.  We first subtract sea level and now z goes from -5000 to -4000.  We find mid level to be -4500 and keep the absolute value z_level = 4500.  It is used in Parker exponent exp (-k*z_level) for upward continuation. We then compute slab thickness as h = W - z_level = 5000 - 4500 = 500m. We need to add the 500m slab gravity (a positive value) to the undulations obtained by the FFT to make sure all values are positive.
2. Input z is a negative bump on the seafloor, varying from 0 to -1000 m.  We first subtract sea level and now z goes from -6000 to -5000.  We find mid level to be -5500 and keep the absolute value z_level = 5500. It is used in Parker exponent exp (-k*z_level) for upward continuation. We then compute slab thickness as h = W - z_level = 5000 - 5500 = -500m. We need to add the -500m slab gravity (a negative value) to the undulations obtained by the FFT to make sure all values are negative.

In master, h is computed as absolute value of (W - z_level), hence both of the above cases end up with a positive correction due to a 500 m slab, yet the second case needs it subtracted.  Hence the sign of W - z_level needs to be retained.

No tests were affected, apparently.  We need more tests, obviously.

### DIFF
--- a/doc/rst/source/explain_fft.rst_
+++ b/doc/rst/source/explain_fft.rst_
@@ -16,9 +16,9 @@
         dimensions >= data which optimize speed and accuracy of FFT. If FFT
         dimensions > grid file dimensions, data are extended and tapered to zero.
 
-    Control detrending of data: Append modifiers for removing a linear trend:
+    Control detrending of data: Append modifiers for removing a linear trend. Consult module documentation for the default action:
 
-        **+d**: Detrend data, i.e. remove best-fitting linear trend [Default].
+        **+d**: Detrend data, i.e. remove best-fitting linear trend.
 
         **+a**: Only remove mean value.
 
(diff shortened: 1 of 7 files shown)


joa-quim (approved):

joa-quim:
One other thing. The man says

> -N+d: Detrend data, i.e. remove best-fitting linear trend [Default].

but it better not be the default because we want

> +h: Only remove mid value, i.e. 0.5 * (max + min).

otherwise the slab would be an wedge.

PaulWessel:
I think the default is +h but we are seeing the standard -N docs. Fix
later; on family zoom

Paul Wessel
Professor, Department of Earth Sciences
(808) 956-4778 | ***@***.***
http://www.soest.hawaii.edu/earthsciences

PR #6844, opened 2022-07-02
Improve subplot setting of -X -Y

PaulWessel:
See this forum [post](https://forum.generic-mapping-tools.org/t/y-axis-label-clipped-when-ticks-are-too-big/3072/8) for details.  Basically, If the very first GMT plot command has explicit **-X -Y** settings then we use those settings as given else (if we are in modern mode) we use 40 inches.  However, **subplot** would always add -X1i -Y1i to the internal **plot** or **basemap** commands even when there was no need, and then those were interpreted as hardwired settings and the 40 inches offset got reset to (1,1) inch.

This PR tries harder by checking if **-X -Y** was in fact given and if there actually is a shift due to clearance.


seisman (approved):
Works for me.

PR #6849, opened 2022-07-04
Fix documentation and handing of -J in histogram

### LINKED ISSUE #6847: histogram behaves differently in classic vs modern
In modern this works despite the man saying that **-J** is mandatory.
```
gmt histogram   @v3206_06.txt -R-6000/0/0/30 -Bx+l"Topography (m)" -By+l"Frequency"+u" %" -BWSne+t"Histograms"+glightblue -Gorange -W1p -Z1 -T250 -png lixo
```
in classic 

```
gmt pshistogram @v3206_06.txt -R-6000/0/0/30 -Bx+l"Topography (m)" -By+l"Frequency"+u" %" -BWSen+t"Histograms"+glightblue -Gorange -W1p -Z1 -T250 -P > lixo.ps

pshistogram [ERROR]: Option -J: Only Cartesian scaling available in this module.
```
and even after adding a -J it still doesn't automatic annotations/ticks but modern version does it.
So, modern:

- Automatic -J and -B

Clsssic:

- Needs -J & -B

PaulWessel:
See #6847 for background.  The handling of **-J** in **histogram** is special since only a Cartesian scaling or dimensions is allowed (i.e., map projections are not allowed).  Hence we need specific checks.  This PR does the following:

1. Updates the documentation for **histogram** to state that **-J** is optional in modern mode (it is required for classic mode).
2. Gives an error if no **-J** given in classic mode.
3. Gives an error if an unsuitable projection was specified via **-J**.

Closes #6847.

### DIFF
--- a/doc/rst/source/histogram.rst
+++ b/doc/rst/source/histogram.rst
@@ -12,7 +12,7 @@ Synopsis
 
 .. include:: common_SYN_OPTs.rst_
 
-**gmt histogram** [ *table* ] |-J|\ **x**\|\ **X**\ *parameters*
+**gmt histogram** [ *table* ]
 |-T|\ [*min/max*\ /]\ *inc*\ [**+i**\|\ **n**] \|\ |-T|\ *file*\|\ *list*
 [ |-A| ]
 [ |SYN_OPT-B| ]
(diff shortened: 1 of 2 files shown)


joa-quim (approved):
A little confusion in this is that in classic what is needed is the **size** not the projection that is only the linear one.

PaulWessel:
> A little confusion in this is that in classic what is needed is the **size** not the projection that is only the linear one.

I do not understand this sentence.  You can give -JX or -Jx in either mode.

joa-quim:
Yes, but if it wasn't for the size even classic could spare user to state the projection. That is, to use a default one, let's say, -JX

PaulWessel:
Yes, but would break the separation between church and state, eeh classic and modern.

PR #6863, opened 2022-07-08
CI: Install dvc using homebrew on macOS

seisman:

dvc is available in homebrew, so we can install it using homebrew, rather than pip.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,7 +4,7 @@
 name: Docs
 
 on:
-  # pull_request:
+  pull_request:
   push:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -4,7 +4,7 @@
 name: Docs
 
 on:
-  # pull_request:
+  pull_request:
```
```suggestion
  # pull_request:
```

.github/workflows/tests.yml
```diff
@@ -4,7 +4,7 @@
 name: Tests
 
 on:
-  # pull_request:
+  pull_request:
```
```suggestion
  # pull_request:
```

PaulWessel (approved):

PR #6868, opened 2022-07-09
CI: Skip the installation of pngquant on Windows CI

seisman:

`choco install pngquant` doesn't work due to unknown network connection issues of GitHub Actions, which makes the docs building CI fail on Windows.

In this PR, `choco install pngquant` is disabled to make the CI pass again.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,7 +4,7 @@
 name: Docs
 
 on:
-  # pull_request:
+  pull_request:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -4,7 +4,7 @@
 name: Docs
 
 on:
-  # pull_request:
+  pull_request:
```
```suggestion
  # pull_request:
```

PaulWessel (approved):

Issue #5595, opened 2021-08-05
Update grdimage example to use recommended syntax


maxrjones:
**Description of the problem**

The current documentation for grdimage includes an example that uses three grids as input:

> To create an false color plot from the three grid files red.nc, green.nc, and blue.nc, with
> linear scaling at 10 inch/x-unit, tickmarks every 5 units:
> 
> `gmt grdimage red.nc green.nc blue.nc -Jx10i -B5 -pdf rgbimage`

The use of grdimage with three grid files as input is no longer recommended/documented (following discussion in https://github.com/GenericMappingTools/gmt/pull/4044).

It would be good to update this example to include a call to grdmix first to produce the image that is provided to grdimage.

PR #6878, opened 2022-07-12
Remove one deprecated example in grdimage manpage

seisman:

We already have a similar example in the grdmix manpage, so we don't need to keep the deprecated example in the grdimage manpage.

Fixes https://github.com/GenericMappingTools/gmt/issues/5595.


PaulWessel (approved):
Sorry, missed this somehow.

PR #6890, opened 2022-07-16
Add legend symbol for confidence lines

PaulWessel:

Both **plot** and **plot3d** can plot lines with build-in confidence bands.  This is done via the option **-L+b**|**d**|**D** that reads extra column(s) to build the envelope around the line (drawn with **-W**_pen_) and the envelope is painted via **-G**_fill_.  The envelope may also take an outline pen via **-L**'s **+p**_pen_ modifier.  However, we had no way to handle this in a legend.  The sad and painful process was to draw them as two separate things, then struggle a bit with a manual legend setup to draw a line on top of the rectangle for the envelope fill.

This PR detects the **-L** situation and basically issues to auto calls for legend entries, the first using the special symbol **L** for a rectangle that should not see a legend position change after it, and then the line as the second symbol which then overprints the rectangle.  With this PR, simple examples like this:

```
#!/bin/bash
# Test CL legend
gmt math -T0/3/0.1 -N3/0 T 2 POW -C2 10 DIV 0.2 ADD = line.txt
gmt begin CL
	gmt plot -R-0.5/3.5/-1/11 -JX15c -B -L+d -W1p -Gpink line.txt -l"My trend and 2@~s@~ bounds"+jTL
gmt end show
```
will give a legend entry like this:

CL


joa-quim (approved):

seisman:
Looks great, but why is the rectangle a little shorter than the line?

PaulWessel:
Couple of reasons:

- I happen to like it over the alternative
- This required no code changes for the rectangle, it is the same used for polygons

To make it the same width as the line would require more coding of a special case.  Seems not worth it to me.

seisman:
> Couple of reasons:
> 
> * I happen to like it over the alternative
> * This required no code changes for the rectangle, it is the same used for polygons
> 
> To make it the same width as the line would require more coding of a special case. Seems not worth it to me.

Sounds reasonable to me.

Issue #6889, opened 2022-07-16
Remove grid registration annoying warning


joa-quim:
Grid registration should not matter for ``grd2cpt``
```
julia> C,cdf = grd2cpt("@earth_relief_10m", nlevels="11+c+f");
GMT [WARNING]: Remote dataset given to a data processing module but no registration was specified - default to gridline registration (if available)
```

PaulWessel:
True. Would need to intercept that deep in the API.  Other grdmodules that fall in the same category?

joa-quim:
Can't remember others now but maybe just move this warning one level up.

PR #6891, opened 2022-07-17
Let grd2cpt and grd2kml not care about remote registration

PaulWessel:
See #6889 fro background.  Since **grd2kml** is a plotter (just not making _PostScript_ but kml) and **grd2cpt** does not produce a grid we will let these two not care about the missing registration.
Note: Requires @joa-quim for testing.


joa-quim (approved):
Yep, those are silent now.

PR #6894, opened 2022-07-17
Enhance gravprisms -D with +c for computing contrasts

PaulWessel:
It occurred to me that it is convenient to allow **-D** to be used not only to set a constant density contrast for all prisms, but with a modifier (**+c**) instead compute the density _contrast_ between the variable prism densities and the constant given in **-D.**  Otherwise, I need to compute two sets of prisms just to be able to display the densities and to compute gravity using a contrast with sea water.  This PR adds the **+c** modifier.  It basically reverts very recent behavior implemented after 6.4 so it has no backwards complications. FYI, @joa-quim.

### DIFF
--- a/doc/rst/source/supplements/potential/gravprisms.rst
+++ b/doc/rst/source/supplements/potential/gravprisms.rst
@@ -15,7 +15,7 @@ Synopsis
 **gmt gravprisms** [ *table* ]
 [ |-A| ]
 [ |-C|\ [**+q**][**+w**\ *file*][**+z**\ *dz*] ]
-[ |-D|\ *density* ]
+[ |-D|\ *density*\ [**+c**] ]
 [ |-E|\ *dx*\ [/*dy*] ]
 [ |-F|\ **f**\|\ **n**\ [*lat*]\|\ **v** ]
 [ |-G|\ *outfile* ]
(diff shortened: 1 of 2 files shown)


joa-quim (approved):

PaulWessel:
Note FYI: I also added a modifier **+r**_rhor_ to **-D**.  The reason is as follows:

I have a variable density grid used with topography to drive flexure. This is done via **-H**rho.grd.  However, the density structure `rho(x,y)` is not perfectly known and I will be running numerical experiments to modify the rho.grd grid by a small scalar _1+f_ (with _f_ in the ±0.2 range and determine the _f_ that minimizes some rms).  So, multiplying _rho(x,y)_ by _f_ changes the mean load density. When the infill density and the load density are set separately then we do the Wessel [2001;2016] approximation which requires these two densities, and now the rho load density will change.  However, I wish to keep the same root density (and infill density) in all these experiments, hence I wish to pass that constant root density and the scaled rho.grd so that the approximation stays the same by just adjusting the topography by _1+f_.  That way the flexural response to me changing _1+f_ is given by _(1+f)_ times _w(x,y)_.

Anyway, I suspect I am the only one doing this, and it does not affect any tests.

PR #6901, opened 2022-07-19
Remove a trailing dot in the -x option syntax

seisman:

Run `gmt grdimage`, you will see a trailing dot in the `-x` option syntax.
```
[-x[[-]<n>]] .
```
This PR removes it.


PaulWessel (approved):
As Joaquim would say, you could just commit that directly to master.

PR #6913, opened 2022-07-25
Changes to support longopts on pswiggle and grdimage.

### LINKED ISSUE #6912: pswiggle+grdimage long-options support

Add pswiggle+grdimage long-options support.

rbdavis:

Changes to support longopts on pswiggle and grdimage.

Fixes #6912

**Reminders**


src/grdimage.c
```diff
@@ -35,6 +35,36 @@
...
 #define THIS_MODULE_OPTIONS "->BJKOPRUVXYfnptxy" GMT_OPT("Sc") GMT_ADD_x_OPT
 
+static struct GMT_KEYWORD_DICTIONARY module_kw[] = { /* Local options for this module */
+	/* separator, short_option, long_option,
+	          short_directives,    long_directives,
+	          short_modifiers,     long_modifiers */
+/* ?? -A not possible because of = usage, e.g., -Aout_img=driver ? */
+	{ 0, 'C', "cpt",
+                  "",                  "",
+                  "h,i,u,U",           "hinge,zinc,tometers,frommeters" },
```
PaulWessel, line 45: `                  "h,i,u,U",           "hinge,zinc,tometers,frommeters" },
`

Reading the documentation on this I think perhaps inverting these are better?  E.g., use _tounit_ and _fromunit_ since they both need to take a unit as argument . So I think `--cpt:=mine.cpt+fromunit=k` might be clearer than `--cpt:=mine.cpt+tometers=k` ?

PaulWessel (commented):
Very nice, see question regarding the -C

src/gmt_common_longoptions.h
```diff
@@ -107,21 +107,26 @@
...
-                "a,c,f,s",               "byset,column,byfile,bysegment" },    /* Actually -qo */
+                "a,c,t,s",               "byset,column,bytable,bysegment" },
     {   0, 'r', "registration",
                 "g,p",                   "gridline,pixel",
                 "",                      "" },
     {   0, 's', "skiprows",
                 "",                      "",
                 "a,r",                   "any,reverse" },
-    {   0, 't', "transparency",          "", "", "", "" },
+    {   0, 't', "transpercent",
```
joa-quim, line 122: ``transpercent``? Firs time I see this word. I don't think it's better then ``transparency`` or ``alpha``.

PaulWessel: I missed that.  Let's stick with transparency here, @rbdavis.

rbdavis: Paul originally used 'transparency' for the -t common option. This was fine by itself, but now working in grdimage I see there is a -Q option that also sets a transparency parameter. These need to be distinguished, I think, so I (i) changed the -t to 'transpercent' to indicate that it specifies a transparency percentage, and (ii) used 'transnode' for grdimage -Q to indicate that it controls per-node transparency. I think that 'transparency' is probably too general a term to be used in either case.

PaulWessel: I think it is better to change what -Q means rather than -t.  -Q does related to transparency in the sense nodes that are NaN will be see-through.  But it is therefore on/off, not like transparency which can be partial.  So maybe --nanmask or just --mask is better here.

src/grdimage.c
PaulWessel: Yes, since tounit=k would then mean change from meters to km, consistent with the meaning you had for frommeters.

src/gmt_common_longoptions.h
rbdavis: OK.

PaulWessel (approved):

Issue #6909, opened 2022-07-23
When to keep or remove a grids default CPT argument


PaulWessel:
As I was testing a **grdmath** macro to obtain _Parsons and Sclater_ [1977] depths from an age grid, I noticed something odd.  If I run

` gmt grdmath -R200/240/40/60 @earth_age_01m_g PS77 = depth_ps77.grd`

then the grid is correct with a range in depths from 2535 to 5356 meters.  However, grdinfo says a bit more:

```
...
depth_ps77.grd: v_min: 2535.00390625 v_max: 5356.30371094 name: z
depth_ps77.grd: scale_factor: 1 add_offset: 0
depth_ps77.grd: Default CPT: @earth_age.cpt
```
 
Because the default CPT inherited from the age grid is passed along, and that CPT is set to go from 0 to 340 Myr, trying to make a simple map fails:

![bad](https://user-images.githubusercontent.com/26473567/180605671-f59548c6-d82e-47b3-a82e-4a209f96a05b.png)

I have to specify a master CPT for this to work:

![good](https://user-images.githubusercontent.com/26473567/180605692-417388fa-c407-440a-9dcd-c6fe0db6026e.png)

It is not immediately obvious to me when we should wipe the incoming CPT.  Well, it is in the sense that we are concerting unit from Myr to m and thus the old CPT should be irrelevant, but **grdmath** does not know that.  Even a simple operation that does not change the units will fail due to the new range:

`gmt grdmath -R160W/120W/40/60 @earth_age_01m_g/ 500 SUB = t.grd`

would not plot a black ocean map.  So there seems to be two possibilities here:

1. **grdmath** should always wipe the default CPT setting in the output grid
2. **grdmath** should leave it along and we write a Note in the documentation to explain the problem.

I think we have to go with (1) here.  It is a relatively obscure problem and people will not read that section and just be frustrated why their basic plot is blank.  I think this is the only module where this sort of thing can happen (unit change or wild range change). Any opposing voices, @GenericMappingTools/core ?

joa-quim:
This is a hint on a more general problem of *metadata propagation* which a not easy problem at all.
Solution 1. is the only one to go here.

PaulWessel:
[...]

PR #6914, opened 2022-07-25
Must erase any input grid CPT in grdmath

PaulWessel:
See https://github.com/GenericMappingTools/gmt/issues/6909 for background.

This PR introduces a hidden parameters that some modules (right now just **grdmath**) can set to prevent a grid deriving from a remote data set to inherit (and save) the CPT name in the grid file.

Closes https://github.com/GenericMappingTools/gmt/issues/6909.

### DIFF
--- a/src/gmt_nc.c
+++ b/src/gmt_nc.c
@@ -1035,7 +1035,7 @@ L100:
 
 		/* Define z variable. Attempt to remove "scale_factor" or "add_offset" when no longer needed */
 		gmtnc_put_units (ncid, z_id, header->z_units);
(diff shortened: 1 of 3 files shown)


joa-quim:
I think it would be better to introduce a new struct for the metadata propagation (or not) instead of that new ``bool ignore_remote_cpt;`` Otherwise for each new control a new member will have to be added to the ``GMTAPI_CTRL``

PaulWessel:
OK, so a sub-structure of GMTAPI_CTRL with some name that would currently only have one bool variable?

joa-quim:
Yes.

joa-quim (approved):

PR #6916, opened 2022-07-25
Let grdmath control what happens to grid CPT settings

PaulWessel:
See #6909 to which this is a follow up.  It adds the new **-C** option that can be used to keep or update the output grid's default CPT setting.  By default, those CPTs are wiped on output.


Esteban82:
I try these commands and I got. 

It is ok, right?

```
gmt grdmath -R200/240/40/60 -Cturbo @earth_age_01m_g PS77 = depth_ps77.grd
gmt grdimage depth_ps77.grd -png ps77
```

ps77

Esteban82 (approved):

PR #6920, opened 2022-07-26
CI: Install gdal with default features

seisman:
Recent tests fail on Windows, because the gdal package installed by vckpg doesn't support features like hdf5 and png by default.

This PR fixes the issue by install gdal with default features, i.e., `vcpkg install gdal[core,tools,default-features]`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,7 @@
 name: Tests
 
 on:
-  pull_request:
+  # pull_request:
   push:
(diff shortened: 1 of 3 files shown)


PaulWessel (approved):

.github/workflows/tests.yml
```diff
@@ -4,7 +4,7 @@
 name: Tests
 
 on:
-  # pull_request:
+  pull_request:
```
```suggestion
  # pull_request:
```

PR #6921, opened 2022-07-26
Update images to show GDAL is required

PaulWessel:
These are old images made in illustrator I think, I kludged the solid lines for GDA in photoshop.  Note the third case you mentioned (GMT4_*) is kept as is since GMT 4 had an optional dependency on GDAL.


anbj (approved):

anbj:
Looks good!

PR #6930, opened 2022-08-03
Let grdimage handle mask grids

PaulWessel:
If you create a mask grid (NaN and 1) and want to see how it looks in **grdimage** you get this annoyance:

```
gmt grdimage E_obl_inside_mask.grd -B -pdf t
grdimage [ERROR]: Failed to read CPT (null).
psconvert [ERROR]: Unable to decode BoundingBox file /Users/pwessel/.gmt/sessions/gmt_session.11534/psconvert_30470c.bb (maybe no non-white features were plotted?)
```
and a failed plot.   The reason is that _zmin_ = _zmax_ = 1 and no CPT can be crafted.  Since this is a common situation, this PR allows for this by checking if

- We have a grid with NaNs
- The rest of the data are all 1.0

If so we have a NaN mask and we can then set the internal _zmax_ to 2 so that we get a map when no CPT is given.

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -1631,6 +1631,10 @@ EXTERN_MSC int GMT_grdimage (void *V_API, int mode, void *args) {
 			double zmin = Grid_orig->header->z_min, zmax = Grid_orig->header->z_max;
 			char *cpt = gmt_cpt_default (API, Ctrl->C.file, Ctrl->In.file, Grid_orig->header);
 			grdimage_reset_grd_minmax (GMT, Grid_orig, &zmin, &zmax);
+			HH = gmt_get_H_hidden (Grid_orig->header);
+			if (HH->has_NaNs == GMT_GRID_HAS_NANS && doubleAlmostEqual (zmin, 1.0) && doubleAlmostEqual (zmax, 1.0)) {	/* Mask grid, just nudge max to 2 */
+				zmax = 2.0;	/* Otherwise we get annoying warning */
+			}
 			if ((P = gmt_get_palette (GMT, cpt, GMT_CPT_OPTIONAL, zmin, zmax, Ctrl->C.dz)) == NULL) {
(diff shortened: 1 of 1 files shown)


Esteban82 (approved):
Approving without testing (in case you need it).
Let me know if you have a grid to test it.

PaulWessel:
Sure:

```
gmt grdmath -R-50/50/-50/50 -I1 0 0 CDIST 25 LE 0 NAN = mask.grd
gmt grdimage mask.grd -B -png mask
```

Esteban82:
> Sure:
> 
> ```
> gmt grdmath -R-50/50/-50/50 -I1 0 0 CDIST 25 LE 0 NAN = mask.grd
> gmt grdimage mask.grd -B -png mask
> ```

Thanks. I will merge it because it works.

seisman:
Why not increase zmax by one for all zmin=zmax grids?

PaulWessel:
Hm, yes I see where you are going with this.  It is very possible that a user could create a grid that is constant, say 4.67.  Then what? Well, we then get the annoying message.  But having a constant grid is not a sin that requires lecturing by us.  SO I agree, we should extend this to any grid, mask or not, and without having more info, just add 1.

PaulWessel:
Assuming that is what you meant, I will make another PR that is more general and does this inside the gmt_get_pallete instead.

seisman:
> It is very possible that a user could create a grid that is constant, say 4.67. Then what? Well, we then get the annoying message. But having a constant grid is not a sin that requires lecturing by us. SO I agree, we should extend this to any grid, mask or not, and without having more info, just add 1.

Yes, that's exactly what I mean. But adding 1 may not a good idea for small grid values like zmin=zmax=1.0e-5. Perhaps zmax = 2 * zmin is better?

PaulWessel:
Yes, but depends on the sign I think

PR #6931, opened 2022-08-03
Handle automatic CPT for constant grids

PaulWessel:
This PR overwrites the changes in #6930 following the discussion therein.  A mask grid is just a special case of a constant grid and such grids are allowed.  However, without upgrading the automatic CPT creation for grids with zero range we run into trouble.

This PR always produces a valid CPT for grids with no range.


seisman (approved):
Looks good.

Issue #6932, opened 2022-08-03
Suppress batch default handling of moving products out


PaulWessel:
**Description of the desired feature**

The **batch** module runs jobs in parallel and may produce products with unique names, similar to that of **movie**.  When a single job completes it tries to move any named products up to the top directory, which for my case may give scripts that end with

`mv -f ${BATCH_NAME}.* /Users/pwessel/Dropbox/UH/ACTIVE/PROJECTS/OXFORD2022/EMPERORS`

However, not every possibly **batch** scenario will produce single products that are called ${BATCH_NAME}.* (here, EFLX3_1000.*).  In my case, **batch** is given a file via **-T** with 3 columns of parameters (two densities and an elastic thickness) and the output of the script is a single grid that has those parameters in the file name, since `flex_2510_2930_15.0k.grd` is much more helpful to me (I can see the densities and the elastic thickness I used via the file name) than `EFLX3_1433.grd`, where I would have to run **grdinfo** to learn which solution this is.  Thus, there are clearly cases where the single number naming is not helpful (whereas it is _always_ used for movies).

I will think more about this but some initial thoughts are listed here:

1. We could add an option **-C** that does _not_ issue those move commands and then the user is on the hook to do something specific in the main script.  In my case I write directly to an outside final directory, so no need to move.
2. Alternatively, we add an option that takes a format statement and then **BATCH_NAME** and others are formed from the template and the various columns in the -Tfile. In my case, that would mean I pass something like `-Fflex_%4.0lf_%4.0lf_%4.1lfk` and then this template is used with the three items.  If the file had trailing texts etc then it would use %s instead.

I think the second one is getting a bit too complicated and will experiment with the first for now, but I do like the power of the 2nd option.

Maybe @Esteban82 has thoughts on this?

PaulWessel:
Also, since **batch** was based on **movie** there is the assumption that a job only produces products with the same name but may have different extensions.  That may not be true in general, hence the move command needs some updates anyway.  In the case above it should probably be

`mv -f ${BATCH_NAME}* /Users/pwessel/Dropbox/UH/ACTIVE/PROJECTS/OXFORD2022/EMPERORS`

i.e.,

- Not insist that all products must have the same complete filename and only vary by extension.  What if I want to make two grids?
- Not insist there is an extension at all.  What if I want to make a file called junk_01234?

[...]

PR #6933, opened 2022-08-03
Let batch -D drop the mandatory moving of product files

PaulWessel:
See #6932 for discussion.  This PR implements a **-D** option that stops any copying of batch products to the top directory.  This solution does not preempt further consideration of the other suggestion which is to generalize the naming of product files via a **-F**_template_ option.

### DIFF
--- a/doc/rst/source/batch.rst
+++ b/doc/rst/source/batch.rst
@@ -15,6 +15,7 @@ Synopsis
 **gmt batch** *mainscript*
 |-N|\ *prefix*
 |-T|\ *njobs*\|\ *min*/*max*/*inc*\ [**+n**]\|\ *timefile*\ [**+p**\ *width*]\ [**+s**\ *first*]\ [**+w**\ [*str*]\|\ **W**]
+[ |-D| ]
 [ |-I|\ *includefile* ]
 [ |-M|\ [*job*] ]
 [ |-Q|\ [**s**] ]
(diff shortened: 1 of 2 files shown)


Esteban82:
It would be possible that -D should be used with -Q?

If I don't use -Q, I think that all files are deleted. It could be possible?

PaulWessel:
I have also implemented **-F** and testing it.  The **-Q** should strictly be for debugging.  With **-D** (this PR) and **-F** (next PR coming once this is approved) we have full control:

1. If you are happy with producing single named products like job_00012.txt and job_0012.grd etc, then there are no changes.  These products are moved up to the top diff when the jobs complete.
2. If your script is not relying on this naming and is doing something custom (like mine) then you use **-D** to avoid **batch** trying to consolidate files that are not there.  batch cannot know if there should be files or not so this takes an option.
3. If your script wishes to name the products using the input **-T**_timefile_ then you use **-F**_template_ to do so.  That adds more flexibility for batch jobs that do not just depend on a single parameter.  **-F** can be combined with **-D** if you are writing the files to some custom place yourself.

With these changes I can write batch jobs that produces parameter files like this:

```
# Parameter file for job 2109
BATCH_DIR=EFLX3_2109
BATCH_NAME=E_obl_flex_G1.04_I2660_10.0k
BATCH_JOB=2109
BATCH_ITEM=2109
BATCH_COL0=1.04
BATCH_COL1=2660
BATCH_COL2=10
```

You will note I have introduced a new parameter called BATCH_DIR which basically replaces the old use of BATCH_NAME and BATCH_NAME is now exclusively used for product naming.

Esteban82 (approved):

PR #6934, opened 2022-08-03
Allow batch -F to set a more specific naming template

PaulWessel:
Rather than have a single running number as the product prefix we may optionally build one from a C-style template and the data columns provided via the **-T** timefile.  This allows products to be named to reflect more than just a running number.  E.g. `-FE_obl_flex_G%4.2lf_I%4.0lf_%4.1lfk` will let me use `${BATCH_NAME}.grd` to make more meaningful product names.

This PR also fixes the bug mentioned in #6932 which assumed all product files must have extensions.

Documentation has been updated to reflect the enhancements.  Apart from the bug fix, no default behavior has changed.

### DIFF
--- a/doc/rst/source/batch.rst
+++ b/doc/rst/source/batch.rst
@@ -15,6 +15,7 @@ Synopsis
 **gmt batch** *mainscript*
 |-N|\ *prefix*
 |-T|\ *njobs*\|\ *min*/*max*/*inc*\ [**+n**]\|\ *timefile*\ [**+p**\ *width*]\ [**+s**\ *first*]\ [**+w**\ [*str*]\|\ **W**]
+[ |-D| ]
 [ |-I|\ *includefile* ]
 [ |-M|\ [*job*] ]
 [ |-Q|\ [**s**] ]
@@ -81,6 +82,12 @@ Required Arguments
 
 Optional Arguments
 ------------------
+
+.. _-D:
+
+**-D**
+    The main script does not produce products that are named using the prefix **BATCH_NAME**, so we do not attempt
+    to move such files to the top directory.  The main script will instead handle this in a different way.
 
 .. _-I:
 
--- a/doc/rst/source/batch.rst
+++ b/doc/rst/source/batch.rst
@@ -87,8 +87,9 @@ Optional Arguments
 .. _-D:
 
 **-D**
-    The main script does not produce products that are named using the prefix **BATCH_NAME**, so we do not attempt
-    to move such files to the top directory.  The main script will instead handle this in a different way.
+    The main script does not produce products named using the prefix **BATCH_NAME**, so we do not attempt
+    to move such files to the top directory, or the main script will handle the placement of any such
+    product files directly.
 
 .. _-F:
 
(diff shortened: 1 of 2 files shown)


PaulWessel:
Happy to report that my command 

`gmt batch main.sh -IE_constants2.sh -Sbpre.sh -W -Tthis_flx_job.txt -NEFLX3 -FE_obl_flex_G%04.2lf_I%04.0lf_%04.1lfk -D`

relatively quickly produced 2970 grids with the new naming convention.  The **batch** module is great for these things.

Esteban82:
If I use `-Fmy_data_%s`, it should take the trailing test from the timefile as used as part of the name, right?

PaulWessel:
Yes, not tested but if your input file has trailing text and there are no other % except %s then that should be what it uses.  Let me know if otherwise.

PaulWessel:
YOu might even be able to give the template as -F%s and have the trialing text have the names you wanted, for example

Esteban82:
Ok, I try this script to make a png map for every european country. 

If I use  `-FCountries_%s`, then I should get the name of each country with the prefix "Countries_"? Because I didn't

Also, I only could get a map for countries with name with only one word (like "Andorra"), It failed with country like "San Marino". 

I got this error

```
begin [ERROR]: Unrecognized graphics format Marino
gmt [ERROR]: Shared GMT module not found: coast

ERROR: No module named coast was found.  This could mean one of four cases:

gmt [ERROR]: Shared GMT module not found: text

ERROR: No module named text was found.  This could mean one of four cases:

gmt [ERROR]: Not available in classic mode
```



Full script:
```
Title=Europe

cat << EOF > pre.sh
gmt begin
    gmt coast -E=EU+l > $Title
gmt end
EOF

cat << EOF > main.sh
gmt begin \${BATCH_WORD1} png
    gmt coast -R\${BATCH_WORD0}+e0.5 -JM10c -Glightgray -Slightblue -B -B+t"\${BATCH_WORD1}" -E\${BATCH_WORD0}+gred+p0.5p
    echo \${BATCH_WORD0} | gmt text -F+f16p+jTL+cTL -Gwhite -W1p
gmt end
EOF

gmt batch main.sh -Sbpre.sh -T$Title+w"\t" -N$Title -Wd -V -Q -FCountries_%s
```

PaulWessel:
I dont see you using BATCH_NAME anywhere though?

PaulWessel:
I would think you need to do

gmt begin ${BATCH_NAME} png
...

PaulWessel:
The space issue is something I was thinking about for regular numerical stuff too.  My first effort had things like %4.1lf and once the argument was < 10 I got a space.  So using %04.1lf fixed that.  But I was thinking that when -F is used perhaps we should replace any resulting space with underscore.

PaulWessel:
Now the spaces in BATCH_NAME become underscrores.

Esteban82:
Nop. I try with  `gmt begin \${BATCH_NAME} png` and I got: 



```
begin [ERROR]: Unrecognized graphics format San_Marino
gmt [ERROR]: Shared GMT module not found: text

ERROR: No module named text was found.  This could mean one of four cases:

gmt [ERROR]: Not available in classic mode
```

PaulWessel:
Right.  So the problem is that the input -T file is of course all trailing text but that means you have a single word with a tab in it.  I did not consider that case.  Now, I replace both spaces and tabs with undercores.  Now the parameter file look like this, e.g.

```
# Parameter file for job 15
BATCH_DIR=Europe_15
BATCH_NAME=Countries_FI_Finland
BATCH_JOB=15
BATCH_ITEM=15
BATCH_TEXT="FI  Finland"
BATCH_WORD0=FI
BATCH_WORD1=Finland

```

Esteban82:
Great. 

Now I only got an error with the Vatican City becuse the name has ():

```
batch_params_51.sh: línea 3: error sintáctico cerca del elemento inesperado `('
batch_params_51.sh: línea 3: `BATCH_NAME=Country_VA_Holy_See_(Vatican_City_State)'
coast [ERROR]: Could not decode +e0.5, return NaN.
```

Esteban82:
```
# Parameter file for job 51
BATCH_DIR=Europe_51
BATCH_NAME=Country_VA_Holy_See_(Vatican_City_State)
BATCH_JOB=51
BATCH_ITEM=51
BATCH_TEXT="VA	Holy See (Vatican City State)"
BATCH_WORD0=VA
BATCH_WORD1="Holy See (Vatican City State)"
```

PaulWessel:
That I cannot fix - needs to be a user fix I think.

Esteban82:
Ok. So I think it can be approved.

Esteban82 (approved):

PR #6935, opened 2022-08-04
Fix misc. links

anbj:


anbj:
http->https where possible, and make clickable links where appropriate. Built and tested.

seisman (approved):

PR #6937, opened 2022-08-05
Clarify what happens in gravfft with NaNs

PaulWessel:
Because residual grids with NaNs in areas of no data are common in isostatic studies, we allow **gravfft** to handle such grids rather than having to use **grdmath** **DENAN**.  This PR:

1. Replaces NaNs in the 1-2 input grids with zeros
2. WIth **-Vl** we report the number of NaNs as well
3. The documentation now has a new section that discussion what happens with the NaN grids in input and via **-D** as a density grid.
4. Minor typographical fixes in **-N** and **gravfft** documentations.


PaulWessel (commented):
Tested adn works

PR #6939, opened 2022-08-06
Mapproject make all links

anbj:
Make clickable links to arg. desc. where applicable. Please review if options should be in bold; see e.g. ``-Qd``, the -Q is a link and the *d* is in normal font that could be changed to **bold**.

### DIFF
--- a/doc/rst/source/mapproject.rst
+++ b/doc/rst/source/mapproject.rst
@@ -65,7 +65,7 @@ Finally, **mapproject** can compute a variety of auxiliary output
 data from input coordinates that make up a track.  Items like
 azimuth, distances, distances to other lines, and travel-times
 along lines can all be computed by using one or more of the options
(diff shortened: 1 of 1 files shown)


anbj:
Any comments on this? If it's too much I'll just close the PR.

PaulWessel:
Sorry, missed this one.  The d in -Q etc are called directives and they should be bold.  Arguments, e.g,, -Tname are in italics.

anbj:
Ok, otherwise good? If so I'll do those changes and push. Afraid it was a bit over the top making links of 'everything'.

PaulWessel:
No, I think more links the merrier - it is just a long of work of course.

anbj:
Great! At least it's a start..

anbj:
Ok; updated. A bit confused by directives vs. arguments and different formatting. I've kept what was bold as bold etc. so should be OK.

PR #6946, opened 2022-08-12
Tighten the auto-precision in movie and batch

PaulWessel:
Since it is important to get lexically sorted names for movie frames and batch jobs we compute the format precision from the number of frames or jobs.  However, since we start numbering from 1 we need to subtract 1 from the algorithm, otherwise n_frames = 1000 will yield a precession of 4 even though the largest frame number will only be 999 and require 3.


Esteban82 (approved):
Looks fine.

Issue #6955, opened 2022-08-15
Malformed input cause gmt to crash


anbj:
I forgot the `n` in `\n` and met a crash:

```
$ gmt begin test png; echo -e "0 0\10 10" | gmt plot -Gred; gmt end show
ERROR: Caught signal number 11 (Segmentation fault) at
/home/anbj/gmt/build/lib/libgmt.so.6(+0x1647dc)[0x7f4afb4b57dc]
[0x0]
Stack backtrace:
/home/anbj/gmt/build/lib/libgmt.so.6(sig_handler_unix+0xc0)[0x7f4afb49e620]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x14140)[0x7f4afb322140]
/home/anbj/gmt/build/lib/libgmt.so.6(+0x1647dc)[0x7f4afb4b57dc]
/home/anbj/gmt/build/lib/libgmt.so.6(+0x164f36)[0x7f4afb4b5f36]
/home/anbj/gmt/build/lib/libgmt.so.6(gmt_init_module+0x367f)[0x7f4afb4d316f]
/home/anbj/gmt/build/lib/libgmt.so.6(GMT_psxy+0x14e)[0x7f4afb7096ce]
/home/anbj/gmt/build/lib/libgmt.so.6(GMT_Call_Module+0x38e)[0x7f4afb3abc2e]
gmt(main+0x546)[0x55687aa77736]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xea)[0x7f4afb16fd0a]
gmt(_start+0x2a)[0x55687aa785aa]
```

If the input are numbers only, or the symbol is placed differently, an error message is given:

```
$ gmt begin test png; echo -e "0 0110 10" | gmt plot -Gred; gmt end show
plot [ERROR]: Could not parse -nan into Geographical, Cartesian, or Temporal coordinates!
plot [ERROR]: Option -R parsing failure. Correct syntax:
[...]
```

```
$ gmt begin test png; echo -e "0 0 \10 10" | gmt plot -Gred; gmt end show
plot [ERROR]: Could not parse -nan into Geographical, Cartesian, or Temporal coordinates!
plot [ERROR]: Option -R parsing failure. Correct syntax:
```

anbj:
Fixed by https://github.com/GenericMappingTools/gmt/pull/6956

PR #6956, opened 2022-08-15
Add required check in gmtinit_get_region_from_data

PaulWessel:
See https://github.com/GenericMappingTools/gmt/issues/6955 for background. The reason for the crash is that the malformed input record sets n col = 1 and then later when we run gmt info on it to detect **-R** we failed to check properly that we got 4 data columns back (xmin, xmax, ymin, ymax).  In fact, we pass the hidden -0 option to also return the data column types and our check failed to detect this problem.  Also fixed a formatting but in the error message.
This PR checks that we actually get (4+2 = 6) columns back from gmt info before we try to access those.  The crash is gone and we get this error instead:

```
gmt begin test png; echo -e "0 0\10 10" | gmt plot -Gred; gmt end show
gmt [ERROR]: gmtinit_get_region_from_data: Not enough data columns (1) to determine region.
plot [ERROR]: Must specify -R option
```


anbj (approved):

anbj:
Looks good as far as I can tell. No more crash.

PR #6966, opened 2022-08-19
Long-options support for movie/grdinterpolate/psevents/seismodules.

### LINKED ISSUE #6965: long-options support for movie/grdinterpolate/psevents/seis

long-options support for movie/grdinterpolate/psevents/seis

rbdavis:

Long-options support for movie/grdinterpolate/psevents/seismodules.

Sorry this is another multi-module PR, it should be the last one. (It was already in progress and I would have made more of a mess if I'd tried to split it up.) I will try to keep future long-option PRs to a single module where possible.

Note that there are interdependencies between psevents.c and psmeca.c/pscoupe.c, as the former requires specific knowledge of the latter's option strings.

There is also one unexplained change in movie.c (lines 1784 against line 1830, see below) that seems to have crept in somehow due to some git mis-synchronization, I suppose. Please correct as appropriate.

Fixes #6965

**Reminders**

### DIFF
--- a/doc/rst/source/grdmath.rst
+++ b/doc/rst/source/grdmath.rst
@@ -721,12 +721,11 @@ Notes On Operators
    two grids, the real (cosine) and imaginary (sine) component of the
    complex spherical harmonic. Use the **POP** operator (and **EXCH**) to
    get rid of one of them, or save both by giving two consecutive = file.nc calls.
-
    The orthonormalized complex harmonics **YLM** are most commonly used in
    physics and seismology. The square of **YLM** integrates to 1 over a
    sphere. In geophysics, **YLMg** is normalized to produce unit power when
    averaging the cosine and sine terms (separately!) over a sphere (i.e.,
-   their squares each integrate to 4 pi). The Condon-Shortley phase :math:`(-1)^M`
+   their squares each integrate to :math:`4 \pi`). The Condon-Shortley phase :math:`(-1)^M`
    is not included in **YLM** or **YLMg**, but it can be added by using *-M*
    as argument.
 
@@ -792,17 +791,36 @@ Macros
 ------
 
 Users may save their favorite operator combinations as macros via the
-file *grdmath.macros* in their current or user directory. The file may contain
+file *grdmath.macros* in their current or user (~/.gmt) directory. The file may contain
 any number of macros (one per record); comment lines starting with # are
 skipped. The format for the macros is **name** = **arg1 arg2 ... arg2**
 : *comment* where **name** is how the macro will be used. When this
 operator appears on the command line we simply replace it with the
 listed argument list. No macro may call another macro. As an example,
 the following macro expects three arguments (radius x0 y0) and sets the
-nodes that are inside the given circle to 1 and those outside to 0:
+nodes that are inside the given Cartesian circle to 1 and those outside to 0:
 
 **INCIRCLE** = **CDIST EXCH DIV** 1 **LE** : usage: r x y INCIRCLE to return 1
 inside circle
+
+Marine geophysicist often need to evaluate predicted seafloor depth from an age grid.
+One such model is the classic *Parsons and Sclater* [1977] curve.  It may be written
+
+.. math::
+
+    z(t) = \left \{ \begin{array}{rr} 
+            2500 + 350 \sqrt{t}, & t \leq 70 \\
+            6400 - 3200 \exp{\left (\frac{-t}{62.8}\right)}, & t > 20
+        \end{array} \right.
+
+A good cross-over age for these curves is 26.2682 Myr. A macro for this system is a bit awkward due to the split but can be written
+
+**PS77** = **STO@T POP** 6400 **RCL@T** 62.8 **DIV NEG EXP** 3200 **MUL SUB RCL@T** 26.2682 **GT MUL** 2500 350 **RCL@T SQRT MUL ADD RCL@T** 26.2682 **LE MUL ADD** : usage: age PS77 returns depth
+
+i.e., we evaluate both expressions and multiply them by 1 or 0 depending where they apply, and then add them.  With this macro
+installed you can compute predicted depths in the Pacific northwest via::
+
+    gmt grdmath -R200/240/40/60 @earth_age_01m_g PS77 = depth_ps77.grd
 
 **Note**: Because geographic or time constants may be present in a macro, it
 is required that the optional comment flag (:) must be followed by a space.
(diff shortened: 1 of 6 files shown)


PaulWessel (approved):
local static variables are not thread safe.  Please move ZSlongopt into the struct CTRL_PSEVENTS Ctrl so it is allocated with the rest, and you will have to pass it of the pointer to the struct to any function that needs access.

PaulWessel:
Oh I forgot: The change in movie was done ~1 week ago so remember you should always merge master into your branch to pick up any changes since you started working on the branch, otherwise we end up reverting things unintentionally.  That line 1784 should have the -1 in it - that was a bug fix.  Please add taht back in to your branch before merging.

joa-quim:
I note also that there barely any longoption name that is common to the those in GMT.jl. This is not *sustainable*

rbdavis:
> local static variables are not thread safe. Please move ZSlongopt ...

Will do, was not aware of this issue.

rbdavis:
> Oh I forgot: The change in movie was done ~1 week ago so remember you should always merge master into your branch to pick up any changes since you started working on the branch, otherwise we end up reverting things unintentionally. That line 1784 should have the -1 in it - that was a bug fix. Please add taht back in to your branch before merging.

Will restore as directed, sorry about that. I did do a pull on master right before beginning work on movie (about a week ago as you said). If you can explain to me how to do a merge on my system (which branch, which git command, etc.) I'll be happy to do that.

rbdavis:
> I note also that there are barely any longoption names common to the those in GMT.jl. This is not _sustainable_

This is the first I've heard of GMT.jl. I'm happy to make these new strings conform to others where possible, but I need to know what those pre-existing strings are. Also, this file does not appear to be part of what I've downloaded from the master branch onto my system. Where is GMT.jl? And is there anything else I should be looking at?

PaulWessel:
I thought I had mentioned this, but anyway.  GMT/Julia is at https://github.com/GenericMappingTools/GMT.jl and PyGMT is at https://github.com/GenericMappingTools/pygmt.

joa-quim:
Some more background. GMT.jl exists since ~2014-2015 and started to expand to use long options around 2016. PyGMT came later and also started to use long options but it doesn't yet wrap all GMT modules nor all the options in a given module. There is no full compliance between GMT.jl and PyGMT but in large most of the options names are the same (I've not done a full comparison though). Another difference between the two is that PyGMT is more strict but admitting only one name whilst GMT.jl accepts some aliases for the same option.

Knowing what GMT.jl uses as longoptions is easy. Just type ``gmthelp(module_name)`` or ``? module_name`` in the Julia REPL. The info between these two is not exactly the same because the ``gmthelp(...)`` intends to show all supported names (even the non-documented ones) but fails some times.

rbdavis:
> Some more background. GMT.jl exists since ~2014-2015 and started to expand to use long options around 2016. PyGMT came later and also started to use long options but it doesn't yet wrap all GMT modules nor all the options in a given module. There is no full compliance between GMT.jl and PyGMT but in large most of the options names are the same (I've not done a full comparison though). Another difference between the two is that PyGMT is more strict but admitting only one name whilst GMT.jl accepts some aliases for the same option.
> 
> Knowing what GMT.jl uses as longoptions is easy. Just type `gmthelp(module_name)` or `? module_name` in the Julia REPL. The info between these two is not exactly the same because the `gmthelp(...)` intends to show all supported names (even the non-documented ones) but fails some times.

Thanks Joaquim. Stupidity factor here, though, as I have no idea what is Julia REPL, nor how to obtain or run it. Any further hints?

joa-quim:
``REPL`` (Read Evaluate Print Loop) is a new slang for ``shell``. As for installing GMT.jl and its docs, see the link Paul posted above (same for PyGMT).

joa-quim:
And for installing Julia in other than Windows I recommend https://github.com/JuliaLang/juliaup

rbdavis:
> And for installing Julia in other than Windows I recommend https://github.com/JuliaLang/juliaup

Thanks. Any reason not to just use MacPorts to install Julia? (Monterey on M1)

joa-quim:
Don't know. If it works ... It's not uncommon to see comments on how Homebrew has screwed.
There is also https://julialang.org/downloads/ but ``juliaup`` takes care of adding to the path and knows how to update.

rbdavis:
> Don't know. If it works ... It's not uncommon to see comments on how Homebrew has screwed. There is also https://julialang.org/downloads/ but `juliaup` takes care of adding to the path and knows how to update.

OK, thanks. I'll try MacPorts first as I get many other external packages from there, easier to just add this to my list of must-have packages. If it makes a mess I'll throw it out and go the other route. I've never used Homebrew -- started with MP before it was around and never heard of a good enough reason to switch.

joa-quim:
Another thing. I'm blocked to release new versions because Conda screw by using the latest GCC and libstdc++ is incompatible with the one that Julia uses. Because of it my many changes are still waiting for that shit to be solved and leave only in the master version. Those changes includes some aliases renaming/adding and that's why I'm mentioning it.

All of this to say, when installing GMT.jl, please do (in the Julia shell/REPL) ``] add GMT#master`` and that will install the master version. After this, updating (later, when needed) it's done by only typing ``] up``

PaulWessel (approved):

rbdavis:
> Another thing. I'm blocked to release new versions because Conda screw by using the latest GCC and libstdc++ is incompatible with the one that Julia uses. Because of it my many changes are still waiting for that shit to be solved and live only in the master version. Those changes includes some aliases renaming/adding and that's why I'm mentioning it.
> 
> All of this to say, when installing GMT.jl, please do (in the Julia shell/REPL) `] add GMT#master` and that will install the master version. After this, updating (later, when needed) it's done by only typing `] up`

OK, some progress here, but not all the way there yet:

(1) MacPorts julia is apparently no better than Homebrew. The package installed with no errors but when julia is run it immediately exits with the message `Killed: 9`. I uninstalled, went to https://github.com/JuliaLang/juliaup as recommended by Joaquim and used the curl installation instruction for macOS. That worked fine, and the julia interpreter now starts properly.

(2)` ] GMT#master` installation of GMT under Julia also worked fine.

(3) gmthelp(modulename) does nothing useful for me:

> julia> gmthelp(blockmean)
> ERROR: UndefVarError: gmthelp not defined
> Stacktrace:
>  [1] top-level scope
>    @ REPL[4]:1

I expect there's a trivial environmental fix for this but it's not going to be figured out by me.

PaulWessel:
You have to also say
`using GMT`

rbdavis:
`using GMT`

That did it, thanks!

joa-quim:
Sorry, forgot to tell that. All packages in Julia need to be loaded first with “using”. 
Remember that you can also do

? modulename

rbdavis:
Thanks, Joaquim, no worries. All working now.

PR #6967, opened 2022-08-20
Clarify that the custom symbol can be read from input

PaulWessel:
The documentation for the custom symbol **k**, unlike the others, did not indicate that the name and/or size were optional on the command line and can be given via the input table instead  This PR clarifies those issues.

### DIFF
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -306,14 +306,18 @@
         examples to use as a template.  The language allows you to design symbols that
         take many parameters and can make decisions based on these parameters.
 
-    **-Sk**\ *name*\ /\ *size*
-        **k**\ ustom symbol. We will look for a symbol definition file called *name*\ .def in
+    **-S**\ [**k**\ *name*\ [/\ *size*]]
+        A **k**\ ustom symbol. We will look for a symbol definition file called *name*\ .def in
         (1) the current directory, (2) in ~/.gmt or (3) in **$GMT_SHAREDIR**/custom.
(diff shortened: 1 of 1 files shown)


Esteban82 (approved):

doc/rst/source/explain_symbols.rst_
```diff
@@ -306,14 +306,18 @@
         examples to use as a template.  The language allows you to design symbols that
         take many parameters and can make decisions based on these parameters.
 
-    **-Sk**\ *name*\ /\ *size*
-        **k**\ ustom symbol. We will look for a symbol definition file called *name*\ .def in
+    **-S**\ [**k**\ *name*\ [/\ *size*]]
+        A **k**\ ustom symbol. We will look for a symbol definition file called *name*\ .def in
         (1) the current directory, (2) in ~/.gmt or (3) in **$GMT_SHAREDIR**/custom.
-        The symbol defined in the definition file is of normalized unit size by default;
+        The symbol declared in the definition file is of normalized unit size by default;
```
joa-quim, line 312: _defined_ is more correct than _declared_

PaulWessel: Yes, but I was trying to avoid defined in definition file.  It is very repetitive...

PaulWessel: COuld say "The symbol in the definition file is ..."

joa-quim: OK, better because they are really being defined in those files.

PaulWessel: Done in master.

PR #6983, opened 2022-08-23
grdinterpolate longopts adjustments

### LINKED ISSUE #6982: grdinterpolate longopts adjustments

Adjust grdinterpolate longopts for better match to GMT.jl

rbdavis:

Adjust grdinterpolate longopts for better match with GMT.jl

Fixes #6982

**Reminders**


PaulWessel (approved):
Hi Roger, here we will do a few things differently:

1. **-F**: Use _interpolator_ instead of _interptype_.
2. The old modifiers 1,2 are no longer there since we added **+d**_level_ so only d as derivative=1|2.  I think you have not done sample1d yet but it will have the same option.

PR #6992, opened 2022-08-23
pswiggle longopts adjustments

### LINKED ISSUE #6991: pswiggle longopts adjustments
Adjust pswiggle longopts to better match GMT.jl

rbdavis:
Adjust pswiggle longopts to better match GMT.jl

Fixes #6991

**Reminders**

### DIFF
--- a/src/pswiggle.c
+++ b/src/pswiggle.c
@@ -46,19 +46,19 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = { /* Local options for this m
 	          short_modifiers,     long_modifiers */
 	{ 0, 'A', "azimuth",           "", "", "", "" },
 	{ 0, 'C', "center",            "", "", "", "" },
-	{ 0, 'D', "mapscale",
+	{ 0, 'D', "scalebar",
 	          "g,j,J,n,x",         "mapcoords,inside,outside,boxcoords,plotcoords",
 	          "w,j,a,o,l",         "length,janchor,side,anchoroffset,label" },
-	{ 0, 'F', "frmpen",
+	{ 0, 'F', "framepen",
 	          "",                  "",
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
-F is to set a frame (filled, outlined, shaded, etc) behind the scale.  Joaquim's framepen makes no sense since it has nothing to do specifically with pen.  I would think _scalepanel_ is better since pen and fill etc are all modifiers.  Or perhaps just --panel is even better.

rbdavis:
> -F is to set a frame (filled, outlined, shaded, etc) behind the scale. Joaquim's framepen makes no sense since it has nothing to do specifically with pen. I would think _scalepanel_ is better since pen and fill etc are all modifiers. Or perhaps just --panel is even better.

Actually, framepen was my string, not Joaquim's. I chose it because of the man page's reference to MAP_FRAME_PEN. If you want --panel instead just LMK and I will change it. This occurs in at least one other module as well as I recall.

PaulWessel:
Yes in basemap as well.

rbdavis:
basemap has not been longopted yet. I just double-checked, turns out I was 
referring to pscoast.

Do you want me to change --framepen to --panel in both pscoast and 
pswiggle?

PaulWessel:
Yes please, that is what -F does (sets a panel which has lots of attributes given by modifiers)

joa-quim:
GMT.jl uses ``box`` for the several modules that have -F

PaulWessel:
> GMT.jl uses `box` for the several modules that have -F

Yes, but panel is a better word for what it is.

joa-quim:
I knew there had to be a reason to have choose ``box`` instead of ``panel``.
``panel`` is already taken by **-c**

PaulWessel:
thanks, We will give this a think.

PR #7008, opened 2022-08-29
Increase length of var holding the custom symbol names.

joa-quim:
Custom symbol names can carry their path and a length of 64 is way too short.
This was damn tricky to find and only fantastic VSC searching capabilities helped.


PaulWessel (approved):
Good

PR #7011, opened 2022-08-31
Fix logic on how to detect a -L deprecated syntax.

joa-quim:
This fixes a bug where -L was not usable in modern mode.

### DIFF
--- a/src/mapproject.c
+++ b/src/mapproject.c
@@ -515,7 +515,7 @@ static int parse (struct GMT_CTRL *GMT, struct MAPPROJECT_CTRL *Ctrl, struct GMT
 
 	unsigned int n_slash, k, n_errors = 0, pos;
 	int n;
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
I think you may want to use **GMT_LEN_UNITS** instead of all those individual checks, e.g

isoldL = strchr (GMT_LEN_UNITS, *pch) != NULL;

joa-quim:
I knew something like that had to exist. But does it include 'c' and 'C'?

PaulWessel:
You can do stuff line strchr (GMT_LEN_UNITS "cC", *pch) since string constants concatenate

joa-quim:
OK, but that's much less readable.

PaulWessel:
Well, standard C so not that unreadable I think.

PR #7020, opened 2022-09-08
Add boilerplate long-option template structures to all modules

PaulWessel:
While at this point only a dozen or so modules have been given long-option equivalents, we expect to solicit help from contributors to complete the project.  To that end, this PR adds all the missing translation structures but sets them to NULL and blank.  Hence, all behavior and tests work exactly as before.  The idea is to simplify what contributors need to work on by isolating the translation tables in separate include files kept in a longopt sub-directory in src and the supplemental dirs.  All modules now include their own longopt include file.  CMakeLists.txt have been updated to add the new include files to the list of source files.  All seems OK.  I know there are almost 300 files that have changed but you can quickly see it is just one new line added to all modules and a few new directories with the (mostly) blank include files.  These were all autogenerated by the script sandbox/guru/build-long-options.sh, and then the dozen or so completed tables were moved from the module file to the include file.

I hope @joa-quim can build this branch on Windows to ensure there are no side-effects, and perhaps @seisman or the CI will do Linux.


PaulWessel:
Note there are a few modern mode only modules (subplot, etc) that are not yet included due to limitation in my script (needed source file names like psxy.c hence gmt --show-classic-modules but that excludes subplot etc. Maybe I can do these quickly now.

PaulWessel:
Also added the hidden **gmtread** and **gmtwrite**, and now pass _module_kw_ instead of NULL into _gmt_init_module_.  I have built with or without setting the **LONG** configuration settings in cmake so should be fine now.

joa-quim:
I build it but how is this going to work? There are currently 95 of them and that's too many for us to track what has been changed. One option is that after each has been approved we manually insert the .h back in the corresponding .c and delete the .h. But there is also the *approval* step, not that easy. I guess it will need an attentive look to both GMT.jl and PyGMT but specially to GMT.jl as I have names for everything, though I don't like them all.

PaulWessel:
Each person who wish to implement long-options for a module will make one pull request for that module, modifying just the include file.  We review, edit ir needed, and approve/merge.  We can discuss if we eventually merge back in to the source.
Yes, the review needs to look at what is done in Py and JL.  The instructions we post will have those links.

FYI, we will also allow for aliases in all of this so we can hopefully cover discrepancies in words between wrappers when there is no clear answer.

PaulWessel:
We made a checkbox list for 150 modules when we did the reformat of usage messages last summer (@maxrjones will recall that) so that we knew what had been done.  We can do the same here.

joa-quim:
Merging back into the C code right away after approval would have the benefit of visually showing the "how many to go".

PaulWessel:
Yes, if you are bothered by that directory.

joa-quim:
I find it quite different to always see that directory full instead of it gradually becoming empty.

PaulWessel:
Any reason to not approve this one now, team?

joa-quim (approved):

PR #7026, opened 2022-09-13
See if we can trick the CI Win compiler

PaulWessel:
Tired of getting dumb meaningliess "internal compiler error" on every PR.  Let's see if this rearrangement works.


PaulWessel:
Silly but this shuffle made the Win compiler not crash so we will do it this way!

seisman (approved):
Great job!

Issue #7039, opened 2022-09-22
Wrong scale in vector attributes +z[scale]


joa-quim:
This plots a vector with the correct length
```
echo 0 0 5 5 | gmt plot -Sv0.5+e+z -R0/5/0/5 -Jx1 -Bpa0.5f0.25g0.5 -BWSen -png lixo
```
but this not (seems divided by 2.54). Note the **+z1** instead of **+z**
```
echo 0 0 5 5 | gmt plot -Sv0.5+e+z1 -R0/5/0/5 -Jx1 -Bpa0.5f0.25g0.5 -BWSen -png lixo
```
![lixo](https://user-images.githubusercontent.com/537321/191632733-8d274670-40a7-4616-aa1e-055301722e9f.png)

joa-quim:
Also don't understand why I have to apply a scale factor when the fig scale is not 1:1
Again, this is correct (after compensating for the bug effect)
```
echo 0 0 5 5 | gmt plot -Sv0.5+e+z2i -R0/5/0/7 -Jx2 -Bpa0.5f0.25g0.5 -BWSen -png lixo
```
![lixo](https://user-images.githubusercontent.com/537321/191634289-6c3c1152-5a6d-4abe-ac69-339ca2357198.png)

but when the x,y scales are not the same, the result is wrong again and here I don't see anything that can be done since **+z** takes only one scale
```
echo 0 0 5 5 | gmt plot -Sv0.5+e+z2i -R0/5/0/7 -JX10 -Bpa0.5f0.25g0.5 -BWSen -png lixo
```
![lixo](https://user-images.githubusercontent.com/537321/191634434-5e7041c5-0068-4276-be2f-8dfa143c2c2f.png)

joa-quim:
In fact, the workaround for now is to use **+s** instead of **+z** (no need to change the input because it starts at 0,0)
```
echo 0 0 5 5 | gmt plot -Sv0.5+e+s -R0/5/0/7 -JX10 -Bpa0.5f0.25g0.5 -BWSen -png lixo
```
![lixo](https://user-images.githubusercontent.com/537321/191635166-3e151fc0-f9c1-44b3-b146-4257172e071a.png)

PaulWessel:
Your first case gives a correct 5,5 vector for me.  If I use +z1 then I get a short vector, and I have to use +z1i to get the correct answer.  So clearly there is some unit mischief here.

joa-quim:
[...]

PR #7047, opened 2022-09-30
WIP: Ensure vector +z is nondimensional or uses unit q only

PaulWessel:
See #7039 for discussion. The coding of modifier **+z** was expecting plot units but that makes no sense.  This PR checks that either no unit or q is appended and errors otherwise.

Hi @joa-quim, if you have time, please note that _vector2d_mods.sh_ and _vector3d_mods.sh_ now fail (left column row 4) but I think those are not failures but confusion about what units are used etc.  We are doing 

` Input: 2.20 1.27 Vector: -Sv20p+e+z1q`

So with **q** that means the input is user units to be converted via **-R -J** to plot units, then scaled by 1.  So given the previous plot row has

`Input: 2.20 1.27 Vector: -Sv20p+e+z0.394q+c --PROJ_LENGTH_UNIT=inch`

it should be longer by 2.54.  The **--PROJ_LENGTH_UNIT** should not have any implication here since input is _not_ in plot units but in user units.  Seems to me the original PS is wrong due to the bug we are trying to fix.  Let me know what you think.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -16485,10 +16485,24 @@ int gmt_parse_vector (struct GMT_CTRL *GMT, char symbol, char *text, struct GMT_
 					}
 				}
 				break;
-			case 'z':	/* Input (angle,length) are vector components (dx,dy) instead */
+			case 'z':	/* Input (angle,length) are vector components (dx,dy) instead. Either the components are
+							   already in plot units (and scale is nondimensional) or we append unit q to scale from
+							   user units to plot units. */
 				S->v.status |= PSL_VEC_COMPONENTS;
 				S->v.status |= PSL_VEC_MAGNIFY;
-				if (p[1]) error += gmtinit_get_length (GMT, symbol, &p[1], false, is_grdvector, &(S->v.comp_scale), &(S->v.v_unit_d));
+				if (p[1]) {	/* Got an argument [Default is 1, i.e., do nothing] */
+					size_t last = strlen (p) - 1;	/* Position of last character in p */
+					if (p[last] == 'q') {	/* User data should be scaled by -J -R */
+						S->v.v_unit_d = true;
+						p[last] = '\0';	/* Hide the q for now */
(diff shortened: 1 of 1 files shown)


joa-quim:
I have implemented patches Julia plot to workaround this bug and I think what I get is the right arrow. So, yes those tests have the wrong plots.
```
arrows([0 0 2.2 1.27], show=1, uv=true, Vd=1)
        psxy  -R-0.1/2.3/-0.04/1.32 -JX14c/9.5c -Baf -BWSen -Sv8p+e+h0.5+z5.833333333333333iq -P -K > C:\TEMP\GMTjl_tmp.ps
```

GMTjl_tmp

PaulWessel:
OK, then I need a review t approve and I can update the PS later.

joa-quim:
But the 2.54 factor is still there, as well as the problem when the axes are not isometric. See, I need to multiply by 2.54.

```
 echo 0 0 2.2 1.27 | gmt plot -R-0.1/2.3/-0.04/1.32 -JX14c/9.5c -Baf -BWSen -N -Sv8p+e+h0.5+z2.54q -png lixo
```
lixo


Regarding the non-isometry, I found that multiplying the *yy** by a scale factor given by 

*(map_height / map_width)  /  ((y_max - y_min) / (x_max - x_min))*

 (9.5/14) / ((1.32+0.04)/(2.3+0.1)) * 1.27 = 1.5207983193277308

and a zscale given by *map_width / (x_max - x_min)*

14 / (2.3+0.1) / 2.54 = 2.2966
 solves it. 

```
echo 0 0 2.2 1.52 | gmt plot -R-0.1/2.3/-0.04/1.32 -JX14c/9.5c -Baf -BWSen -N -Sv8p+e+h0.5+z2.2966q -png lixo
```
lixo

PaulWessel:
Added WIP back on.  Maybe have time to look at this tomorrow, but times are busy.

PR #7064, opened 2022-10-25
Fix expression that computes the uncertainty in the median.

joa-quim:
Some context in this forum [post](https://forum.generic-mapping-tools.org/t/what-is-the-notch-in-psxy-e-n/3381/6).


PaulWessel (approved):
looks good

PR #7065, opened 2022-10-26
Use correct named integer in notched whisker check

PaulWessel:
The **psxy** code was not properly ported and used integers like 2 instead of **EBAR_NOTCHED_WHISKER**.  Now seems to work OK.

map


joa-quim (approved):

PaulWessel:
Maybe I'll make a test for this later.

joa-quim:
Just add an extra number in ``boxwhisk.sh``

PR #7068, opened 2022-10-30
Add translation of compass directions in gmt_cn1.locale

ZMAlt:
Add translation of compass directions in `gmt_cn1.locale` to get Chinese directions when `gmt set GMT_LANGUAGE  cn1`

### DIFF
--- a/share/localization/gmt_cn1.locale
+++ b/share/localization/gmt_cn1.locale
@@ -25,9 +25,9 @@ W	4	星期三		周三	W
 W	5	星期四		周四	T
 W	6	星期五		周五	F
 W	7	星期六		周六	S
-# The following lines give the translation of the 4 major compass directions (WESN).
-# Name in full (optionally capitalized), abbreviation of longitude and latitude on map edge and wind direction (may be more than one letter).
-C	1	West		W	W
-C	2	East		E	E
-C	3	South		S	S
-C	4	North		N	N
+# 下面给出玫瑰图四个方向的中文名称
+# 由于玫瑰图仅使用缩写，因此只修改最后一行即可
+C	1	West		W	西
+C	2	East		E	东
+C	3	South		S	南
+C	4	North		N	北


share/localization/gmt_cn1.locale
```diff
@@ -25,9 +25,9 @@ W	4	
...
-C	1	West		W	W
-C	2	East		E	E
-C	3	South		S	S
-C	4	North		N	N
+# ������õ��ͼ�ĸ�����������
+# ����õ��ͼ��ʹ����д�����ֻ�޸����һ�м���
+C	1	West		W	��
+C	2	East		E	��
+C	3	South		S	��
+C	4	North		N	��
```
seisman, line 33: @ZMAlt  I think you need to update both the 3rd and the 5th columns to Chinese.

ZMAlt: Done

seisman:
```
# The following lines give the translation of the 4 major compass directions (WESN).
# Name in full (optionally capitalized), abbreviation of longitude and latitude on map edge and wind direction (may be more than one letter).
C	1	West		W	W
C	2	East		E	E
C	3	South		S	S
C	4	North		N	N
```
@PaulWessel Could you please explain what "abbreviation of longitude and latitude on map edge" means? Are they annotation suffix for geographic maps?

PaulWessel:
I think it just means the single letter one place after a positive definite longitude or latitude to indicate it is negative of positive.  In English we do 130W and 20N etc.  If there is an equivalent in other language they can change those letters, but if not then leave as is.  I think a single letter is expected if you change it.  So yes, annotation suffix.

seisman:
@PaulWessel Thanks for your explanation.

@ZMAlt Sorry for the delay. Although the changes looks messy in GitHub PR (https://github.com/GenericMappingTools/gmt/pull/7068/files), they look great here (https://github.com/ZMAlt/gmt/blob/Chinese_compass/share/localization/gmt_cn1.locale).

The only minor issue here is that, tabs are used in this file, but your changes use whitespaces (see the screenshot below). Please convert your whitespaces into tabs instead.

image

seisman (approved):

Issue #7086, opened 2022-11-09
Adding mb-system libs to GMT_CUSTOM_LIBS causes crash


anbj:
Not sure if this is a mb-system or gmt issue, but as @joa-quim once wrote, "every crash is a bug".

After installing git head of mb-system and adding the path to the libs (e.g. ``mbsystem.so``), with ``gmt set GMT_CUSTOM_LIBS /home/anbj/MB-System/build/lib/``, gmt crashes after every command, e.g.:

```
$ gmt

	GMT - The Generic Mapping Tools, Version 6.5.0_67dd5bd_2022.11.09 [64-bit] [MP] [12 cores]
	(c) 1991-2022 The GMT Team (https://www.generic-mapping-tools.org/team.html).

... snip...

ERROR: Caught signal number 7 (Bus error) at
/lib64/ld-linux-x86-64.so.2(+0x15961)[0x7f8522235961]
[0x0]
Stack backtrace:
/home/anbj/gmt/build/lib/libgmt.so.6(sig_handler_unix+0xc0)[0x7f8521d9c7b0]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x13140)[0x7f8521c1d140]
/lib64/ld-linux-x86-64.so.2(+0x15961)[0x7f8522235961]
/lib/x86_64-linux-gnu/libc.so.6(_dl_catch_exception+0x80)[0x7f8521b6bb50]
/lib/x86_64-linux-gnu/libc.so.6(_dl_catch_error+0x2f)[0x7f8521b6bc0f]
/lib/x86_64-linux-gnu/libdl.so.2(+0x1a65)[0x7f851f577a65]
/lib/x86_64-linux-gnu/libdl.so.2(dlclose+0x24)[0x7f851f577374]
/home/anbj/gmt/build/lib/libgmt.so.6(GMT_Destroy_Session+0xda)[0x7f8521cac16a]
gmt(main+0x487)[0x5576ce543677]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xea)[0x7f8521a58d0a]
gmt(_start+0x2a)[0x5576ce5445aa]
```

```
$ gmt pscoast
gmt pscoast [core] 6.5.0_67dd5bd_2022.11.09 [64-bit] [MP] - Plot continents, countries, shorelines, rivers, and borders

.. snip ..

ERROR: Caught signal number 7 (Bus error) at
[...]

PR #7089, opened 2022-11-12
Check that a custom shared library given has extension

PaulWessel:
See #7086 for background.  Turns out the documentation is corrrect, and one must end a directory entry with trailing slash.  Alternatively, give the full name (or comma-separated list of full names) to shared libraries.

This PR strengthens the parsing by checking that a non-directory argument in fact ends in the valid extension before trying to open it as a shared library and instead yield an informative error.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -1155,12 +1155,20 @@ GMT_LOCAL int gmtapi_init_sharedlibs (struct GMTAPI_CTRL *API) {
 			}
 		}
 		else {	/* Just a list with one or more comma-separated library paths */
-			unsigned int pos = 0;
+			unsigned int pos = 0, ok;
 			while (gmt_strtok (GMT->session.CUSTOM_LIBS, ",", &pos, text)) {
+				for (e = ok = 0; e < n_extensions; e++) {
+					if (strstr (text, extension[e])) ok++;
+				}
+				if (ok != 1) {	/* Maybe did not append proper shared library extension */
+					GMT_Report (API, GMT_MSG_ERROR, "Shared Library %s lacks proper extension\n", text);
+					GMT_Report (API, GMT_MSG_ERROR, "Check that your GMT_CUSTOM_LIBS (in %s, perhaps) is correct; if a directory it must end in /\n", GMT_SETTINGS_FILE);
+					continue;
+				}
 				libname = strdup (basename (text));		/* Last component from the pathname */
 				if (access (text, R_OK)) {
 					GMT_Report (API, GMT_MSG_ERROR, "Shared Library %s cannot be found or read!\n", text);
-					GMT_Report (API, GMT_MSG_ERROR, "Check that your GMT_CUSTOM_LIBS (in %s, perhaps) is correct\n", GMT_SETTINGS_FILE);
+					GMT_Report (API, GMT_MSG_ERROR, "Check that your GMT_CUSTOM_LIBS (in %s, perhaps) is correct; if a directory it must end in /\n", GMT_SETTINGS_FILE);
 				}
 				else if ((API->lib[n_custom_libs].name = gmtapi_lib_tag (libname))) {
 					API->lib[n_custom_libs].path = strdup (text);


anbj (commented):
When defining a directory, I'm still getting;

```
$ gmt set GMT_CUSTOM_LIBS /home/anbj/MB-System/build/lib/
ERROR: Caught signal number 7 (Bus error) at
/lib64/ld-linux-x86-64.so.2(+0x15961)[0x7f8699959961]
[0x0]
Stack backtrace:
/home/anbj/gmt/build/lib/libgmt.so.6(sig_handler_unix+0xc0)[0x7f86994c07f0]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x13140)[0x7f8699344140]
/lib64/ld-linux-x86-64.so.2(+0x15961)[0x7f8699959961]
/lib/x86_64-linux-gnu/libc.so.6(_dl_catch_exception+0x80)[0x7f8699292b50]
/lib/x86_64-linux-gnu/libc.so.6(_dl_catch_error+0x2f)[0x7f8699292c0f]
/lib/x86_64-linux-gnu/libdl.so.2(+0x1a65)[0x7f8696c9ea65]
/lib/x86_64-linux-gnu/libdl.so.2(dlclose+0x24)[0x7f8696c9e374]
/home/anbj/gmt/build/lib/libgmt.so.6(GMT_Destroy_Session+0xda)[0x7f86993d01aa]
gmt(main+0x487)[0x563224235677]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xea)[0x7f869917fd0a]
gmt(_start+0x2a)[0x5632242365aa]
```

PaulWessel:
OK, probably because my PR only addresses the lack of checking on paths only.  There was no crash on macOS so Cannot test what is going on.  We need a ddd debug session on Linux to see what is happening I think. Also need a Windows check.

joa-quim (approved):

joa-quim:
I think that on Windows the only documented way should as is now. **All** shared/supplement dlls must go into the ``bin/gmt_plugins`` dir. We don't want that some *clever* users put things in dirs with spaces or, worse, use accent chars and then come to us saying that *it's not working*.

anbj:
(I have no idea what I'm doing).

Installed `ddd` and compiled with *DCMAKE_BUILD_TYPE=Debug*. This is what I've got:
Screenshot from 2022-11-12 21-32-40


Screenshot from 2022-11-12 21-32-14

PaulWessel:
IT has been some years but I think you start it by giving it the argument of the program, i.e.

`ddd gmt &`

and gmt needs to be built in DEBUG mode.  Sorry, at conference at Sundvolden so no access to any Linux.

PaulWessel:
Maybe this [link](http://www.science.smith.edu/~jfrankli/231f13/231/doc/quickstart.html) can get you going?

anbj:
From Hawaii to Sundvolden, love it!

anbj:
Thanks, I'll see what I can do. Have to start learning C soon...

anbj:
PS! Is what I posted from `ddd` of any help? https://github.com/GenericMappingTools/gmt/pull/7089#issuecomment-1312566702

PaulWessel:
Maybe related to bus factor?

PaulWessel:
@anbj the message only says it crashes at line 193 which is the start of the signal handler.  So only hope is to set a stop point and step in until it crashes and hopefully there is something to be learned at that spot.

PaulWessel:
But apart from that, I think this specific PR can be merged, yes? It only adds more checking and better messaging b ut does not address the (for me) non-reproducible crash.

anbj:
Don’t underestimate the power of public transportation.

anbj:
Yes, just merge it. I’ll try digging for crash if time and ability permits. Thanks Paul.

joa-quim:
In what was not meant as a curse but as premonition, I told Andreas "if you had to debug on Linux ..."

anbj:
I remember @joa-quim. We’ll see..

PR #7093, opened 2022-11-15
Enclose filenames given to gdal_translate with quotes

PaulWessel:
Addresses this issue posted on the [forum](https://forum.generic-mapping-tools.org/t/internal-call-gdal-translate-from-grdcut-source-grid-filename-with-spaces-not-working/3439/2).  This PR adds the quotes in the calling of this command line via the system function from **grdcut**.
Note: I have not actually tested this on a file/directory with spaces but the fix is identical to what is used in **psconvert**.

### DIFF
--- a/src/grdcut.c
+++ b/src/grdcut.c
@@ -37,6 +37,12 @@
 #define THIS_MODULE_NEEDS	""
 #define THIS_MODULE_OPTIONS "-JRVf"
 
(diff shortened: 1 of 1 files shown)


joa-quim:
No, ofc 😄 
In my opinion this type of issues barely worth the time they take to fix the **users bug**.

PR #7096, opened 2022-11-18
Update psconvert to handle gs version 10

PaulWessel:
We had specific tests that did not anticipate a **gs** version 10 and hence turned off transparency flags.  This PR addresses this and isolates the checks in a bool function that we can update as specific future gs versions have trouble.  So this time it was not gs that was the culprit but us.

### DIFF
--- a/src/psconvert.c
+++ b/src/psconvert.c
@@ -1566,7 +1566,7 @@ GMT_LOCAL int psconvert_make_dir_if_needed (struct GMTAPI_CTRL *API, char *dir) 
 
 GMT_LOCAL psconvert_gs_is_good (int major, int minor) {
 		/* Return true if the gs version works with transparency */
-	if (major > 9) return true;		/* 10 should work as of 10.0.0 unless there are future regressions */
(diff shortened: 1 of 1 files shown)


Esteban82:
Is there any test?

PaulWessel:
You will need access to gs 10 and then try to make a transparent PDF or PNG using -G/path/to/gs10 in psconvert.  So no test can be added here.

PaulWessel:
You can make a PS file for testing this way:

echo 0 0 | gmt psxy -R-2/2/-2/2 -JX10c -B0 -Sc5c -Gred@50 -P > t.ps

Then gs 9 (my default) gives correct transparency:

`gmt psconvert t.ps -Tf`

If I build gs 10 and us it:

`gmt psconvert t.ps -Tf -G/Users/pwessel/Downloads/ghostscript-10.0.0/bin/gs`

then master fails while this branch works.

joa-quim (approved):
Despite having no gs 10 installed to test it, given the nature of the bug, if it works in one OS there is no reason to not work in all the others.

PR #7099, opened 2022-11-21
Improve modern onliner graphic check

### LINKED ISSUE #7098: Improve the error message when invalid formats are used for one-liners

Note the typo `-pnf` below. It should be `-png`. Ideally GMT should report an error like:
"Unrecognized graphics format pnf".
```
$ gmt basemap -Rg -JH20c -Ba -pnf map
gmt [ERROR]: Shared GMT module not found: basemap

ERROR: No module named basemap was found.  This could mean one of four cases:
```

**System information**

* Operating system: Linux
* GMT version (`gmt --version`): 6.5.0_111fa6d_2022.08.11

PaulWessel:
Closes #7098, hopefully.  Basically, if option starts with one of the first letters in a valid graphics format, has an acceptable length (<= 4) but fails to match a valid format then we assume it is a typo and report that message.  E.g.,

```
gmt basemap -Rg -JH20c -Ba -pnf map
gmt [ERROR]: Unrecognized graphics format pnf for modern mode one-liner command

```

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -187,6 +187,7 @@
 #include "gmt_internals.h"
 #include "gmt_sharedlibs.h" 	/* Common shared libs structures */
 #include <stdarg.h>
+#include "gmt_gsformats.h"
 
 #ifdef HAVE_DIRENT_H_
 #	include <dirent.h>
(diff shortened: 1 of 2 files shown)


seisman (approved):

seisman:
It seems this PR causes breaking changes:
```
gmt basemap -Rg -JH20c -Ba -t15 -png map
gmt [ERROR]: Unrecognized graphics format t15 for modern mode one-liner command
```

PaulWessel:
Lovely.  Yes, I guess I should add the check that the characters following the initial valid start letter (e.g., -t) must be same case and all letters, e.g., -PNG for transparent png, -png, -tiff etc).  I do not think we have a GMT module option that simply is two to four letters with those starting characters (e.g., -ps is the shortest, -tiff is longest).  And next option should be a filename (e.g., -png map) and not an option.

PR #7103, opened 2022-11-21
Update decimals of PI, E, EULER, PHI.

Esteban82:
I add all the decimals that are included in GMT for PI, E, EULER, PHI.

I think it is important to clarify this in case a user wants to use more decimals.


PaulWessel:
Not sure I understand.  Those constants go on and on, like PI.  What is the significance of writing 3.1415926535 versus 3.1415926... ?

Esteban82:
> Not sure I understand. Those constants go on and on, like PI. What is the significance of writing 3.1415926535 versus 3.1415926... ?

Yes, those contansts go on and on, but we just use a finite numbers of decimals. Maybe it is not important, but I think it is better to show what is the number of PI that we are using.

Esteban82:
I put those decimals because I run:

```
gmt math -Q PI 1 MUL =
3.14159265359
```
But, since is wrong, then we could leave the original table and add a text saying how many decimals we are using.

PaulWessel:
That is output format related (FORMAT_FLOAT_OUT etc).
I think it is enough to state that all items are used in full double precision internally (for gmtmath) and to single precision internally (grdmatn).

PR #7114, opened 2022-11-25
Ensure modern mode does not create compatibiltiy level 4

### LINKED ISSUE #7112: Useless warnings about compatibility level
If we run a modern mode script with a gmt.conf file that has compatibility 4 [Default], we may get pointless and confusing messages:

gmt [WARNING]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.

This is because of these tests:

```
ival = (int)atof (value);
limit = (GMT->current.setting.run_mode == GMT_CLASSIC) ? 4 : 6;
if (ival < limit) {
	GMT_Report (GMT->parent, GMT_MSG_WARNING, "GMT_COMPATIBILITY: Expects values from %d to %d; reset to %d.\n", limit, GMT_MAJOR_VERSION, limit);
	GMT->current.setting.compatibility = 4;
}

```

So when gmt begin initializes and sets up the gmt.conf file it ends up with compatibility (_ival_) 4.  This makes no sense for a modern mode script since compatibility is really only for GMT 4 classic scripts.  I think the solution is to ensure that a modern mode script writing a gmt.conf file writes 6 so that we dont get these messages.

PaulWessel:
See #7112 for background.  This applies the same fix when writing GMT_COMPATIBILITY as we do when reading it under a modern mode: Setting it to 6.  Basically, there is no level 4 when we are doing modern mode, only when dealing with classic GMT. Closes #7112.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -13178,6 +13178,7 @@ char *gmtlib_getparameter (struct GMT_CTRL *GMT, const char *keyword) {
 			/* Deprecated as of 6.2: we only use GMT_DATA_UPDATE_INTERVAL to control this feature now, so just break here */
 			break;
 		case GMTCASE_GMT_COMPATIBILITY:
+			if (GMT->current.setting.run_mode == GMT_MODERN) GMT->current.setting.compatibility = MAX (GMT->current.setting.compatibility, 6);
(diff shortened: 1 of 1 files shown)


Esteban82 (approved):
Looks fine!

anbj (approved):
Looks good. No more compat babble.

PR #7117, opened 2022-11-27
Do not turn on polygon flag when -L args are given

### LINKED ISSUE #7116: Fill color in header screws -L+D polygons
Data
```
>  -W,#0072BD -G#0072BD
1	5	0.2	0.2
2	3.4	0.2	0.275
3	1.5	0.1	0.075
4	0.2	0	0.1
>  -W,#D95319 -G#0072BD
1	5.9	0.3	0.4
2	2.8	0.275	0.2
3	4.35	0.35	0.25
4	1.3	0.1	0.2
>  -W,#EDB120 -G#0072BD
1	6.5	0.275	0.4
2	3	0.2	0.175
3	5.55	0.45	0.325
4	2	0.2	0.3
```

```
gmt plot -JX14c/9.5c -Baf -BWSen -R1/4/1/8 -L+D+p0.5p -W0.5 G4par.dat -png lixo
```
lixo

But if one removes the ``-G#...`` from the headers we get the right polygons, except not filled ofc

lixo2

PaulWessel:
See #7116 for background.  The trouble was this line:

`if (change & 1) polygon = true;`

which detected a fill change in the segment header and turns the polygon flag to true even though in this case (**-L+D**) the polygon has not yet been built from the line, hence the automatic closing of the open polygon is premature.  As it was, we wrongly closed the "polygon" (which was just a line) and wreaked havoc further down.  The bug affected both psxy.c and psxyz.c.

Once approved and merged I will add a test based on this case.  Closes #7116.

### DIFF
--- a/src/psxy.c
+++ b/src/psxy.c
@@ -2582,7 +2582,7 @@ EXTERN_MSC int GMT_psxy (void *V_API, int mode, void *args) {
 						current_fill = Ctrl->G.fill;
 				}
 				else if (z_for_cpt == NULL) {
-					if (change & 1) polygon = true;
+					if (change & 1 && Ctrl->L.anchor == 0) polygon = true;
 					if (change & 2 && !Ctrl->L.polygon) {
 						polygon = false;
 						PSL_setcolor (PSL, current_fill.rgb, PSL_IS_STROKE);
--- a/src/psxyz.c
+++ b/src/psxyz.c
@@ -2133,7 +2133,7 @@ EXTERN_MSC int GMT_psxyz (void *V_API, int mode, void *args) {
 						current_fill = Ctrl->G.fill;
 				}
 				else if (z_for_cpt == NULL) {
-					if (change & 1) polygon = true;
+					if (change & 1 && Ctrl->L.anchor == 0) polygon = true;
 					if (change & 2 && !Ctrl->L.polygon) {
 						polygon = false;
 						PSL_setcolor (PSL, current_fill.rgb, PSL_IS_STROKE);


PaulWessel:
Note that GMT requires you to give **-W** and or **-G** on the command line to actually use any **-W -G** in the segment header.  You cannot turn on fill just by adding **-G** to a header - it is required that you also specify **-G**_fill_ on the command line so it can be overridden.

joa-quim (approved):

joa-quim:
> Once approved and merged I will add a test based on this case. Closes https://github.com/GenericMappingTools/gmt/issues/7116.

Note that there are two issues om #7116

PaulWessel:
I take it your Julia problem persists after this PR then?  And you are issuing -W and -G on the command line so that those can be overridden by things in the segment headers?

joa-quim:
Didn't try yet but it but it's not a polygon issue. I posted (always do) the generated command.

```
julia> parallelplot("iris.dat", groupvar="text", quantile=0.25, band=true, pen=1, show=1)
-W,#0072BD
        psxy  -R1/4/0/1 -J -G#0072BD@70.0 -L+D -W1 -K -O >> C:\TEMP\GMTjl_tmp.ps
        psxy  -R1/4/0/1 -J -G#D95319@70.0 -L+D -W1 -K -O >> C:\TEMP\GMTjl_tmp.ps
        psxy  -R1/4/0/1 -J -G#EDB120@70.0 -L+D -W1 -K -O >> C:\TEMP\GMTjl_tmp.ps
```

PaulWessel:
Not sure I understand the -W,#0072BD setting, i.e., no pen width specified?  Does it still fail (in Julia) if you give an actual width? It will all be reset to whatever is in the header anyway (or should).

joa-quim:
No pen in ``-W,#0072BD`` means use the default pen.
After rebuilding with polygon fix the color is still not obeyed even with an explicit pen width.
 
```
julia> plot(mat2ds([1 1 0.2 0.3; 2 2 0.3 0.2], color="red", fill="#0072BD", lt=1), L="+D", W=0.5, fill=true, Vd=1, show=1)
(D[1]).header = " -W1.0,#0072BD -G#0072BD"
        psxy  -JX14c/9.5c -Baf -BWSen -R0.96/2.04/0.96/2.04 -G#0072BD -L+D -W0.5 -P -K > C:\TEMP\GMTjl_tmp.ps
```

GMTjl_tmp

PaulWessel:
OK, so I see that the equivalent CLI works fine.  Hence, fails via Julia.  I take it if you pass -W0.5p,pink you get a pink line and not the red you want, first?

joa-quim:
Right, pink it is.

PaulWessel:
So I guess if you pass a different -G value than the one in fill you get that color instead as well? Seems like the Julia dataset being passed in does not have the right stuff in the segment header?  E.g., being empty?

joa-quim:
No, ofc, it's empty but I'm more and more confuse with this
```
julia> D = mat2ds([1 1 0.2 0.3; 2 2 0.3 0.2], color=["red"], fill=["#0072BD"], lt=1)
BoundingBox: [1.0, 2.0, 1.0, 2.0, 0.2, 0.3, 0.2, 0.3]
Header:  -W1.0,red -G#0072BD
2×4 GMTdataset{Float64, 2}
 Row │       X        Y       Z1       Z2
     │ Float64  Float64  Float64  Float64
─────┼────────────────────────────────────
   1 │     1.0      1.0      0.2      0.3
   2 │     2.0      2.0      0.3      0.2
julia> plot(D, lc=:blue, Vd=1, show=1)
        psxy  -JX14c/9.5c -Baf -BWSen -R0.96/2.04/0.96/2.04 -W,blue -P -K > C:\TEMP\GMTjl_tmp.ps
```
You can see the header and in this case the online command overwrites that in header. The line is blue.
GMTjl_tmp

But if I now use a vector of datasets instead of a single dataset, it works. Well it prints what the header says and I don't even have to specify any -W.

```
julia> plot([D], Vd=1, show=1)
        psxy  -JX14c/9.5c -Baf -BWSen -R0.96/2.04/0.96/2.04 -P -K > C:\TEMP\GMTjl_tmp.ps
```
GMTjl_tmp

joa-quim:
Hmm, solved half the mystery. Sometime ago I made  a change where a GMTdaset without Text would be passed as a Matrix, so the *scalar* case above was being passed as a matrix and loose the header info. Now that is fixed I have only the case that no -W is needed to plot the line.

PaulWessel:
I believe psxy long-standing default is to draw a line if no symbol nor -W is given.  So looks like there is nothing left here but closing?

joa-quim:
There is nothing opened about this. Except the talk. And yes, it's probably that.

Issue #7109, opened 2022-11-24
coast -Di -A falsely turning whole tiles to water iso land


KristofKoch:
**Description of the problem**
When using `coast -Di -A` with `-A` in the range of 2460 – 6920 inclusive, some tiles west of Hudson Bay in Canada (around N60° W102°) get turned into “water” instead of “land”. 

**Full script that generated the error**

```
gmt begin globe_Di_A2460 png
  # First occurrence at -A2460
  gmt coast -Rg -JG-102/60/10c -Di -A2460 -Gblack -Sgrey95

  # Last occurrence at -A6920
  # gmt coast -Rg -JG-102/60/10c -Di -A6920 -Gblack -Sgrey95
gmt end show
```

**Actual outcome**

First occurrence at `-Di -A2460`:
![2460](https://forum.generic-mapping-tools.org/uploads/default/original/2X/1/12c2caea4a6e61835a56a23523c19d226e282c91.png)

Last occurrence at `-Di -A6920`:
![6920](https://forum.generic-mapping-tools.org/uploads/default/original/2X/1/14bcc34c4e5416237879572c1eae9d1a437bd85a.png)

**Expected outcome**

Have those tiles show as “land” instead of “water”. 

**System information**

* Operating system: macOS 13.0.1
* GMT version (`gmt --version`): 6.2.0_a70eacd_2021.05.18
* GSHHG version 2.3.7

KristofKoch:
Fixed in #7121

Issue #1295, opened 2019-08-02
pscoast -G fill missing slice in Antarctica


btozer:
**Description of the problem**

Hey @PaulWessel  As mentioned earlier in the week - Here is a reproduction of the fill slice missing in Antarctica.

**Full script that generated the error**

gmt pscoast -Rg -JG176/-41/10c -Ggrey -Wblack -Bg45 -Dl -A8000  -Xc -Yc -P 
![globe](https://user-images.githubusercontent.com/8233693/62403434-f20dc080-b541-11e9-8fde-fc9fc03de459.png)

> globe.ps

**Full error message**

A slice is not filled by -G flag fill color in Antarctica.

**System information**

* Operating system: MacOs Mojave 10.14.5
* Version of GMT: 5.4.5

seisman:
Just tried the same command with gmt 6.1.0_22d70a8_2019.08.01. It's even worse.

Note there are some white lines in this plot.
![image](https://user-images.githubusercontent.com/3974108/62407361-1ed6cd80-b56c-11e9-9378-4eeb63b2fa48.png)

btozer:
The white lines are also in the 5.4.5 version, it's just they can't be seen at the scale of the plot.

PaulWessel:
The missing pieces is clearly a bug, but the thin lines might be result of aliasing in gs.

PaulWessel:
I have not figured out why it happens, but it is related to the -A8000 option.  Selecting a smaller value, e.g. -A7000 works.  So the workaround is to reduce the minimum area cutoff.

PaulWessel:
Keep having to rediscover which bin is failing... For the record (in case not solving it this weekend), one bin that is failing to appear is number 641 (selected as only bin via debug **-+**_bin_ option), but see **-A** discussion above.  Here is the failing example:

`gmt pscoast -Rg -JG176/-41/18c -Gred -Wblack -Bg30 -Dl -A8000  -+641 > t.ps
`

![t](https://user-images.githubusercontent.com/26473567/65378836-1e001500-dc5a-11e9-8c83-78cf84749143.png)

PaulWessel:
[...]

PR #7121, opened 2022-11-27
Attempt to fix coast holes in Canada for some -A settings

PaulWessel:
See #7109 for background.  We did have checks for this but it seemed only tailored to the Antarctica case and the code refers to #1295 which we tried to fix.  This PR restricts that Antarctica check to Antarctica only (< 60S) and applies the general fix elsewhere, which is to reduce the node level of the corner if a polygon covering it is skipped.  This seems to fix the problem and older tests still work.


joa-quim (approved):

KristofKoch (approved):
Works as expected for me now in 6.5.0_578b7c8_2022.11.27 Thank you Paul!

PaulWessel:
Nice test, @KristofKoch !  A bit too big to add to our suite though.  Looks like this is fixed for now.

Issue #2227, opened 2019-12-08
coast -Td, -Tm: Don't repeat the option, just say Draw…


tparker37:
Example task for AGU2019 workshop

**Description of the problem**

**Full script that generated the error**

```
# PASTE CODE HERE
```

**Full error message**

```
# PASTE ERROR MESSAGE HERE
```

**Actual outcome**

**Expected outcome**

**System information**

* Operating system:
* GMT version (`gmt --version`):

PR #7135, opened 2022-12-06
Correct column naming and width unit in ternary example

michaelgrund:

Based on [#2227](https://github.com/GenericMappingTools/pygmt/pull/2227) this PR corrects the column naming in the ternary example. 

**Reminders**

### DIFF
--- a/doc/rst/source/psternary.rst
+++ b/doc/rst/source/psternary.rst
@@ -53,13 +53,13 @@ Examples
 
 .. include:: explain_example.rst_
 
-To plot circles (diameter = 0.1 cm) on a 6-inch-wide ternary diagram at the positions listed
+To plot circles (diameter = 0.1 cm) on a 15-centimeter-wide ternary diagram at the positions listed
 in the file ternary.txt, with default annotations and gridline spacings, using the
 specified labeling, try::
 
     gmt makecpt -Cturbo -T0/80/10 > t.cpt
-    gmt psternary @ternary.txt -R0/100/0/100/0/100 -JX6i -P -Sc0.1c -Ct.cpt -LWater/Air/Limestone \
-        -Baafg+l"Water component"+u" %" -Bbafg+l"Air component"+u" %" -Bcagf+l"Limestone component"+u" %" \
+    gmt psternary @ternary.txt -R0/100/0/100/0/100 -JX15c -P -Sc0.1c -Ct.cpt -LLimestone/Water/Air \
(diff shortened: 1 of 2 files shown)


yvonnefroehlich:
Also the [example for `psteneray`](https://docs.generic-mapping-tools.org/dev/psternary.html#examples) has to be corrected, please.

yvonnefroehlich:
> Note that ternary.sh test produces the same output as the one shown in the Matlab central example.

I do not get the same figures. Please note the axis directions highlighted by the purple ellipses.
_Test psternary output_:
gmt_psternary_test_marked

_MATLAB file exchange output_:
teneray_matlab_fileexchange_marked

michaelgrund:
I agree with @yvonnefroehlich that after the renaming and changing `-JX-6i` to `-JX6i ` all figures (from FEX, ternary example and fig from the test) are identical, with the exception that the whole diagram is rotated. The difference between @joa-quim s last figure and the one from FEX is that the values (0 - 100% or 0-1) are in opposite directions.

So, to conclude what needs to be changed:

1) renaming (as already done in this PR for the example) for all appearances
2) change `-JX-6i` to `-JX6i `in `ternary.sh`

yvonnefroehlich:
> I agree with @yvonnefroehlich that after the renaming and changing `-JX-6i` to `-JX6i ` all figures (from FEX, ternary example and fig from the test) are identical, with the exception that the whole diagram is rotated. The difference between @joa-quim s last figure and the one from FEX is that the values (0 - 100% or 0-1) are in opposite directions.
> 
The second point is what I meant with "Please note the axis directions" and tried to highlight with the purple ellipses in https://github.com/GenericMappingTools/gmt/pull/7135#issuecomment-1339494943 🙂.

> So, to conclude what needs to be changed:
> 
> 1. renaming (as already done in this PR for the example) for all appearances
> 2. change `-JX-6i` to `-JX6i `in `ternary.sh`

Regarding the second point, I feel it makes sense to addionally change the unit from inches to centimeters as mention by  @joa-quim in https://github.com/GenericMappingTools/gmt/pull/7135#issuecomment-1339384514.

Hm. Probably adding a colorbar also in the examples in the docs (in the test there is already a colorbar) would make the figure better interpretable?

PaulWessel:
I agree changing the test script only adds more bellyache with PS update.  But go ahead and update the other things.

joa-quim (approved):

PR #7139, opened 2022-12-08
Improve green spline related to slopes

PaulWessel:
Ensure that both slopes and data are scaled by inverse range. Use new named constant for just removing the mean.  Update man page to make it clear that the sopes data in **-A** must be at different locations than the data values.

### DIFF
--- a/src/greenspline.c
+++ b/src/greenspline.c
@@ -164,6 +164,7 @@ enum Greenspline_modes {	/* Various integer mode flags */
 	LINEAR_2D			= 11,
 	N_METHODS			= 12,
 	N_PARAMS			= 11,
+	GREENSPLINE_MEAN		= 0,		/* Remove/Restore mean value only */
 	GREENSPLINE_TREND		= 1,		/* Remove/Restore linear trend */
 	GREENSPLINE_NORM		= 2,		/* Normalize residual data to 0-1 range */
 	SQ_N_NODES 			= 10001,	/* Default number of nodes in the precalculated -Sq spline */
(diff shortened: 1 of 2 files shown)


src/greenspline.c
```diff
@@ -1975,13 +1976,13 @@ EXTERN_MSC int GMT_greenspline (void *V_API, int mode, void *args) {
...
-		normalize = GREENSPLINE_NORM;	/* Only allow taking out data mean for mixed z/slope data */
-		GMT_Report (API, GMT_MSG_WARNING, "Can only remove/restore mean z in mixed {z, grad(z)} data sets\n");
+		normalize = GREENSPLINE_NORM;	/* Only allow taking out data mean and scale residuals for mixed z/slope data */
+		GMT_Report (API, GMT_MSG_WARNING, "Can only remove/restore mean z and scale residuals in mixed {z, grad(z)} data sets\n");
 	}
 
 	if (m == 0)
 		GMT_Report (API, GMT_MSG_INFORMATION, "Found %" PRIu64 " data points, yielding a %" PRIu64 " by %" PRIu64 " set of linear equations\n",
-		            n, nm, nm);
+			n, nm, nm);
```
joa-quim, line 1985: Don't do this. It doesn't align nicely with the first argument in above line. Need to use spaces for that (probably that's how it was)

joa-quim (approved):

src/greenspline.c
PaulWessel: OK done

Issue #7148, opened 2022-12-12
grdvector: order of input files and missing option to -W


anbj:
* Req. [arguments](https://docs.generic-mapping-tools.org/dev/grdvector.html#required-arguments) to [grdvector](https://docs.generic-mapping-tools.org/dev/grdvector.html) are *compx.nc* and  *compy.nc*
    * If [-A](https://docs.generic-mapping-tools.org/dev/grdvector.html#a) is used, the required arguments are *r* and *theta*. Should this be mentioned? (similar issue as in #7031).
    * the order of input is important(?); based on testing is seems like *compx.nc* have to come first, and then *compy.nc*. Using *-A*, the order is *r* follows by *theta*. Should this be mentioned?
* ``-W+c`` works together with ``-C``, but is not mentioned in [docs](https://docs.generic-mapping-tools.org/dev/grdvector.html#w). Should it? (I think so - I wanted this 'feature', so based on experience from other modules I tried it - and it worked).

anbj:
* Also, in the documentation - in rst - it's possible to use the mathematical symbol instead of writing it out; e.g. *theta* -> $\theta$ (also possible on github, see [Math support in Markdown](https://github.blog/2022-05-19-math-support-in-markdown/). Worth changing?

I can fix all points mentioned if I get 'go'.

PaulWessel:
Sure, feel free to submit a PR that addresses those issues:

1. Makes it clear what the expected order is for the two grids.
2. Perhaps check plot and consider updating the -W description in grdvector
3. Use math theta except when used in file name (theta.nc)

Thanx!

anbj:
Good!

See [forum](https://forum.generic-mapping-tools.org/t/plotting-vectors-and-calculating-field-lines-and-flux/2771/12?u=andreas) for illustratoin of ``-W+c``.

anbj:
See #7151.

PR #7151, opened 2022-12-12
grdvector.rst: minor improvements

anbj:
Work on #7148.

* describe order of input grids
* use :math: where appropriate
* add -W+c  - note that When ``-W+c`` is given -``Wpen`` specification is ignored. So p.t. ``-Wpen`` and ``-W+c`` are mutually exclusive. Not sure if the formatting is appropriate - the *pen* should maybe also be in ``[]``? A nice feature would be to be able to color vector head+stem and have an outline (set with *pen*, e.g. ``-Wblack+c``).

### DIFF
--- a/doc/rst/source/grdvector.rst
+++ b/doc/rst/source/grdvector.rst
@@ -12,7 +12,7 @@ Synopsis
 
 .. include:: common_SYN_OPTs.rst_
 
-**gmt grdvector** *compx.nc* *compy.nc* |-J|\ *parameters*
+**gmt grdvector** *grid1* *grid2* |-J|\ *parameters*
 [ |-A| ]
 [ |SYN_OPT-B| ]
 [ |-C|\ *cpt* ]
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Looks good but probably need +c in the synopsis too

anbj:
Done.

anbj:
I think thats a good suggestion. In this context, grid formats are/should be abstracted away I guess (read: not interesting).

So in synopsis and req. arguments, change ``.nc`` to ``.grd`` *or* remove extensions. What say you, @PaulWessel?

PaulWessel:
Yes I think in other places we use grid1 grid2? See grdfft grdblend etc

anbj:
Alright - please do a new review.

PaulWessel (approved):
Looks great

anbj:
Glad to hear.
Added one more reference to *-Z*.
If it looks good, I think we're good to go!

(Pardon the fuzzing and force pushing - I'm afraid to fudge up something by doing new commits).

PR #7167, opened 2022-12-18
The -T option had lost the -Tfile info

joa-quim:
But it's still not clear what is **-T**list


PaulWessel:
-Tlist is things like

`
-T13,15,16,22.5`

i'e,m a list of coordinates that you can also pass via a file one per line.

PaulWessel (approved):

PR #7171, opened 2022-12-24
Do not adjust externally allocated segments

PaulWessel:
See this pygmt [post](https://github.com/GenericMappingTools/pygmt/pull/2138#issuecomment-1364463845) for details.  The problem was that **ternary** converts the incoming `a,b,c[,z]` to `x,y[,z]` and in the process tries to remove that extra column. This ends up in _gmtio_adjust_segment_ which did not check the allocation mode of the vectors it was trying to free, thus trying to free memory not belonging to GMT.  So this is an API bug in a function rarely called.

Check allocation status of the affected vectors and only free if allocated by GMT, else set to NULL.

**Question**: If you examine `df` after the plot has finished you will see that the content has changed since **ternary** changes it.  If you feel that is not right and should be considered read-only memory then your **ternary** pyGMT wrapper needs to pass **GMT_IS_DUPLICATE** instead of **GMT_IS_REFERENCE**. You can make a decision on that but at least the SEGV bug in the API has been fixed, I hope.

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -3078,10 +3078,16 @@ GMT_LOCAL void gmtio_adjust_segment (struct GMT_CTRL *GMT, struct GMT_DATASEGMEN
 	/* Change the number of columns in this segment to n_columns (free or allocate as needed) */
 	uint64_t col;
 	struct GMT_DATASEGMENT_HIDDEN *SH = gmt_get_DS_hidden (S);
-	for (col = n_columns; col < S->n_columns; col++) gmt_M_free (GMT, S->data[col]);	/* Free up if n_columns < S->columns */
+	for (col = n_columns; col < S->n_columns; col++) {	/* Free up if n_columns < S->columns if allowed */
+		if (SH->alloc_mode[col] == GMT_ALLOC_INTERNALLY)
+			gmt_M_free (GMT, S->data[col]);
+		else
+			S->data[col] = NULL;	/* Let go of link to an external vector */
+	}
+	/* Reallocate number of columns (increase or decrease lengths) */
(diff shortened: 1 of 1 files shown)


seisman (approved):

seisman:
> **Question**: If you examine `df` after the plot has finished you will see that the content has changed since **ternary** changes it. If you feel that is not right and should be considered read-only memory then your **ternary** pyGMT wrapper needs to pass **GMT_IS_DUPLICATE** instead of **GMT_IS_REFERENCE**. You can make a decision on that but at least the SEGV bug in the API has been fixed, I hope.

Yes, I can confirm that this PR fixes the SEGV bug.

As what you said, the variable `df` changes from:
```
      a      b      c     d
0  0.16  0.331  0.509   9.0
1  0.22  0.625  0.158  39.0
2  0.09  0.180  0.732   2.0
3  0.66  0.258  0.078  12.0
```
to
```
          a         b     c     d
0  0.325500  0.286654   9.0   9.0
1  0.530907  0.539647  39.0  39.0
2  0.179641  0.155573   2.0   2.0
3  0.792169  0.224332  12.0  12.0
```
which is unexpected.

PaulWessel:
Yes, I do agree this is unexpected.  Given that nobody will plot 2 Tb of points in a ternary (we hope) it is probably OK if ternary inside does the duplication to avoid surprises on the other end.  But I think this is possibly also true for other situations.  Many grid operators modify the input grid rather than duplicate, for instance.

seisman:
> Given that nobody will plot 2 Tb of points in a ternary (we hope) it is probably OK if ternary inside does the duplication to avoid surprises on the other end.

It would be better if ternary can do the duplication on the C side.

PR #7172, opened 2022-12-24
Do ternary duplication if input is read-only

PaulWessel:
See pyGMT [post](https://github.com/GenericMappingTools/gmt/pull/7171#pullrequestreview-1229640903) for discussion.

@seisman, please see if this works as expected - I am of for Xmas dinner...


seisman (approved):

PaulWessel:
The PyGMT team may wish to check what other modules modify input.  I do not think we will rewrite all modules to internally do this though, in that case simpler to pass the **DUPLICATE** flag.

PR #7175, opened 2022-12-25
Quiet a bunch of annoying compiler warnings

PaulWessel:
Mostly unused variables.  Removed some, added others to debug output (-**Vd**).  All tests still work as before.

### DIFF
--- a/src/geodesy/psvelo.c
+++ b/src/geodesy/psvelo.c
@@ -991,7 +991,7 @@ EXTERN_MSC int GMT_psvelo (void *V_API, int mode, void *args) {
 	double eps1 = 0.0, eps2 = 0.0, spin = 0.0, spinsig = 0.0, theta = 0.0, *in = NULL;
 	double direction = 0, small_axis = 0, great_axis = 0, sigma_x, sigma_y, corr_xy;
 	double t11 = 1.0, t12 = 0.0, t21 = 0.0, t22 = 1.0, hl, hw, vw, ssize, headpen_width = 0.0;
-	double z_val, e_val, value, scale, size, i_value, nominal_size;
+	double z_val, e_val, value, scale, size, i_value;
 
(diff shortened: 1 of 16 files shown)


joa-quim (approved):

PaulWessel:
@joa-quim perhaps you can elaborate on wh there are these (I think perhaps your comments?):

```
/Users/pwessel/UH/RESEARCH/CVSPROJECTS/GMTdev/gmt-dev/src/gmt_init.c:17908:17: warning: variable 'i' set but not used [-Wunused-but-set-variable]
        int error = 0, i = 0, q = 0;    /* The i and i+= gmt_M_more_than_once are there to avoid compiler warnings... */
                       ^
1 warning generated.
[245/287] Building C object src/CMakeFiles/supplements.dir/potential/gravprisms.c.o
/Users/pwessel/UH/RESEARCH/CVSPROJECTS/GMTdev/gmt-dev/src/potential/gravprisms.c:1066:14: warning: unused variable 'th' [-Wunused-variable]
                openmp_int th = 0, row, col, n_columns = (openmp_int)G->header->n_columns, n_rows = (openmp_int)G->header->n_rows;      /* To shut up compiler warnings */
                           ^
1 warning generated.

```

joa-quim:
No idea about the first case. ``git blame`` says you wrote that line.

Regarding the second, I don't get it why the compiler screams but I have dozens like these. Happens when variables are used only inside OMP blocks.

PaulWessel:
OK, fixing the first like I did for some of the others.  As for the OpenMP Win stuff it is probably just because they are stuck at an old and low OpenMP level so we just have to live with it.

PaulWessel:
Oh, I see, the th is defined and set to 0 even if not using OpenMP and hence it is unused.  I just put it under another ifdef and that takes care of that message for me when openMP is not active.

PR #7178, opened 2022-12-26
Just clean up macro calls and reports

PaulWessel:
Odd sections to avoid various compiler warnings have been clarified a bit. No change in functionailty hence merged to master.


PaulWessel:
Will merge to master if nobody is up yet!

seisman (approved):

PR #7197, opened 2023-01-07
RPN documentation improvements

PaulWessel:
We have two modules (**gmtmath** and **grdmath**) that use Reverse Polish Notation (RPN).  This PR adds a figure with caption to both modules (the same figure/caption) that shows a typical Hewlett-Packard RPN calculator and the inventor of postfix (RPN) notation, with a Wiki link.

It was fun to also find that the [link](https://en.wikipedia.org/wiki/Reverse_Polish_notation) we use to Reverse Polish Notation on Wikipedia lists various computer languages (e.g., PostScript) and programs (e.g., **grdmath**!) that uses RPN.  So a reasonable addition to the documentation I think.


joa-quim (approved):

Esteban82:
Great.

They said that "we are stuck in “print mode”. So, I think that we could also add links to youtube videos.

For example this video helped me to understand RPN:

https://www.youtube.com/watch?v=7ha78yWRDlE&ab_channel=Computerphile

PaulWessel:
I like that video. Maybe we can add a short section to the g*math rat files the links o it for this who want to know more? Maybe a section before SEE ALSO like this

More on Reverse Polish Notation

Reverse Polish Notation (RPN) or postfix notation is widely used in computer science because of the simplicity of implementing a stack-based computation system.  The PostScipt language we used to make GMT graphics is postfix, for instance.  To learn more about RPN or postfix, watch this YouTube video explanation: <link here>

Perhaps @Esteban82 could try to make it? Create an RST include file (with extension .rst_) that we can include in both module rst files.

Issue #5153, opened 2021-04-18
More FORMAT_GEO_MAP/OUT format options


KristofKoch:
**Description of the desired feature**

Currently the formatting options for geographic coordinates offers quite a few options but might benefit from some more. 

### **Allow *F* and *G* in front of string**

Currently the WESN sign is only allowed after the numbers. I could make good use of them in front of the numbers as well.

***Current behaviour:***

`ddd:mm:ss[F|G]`  -> `52°37'19"N` or `52°37'19" N`

***Proposed addition:***

`[F|G]ddd:mm:ss` -> `N52°37'19"` or `N 52°37'19"`

The proposed leading `F` option is common in aviation context.

### **Leading/trailing Zeroes**

In aviation usually all places of a coordinate are written even if they are zero. This is done to make sure not digit is accidentally missing. To my knowledge currently single digit degree latitude and single or double digit degree longitude can only be written as `1°N`, `3°E`, `47°W` and not as `01°N`, `003°E`, `047°W`. Similar with arc minutes and seconds. They are only plotted if not zero. 

***Current behaviour (three degrees north latitude, nine degrees west longitude):***

`ddd:mm:ssF` -> `3°N`, `9°W`

***Proposed addition:***

`DDD:MM:SSF` -> `03°00'00"N`, `009°00'00"W`

While you might have even better ideas how to implement this I found capitalising the letters quite nice as it maintains backwards compatibility 

**Both proposals together**

`FDDD:MM:SS` -> `W009°00'00"`

PaulWessel:
These are good suggestions, @KristofKoch.  Once EGU is over etc I will have a look if your syntax is the best option or not, but I agree with the functionality.  6.3 though.

anbj:
PS! https://github.com/GenericMappingTools/gmt/issues/3562

KristofKoch:
I found myself needing the `W009°00'00"` output format again and was about to write a feature request – turns out I wrote one already. 

Is this still considered a useful addition?

PaulWessel:
Obviously this got lost in the shuffle when there are too many issues pending.  I will refresh my memory of how this could be done.

KristofKoch:
Good morning Paul, thank you for looking into this again. No worries, it is nothing a visit to a vector graphics program can't fix. It is just tedious work doing it manually.

PR #7215, opened 2023-01-14
Allow format flags F and G in FORMAT_GEO_MAP to come first or last

PaulWessel:
See https://github.com/GenericMappingTools/gmt/issues/5153 for background.  This PR seeks to implement that scheme. The **FORMAT_GEO_OUT** documentation now explains that the **F** and **G** can be used with **FORMAT_GEO_MAP** as either prefix or suffix.  Seems to work but need @KristofKoch to give it a better test.  Note that it does not work with **FORMAT_GEO_OUT** since the result would be data that GMT would only recognise as text (since it starts with a letter).  If there is a need to produce data output like

```
W23:55 S19:44
E16:05 N05:15

```
then we would need to recognise the above as valid data ascii input when reading it.  Not sure if we want to go that road yet.

### Example:

`gmt basemap -R-1/1/-1/1 -JM15c -Baf -png map --FORMAT_GEO_MAP=Gddd.mm`

map

We should probably add a test.  I can imagine a 2x2 subplot with this example using F and G as suffix or prefix which gives 4 cases.

### DIFF
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -292,6 +292,8 @@ FORMAT Parameters
         **G**      Same as **F** but with a leading space before suffix
         ========   =================================================================
 
+        **Note**: **F** and **G** can either be the first or last character in the template.
+
     **FORMAT_FLOAT_MAP**
         Format (C language printf syntax, see :term:`FORMAT_FLOAT_OUT`) to be used when plotting double
         precision floating point numbers along plot frames and contours [default is **%.12g**].
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -288,11 +288,11 @@ FORMAT Parameters
         **mm**     Fixed format integer arc minutes
         **ss**     Fixed format integer arc seconds
         **.xxx**   Floating fraction of previous integer field, fixed width
-        **F**      Encode sign using WESN suffix or prefix
-        **G**      As **F** but with a leading space before suffix or after prefix
+        **F**      Encode sign using WESN suffix
+        **G**      Same as **F** but with a leading space before suffix
         ========   =================================================================
 
-        **Note**: If used, **F** and **G** must be the first or last character in the template.
+        **Note**: With :term:`FORMAT_GEO_MAP`, **F** and **G** may also be used as a prefix.
 
     **FORMAT_FLOAT_MAP**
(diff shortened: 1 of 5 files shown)


PaulWessel:
I am thinking perhaps we either ban outright or warn that if **FORMAT_GEO_OUT** uses G (which adds a space) then the output is not compatible with GMT data format anymore and cannot be recognised properly:

```
echo -13 14 | gmt convert --FORMAT_GEO_OUT=ddd:mmG -fg
13:00 W	14:00 N

```
Output is geographic but because of **G** has that extra space between numbers and suffices.  If we then try to read that back in:

```
echo -13 14 | gmt convert --FORMAT_GEO_OUT=ddd:mmG -fg | gmt convert -fg
13	W	14:00 N
```

We cannot get past the columns with just a W or E and the whole thing becomes just trailing text as is.  I see no use for this.  What do you think: Prevent it or just warn when parsing **FORMAT_GEO_OUT**?  I guess the latter is the most flexible since you may want to format a data set for a completely other non-GMT use.

KristofKoch:
Hi Paul, leading `[F|G]` worked great in my limited testing so far. I noticed one thing:

When using `--FORMAT_GEO_MAP=Fddd:mmF` (which is clearly not formatted correctly with the double occurrence `F` at the beginning and at the end) only the last occurrence is honoured. Maybe a warning would be a nice idea? Just as a heads up in the sense of "Hey, your formatting can't work this way! Figure it out, in the meantime we just use the last occurrence of [F|G].".

KristofKoch:
> What do you think: Prevent it or just warn when parsing **FORMAT_GEO_OUT**? I guess the latter is the most flexible since you may want to format a data set for a completely other non-GMT use.

I agree with you and would just warn. While I mostly use floating point for the internal stuff there might be non-GMT use cases where this could be desirable.

PaulWessel:
Thanks @KristofKoch.  For now, I have added checks for multiple F, G or combinations and they get the general error message (there are too many error possibilities we check for to give individual messages).  I now warn if you use trailing G in FORMAT_GEO_OUT since it adds that space.  For now, since the output function does not yet handle pre-number hemisphere I have just added commented-out warning to remind me to eventually let GMT write out the bad format and open this message.

KristofKoch (approved):
I like it. That error message gives me a nice starting point for investigating what I screwed up. Thank you, @PaulWessel!

PR #7221, opened 2023-01-24
Centralize info on character escape sequences in separate file

### LINKED ISSUE #7210: Documentation of character escape sequences in text module and Cookbook not ideal

The documentation for the character escape sequences exists in both the [`text` module](https://docs.generic-mapping-tools.org/dev/text.html#description) and in the [Cookbook](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#character-escape-sequences). Neither one is really good and I would like to improve the documentation.

**Identified areas for improvement:**
_(in no particular order)_

1. The wording in the tables is different (_toggle_ vs _turn on/off_, I prefer _turn on/off_ as it is easier to understand for people with basic English skills)
2. LaTex option is missing completely in the Cookbook and is not in the table in `text`, just some paragraphs down as text. Easy to overlook that powerful option this way.
3.  The various accented European characters are presented in a nice table in the Cookbook but for some reason in `text` only the shorthands are given instead of the resulting characters. Would be better to have the supported characters on hand and then link to the table for the shorthand?
4. Maybe move the details completely into the Cookbook and just link to it in `text`? Only one place to maintain them.

The issue was brought to my attention [by user Andreas in the forum](https://forum.generic-mapping-tools.org/t/greek-symbols-and-escape-characters/3588).

What do @GenericMappingTools/core and @GenericMappingTools/gmt-maintainers think about this? Do you consider it worthwhile? If yes, I'm happy to work on the improvement of the documentation.

KristofKoch:

The details on how to use the character escape sequences were available in the Cookbook and in the documentation of the  `text` module. There were some differences between them. I summed them up in a separate file and included it in both places. Slight re-wording was necessary to make it fit into the respective page.

Please check especially for language and coherence before merging.

Fixes #7210


joa-quim:
@KristofKoch could look at the conflicts and fix them? We should go ahead with this and merge it.

KristofKoch:
Will have a look at it & report back

PR #7237, opened 2023-02-11
Fix the -V and @remotes case

### LINKED ISSUE #7236: -V is ignored when processing @remote files
Example
```
grdcut @earth_relief_10m  -R-12.5/-10/35.5/37.5 -Vq -Glixoo.grd
unknown [WARNING]: Remote dataset given to a data processing module but no registration was specified - default to gridline registration (if available)
```
That happens because when ``gmt_set_unspecified_remote_registration()`` is called the common options have not yet been parsed.

joa-quim:
Fixes #7236

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -15224,6 +15224,11 @@ struct GMT_CTRL *gmt_init_module (struct GMTAPI_CTRL *API, const char *lib_name,
 	/* First handle any halfhearted naming of remote datasets where _g or _p should be appended */
 
 	if (options) {
+		/* Treat -V in advance because gmt_set_unspecified_remote_registration needs to already know about it.*/
+		for (opt = *options; opt; opt = opt->next) {
+			if (opt->option == 'V')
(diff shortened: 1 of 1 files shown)


PaulWessel (commented):
Since gmt.c checks --V up front and passes the mode to GMT_Create_Session, which then sets API->verbose, I suspect you could just use API->verbose here instead of read-scan the entire option string.  See how v_mode is set in gmt.c and how that changes mode which is passed to GMT_Create_Session and then picket up right away.

joa-quim:
I did see with the debugger that at that point where I inserted the call to ``gmt_parse_common_options`` GMT->current.setting.verbose was still == 3, which means **-Vq**  has not been  seen (or was ignored) before this.
I don't know how to solve this in anther way.

PaulWessel:
Probably, but what does API->verbose say?

joa-quim:
I think it said = 0.

PaulWessel (approved):
Okay

PR #7243, opened 2023-02-12
Remove references to ncBrowse that seems to have evaporated from internet

joa-quim:
and is causing consecutive failures from the link checker bot.


PaulWessel (approved):
Fine. I had some emails with them but nothing really resolved, so good riddance

PR #7245, opened 2023-02-13
Add a sub-structure to the API main structure to hold info of interest to GMT.jl

### LINKED ISSUE #7244: Smooth communication with Julia
Julia offers a very powerful mechanism to communicate with C (the ``ccall`` function), which is the base of ``GMT.jl`` deep communication between Julia and the GMT C lib. However, even with that facility there are chunks of info that are quite challenging to intercommunicate. I'm, referring specially to the info that is stored in the ``API`` structure and all of it's dependents. For one side that struct is **big** and for the other it has changed along the time to respond to new needs. Such a moving target is nearly impossible to fully wrap in practice. But we can make some compromises if a transfer structure is put as the first member of ``API`` and never move it out of that position. That transfer struct member can then hold interesting info, like the name of the session dir and the type of table columns (time, geog, cartesians, etc...) and it can even be changed according to future needs without changing any of the GMT internal machinery.

A PR will follow.

joa-quim:
Fixes #7244

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -7556,6 +7556,11 @@ GMT_LOCAL void * gmtapi_get_data (void *V_API, int object_ID, unsigned int mode,
 	gmtapi_set_object (API, S_obj);
 	//gmtapi_list_objects (API, "gmtapi_get_data");
 #endif
+	
+	/* 4GMT.jl For start, copy only 64 columns instead of the GMT_MAX_COLUMNS */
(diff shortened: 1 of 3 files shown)


PaulWessel (approved):
OK, but would prefer things like

gmt_M_memcpy (&API->jl_pocket.col_type[0], &API->GMT->current.io.col_type[0], 64,  int);

And spell "Ideally, " line 129, and perhaps add something like "Please do not change anything here without first checking with Joaquim Luis" or whatever you thing is reasonable.

PR #7252, opened 2023-02-18
Change order of GMT_CTRL members.

joa-quim:
To fully wrap structures from Julia one must know exactly the size of each of those members. That is simple for scalars and pointers types but may become prohibitively complicated for members that structs that contain themselves a chain of substructures. That is the case of the GMT_CTRL struct. But we can do a kind of partial wrap where we care to access only to a subset of its members. Moving ``struct PSL_CTRL *PSL;`` to first position allows wrapping the  ``PSL_CTRL`` and hence have access to its own members.

This change should not matter for the C side of things and likely will even make it slightly more memory friendly.
Run a subset of the tests and nothing bad showed up.

### DIFF
--- a/src/gmt_types.h
+++ b/src/gmt_types.h
@@ -460,13 +460,13 @@ struct GMT_SESSION {
 
 struct GMT_CTRL {
 	/* Master structure for a GMT invocation.  All internal settings for GMT is accessed here */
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Yes, can't imagine this will have any effect on CLI - but hope you ran all tests just the same before pushing :-)

PR #7283, opened 2023-02-28
Better solution for the grdview draping issues.

joa-quim:
Now the gmt_M_file_is_image(file) macro is capable of detecting if  file is a in-memory file containing an GMTimage and hence the previous patches in grdview are no longer necessary.


joa-quim:
Tests pass. The new ones were failing in because the Orig PS were generated with my local page settings. 
Also updated the PS files in DVC, so let's see if all those bots are finally satisfied.

Issue #7282, opened 2023-02-28
van der Grinten projection (-JV) fails in non -Rg domains


KristofKoch:
**Description of the problem**

When plotting a global map with the *van der Grinten* projection (`-JV`) everything behaves as expected, when using it to plot quadrants or hemispheres things go wild: Instead of quadrants or hemispheres I get modern art.

**Full script that generated the error**

```
gmt begin van_der_Grinten_quadrants_hemispheres png
  gmt subplot begin 2x4 -Fs8c/16c -M8c/1c -T"van der Grinten quadrant & hemisphere maps"
    while read minX maxX minY maxY title
    do
    gmt subplot set
      gmt coast -R${minX}/${maxX}/${minY}/${maxY} -JV0/8c -Bg -Dc -Glightgray -Scornsilk -A10000 -Wthinnest -B+t"${title}"
    done << EOF
-180   0   0 90 NW
   0 180   0 90 NE
   0 180 -90  0 SE
-180   0 -90  0 SW
-180   0 -90 90 W
-180 180   0 90 N
   0 180 -90 90 E
-180 180 -90  0 S
EOF
  gmt subplot end
gmt end show
```
If you prefer a one-liner use the following for a plot of the NW quadrant:
```
gmt coast -R-180/0/0/90 -JV0/10c -Bg -Dc -Glightgray -Scornsilk -A10000 -Wthinnest -view van_der_Grinten_NW_quadrant
```

**Full error message**

*No error messages generated*

**Actual outcome**

![van_der_Grinten_quadrants_hemispheres](https://user-images.githubusercontent.com/45796239/221960467-3361f95d-2de9-4fa8-8acc-522adb6ae73d.png)

**Expected outcome**

Four quadrant maps in the first row and four hemisphere maps in the second row in above plot.

**System information**

* Operating system: MacOS 13.2.1
* GMT version (`gmt --version`): 6.5.0_51b8cba_2023.02.25 also fails with 6.4.0

KristofKoch:
Additionally similar things happen with the *Winkel Tripel* projection (`-JR`). Maybe related?

Trying to plot a western hemisphere:

Central meridian is on the domain border (0°): **fail**
![winkel-0](https://user-images.githubusercontent.com/45796239/221969239-57bd66ef-1da6-44aa-9a4f-00e2ee9432fc.png)
```
[...]

PR #7287, opened 2023-03-02
Fix incorrect Van Grinten map clip path

PaulWessel:
The Van Grinten is similar to, say, Hammer, in that it has curved meridians and parallels.  However, the code had it placed with other cases who either had pole point or straight parallels, leading to stuff @KristofKoch reported in https://github.com/GenericMappingTools/gmt/issues/7282.  This PR just fixes the calculation of the map clip path - there may be other issues but tackling those next (assuming they are still there - I think so since not giving central longitude works in general OK...).

Below is from master, then from this PR.  All tests are unaffected (since not testing the central longitude).:

Master:

bad

This PR:
good

### DIFF
--- a/src/gmt_map.c
+++ b/src/gmt_map.c
@@ -8654,7 +8654,6 @@ uint64_t gmt_map_clip_path (struct GMT_CTRL *GMT, double **x, double **y, bool *
 			case GMT_AZ_EQDIST:
 			case GMT_ALBERS:
 			case GMT_ECONIC:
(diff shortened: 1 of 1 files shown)


anbj:
Not completely acquainted with this issue, but this is how the script in #7282 looks at my end.
van_der_Grinten_quadrants_hemispheres

joa-quim (approved):

joa-quim (approved):

PR #7291, opened 2023-03-03
Let -JV take optional meridian without failure

PaulWessel:
Well, at least the original troubles reported in #7282 are fixed.  The final issue appreas to have been that when the region was negative we shifted it in **coast** by 360 but unfortunately the check for doing the same to the central meridian had `<` instead of `<= `so it was left behind to cause chaos, I added another test _pscoast_JV_set.sh_ which uses a fixed Greenwich as the central meridian and those quadrants and hemispheres now work:

pscoast_JV_set

### DIFF
--- a/src/pscoast.c
+++ b/src/pscoast.c
@@ -1024,7 +1024,7 @@ EXTERN_MSC int GMT_pscoast (void *V_API, int mode, void *args) {
 	if (GMT->common.R.wesn[XLO] < 0.0 && GMT->common.R.wesn[XHI] <= 0.0) {	/* Temporarily shift boundaries */
 		GMT->common.R.wesn[XLO] += 360.0;
 		GMT->common.R.wesn[XHI] += 360.0;
-		if (GMT->current.proj.central_meridian < 0.0) GMT->current.proj.central_meridian += 360.0;
+		if (GMT->current.proj.central_meridian <= 0.0) GMT->current.proj.central_meridian += 360.0;
 	}
 	if (need_coast_base) {
 		west_border = floor (GMT->common.R.wesn[XLO] / c.bsize) * c.bsize;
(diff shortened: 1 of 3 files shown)


PaulWessel:
If I change the projection from V to R, N, K, H, W etc they all work it seems

Esteban82 (approved):

joa-quim:
But we still have more problems of this kind
```
gmt coast -R-180/2/-90/90 -JR1/12c -Bg -W0.5 -png lixo
```
lixo

PaulWessel:
Yep. Nothing is foolproof yet but let me know if **basemap** does the same or if it is a **coast** thing.

PaulWessel:
So, what is happening here is taht if you set a central meridian at lon = 1, then the most western longitude on a global map is lon = -179 since then the distance from central meridian to west is 180 degrees.  Setting it to -180 is actually a point on the east end and GMT goes nuts.  For instance, if I modify your example slightly along these lines:

gmt coast -R-179/2/-90/90 -JR1/12c -Bg -W0.5 -png lixo

then we get

lixo

So I think these failures should be caught with a check that says

`basemap [ERROR]: Western boundary is > 180 degrees from central meridian and thus invalid.`

PaulWessel:
Just noticing that for most of the cylindrical projections, we have a check that override bogus things like this and recenter the central longitude in the middle.  If I add the same call (_gmtmap_cyl_validate_clon_) to _gmtmap_init_winkel_ then you sample yields

lixo

with central longitude forced to be half-way from w to e.  Given this has been the default action in Mercator, Cylindrical Equal-Area, Cylindrical Equidistance, Miller, Cylindrical Stereographic, I guess some options are:

1. Impose the same on similar projections (Winkel, Robinson, Grinten, etc etc which have the same issue)
2. Decide that if user specified a central longitude and it is > 180 degrees from west or east we yield a (new) error message

Thoughts?

PR #7292, opened 2023-03-03
Use lower case header name that seems to bother some MinGW versions

joa-quim:


PaulWessel:
no worries as log as also working on windows for you.

PaulWessel (approved):

joa-quim:
It the wonderful world of MinGW. It doesn't bother my MinGW's neither the local one of Boris but it *maybe* does on the bot. BTW, is the Boris the same one you talked in Norway about GMT.jl?

https://github.com/JuliaPackaging/Yggdrasil/pull/5199#issuecomment-1453573045

PR #7295, opened 2023-03-05
Make stricter checks on central longitude

PaulWessel:
There are a few cases where GMT was inconsistent and wrong - we will label these as bugs:

1. When the user specified a specific central longitude, it is to be respected, and for a global map the w/e borders are set at 180 distance from that longitude.
2. If _not_ a global map then if the distance from the specified central meridian to either border exceeds 180 then we yield a map region error.  Currently, something slightly different may happen if `-R's +r` is used for rectangular areas.
3. If _no_ central meridian is set then it continues to default to the mid longitude.

Tests pass and some of the odd and strange plots found in #7291 are now dealt with.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -5249,7 +5249,7 @@ GMT_LOCAL bool gmtinit_parse_J_option (struct GMT_CTRL *GMT, char *args_in) {
 	gmt_M_memset (l_pos, 3, int);	gmt_M_memset (p_pos, 3, int);
 	gmt_M_memset (t_pos, 3, int);	gmt_M_memset (d_pos, 3, int);
 	if (!GMT->common.J.active)	/* Down want to clobber this during -Jz/Z after the horizontal part has been set */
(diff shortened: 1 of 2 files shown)


joa-quim (approved):

seisman:
The following command no longer works after this PR:
```
gmt basemap -Bag -Bzag -Ggreen -JR270/10c -Jz5 -R10/70/-5/10/0/1 -p225/30 -pdf map                           
basemap [ERROR]: Western boundary is > 180 degrees from specified central meridian and thus your region is invalid
basemap [ERROR]: General map projection error
```

seisman:
Before this PR was merged, the command produces the following figure. As you can see, the location of the z-axis doesn't match well with the map boundary. Is it intentional?
map

PaulWessel:
Couple of issues I guess:

1. Since Winkel Tripel has separate west and east borders, they cannot be more than 180 degrees from the center, and the center longitude cannot be outside the range.  This is a new check to prevent crazy situations.
2. Since no **+r** is used in setting the region, the sliver of Winkel map given w/e/s/n has variable height.  The back wall mechanism has no knowledge of this and will not draw some curved back wall.

Probably best to improve the docs for -B on this.

Issue #7226, opened 2023-01-31
Use more Math markup in documentation


PaulWessel:
**Description of the desired feature**

Given that our Restructured Text documentation fully support mathematics, it would be nice to try to use it when we do math.  If I cd into the doc source directory and search for things like "raised to the second power" which we do for area units and simple math, I find lots of places that could be improved very simply:

```
cd doc/rst/source/
find . -type f -exec grep -H '\^2' {} \;


./supplements/gshhg/gshhg.rst:north* is the enclosing rectangle, *area* is the polygon area in km^2
./supplements/mgd77/mgd77list.rst:        7.5038 \* V \* cos(lat) \* sin(az) + 0.004154 \* V^2
./supplements/mgd77/mgd77list.rst:(1) g = 978052.0 \* [1 + 0.005285 \* sin^2(lat) - 7e-6 \* sin^2(2\*lat)
./supplements/mgd77/mgd77list.rst:+ 27e-6 \* cos^2(lat) \* cos^2(lon-18)]
./supplements/mgd77/mgd77list.rst:(2) g = 978049.0 \* [1 + 0.0052884 \* sin^2(lat) - 0.0000059 \*
./supplements/mgd77/mgd77list.rst:sin^2(2\*lat)]
./supplements/mgd77/mgd77list.rst:(3) g = 978031.846 \* [1 + 0.0053024 \* sin^2(lat) - 0.0000058 \*
./supplements/mgd77/mgd77list.rst:sin^2(2\*lat)]
./supplements/mgd77/mgd77list.rst:(4) g = 978032.67714 \* [(1 + 0.00193185138639 \* sin^2(lat)) / sqrt (1
./supplements/mgd77/mgd77list.rst:- 0.00669437999013 \* sin^2(lat))]
./supplements/mgd77/mgd77list.rst:99999999 mag 0.5\*cos(0.5\*(azim-19))^2 1.0\*exp(-1e-3(lat))^1.5
./supplements/potential/gmtflexure.rst:    or Young's modulus [7.0e10 N/m^2], respectively.
./supplements/potential/grdflexure.rst:    or Young's modulus [7.0e10 N/m^2], respectively.
./supplements/x2sys/x2sys_solve.rst:    **g** will fit f(**p**) = *a* + *b* sin(y)^2
[...]

PR #7302, opened 2023-03-07
DwM gmtbinstats.rst

anbj:
More work on #7226.

Please check that the abbreviations are correct:

LMS = least mean square (LMS) scale
RMS = root mean square (RMS)


PaulWessel (commented):
LMS stands for Least Median Square.  It is a robust measure unlike a mean measure,  So please do not mix those up.  There is no Least mean square operator in GMT since that is the same as basic standard deviation, last-squares fitting etc,

anbj:
What the heck just happened with all these commits?

Thanks Paul - updated. May have to annihilate this PR and create a new one...

anbj:
Making a new one.

PR #7304, opened 2023-03-07
Do not allow -Xc|f in modern mode

PaulWessel:
A handful of stray modern mode scripts had left-over **-Xc** settings which does northing to affect the final cropped plot but does cause failure when comparing to orig PS since those are US letter size and the -Xc shifts things.  I have updated those few PS in dvc after removing the -Xc and recreating the outputs.  Other scripts were not affected and tests are similarly working.  The documentation and usage message now mention the limitation in modern mode.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -1133,6 +1133,10 @@ GMT_LOCAL int gmtinit_parse_X_option (struct GMT_CTRL *GMT, char *text) {
 		default:
 			GMT->current.ps.origin[GMT_X] = GMT->common.X.mode = 'r'; break;
 	}
+	if (GMT->current.setting.run_mode == GMT_MODERN && strchr ("cf", GMT->common.X.mode)) {	/* Modern mode: Disallow -Xc|f */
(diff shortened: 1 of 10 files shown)


src/gmt_init.c
```diff
@@ -1185,6 +1189,10 @@ GMT_LOCAL int gmtinit_parse_Y_option (struct GMT_CTRL *GMT, char *text) {
 		default:
 			GMT->current.ps.origin[GMT_Y] = GMT->common.Y.mode = 'r'; break;
 	}
+	if (GMT->current.setting.run_mode == GMT_MODERN && strchr ("cf", GMT->common.Y.mode)) {	/* Modern mode: Disallow -Yc|f */
+		GMT_Report (GMT->parent, GMT_MSG_ERROR, "Centered (-Yc) and fixed (-Tf) shifts are not available in modern mode\n");
```
seisman, line 1193:
```suggestion
		GMT_Report (GMT->parent, GMT_MSG_ERROR, "Centered (-Yc) and fixed (-Yf) shifts are not available in modern mode\n");
```

seisman (approved):

PR #7308, opened 2023-03-08
explain_-A.rst_, explain_gshhg.rst_ and gmtselect.rst: create link to…

anbj:
* create link in *explain_gshhg.rst_* and link from *explain_-A.rst_*
* remove double dot in *gmtselect.rst*


anbj:
No math markup needed, so this can be marked off in #7226.

PR #7309, opened 2023-03-08
Add modifier +h[k|s] to gmt select -Z

PaulWessel:
See the Forum [post](https://forum.generic-mapping-tools.org/t/selection-based-on-segment-header-value-zval/3733) for background.  This modifier allows us to pick the segments fixed z-value via `-Z<string>` in the segment header to be used in the **-Z** range or value check.  The optional argument **k** means "keep records whose segment header has no `-Z<value>` statement".  The default is to skip the entire segment, i.e., **+hs**.

I added a test to demonstrate that it works.


Esteban82 (approved):
Works fine!

joa-quim (approved):
I take it that you got tired of fighting the BitLocker.

PaulWessel:
> I take it that you got tired of fighting the BitLocker.

Yes, sent a bunch of question to SOEST IT since they should know how - maybe even a zoom remote view assisted thing could work if they tell me what to push and type.

PR #7312, opened 2023-03-08
Replace mktemp with secure versions

PaulWessel:
See #7101 for background. Means an `ifdef` on Windows to get the underscored version __mkdemp_s_.  Hopefully, @joa-quim can test this branch and let us know if this works OK under both pure Windows and MinGW. Obviously no change on macOS or Linux.

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -87,6 +87,7 @@
 #include <glob.h>
 #else
 #include <windows.h>
+#include <io.h>
 #include <tchar.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


joa-quim:
Yep, things got a bit more complicated now that I have to deal also with MinGW. How can I test this?

joa-quim:
Yep but tests on Windows (at leas mine) are not very trustful. Many failures for various stupid path errors.

PaulWessel:
I think it would be great if Windows testing was more robust.  Can anything be done?  Is it the same in the CI as on your laptop?

remkos:
Sorry for the late response on this and not doing the actual implementation. I wanted, in fact, to remove the race condition that still exists this way. This solution leaves the race condition in place, because `gmt_support_make_template` creates the file name, and only *another* `fopen` will open the file. To really get rid of the race condition you need to use `fd = mkstemp (path)` instead.

To do this right, we need to get rid of `gmt_support_make_template` in its entirety. But I'm not sure how to do it.

PaulWessel:
I see.  Well, I was more concerned about eliminating the warning which is now gone.  I cannot imagine we well have race conditions in GMT where this function is called maybe a few times in a single process?

PR #7315, opened 2023-03-09
Let grdimage share -T with grdview

PaulWessel:
This PR is creating a function of out **grdview -T** so it can be called from grdimage as well as **grdview**.  We can consider extending this to handing images (which **grdimage** may be fed) but for **now** it is like **grdview -T.**

Tested on @anbj script and seems to work fine - no test issues.  However, need @joa-quim to try this from Julia and passing memory since it is at least possible that something different needs to be done when I call `goto basemap_and_free` he end.

### DIFF
--- a/doc/rst/source/grdimage.rst
+++ b/doc/rst/source/grdimage.rst
@@ -238,7 +238,7 @@ place with most map projections. Because **grdimage** uses the
 PostScript colorimage operator, for most non-linear projections we
 must resample your grid onto an equidistant rectangular lattice. If you
 find that the NaN areas are not treated adequately, consider (a) use a
-linear projection, or (b) use :doc:`grdview` **-Ts** instead.
+linear projection, or (b) use **-T+s** instead to plot graticule polygons.
 
 .. include:: explain_grdresample.rst_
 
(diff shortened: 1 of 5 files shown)


joa-quim (approved):
No problems from the Julia side.

anbj:
Fantastic - thank you Paul!

Just a PS: Skip nodes with NaN/NaN values transparent: ``-Q`` has no effect it ``-T`` is used, and basically replaced with ``T+s``. Worth noting in the docs?

anbj:
One more note: ``-T+oblack+s -t50`` makes the cells transparent, as expected. ``-T+oblack@50+s -t50`` makes nothing transparent ('cancels' the ``-t50``). I don't know if ``-T+oblack`` should even accept a ``@50``. Just testing!

PaulWessel:
You could see if #7322 makes things more stable?

anbj:
``-T+oblack+s -t40`` - works as expected
These do not;
``-T+oblack@20+s -t50``:

tell-antall-jordskjelv-i-noder

``-T+oblack@20+s``:

tell-antall-jordskjelv-i-noder

Issue #7316, opened 2023-03-09
Add a test program to show how to correctly pass a user-created GMT_GRID structure to GMT API?


seisman:
Recently I'm working on wrapping the GMT_GRID structure in PyGMT. I want to convert a `xarray.DataArray` object (grid data in the Python ecosystem) to `GMT_GRID` and pass it to GMT modules, but I'm always getting `pointer being freed was not allocated` or `double free` errors.

Here is a C program that shows what I'm doing in PyGMT. 
```C
#include "gmt.h"

int main () {
	unsigned int mode = GMT_SESSION_EXTERNAL;
	struct GMT_GRID *G = NULL;
	struct GMTAPI_CTRL *API = NULL;
	uint64_t dim[2] = {5, 4};
	char args[1000] = {""};
	char input[GMT_VF_LEN] = {""};
	float *data = NULL;

	API = GMT_Create_Session ("test", 2U, mode, NULL);

	/* create a container for the grid */
	G = GMT_Create_Data(API, GMT_IS_GRID, GMT_IS_SURFACE, GMT_CONTAINER_ONLY, dim, NULL, NULL, 0, 0, NULL);

	/* allocate memory of the data array in the external program (C or PyGMT) */
	data = (float *)malloc(sizeof(float) * 5 * 4);
	for (int i=0; i<5*4; i++) data[i] = (float)i;

	/* assign the user data to the GMT_GRID structure */
	G->data = data;

	/* Using the grid */
	GMT_Open_VirtualFile (API, GMT_IS_GRID, GMT_IS_SURFACE, GMT_IN|GMT_IS_REFERENCE, G, input);
	sprintf (args, "%s", input);
	GMT_Call_Module(API, "grd2xyz", GMT_MODULE_CMD, args);
	GMT_Close_VirtualFile (API, input);

	/* Trying to free the data array because it's allocated by the external program. */
	free(data);

	if (GMT_Destroy_Session (API)) return EXIT_FAILURE;
	exit (0);
}
```
After compiling and running the program, I'm getting the following results:
```
./a.out
0	3	0
1	3	1
2	3	2
3	3	3
4	3	4
0	2	5
1	2	6
2	2	7
3	2	8
4	2	9
0	1	10
1	1	11
2	1	12
3	1	13
4	1	14
0	0	15
1	0	16
2	0	17
3	0	18
4	0	19
[...]

PR #7320, opened 2023-03-10
Redo how alloc_mode is set in objects

PaulWessel:
See #7316 for background.  This PR lets the _alloc_mode_ remain unset (**GMT_ALLOC_EXTERNALLY**) until a grid, cube ,etc are actually allocated by GMT, at which point we set it to **GMT_ALLOC_INTERNALLY**.  Quite a few other changes had to be made to work with the new scheme, especially odd handling in **grdmath** and **grdblend**, but now all tests pass.

However, it is important that the wrappers try this new branch to see if anything else shows up since it is more likely that there are subtle things related to passing memory that the CLI tests dont exercise.  So I need an approval from you all before we merge this one.

@seisman, I removed the _GMT_Set_AllocMode_ call from `testapi_gmtgrid.c` since the new scheme no longer incorrectly sets that alloc_mode to  **GMT_ALLOC_INTERNALLY**.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -608,13 +608,18 @@ GMT_LOCAL inline void gmtapi_update_prev_rec (struct GMT_CTRL *GMT, uint64_t n_u
 GMT_LOCAL int gmtapi_alloc_grid (struct GMT_CTRL *GMT, struct GMT_GRID *G) {
 	/* Use information in Grid header to allocate the grid data array.
 	 * We assume gmtapi_init_grdheader has already been called. */
+	struct GMT_GRID_HIDDEN *GG = NULL;
 	struct GMT_GRID_HEADER_HIDDEN *GH = NULL;
+
 	if (G == NULL) return (GMT_PTR_IS_NULL);
-	GH = gmt_get_H_hidden (G->header);
 	if (G->data) return (GMT_PTR_NOT_NULL);
 	if (G->header->size == 0U) return (GMT_SIZE_IS_ZERO);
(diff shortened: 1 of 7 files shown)


joa-quim:
I'm not sure of this, but it looks that ``gmt_create_grid`` does not set **GMT_ALLOC_EXTERNALLY** leaving it in *undefined* sate. That is not good. The ``alloc_mod`` has to be a binary thing only. Either ``EXTERNALLY`` or ``INTERNALLY``, not ternary because we want to be able to do ``if (alloc_mode == GMT_ALLOC_EXTERNALLY)`` and if false it means ``INTERNALLY``

PaulWessel:
Default is **GMT_ALLOC_EXTERNALLY** since that is 0.

joa-quim:
OK, but still better to set it explicitly in ``gmt_create_grid`` so even if the enum changes its value codes won't break.

PaulWessel:
OK, doing that now plus some other remaining assignments...

PaulWessel:
OK, have a look now @joa-quim

PaulWessel:
Great, we will await the @GenericMappingTools/pygmt-admin check of this branch first.

PaulWessel:
Hi @seisman @maxrjones @weiji14 etc.  Have you tried this branch with your PyGMT tests? Need to know if there are any odd side-effects for the wrappers.  Nothing in Julia so most likely no, but I dont want to merge until I hear from you.

seisman:
I'll run the full PyGMT tests tomorrow.

seisman (approved):
No side-effect on the PyGMT side.

PR #7326, opened 2023-03-12
Improve -T in grdimage/view for transparency

PaulWessel:
Hoping this helps with #7315. Perhaps @anbj can give it a spin.

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -10806,13 +10806,15 @@ void gmt_plot_image_graticules (struct GMT_CTRL *GMT, struct GMT_GRID *G, struct
 	 * G is the data grid
 	 * I is an optional intensity grid.  If NULL then either intensity points to a
 	 *    constant intensity or it is also NULL, meaning no intensity adjustment for colors.
-	 * P is the CPT in use
+	 * P is the CPT in use for fills
 	 * pen is an optional pen for drawing the graticules, or NULL
 	 * skip determines if we paint NaN polygons or not
-	 * intensity is pointer to a constant intensity or NULL
+	 * intensity is pointer to a constant intensity or NULL.
 	 */
 	openmp_int row, col;
 	uint64_t ij, n;
+	int outline = 0;
+	bool delay_outline = false;
 	double *xx = NULL, *yy = NULL, inc2[2] = {0.0, 0.0};
 	struct GMT_FILL fill;
 	struct GMT_DATASEGMENT *S = gmt_get_segment (GMT, 2);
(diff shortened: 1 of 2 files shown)


src/gmt_plot.c
```diff
@@ -10833,12 +10844,26 @@ void gmt_plot_image_graticules (struct GMT_CTRL *GMT, struct GMT_GRID *G, struct
...
+	if (delay_outline) {	/* Just get graticule outlines and draw them */
+ 		gmt_setpen (GMT, pen);	/* Set outline pen */
+		gmt_setfill (GMT, NULL, 1);	/* Turn off fill */
+	 	gmt_M_grd_loop (GMT, G, row, col, ij) { /* Visit each node */
+			if (gmt_M_is_fnan (G->data[ij]) && skip) continue;
+			if (I && skip && gmt_M_is_fnan (I->data[ij])) continue;
+			n = gmt_graticule_path (GMT, &xx, &yy, 1, true, G->x[col] - inc2[GMT_X], G->x[col] + inc2[GMT_X], G->y[row] - inc2[GMT_Y], G->y[row] + inc2[GMT_Y]);
+			S->data[GMT_X] = xx;    S->data[GMT_Y] = yy;    S->n_rows = n;
+			gmt_geo_polygons (GMT, S);
+			gmt_M_free (GMT, xx);
```
joa-quim, line 10862: Can't those guys be freed **outside** the loop?

anbj (approved):
I think this is good to go.

``-T+oblack+s -t40``, ``-T+oblack@20+s -t50`` and ``-T+oblack@20+s`` from https://github.com/GenericMappingTools/gmt/pull/7315#issuecomment-1464309852 now all give expected results. Let me know if you want graphical evidence.

src/gmt_plot.c
PaulWessel: No, unless you rewrite it. _n_ can vary from a few to tens or more depending on projection and **MAP_LINE_STEP**.  I did  not notice that it was particularly slow?

joa-quim:
Allocating and freeing inside a double loop can never be cheap. Couldn't ``gmt_graticule_path()`` receive static *x,y* arrays, sufficiently large, set the non-used elements to 0 for safety, and return the number of used elements?

PaulWessel:
Since you don't know the number needed, I would recommend something like this:

1. Pass the same arrays in, no freeing inside loop
2. Pass in a `size_t *n_alloc` argument, initially 0
3. Inside we check before each alloc if `n < *n_alloc`, and only alloc if too small.
4. Increase `*n_alloc` for each alloc (there are 4 sides)
5. Free arrays outside the loop in the end

This means in practice that we do most allocations early on and then we just recycle.  Or as you suggest, start with `n_alloc = GMT_LEN256` or something and then perhaps very little allocations happen.

If you have the time to do this - which I think is pretty simple, knock yourself out!

joa-quim:
These two weeks are quite busy but I'll see.

PR #7329, opened 2023-03-16
Watch if convert is getting -g with X, Y, or D

PaulWessel:
Since those measures requires a map projection which is not available in **convert** we cannot accept those arguments and thus give an error.


anbj:
`gD` fixed, but same issue for `-gX` and `-gY`.

anbj:
That is, ``-gD`` gives ``gmtconvert [ERROR]: The -g option cannot use X, Y, or D since no projection can be set. Use mapproject first``. ``-g[X|Y]`` still crashes.

PaulWessel:
You are right - i did not test that since it seemed so trivial. Hm, will see why.

anbj:
Yes, no crash for any of `-g[X|Y|D]` anymore!

PR #7334, opened 2023-03-23
Wrong divisor in grdmath D2DXY

PaulWessel:
See this forum [post](https://forum.generic-mapping-tools.org/t/grdmath-d2dxy-operator-usage/3773) for background on this **grdmath** bug.  Classic copy-paste error where we ended up dividing by dx^2 and dy^2 instead of just dx*dy.

After the fix proposed in this PR and setting the CPT for the **D2DXY** case to that of the others, we get

gaussian_test_diff

### DIFF
--- a/src/grdmath.c
+++ b/src/grdmath.c
@@ -2023,10 +2023,10 @@ GMT_LOCAL void grdmath_D2DXY (struct GMT_CTRL *GMT, struct GRDMATH_INFO *info, s
 
 	z = gmt_M_memory (GMT, NULL, info->size, float);
 	cx = gmt_M_memory (GMT, NULL, info->G->header->n_rows, double);
-	gmt_M_row_loop (GMT, info->G, row) cx[row] = 0.5 / (info->dx[row] * info->dx[row]);
+	gmt_M_row_loop (GMT, info->G, row) cx[row] = 0.5 / info->dx[row];
 
 	mx = info->G->header->mx;
-	cy = 0.5 / (info->dy * info->dy);
+	cy = 0.5 / info->dy;
 
 	gmt_M_grd_loop (GMT, info->G, row, col, node) {
 		z[node] = (float)(cx[row] * cy * (stack[last]->G->data[node-mx+1] - stack[last]->G->data[node-mx-1] + \


test/grdmath/d2dxy.sh
```diff
@@ -0,0 +1,39 @@
...
+		gmt colorbar -DJBC+o0c/1.25c+h -Ba0.1+l"@[ f(x,y) @[ " -Cz.cpt
+
+		gmt grdimage d2gaussdxdy_ex.nc -Ba2f1g1 -BWeSn+t"Exact Formula" -Bx+l"x" -By+l"y" -c
+		gmt colorbar -DJBC+o0c/1.25c+h -Ba0.1+l"@[ \frac{\partial^2 f}{\partial x \partial y} @[ "
+
+		gmt grdimage gauss_d2dxy.nc -Ba2f1g1 -BWeSn+t"D2DXY" -Bx+l"x" -By+l"y" -c
+		gmt colorbar -DJBC+o0c/1.25c+h -Ba0.1+l"@[ \frac{\partial^2 f}{\partial x \partial y} @[ "
+
+		gmt grdimage gauss_ddx_ddy.nc -Ba2f1g1 -BWeSn+t"DDY then DDX" -Bx+l"x" -By+l"y" -c
+		gmt colorbar -DJBC+o0c/1.25c+h -Ba0.1+l"@[ \frac{\partial }{\partial x } \left( \frac{\partial f}{\partial y} \right) @[ "
```
joa-quim, line 37: Please, drop the Latex part. It brings nothing and will cause another failure for me on Windows.

PaulWessel: Why is it failing and why are you not fixing it? Or is this the CI tests you are talking about?

joa-quim: CI test that involve Latex fail for me. TexLive or whatever wanted to install a 6 GB monster. I refused, install what it called some *mini* thing and it fails.

Anyway this test is testing the D2DXY operator so it should test only that. No need to compute and plot the other figures. This only contributes to makes the test set longer and longer to run.

joa-quim (approved):
Smaller it is lower is the maintenance.

test/grdmath/d2dxy.sh
PaulWessel: Works for you now?

PR #7337, opened 2023-03-24
Special case to find the gs executable full name from within Julia artifacts

joa-quim:
In Julia when using the precompiled Artifacts, the Ghost is also shipped with but when gmt_init makes calls to psconvert it doesn't set -G with the path to the Ghostscript_jll executable and those processes fail (mostly modern mode stuff). The solution is to try to read the gs path from ~/.gmt/ghost_jll_path.txt

This code should only be executed by binaries created with Julia's BinaryBuilder.

### DIFF
--- a/cmake/ConfigDefault.cmake
+++ b/cmake/ConfigDefault.cmake
@@ -215,3 +215,8 @@ set (CMAKE_FIND_FRAMEWORK LAST)
 if (NOT DEFINED BUILD_DEVELOPER)
 	set (BUILD_DEVELOPER true)
 endif (NOT DEFINED BUILD_DEVELOPER)
+
+# To allow setting this PreProcessor on command line when BinaryBuilder (Julia) builds GMT
+if (JULIA_GHOST_JLL)
+	add_definitions(/DJULIA_GHOST_JLL)
+endif (JULIA_GHOST_JLL)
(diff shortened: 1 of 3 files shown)


PaulWessel (commented):
I think the add_definition statements all go into cmake/ConfigUsersAdvancedTemplate etc and not in CMakeLists.txt.  That would be very unusual.  Please move it there and update your copy of the template.

joa-quim:
The only other place where the ``add_definition`` statement can go is to ``ConfigDefault.cmake``. The *templates* are just indications and rely on a ``ConfigUser.cmake`` but for builds that don't use a ``ConfigUser.cmake`` like GMT Julia artifact
https://github.com/boriskaus/Yggdrasil-1/blob/4c4171e0c7df698870ffe00403faa6b7c383c97b/G/GMT/build_tarballs.jl
we need something that can be specified on the ``cmake ... `` command line.

Not sure that ``ConfigDefault.cmake`` is particularly more appropriate for this default (non)-setting than ``CMakeLists.txt`` but don't care either.

PaulWessel:
Put it in the Default file then

PaulWessel (approved):

joa-quim:
OK, but let's not merge this until it proves useful. Creating and testing those *Artifacts* is far from trivial.

PR #7349, opened 2023-03-29
Remove the restriction in -p that Elevation must be in 0-90 range

joa-quim:
This allows for example to do mirror maps like this. Limiting to positive elevations only lets do rotations, nor mirroring.

```
gmt pscoast -W0.3 -R-180/0/-90/90 -JV0/8c -Bg -P -K > lixo.ps
gmt pscoast -W0.3 -R-180/0/-90/90 -JV0/8c -Bg -p0/-90 -X8 -O >> lixo.ps
```

lixo

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -2472,10 +2472,6 @@ GMT_LOCAL int gmtinit_parse_p_option (struct GMT_CTRL *GMT, char *item) {
 		return GMT_PARSE_ERROR;
 	}
 	if (k == 1) { GMT->common.p.do_z_rotation = true; el = 90.0;}
-	if (el <= 0.0 || el > 90.0) {
-		GMT_Report (GMT->parent, GMT_MSG_ERROR, "Option -p: Elevation must be in 0-90 range\n");
-		return GMT_PARSE_ERROR;
-	}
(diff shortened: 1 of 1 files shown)


seisman:
Quick comment: 

1. Perhaps we should check if elevation is in the -90 to 90 range?
2. ALso need to update the documentation because I think there are many places that we should elevation must be in 0 to 90.

seisman:
> I don't think we should impose any restriction in the elevation angle. It would only make sense if we were allowing something that does not work. But that is apparently not the case.

I'm not sure about this. It seems "elevation angle" is defined to be in the 0-90 range.
 
> Regarding the docs, my first thought was "sure, I forgot them* but then I can't find any reference to a place were we say that _elev_ must be in the [0 90] interval. Do you know any?

My search also finds nothing. So the [0 90] is never mentioned in the documentation.

joa-quim:
> I'm not sure about this. It seems "elevation angle" is defined to be in the 0-90 range.

Sorry, don't get this. Now you say it should stay above horizon because of the meaning of the word "elevation"?
Why should we restrict the range of *elevation* when that prevents us from doing things like mirroring? In name of what?

PaulWessel (approved):
Seems harmless

PR #7358, opened 2023-04-04
WIP Allow -E+n for modelnorm output

PaulWessel:
See forum [post](https://forum.generic-mapping-tools.org/t/access-greensplines-model-coefficients/3804/10) for the origin of this PR.

Needs much work still - such as matching what the formum post seems to do?

### DIFF
--- a/src/greenspline.c
+++ b/src/greenspline.c
@@ -82,8 +82,9 @@ struct GREENSPLINE_CTRL {
 		bool active;
 		char *information;
 	} D;
-	struct GREENSPLINE_E {	/* -E[<file>] */
+	struct GREENSPLINE_E {	/* -E[<file>][+n] */
 		bool active;
+		bool norm;
 		unsigned int mode;
 		char *file;
 	} E;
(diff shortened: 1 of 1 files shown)


PaulWessel:
Can any of @GenericMappingTools/core see whaT I am missing repeatedly?

wsja:
I have successfully extracted the model variance using this branch. In order to plot an L-curve to determine the optimal number of singular values to use, I computed the data variance from the output grids as (python syntax):

```
datnorm = sqrt(sum(((pre - obs) / uncert) ** 2))
```
with predicted data values `pre`, observed data values `obs` and 1-sigma uncertainties (as supplied in the 4th column via  `-W`) `uncert`.

It would be convenient to also output the data norm as an additional column into *misfitfile*.

During the course of my work, it turned out that the L-curve is not as easy to interpret as I wished to, so I ended up computing Akaikes Information Criterion (AIC)

```
aic = 2*k - 2*ln(L)
```

with number of singular values used `k` and the model likelihood `L` (of `k`). I assumed Gaussian uncertainties and computed `L` as:

```
L = product( exp(-(obs - pred)**2 /(2 * uncert)**2) / sqrt(2 * pi * uncert**2) )
```

The extra steps needed (apart from calculating the model norm `modnorm`) can all be done manually outside GMT. However, writing out all the grids and extracting `pred` can be quite time consuming, so it would be really nice if these metrics could be provided in dry-run mode, that is in convert with `-C+n`. I am unsure, however, if `-C+n` currently does allow to compute `pred`.

One glitch in the current implementation: `modnorm` should be `sqrt(modnorm)`.

wsja:
I could try getting started, but an important prerequisite would be to get the file output working. Until now, `modnorm` does not end up in *misfitfile*, but only `0`s are written to the last column.

PaulWessel:
OK, I have a fixed a bunch of issues:

1. modnorm is now sort (mod norm) and is written properly in the correct column.
2. The column headers are now written to output file

PaulWessel:
I also merged in master so the branch is up to date with master plus the specific changes above.

PR #7362, opened 2023-04-07
Add safety valve in iteration loop

PaulWessel:
Seems like latest Xcode/clang has slightly different roundoff that suddenly (after upgrading Xcode yesterday)  became a hung loop. Since the code has not changed in a year a compiler "improvement" is my only explanation.  Nevertheless, the code fix is plain sanity check anyway and for whatever reason it has never caused a problem with ending the iterations.


PaulWessel:
Note: Many new failures for me on macOS with new Xcode that all seams related to round-off treatments.  @GenericMappingTools/core may wish to run tests before/after updating Xcode.

PR #7363, opened 2023-04-07
WIP Update admin/build-release.sh for 6.5

PaulWessel:
Given the macport's GDAL seems perpetually f**ed  I am trying to see if we can build the macOS installer with homebrew.  It goes a ways but then I get this rpath error from

```
 error:
  /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool-classic:
  can't open file: @rpath/libIlmThread-3_1.30.dylib (No such file or
  directory)

```
The file do exist but in /opt/homebrew/lib

However, homebrew is not always installed there but in /usr/local/bin?

Hoping @seisman and @maxrjones might have some insight here since we've waited several months for macports to get their act together.

### DIFF
--- a/admin/build-release.sh
+++ b/admin/build-release.sh
@@ -51,15 +51,15 @@ elif [ "X${1}" = "X-t" ]; then
 	release=0
 elif [ $# -gt 0 ]; then
 	cat <<- EOF  >&2
-	Usage: build-release.sh [-p|m]
+	Usage: build-release.sh [-p|m|t]
 	
 	build-release.sh must be run from top-level gmt directory.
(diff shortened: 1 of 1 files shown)


seisman:
> However, homebrew is not always installed there but in /usr/local/bin?

Homebrew is installed in `/opt/homebrew` for Apple M1 and in `/usr/local` for Apple Intel chips.

PaulWessel:
BTW, I tried 

`export DYLD_LIBRARY_PATH=/opt/homebrew/lib`

but same `@rpath` error.

PR #7364, opened 2023-04-07
Fix doc search for docs built with sphinx 6.x

### LINKED ISSUE #5039: Documentation search not working in dev branch
Not sure if correct place for this but when using the search box in the documentation I get no results when searching in "dev" branch.

"latest" works:
https://docs.generic-mapping-tools.org/latest/search.html?q=plot&check_keywords=yes&area=default#

"dev" doesn't work:
https://docs.generic-mapping-tools.org/dev/search.html?q=plot&check_keywords=yes&area=default#

seisman:

The doc search is broken due to incompatible changes in Sphinx 6.x (see https://www.sphinx-doc.org/en/master/changes.html#release-6-0-0-released-dec-29-2022 for the full changelog and the official solution).

Fixes #5039

**Reminders**


PaulWessel (approved):
blind review

PR #7369, opened 2023-04-10
Add function list to gmt_api.c comments

PaulWessel:
Also, move 1 function to static and test if the four `api_get_record_*` functions still work via the pointers if they are declared static to this file.  Works fine on macOS.  Suggest if nothing strange happens on Windows (@joa-quim) or Linux (@Esteban82, @anbj, @seisman) then we can be sure this works.

Also fixed two odd places where we introduced new variables (e.g., int n) _after_ we already have had statements.


joa-quim:
It's fine on Windows

joa-quim (approved):

Esteban82 (approved):

Esteban82:
I am not sure how to test it. This code worked on linux. 

`gmt basemap -R0/1/0/1 -Bafg -png test`

Let me know if there is another test.

PR #7379, opened 2023-04-12
Option to use build-release no gdal on the macports

PaulWessel:
This PR adds a check to determine if we do not have **GDAL** installed in **macports** (since they screwed that port up) and instead get that library, include files, and executables from a **homebrew** installation.  This seems to work well except of course the new problem of _otool_ not finding a library:

```
 error:
  /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool-classic:
  can't open file: @rpath/libIlmThread-3_1.30.dylib (No such file or
  directory)

```
Since our script always worked before and now fails with this error I can only assume that there is something stupid somewhere else that is causing this @rpath confusion.  Ideas are welcome, but this PR should be straightforward.

### DIFF
--- a/admin/build-macos-external-list.sh
+++ b/admin/build-macos-external-list.sh
@@ -21,7 +21,7 @@ GDAL_VIA_HOMEBREW=$3
 
 if [ ${DISTRO} = "MacPorts" ]; then
 	top=/opt/local
-	if [ ${GDAL_VIA_HOMEBREW} = "Y" ]; then
+	if [ "X${GDAL_VIA_HOMEBREW}" = "XY" ]; then
 		topgdal=/opt/homebrew
 	else
 		topgdal=/opt/local
@@ -60,7 +60,7 @@ EXELINKS=${top}/bin/gs
 EXEONLY=
 # 1d. Shared directories to be added (except ghostscript which we do separately)
 #     Use full path if you need something not in your path
-if [ ${GDAL_VIA_HOMEBREW} = "Y" ]; then
+if [ "X${GDAL_VIA_HOMEBREW}" = "XY" ]; then
 	EXESHARED="${proj}"
 else
 	EXESHARED="gdal ${proj}"
@@ -114,7 +114,7 @@ if [ ! "X$EXESHARED" = "X" ]; then
 	echo ""
 	echo "install (DIRECTORY"
 fi
-if [ ${GDAL_VIA_HOMEBREW} = "Y" ]; then
+if [ "X${GDAL_VIA_HOMEBREW}" = "XY" ]; then
 	echo "	$topgdal/share/gdal"
 fi
 for P in $EXESHARED; do
--- a/admin/build-macos-external-list.sh
+++ b/admin/build-macos-external-list.sh
@@ -12,12 +12,20 @@
 #
 # Notes:
 #   1. This is tested on macports where gs is a symbolic link to gsc.
+#	2. Since the latest macports is fucked related to GDAL we must get GDAL and executables
+#	   from homebrew installation instead.
 
 GSVERSION=$1	# Get the version of gs from build-release.sh
 DISTRO=$2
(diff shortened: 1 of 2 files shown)


PaulWessel:
Some googling revealed that the missing library is something graphics related developed by Industrial Light and Magic (George Lucas), hence the **Ilm** tag.  Nothing we have relies on it as far as I can tell, but perhaps **ffmpeg**? I noticed earlier that the library is present in /opt/homebrew but was not in macports until I installed it explicitly.  In either case the library is not found due to the `@rpath `issue above.

seisman:
Can you figure out which package relies on `@rpath/libIlmThread-3_1.30.dylib` by 

> This PR adds a check to determine if we do not have **GDAL** installed in **macports** (since they screwed that port up) and instead get that library, include files, and executables from a **homebrew** installation. This seems to work well except of course the new problem of _otool_ not finding a library:

Both GMT and GDAL rely on netCDF. In this case, GMT may be linked to the netCDF library from MacPorts and GDAL from Homebrew is linked to the Homebrew's netCDF library. Will it make a mess?

joa-quim:
> Both GMT and GDAL rely on netCDF. In this case, GMT may be linked to the netCDF library from MacPorts and GDAL from Homebrew is linked to the Homebrew's netCDF library. Will it make a mess?

Potentially yes. netCDF depends on HDF5 and if they were not build with the **exactly** same HDF5 version, shit will follow.

seisman:
@PaulWessel Can you figure out which package rely on `libIlmThread-3_1.30.dylib` by adding more debug message to your `otoolr.c` code?

PaulWessel:
No package depends on the ILM Threads library.  I agree with the HDF5/netcdf/GDAL issue - my experiment was mostly to see if the ILM was something that only happen when using homebrew, but given it is the same error for this hybrid macports setup there is no need to involve macports provided we can fix the ILM library issues that originates in the Xcode command above.

PaulWessel:
Sorry, take that back.  ligdal depends on it via another library:

gdal:

	`/opt/homebrew/opt/openexr/lib/libOpenEXRUtil-3_1.30.dylib (compatibility version 30.0.0, current version 30.6.1)`

and that one lists this:

	`@rpath/libIlmThread-3_1.30.dylib (compatibility version 30.0.0, current version 30.6.1)`

So because **GDAL** depends on the _openexr_ library and it has lovely `rpath` statements we get screwed?

PaulWessel:
I dont understand why some entries have full paths and others just have `@rpath `in the libraries from homebrew:

```
/opt/homebrew/opt/openexr/lib/libOpenEXRUtil-3_1.30.dylib:
	/opt/homebrew/opt/openexr/lib/libOpenEXRUtil-3_1.30.dylib (compatibility version 30.0.0, current version 30.6.1)
	@rpath/libOpenEXR-3_1.30.dylib (compatibility version 30.0.0, current version 30.6.1)
	@rpath/libOpenEXRCore-3_1.30.dylib (compatibility version 30.0.0, current version 30.6.1)
	/opt/homebrew/opt/imath/lib/libImath-3_1.30.dylib (compatibility version 30.0.0, current version 30.0.1)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)
	@rpath/libIlmThread-3_1.30.dylib (compatibility version 30.0.0, current version 30.6.1)
	@rpath/libIex-3_1.30.dylib (compatibility version 30.0.0, current version 30.6.1)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.36.0)

```

joa-quim:
I think the @rpath results from the use of a compiling/linking option and different packages do it differently.

joa-quim (approved):

PR #7386, opened 2023-04-14
Add gdalread func list

PaulWessel:
Also rename static local functions to follow GMT standard naming convention.

### DIFF
--- a/src/gmt_gdalread.c
+++ b/src/gmt_gdalread.c
@@ -21,17 +21,19 @@
  * Author:	Joaquim Luis
  * Date:	27-Aug-2009
  * Revision: 1		Based on gdalread.c MEX from Mirone
- *			The populate_metadata() was modified from mexgdal of John Evans (johnevans@acm.org)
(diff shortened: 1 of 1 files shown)


joa-quim (commented):
They are local funs, what's the interest of making their names longer?

joa-quim:
In my view this only makes the code a bit more difficult read (longer lines that wrap or 'outflow' on the right) with no gain in extra information.

PaulWessel:
Well, what is the point of having a convention and then we start making exceptions here and there.  This is to help future GMT developers better navigate things.

PaulWessel:
If you wish to shorten some of those names, feel free to do so as long as they start as indicated.

joa-quim (approved):
I am not convinced that this convention helps code navigation. Static local files are only seen when reading the code, and than the GMT_LOCAL prefix already tells reader about their nature without making the fun names longer.

PaulWessel:
Well, true but I am not going back to removing gmtfile_ from all those functions everywhere,  So a bit of inconvenience you will have to endure.

PR #7413, opened 2023-04-25
Add -Io to grdinfo for oblique format

PaulWessel:
See this forum [post](https://forum.generic-mapping-tools.org/t/grdimage-of-projected-grid-fails/3882) for background.  There are times we need **grdinfo** to report  the grid's region in the _lower left upper right_ **+r** syntax.  This PR adds **-Io** to format the region that way.

### DIFF
--- a/doc/rst/source/grdinfo.rst
+++ b/doc/rst/source/grdinfo.rst
@@ -17,7 +17,7 @@ Synopsis
 [ |-E|\ [**x**\|\ **y**][**+l**\|\ **L**\|\ **u**\|\ **U**] ]
 [ |-F| ]
 [ |-G| ]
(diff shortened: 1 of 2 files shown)


anbj (approved):
Looks good:
```
$ gmt grdcut -RNO -Gno.nc @earth_relief_10m
gmt [WARNING]: Remote dataset given to a data processing module but no registration was specified - default to gridline registration (if available)

$ gmt grdinfo -Io no.nc
-R4.5/57.8333333333/31.1666666667/71.3333333333+r
```

Q: It is not possible to use ``[dx[/dy]`` with the modifiers (e.g. ``-Io``)?

PaulWessel:
No.  Same with -Ir.  They are just there to give the exact.

PR #7417, opened 2023-04-25
Fix example discussion

PaulWessel:
The text said the postflight script was not even a GMT script but that changed over time so now it is.  Also, let the 2nd example use quoted 'EOF' tags to show contrast with 1st example.


anbj (approved):
Looks good to me.

PR #7423, opened 2023-04-27
More robust checking for single L|U|B flags in CPTs

PaulWessel:
See this form [post](https://forum.generic-mapping-tools.org/t/issue-creating-cpt/3890) for background.  The problem was that the code which checked for a trailing L|U|B flag did not consider that we might be reading a hex color string which might actually end in **B**:

```
-1000 #5CC3EB -500 #5CC3EB
...
500 #99762B 600 #99762B
```

and thus be flagged as as bottom annotation setting.  This PR checks that the previous character is white space before interpreting any L|U|B flags. **Note:** The L|U|B logic was implemented a long time before we started to consider hex colors so we never anticipated letters in those positions.

While at it I noticed that in gmt_support.c _gmtlib_read_cpt_ function the annotation flags were still 1-3 while elsewhere we had switched to named constants such as **GMT_CPT_B_ANNOT**, so I updated those assignments for better readability.


anbj (approved):
Now gives the expected cpt. Great. Thank you!

PR #7426, opened 2023-04-27
Update docs for "G" of "gmtwhich" regarding remote datasets

yvonnefroehlich:

GMT now supports several remote datasets, not only `@earth_relief_xxy`. This PR aims to update the documentation for **G** of `gmtwhich`. This was first mentioned by @seisman in https://github.com/GenericMappingTools/pygmt/pull/2519#discussion_r1178599444.

**Reminders**

### DIFF
--- a/doc/rst/source/gmtwhich.rst
+++ b/doc/rst/source/gmtwhich.rst
@@ -62,9 +62,9 @@ Optional Arguments
 .. _-G:
 
 **-G**\ [**a**\|\ **c**\|\ **l**\|\ **u**]
-    If a file argument is a downloadable file (either a complete URL or a @file for
-    downloading from the GMT data server, i.e., one of the sample datasets or one of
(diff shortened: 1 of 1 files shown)


joa-quim (commented):
I don't think it's good to drop reference to the example ``@earth_relief_xxy`` name. Sure we can extend info with "or any other of the sample datasets at https://www.generic-mapping-tools.org/remote-datasets/", but not drop the *iconic* @earth_relief...

yvonnefroehlich:
> I don't think it's good to drop reference to the example `@earth_relief_xxy` name. Sure we can extend info with "or any other of the sample datasets at https://www.generic-mapping-tools.org/remote-datasets/", but not drop the _iconic_ @earth_relief...

I updated the formulation in commit https://github.com/GenericMappingTools/gmt/pull/7426/commits/a5030895f3f5fa1048bb84060860545d010337d9.

joa-quim (approved):
Thanks

PR #7457, opened 2023-05-11
Pause 1 sec before viewing master frame

PaulWessel:
We do something similar before assembling frmes to video since we cannot be 100% that the image(s) have been finished written and ready to be used.  Before this PR we would sometimes (but not aways) get messages like

`mv: rename 1.png to /Users/pwessel/1.png: No such file or directory`

This waits a second before trying to access the master frame.

Let me know if this works on Linux as wells it does on macOS


Esteban82:
It is not working for me.

PaulWessel:
Maybe it happens when the master ping is there from a previous run? Please delete and try ahain

Esteban82:
It is trying to open the file in a sub-directory (`title/master/`). Not in the main directory.

image

PaulWessel:
Closing this since the 1-sec wait is no longer needed when we use gmt docs to open files.

PR #7458, opened 2023-05-11
Add modifier +hheight to set logo height instead of width

PaulWessel:
Since the logo aspect ratio is fixed, one can only specify one of the dimensions.  This PR adds **+h** and makes it clear in the synopsis and usage that both are optional [Default is 144p wide logo].


Esteban82 (approved):

Esteban82:
I can't merge it. There are two checks errors.

PR #7461, opened 2023-05-14
Let movie -A handle addition of audio track

PaulWessel:
Thie enhancement allows users to add a narration (or any sound) track to their animation.  Via **+e** the length of the audio can be stretched to exactly fit the animation length.

@Esteban82, this PR compiles but I have not tested any of it.  If you have time then please do.  The key coding is in movie.c starting around line 2620.

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -466,6 +466,18 @@ able to verify the latter by viewing a transparent webm movie in Chrome. Animate
 transparent PNGs as well and here each additional frame accumulate in the final movie. Experts may create
 transparent PNGs and create movies in professional tools that support a movie alpha channel.
 
+. _Audio:
+
+Adding an Audio Track
+---------------------
+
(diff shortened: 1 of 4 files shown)


Esteban82:
Ok, great @PaulWessel. I will test it. 

Will we need to make a test for gmt/test/movie also?

PaulWessel:
Hard to test movies. Maybe eventually we only create master frame pdf and compare with original

anbj:
I tried adding an mp3-file to an mp4 movie. This resulted in the movie containing only the audio, and not any video.

(Sorry  - couldn't wait testing)

PaulWessel:
I just updated movie.c and also the Indiana Jones track with audio and it worked brilliantly.

Esteban82:
I inverse the ratio formula. I test it with the Indiana Jones script. When I reduce length of movie (`animation_duration` ), the audio was strecht. Now works well with my commit.

Esteban82 (approved):

PaulWessel:
Thanks, good catch.

PR #7468, opened 2023-05-20
Add option to show the build date

joa-quim:
We show the build date for commit builds but not for release. New option ``gmt --show-date`` shows building date for every build.


PaulWessel (approved):
Works on macOS as well. Requesting 2 changes:

1. --show-date         Show GMT binary building date [there is only one binary, no?]
2. Add the above to the gmt.rst documentation
3.

joa-quim:
> [there is only one binary, no?]

Well, an exe and two dlls.

Issue #7453, opened 2023-05-08
One of the grdcut bugs


joa-quim:
The bug shown in
```
grdcut @earth_relief_03s_g -R-10/-9.8/4.9/5.0 -Glixo.grd
```

Was introduced in the commit

4/7/2023
SHA-1: bf4918172f8f627963180898334740887e67f42a

* allow slop in wesn (#7356)

![image](https://user-images.githubusercontent.com/537321/236909912-2eda40e2-6078-4682-b1ad-013ba5ad4ec1.png)

joa-quim:
The other bug is (it crashes).
```
gmt grdcut @earth_relief_03s_g -R-11/-9/4/6.0 -Glixo.grd
grdblend [ERROR]: Failed to remove c:\TMP/grdblend_resampled_a28116.nc! [remove error: Permission denied]
grdblend [ERROR]: Failed to delete file c:\TMP/grdblend_resampled_a28116.nc
```

joa-quim:
Have to leave right now, but the permission to remove is not the problem . That happens because the struct has twice the same file name and the error message come from the the first time it tries to remove it but the it's still open in grdcut.

seisman:
Maybe it's also related to https://github.com/GenericMappingTools/pygmt/issues/2511?

joa-quim:
I don't get the second error on WSL either.

joa-quim:
And there is also #7012

PaulWessel:
First bug issue. So **grdcut** calls **grdblend** to make a blend of 15s pixel and 3s gridline. The z-range of the entire 1x1 degree 3s tile is -1 to 33 m and inside the selected area it is all zeros.  The 15s grid in this area has z = -2048.5 to -940.  So, what happens is that **grdblend** runs **grdsample** on the 15s subset to bring it onto the 3s nodes, and then we blend, which of course is adding 0 to the 15s value and taking the average so you think it would be close to -1000.  Yet the lixo grid values are about or matches matches the 3s grid.

I attach a script that shows a bunch of plots and has those two peaks in the corners.

```
#!/bin/bash
# Cut&blend the 3s and 15s subsets
gmt grdcut @earth_relief_03s_g -R-10/-9.8/4.9/5.0 -Glixo03s.grd
gmt grdcut @earth_relief_15s_p -R-10/-9.8/4.9/5.0 -Glixo15s.grd
# Cut the raw grids
[...]

PR #7469, opened 2023-05-21
Fix wrong noise direction in grdsample wesn checking

PaulWessel:
This relates to #7453 but does not fix it.  When **grdsample** was changed by #7356 we got the sign of the error slop (_ince_) wrong for the x-direction.  I addition to fixing that, I have also broken some long lines into 2-lines so easier to do debug.  With this fix the original problem fixed by #7356 is still fixed but now the **grdcut** command yields the correct sub-region.  However, the spikes in the corners are still there and are there for different reasons, as discussed below.  I recommend we approve and merge this PR though.

The commented out verbose statements in **grdsample** would print this before the April 7 "fix":
```
gmt grdcut @earth_relief_03s_g -R-10/-9.8/4.9/5.0 -Glixo03s.grd
wesn_i = -10/-9.8/4.89583/5
wesn_o = -10/-9.8/4.9/5

```
which is incorrect for the south limit in _wesn_i_ (but fortuitously gives a better plot).  They should be the same as _wesn_o_.  After this PR we get the correct

```
wesn_i = -10/-9.8/4.9/5
wesn_o = -10/-9.8/4.9/5

```
I am pretty sure the spikes at the WS and NW corners have the same source.  Remember, we are resampling the 15s grid (pixel) at the 3s grid (gridline). Here are items that conspire:

1. The 3s tile ends at west = -10 and hence the boundary rows/cols are produced by the boundary conditions, not neighbour tile data.
2. We are extrapolating the 15s grid beyond the last nodes in the corners and hence things are driven almost exclusively by the boundary conditions.  Just like for the 3s grid on even degrees, the 15s data is tiled at 10x10 degrees and hence at -10 we have the same situation: padding completely determined by BC and not neighbor tile data.

Here is a zoom in on the SW corner and where we wish to estimate 15s (red dots) from the blue dots.

zoom

Thus, the fix to this issue may be complicated, but perhaps something like this:

- When processing the remote data set tiles, extend exact tile by 2 rows/cols.
- When reading/processing these tiles, GMT knows that it should:
- a) Read the tile with no pad requested (so we read it as is including the extra nodes)
- b) Change the region inwards by 2 steps and set padding to 2.  Now we have data in the internal grids boundary pad
- c) The interpolation would now not have crazy artefacts of PCs imposed (natural BC) since the underlying data is continuous and now only data are used as BC nodes.

This obviously means a wholesale revision of the remote data set structures, how _gmt_remote.c_ operators, etc. but probably the best solution. In fact, we should have been smart enough to think of this when the global tiling was first proposed and introduced.

Other solutions seems much messier, e.g., to somehow get the neighbouring tile data into the boundary pads, i.e., read in strips of the neighbor tiles and insert into pad.  I would hate to be the coder for that.

FYI, if we just decide to cut out the 15s subset (the raw grid) and then use **grdsample** to sample it at the 3s nodes (i.e., no blending at all, just 15s resampled), you can still see the four corners having "lips":

```
gmt grdcut ~/.gmt/server/earth/earth_relief/earth_relief_15s_p/N00W010.earth_relief_15s_p.nc -R-10/-9.8/4.9/5.0 -Glixo15s_raw.grd
gmt grdsample lixo15s_raw.grd -T -I3s -Gtmp.grd
gmt grdview tmp.grd -R-10/-9.75/4.85/5/-2200/-900 -Bxaf -Byaf -Bzaf -BWSZ+t"Blend" -JZ2i -p135/35 -png p

```
p

So even without any **grdblend** calls happening you can see the 2-D BCR extrapolation at the corners lack real BC constraints.  If we just plot the 15s as is: No lips:

map15s

So the spikes are all produced by teh BCR extrapolation into the voids.

### DIFF
--- a/src/grdsample.c
+++ b/src/grdsample.c
@@ -264,11 +264,15 @@ EXTERN_MSC int GMT_grdsample (void *V_API, int mode, void *args) {
 		/* Adjust wesn used to READ subset unless 360 geographic */
 		if (!wrap_360_i) {
 			k = 0;
-			while (wesn_i[YLO] < (wesn_o[YLO] + ince[GMT_Y])) wesn_i[YLO] += Gin->header->inc[GMT_Y], k++;	/* Now on or inside grid boundary */
-			if (wesn_i[YLO] > (Gin->header->wesn[YLO] + ince[GMT_Y]) && k) wesn_i[YLO] -= Gin->header->inc[GMT_Y];	/* Now exactly on boundary or just outside but still inside input grid south boundary */
+			while (wesn_i[YLO] < (wesn_o[YLO] + ince[GMT_Y]))
+				wesn_i[YLO] += Gin->header->inc[GMT_Y], k++;	/* Now on or inside grid boundary */
+			if (wesn_i[YLO] > (Gin->header->wesn[YLO] + ince[GMT_Y]) && k)
+				wesn_i[YLO] -= Gin->header->inc[GMT_Y];	/* Now exactly on boundary or just outside but still inside input grid south boundary */
 			k = 0;
-			while (wesn_i[YHI] > (wesn_o[YHI] - ince[GMT_Y])) wesn_i[YHI] -= Gin->header->inc[GMT_Y], k++;	/* Now on or inside boundary */
-			if (wesn_i[YHI] < (Gin->header->wesn[YHI] + ince[GMT_Y]) && k) wesn_i[YHI] += Gin->header->inc[GMT_Y];	/* Now exactly on boundary or just outside but still inside input grid north boundary */
+			while (wesn_i[YHI] > (wesn_o[YHI] - ince[GMT_Y]))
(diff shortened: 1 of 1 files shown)


PaulWessel:
If you extend the west boundary to further west, e.g. -10:02, then the surface will show a visible “seam” along Lon = 10, and ending in the two spike lips.  While I think this is BCs, we should also examine the BCR for outside corner cases as here

Esteban82 (approved):
Blind approval. Let me know if I do you want to test it.

PaulWessel:
> Blind approval. Let me know if I do you want to test it.

You could try to play with view points etc on something like this

`gmt grdview @earth_relief_03s_g -R-10.1/-9.9/4.9/5.0/-2200/-900  -Bxaf -Byaf -Bzaf  -JZ2i -p135/35  -pdf t
`

which I think shows the seam + the lips

PaulWessel:
But no merging yet.

joa-quim (approved):

PR #7488, opened 2023-06-01
Add -Fe for step-up sampling in sample1d

PaulWessel:
See the forum [post](https://forum.generic-mapping-tools.org/t/interpolate-data-values-with-dates-vlookup/3985) for details.  This interpolator keeps the previous value until we pass a new input node and then we step up to that value.

@Esteban82, I think this reproduces your example but I do not know if it is generally correct since I only tested it on yours.

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -1849,7 +1849,7 @@ GMT_LOCAL int gmtsupport_intpol_sub (struct GMT_CTRL *GMT, double *x, double *y,
 			case GMT_SPLINE_NN+GMT_SPLINE_CURVATURE:	/* Nearest neighbor curvatures are zero  */
 				v[i] = 0.0;
 				break;
-			case GMT_SPLINE_STEO+GMT_SPLINE_CURVATURE:	/* Step curvatures are zero  */
+			case GMT_SPLINE_STEP+GMT_SPLINE_CURVATURE:	/* Step curvatures are zero  */
 				v[i] = 0.0;
 				break;
 			case GMT_SPLINE_SMOOTH:	/* Smoothing spline */
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -1358,8 +1358,8 @@ GMT_LOCAL bool gmtsupport_is_penstyle (char *word) {
 	if (n == 0) return (false);
 
 	if (!strncmp (word, "dotdash", 7U) || !strncmp (word, "dashdot", 7U) \
-		|| !strncmp (word, "dash", 4U) || !strncmp (word, "dot", 3U) \
-		|| !strncmp (word, "solid", 5U)) return (true);
+        || !strncmp (word, "dash", 4U) || !strncmp (word, "dot", 3U) \
+        || !strncmp (word, "solid", 5U)) return (true);
 
 	n--;
 	if (strchr (GMT_DIM_UNITS, word[n])) n--;	/* Reduce length by 1; the unit character */
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -1813,10 +1813,10 @@ GMT_LOCAL int gmtsupport_intpol_sub (struct GMT_CTRL *GMT, double *x, double *y,
 				v[i] = ((u[i] - x[j]) < (x[j+1] - u[i])) ? y[j] : y[j+1];
 				break;
 			case GMT_SPLINE_STEP: /* Step up to next value when passing it */
-				if (j == 0)	/* Start at the beginning */
+                if (i == (m-1))    /* End at the end */
+                    v[i] = y[n-1];
+				else if (j == 0)	/* Start at the beginning */
 					v[i] = y[0];
-				else if (i == (m-1))	/* End at the end */
-					v[i] = y[n-1];
 				else	/* Pick the smallest before the step happens */
 					v[i] = MIN (y[j], y[j+1]);
 				break;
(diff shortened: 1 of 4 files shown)


src/gmt_support.c
```diff
@@ -1358,8 +1358,8 @@ GMT_LOCAL bool gmtsupport_is_penstyle (char *word) {
 	if (n == 0) return (false);
 
 	if (!strncmp (word, "dotdash", 7U) || !strncmp (word, "dashdot", 7U) \
-	    || !strncmp (word, "dash", 4U) || !strncmp (word, "dot", 3U) \
-	    || !strncmp (word, "solid", 5U)) return (true);
+		|| !strncmp (word, "dash", 4U) || !strncmp (word, "dot", 3U) \
```
joa-quim, line 1361: Don't replace those spaces by tabs. Using the spaces like that is the only way of keep the alignment with the first character after the parentesis.

PaulWessel: Sorry, it was an auto think.  I will fix it.

src/gmt_support.c
```diff
@@ -10323,7 +10338,7 @@ GMT_LOCAL unsigned int gmtsupport_contour_old_T_parser (struct GMT_CTRL *GMT, ch
 		}
 		else {
 			GMT_Report (GMT->parent, GMT_MSG_ERROR,
-			            "Option -T: Give low and high labels either as :LH or :<low>,<high>.\n");
+						"Option -T: Give low and high labels either as :LH or :<low>,<high>.\n");
```
joa-quim, line 10341: Likewise here and below.

PaulWessel: I did a few more but dont want to go over all of this manually.  As you know I am not in favour of 80character screens so feel these break-the-line adds longer code and more opportunities for tab/space shits.

joa-quim: It's has nothing to do about the the line width but only on a good alignment with above line. Tee price of indenting with tabs is that we must do a mix of tabs and spaces to keep that alignment.

Esteban82:
Great Paul!! I test it. It works. But I found a bug when the input has two records

```
cat << EOF | gmt sample1d xy.txt -I1 -Fe
1	1
5   4
EOF
Output:
1       1
2       1
3       1
4       1
5       1
```

PaulWessel:
OK, I was checking for the first point first but need to do the end first, then first, then the rest.  Now works, I think?

src/gmt_support.c
PaulWessel: If you want to revert some of these then no problem, go right ahead.

Esteban82 (approved):

Esteban82:
I found a new issue. 

```
cat << EOF | gmt sample1d -I2 -Fe -Am
1	1
3   3
7   4
EOF
output:
2       1
4       3
6       4   # here I would expect 3.
```

Esteban82:
With the GMT_Splines.sh also found some issues. Maybe the sames (see #7490).
Here is the figure. 

GMT_splines

```
gmt sample1d t.txt -Fe -T0.5
0       0
0.5     0
1       1
1.5     1
2       1.25
2.5     1.25
3       1.25
3.5     1.25
4       1.5
4.5     2
5       2
5.5     2
6       2.5
```

PaulWessel:
-Fe requires monotonously increasing data but this data set is not that, so not an appropriate test

Esteban82:
Ok, I see. So, would I need to pre process my data with another tool?

PaulWessel:
Well, it only make sense for cumulative data since we step up

Esteban82:
In the case of the forum, yes, it is cumulative. But it might not be always the case specially with dates.

PaulWessel:
Perhaps you can give me some examples of how -Fe would work on regular data such as in GMT_splines.sh.  I approached the problem as specific to cumulative data, or at least a monotonically increasing data set, but presumably we could do something similar for monotonically _decreasing_ data.  Not sure what expectation is for other data sets though.

Esteban82:
This is a simple animation that I want to make. Look at the labels.

https://github.com/GenericMappingTools/gmt/assets/17908951/851a2b0d-4ff9-4eef-84bf-8648d5c38705

When I interpolate the data (with sample1d -Fe) I got this table. 
There are some issues:
- Why the list doesn't start at 19-01 as in the dataset? It does if the first data is 2004-01-18.
- In the last record says 5 ahead of time (14-Feb) when in the data table it says 16-Feb.


**times.txt:**
```
2004-01-21T00:00:00	1
2004-01-24T00:00:00	1
2004-01-27T00:00:00	1
2004-01-30T00:00:00	2
2004-02-02T00:00:00	2
2004-02-05T00:00:00	2
2004-02-08T00:00:00	2
2004-02-11T00:00:00	2
2004-02-14T00:00:00	5

```

## Full script 

```
#!/usr/bin/env bash
# Data
cat << EOF > Accumulated_Goals.txt
2004-01-19 1
2004-01-28 2
2004-02-16 5
EOF

gmt sample1d Accumulated_Goals.txt -fT --TIME_UNIT=d -I3 -Fe > times.txt

# 	2. Set up main script
cat << EOF > main.sh
gmt begin
	gmt basemap -Rd -JW24c -Bf -Y0 -X0
gmt end
EOF

gmt movie main.sh -Cfhd -Ttimes.txt -NMessi -Ml,png -Vi -Zs -Fmp4 -Lc0+jTR+o0.3/0.3+gwhite+h+r \
	-Lc1+jTL+o0.3/0.3+gwhite+h+r
```

PaulWessel:
I guess 19 is not a multiple of 3, but 21 is.  Remember, you are stepping every 3 and nodes are always a multiple of the increment.

Esteban82:
> I guess 19 is not a multiple of 3, but 21 is. Remember, you are stepping every 3 and nodes are always a multiple of the increment.

Ok, thanks, I didn't know that. Altough with dates it doesn't make much sense to have steps that are multiple of the increment.

PR #7496, opened 2023-06-06
Update gmtmath.rst

Esteban82:
Fix indentation of examples.

BTW, I noticed that some examples have an indentation and others don't. Should we use them or not?

### DIFF
--- a/doc/rst/source/gmtmath.rst
+++ b/doc/rst/source/gmtmath.rst
@@ -565,26 +565,26 @@ Examples
 
 To add two plot dimensions of different units, we can run
 
-   ::
+::
 
     length=$(gmt math -Q 15c 2i SUB =)
 
 To compute the ratio of two plot dimensions of different units, we select *non-dimensional* output and run
 
-   ::
+::
 
     ratio=$(gmt math -Qn 15c 2i DIV =)
 
 To take the square root of the content of the second data column being
 piped through **gmtmath** by process1 and pipe it through a 3rd process, use
 
-   ::
+::
 
     process1 | gmt math STDIN SQRT = | process3
 
 To take log10 of the average of 2 data files, use
 
-   ::
+::
 
     gmt math file1.txt file2.txt ADD 0.5 MUL LOG10 = file3.txt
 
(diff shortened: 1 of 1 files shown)


joa-quim:
No, better not indent the examples. They are highlighted already.

Esteban82:
> No, better not indent the examples. They are highlighted already.

Ok, I just did it.

anbj:
I think I asked Paul the same once about indentation, and he preferred to have them.

anbj:
Yes; see [here](https://github.com/GenericMappingTools/gmt/issues/7226#issuecomment-1418929932) and response below.

Esteban82:
Ok, so I have to undo all the last changes.  Do you agree @joa-quim ?

 How can I undo the last commit?

seisman:
> I still prefer the non-indented because it may save us to break a command line example. 

I also prefer the non-indented version, because the code blocks are already highlighted with a colorful background. 

> But what we should do is whatever is done in the majority of of other examples.

It seems possible to batch convert all indented code blocks to non-indneted code blocks by replacing `  ::` with `::`.

> How can I undo the last commit?

If you really want to undo the last commit, you can use `git revert 230a7290f1a6cf29e12f45a2973b32aef1c81b7d` in a terminal.

Esteban82:
Ok, I am not sure if I want to undo the last commit. I don't have a strong opinion on indetaction.

BTW, I notice that the table of operator don't fit in the screen. How can I remove the identation of the table? By removing the whitespace at the beginnig of each line?

PaulWessel:
So long as I don’t have to do it I am fine with whatever system is decided

joa-quim:
I also don't care much about this so minor thing but if someone wants to make the change my vote, as I said before, goes to no indentation. The highlighting is more than enough.

anbj:
I care. The appearance of the docs is imporant. Even if it’s a minor thing, consistency gives a good impression and indicates a well-taken-care-of project. Gmt deserves it.

PaulWessel:
True. If you are motivated please do it

Esteban82:
> So no indentation, @Esteban82, @joa-quim?

No indentation. 

So, I think that this PR can be approved.

anbj (approved):

Issue #6778, opened 2022-06-08
Improve description of table operators in gmt/grd math modules


Esteban82:
As I have suggested [here](https://github.com/GenericMappingTools/gmt/discussions/4678#discussioncomment-606312) I want to improve the description of the [math operators](https://docs.generic-mapping-tools.org/dev/gmtmath.html#operators).  I will work on this. It can wait until gmt 6.5.

I open this issue so you know I am working on it, and more important to ask some doubts. Any suggestions is welcome. 

I am planning to:

1. Improve the description of the `Return` column 
2. Add a link to [DLMF]( https://dlmf.nist.gov/) for complex functions.

Esteban82:
I am thinking of adding a new column with the type of formula (like trigonometric, logic, basic math, statistical, etc). I think it would help the users. 

Is there any way to add a filter in the table (so I could ask to only show the trigonometric formulas) or to sort it?

seisman:
> Is there any way to add a filter in the table (so I could ask to only show the trigonometric formulas) or to sort it?

Sphinx doesn't support it, but it's possible to enable filtering/sorting with [this extension](https://sphinxcontrib-needs.readthedocs.io/en/latest/directives/needtable.html).

Esteban82:
I create a table in [google docs](https://docs.google.com/spreadsheets/d/1kLIWXPxMsqNU_1WB9fu9zT30ayL8VPS-NwFDwfEkQ9E/edit?usp=sharing) with the info of the table.

There I add a new column with the `Type of Function`. For some cells (highligthed in color), I am not sure what to put (e.g. Psi (or Digamma) of A). I also add a more detailed version for the `return` column to replace the orignal. There is also a column with wikepedia and https://dlmf.nist.gov/idx/ links but that is WIP.

Let me know if you prefer that I post my doubts directly here.

PaulWessel:
Nice start, @Esteban82.  I have some comments:

1. I would lump probability distributions, statistical, combinatorics, **LMSSCL** etc etc into a single category: Probability
2. All the crazy math functions (Bessel, Kelvin, Dilog, Legendre etc) I would place in category: Special Functions (that is how they are often referred to in math books, for instance).
3. I think some of the **BIT** operators should go into Logic (**AND**, **NOT**, **OR**, **TEST**, **XOR**) and the rest (**LEFT**, **RIGHT**) goes into Arithmetic. Computer Programming is too specific.
[...]

PR #7500, opened 2023-06-07
Update grdmath.rst table of operators

Esteban82:
I update the table of operators. Details in #6778.


PaulWessel (approved):

joa-quim:
This commit must be reversed. It *broke* part of the man page.

Esteban82:
I think that the problem is that a remove a whitespace at the beggining of the table. I tought that that was the caused for the indetation of the table (as seen in gmtmath).

PR #7505, opened 2023-06-09
Improve math formatting for IGF in mgd77list do

PaulWessel:
There were too many \times symbols in those equations and the trig functions were not set as Latex (need \sin not just sin), etc. Also, need to use 10^exp and not e-6 etc in equations.


anbj (approved):
Thanks for reviewing, Paul.

anbj:
I didn’t want to mess too much with the equations in fear of introducing errors. But should’ve caught sin -> \sin, lat -> \phi, etc. Thanks!

PR #7530, opened 2023-06-13
By "basemap" we mean empty basemap

jidanni:
Because the majority of non-GMT basemaps contain actual maps,
https://www.google.com/search?q=basemap&tbm=isch
therefore GMT must make sure the user understands that GMT's "basemaps" do non contain actual maps.

Feel free to edit my edit.

E.g., me saying "frame" isn't always very exact either.

### DIFF
--- a/doc/rst/source/basemap.rst
+++ b/doc/rst/source/basemap.rst
@@ -35,7 +35,7 @@ Synopsis
 Description
 -----------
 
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Thanks. Too late for us to rename basemap to map frame or something more descriptive but I approved and merged your suggestion.

PR #7545, opened 2023-06-13
Report the true number of points that are non NaN

joa-quim:
See forum [post](https://forum.generic-mapping-tools.org/t/gmtinfo-on-binary-file-incorrect-count-for-non-nan-records/4033)


PaulWessel (approved):
Maybe mention this in the doc page?

joa-quim:
Added.

PR #7548, opened 2023-06-14
Update coast.rst

Esteban82:
Add link in dcw-collections.txt to the DCW-GMT repository


PaulWessel (approved):

anbj:
Consider using a double underscore to make an "anonymous" link. If you don't plan to refer to it, having two underscore will not cause trouble if you have other links with the same target text. See [here](https://stackoverflow.com/a/27435441) for a good explanation. This is mostly (always?) how it is used in the gmt docs; you create a nice external link and forget about it.

E.g.:

```
See more examples `here <https://interesing.com>`_ and `here <https://fabolousfigures.org>`_
```

will cause a warning/error since 'here' is used for multiple targets. Add another underscore and this problem goes away.

PR #7550, opened 2023-06-16
Also check -Y for c|f in modern mode

PaulWessel:
Serious lack of coffee.  Forgot to copy/paste the solution from **-X** parser to **-Y** parser, which leda to 100 extra failures where -Y was used...


Esteban82:
Blind approval.

PR #7567, opened 2023-06-23
Let -F work for -T as well

PaulWessel:
Option **-T** computes plate flexure and then depending on **-F** a geopotential value.  But looks like only FAA and maybe geoid was implemented?  This PR tries to:

1. Add in the two deflections and VGG as well, but not sure if correct.  Since you derived these expressions, perhaps you kan do the derivatives.
2. Remove two_pi with GMT constant **TWO_PI**

### DIFF
--- a/src/potential/gravfft.c
+++ b/src/potential/gravfft.c
@@ -1099,7 +1099,7 @@ GMT_LOCAL void gravfft_load_from_below_admitt(struct GMT_CTRL *GMT, struct GRAVF
 	   with a size of "nk" like that variable is computed here.	*/
 
 	unsigned int k, nk;
-	double	earth_curvature, alfa, delta_k, freq, D, twopi, t1, t2, t3;
+	double	earth_curvature, alfa, delta_k, freq, D, t1, t2, t3;
 	gmt_M_unused(GMT);
 
 	if (K->delta_kx < K->delta_ky)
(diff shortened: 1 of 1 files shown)


src/potential/gravfft.c
```diff
@@ -1178,28 +1176,58 @@ GMT_LOCAL void gravfft_load_from_top_grid (struct GMT_CTRL *GMT, struct GMT_GRID
...
+		 * THose where the only two cases covered by JL's code */
+
+		switch (Ctrl->F.mode) {
+			case GRAVFFT_FAA:
+				t1 *= 1.0e5;     /* To have it in mGals */
+				datac[k]   += (gmt_grdfloat) ((Ctrl->T.rho_cw * t1 * t2) * t / f * raised[k]);
+				datac[k+1] += (gmt_grdfloat) ((Ctrl->T.rho_cw * t1 * t2) * t / f * raised[k+1]);
+				break;
+			case GRAVFFT_GEOID:
+				if (mk > 0.0) t1 /= (NORMAL_GRAVITY * mk * TWO_PI);
```
joa-quim, line 1205: This ``if (mk > 0.0)`` condition was not in the old code, which was probably a bug.

joa-quim (approved):
Problem is that I also don't know what should be the result of these new cases.

joa-quim:
Ping

PaulWessel:
Hm, yes, I think something crashes? if mk == 0 and we divide by it, nothing good will happen.

PaulWessel:
Should I just merge this and let users eventually find a problem and give us a case?

joa-quim:
Likely yes. I may revisit these matters beginning of next year.

PR #7568, opened 2023-06-23
Remove empty first row

### LINKED ISSUE #7527: Tables with empty row at the beginning
I see that in General Features there are some tables like these two:

https://docs.generic-mapping-tools.org/dev/cookbook/features.html#distance-units
https://docs.generic-mapping-tools.org/dev/cookbook/features.html#specifying-pen-attributes

I wonder how that happened. It seems that some configuration was modified.

Esteban82:
Remove first empty rows of tables. I checked that version 5.4 of the docs there was no header.

In the preview looks fine.

Fixes #7527 (at least in this rst file).


joa-quim (approved):

Esteban82:
I only find those tables in two pages.

Issue #7558, opened 2023-06-19
Odd makecpt behavior from Julia


joa-quim:
```
# Correct

julia> makecpt(cmap=(:red,:green,:blue), range=[0 100 200 450])
Extract of a GMTcpt exposed as a GMTdataset for display. Colors converted to [0-255]
Model: rgb
Color depth: 24
3×9 GMTdataset{Float64, 2}
 Row │      r1       g1       b1       r2       g2       b2    alpha       z1       z2
     │ Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64
─────┼─────────────────────────────────────────────────────────────────────────────────
   1 │   255.0      0.0      0.0    255.0      0.0      0.0      0.0      0.0    100.0
   2 │     0.0    255.0      0.0      0.0    255.0      0.0      0.0    100.0    200.0
   3 │     0.0      0.0    255.0      0.0      0.0    255.0      0.0    200.0    450.0

# Inventive

julia> makecpt(cmap=(:red,:green,:blue), range=[0 100 200 451])
Extract of a GMTcpt exposed as a GMTdataset for display. Colors converted to [0-255]
Model: rgb
Color depth: 24
3×9 GMTdataset{Float64, 2}
 Row │      r1       g1       b1       r2       g2       b2    alpha       z1       z2
     │ Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64
─────┼─────────────────────────────────────────────────────────────────────────────────
   1 │   255.0      0.0      0.0    255.0      0.0      0.0      0.0      0.0    100.0
   2 │   255.0      0.0      0.0    255.0      0.0      0.0      0.0    100.0    200.0
   3 │     0.0      0.0    255.0      0.0      0.0    255.0      0.0    200.0    451.0
```

joa-quim:
CLI passes a string, not an array. That works fine in GMT.jl too
```
julia> makecpt(cmap=(:red,:green,:blue), range="0,100,200,451")
Extract of a GMTcpt exposed as a GMTdataset for display. Colors converted to [0-255]
Model: rgb
Color depth: 24
3×9 GMTdataset{Float64, 2}
 Row │      r1       g1       b1       r2       g2       b2    alpha       z1       z2
     │ Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64  Float64
─────┼─────────────────────────────────────────────────────────────────────────────────
[...]

PR #7569, opened 2023-06-23
Mechanism for passing -Tlist via a file

PaulWessel:
Se #7558 for background.  This WIP branch will examine the input data header record and if it contains "LIST" then we treat the nodes as we do a list.  WIth that, a file like

```
# LIST
0
100
200
451

```
called t.txt will give

```
gmt makecpt -Cred,green,blue -Tt.txt
0	red	100	red	L
100	green	200	green	L
200	blue	451	blue	B
B	black
F	white
N	128

```
i.e., same as `gmt makecpt -Cred,green,blue -T0,100,200,451`, while without the header record  we get the normal interpolation:

```
gmt makecpt -Cred,green,blue -Tf.txt
0	red	100	red	L
100	red	200	red	L
200	blue	451	blue	B
B	black
F	white
N	128
```

@joa-quim, if this allows you to do what you need to do from Julia, perhaps this is an acceptable solution? If so let me know and I will add documentation of the feature.


joa-quim (approved):
Yes, I may be able to use this from Julia (or do the conversion there directly), but this fixes CLI bug so good to go.

PR #7575, opened 2023-06-26
Update Scientific Color Maps to version 8

PaulWessel:
There are 4 new color maps added in v 8: glasgow, lipari, managua, and navia.  They have also messed around with details in almost a dozen cuts so I expect a few new failures.  We will deal with all of that when we tackle gs 10 for testing.  For now I jus updated Appendix 1b,c figures.


joa-quim (approved):
With their frequency of changes we should probably only use it in one single test.

PaulWessel:
Yes, that is a good idea.  In fact, since the majority of tests predateSCM I think there are only a few.  I typically get 52 failures but with the new SCM I think it might have been 55, and 2 of those were the App. Maybe, so not bad.

PR #7579, opened 2023-06-29
Update basemap.rst

jidanni:
Current example raises error message.

`basemap [WARNING]: Length <unit> d not supported - revert to default unit [cm]`


jidanni:
You may also want to fix the psbasemap page.

PaulWessel (approved):
Thanks!

PR #7580, opened 2023-06-29
Update explain_-J.rst_

jidanni:
1. The previous author had confused o/O with abc/ABC !
2. Note that attempting to use 1:xxxx syntax will result in:
```
basemap [ERROR]: Cannot specify map width with 1:xxxx format in -J option
basemap [ERROR]: Option -J parsing failure. Correct syntax:

   -J Select map proJection (<scale> in cm/degree, <width> in cm).
```


PaulWessel:
Thanks, I ended up with a separate PR so you can delete this one as fixed in #7584.

Esteban82:
By last Paul message I think that this PR can be close.

PR #7599, opened 2023-07-03
Handle commas in meca if -A and CSV

PaulWessel:
See pyGMT [issue](https://github.com/GenericMappingTools/pygmt/issues/1365#issuecomment-1336192548).

With **-A** we examine the trailing text, but if those items are comma-separated then the scanning will fail since the format assumes white-space.  This PR counts commas in trailing text and if 2 we change the scanning.

@seisman:

Please see if it works as is.

1. If not, please add a fprintf (stderr, trailing = %s\n", S->text[row]); so we can see the trailing text and manually check commas (not sure if there might be a leading one).
2. Once working, perhaps same is needed in other seis modules?

### DIFF
--- a/src/seis/pscoupe.c
+++ b/src/seis/pscoupe.c
@@ -1294,7 +1294,12 @@ EXTERN_MSC int GMT_pscoupe (void *V_API, int mode, void *args) {
 				/* Must examine the trailing text for optional columns: newX, newY and title */
 				/* newX and newY are not used in pscoupe, but we promised psmeca and pscoupe can use the same input file */
 				if (has_text) {
-					n_scanned = sscanf (S->text[row], "%s %s %[^\n]s\n", Xstring, Ystring, event_title);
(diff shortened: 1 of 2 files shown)


PaulWessel:
Sorry, did not notice at first this was a CLI example.  My fix seems to work.

seisman:
It fixes the original issue, but what if the columns are space-separated and the trailing text contains commas, e.g., 
```
gmt meca -R-5/5/-5/5 -JX10c -Sa2c -Baf -A -png map << EOF
0 0 10 0 90 0 5.0 2,2,eventname
EOF
```
In the above case, users expect to see `2,2,eventname` as the beachball title, but only `eventname` is shown.

seisman (approved):

seisman:
> Hm, have no way of knowing if the entire file was CSV since that reading was done earlier via GMT_Read_Data. ?

I believe so.

> 2\. Once working, perhaps same is needed in other seis modules?

I think `coupe` also needs update.

PR #7606, opened 2023-07-04
Document the -R+U option in the global -R

### LINKED ISSUE #7537: Document basemap -R...+U
https://docs.generic-mapping-tools.org/dev/gallery/ex28.html#example-28 says:
> `gmt basemap -R@Kilauea.utm.nc+Uk`

Alas, the "+U" is undocumented.

The best one can dig up is "+u".

joa-quim:
Fix #7537

### DIFF
--- a/doc/rst/source/explain_-R.rst_
+++ b/doc/rst/source/explain_-R.rst_
@@ -5,7 +5,7 @@ The **-R** option
 
 .. _-R:
 
-**-R**\ *xmin*/*xmax*/*ymin*/*ymax*\ [**+r**][**+u**\ *unit*]
+**-R**\ *xmin*/*xmax*/*ymin*/*ymax*\ [**+r**][**+u|U**\ *unit*]
     Specify the region of interest. |Add_-R|
 
 **Description**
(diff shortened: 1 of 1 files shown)


PaulWessel:
Think you are mistaken here.  The +U|u at the end of a grid does that scaling, but it is not the +u that belongs to -R. So the docs you linked is the right place and should not be added to -R

joa-quim:
I added stuff about **+U** not **+u** (which was already there). So, where do we document **+U**? The CookBook is not the right place for it.

PaulWessel:
Well, first half of cookbook is documentation and the examples etc are cookbook.  We have tons of explanations (pen, fill, text, etc) in the technical reference section.  We have extensive discussion of netCDF files and how to pull out a named variable, how to scale CPT z-values etc.  This is no different (**+U**|**u**_unit_ is also used for CPTs).

joa-quim:
But we already have **+u** in **-R**. Why not have the **+U** too? Relying in the CookBook to find an obscure **-R** flag is the same ~= nothing.

PaulWessel:
In -R, +uk means xmin/xmax/ymin/ymax are projected coordinates given in km and we scale by 1000 and then inverse project to get lon/lat.

As grid file modifier, +uk means scale grid xmin/xmax/ymin/ymax by 1000, but no inverting takes place. +u|U is also available for things like -Cmy.cpt+Uk.

We can chat about these things at the summit and make a list of tasks that are good things for some of our collaborators to deal with.  I also think we should break the cookbook into separate documents: Technical Reference for all this stuff, and Gallery for the examples.

joa-quim:
https://documentation.divio.com/

Esteban82:
Could this be merged?

seisman:
I agree with Paul that this is not the correct place to document **-R**. Perhaps it's better to document it at line 35 in the same file?

PR #7608, opened 2023-07-04
Update GMT_obl_nz.sh

jidanni:
Be consistent, within the same line, and with the rest of the example.

Or, make all four, and the rest of the file, full of unneeded {}. It is up to you.


PaulWessel:
We do this on purpose since there are times you must use the {} and there are times you don't have to, like here.  But it is a good habit to always use ${var} since you then never run into troubles like -T$valuek, where the k is mean to be a unit to $value...

joa-quim:
And this file is not public. 
Honestly, we cannot go to this level of details in opening issues. We simply have no resources for it.

jidanni:
> And this file is not public.

It is a example that gets included in the documentation.

joa-quim:
Won't change

PR #7623, opened 2023-07-11
Fix size of buffer sequence

ZMAlt:
Hi, everyone.

I found that the buffer generated by GMT has some weird points, and the result is different from Julialang's LibGEOS.jl

```bash
    gmt begin coast_buffer

        gmt coast -ETW -M -Dc > tw.geo
        gmt spatial -Qc1000+h tw.geo -fg > tw_temp.geo
        # buffer
        gmt spatial tw_temp.geo -Sb0.5 > tw_buffer.geo
        gmt plot -R116/124/20/26 -JQ5i tw.geo -W0.5p,black -B
        gmt plot tw_buffer.geo -W0.5p,red

    gmt end show
```
```julia
# julia code
using LibGEOS
# coastline from GMT
p1 = readgom("POLYGON((
       121.601206488        23.945428759 ,
       121.605109656        23.9400906133 ,
       121.603835991        23.9382766802 ,
       121.603835991        23.9382766802 ,
       ...
))")
g1 = buffer(p1, 0.5)
```

GMT in red  and LibGEOS.jl in blue
image

At first, I found the numbers of buffer squeence between GMT and LibGEOS are different. 
GMT sets the `quadsegs` to 30.  `geom_out = GEOSBuffer_r(handle, geom, buf_dist, 30)`. LibGEOS sets it to 8.
When changing the parameter in GMT from 30 to 8, there are still some strange points in GMT‘s result.
Then, I found the problem is with the size of `seq_out` which is different from `geom_out`. 
I fix it and weird points from GMT disappear as follows.

coast_buffer

**Reminders**

### DIFF
--- a/src/gmtspatial.c
+++ b/src/gmtspatial.c
@@ -2501,11 +2501,11 @@ int geos_method_linestring(struct GMT_CTRL *GMT, struct GMT_DATASET *Din, struct
 				GMT_Report (GMT->parent, GMT_MSG_NORMAL, "Failed to create output GEOS sequence for table %d, segment %d.\n", nt, ns);
 				continue;
 			}
-            /* geom_out's size and seq_out's size are different */
-            if (GEOSCoordSeq_getSize_r(handle, seq_out, &n_pts) == 0 ) {
+            		/* geom_out's size and seq_out's size are different */
+            		if (GEOSCoordSeq_getSize_r(handle, seq_out, &n_pts) == 0 ) {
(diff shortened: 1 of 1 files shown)


joa-quim (approved):
Thanks for fixing it. I would have had difficulties remembering again how to use these GEOS functions.

And, BTW, instead of ``libGEOS.jl`` you can do the buffering easily  with ``GMT.jl``

```
TW = coast(E=:TW, dump=true);
D = buffer(TW, 0.5);
plot(D, proj=:guess)
plot!(TW, show=true)
```

ZMAlt:
> And, BTW, instead of libGEOS.jl you can do the buffering easily with GMT.jl

Got it. Always keep an eye on GMT.jl

PR #7631, opened 2023-07-12
Left-over fprint message now removed

### LINKED ISSUE #7630: Temperamental?
```
gmtset  FONT_ANNOT_PRIMARY +10p
Not doing it
```

PaulWessel:
See #7630.  I am guessing I had forgotten about a temporary fprintf during debugging.  It is harmless but now changed to a DBG message so not seen unless you want to via **-Vd**. Closes #7630.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -10379,7 +10379,10 @@ void gmt_set_undefined_defaults (struct GMT_CTRL *GMT, double plot_dim, bool con
 #endif
 
 	/* Refuse to do this in gmtset */
-	if (!strcmp (GMT->init.module_name, "gmtset")) {fprintf (stderr, "Not doing it\n"); return; }
+	if (!strcmp (GMT->init.module_name, "gmtset")) {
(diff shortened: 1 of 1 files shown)


joa-quim (commented):
But +10p is a valid setting and it’s refusing to accept it.

PaulWessel:
No, it is set already earlier so just exiting that function - with a bad message.  See the gmt.conf I gave - all dimensions are set correct, no?

joa-quim:
In mobile only till the afternoon

joa-quim:
In the gmt.conf, it lost the "'+"

```
FONT_ANNOT_PRIMARY             = 10p,Helvetica,black
```

PaulWessel:
Pretty sure the leading plus is just a convenience operator to adjust all those settings proportionally. It does that, but does not write out that operator. That was how it was implemented

joa-quim:
Still confused. That means the ``gmt set FONT_ANNOT_PRIMARY +10p`` does one thing when is executed but a different one when the ``gmt.conf`` is read.

PaulWessel:
Just search for ANNOT_FONT in gmt_init.c and see what it does then. There is a reason why it became this way.

joa-quim (approved):

PR #7637, opened 2023-07-13
Update session-2.rst

jidanni:
Else we think, "Oh he's talking about a different column", and need to count them again.


joa-quim (commented):
In fact there are more serious errors there. Instead of
```
4th, 3rd, 5th, and 6th column and scale the last column by 0.1
```
it should be
```
5th, 4rd, 6th, and 7th column and scale the 7th column by 0.1
```

jidanni:
Yes, please fix it. All those columns and my head gets hazy.

joa-quim:
Done directly on master.

PR #7641, opened 2023-07-14
Give error if -i does not list at least one column

PaulWessel:
Per syntax and parsing **-i** needs a column or column rnge.  So things like **-i+o**1000 is an error since no columns were specified (as in this [forum](https://forum.generic-mapping-tools.org/t/gmt-math-i-t-should-print-warning/4123) post). Now we check if there is an argument after stripping off modifiers.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -9522,6 +9522,11 @@ int gmt_parse_i_option (struct GMT_CTRL *GMT, char *arg) {
 			}
 		}
 
+		if (p[0] == '\0') {	/* No range given */
+				GMT_Report (GMT->parent, GMT_MSG_ERROR, "-i: No columns specified\n");
+				return (GMT_PARSE_ERROR);
+		}
+
 		if (p[0] == 't') {	/* Got the trailing test "column" */
 			GMT->current.io.trailing_text[GMT_IN] = GMT->current.io.trailing_text[GMT_OUT] = true;
 			if (p[1]) {	/* Want a specific word (0-(nwords-1)) from the trailing text */


jidanni:
All I know is
https://docs.generic-mapping-tools.org/dev/cookbook/options.html#input-columns-selection-the-i-option
says
>  [Default reads all columns in order, starting with the first column (i.e., column 0)]. 

So maybe there needs to now be a way to say "all columns" -ia? Without needing to count them. To be able to do
"multiply all columns by 1.2" without needing to know how many are being input.

PaulWessel:
Why would you use -i at all if you want to use all columns? That has always been the default.  E.g.

`echo 1 2 3 4 5 6 7 | gmt math STDIN 10 MUL =`

and if the frist columns is not "time" (so not scaled) you add -Ca:

`echo 1 2 3 4 5 6 7 | gmt math -Ca  STDIN 10 MUL =`

Can do -i0: but why since that is the same as no -i at all.

joa-quim:
Maybe what is wanted is

echo 1 2 3 4 5 6 7 | gmt math ia STDERR 10 MUL =

jidanni:
I was thinking of when -i is used outside of gmt math.

PaulWessel:
outside of gmt math the number of expected columns is typically fixed (e.g., lon, lat, maybe one z) and if you were thinking of having many lon, lat pairs on one row then that is not going to happen for coding reasons. Perhaps you can show a line of input to some module you would imagine more columns so I can see what you mean?

PaulWessel:
> Maybe what is wanted is
> 
> echo 1 2 3 4 5 6 7 | gmt math ia STDERR 10 MUL =

STDERR is an output source (like STDOUT) so opening that source as read does not sound like a good idea.

jidanni:
This is getting over my head. I made
https://github.com/GenericMappingTools/gmt/pull/7649
to help match what I think you are saying -i does.

PR #7644, opened 2023-07-14
Update mapproject.rst

jidanni:
Fully document -Q. (I wish STDERR wasn't used.)

### DIFF
--- a/doc/rst/source/mapproject.rst
+++ b/doc/rst/source/mapproject.rst
@@ -199,8 +199,8 @@ Optional Arguments
 .. _-Q:
 
 **-Q**\ [**d**\|\ **e**]
-    List all projection parameters. To only list datums, use |-Q|\d. To
-    only list ellipsoids, use |-Q|\e.
+    List all projection parameters to STDERR and exit. No input files are read nor are other options checked.
+    To only list datums, use |-Q|\d. To only list ellipsoids, use |-Q|\e.
 
 .. _-S:
 


doc/rst/source/mapproject.rst
```diff
@@ -199,8 +199,8 @@ Optional Arguments
 .. _-Q:
 
 **-Q**\ [**d**\|\ **e**]
-    List all projection parameters. To only list datums, use |-Q|\d. To
-    only list ellipsoids, use |-Q|\e.
+    List all projection parameters to STDERR and exit. No input files are read nor are other options checked.
```
joa-quim, line 202: I don't agree with the mention to STDERR. First, very few users know what that is. Second, we don't mention it in all other printed statements.

jidanni: You are very correct that this shouldn't be mentioned... Because indeed I have "_proved_"  that it shouldn't go to standard error in the very first place! You guys are breaking the Unix traditions:

https://github.com/GenericMappingTools/gmt/discussions/7616

jidanni: Well, even if it's not a man page, but commands asking for information should put it  on standard output not standard error.

PaulWessel: We are not switching to stdout for these.  Unbeknownst to you, our i/o model is (overly) complex since we are enabling i/o via memory held by Julia, MATLAB, or Python.  They do not expect text output when they call GMT modules directly from the API.

jidanni: All I know is
```
$ proj -l | wc
    180     829    5210
$ xrandr --query | wc
     51     223    2177
```
and then the user can parse that output.
It's different than gmt ... -V,
where STDERR is good because there is already something else on STDOUT.

PR #7645, opened 2023-07-14
Update mapproject.rst

jidanni:
Mention what happens with FBO conflicts.


PaulWessel (commented):
No, we won't start doing that.  Throughout GMT, it is understood that a|b|c|d... means that only one of these codes can be given.  I am sure you can give all three and as you say it picks the first but that is because we never felt spending time on checking all the things a user might place there and give warnings and errors.

PR #7648, opened 2023-07-16
Update cookbook.rst

jidanni:
Yes, the first time the user looks at this page, it would be nice to have the picture greet the user and read the credits, to indeed put a long-needed human face to GMT.

But on second and subsequent visits it just becomes a long block the user must scroll past in order to find the item in the TOC that s/he wants to click on.


jidanni:
Actually, this chunk should in fact be moved to a separate "faces.html" or "people.html" with title "The people behind GMT" .
And it should be the first page after the TOC.

PR #7649, opened 2023-07-16
Update explain_-icols_full.rst_

jidanni:
For https://github.com/GenericMappingTools/gmt/pull/7641 . I am just trying to make the words match what I think what is being said there. I did not test any of this.


joa-quim:
I prefer the way we have it. 
And seriously, we could create thousands on nano-things like this but we also have much more important things where to spend our time.

joa-quim:
Won't change

Issue #7626, opened 2023-07-11
subplot -C  crashes GMT


joa-quim:
Although this happened to me by accident, the man page  says

-C[side]clearance

But if we do not specify a side, at line subplot#L454
```
n_errors += gmt_M_repeated_module_option (API, Ctrl->C.set[side]);
```
``side`` is used without being defined. Not sure what's the fix here.

PaulWessel:
The intention was if no side then set all sides to given value but looks like I forgot. Does doc say anything?

joa-quim:
" No side means all sides."

joa-quim:
The fix was not good enough. Now ``-C0.25c`` raises
```
subplot [ERROR]: Option -C: 0.25c is not a valid side argument.
```

joa-quim:
Fixed

PR #7665, opened 2023-07-21
New fix for the #7626 issue (-C alone or without _side_)

joa-quim:
This seams to fix both previously identified issues.


PaulWessel (commented):
Not sure. I think the NULL char is part of any string and thus. If opt->arg[0] is NULL the in returns true. That is why I check it before calling strchr.

joa-quim:
It works fine now
```
C:\v\build>gmt begin

C:\v\build>gmt subplot begin 2x2 -Fs6c -C -M0.5/1.5c
subplot [ERROR]: Option -C without arguments is not a valid syntax.
```

PaulWessel (approved):

PR #7666, opened 2023-07-23
Update triangulate docs to explain why Shewchuk is missing

PaulWessel:
See the [forum](https://forum.generic-mapping-tools.org/t/update-the-apt-get-version-of-gmt/4136/10) for the silliness.


joa-quim (approved):
Yes, but what I had in mind is a runtime warning.

PaulWessel:
Seems a bit hash if you dont really care about this and still get that message every time you need some basic triangulation.  I think this is a reasonable compromise (clear docs).

PR #7669, opened 2023-07-23
Remove test pragmas and old code for qsort_r

PaulWessel:
Now that we redid how qsort_r|s are handled we can remove the commented-out old code as well as the excessive pragma warnings.


joa-quim (approved):
Not tested yet though.

PaulWessel:
Probably should give it a check since I cannot. Ping here when you get there.

joa-quim:
Tests passed.

PR #7680, opened 2023-07-25
WIP Enhance the curves in events to do linear, quadratic or cosine ramp

PaulWessel:
So far all of the magnification curves have been quadratic.  This PR adds optional directives **l**(inear), **q**(uadratic) or **c**(osine ramp) to the rise, decay and fade sections.  E.g, cosine ramp rise of 2 time units would use the **-E** modifier **+rc**2. Below shows the 4 curves for each type (the dashed curve is the default size curve when no curve tweaks have been requested (i.e., no rise, plateau, etc):

lines

It is a WIP for now since we are playing with this a bit and may add this one as new test eventually.


Esteban82:
It is working fine. I run psevents.sh and I got the same figure.

lines

Esteban82 (approved):

PR #7694, opened 2023-07-29
Update mapproject.rst

jidanni:
Indent to get yellow examples, not white.

PR #7700, opened 2023-07-31
mapproject.rst: fix code blocks

anbj:
Fix code blocks in mapproject.rst.
See #7694.

### DIFF
--- a/doc/rst/source/mapproject.rst
+++ b/doc/rst/source/mapproject.rst
@@ -334,97 +334,72 @@ Examples
 .. include:: explain_example.rst_
 
 To transform a remote file with (latitude,longitude) into (x,y) positions in cm
-on a Mercator grid for a given scale of 0.5 cm per degree and selected region, run
+on a Mercator grid for a given scale of 0.5 cm per degree and selected region, run::
 
-::
-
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):

joa-quim:
But we decided to do exactly the contrary of this. 
No tabs at beginning of examples.

PaulWessel:
@joa-quim is right. I am at too high altitude to be trusted and @jidanni is posting too much

anbj:
But look at the docs (mapproject) now; this is how we want it, yes?

anbj:
You need to indent the stuff that you want inside a code block following the ``::``. This is in the markup (rst) source. The rendered code block (html) will not be indented. I think it is correct now, but please [check](https://docs.generic-mapping-tools.org/dev/mapproject.html#examples).

joa-quim:
No. It is indented.

anbj:
This is what I see. Isn't this the way it is supposed to look?

Screenshot 2023-07-31 152602

joa-quim:
Well I see the same, there is a small indent right?
But this is not important and I’m not near a computer for now and next week.

PR #7711, opened 2023-08-09
Let Glib threads be true as default

PaulWessel:
LIke the template, both openMP and gthreads support was by default off.

### DIFF
--- a/src/potential/gmtgravmag3d.c
+++ b/src/potential/gmtgravmag3d.c
@@ -245,7 +245,7 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 #ifdef HAVE_GLIB_GTHREAD
 	GMT_Option (API, "x");
 #else
-	GMT_Usage (API, 1, "\n-x Not available since this binary was not build with multi-threading support.");
+	GMT_Usage (API, 1, "\n-x Not available since this binary was not build with GLIB multi-threading support.");
 #endif
 	GMT_Option (API, ":,.");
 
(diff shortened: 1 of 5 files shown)


PaulWessel:
@joa-quim, I need to sort out if -x is available.  Right now we have code that adds "-x" or "" depending on OpenMP, but if GLIB is used that is separate.  So adding a new define constant that will be -x or NULL, set in gmt_glib.h.  Other than grdfilter, which potential programs are using threads?

joa-quim:
grdfilter and gmt|grdgravmag3d are the ones that use gthreads and they have **-x**

The **-x** setting is done in

```
#ifdef HAVE_GLIB_GTHREAD
	if (GMT->common.x.n_threads > 1)
		threads = gmt_M_memory (GMT, NULL, GMT->common.x.n_threads, GThread *);
#endif
```

PaulWessel:
@joa-quim, please have a look at this PR.  What it tries to do is:

1. set using GLIB cmake setting to TRUE as default in the cmake/advanced file (was FALSE)
2. Duplicate the constants `GMT_x_OPT` and `GMT_ADD_x_OPT` to `GMT_xg_OPT` and `GMT_ADD_xg_OPT` and set then separately from the OpenMP statements (in gmt_glib.h)
3. Only display **-x** usage in synopsis and usage if indeed compiled with GLIB.

Seems OK here?

src/potential/gmtgravmag3d.c
```diff
@@ -241,7 +241,13 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Usage (API, -2, "Set z level of reference plane [Default = 0].");
 	GMT_Option (API, "bi");
 	GMT_Usage (API, 1, "\n-fg Converts geographic grids to meters using a \"Flat Earth\" approximation.");
-	GMT_Option (API, "h,i,o,r,:,.");
+	GMT_Option (API, "h,i,o,r");
+#ifdef HAVE_GLIB_GTHREAD
+	GMT_Option (API, "x");
+#else
+	GMT_Usage (API, 1, "\n-x Not available since this binary was not build with multi-threading support.");
```
joa-quim, line 248: Maybe make this message more clear that we are talking of GLIB muti-threading and not OMP.

PaulWessel: OK, done

joa-quim (approved):

PR #7715, opened 2023-08-12
Use plus/minus char instead of -1/+1 kludges

PaulWessel:
Since RST man pages can handle that character, we use ±1 instead.


anbj (approved):
Looks good.

PR #7746, opened 2023-08-20
Fix errors in gmtspatial related to -N

### LINKED ISSUE #7745: gmtspatial -N fails to indentify the points-in-polygon
See forum post https://forum.generic-mapping-tools.org/t/gmtspatial-issue/4165/5

PaulWessel:
See #7745 for background. There were three problems in **gmt spatial -N** that have been addressed:

1. Failed to exclude features completely outside the polygon
2. Flawed implementation of +a could easily failed if the middle point was outside
3. Failed to turn off output from segments that failed the test

I also added a simple text test called test/gmtspatial/inout_features.sh

Closes #7745.

### DIFF
--- a/src/gmtspatial.c
+++ b/src/gmtspatial.c
@@ -38,7 +38,10 @@
 #define THIS_MODULE_NEEDS	""
 #define THIS_MODULE_OPTIONS "-:RVabdefghijoqs" GMT_OPT("HMm")
 
-#define GMT_W	3
+#define GMT_N_MODE_NOTSET	0
+#define GMT_N_MODE_REPORT	1
+#define GMT_N_MODE_ADD_ID	2
+#define GMT_W   3
 
 #define POL_UNION		1
 #define POL_INTERSECTION	2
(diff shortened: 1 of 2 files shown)


joa-quim (approved):

macpurity:
I appreciate that this issue has been closed, but I now see error messages with -N when specifying +p, without setting a "start" value.

The on-line documentation indicates the +p will default to zero, but this command (part of a much larger piece of code) causes an error message:

`echo ${flonlat[@]} | gmt gmtspatial -N${usbdy80k}+p+r`

What this line does is determine whether a lon,lat point falls inside a simple conterminous boundary of the USA. This command used to work, but now it issues the following:

`gmtspatial [ERROR]: Option -N: Modifier +p given no value`

The easy work-around is to specify +p0. Which is what I did to get the script to work properly. If there no longer a default value, then the on-line documentation should be changed.

The on-line documentation currently reads: "or it is assigned the running number that is initialized to start [0]." This suggests that zero is the default value if none is provided.

PaulWessel:
Fixed in #8039.

macpurity:
Thanks, Paul. That did the trick!

PR #7748, opened 2023-08-21
gmtspatial -N+i+pstart is allowed

PaulWessel:
The updated **gmtspatial** checks would refuse **+p**_start_ if **+i** is used.  Now that is relaxed, the usage messages are cleaned up a bit, as is the docs.

PR #7749, opened 2023-08-21
Alternative to #7748 that also takes care of printing correct IDs when +i+p

joa-quim:


joa-quim (commented):
OK, the +p is working now.

PR #7750, opened 2023-08-21
Add a modifier to -S to let polygon counting in headers start at 1 instead of 0

joa-quim:


PaulWessel (changes requested):
No, I dont like it, hardwired to 1.  Just do the same as for -N: +p<start> with default = 0.

joa-quim:
It's nor hardwired to 1. Ctrls->S.onebased defaults to 0

PaulWessel:
I mean, I cannot make it start at 100 if I wanted to.  So why not use a variable argument via +p.

joa-quim:
Ok, changed it to -S[<n>].

PaulWessel (approved):
Good, yes no need for +p in this case, just n.

PR #7752, opened 2023-08-21
CI: Install ghostscript 9.56.1 via conda on Linux/macOS to have a stable/consistent gs version

seisman:

We're having hundreds of failing tests on macOS, because Homebrew is providing the latest and buggy Ghostscript 10.01.2. Even if future gs versions fix all the bugs, we may still have issues because Ubuntu and macOS may provide different gs versions. For example, currently, Ubuntu 22.04 has gs 9.55, and Homebrew on macOS provides gs 10.01.2.

To avoid issues with different gs versions on different platforms, we can install Ghostscript from conda instead. In this PR, I modify the CI scripts so that we install Ghostscript 9.56.1 from conda on Linux/macOS. 

Windows is more tricky to support. Currently, it uses gs 9.50 from [Chocolatey](https://chocolatey.org/).

**Reminders**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,7 @@ on:
     branches:
       - master
       - 6.[0-9]+
+      - pin-gs
 
 defaults:
   run:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,6 @@ on:
     branches:
       - master
       - 6.[0-9]+
-      - pin-gs
 
 defaults:
   run:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,7 +9,6 @@ on:
     branches:
       - master
       - 6.[0-9]+
-      - pin-gs
     paths:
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - master
       - 6.[0-9]+
+      - pin-gs
```
seisman, line 12:
```suggestion
```

.github/workflows/docs.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - master
       - 6.[0-9]+
+      - pin-gs
```
seisman, line 12:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - master
       - 6.[0-9]+
+      - pin-gs
```
seisman, line 12:
```suggestion
```

PaulWessel (approved):

PR #7760, opened 2023-08-24
Add test program testapi_modern.c

PaulWessel:
This makes sure the issue #4518 reported by @sarzeaud is preserved.  The test fails for now (well we need to add an actual comparison of the output) but this lets us play in the debugger. We need to determine why **-A** is not given to **psconvert** in **gmt end**. Passing the **GMT_SESSION_RUNMODE** is not enough.

**Note**: Any C program that is called `testapi_*.c` will automatically be built.

### DIFF
--- a/src/testapi_modern.c
+++ b/src/testapi_modern.c
@@ -1,0 +1,21 @@
+/* Issue from GitHub: https://github.com/GenericMappingTools/gmt/issues/4518
+ * Problem seems to be that gmt begin comes after GMT_Create_Session in the
+ * code so there are orientation and page size issues. In gmt end, the
+ * psconvert call is missing the -A.
+ */
+
+#include "gmt.h"
+
+int main () {
+	void *API = NULL;
+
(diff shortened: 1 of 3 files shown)


PaulWessel:
> Shouldn't you add the A option to the call to the begin module in testap_modern.c?
> Moreover, I have seen in the GMT code that the A option is sometimes automagically given to psconvert. I think it could be interesting to add the option A,I+m0.5c to the call to the begin module, in order to check that all the options are really given to psconvert, and that the A option is not added another way.

We will play with it.  In modern mode default is portrait and A is on (as the CLI example shows).  So just need to dive in there...

PaulWessel:
@Esteban82 perhaps you can approve this - it is just adding a test of a small C program that should make the same plot as a CLI script, but it does not.  having this in master makes it easier for me to debug GMT changes needed to fix this.

seisman (approved):

PR #7772, opened 2023-08-30
Ensure given distances in gmtvector are positive

### LINKED ISSUE #7769: gmt vector -Tt: negative distances read as positive unless entered via column
```
$ echo 0 0 45 -1111111 | gmt vector -Tt -jg
-7.1016558127   -7.04772447977
$ echo 0 0 | gmt vector -Tt45/-1111111 -jg
7.1016558127    7.04772447977
$ gmt --version
6.4.0
```
Probably related to https://docs.generic-mapping-tools.org/dev/changes.html
> We have a new GMT common option -j that clarifies how to select flat Earth, great circle, and geodesic calculations and thus eliminates awkward, sign-based increments.

PaulWessel:
No such thing as a negative distance.  Closes #7769.

### DIFF
--- a/doc/rst/source/gmtvector.rst
+++ b/doc/rst/source/gmtvector.rst
@@ -119,14 +119,15 @@ Optional Arguments
     angle for 2-D Cartesian data and *lon/lat/angle* for a 3-D rotation
     pole and angle), **R** will instead rotate the fixed secondary vector
     by the rotations implied by the input records, **t** to translate
-    the input point by a positive *distance* in the *azimuth* direction (append *azimuth*/*distance*\ [*unit*]
+    the input point by a *distance* in the *azimuth* direction (append *azimuth*/*distance*\ [*unit*]
     for the same translation for all input points, or just append
     *unit* to read *azimuth* and *distance* (in specified *unit* [**e**])
     from the third and fourth data column in the file, and **x** for cross-product.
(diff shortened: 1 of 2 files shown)


PaulWessel:
Actually, perhaps it is more user friendly to simply stay that if distance is negative we make it positive and add 180 to the azimuth?

joa-quim (approved):

PaulWessel:
I am pretty sure the separation of two points on the earth is a positive definite quantity.  But as you mention, one could pick a point as an origin and then one could have a positional coordinate with sign.  Anyway ,as you see above I allow negative distances.

jidanni:
All I know is now I see in the code changes of today a new error message when a person gives a negative distance. Also there's new documentation requiring a positive distance. All I can conclude is negative distance are now "criminalized."

Consider a person who takes two steps forward, and one step back, but never turns around.

Consider an automobile. That drives 150 ft forward and then puts the gear into reverse and drives 50 ft backward.

That is much different than turning the car around 180° and going backward.

PaulWessel:
Those who knows what they are doing will pull the very latest from master which was this morning, about 12 hours ago:

echo 0 0 | gmt vector -Tt45/-1
-0.707106781187	-0.707106781187
echo 0 0 | gmt vector -Tt45/1
0.707106781187	0.707106781187

My patience with you is wearing thin.

Issue #7777, opened 2023-09-02
using coupe to map focal mechanism, don't move beach ball location when providing a new coordinates.


dh4931:
**Description of the problem**
using coupe to map focal mechanism, don't move beach ball location when providing a new coordinates.
using moment tensor to map focal mechanism and providing -A option, I want to the new location(400, 200), but the location of beach ball don't move.

**Full script that generated the error**

```
#!/bin/bash
gmt begin profile2
gmt coupe -JX15c/-6c -Sd0.8 -Aa130/43/140/36/90/100/0/700 -R0/600/0/900  -Gblack -Q -N << EOF
131.55 41.48 579  1.14 -0.10 -1.04 -0.51 -2.21 -0.99 26 400 200
EOF
gmt basemap -BWS -Byaf+l"Focal depth (km)" -Bxaf+l"Distance (km)"
gmt end show
```

**Actual outcome**

![image](https://github.com/GenericMappingTools/gmt/assets/53483302/836605bd-3e6d-44ad-8de4-83143371c288)

**Expected outcome**

i want to move the beach ball to the new location(400,200)

**System information**

* Operating system:  Ubuntu 20.04.6 LTS
* GMT version (`gmt --version`): Version 6.2.0

PaulWessel:
The docs says give alternative long, lat, not projected km. But I can see it could be useful with a modifier that says “use those coordinates as already projected. @seisman?

seisman:
Looking at the old GMT 4 (https://docs.generic-mapping-tools.org/4/gmt/html/man/pscoupe.html) and GMT 5 docs (https://docs.generic-mapping-tools.org/5.4/supplements/meca/pscoupe.html), they all say:

> these two columns are not used; can be 0 0; allows use of the psmeca file format

Since GMT 6.0, the `coupe` docs started to say:

> 8,9: longitude, latitude at which to place beachball if -C is used (optional). Using 0,0 in columns 8 and 9 will plot the beach ball at the longitude, latitude given in columns 1 and 2. The -: option will interchange the order of columns (1,2) and (8,9).

It's because we incorrectly use the same file `explain_meca_-S.rst_` in both `meca` and `coupe` docs.

In my opinion, giving alternative long/lat doesn't make sense in `coupe`. It should always be in the projected coordinates.

PaulWessel:
[...]

PR #7778, opened 2023-09-02
Enhance meca and coupe w.r.t. adjusting positions and connecting with a line

PaulWessel:
This PR started with the discussions on #7777.  To prepare this, I found that there were much duplicated `#define` statements in _psmeca.c_ and _pscoupe.c_ so I pulled those into _meca.h_. Then, given we want a "-A"-like option in **coupe** I placed the **-A** parser from _psmeca.c_ into _utilmeca.c_ as well as the usage message now that it will have more modifiers.  Then there is a common structure defined in meca that is included in Ctrl as member A or D.  This reduces the amount of code needed to implement the desired features. This initial commit just does that - no actual "-D" option will be acted on yet.  However, the parser has been extended to process +c and +o but they are basically ignored.


PaulWessel:
Hi @seisman, draft version 1 is pushed.  I tested coupe with -D+o (reads the alternative coordinates as offsets instead) and -Dodx/dy (same offset from projected location for all points).  Things not tested (but hope they work as is):

Use CPT in coupe and with -D (should be completely decoupled)
Do same tests with meca

Questions:

1. Would be easy to add modifier **-D +g**_fill_ if you want to paint that circle with another color than the beachball. Could also all **+g-** which would _not_ fill the circle.
2. I did not implement a **+c** to affect conversions. meca will always expect an alternative plot location in geographical coordinates, whereas coupe should aways expect offsets in plot units.  So unless I am missing something I do not see a need for any **+c**.. When +o is given we read those two columns as plot units in both modules.

PaulWessel:
Once things work we should make a test or two.

PaulWessel:
OK, I added documentation for coupe -D and updated meca -A docs as well.
Have a look.  I decided it only made sense in coupe that the two alternative coordinates are in distance, depth since if they were lon lat then the conversion to distance,depth requires (lon, lat, depth) and we would presumably have to use epicenter depth to do that.  If would be simple to do but would require a new +c modifier just for coupe.  Then coupe could do (with -D):

Default assume the two cols are distance, depth in the crossection
+c No, they are lon, lat and with epicentre depth compute distance, depth for the alternate location in the crossection
+o offset the final alternate location by dx/dy:
  With +c, only +odx/dy is allowed since already read two cols to compute alternative coordinates.
  Without +c, if just +o then the two cols have variable dx dy in plot units.
  Without +c, +odx/dy adjusts the alternate locations by this fixed offset.

Comments? Shall I add +c to coupe.

PaulWessel:
Note I got bored and added **+g**[_fill_] for how to fill the circle [Default is same as beachball as before] but also modified **+s** to be  **+s**[_symbol_]_size_ where symbol can be any of the standard geometric plot symbols [**c**].
Might be done?

PaulWessel:
`gmt coupe meca.dat -Sa1c -Aa111/33/119/33+d90+w500+z0/50+r -Q -JX15c/-8c -B -D+o+sa6p+gyellow -Gred -png map
`
map

Should we do something with the label justification if **+o** gives a negative _dy_? i.e., TC -> BC?

PaulWessel:
Hi @seisman, I ended up adding **+c** to **coupe -D** since people plotting with meca may have alternative geographic coordinates which they use with **meca -A** but **coupe** would normally expect _distance, dept_h coordinates instead.  The **-D+c** projects the alternative geographic coordinates into the crossectional plane.  After than, **+**odx/_dy_ can be used to shift any further since hard to know where things project.  With no **+c** then we expect distance, depth in those columns.  In that case, **+**odx/_dy_ can be used to apply a constant shift to all alternative events. If just **+o** is given then we initialise the alternate plot locations as the same as the original plot location, then add variable _dx dy_ read from those two special columns.

See what you think.  I have updated docs.

src/seis/meca.h
```diff
@@ -27,45 +27,81 @@
...
+
+#define SEIS_PLOT_DC		1
+#define SEIS_PLOT_AXIS	2
+#define SEIS_PLOT_TRACE	4
+#define SEIS_PLOT_TENSOR	8
+
+#define SEIS_CART_OFFSET	1	/* Cartesian plot offset in optional|trailing text */
+#define SEIS_CART_OFFSET_FIX	2	/* Same, but given as fixed offset with +o on option line */
+
+#define SEIS_EVENT_FILL		0	/* Default is to fill optional symbol with even color */
```
seisman, line 49:
```suggestion
#define SEIS_EVENT_FILL		0	/* Default is to fill optional symbol with event color */
```

doc/rst/source/supplements/seis/psmeca.rst
```diff
@@ -16,7 +16,7 @@ Synopsis
 |-J|\ *parameters*
 |SYN_OPT-R|
 |-S|\ *format*\ [*scale*][**+a**\ *angle*][**+f**\ *font*][**+j**\ *justify*][**+l**][**+m**][**+o**\ *dx*\ [/*dy*]][**+s**\ *reference*]
-[ |-A|\ [**+p**\ *pen*][**+s**\ *size*] ]
+[ |-A|\ [**+g**\ [*fill*][**+o**\ *dx*\ [/*dy*]][**+p**\ *pen*][**+s**\ [*symbol*]\ *size*] ]
```
seisman, line 19:
```suggestion
[ |-A|\ [**+g**\ [*fill*]][**+o**\ *dx*\ [/*dy*]][**+p**\ *pen*][**+s**\ [*symbol*]\ *size*] ]
```

doc/rst/source/supplements/seis/pscoupe.rst
```diff
@@ -20,6 +20,7 @@ Synopsis
 |-S|\ *format*\ [*scale*][**+a**\ *angle*][**+f**\ *font*][**+j**\ *justify*][**+l**][**+m**][**+o**\ *dx*\ [/*dy*]][**+s**\ *reference*]
 [ |SYN_OPT-B| ]
 [ |-C|\ *cpt* ]
+[ |-D|\ [**+c**][**+g**\ [*fill*][**+o**\ *dx*\ [/*dy*]][**+p**\ *pen*][**+s**\ [*symbol*]\ *size*] ]
```
seisman, line 23:
```suggestion
[ |-D|\ [**+c**][**+g**\ [*fill*]][**+o**\ *dx*\ [/*dy*]][**+p**\ *pen*][**+s**\ [*symbol*]\ *size*] ]
```

src/seis/pscoupe.c
```diff
@@ -72,13 +57,14 @@ struct PSCOUPE_CTRL {
...
 		struct GMT_PEN pen;
-		struct nodal_plane PREF;
+		struct SEIS_NODAL_PLANE PREF;
 		char newfile[PATH_MAX], extfile[PATH_MAX];
 	} A;
 	struct PSCOUPE_C {	/* -C<cpt> */
 		bool active;
 		char *file;
 	} C;
+	struct SEIS_OFFSET_LINE D;	/* -D[+c][+o[<dx>[/<dy>]]][+p<pen>][+s<size>] */
```
seisman, line 67:
```suggestion
	struct SEIS_OFFSET_LINE D;	/* -D[+c][+g[<fill>]][+o[<dx>[/<dy>]]][+p<pen>][+s<size>] */
```

src/seis/pscoupe.c
```diff
@@ -1302,7 +1294,12 @@ EXTERN_MSC int GMT_pscoupe (void *V_API, int mode, void *args) {
 					n_scanned = sscanf (tmp_buffer, "%s %s %[^\n]s\n", Xstring, Ystring, event_title);
 					if (n_scanned >= 2) { /* Got new x,y coordinates and possibly event title */
 						unsigned int type;
-						if (GMT->current.setting.io_lonlat_toggle[GMT_IN]) {	/* Expect lat lon but watch for junk */
+						if (Ctrl->D.mode == SEIS_CART_OFFSET) {	/* Cartesian dx and dy in plot units */
+							Ctrl->D.off[GMT_X] = gmt_M_to_inch (GMT, Xstring);
+							Ctrl->D.off[GMT_Y] = gmt_M_to_inch (GMT, Ystring);
+							fprintf (stderr, "%s %s\n", Xstring, Ystring);
```
seisman, line 1300:
```suggestion
```

src/seis/psmeca.c
```diff
@@ -44,27 +44,9 @@ PostScript code is written to stdout.
...
-#define PLOT_TRACE	4
-
 /* Control structure for psmeca */
 struct PSMECA_CTRL {
-	struct PSMECA_A {	/* -A[+p<pen>][+s<size>] */
-		bool active;
-		double size;
-		struct GMT_PEN pen;
-	} A;
+	struct SEIS_OFFSET_LINE A; 	/* -A[+o[<dx>[/<dy>]]][+p<pen>][+s<size>] */
```
seisman, line 49:
```suggestion
	struct SEIS_OFFSET_LINE A; 	/* -A[+g<fill>][+o[<dx>[/<dy>]]][+p<pen>][+s<size>] */
```

seisman:
The codes look pretty good to me. I still need some time for more testings.

PaulWessel:
Thanks for spotting those glitches.

seisman:
Bug in parsing the trailing text:
```
gmt meca meca.dat -Sa1c -JM10c -B -R110/120/30/36 -A+o -png map << EOF
112 31.05  30  90   0  4  2.0 1.0 Strike-slip                                   
115 34 15  30  60  90  5  1.0 1.0 Reverse
118 32 45  30  60 -90  6  1.0 1.0 Normal
EOF
```

PaulWessel:
Will look, but you are giving both a file (meca.dat) and feed from stdin. Intentional?

PaulWessel:
Good test, It tries to convert to a number.  But help me with this:

In record looks like this:

112 31.05  30  90   0  4  2.0 1.0 Strike-slip                                   

and -Sa sets the number of input columns to 7 But you are giving only 6 and then the two extra. So se set GMT_Set_Columns to 7 and then the trailing text is "1.0 Strike-slip".

Is the n_cols settings when parsing 'S' expecting 6 ir 7?

seisman:
> Good test, It tries to convert to a number. But help me with this:
> 
> In record looks like this:
> 
> 112 31.05 30 90 0 4 2.0 1.0 Strike-slip
> 
> and -Sa sets the number of input columns to 7 But you are giving only 6 and then the two extra. So se set GMT_Set_Columns to 7 and then the trailing text is "1.0 Strike-slip".
> 
> Is the n_cols settings when parsing 'S' expecting 6 ir 7?

Sorry, my mistake. It's weird that I lost the depth column when copying and paste the data.

PaulWessel:
Oh, now I see it; the decimals tricked you.  Looks OK?

map

seisman:
Yes, it looks good. Will do more tests.

PaulWessel:
Some issues

using +s and x (cross) requires -W to change pen and +c to set pen color to event depth color. Maybe computer cross pen width from size is more practical. At least as default 
The issue with label justifications and offsets dc day

doc/rst/source/supplements/seis/coupe.rst
```diff
@@ -136,6 +137,27 @@ Optional Arguments
...
+    We will draw a line connecting the original and relocated
+    beachball positions and optionally place a small symbol [circle] at the original
+    location.  Use **+s**\ *size* to set the diameter of the symbol [no symbol].
+    Change from circle to any of the standard geometric symbols in :doc:`plot </plot>`,
+    i.e., **a**\|\ **c**\|\ **d**\|\ **g**\|\ **h**\|\ **i**\|\ **n**\|\ **p**\|\ **s**\|\ **t**\|\ **x** [**c**].
+    The symbol will be filled with the beachball color, but a fixed color (**+g**\ *fill*)
+    or no fill (**+g**) can be selected.
+    The line pen defaults to that given via |-W| but can be overridden
+    by using **+p**\ *pen* [0.25p]. **Note**: If the cross symbol (**x**) is selected there is no fill since it is
+    just two stroked lines. Use |-W| to set line thickness and append **+c** to color the crosses as other symbols.
```
seisman, line 159: >  **Note**: If the cross symbol (**x**) is selected there is no fill since it is
    just two stroked lines. Use |-W| to set line thickness and append **+c** to color the crosses as other symbols.

This sentence is no longer valid after PR https://github.com/GenericMappingTools/gmt/pull/7789 is merged, right?

seisman:
This PR looks pretty good to me.

PaulWessel:
Hi @seisman, I am done with this branch.  All 13 seis scripts passes and unless you see anything funny I hope you can approve it and we can merge.  Then onto stroke-symbol revitalization!

seisman (approved):

PR #7782, opened 2023-09-03
Update gmtspatial.rst

jidanni:
Mention also truncates lines.


PaulWessel:
Code actually reads the input as a polygon, so I assume if a line it will connect first/last and make a bad polygon.  So I cannot accept this one - would require testing and working with lines....

jidanni:
All I know is it works fine with lines.

PaulWessel (approved):
Taking your word on it.

joa-quim:
I remind that each of these absolutetely needless changes have launched a VM somewhere that worked full steam during some half an hour.

PR #7786, opened 2023-09-05
Replace CookBook refereneces with Technical Reference

### LINKED ISSUE #7776: Rename Bookbook to Technical Reference

During the summit we discussed the organisation of the documentation, and one issue is that the file/link called CookBook isn't truthful.  A cookbook as we know it will contain lots of _recipes_, but those are actually in the illustration galleries.

I suggest we just rename the cookbook directory in the docs to techref and then update the word Cookbook (wherever it occurs, such as sidebar, links, etc.) to Technical Reference.

Ifi you agree, might there be an RST expert willing to do this PR? E.g. @GenericMappingTools/gmt-docs ?

PaulWessel:
The Cookbook is actually our tecthnical reference and should be named as such.  Any cooking takes place in the galleries.  Closes #7776.

Fixed CookBook, cookbook, Cook Book etc and other variations, plus /cookbook -> /reference.  Seems to build fine but probably some subtle mistakes.  See what you think.


seisman:
I can still find many "cookbook"s that need changes. I'll push more changes directly to this PR.

PaulWessel:
OK,  thanks. Must be hiding somewhere outside doc then:

```
find . -name '*.rst*' -exec grep -H '/cookbook' {} \; | wc  
0
```

seisman:
@PaulWessel Should `gmt docs cookbook` be changed to `gmt docs reference`?

PaulWessel:
Yes, but we should accept either for backward protection.  I can look at that.

seisman:
@PaulWessel This PR should be good to merge, but please rebuild the docs to see if the docs building is successful. I can't build GMT recently due to my broken Homebrew pacakges.

PaulWessel:
Heads up @joa-quim if your docs specifically looks for stuff. Both gmt docs cookbook and gmt docs reference will work.

Issue #7788, opened 2023-09-05
Treat stroke-symbols as other symbols unless -W is set


PaulWessel:
**Description of the desired feature**

It has annoyed me for some time that using symbols like cross, dash, etc., i.e., just line features, requires **-W** to set the pen thickness, making it very complicated to have variable pen widths.  Also, color is not set by **-G** as for other symbols but taken from the pen.  That color could come from a z/cpt lookup if **-W+c** is given.  Etc.  So while they are symbols they re second-class citizens in many ways.

Because of history an upgrade needs to be backwards compatible.  I think this is:

- If **-W** is used to set pen with constant color or variable via **-W+c** we honour that as the current solution as backwards compatibility.
- If _no_ **-W** then stroke symbol pen width is` f * symbol_size`, and if **-G** is set we set the pen to that color (same if z/cpt lookup)

This would allow `-Sx14p -Gred` to draw a red cross with pen width `f *size` and `-Sx14p -Ct.cpt` will do the same `f * size` but with pen color set via z/cpt.

So what is a reasonable _f_?  Here are three candidates (ignore the other but reported in #7787)  - I think going beyond this range gets odd:  Please give your votes:

**Note:** As they say, _different strokes for different blokes_, one could also entertain a new GMT defaults **MAP_STROKE_WIDTH** that could default to our preferred choice but the users could override it for some project.

![line_symbols_10](https://github.com/GenericMappingTools/gmt/assets/26473567/232c7aad-2281-4bc7-a85d-86a0be2e8fb9)

![line_symbols_15](https://github.com/GenericMappingTools/gmt/assets/26473567/a0c3e64a-5cc8-4da1-9f2c-5f2fa7f1e862)

![line_symbols_20](https://github.com/GenericMappingTools/gmt/assets/26473567/be5036ce-108d-45d2-afbf-e147439c0c33)

PaulWessel:
There is no way to let Sx14p either be a cross with pen = default or pen = 15% of size.  I know it has been like you state for 20 years but that was because we did not want users to plot nothing (no pen no fill) and be confused.  I think we will have to let that go.

-Sx14p -Gred -W1p would be a red, 1p cross. Basically, we are treating all symbols the same way

PaulWessel:
Hei, @joa-quim @anbj @Esteban82 @maxrjones @seisman @remkos : Any opinion on this - otherwise I will

Implement at 15% size and no **MAP_STROKE_WIDTH** or similar.

[...]

Issue #3929, opened 2020-08-11
Creating Custom Pens


PaulWessel:
## Rationale

For symbols we have custom symbols which allows users to create macros that produce any symbol imaginable, with one or more extra data parameters affecting the symbol. We do not have a similar mechanism for lines. One idea therefore is to create a custom pen. The information would be stored in name.pen files, similar to symbol.def files for symbols, and name will be used as the style argument in pen specifications.

![custompen](https://user-images.githubusercontent.com/26473567/89953442-c7c9f680-dbca-11ea-8c03-40f389ea7199.png)

## Syntax

In GMT, a pen is specified via the syntax

[_width_][,_color_][,_style_]

where commas are only used to separate items if more than one is given. All items are optional and have GMT defaults: 0.25p,black,solid. A custom pen will be requested via the style argument in a pen assignment. All custom pens inherit the above defaults if width or color are not set (the style cannot be overridden since it is the name of the custom pen). A given pen width will be distributed proportionally since custom pens must use nondimensional widths (and dash lengths). To override custom pen colors a series of colors may be provided to change them all, separated by + characters. A missing color means keep the default color.

## User-level examples

Here is a set of syntax examples as to how users may specify such custom pens:

```
-Wrailroad
-W4p,ridge
-Wred+green,border
-W6p,black+red+yellow,route
-W1p,@path
```

The last example gets the custom pen path from the GMT cache server and scales it to 1p.

## Implementation

The name.pen file format consists of an optional header with comments, followed by two or more nondimensional pen specifications. The pens will be applied in the order given, hence the order is important. Here is an example of a custom pen used to draw railroad tracks:

```
# Name: railroad
# Purpose: Draw a line that looks like a checkered railroad line.
# Lay down a black solid line of unit width
1,black,solid
# Overlay a white thinner dashed line
0.6666666666,white,5_5:0
```

Note the following:

1. The actual pens are given in multiples of unit pen widths and are normally scaled by the width given on the command line [0.25p].
2. The assigned color of the pens may be overridden on the command line [black]
3. Any dashes, gaps, offsets must all be given in nondimensional and scaled up accordingly.
4. To simplify parsing we specify all of width,color,style even if black and solid.

[...]

PR #7789, opened 2023-09-05
Treat stroke symbols like other symbols

PaulWessel:
Refer to #7788 for background.  This WIP PR covers **psxy** and seems to work as explained.  Will do the same for **psxyz**. Only unsettled issue is default and/or variable scaling from symbol size to pen width.


joa-quim (approved):
Almost blind approve. Will try to look better when finish to fix damn convoluted GMT<->GDAL mem layout shits in GMT.jl

... and also bring back the ``Custom pens issue`` #3929

Esteban82 (approved):

PaulWessel:
Unless objections the new 30 minutes, I will merge this branch into master. I am then down to 45 failures on macOS, most involving grid images but there are a few that probably are odd roundoff.

PR #7793, opened 2023-09-07
Add condition to build the texture shading code in the GMT lib.

joa-quim:
Don't know if it worth to doc this in ``ConfigUserTemplate.cmake``

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -377,6 +377,12 @@ set (GMT_MBSYSTEM_SRCS gmt_mbsystem_glue.c)
 
 AUX_SOURCE_DIRECTORY (longopt GMT_LONG_OPT_H)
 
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Do we really need yet another FFT package for this? We already have kiss, old Fortran-to-C versions and macs have Accelerate.

joa-quim:
From the time I read the code I have the impression that he called FFT to a cosine transform. But, yes, this something to look at (it was not an obvious replacement otherwise I would(?) have done it). Specially because this C code is not particularly fast. It takes ~8 seconds here to run Kristof's larger grid example.

PaulWessel:
OK, once we get 6.5 out we can consider that. I need 6.5 out so I can start working on Roger's long option branches.

PR #7794, opened 2023-09-08
Add admin script to explore failure distribution

PaulWessel:
This PR only adds a new admin/examin-failure-log.sh script that developers can use after running all the tests (we assume a _build_ or _rbuild_ directory in the top gmt directory). We hvae a range of failures that depends on at least two things:

- OS architecture (i.e., Apple arm64 vs Intel x86_64 etc.)
- What version of remote data sets (e.g., @earth_relief_02m) are actually accessed.

Below is my result on Mac arm64 running on master, where I get 84 failures:

rms-errors

Some notes:

1. The known failures do not even run the script so I just plotted them as blue dots at some high rms value
2. We have a range of passes (gray zone) that do so because the GRAPHICSMAGICK_RMS were adjusted individually for those scripts.
3. Over half the script pass with zero rms difference with the original

I would like other @GenericMappingTools/core to quickly approve this PR (since it does not affect any code) and then give it a try (from this branch or as soon as approved and merged).

### DIFF
--- a/admin/examine-failure-log.sh
+++ b/admin/examine-failure-log.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+#
+# Script that can be run after full testing to examining the RMS
+# failures.  We assume this script is run from the top GMT directory
+# and that there is a build or rbuild directory there.
+BLDDIR=rbuild
+if [ ! -d admin ]; then
+	echo "examine-failure-log.sh: Must be run from top-level gmt directory" >&2
+	exit 1
+fi
+if [ ! -d ${BLDDIR} ]; then
+	BLDDIR=build
+	if [ ! -d ${BLDDIR} ]; then
+		echo "examine-failure-log.sh: Neither build nor rbuild directory found" >&2
+	exit 1
+	fi
+fi
+
+# 0. Change into build directory
+
+cd ${BLDDIR}
+
+# 1. Get one-record-per test results
+
+grep "RMS Error" Testing/Temporary/LastTest.log | egrep -v "Thread" | sed -e 'sBN/AB-1Bgp '> /tmp/raw.log
+
+# 2. Extract info from raw log
+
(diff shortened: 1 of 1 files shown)


joa-quim (approved):

PaulWessel:
Updated to leave the sorted rank, rms, script name in _file failure-log.txt_.

PaulWessel:
Googling about about arm vs x8086 and rounding, some users reported using a few compiler flags that seems to help: 

`-ffp-contract=off -fno-fast-math`

With these I get fewer failures (29 additional scripts pass with zero rms difference - including many that passed earlier but not at zero rms) and a total of 21 more tests passed:

rms-errors

Could someone on Intel (or @seisman from CI) send me the full Testing/Temporary/LastTest.log file.  This is using master.

seisman:
The master branch has 46 failures on macOS, but please note that the macOS CI is using gs 9.56.1.

The [LastTest.log](https://github.com/GenericMappingTools/gmt/files/12565276/LastTest.log) is attached. The log file only contains the 46 failures, because in CI we rerun the failed tests multiple times.

PaulWessel:
Thanks. I am curious about running those we have flagged as KNOWN_FAILURES.  Is there a simple way to activate the others than commenting out that seeing in each of those scripts?

seisman:
The `cmake/ConfigUserAdvancedTemplate.cmake` files says you can `set (GMT_ENABLE_KNOWN2FAIL OFF)`, but I'm not sure if it works.

PaulWessel:
Thanks, will give it a try.  We probably should revisit those failures more often, maybe a reminder every 6 months or so?

seisman:
> Thanks, will give it a try. We probably should revisit those failures more often, maybe a reminder every 6 months or so?

Sounds a good idea. I've opened an issue and will work on it later.

joa-quim:
I got 34 failures that include old-never-resolved shits of msys screwing the custom symbols paths and 3 latex failures.
My build dir is detached from gmt source.

PR #7798, opened 2023-09-09
Add admin script that determines which remote data sets are used

PaulWessel:
Brute force search for all available data sets (e.g., @earth_age_03m_g) in all our doc and test scripts.  Currently finds 35.  No effect on GMT or tests so can be approved and merged - improvements are appreciated - it runs a bit slow.  Script is `admin/remote-files-used.sh`.


seisman:
> it runs a bit slow.

If we assume that all remote datasets used in docs/tests start with `@earth_`, then we can run：
```
$ grep '@earth_' test/**/*.sh doc/scripts/**/*.sh | awk -F'@earth_' '{print $2}' | awk '{printf("@earth_%s\n", $1)}' | sort -u
@earth_age
@earth_age_02m_p
@earth_age_06m_p
@earth_age_10m_p
@earth_day
@earth_day_01d
@earth_day_01m_p.tif
@earth_day_15m
@earth_mask
@earth_night
@earth_relief
@earth_relief_01d
@earth_relief_01d_g
@earth_relief_01m
@earth_relief_01m_p
@earth_relief_02m_p
@earth_relief_03s
@earth_relief_04m_p
@earth_relief_05m
@earth_relief_05m_g
@earth_relief_05m_p
@earth_relief_06m_p
@earth_relief_10m
@earth_relief_10m_g
@earth_relief_10m_p
@earth_relief_15m
@earth_relief_15m_p
@earth_relief_20m
@earth_relief_20m_g
@earth_relief_20m_holes.grd
@earth_relief_30m
@earth_relief_30m_p
@earth_relief_30s
@earth_relief_30s_p
```

PaulWessel:
Yours is better - will update soon, and grep away the _holes file

PaulWessel:
OK, works fast and skips the one _hole file.
I will update the two scripts using earth_day_01m_p.tif since they work just as well with earth_day_01m in a separate branch

seisman:
I just found that we already have the `admin/remote_data_use.sh` script, which does similar job.

seisman:
> I just found that we already have the `admin/remote_data_use.sh` script, which does similar job.

The new script is simpler and faster. So please remove the old `admin/remote_data_use.sh` script.

seisman (approved):

PR #7810, opened 2023-09-12
Update WL_example_3.sh to use NETCDF driver

PaulWessel:
Still does not work but at least it opens the file correctly.  More work needed, @joa-quim.

### DIFF
--- a/test/gmtmex/WL_example_3.sh
+++ b/test/gmtmex/WL_example_3.sh
@@ -9,7 +9,7 @@ gmt begin WL_example_3 ps
   # Import sea surface temperature grids from several HDF5 layers (lon, lat, sst, sst_qual)
   # Speed up processing by using native binary intermediary files
   file="A2016152023000.L2_LAC_SST.nc"
-  args="=gd?HDF5:A2016152023000.L2_LAC_SST.nc"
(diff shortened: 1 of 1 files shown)


joa-quim:
The failing case is not our fault at all. If, in `gmt_gdalread`, we change the two occurrences to
```
switch (GDALGetRasterDataType(hBand)) {
	case GDT_Byte:
	case 14:
```

than it works.

I wonder if we should not add this ugly hack since we do not have code to read complex grids from GDAL.

joa-quim (approved):

PaulWessel:
So the data are complex in this case? Or do you worry we will crash on such a grid (which not that many would have, no)?

joa-quim:
No, data is not complex but is misinterpreted to be so. The worry, is if we pass in a true complex layer and that change will make it seem as if it is a char. But that case would fail anyway because we don't (?) deal with complex grids read by GDAL:

PaulWessel:
Yes, more important that a published WL_example_3 (presumably both .m and .sh) works again

joa-quim:
The may work but result is wrong. Mirone is using an older GDAL and can read that layer just fine.

Issue #6736, opened 2022-05-22
Backticks vesus $(...) in shell scripting


PaulWessel:
**Description of the desired feature**

Please see this [StackExchange](https://unix.stackexchange.com/questions/126927/have-backticks-i-e-cmd-in-sh-shells-been-deprecated#:~:text=In%20csh%2Ftcsh%2C%20the%20backticks,ever%2C%20for%20that%20matter).) note for full background.

Basically, the argument is that we should stick to `$(..)` as the recommended way to run a subshell and store the value in a variable.
There are many good arguments for this (see above).

The only argument against is that` $(...)` is not a known thing in csh/tcsh and apparently there are still folks out there using those shells.  They have back-ticks for this only.  Also, note backticks are not going away so it is not a transition thing.  Finallyl, there is nothing like this in DOS batch scripts I believe.

Please make sure you read the StackExchange summary before you voice your (now) informed opinion on this.  I was about to make a PR to replace all occurrences of this in scripts and docs but decided to go about it in this way first.

anbj:
I replaced alot of these in #2595.

It's too bad it's not supported by csh/tcsh. Just out of curiosity; how many users are complaining abour this?

See also [this post on stackoverflow](https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts). `$()` is POSIX; *$() is POSIX and supported by all modern Bourne shells, e.g. ksh, bash, ash, dash, zsh, busybox, you name it.*

But csh/tsch not listed.

Esteban82:
OK, we leave the docs. 


I had to use double backticks. With <> I get this: 

`gmt plot `gmt info -I5 -D1 @ship_15.txt` @ship_15.txt -B -Sc2p -pdf map`

PaulWessel:
AH, I see.  Interesting failure,  maybe there is an escape sequence to use instead.  Not important.

Esteban82:
> AH, I see. Interesting failure, maybe there is an escape sequence to use instead. Not important.

Yes, 2 backticks instead of 1.

seisman:
Perhaps we can try [the `tab-set` directive](https://sphinx-design.readthedocs.io/en/latest/tabs.html).

For commands that need backticks/`$()`, we can show two tabs, one for bash (using `$()`), another one for csh/tcsh (using backticks).

seisman:
I think we should use `$()` consistently in all scripts. Here are my reasons：

[...]

PR #7817, opened 2023-09-15
Replace backticks with $() in shell scripts

seisman:
Closes https://github.com/GenericMappingTools/gmt/issues/6736.


PaulWessel (approved):
Looks good.  Most of these are old scripts inherited from user reports and predates our own switch to `$()`.

PaulWessel:
Noticed this in gmtinfo.rst:

gmt plot \`gmt info -I5 -D1 @ship_15.txt\` @ship_15.txt -B -Sc2p -pdf map

Maybe need to look at docs too?

PR #7821, opened 2023-09-20
Fix dumb bug in -x parser

PaulWessel:
I tried (perhaps for the first time) **-x**-1 which is to give me all cores minus 1, in my case 10 -1 = 9, but I got 11 (!) since the code subtracted the -1 from 10.... You gotta watch out for me lack of coffee - it does strange things.


joa-quim:
> , but I got 11 (!) since the code subtracted the -1 from 10.... You gotta watch out for me lack of coffee - it does strange things.

I'm probably the one to blame, since those negative numbers were my idea (if I still can recall).

anbj:
> I tried (perhaps for the first time) **-x**-1 which is to give me all cores minus 1, in my case 10 -1 = 9, but I got 11 (!) since the code subtracted the -1 from 10.... You gotta watch out for me lack of coffee - it does strange things.

At least you get the job done helluva fast.

PR #7822, opened 2023-09-20
Use MAP_SYMBOL_PEN_SCALE instead of MAP_STROKE_WIDTH

### LINKED ISSUE #7788: Treat stroke-symbols as other symbols unless -W is set

PaulWessel:
See #7788 for discussion at end.  Note the internal variable is _map_stroke_width_. Two questions:

1. Rename _map_stroke_width_ to _map_stroke_factor_
2. Consider SCALE instead of FACTOR?

Not sure what is best but will go with the flow of comments. Closes #7788.

### DIFF
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -90,7 +90,7 @@
     **x**, **y**, **+** and **-** will derive their pen widths from the symbol size
     and their pen color from |-G| or |-C|. In that sense they behave as the other
     symbols (but no outline).  You can override this behavior by specifying |-W|,
-    modify it by adjusting :term:`MAP_SYMBOL_STROKE_FACTOR` [15%], or set it to zero to
+    modify it by adjusting :term:`MAP_SYMBOL_PEN_SCALE` [15%], or set it to zero to
     rely on pen defaults (and |-W|).
 
     The next collection shows five symbols that require two or more parameters, and
--- a/doc/rst/source/explain_symbols.rst_
+++ b/doc/rst/source/explain_symbols.rst_
@@ -90,7 +90,7 @@
     **x**, **y**, **+** and **-** will derive their pen widths from the symbol size
     and their pen color from |-G| or |-C|. In that sense they behave as the other
     symbols (but no outline).  You can override this behavior by specifying |-W|,
-    modify it by adjusting :term:`MAP_STROKE_WIDTH` [15%], or set it to zero to
+    modify it by adjusting :term:`MAP_SYMBOL_STROKE_FACTOR` [15%], or set it to zero to
     rely on pen defaults (and |-W|).
 
     The next collection shows five symbols that require two or more parameters, and
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -868,16 +868,16 @@ MAP Parameters
         cap is reversed, i.e., *85/90* will draw a r = 5 radius circle at the
         center of the map with less frequent radial lines there.
 
-    **MAP_STROKE_WIDTH**
+    **MAP_SCALE_HEIGHT**
+        Sets the height (> 0) on the map of the map scale bars drawn by
+        various programs [default is **5p**].
+
+    **MAP_SYMBOL_STROKE_FACTOR**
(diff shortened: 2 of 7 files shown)


remkos:
I'm fine with FACTOR.

remkos (approved):

joa-quim:
Can we drop the `STROKE`? The variable name is very long and will look ugly if used in command with the ``--`` mechanism. And many may think the symbol had and hart attack 😄

remkos:
> Can we drop the `STROKE`? The variable name is very long and will look ugly if used in command with the `--` mechanism. And many may think the symbol had and hart attack 😄

It's true that this would be getting long. But `MAP_SYMBOL_FACTOR` is not very clear, is it? It would suggest some dependent scaling of the whole symbol. If you want to avoid STROKE, it ought probably be `MAP_SYMBOL_PEN_FACTOR`.
But in that case, I start to feel with @PaulWessel that it should be `MAP_SYMBOL_PEN_SCALE`.

Look @joa-quim : I saved just 3 characters 😄. It is shorter than many of the others that we have, e.g.: `MAP_GRID_CROSS_SIZE_SECONDARY`

joa-quim:
I prefer `MAP_SYMBOL_PEN_SCALE` to `MAP_SYMBOL_PEN_FACTOR`, and I definitely prefer `PEN` to `STROKE` and not because it's 3 chars shorter but because the meaning is much more clear (no coronaries involved).

seisman:
Do we really need the new default MAP_SYMBOL_STROKE_FACTOR? It's easy to change the scale via the **-W** option so I feel the new default will be rarely used.

remkos:
@seisman : You can set the pen width with -W, but that would then apply to any size. The thing is that Paul wanted to assure that if someone did not supply -W, some pen width would be set automatically. For example `-Sx30p -Gred` should automatically result in `-Sx20p -Gred -W3p`, assuming a scale factor of 0.15 (15%).

But in any case you are correct: it would be a barely used setting.

PaulWessel:
I aded WIP until we have a solution.  Just so we are on the same page.  Unless MAP_SYMBOL_STROKE_FACTOR is set to 0, we do

`GMT->current.setting.map_stroke_width * S.size_x * PSL_POINTS_PER_INCH`

So the pen thickness does not enter I think, @seisman ? The whole point was to scale the pen to a fraction of symbol size.  Adding -W4p makes all symbols fat.

seisman:
> The whole point was to scale the pen to a fraction of symbol size. Adding -W4p makes all symbols fat.

That makes sense to me now.

> If you want to avoid STROKE, it ought probably be `MAP_SYMBOL_PEN_FACTOR`

My first impression about `MAP_SYMBOL_PEN_FACTOR` is that it is applied to all symbols like circles.

PaulWessel:
Must say **MAP_SYMBOL_PEN_SCALE** is not bad: Avoids cardiac arrests and shaves a few letters.

remkos:
I agree: RTFM.

PR #7827, opened 2023-09-22
Allow grdimage to process byte images without indexed colors, via CPT

PaulWessel:
See [forum](https://forum.generic-mapping-tools.org/t/thematic-mapping-of-land-cover-by-pygmt/4271/20) discussion for background. This PR is almost there - 45 minutes of work is not bad. Perhaps @joa-quim knows what to improve.  There are an issue to resolve:

The image has values 0-8 for a total of 9, but CPT has 8. Since these are "keys" to categories 1-8 there are only 8 but GMT does not know the image contains categories and not index into a color table. I inserted a 0 row (white) to make this work for now.

Perhaps a 1-byte image without color table is expected to hold categories? If so we can subtract one and be OK

Should not be too hard to fix this but want @joa-quim input on this.  See new renderer _grdimage_img_byte_index_.

t

PR #7828, opened 2023-09-22
Take into account the value reported by GDAL for noData in Images

joa-quim:
We were ignoring this value before (in images) so it shouldn't break anything.


PaulWessel (approved):

PaulWessel:
Merged into #7827

PaulWessel:
I can look at this tomorrow unless you want to finish tonight.

PR #7829, opened 2023-09-23
CI: Use conda to install ghostscript/ninja/dvc/sphinx on Windows

seisman:
Use conda to install GMT dependencies on Windows. The main benefit is that now gs is installed via conda on all three operating systems, so it's much easier for us to have the same gs version on the three platforms.


seisman:
Ping @PaulWessel

PaulWessel (approved):

PR #7831, opened 2023-09-23
Default to color bars with gaps for categorical CPTs

### LINKED ISSUE #7830: Automatically select -Lgap layout in colobar if categorical CPT
Currently, if you plot a color bar with a categorical CPT:

`gmt colorbar -Cclasscolor.cpt -Dx4i/4i+w5i+h -png a`

you will by default get this:

a

Not liking this, we usually decide to add **-L**_gap_:

`gmt colorbar -Cclasscolor.cpt -Dx4i/4i+w5i+h -L6p -png b`

b

But, since categories are never continuous we should automatically detect that and behave as if **-L** was set.  Changing this would be a "bug fix" in the sense we cannot secure backward compatibility.  Alternatively, we invent a modifier **+c** for **-L** that means "If CPT is categorical, auto-apply **-L**, else ignore it". For out one-liner plots we would certainly then add **-L+c** to the internal call. Also if gap is not set we should have a default, e.g., The `n-1 `gaps should add to 15% of the bar width (for example, here it is ~15%).

Comments?

PaulWessel:
See #7830 for background.  This PR implements the change.  Old (and in our opinion wrong) behavior can be retained by just giving **-L** with no or zero gap.

Part 1 (implementation) is done
Part 2 (hence WIP) is to update some scripts that currently plot no gaps.  I will let a few of those add **-**L so the PS does not change but at least one should give the new behaviour and I will dvc that plot only.

Closes #7830.

### DIFF
--- a/doc/rst/source/colorbar.rst
+++ b/doc/rst/source/colorbar.rst
@@ -183,7 +183,10 @@ Optional Arguments
     annotation on each rectangle, using the lower boundary z-value for
     the annotation. If **i** is prepended we annotate the interval range
     instead. If |-I| is used then each rectangle will have its
-    constant color modified by the specified intensity.
+    constant color modified by the specified intensity.  **Note**: For
+    categorical CPTs we default to activating |-L| with a *gap* such
+    that the sum of all the gaps equal 15% of the bar width.  You may
+    chose no gaps by giving |-L| only or explicitly set *gap = 0*.
 
 .. _-M:
 
--- a/src/psscale.c
+++ b/src/psscale.c
@@ -826,6 +826,11 @@ GMT_LOCAL void psscale_draw_colorbar (struct GMT_CTRL *GMT, struct PSSCALE_CTRL 
 	unsigned int dump_k_val = 0;
 #endif
 
+	if (P->categorical && !Ctrl->L.active) {	/* For categorical CPTs the default is -L<gap> with sum of all gaps = 15% of bar length  */
+		Ctrl->L.active = true;
+		Ctrl->L.spacing = 0.01 * PSSCALE_GAP_PERCENT * Ctrl->D.dim[GMT_X] / (P->n_colors - 1);
+		fprintf (stderr, "Gap set to %lg\n", Ctrl->L.spacing);
+	}
 	max_intens[0] = Ctrl->I.min;
 	max_intens[1] = Ctrl->I.max;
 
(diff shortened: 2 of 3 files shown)


joa-quim (approved):

seisman:
It's likely this PR breaks `GMT_App_M_1c`:

GMT_App_M_1c

PaulWessel:
Sure, can see why. But what is a good solution here? This is a page of all categorical CPTs but we treat them as any other CPT. Perhaps it _should_ look different since it is categorical and the script has to adjust. In M_1a we have  the paired.cpt who is the same kind, but it is also placed in M_1c so we should remove it so it does not show twice.  For all of those the "top continuous, bottom discrete" makes no sense anyway so best to just show those in a separate (new) format. Agree? Thoughts on this, @joa-quim ?

PaulWessel:
@Esteban82, could you please send me the Rust to Silver CPT from Mercury(?). I cannot find it but I know I kept it....

joa-quim:
I have never played with the categorical before and so I'm in the learning process. My doubt is even more basic. What are those SCM categorical cpts good for? They look absolutely useless to me. Who wants 100 classes? And if we interpolate them to small number (8 in the bottom palettes) the result is garbage with several classes with indistinguishable colors.

PaulWessel:
All true. I am reducing to 25 cases and 12 for our paired.cpt.  We are just passing SCM to users who hate rainbows and turbo.

joa-quim:
Right for the SCM's, but the SCM's-S (the categoricals) are garbage to me (until I'm instructed on its validity) and we should simply drop them, as well as the `GMT_App_M_1c` figure.

PaulWessel:
No reason to do so.  It does let people see them.

joa-quim:
They can be see in ``GMT_App_M_1b``. If they are **absolutely useless** as, until further proof, is the case there is no reason to show them. On the contrary, there is a compelling reason to **not** show them.

PaulWessel:
I sense your anger so WTF, we will trash 1c. I am angry too, spent 90 minutes on fixing that plot!

joa-quim:
You should know that Sunday lunches here last for long. You should only asked my thoughts only after I had time to do a little siesta 😄

Esteban82:
> @Esteban82, could you please send me the Rust to Silver CPT from Mercury(?). I cannot find it but I know I kept it....

Hi Paul. Here is the CPT (I add the extension .txt so I can upload it here). [RustToSilver_180207.cpt.txt](https://github.com/GenericMappingTools/gmt/files/12710553/RustToSilver_180207.cpt.txt)


I used this script to create it from the original CLR file. If you want to use it for pluto, them you only will have to adjust the values (I think).

```
CLR=RustToSilver_180207.clr
CPT=$(basename $CLR .clr).cpt

# 1. Get values from CLR file
gmt convert $CLR -hi1+d -o0  > values.txt

# 2. Get colors (RGB values) from CLR file
gmt convert $CLR -hi1+d -o1-3  | sed 's/\t/\//g' > colors.txt

# 3. Rearrenge boths lists
colors=$(cat colors.txt | tr -s "\n\r" "," | sed 's/,$//')
values=$(cat values.txt | tr -s "\n\r" "," | sed 's/,$//')

# 4. Create continuos CPT
gmt makecpt -C$colors -T$values -Z > $CPT
```

PR #7835, opened 2023-09-23
Read and write a new "crs_wkt" attribute which is equal to "spatial_ref"

joa-quim:
GDAL is threatening that in its future4.0 version the netCDF attribute ``spatial_ref`` will be dropped because the standard should now be ``crs_wkt``. This PR adds reading and writting that attribute too. However, although tests are unaffected, I'm not sure that this will completely solve the trouble of the future 4.0. Looking at a GDAL generated nc they use a global variable called ``crs`` while we use one called (like it was called for eons in GDAL) ``grid_mapping``. I guess we'll have to wait and see if future still bring a Linux beloved breaking in backward compatibility or if this is enough to keep it working in-out with GDAL. I suspect, however, that other non GDAL based may have troubles to recognize the referencing information in nc grids created by GMT.

### DIFF
--- a/src/gmt_nc.c
+++ b/src/gmt_nc.c
@@ -641,6 +641,14 @@ GMT_LOCAL int gmtnc_grd_info (struct GMT_CTRL *GMT, struct GMT_GRID_HEADER *head
 			header->ProjRefWKT = strdup (attrib);	/* Turn it into a strdup allocation to be compatible with other instances elsewhere */
 			gmt_M_free (GMT, attrib);
 		}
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Are you sure of the orders of lines 1007-1008? I would have thought you'd do the new one first and if still blank you do the older ones.

joa-quim:
Lines 1007-8 are for writing so the order shouldn't matter (I think). That could be raised on reading but given that we are mostly reading nc's that we have produced, starting by the old one seemed a tinny bit more efficient.

PR #7843, opened 2023-09-24
Remove the restriction that -C & -D cannot be used together.

joa-quim:
Since the grdimage with images accepts -C this is no longer true. However, my quick trials with Julia (where I'm forced to set -D when passing in an image) shown that in that case the cpt is ignored.


PaulWessel (approved):
I assume no tests get broken by this

PR #7845, opened 2023-09-25
Allow building GMT documentation without compiling GMT

seisman:
Currently, building the GMT documentation requires compiling GMT first, because we need `gmt` to convert PS images to PNG/PDF. However, as GMT developers and contributors, we definitely already have `gmt` installed in our PATH (either a stable version or a dev version). So, why not just call the `gmt` executable in the **PATH**?

I found the cmake option `GMT_DOCS_DEPEND_ON_GMT` which makes the `docs_html` target independent of the `gmt` executable. However, it still requires us to build the `gmt` executable manually (via `ninja gmt`) before building the documentation (via `ninja docs_html)`. It's useful if you already have the `gmt` executable in the build directory and want to update and build docs, but `ninja gmt` should be very fast in this case because there are no code changes. So, the `GMT_DOCS_DEPEND_ON_GMT` option seems useless to me.

This PR changes the behavior of the `GMT_DOCS_DEPEND_ON_GMT` cmake option. If `GMT_DOCS_DEPEND_ON_GMT=FALSE`, cmake will call the `gmt` executable in your **PATH** to do the PS conversion. Thus, even in a clean build directory, you can still build the GMT documentation without compiling GMT. I.e., the following commands will build the documentation only:
```
mkdir rbuild
cd rbuild
cmake .. -G Ninja
ninja docs_html 
```

I think the new behavior makes more sense. What do you think?

### DIFF
--- a/cmake/ConfigUserAdvancedTemplate.cmake
+++ b/cmake/ConfigUserAdvancedTemplate.cmake
@@ -219,8 +219,8 @@ set (GMT_ENABLE_OPENMP TRUE)
 #set (GMT_RELEASE_PREFIX "release-src-prefix")
 
 # If set to false, image conversion from PS images to PNG and PDF does
-# not depend on the gmt binary target. Note: "make gmt" is then required
-# before docs_depends [TRUE].
+# not depend on the gmt binary target. It assumes that you already have the
+# gmt executable in your PATH [TRUE].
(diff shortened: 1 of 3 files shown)


joa-quim (approved):
I like it for another reason. This will allow having a build dir only for the docs and hence the ``ninja install`` will spend more time updating the docs (I mean checking if copying the files is needed or not) then to compile GMT.

Issue #7227, opened 2023-02-03
Specify grid registration in GMT examples/animations


maxrjones:
**Description of the desired feature**

Since GMT 6.4 (https://github.com/GenericMappingTools/gmt/pull/6319), the default grid registration for remote files are as follows:

    1. All plotting modules: pixel-registration
    2. All processing modules: gridline-registration

This causes problems in some of the animation scripts that create an illumination file (processing=gridline) that will be applied to a relief map (plotting). The solution is to specify the registration for all the remote files used in the [examples](https://github.com/GenericMappingTools/gmt/tree/master/doc/examples).

**Are you willing to help implement and maintain this feature?** I can review, but would really appreciate help!

joa-quim:
If someone wants to do this, please do.

PR #7850, opened 2023-09-29
Specify grid registration in example scripts

seisman:

Address https://github.com/GenericMappingTools/gmt/issues/7227

### DIFF
--- a/doc/examples/ex08/ex08.sh
+++ b/doc/examples/ex08/ex08.sh
@@ -7,7 +7,7 @@
 #
 gmt begin ex08
 	gmt makecpt -Ccubhelix -T-5000/0
-	gmt grd2xyz @earth_relief_05m -R0/5/0/5 | gmt plot3d -B -Bz1000+l"Topography (m)" -BWSneZ+b+tETOPO5 \
+	gmt grd2xyz @earth_relief_05m_g -R0/5/0/5 | gmt plot3d -B -Bz1000+l"Topography (m)" -BWSneZ+b+tETOPO5 \
 		-R0/5/0/5/-5000/0 -JM12c -JZ14c -p200/30 -So0.0833333ub-5000 -Wthinnest -C -i0:2,2
 	echo '0.1 4.9 This is the surface of cube' | gmt text -JZ -Z0 -F+f24p,Helvetica-Bold+jTL -p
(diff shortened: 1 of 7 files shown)


PaulWessel (commented):
Agree with the non-plotting modules (grdcut, edit etc) but grdimage? It defaults to pixel registration if now given so I dont think we want to complicate the users most likely interaction which is to make plots.  People who need to cut and process should know what they are doing.

joa-quim:
I go even further. What we should have is that annoying warning about *since you didn't specify registration I'm defaulting to ...* moved into the ``info`` verbose level. Most people don't know, or don't care (and the truth is, do we as users?) what we are talking about. If that's ok for them just let go and don't complicate the system.

PaulWessel:
I am OK with muting that annoying warning.  If you dont know what you are doing then adding -V is a reasonable thing to be optional.  I will take care of it.

PaulWessel:
So we have this:

```
is_PS = gmtinit_is_PS_module (API, mod_name, keys, options);	/* true if module will produce PS */
if (!is_PS) {	/* Override API default since module is a data processor */
	API->use_gridline_registration = true;
	API->use_gridline_registration_warn = (strcmp (mod_name, "grd2cpt") && strcmp (mod_name, "grd2kml"));	/* Give warning unless it is these two clowns */
}
```

so that warning only comes when you are doing data processing, in which case we felt the user need to know what they are doing and actually specify the registration.  This is different from making just a map.  Have you changed your mind? Don't you care if you get p or g when you do stuff with non-plotters?

joa-quim:
My stanza on it is *if you care than be explicit*. But note that many data processing steps are nothing but previous steps for making an image. I GMT.jl I had to add this condition to be able to read ``@earth_`` files without having to remember every time that I had to add that ``-g``.
https://github.com/GenericMappingTools/GMT.jl/blob/master/src/gmtreadwrite.jl#L125

PaulWessel:
OK, find I will make it INFORMATION. Not bothered either way.

PaulWessel:
Now in master

joa-quim:
Nice, and I now go catch a train.

seisman:
I think we can close this PR now.

PR #7855, opened 2023-10-02
Prevent pstext crash is given stoopid things on command line

PaulWessel:
An unnamed rookie tried this:

`gmt pstext '@:20.0:@;70/134/251;A@;;@::@:10.0:@;162/252/60; @;;@::@:20.000000000000004:@;229/70/11;B@;;@::' -R0/5/0/5 -JX6i -F+fHelvetica-Bold+cMC -P -B > t.ps`

and got a "remote file not found" error.  To avoid a lengthy argument I now check that a file cannot have that format, and now we get

`gmt pstext '@:20.0:@;70/134/251;A@;;@::@:10.0:@;162/252/60; @;;@::@:20.000000000000004:@;229/70/11;B@;;@::' -R0/5/0/5 -JX6i -F+fHelvetica-Bold+cMC -P -B > t.ps gmt [ERROR]: File @:20.0:@;70/134/251;A@;;@::@:10.0:@;162/252/60; @;;@::@:20.000000000000004:@;229/70/11;B@;;@:: is not a file and looks like pstext strings.`

PR #7857, opened 2023-10-02
Increment in -T seen as zero if < 1e08

### LINKED ISSUE #7856: makecpt fails on small increments
Small increments (<~1e-8) cause makecpt to fail with the message
```
makecpt [ERROR]: Option T: Increment is zero
```

Example that fails:
```
 gmt makecpt -D -Z -Cjet -T-1.59947e-08/1.59947e-08/3.554379945e-09 >test.cpt
```
Example that succeeds (previous times 10): 

```
 gmt makecpt -D -Z -Cjet -T-1.59947e-07/1.59947e-07/3.554379945e-08 >test.cpt
```

* Operating system: SLES 15.1 (kernel 4.12.14-197.108)
* GMT version: 6.4.0

PaulWessel:
We had too hash **-T**_min/max/inc_ checks, as pointed out in #7856. We used _gmt_M_is_zero_ indiscriminantly (e.g., on the increment) causing small but not EPS-like increments to fail as "zero".  This PR addresses that and closes #7856.

Unrelated, I also found the issue at the end of #7855 where we checked if an argument was a remote file even though the option was actually set, in this case` -l"string"` for a legend.  Clearly, "string" should not be checked for being a remote file.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -15392,6 +15392,7 @@ struct GMT_CTRL *gmt_init_module (struct GMTAPI_CTRL *API, const char *lib_name,
 				gmt_parse_common_options (API->GMT, "V", opt->option, opt->arg);
 		}
 		for (opt = *options; opt; opt = opt->next) {	/* Loop over all options */
+			if (opt->option != GMT_OPT_INFILE) continue;	/* Only check command line input files */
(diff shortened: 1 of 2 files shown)


joa-quim (approved):
Ah, so my rookie junk was in fact a *fuzzy test*

Issue #6257, opened 2022-01-23
Add grdclip functionality to grdmath


anbj:
Why is the functionality in [grdclip](https://docs.generic-mapping-tools.org/dev/grdclip.html) not included (or exclusively) in [grdmath](https://docs.generic-mapping-tools.org/dev/grdmath.html)?

Today I had to do some calculations with *grdmath* and then set all cell values < 0 to *NaN*. To do this I had to:

```
grdmath ........
grdclip -Sb0/NAN .........
```

Ideally, this should be possible to do with a single call to *grdmath*.
In my example only `-Sb` from *grdclip* was used, but all the options in *grdclip* could be included in *grdmath*.

PaulWessel:
Because **grdclip** predates **grdmath**.  But I agree we could add those features to **grdmath** via new operators, e.g.,

_low replace_ **CLIPL** (**-Sb**)
_high replace_ **CLIPH** (**-S**a)
_low high replace_ **RANGEFILL** (**-Si**)
_value replace_ **REPLACE** (**-Sr**)

Since for backwards reasons we cannot remove grdclip we now would have to keep grdclip around but perhaps

1. Not advertise it (remove from docs and examples)
2. Replace it with short program that calls grdmath instead.

anbj:
That explains it; I was wondering why *grdclip* existed - all of this could(/should) be done in *grdmath*.

I think your suggestions are good.

joa-quim:
``grdmath`` is an impossible module to wrap. But no problems making ``grdclip`` a mock-up module. One of these days I'll have to implement it directly in Julia, which will be much more efficient (no need to care with the ``pad`` and no memory copies).

PaulWessel:
I think pyGMT also are not able to wrap grdmath and gmtmath.  But we did so in GMT/MEX but of course there we have the text command to work with.

joa-quim:
Wrapping is not a problem (Julia has it). The problem is to give it a friendly syntax.

Esteban82:
Just a question. With `grdmath` it would be possible to reclassify several ranges of values like this:  

`gmt grdclip classes.nc -Gnew_classes.nc -Si25/30/99 -Si35/39/55 -Sr17/11 -Sb10/0 -V`

PaulWessel:
That is correct, but that is what grdclip allows as well (-Sr and -Si are repeatable options in grdclip).

PaulWessel:
[...]

PR #7858, opened 2023-10-02
Enhance grdclip -Sa -Sb with modifier +e for equality

PaulWessel:
Since the ancient times, **grdclip**'s **-Sa**_low/value_ and **-Sb**_above/value_ used `<` and` >` in the tests.  But sometimes one wishes `<=` and` >=`.  This PR enhances those to options with modifier **+e** for equality that then does what is desired.

Fixed a wrong indentation level which made many lines change, and a few typos.


joa-quim (approved):

anbj:
PS! #6257.

PR #7865, opened 2023-10-04
psconvert should ignore writing world file if geotiff output is requested

### LINKED ISSUE #7844: Incorrect world file name when generating GeoTIFF file
Issue first reported at https://github.com/GenericMappingTools/pygmt/pull/2698#discussion_r1335248729.
```
gmt psbasemap -R0/10/0/10 -JM10c -Baf > map.ps
gmt psconvert -A -P -W+g map.ps
```

The commands above produce two files: `map.tiff` and `map.pgw`. 

The world file `map.pgw` makes no sense here:

1. The output file is `.tiff`, so the expected world file extension should be `.tfw`, not `.pgw`
2. Since the GeoTIFF file already has the georeference information embedded, a world file seems useless. Perhaps we should delete the world file in this case?

PaulWessel:
See #7844 for discussion.  Closes #7844. Note to @joa-quim : In messages that says

`...the Proj4 command...`

should that be PROJ or are these things still called Proj4 even though we are at Proj9? Let me know if we should update such language in source and rst (separate from this PR).  I just added an extra check that we only do the world_file output if Ctrl->W.warp (**-W+g**) is not set.  See right, no?

### DIFF
--- a/src/psconvert.c
+++ b/src/psconvert.c
@@ -2715,38 +2715,37 @@ EXTERN_MSC int GMT_psconvert (void *V_API, int mode, void *args) {
 				So we'll move halph pixel inward. */
 			west  += x_inc / 2.0;	north -= y_inc / 2.0;
 
-			if (!Ctrl->W.warp) {	/* No need to write a world file for GeoTiff */
-				if (Ctrl->D.active) sprintf (world_file, "%s/", Ctrl->D.dir);	/* Use specified output directory */
-				if (Ctrl->F.active) {		/* Must rip the raster file extension before adding the world one */
-					pos_ext = get_extension_period (out_file);	/* Get beginning of file extension */
-					out_file[pos_ext] = '\0';	/* Remove/restore extension to get world_file prefix */
-					strcat(world_file, out_file);
-					out_file[pos_ext] = '.';
-				}
-				else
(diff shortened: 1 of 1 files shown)


seisman (approved):

joa-quim:
Does it work? Because when it calls `gdal_translate` it's not impossible that a world fine is needed.

The Proj4 *vs* PROJ. There is only proj4 , the name, or PROJ (and don't pay attention to the capitalization. I don't know what is the correct way). No proj5,6,7,8,9. Functions variables, including in official PostGIS docs are still called ``proj4...`` The guys that took care of prof development decided to drop the 4 from the name at around proj v6, but there are still tons of ``proj4`` references in the internet. So, in summary, for my taste, I still prefer to call the ``proj4`` strings as is still used in, for example,
https://spatialreference.org/ref/epsg/4326/

seisman:
> Does it work? Because when it calls `gdal_translate` it's not impossible that a world fine is needed.

You're right. It needs the world file to produce the GeoTIFF file. It means we need to delete the world file after calling gdal_translate.

PaulWessel:
OK, I will fix that. Will leave proj4 of course.

PaulWessel:
How about this one? Had to duplicate world_file since you recycled it for something else and it got lost.

seisman (approved):

PR #7874, opened 2023-10-05
grdcontour.rst: refer to constant contour interval

anbj:
Had to read up on **grdcontour** again, after ``-C+<contour>`` didn't work.

Hopefully makes it a tiny bit clearer.


PaulWessel (approved):
Fine

PR #7876, opened 2023-10-05
Update filter1d.rst. Add reference to technical references

Esteban82:
I just want to add a link to the url below.
https://docs.generic-mapping-tools.org/dev/reference/data-filtering.html

I am not sure if the reference is ok. Please check.


anbj:
I cannot see a 'filtering-of-data-in-gmt' reference in https://docs.generic-mapping-tools.org/dev/reference/data-filtering.html? I think you have to add that to be able to do your link.

If it builds, then discard this.
If not, try adding ``.. _filtering-of-data-in-gmt`` above the title on https://docs.generic-mapping-tools.org/dev/reference/data-filtering.html.

PaulWessel:
Should be simpe I think.  Just grep for http in *.rst and see what is done elsewhere.

Esteban82:
I updated the link. Is it ok now?

anbj:
Builds fine!

I tested your initial commit as well, and that builds fine as well.
I thought you had to define the reference first. Maybe it deduces it from the heading.

anbj (approved):

Esteban82:
Is there any problem with the checks? Or should I just wait? I don't understand why I can't merge this.

seisman:
> Is there any problem with the checks? Or should I just wait? I don't understand why I can't merge this.

The Build workflow is required to pass to make sure that the master branch codes always build successfully. However, to save CI resources, the workflow is skipped for docs-only changes like this one, which means you need admin permission (which I think you don't have) to merge this PR.

Esteban82:
Ok, I see. And is not possible to de-activate those checks for docs-only changes (so I could merge it)?

seisman:
As I understand it, it's impossible without admin permission.

PaulWessel:
If @seisman cannot give admin permission to @Esteban82, how do I give it?

PR #7880, opened 2023-10-07
Prepare -i -o for common modifiers

PaulWessel:
In order to have **-i** and **-o** behave similarly, I first wanted to standardise those two common structures, since **-o** does not have the same variables as **-i** (yet).  I isolated such variables in a **GMT_COL_IO** substructure that both `GMT->common.o` and `GMT->common.o` now use.  This PR simply replaces all the `GMT->common.i.*` references with `GMT-> common.i.col.*` and the same for `GMT-> common.o`.

No changes to functionality yet; just want to get these numerous changes into master before I try to expand what **-o** can do so that those changes are not buried in all the ones in this PR.  Test are unchanged.


PaulWessel:
Not sure if @joa-quim is wrapping` gmt_common.h` which has changed.

joa-quim (approved):
A blind aproval since I’m in a spiritual retreat. 
image

Issue #6053, opened 2021-11-23
Let modifiers of -i (+d|l|o|s) available for -o.


Esteban82:
But for some modules it is not possible to use -i (at least in direct way). 

For example, I will have to do this:

```
gmt pshistogram @v3206_06.txt -F -T50 -I -o3 | gmt info -i+d100 -C -o0
5.26
```

Instead of 
```
gmt pshistogram @v3206_06.txt -F -T50 -I -o3+d100
pshistogram [ERROR]: The -o option does not take +d|l|o|s modifiers; consider -i instead.

```

PaulWessel:
As a command-line user I can sympathize with you on this.  I may decide to add it.  However, the external wrapper types like @joa-quim probably has no use for this as they can do those conversations easily in Julia (and Python and Matlab).  Unlike the input (which is pretty convoluted), the output is much cleaner and I may wish to add a test on this to ensure it does not add much slowness to the execution time.  E.g., if outputting a million records takes some nanoseconds longer than that is fine, etc.

PaulWessel:
See WIP implementation in #7887.

Esteban82:
Now it is working!! Thanks Paul!

PR #7887, opened 2023-10-08
Enhance -o to accept the same transformations as -i

PaulWessel:
See this recent forum [post](https://forum.generic-mapping-tools.org/t/is-there-a-way-to-add-an-offset-to-output-columns/4324/3) for background.  Long overdue upgrade to modify output columns via transformations as we are able to do on input. This is particularly useful when writing table output that originate from grids (e.g., grdtrack).  I have updated the **-o** parser (shared code with **-i** parser), synopsis, documentation, and added a test script _gmt_core/gmt_io_cols.sh_ which compares output from convert with the same settings for **-i** to the output from **-o**.  Passes.
WIP since I want some more testing first.


Esteban82 (approved):
I test it with #6053

Esteban82:
Could I merge this?

PaulWessel:
Already merged

PR #7891, opened 2023-10-09
Skip input ascii records with not enough columns

PaulWessel:
See discussion on the forum [post](https://forum.generic-mapping-tools.org/t/skip-lines-with-mismatch-between-actual-2-and-expected-3-fields/4327/11).  This simply skips the bad line(s) rather than exit.

```
cat << EOF > t.txt
> First line
1 2 3
4 5
3 2 1
> second
6 5
6 7 8
9 9 9
EOF

gmt convert t.txt
gmtconvert [WARNING]: Mismatch between actual (2) and expected (3) fields near line 3 in file t.txt
gmtconvert [WARNING]: Mismatch between actual (2) and expected (3) fields near line 6 in file t.txt
> First line
1	2	3
3	2	1
> second
6	7	8
9	9	9
```

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -8187,6 +8187,7 @@ struct GMT_DATATABLE * gmtlib_read_table (struct GMT_CTRL *GMT, void *source, un
 		return (NULL);
 	}
 
+	strcpy (GMT->current.io.filename[GMT_IN], file);
 	/* Skip binary header first, if binary and there is a header given in # of bytes */
 	if (GMT->common.b.active[GMT_IN]) {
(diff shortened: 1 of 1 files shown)


joa-quim (approved):

anbj:
Hmm, does this only apply to e.g. ``gmt convert``? I'm trying to do the same in ``gmt info``, but same result as before:

```
$ gmt info data.csv -I1 -i8,9,10
gmtinfo [WARNING]: Mismatch between actual (2) and expected (3) fields near line 31695 in file data.csv
```

PR #7892, opened 2023-10-09
Reset a bit more of the API fresh state in gmt_end_module()

joa-quim:
This affects externals only.

Some modules like ``grdimage`` may leave the API expecting further data to be geographic and error when receiving cartesian tables.  This PR resets the internal state into a closer to what ``gmtlib_io_init()`` would have left it. Can't call it though because then some other mysterious errors would popup.


PaulWessel (commented):
Did you run the test set? I am worried it could break something...

PaulWessel (approved):
Taking your word for it.

PR #7893, opened 2023-10-09
Update long-options for -o

PaulWessel:
@rbdavis, given we now merged in #7887 we need to update -o, as -i.


rbdavis:
Hi Paul,

I'm not sure what this is about, but looking at your code change on the
git page I wonder if you should have preserved the original 'o'
short-option string in the modified long-option entry rather than changing
it to 'i'?

I'm working at ARL today so I can't really look at anything easily, sorry.

Roger

rbdavis (approved):

PR #7898, opened 2023-10-10
Disallow -B when grdimage -A is used

PaulWessel:
See pyGMT [issue](https://github.com/GenericMappingTools/pygmt/issues/2599) for background.  This PR checks that **-A** and **-B** are not both used..


joa-quim (approved):
But it could as well just warn and ignore.

PaulWessel:
Yes, but I think better to suppress wrong things since the -B output just confuses users.

PR #7899, opened 2023-10-11
add -F information to description

mlees0209:
It looks like the documentation has not been updated along with the addition of -F functionality -- I am proposing a small change to correct that.


PaulWessel (changes requested):
Please instead say. "Finally, use option |-F| to set all nodes inside or outside the specified polygon to NaN.

PaulWessel:
I have done this but a bit differently  Closing as finished,

PR #7901, opened 2023-10-13
Try harder to guess date format and update template

PaulWessel:
While desirable, it is just not possible to expect random people to pay attention to ISO format specifications, so they give `dd/mm/yyyy` while we expect `yyyy-mm-dd`.  This PR tries a bit harder in _gmtio_examine_current_record_ to determine what we have and update the template if needed.  If the user have set **FORMAT_DATE_IN** then we do not change it, but if it is set to the default ISO template (`yyyy-mm-dd`) then we check if a candidate absolute date-time input can work with that or if we got something else, e.g., `dd/mm/yyyy` or `dd-monthname-yyyy` and need to rebuild the active **FORMAT_DATE_IN** format.

@joa-quim, have a look at the gmtio_check_abstime_format_is_suitable function in gmt_io.c.  It should catch most common things but not idiocy like OCT-2001-11, but we could add even that...

### DIFF
--- a/src/gmt_io.c
+++ b/src/gmt_io.c
@@ -3378,17 +3378,21 @@ GMT_LOCAL unsigned int gmtio_get_coltype_name_index (struct GMT_CTRL *GMT, unsig
 bool gmtlib_maybe_abstime (struct GMT_CTRL *GMT, char *txt) {
 	size_t k;
 	unsigned int n_dash, n_slash;
+	int start = -1;
 	gmt_M_unused (GMT);
-	if (strchr (txt, 'T')) return true;	/* Might be [<date>]T[<clock>], else if is likely text */
+	if (strchr (txt, 'T')) return true;	/* Might be [<date>]T[<clock>], else it is likely text */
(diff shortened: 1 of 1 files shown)


PaulWessel:
```
(base) pwessel@MacAttack-2:~-> gmt info enso_new.txt
enso_new.txt: N = 1824	<1870-01-01T00:00:00/2021-12-01T00:00:00>	<-2.49/2.57>
(base) pwessel@MacAttack-2:~-> gmt info enso_new.txt --FORMAT_DATE_IN=yyyy/mm/dd
enso_new.txt: N = 1824	<1870-01-01T00:00:00/2021-12-01T00:00:00>	<-2.49/2.57>

echo 21-JAN-2006 | gmt convert
2006-01-21T00:00:00
echo 2006-JAN-21 | gmt convert
2006-01-21T00:00:00
echo 12-12-2019 | gmt convert
2019-12-12T00:00:00

```

joa-quim (approved):

PR #7490, opened 2023-06-02
Add -Fe to GMT_splines.sh

Esteban82:
Update script to add the new method to interpolate (#7488).

I make the figure taller so the legend doesn't cover the curves.

PR #7902, opened 2023-10-13
Update gmt_support.c

PaulWessel:
The **sample1d -Fe** interpolator could only go up but now it can go down as well.  Please check, @federico if this is what you expected:

GMT_splines

Related to #7490.  If you agree please update the script and plot (dvc if you have tried that before).


Esteban82 (approved):

Esteban82:
@PaulWessel now with this PR this _-Fe requires momotonically increasing data_ is false?

PaulWessel:
Yes, monotonic is now obsolete

PaulWessel:
Sorry, going to bed but the legend should not sayStep-up curve. Maybe just Step curve.

PR #7918, opened 2023-10-16
Fix grdcut tif grids

joa-quim:
Cutting geotiff float files was failing because the .tif extension was forcing trying to save as images when input is clearly a grid (determined by data type).

I still left non-working commented lines to see if brings some memory of what was meant in that line.


joa-quim:
Ping

PR #7923, opened 2023-10-17
Use --metric (and aliases) for -j

PaulWessel:
I had selected **--distnce** for **-j** but not used in PyJu which also differ.  But since **-D** is **--distance** in the two grid filter functions we want to preserve that so **-j** changes instead.

### DIFF
--- a/src/gmt_common_longoptions.h
+++ b/src/gmt_common_longoptions.h
@@ -107,7 +107,7 @@
     { ',', 'i', "incols",
                 "",                      "",
                 "l,d,o,s",               "log10,divide,offset,scale" },
-    {   0, 'j', "distance",
+    {   0, 'j', "metric|spherical|distcalc",
                 "e,f,g",                 "ellipsoidal,flatearth,spherical",
                 "",                      "" },
     {   0, 'l', "legend",


remkos (commented):
I don't know what *-j* is really about since it is not in the manual (oversight?).
But with this new spec, you could create --spherical=spherical. What does that mean?

PaulWessel:
I think metric is superior 
—spherical is in Julia so we should add it as an Alia’s and encourage @joa-quim and @seisman to  add metric as alias as well

joa-quim:
Added ``metric`` to GMT.jl **-j** aliases list

seisman:
> I think metric is superior —spherical is in Julia so we should add it as an Alia’s and encourage @joa-quim and @seisman to add metric as alias as well

Unfortunately, PyGMT doesn't allow two aliases for the same option flag. I'll open an issue report in the PyGMT repository.

joa-quim:
That is the same issue as ``R`` *vs* ``region``, no? Don't you guys want to revert that decision? I think is a really bad one.

PaulWessel:
These are good reasons to have an all-GMT summit next year, ideally spring

rbdavis (approved):

PR #7928, opened 2023-10-17
Change var name in function prototype to avoid build error.

joa-quim:
It turns out that having `struct GMT_PEN *PR` in the `gmt_two_curve_fill()` prototype was conflicting with a `#define PR` that I have in one of my mb modules (`mbclean`). Changing that var name in the prototype fixed the build error. WTF!!!


PaulWessel (approved):
Silliness

PR #7931, opened 2023-10-17
Remove test figure that also appears in documention

PaulWessel:
The bezierline.ps file was also identically added as GMT_bezier.ps in the pen documentation so removing the test.


seisman (approved):

seisman:
@PaulWessel Did you forget to run `dvc push`?

PaulWessel:
Don't think so but just did it.

Issue #7934, opened 2023-10-18
grdinterpolate -o is confused


joa-quim:
Full output
```
C:\v>grdinterpolate LC08_044034_20170614.nc  -S606200.0/4.2134405e6
> Location 606200,4.21344e+06
606200  4213440.5       1       0.126504148625
606200  4213440.5       2       0.113188595884
606200  4213440.5       3       0.0914816445611
606200  4213440.5       4       0.0569632023659
606200  4213440.5       5       0.0409598219348
606200  4213440.5       6       0.0330643219513
```

Requesting last 2 columns prints first two
```
C:\v>grdinterpolate LC08_044034_20170614.nc  -S606200.0/4.2134405e6 -o2,3
> Location 606200,4.21344e+06
606200  4213440.5
606200  4213440.5
606200  4213440.5
606200  4213440.5
606200  4213440.5
606200  4213440.5
```
The nc file is 28 MB so not attaching it.

PR #7936, opened 2023-10-18
Need to reparse -o when output received from module call

PaulWessel:
See #7934 for background.  This PR checks if **-o** was set and reparses the **-o** option since it may get mangled by intermediate _GMT_Call_Module_ calculations and returns.


joa-quim (approved):
Chcked that it work. Thanks

PR #7947, opened 2023-10-19
Add longoption support for gmt2kml

### LINKED ISSUE #7946: longoption support for gmt2kml

longoption support for gmt2kml

rbdavis:

Add longoption support for gmt2kml.

Python differences:

  No Python module found.

Julia differences:

  -C aliases limited to cpt|cmap to match other modules

  extrude (-E) changed to line_render
    Extrusion is only one of multiple possibilities with -E
    '+extrude' is used as a modifier longoption (+e)

  -Q{a,i,s} option variants (wiggles/wiggle_fixedazim/wiggle_scale)
  combined into a single -Q wiggle longoption

  ilscale (-S) changed to scale. Use +icon and/or +label as needed

Fixes #7946

**Reminders**


remkos (commented):
-A : I do not see how the optional <altitude> or x<scale> are being dealt with.
-D : I prefer "description"
-E : I prefer "extend,connect"
-Q : I prefer "azimuth,fixed,scale"

joa-quim:
We may have some multiple opinions here. I still think `extrude` is the right one. I only know this word from the KML slang, but that's what is used there and that's what this option do. `connection` is a sub-option of `extrude`, not an alternative.

remkos:
@joa-quim : I agree on the `extrude`. Had to look that up again. Long since I worked on that.

joa-quim:
We have many instances of **-G** meaning a fill (color or pattern). I think it's better that they all have the same long option name. Even if only a color can used as the fill.

PaulWessel:
> We have many instances of **-G** meaning a fill (color or pattern). I think it's better that they all have the same long option name. Even if only a color can used as the fill.

We could do --color as alias with fill then

joa-quim:
Hmm, just by reading the docs _Fill sets infill color for symbols, extrusions, polygons and positive anomaly wiggles_ I would think polygons can be filled with patterns as well, no?

PaulWessel:
Beyond my understanding.  I did not implement that since I saw no way of doing it, but if you think it is doable then maybe it is.

joa-quim:
Agree on a ``fill``, ``color`` pair.

remkos:
I second that

remkos:
> @remkos, -A would take the optional argument:
> 
> 
> 
> `--altitude=relative_surface:500`
> 
> 
> 
> Note the man page says exactly the same thing for both -Ag and -As.  I assume it should be ground vs seafloor and we should fix the doc.
> 

Correct. Understood!

PaulWessel (approved):
Made a few updates based on comments, merged in master. Approved, ready to be merged. Watch for any l2s tests that may need to be updated as well

PR #7949, opened 2023-10-19
Add longoptions support for gmtbinstats

### LINKED ISSUE #7948: longoption support for gmtbinstats

longoption support for gmtbinstats

rbdavis:
Add longoptions support for gmtbinstats

Python differences:

  None?

Julia differences:

  gmthelp(gmtbinstats) hangs with minimal output
  on my system (macOS Monterey 12.6.8 on ARM CPU)

Fixes #7948

**Reminders**


joa-quim:
> gmthelp(gmtbinstats) hangs with minimal output

Yep, that `gmthelp()` is complicated and hard to maintain. I don't like this other (very limited) online help and hope to find a way to replace it with something more valued but it's good  for these type of checks. You can see it with ``? module_name``, that is in this case ``? gmtbinstats``. Even better, when a full man page exists (which is not the case yet of `gmtbinstats`) is to do ``@? module_name``.

PaulWessel:
Needs to use

`    GMT_INCREMENT_KW,   /* Defined in gmt_constant.h since not a true GMT common option (but almost) */
`

since it is standard increment parsing.  I updated that and also merged in 21 commits from master.

PaulWessel (approved):
I think this is good to go

PaulWessel:
Ping @GenericMappingTools/long-options-review : I think we decided at the summit in Oslo that this team should review these pull requests and at the _very least_ check the approve button so that we know you actually looked.  So far I only see mostly Joaquim, Remko, and myself doing this.  It takes just a minute to look at one of these longopt/*_inc.h files and your fresh eyes may be used to suggest that `--awkwardoption `is a bit unclear and perhaps` --set_pole` is better?  Or, maybe you wish to argue a bit with me if I suggest `--not_negative` if the original short-option was **-Q** for "select positive values"?

There are five PRs from ROger (see them by selecting the Autor tab above the pull requests.  Roger will merge these in on Monday Hawaii time so please chip in your 2-cents or approve so we know you are involved.

PR #7951, opened 2023-10-19
Add longoptions support for gmtconnect

### LINKED ISSUE #7950: longoption support for gmtconnect

longoption support for gmtconnect

rbdavis:
Add longoptions support for gmtconnect

Python differences:

  No Python module

Julia differences:

  linkfile (-L) changed to link_file to match style with list_file (-Q)
    Better to use underscore for both options or neither.

Fixes #7950

**Reminders**


PaulWessel:
@joa-quim, what do you think about the difference between **-L** (takes a single file name) and **-Q** (takes a format template from which two file names can be created)?  Is list_file and link_file still reasonable?

joa-quim:
To be thrush I don't like much the `file` suffix. In Julia/MEX we can make it return a GMTdataset. Perhaps `link_info`, `list_names`?

PaulWessel:
I see, yes "file" is not always true.  Maybe just --link and --lists ?

joa-quim:
OK

joa-quim:
But should be `links`, not `link`

PaulWessel:
Well, **-L** only accepts a single destination (ie., one file) with no template for several links (unlike **-Q** which can be one or a template with %c in it)

joa-quim:
Yes, but it stores the _links_ (there are many) not just one as a singular would suggest.

PaulWessel:
Good point. plural it shall be

PaulWessel (approved):

Issue #7954, opened 2023-10-20
Inconsistent dimension names (x/y vs lon/lat) for tiled and non-tiled grids


seisman:
To reproduce the issue:
```
gmt grdcut @earth_relief_06m_p -R0/10/0/10 -G06m.nc
gmt grdcut @earth_relief_05m_p -R0/10/0/10 -G05m.nc
```

Run `gdalinfo` to check the two files:
```
$ gdalinfo 05m.nc 
Driver: netCDF/Network Common Data Format
Files: 05m.nc
Size is 120, 120
Origin = (0.000000000000000,10.000000000000000)
Pixel Size = (0.083333333333333,-0.083333333333333)
Metadata:
  NC_GLOBAL#Conventions=CF-1.7
  NC_GLOBAL#description=
  NC_GLOBAL#GMT_version=6.5.0_98221f9-dirty_2023.10.18 [64-bit]
  NC_GLOBAL#history=gmt grdcut @earth_relief_05m_p/ -R0/10/0/10 -G05m.nc
  NC_GLOBAL#node_offset=1
  NC_GLOBAL#title=
  x#actual_range={0,10}
  x#axis=X
  x#long_name=longitude
  x#standard_name=longitude
  x#units=degrees_east
  y#actual_range={0,10}
  y#axis=Y
  y#long_name=latitude
  y#standard_name=latitude
  y#units=degrees_north
  z#actual_range={-5071,2769.5}
  z#cpt=geo
  z#long_name=z
  z#_FillValue=nan
Corner Coordinates:
Upper Left  (   0.0000000,  10.0000000) 
Lower Left  (   0.0000000,   0.0000000) 
Upper Right (  10.0000000,  10.0000000) 
Lower Right (  10.0000000,   0.0000000) 
Center      (   5.0000000,   5.0000000) 
Band 1 Block=120x1 Type=Float32, ColorInterp=Undefined
  NoData Value=nan
  Metadata:
    actual_range={-5071,2769.5}
    cpt=geo
    long_name=z
    NETCDF_VARNAME=z
    _FillValue=nan

$ gdalinfo 06m.nc
Driver: netCDF/Network Common Data Format
Files: 06m.nc
Size is 100, 100
Origin = (0.000000000000000,10.000000000000000)
Pixel Size = (0.100000000000000,-0.100000000000000)
Metadata:
  lat#actual_range={0,10}
  lat#axis=Y
  lat#long_name=latitude
  lat#standard_name=latitude
[...]

PR #7956, opened 2023-10-20
Ensure geographic grids are written as such

PaulWessel:
See #7954 for failure.  This PR addresses this problem related to assembled grids.


seisman (approved):
Works!

PR #7958, opened 2023-10-20
WIP Explore Latex in subplot tags

PaulWessel:
See this forum [post](https://forum.generic-mapping-tools.org/t/latex-in-subplot-title/4364) for background.  Not straightforward since we run a special PSL function to set those.  Tried the obvious (relying on the other places we handle this) but get a crashing PostScript.  Will follow up.


src/gmt_plot.c
```diff
@@ -2675,24 +2675,24 @@ GMT_LOCAL void gmtplot_consider_internal_annotations (struct GMT_CTRL *GMT, stru
 	PSL_settextmode (PSL, PSL_TXTMODE_HYPHEN);	/* Back to leave as is */
 }
 
-GMT_LOCAL void gmtplot_map_label (struct GMT_CTRL *GMT, double x, double y, char *label, double angle, int just, unsigned int axis, bool below) {
+GMT_LOCAL void gmtplot_map_label (struct GMT_CTRL *GMT, double x, double y, char *label, double angle, int just, unsigned int axis, bool below, unsigned int feature) {
```
joa-quim, line 2678: `unsigned int feature`. Aren't ``enums`` signed ints? So this one should be signed int too, no? Besides `unsigned ints` are a pain.

PaulWessel: feature is now a `enum GMT_enum_feature`

PaulWessel:
For testing, here is a small subplot that fails due to Latex code:

```
#!/bin/bash
gmt begin Figure png
  gmt set FONT_TAG 16p,Times-Italic
  gmt subplot begin 1x1 -Fs6i -A+o-0.2c/0 -JW180/6i -Rg
    gmt subplot set 0,0 -A"@[\nabla^2@["
    gmt basemap ${eq_proj} -B0
  gmt subplot end
gmt end show
```

I will update the **subplot** documentation with a note that Latex expressions in the **-A** is currently not allowed.

PaulWessel:
See #8016 for documentation warning.

PR #7961, opened 2023-10-20
Let grdcut -E extract a vertical slice from a 3-D cube

PaulWessel:
**grdcut** **-Ex**|**y**_coord_ lets users extract a vertical slice grid that is either parallel to the x or y direction and goes through the specified coordinate. For example, `gmt grdcut my cube.nc -Gslice.nc -Ey33` will cut along the row corresponding to y = 33. For **grdcut** we require the cube be equidistant in the 3rd dimension, and that _coord_ is a node coordinate so there is no resampling here, just picking the nodes along that slice.  Once tested I will add this to **grdinterpolate** where, if **-T** is set and *coord* does not necessarily fall on a node,. we will resample by providing the coordinates for the output grid.

### DIFF
--- a/doc/rst/source/grdcut.rst
+++ b/doc/rst/source/grdcut.rst
@@ -17,6 +17,7 @@ Synopsis
 |-G|\ *outgrid*
 |SYN_OPT-R|
 [ |-D|\ [**+t**] ]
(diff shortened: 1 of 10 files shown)


joa-quim (approved):

PaulWessel:
Added working test and polished docs.

PR #7972, opened 2023-10-24
Let psconvert -N+kfill set alternate fade fill

### LINKED ISSUE #7593: Movie. Canvas color and -K
It is not possible to set the background color (-G) of the canvas when -K is used. 

1. When I use this, then the canvas is red

`gmt movie main.sh -Nffmpeg1 -T40 -C12cx12cx80 -Z  -D12 -Fmp4 -Vi -Gred`

2. With this command, then the fade in/out are red. And the canvas during the animation is white. I can't find I way to set the canvas color in a case like this.

`gmt movie main.sh -Nffmpeg2 -T40 -C12cx12cx80 -Z  -D12 -Fmp4 -Vi -Gred -K+p`

3. Here, the inicial and final frames are red even when I used `+gyellow`. I attach this mp4. 

`gmt movie main.sh -Nffmpeg3 -T40 -C12cx12cx80 -Z -D12 -Fmp4 -Vi -Gred -K+p+gyellow`

## Solution

I think that `-G` should be only be use to the set the color of canvas for the main animation.

## Full script
```
cat << EOF > main.sh
gmt begin
	echo \${MOVIE_FRAME} | gmt text -R0/1/0/1 -JX10c -F+f200p+cCM -B0 -X0 -Y0
gmt end
EOF

gmt movie main.sh -Nffmpeg1 -T40 -C12cx12cx80 -Z  -D12 -Fmp4 -Vi -Gred
gmt movie main.sh -Nffmpeg2 -T40 -C12cx12cx80 -Z  -D12 -Fmp4 -Vi -Gred -K+p
gmt movie main.sh -Nffmpeg3 -T40 -C12cx12cx80 -Z -D12 -Fmp4 -Vi -Gred -K+p+gyellow
```

https://github.com/GenericMappingTools/gmt/assets/17908951/d85a9545-1bc2-436b-88b7-1613eec17dc6

PaulWessel:
See #7593 for some background. **movie** allows two fill settings that are passed to **psconvert** when building the rasters:

1. With **-G** one can optionally  fill the bounding box with specified colour before other things are painted
2. With **-K** (and **-E**) one can specify what the terminal fade colour should be [Default is black].

Unfortunately, **psconvert** only had **-N+g** to accept the background fill and the fade colour was passed from **movie** the same way via **-N+g**, resulting in two different **+g**_color_ modifier passed to **psconvert**, which then of course used just the last one given.  Thus, fade was always black.

This PR does several things:

1. Adds **+k**_fadefill_ to **-N** and updates **movie** to pass that modifier to **psconvert** when the fading colour is explicitly set.
2. Deprecates **-N+i** for ancient _ghostscript_ versions.   We honour it under the hood but drop the documentation and the reasoning behind it.
3. Fix a few missing `psconvertoptions` in **figure**
4. Use _gmt_rgb_syntax_ not _gmt_pen_syntax_ to complain about bad rgb fill

Closes #7593

https://github.com/GenericMappingTools/gmt/assets/26473567/fd730d46-e3c7-434e-8134-d845152fbcd8

### DIFF
--- a/doc/rst/source/figure.rst
+++ b/doc/rst/source/figure.rst
@@ -59,7 +59,7 @@ Optional Arguments
     Sets one or more comma-separated options (and possibly arguments) that
     can be passed to :doc:`psconvert` when preparing this figure [**A**].
     The valid subset of options are
-    **A**\ [*args*],\ **C**\ *args*,\ **D**\ *dir*,\ **E**\ *dpi*,\ **H**\ *factor*,\ **M**\ *args*,\ **Q**\ *args*,\ **S**.
+    **A**\ [*args*],\ **C**\ *args*,\ **D**\ *dir*,\ **E**\ *dpi*,\ **H**\ *factor*,\ **I**\ *args*,\ **M**\ *args*,\ **N**\ *args*,\ **Q**\ *args*,\ **S**.
     See the :doc:`psconvert` documentation for details on these options.
 
 .. |Add_-V| replace:: |Add_-V_links|
--- a/doc/rst/source/psconvert.rst
+++ b/doc/rst/source/psconvert.rst
@@ -23,7 +23,7 @@ Synopsis
 [ |-I|\ [**+m**\ *margins*][**+s**\ [**m**]\ *width*\ [/\ *height*]][**+S**\ *scale*] ]
 [ |-L|\ *list* ]
 [ |-M|\ **b**\|\ **f**\ *pslayer* ]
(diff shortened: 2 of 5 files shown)


joa-quim (approved):

Esteban82:
I see now that the default color for the fade is black while the canvas color is white (no fill). Wouldn't be better to have the same default color? I know that it is not important.

Esteban82:
> If they both were white or black, you would fade from white to white or black to black.

Mmm, yes, you will see the content of the last frame dissapear (to white or black).

But, I am ok with your solution.

PR #7974, opened 2023-10-24
TM gives widely different answer for spherical and elliptical

PaulWessel:
See forum [post](https://forum.generic-mapping-tools.org/t/importance-of-r-for-mapproject/4363) for the problem.  This PR fixes some errors in the code relative to Snyder, but does not yet fix the 16 km difference between spherical authalic and full ellipsoidal.

Changes so far:

_M_ and _M_0_ used _s2_ and _c2_ in calculation but Snyder has s and even powers of _s_ only (s, s2, s4, s6).

Other changes shown are just better notation (e.g, use _A4_ rather than first set `A3 *= A` and use _A3_ - easy to loose track and the multiply happens anyway.

However, this change only changed the output from 
`-20.7821644863	64.5340165896`
to
`-20.7846140796	64.5338009013`

@joa-quim, you know proj4, could you run it on the equivalent of the **mapproject** command that give the above?

`echo "-137000.0 11000.0" | gmt mapproject -Jt-17.927/64.463/1:1 -R-23/-8/60/66 -I -C -F`


joa-quim:
```
echo -137000.0 11000.0 | gmt mapproject -J+proj=tmerc+lon_0=-17.927+lat_0=64.463 -I
-20.7821644807  64.5340165831
```

PaulWessel:
So after fixing more bugs in our Snyder implementation of TM ellipsoidal than I have fingers, we have a working solution.  Here are some comments:

- Our test snyder.sh (which exercises the various numerical examples printed in Snyder) only tested the spherical TM version and that example passes.  However, there was no test for the ellipsoidal example despite having them for other projections.  When I added it the test failed in master but passes with this branch.
- It may be that some of the things that differed from Snyder was obtained from another source and used Horner series packing, but good to have a check that now passes.

One question remain and would like to get input from @WalterHFSmith on this one:

Decades ago we (i.e., you mostly) added the auxiliary latitudes and we now use Snyder's spherical equation with authalic latitudes (for several projections). So basically we have this setup:

1. If **PROJ_ELLIPSOID** is Sphere then just use spherical equation with normal latitude
2. Else, if w or e on  the longitudinal w/e of the area of interest exceeds _10 degrees_ from central meridian then elliptical series becomes inaccurate and we switch to the spherical equations but insert authalic latitudes
3. Otherwise, we use the ellipsoidal equations with given latitudes.

Because of how GMT was designed 30 years ago, **mapproject** thus needs the **-R** to make the decision, and it is annoying since **-R** is otherwise not needed for these calculations.  We presumably were transfixed on **-R -J** for making maps so let **mapproject** expect teh same.

This means we have two options:

1. Add another way for mapproject to decide between spherical and ellipsoidal, say **-Hs**|**e** [**e** for ellipsoidal is default], and let commands with **-R** be backwards compatible.
2. Stick with current setup but improve documentation with a short Ellipsoidal Series vs Spherical.

Thoughts, comments?

WalterHFSmith:
Hi,

I am not sure why were are doing the below in the context of a Transverse Mercator projection. The TM, if I understand, wraps a cylinder of elliptical cross-section so it is tangent two a meridian on the ellipsoid (or, if scaled peculiarly, I suppose it could be a shrunken or expanded version of such a cylinder). 

If the ellipsoid is a sphere then the TM could be a conformal projection, in the way that the usual Mercator is conformal. To do this one would not need to use the conformal auxiliary latitude, as on a sphere all the auxiliary latitudes are equal to the spherical latitude.

f the ellipsoid is not a sphere, then the best the TM can do is to give true north-south distance along the meridian on which the cylinder is tangent. For this purpose one could use the north-south-distance-true auxiliary latitude and then proceed with a transverse development of the sphere, substituting the N-S-D-T auxiliary latitude.

Notice that neither of these cases uses the “authalic” latitude. The “authalic” latitude is an auxiliary latitude such that, if the geodetic latitude on an ellipsoid is converted to the auxiliary authalic latitude and then this latitude is substituted into an equal-area (that’s what authalic means) development of a sphere, the result will be an equal area development of the ellipsoid. But the Mercator and TM are not equal area projections, so it makes no sense (to me) to use the authalic latitude in this case.

In general we can’t do much with oblique or transverse developments of an ellipsoid. Snyder’s calculations used truncated series which aimed to be accurate enough if the map covered a small domain. For cylindrical, conical, and azimuthal developments where the axis of the development is along the ellipsoid’s north-south axis, there are correct solutions useful over wide domains, and they can be created using the appropriate auxiliary latitude followed by a spherical development formula with the substituted latitude. However, just as  a map can be at most one of {equal area, conformal (equal angle), distance-true along some path} there are auxiliary latitudes that allow at most one of these properties to hold.

Sorry if I am being pedantic. Sorry also for not understanding why we are using authalic for the TM.

W

PaulWessel:
Thanks Walter, a minute after I posted I edited my post to say conformal instead of authalic but I guess you did not get that update somehow.  We are definitively using conformal latitudes for non-spheres for larger regions.  So I think we are good.  It is just the issue of how the user selects this scenario (currently via -R to ensure w or e is>10 degrees from central) or have an explicit setting to pick between ellipsoidal or spherical.  I think that is more flexible than having to determine suitable -R settings just for that.

anbj:
For what it's worth (and if I remember correctly);
the need of having to specify ``-R`` in mapproject when projecting lon,lat to utm was very confusing to me. I discovered that I could dodge this by specifying the zone and/or hemisphere in my ``-J`` command, e.g. ``-Ju+32/1:1`` (which I guess then implicitly gives the ``-R``, which then uses ellipsoidal equations; which I wanted/needed).

Point: If it's not necessary to specify a region, ``-R``, don't require it.

PaulWessel:
Looking at **mapproject** which has gained lots of functionality since the Bush I administration: (e.g., **-A**, **-G**, **-L**, **-N**, **-Z**, none of which needs **-R -J**).

Seems **-R** and **-J** are only required for **-W** (which deals with plot regions that must be computed).

Al other option combinations do not need **-R**.  The only exception is if **-J** is given with map width and not scale: The width must be related to a region for us to convert it to scale.

My plan is now to:

1. Enable **-j** for **mapproject** and **grdproject**.  It lets users force ellipsoidal vs spherical projections.
2. Remove **-R** requirement for **mapproject** [not sure we have that in **grdproject** since a grid is always given]
3. Add a paragraph explaining why there are truncated ellipsoidal series approximations, why there are spherical exact solutions, under what conditions GMT selects one or the other, and what the user can control. Also spell out that this level of detail only pertains to coordinate conversions.  For a map of almost any size the differences between the projections is minuscule and not detectable.
4. Assess how much of this applies to **grdproject**.

PaulWessel:
Ping @joa-quim for review.

joa-quim (approved):

Issue #7846, opened 2023-09-25
Upgrade ex15


Esteban82:
I was looking at [example 15](https://docs.generic-mapping-tools.org/dev/gallery/ex15.html#example-15) and I think it can be improved. 

### Bugs: 
In the docs it says: 
> Our lower left plot illustrates the results of gridding using a nearest neighbor technique ([nearneighbor](https://docs.generic-mapping-tools.org/dev/nearneighbor.html)) which is a local method: **No output is given where there are no data.** 

This is false (see figure). We need to add `-N4+m1` to `nearneighbor` or to modify the text. The default behaviour is "a quadrant search with 100% coverage, i.e., sectors = min_sectors = 4]". So, in the map, only the nodes surrounded by 4 data are highlighted. What should I do? Modify the script or the text?

### Important modification:
* Plot the data in at least one figure.
* Replace contours with images.

### Potencial new version

![ex15_new](https://github.com/GenericMappingTools/gmt/assets/17908951/fd42e37c-ba53-4685-b31a-dc4f96cea94e)

```
gmt begin ex15_new png
	#gmt convert @ship_15.txt -bo > ship.b
	region=$(gmt info ship.b -I1 -bi3d)
	gmt subplot begin 2x2 -M0.3c/0.1c -Fs7.5c/0 $region -JM7.5c -BWSne -T"Gridding with missing data"
		#   Raw nearest neighbor contouring
		gmt nearneighbor $region -I10m -S40k -Gnn.nc ship.b -bi
        gmt grdimage nn.nc -c1,0
        gmt plot ship.b -bi3d -Sc0.005c
		#gmt grdcontour ship.nc -JM -C250 -A1000 -Gd5c -c1,0
		#   Mask out area with no data using coastlines
	    gmt mask -I10m ship.b -T -bi3d -c1,1 -Gorange
        gmt plot ship.b -bi3d -Sc0.005c
		#gmt grdcontour ship.nc -C250 -L-8000/0 -A1000 -Gd5c
		#   Grid via surface but mask out area with no data
		gmt blockmedian ship.b -b3d > ship_10m.b
		gmt surface ship_10m.b -Gship.nc -bi
		gmt mask -I10m ship_10m.b -bi3d -c0,0
   		gmt grdimage ship.nc
		#gmt grdcontour ship.nc -C250 -A1000 -L-8000/0 -Gd5c
		gmt mask -C
		#   Clip data above sealevel then overlay land
[...]

PR #7977, opened 2023-10-26
New version of ex15

Esteban82:
See #7846 for details.

This would be the new version. It is the same that in the link but I  add  `-Scb -Srl` for `subplot`.
ex15

### DIFF
--- a/doc/examples/ex15/ex15.bat
+++ b/doc/examples/ex15/ex15.bat
@@ -12,19 +12,20 @@ gmt begin ex15
 	gmt subplot begin 2x2 -M0.3c/0.1c -Fs7.5c/0 %region% -JM7.5c -BWSne -T"Gridding with missing data"
 		REM   Raw nearest neighbor contouring
 		gmt nearneighbor %region% -I10m -S40k -Gship.nc ship.b -bi
-		gmt grdcontour ship.nc -JM -C250 -A1000 -Gd5c -c1,0
+        gmt grdimage ship.nc -c1,0
+        gmt plot ship.b -bi3d -Sc0.005c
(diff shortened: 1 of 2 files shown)


PaulWessel (approved):
Looks good, but a few issues:

1. You have a mix of spaces and tabs for indent and it looks odd depending on how tabs are converted to spaces.  Better to just use tabs as we do elsewhere
2. No need to comment out the "show" - gmtest.in takes care of such things.
3. There is also ex15.bat?
Please fix those before merging since I am approving.

Esteban82:
> but a few issues:

I think that I fixed all of them.

PaulWessel (approved):

Esteban82:
@PaulWessel could you update the figure within dvc?

PaulWessel:
Yes, tomorrow morning but you need to merge this approved PR into master.  I will do the fig based on the new script.

Esteban82:
Ok, but I can't merge this PR ("some checks haven’t completed yet" and they won't be run since only the documentation was modified).

Esteban82:
In the figure, there is this error:

In the docs it says:

> Our lower left plot illustrates the results of gridding using a nearest neighbor technique ([nearneighbor](https://docs.generic-mapping-tools.org/dev/nearneighbor.html)) which is a local method: No output is given where there are no data.

This is false (see figure). We need to add -N4+m1 to nearneighbor or to modify the text. The default behaviour is "a quadrant search with 100% coverage, i.e., sectors = min_sectors = 4]". So, in the map, only the nodes surrounded by 4 data are highlighted. What should I do? Modify the script or the text?

joa-quim:
Just merged it myself.

PR #7986, opened 2023-10-30
Added new GMT_{CPT,OUTGRID,PEN}_KW definitions for longoptions usage.

### LINKED ISSUE #7983: Add longopt constants for near-common options -C, -G, -I and -W

While the common GMT options have its own _gmt_connon_longopt.h_ file, there are many options that are near-common in that the usually have the exact same meaning in many modules, but were never selected to become common options (which always mean the same in any modules, e.g., **-R**, **-J**, **-U**, etc.).

At the moment we have one longopt constant for **-I**, i.e. **--increment**.  It is defined as a constant in _gmt_constants.h_ thus:

`#define GMT_INCREMENT_KW { '/', 'I', "increment|inc", "", "", "e,n", "exact,number" }`

which allows any _longopt/*_inc.h_ file that requires **--increment** to use that as here in _blockmean_inc.h_:

```
...
{ 0, 'G', "outgrid",   "", "", "", "" },
GMT_INCREMENT_KW,   /* Defined in gmt_constant.h since not a true GMT common option (but almost) */
 { 0, 'S', "statistic",
...
```

@rbdavis, please add similar constants for **-C**_cpt_ [**GMT_CPT_KW**], **-G**_outgrid_ [**GMT_OUTGRID_KW**], and **-W**_pen_ [**GMT_PEN_KW**], and use these constants in module includes that use those options in the sense indicated.  There are certainly modules that use **-C** for something else so we skip those of course.  Also make sure to use **GMT_INCREMENT_KW** whenever grid increments are set. You will come across modules that need to set a pen but must but use some other option that **-W** since it may be taken for _weights_ or other things. Not sure how we handle those since the 'W' is then hardwired into the constant.  Any clever ideas to avoid duplicating GMT_PEN_KW to GMT_PENFp_KW (say for **-Fp**_pen_ in **psmeca**?

rbdavis:
New additions to gmt_constants.h for quasi-global options -C, -G and -W.

Fixes #7983

**Reminders**

### DIFF
--- a/src/gmt_constants.h
+++ b/src/gmt_constants.h
@@ -223,10 +223,12 @@ enum GMT_time_period {
  * supported module documentation (e.g., manual pages) rather than these
  * non-user-facing translation entries.
  */
+/* next line is temporary, remove after all conversions to GMT_I_INCREMENT_KW are complete */
 #define GMT_INCREMENT_KW { '/', 'I', "increment|inc", "", "", "e,n", "exact,number" }
-#define GMT_CPT_KW	{ '/', 'C', "cpt|cmap", "", "", "h,i,u,U", "hinge,zinc,fromunit,tounit" }
-#define GMT_OUTGRID_KW	{ '/', 'G', "outgrid", "", "", "d,n,o,s,c,l", "divide,nan,offset,scale,gdal,list" }
-#define GMT_PEN_KW	{ '/', 'W', "pen", "", "", "c", "color" }
+#define GMT_I_INCREMENT_KW { '/', 'I', "increment|inc", "", "", "e,n", "exact,number" }
+#define GMT_C_CPT_KW	{ '/', 'C', "cpt|cmap", "", "", "h,i,u,U", "hinge,zinc,fromunit,tounit" }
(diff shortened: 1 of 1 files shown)


PaulWessel (approved):
Think this looks OK, -W in plot and plot3 has more things so those are exceptions?

rbdavis:
Paul, I think I'd like to make a further change to these #defines. Now that there are a few of them (with perhaps a bit more to come?), and two of the tokens (OUTGRID and PEN) are not suggestive of the short-option character, I'd like to change them all as follows:

GMT_C_CPT_KW
GMT_G_OUTGRID_KW
GMT_I_INCREMENT_KW
GMT_W_PEN_KW

This will make it easier for someone who is just looking at the ?_inc.h file to tell which short-option character is being translated to. What do you think? I suppose I can just make these changes myself right in this PR? I would also temporarily keep the current  GMT_INCREMENT_KW in place so as not to break any compilations until the rest of the *_inc.h files are updated.

PaulWessel:
Sue, do it here

PaulWessel:
Sure not sue

rbdavis:
Sorry, regarding your plot{3} comment, we can add additional modifiers to these defines as we work through the modules. I guess I would prefer to do that as opposed to going through all modules now just for this particular purpose. I can take a look at these particular ones today and and make changes in this PR if you like, though.

PR #7990, opened 2023-10-31
Support for longoption-related changes to gmt_constants.h

rbdavis:

Modify all longoptions-related *.[ch] files to replace references to GMT_INCREMENT_KW with GMT_I_INCREMENT_KW and make use of the new GMT_{C,G,W}_{CPT,OUTGRID,PEN}_KW definitions.

There are no user-facing longoption string changes in this PR.

Fixes #

**Reminders**


src/gmt_constants.h
```diff
@@ -223,8 +223,6 @@ enum GMT_time_period {
  * supported module documentation (e.g., manual pages) rather than these
  * non-user-facing translation entries.
  */
-/* next line is temporary, remove after all conversions to GMT_I_INCREMENT_KW are complete */
-#define GMT_INCREMENT_KW { '/', 'I', "increment|inc", "", "", "e,n", "exact,number" }
 #define GMT_I_INCREMENT_KW { '/', 'I', "increment|inc", "", "", "e,n", "exact,number" }
 #define GMT_C_CPT_KW	{ '/', 'C', "cpt|cmap", "", "", "h,i,u,U", "hinge,zinc,fromunit,tounit" }
```
joa-quim, line 227: Hmm, why 'zinc' and not just 'inc'?

rbdavis: > Hmm, why 'zinc' and not just 'inc'?
 
All of the longoption strings used here (and hopefully in all future longoption-related PRs that I am submitting for Paul's review only as opposed to the wider longoptions review group) were approved in past PRs, including zinc (from grdimage). In this case zinc was moved from the grdimage_inc.h file and incorporated into the new GMT_C_CPT_KW constant that was recently defined.

joa-quim: Or better, what is `zinc` for? I can't find its meaning in the docs.

joa-quim: Ah, I see. There is a **+i** in **-C** of `grdimage` but not in `makecpt` that I looked at.

rbdavis: From the grdimage man page:

> -C[cpt|master[+h[hinge]][+izinc][+u|Uunit] |color1,color2[,color3,…][+s<fname>]]


Name of a CPT file. Alternatively, supply the name of a GMT color master dynamic CPT [turbo, but we use geo for @earth_relief and srtm for @srtm_relief data] to automatically determine a continuous CPT from the grid’s z-range; you may round the range to the nearest multiple of zinc by adding +izinc.

joa-quim: But still prefer `inc` to `zinc`

PaulWessel: I think it is used in grd2kml.  Not to worry, the modules themselves will not allow +i if it is not a valid modifier in that module.  Remember, it is the 150 parsers that validates there.  Whether you lass --color=hot+10 or -Chot+i10 to grdimage you will get the same error message.  This just simplifies the longopt include files.

rbdavis: FYI, Joaquim, I believe you were not copied on a side-discussion that Paul and I had yesterday concerning these new GMT_?_???_KW constants. We agreed that to make these constants maximally useful they should incorporate as many modifiers as possible from all possible modules. In this case, -C is used by many modules to  reference a colormap, but only a small number of them offer the +i modifier. If we did not include +i in this definition then grdimage (which offers +i) would be unable to use the new constant. As Paul just said in his previous comment, if a user tries to call a module with +i which does not support +i, then that module will return the same error with longoptions as it would if called with shortoptions. It's on the user to read the manual pages and use valid modifiers. The translation tables in the code are NOT intended as user documentation.

joa-quim: Thanks Roger. I saw them passing but didn't look carefully (too many thinks passing by my eyes).

PaulWessel (approved):
Sorry!

PR #7996, opened 2023-10-31
Add longoptions support for gmtinfo.

rbdavis:

Python differences:

  No differences

Julia differences:

  -C numeric not used (not descriptive of per-column behavior)

Fixes #

**Reminders**


PaulWessel:
A few changes via commit:

-A changed to --report
-C changed to --columns

Not happy with --nearest-multiple for -T but cannot come up with a better alternative (yet). May resist later. Ideas?

src/longopt/gmtinfo_inc.h
```diff
@@ -24,10 +24,10 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
 	/* separator, short_option, long_option,
 	          short_directives,    long_directives,
 	          short_modifiers,     long_modifiers */
-	{ 0, 'A', "range|ranges",
+	{ 0, 'A', "report",
```
joa-quim, line 27: `report` what? A `range`.

src/longopt/gmtinfo_inc.h
```diff
@@ -24,10 +24,10 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
 	          "a,t,s",             "all,per_table,per_segment",
 	          "",                  "" },
-	{ 0, 'C', "per_column",        "", "", "", "" },
+	{ 0, 'C', "columns",        "", "", "", "" },
```
joa-quim, line 30: The reason a I named it `numeric` in Julia it's because that is what it really does. It returns a numeric vector in contra-position of no -C that return a text struct. And even in CLI  that difference is truly visible.

src/longopt/gmtinfo_inc.h
```diff
@@ -24,33 +24,25 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
-	          "",                  "",
+	{ 0, 'F', "counts",
+	          "i,d,t",             "totals,segments,segments_reset",
 	          "",                  "" },
-	{ 0, 'S', "",
+	{ 0, 'I', "inc|increment|spacing",
+	          "b,e,f,p,s",         "box,exact,fft,override,surface",
+	          "e,r,R",             "extend_box,adjust,extend_region" },
+	{ 0, 'L', "common_limits",     "", "", "", "" },
+	{ 0, 'S', "errorbar_space|forerrorbars", "", "", "", "" },
```
joa-quim, line 42: In Julia -S is called `for_error_bars`, not `forerrorbars`. For me `errorbar_space` is not intuitive.

src/longopt/gmtinfo_inc.h
PaulWessel: We can do alias with numeric|colums.

src/longopt/gmtinfo_inc.h
PaulWessel: Have you ever used **-S**? I think this was something Stephan Erickson (?) got us to put in decades ago. Does it really serve a purpose? Better to quietly remove but let it live under the hood.

joa-quim:
Nope, never used it and actually I don't even understand what it does _leave space for ..._??? I'm fine with hiding this.

PaulWessel:
The **-S** is now deprecated. Machinery is intact but synopsis, usage, and documentation does not mention it. Also updated gmtinfo_inc.h and removed it.

PaulWessel:
Also needs the new multi-directive scheme.

src/longopt/gmtinfo_inc.h
PaulWessel: Alias done.

PaulWessel (approved):

PR #8004, opened 2023-10-31
Add longoptions support for gmtselect.

rbdavis:

Python differences:

  Used min_area (-A) instead of area_thresh to better match manual page.

  Changed mask (-N) to feature_mask to differentiate from gridmask (-G)
    and changed gridmask to grid_mask.

  Used z_range (-Z) instead of z_subregion (compromise wth Julia in_range,
    perhaps better than either).

Julia differences:

  Replaced revert (-I) alias with invert. (revert has a different and in
    this case inappropriate meaning.)

  Changed mask (-N) to feature_mask to differentiate from gridmask (-G)
    and changed gridmask to grid_mask.

  Used z_range (-Z) instead of in_range (compromise wth Python z_subregion,
    perhaps better than either).

Fixes #

**Reminders**


src/longopt/gmtselect_inc.h
```diff
@@ -24,36 +24,26 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
+	{ 0, 'C', "distance|dist2pt",
+	          "",                  "",
+	          "d",                 "distance" },
+	{ 0, 'D', "resolution",
+	          "f,h,i,l,c",         "full,high,intermediate,low,crude",
+	          "f",                 "lower" },
+	{ 0, 'E', "boundary",          "", "", "", "" },
+	{ 0, 'F', "polygon",           "", "", "", "" },
+	{ 0, 'G', "grid_mask",         "", "", "", "" },
+	{ 0, 'I', "reverse|invert",    "", "", "", "" },
```
remkos, line 39: Since the preferred long option syntax for `-I` is `--invert`, should this not be the first option? I.e. `invert|reverse`?
Since we're interested also in an automated script to convert commands to long options, the order would matter when picking the preferred long option.

PaulWessel: The order matter, but not sure if it affects that parsing. @rbdavis will know, but aesthetically it should be invert|reverse.

PaulWessel:
The **-A** option appears in several modules (gmtselect.c, grdlandmask.c, grdmath.c, pscoast.c).  It sets minimum feature area, and optionally the range of desired feature levels [0-4 is default].  It then has a few modifiers. So not sure if --min_area is the best name here.  Maybe --coast_limits=min_area[minlevel/maxlevel][mods] or -gshhg-limits, or --area_limits?

src/longopt/gmtselect_inc.h
rbdavis: In long-to-short translation alias order is immaterial, i.e., the translator just crawls through the string set left to right and returns the first match it finds or no match at all. Short-to-long translation behavior is undefined now as it does not exist, but Remko's guess at this is certainly reasonable.

The reason these are ordered **reverse|invert** is simply that when I scrolled through the Julia gmtselect help the strings there were ordered as **reverse|revert**. I changed **revert** to **invert** because of the fact that the particular meaning of revert is not appropriate in this context.

src/longopt/gmtselect_inc.h
```diff
@@ -24,36 +24,26 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
+	{ 0, 'L', "dist2line",
+	          "",                  "",
+	          "d,p",               "distance,project" },
+	{ 0, 'N', "feature_mask",      "", "", "", "" },
```
joa-quim, line 43: I don't like the `feature_mask`. It tells me nothing when I read it. Much prefer to keep it just `mask`. We don't have to be consistent with `gridmask` (that I also prefer to `grid_mask`). We have had another case like this but don't remember which right now.

src/longopt/gmtselect_inc.h
```diff
@@ -24,36 +24,26 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
+	{ 0, 'Z', "z_range",
```
joa-quim, line 44: Why `z_` in `z_range`? I had it `in_range` in Julia and don't think it's worse.

src/longopt/gmtselect_inc.h
remkos: Yes, I was referring to the short-to-long translation that we discussed would also be implemented so we can modify the documentation. In that case, I assumed that for that purpose the preferred long option should come before the optional, so we would know which to use in the documentation.

src/longopt/gmtselect_inc.h
PaulWessel: grdlandmask has a **-N** option similar to **gmtselect**. It talks about `maskvalues`.  If we have to have the word "mask" in the long option then it remains unclear (what are we masking?).  Maybe use `--inout` ? Perhaps the silent majority of @GenericMappingTools/long-options-review have some suggestions? We should use the same phrase for both **gmtselect** and **grdlandmask**.  Then, **grdmask** is a bit different, perhaps, but `--inout` might work there too.

As for **-Z** I also think we don't want "z" in the name since can be any column.  And `range` is used for other things (e.g., `--range|series` for the many **-T** array options).  Maybe `--limits`? I think we restrict ourselves too much if we want a long-option name that starts with the short-option letter (here Z).

joa-quim:
Regarding **-Z**, I see no problems in repeating the word `range`, used in other options (**-T**) if that is also appropriate (`limits` is another example of potential multiple uses). But as I mentioned above, GMT.jl uses `in_range`

src/longopt/gmtselect_inc.h
seisman: Please note that range is a Python keyword, so PyGMT can't use it.

PaulWessel: We will minimize its use

PaulWessel:
Range is bad for python so always need an alias

rbdavis:
Please recall that **--limits** is defined in the common longopts .h file (as an alias for **--region**). Any user of any module who invokes **--limits** is going to get **-R**.

In general anyone suggesting any --_longoption_  option-triggering string should first inspect that common .h file for conflicts. (Conflicts with the common .h file for attribute or modifier longoption strings are not an issue.) The translator has been coded to search the common file first for all option-triggering string matches, then the module-specific ?_inc.h file. This is certainly an arguable decision, but it was made very early on (pre-me) and changing it now could be problematic without deeper consideration.

PaulWessel:
Good point. There is no denying that we will screw up at the module level by not keeping our eyes on the common option.  I am sure once we are through the 150 modules we will need to do specific checks to ensure that

We are consistent with long-names that do the same across many modules
That we do not step on the common options and find that --limits and --limits are for two different things that used to take -R and -F etc etc.

joa-quim:
We can't repeat long-options names, but we can reuse those words in directives.

src/longopt/gmtselect_inc.h
PaulWessel: OK, agreed, doing mask|mask_values

src/longopt/gmtselect_inc.h
PaulWessel: Finalised to do mask|maskvalues for -N, invert|reverse for -I, limits|in_range for -Z.

PaulWessel (approved):

PR #8010, opened 2023-11-03
DVC: Migrate from tracking directories to tracking files for doc images

seisman:
DVC provides two different ways to track files: (1) tracking individual files (one `.dvc` file for each .PS file); (2) tracking a directory (one `.dvc` file for the whole directory).

Currently, we're using the 2nd way, which causes many troubles for us (see https://github.com/GenericMappingTools/gmt/issues/5724#issuecomment-1783738234 for context).

In this PR, I've changed the way we tracked PS files in the `docs` directory to the first way. The biggest benefit is that we can have multiple PRs that work on updating documentation images without worrying about any conflicts.

In the future, if someone wants to update ex01.ps, then he needs to:
1. Update the `doc/examples/images/ex01.ps`
2. Run `dvc add doc/examples/images/ex01.ps` rather than `dvc add doc/examples/images/`

Please note that the way we track the PS files for tests are not changed.


seisman:
BTW, please ignore the "DVC diff" workflow failure, which is likely an upstream dvc issue.

doc/examples/images/ex01.ps.dvc
```diff
@@ -0,0 +1,5 @@
+outs:
+- md5: 08a2e5d2a7f9135d82ea0ae2ac99579f
+  size: 338598
+  hash: md5
```
seisman, line 4: @PaulWessel all the new `.dvc` files added by me have the `hash: md5` line, which is something new in newer dvc versions. Maybe it's easier if you can upgrade your dvc version.

PaulWessel (approved):
Sorry @seisman, been busy with many things today, not all GMT..

PR #8011, opened 2023-11-03
Convert test to doc figure for grdmix

PaulWessel:
Replace test script for **grdmix** with doc figure GMT_mixing.sh instead.
New fig referenced  in grdmix.rst. Removed the old test/grdmixmixcompose.sh
Aslo had to update DVC with the new fig as well as issues with ex15.  Now both pass.


seisman:
@PaulWessel Please review https://github.com/GenericMappingTools/gmt/pull/8010 first.

PR #8014, opened 2023-11-05
Add named constants to gmt_api,c

PaulWessel:
Instead of 0,1, and 2 we use **GMTAPI_MOD_NOTSET**, **GMTAPI_MOD_SPECIAL**, and **GMTAPI_MOD_NORMAL** to make code more readable in _gmtapi_extract_argument_.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -309,6 +309,11 @@ enum GMTAPI_enum_status {
 	GMTAPI_GOT_SEGHEADER 	= -1,	/* Read a segment header */
 	GMTAPI_GOT_SEGGAP 	= -2};	/* Detected a gap and insertion of new segment header */
 
+enum GMTAPI_enum_takes_mod {
+	GMTAPI_MOD_NOTSET	= 0,	/* Not assigned */
+	GMTAPI_MOD_SPECIAL	= 1,	/* Found = and possibly modifiers */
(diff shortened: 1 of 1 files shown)


joa-quim (approved):

PaulWessel:
Note this also includes the fix so please run your suite to see if any regressions elsewhere.

joa-quim:
It works, thanks, but where the hell is the fix? I only see variables renaming.

You can run the tests too. For that do ``] test GMT`` and don't be surprised if it sometimes crashes near the end. It's an Heisenbug that I'm not able to track involving pointers GMT<->Julia

PaulWessel:
Line # 13018 in gmt_api.c is the only actual change in behavior.

joa-quim:
Still can't see any difference. Before it was
```
if (!argument[0]) *takes_mod = 2;   /* Flag that option is missing the arg and needs it later */
```
now it is
```
if (!argument[0]) *takes_mod = GMTAPI_MOD_NORMAL;   /* Flag that option is missing the arg and needs it later */
```
but `GMTAPI_MOD_NORMAL` is equal to `2`

But never mind.

PaulWessel:
Well I mistakenly put the 2 directly in master (remember my confusion on GMT#master), so it was there before that PR - just changed to a name.

PR #8016, opened 2023-11-05
Warn that Latex cannot be used in panel tags

PaulWessel:
See #7958 for background.  This updates the documentation until we have a better solution.


doc/rst/source/subplot.rst
```diff
@@ -109,6 +109,8 @@ Optional Arguments (begin mode)
     shift relative to the tag box [default is **2p**\ /**-2p**] and *shade* sets the
     fill style to use for shading [default is **gray50**].
     Append **+v** to increase tag numbers vertically down columns [horizontally across rows].
+    **Note**: Currently you cannot use Latex expressions in the panel tag. The workaround is
+    to place such tags separately via :doc:text.
```
seisman, line 113:
```suggestion
    **Note**: Currently you cannot use LaTeX expressions in the panel tag. The workaround is
    to place such tags separately via :doc:`text`.
```

seisman (approved):

PR #8044, opened 2023-11-16
Be more careful to detect absolute time strings

PaulWessel:
Stuff like " ATL03_20200626034717_00120806_006_01.h5/beam: gt3l" should never be confused with an absolute time string.  This PR tightens that check but is still not fool proof.


src/gmt_io.c
```diff
@@ -3376,18 +3376,29 @@ GMT_LOCAL unsigned int gmtio_get_coltype_name_index (struct GMT_CTRL *GMT, unsig
 }
 
 bool gmtlib_maybe_abstime (struct GMT_CTRL *GMT, char *txt) {
```
seisman, line 3378: `src/t.ps` was mistakenly added.

joa-quim (approved):

PR #8046, opened 2023-11-18
Refit of already-longoptified modules to incorporate multi-directives

rbdavis:

This PR contains new long-option strings for 4 already-longoptified modules (blockmean, blockmode, fitcircle and gmtregress) which use the new multi-directive feature that was recently added. _Please review only the newly added multi-directive long-option strings (i.e., only the green code lines) in these modules_, and not the pre-existing long-option strings which have already been reviewed and approved in previous PRs.

Also, a modified gmtconvert_inc.h file is included here which basically only has some changed code indents from what was already reviewed and approved about a week ago. That earlier PR was not merged until earlier today because it also had multi-directive issues which are now resolved. I'm not sure why the code diff below is showing another old version of gmtconvert_inc.h (undoubtedly something stupid that I did), but the new code (green) is in fact correct and is essentially unchanged from what was reviewed and approved in that earlier gmtconvert PR.

This PR brings all of the already-longoptified modules up-to-date with respect to the new multi-directive stuff.

Fixes #

**Reminders**


rbdavis:
OK, I think I have fixed whatever weird was going on with the gmtconvert_inc.h file diff. It now seems to be completely removed from this PR, which I don't quite understand but I guess is OK. If it turns out there is a problem I will fix later in a separate PR.

PaulWessel (approved):
Since "my" module I chose a different set of names for -W.

Issue #8037, opened 2023-11-13
Trampled-upon image in an otherwise fine animation sequence


PaulWessel:
I am remaking the usual fly-over of the mid-ocean ridge with over 2000 frames. Exactly 2 frames do strange stuff.  It is always the same 2 frames.  I have isolated the movie job down to just make three frames with the middle one going crazy only.  Stripped off any labelling etc so it is just a **grdimage** plot with colors from a global static topo grid (and for this test I ignore the intensity grid).  The correct first frame looks like this:

![MOR_twist_0](https://github.com/GenericMappingTools/gmt/assets/26473567/8494d083-805f-41dc-a39c-5c98217145a6)

The second bad frame looks like this:

![MOR_twist_1](https://github.com/GenericMappingTools/gmt/assets/26473567/7a796100-dec4-45a1-9e10-8defccb77c06)

while the third frame is back to normal:

![MOR_twist_2](https://github.com/GenericMappingTools/gmt/assets/26473567/66b5f281-b6c0-4937-a3fa-e882cba40a72)

I even added **-x1** to ensure each frame was built separately so only one **grdimage** command was running at the same time.  No difference. This data set was the 30s earth_relief version. Lower resolutions do not have the problem. Also, if I run **grdcontour** instead of **grdimage** there are no odd offsets. If I run **grdimage** manually (not via movie) then the 2nd frame is fine.
Summary:

Affects **grdimage** when run via movie, and is reproducible for the same (2) frame out of 2300 frames

1. Not caused by job interference since one frame has to complete before the next starts (**-x1**).
2. Adding **-M** selects grayscale image - get same problem so not likely to be loop code (?)
3. Converting the script and running **batch** with **grdimage -A** instead of _PostScript_ and **ghostscript** yields the same problem; so not a **gs** issue.
4. Most likely something related to how both **movie** and **batch** launches scripts?

Very open to ideas of what to check for and how this can happen!  I have checked the intermediate script files and there is nothing that seems odd in any of them (e.g., the aparemeter files).

anbj:
Mind sharing the code?

anbj:
Thanks.

Getting:
```
$ gmt grdcut @earth_relief_30s_p.grd -Rd -Gearth_relief_p.nc
grdcut [NOTICE]: Remote data courtesy of GMT data server oceania [http://oceania.generic-mapping-tools.org]

[...]

PR #8054, opened 2023-11-22
Fix anim09 to first build global grid

PaulWessel:
Do not want @earth_relief_15s to be called again and again in main.sh. First assemble it in pre and do its gradient.


Esteban82 (approved):

Esteban82:
I wonder if this is still true: `The movie took ~6 hours to render on a 24-core MacPro 2013.`

PaulWessel:
I bet it is but have no time to retry right now.  Irritated by #8037.

Esteban82:
I also update the rst file.

Paul you will have to merge directly.

PR #8059, opened 2023-11-22
New pen meca and coupe -T syntax

### LINKED ISSUE #8057: Improve seis supplement options

Both **meca** and **coupe** have a **-T**_plane_[/_pen_] option but a few issues appear:

1. We use _nplane_ in **coupe** and _plane_ in **meca**.  Should use the same word, probably _plane_ since it is explained that it is nodal plane (long-option could be _nodeplane_ but not sure what is used in pyGMT and Julia since we have not worked on supplement long-options yet).
2. Old way to set _pen_ by separating _plane_ and _pen_ with a slash, which is bad since _pen_ could have a colour like 13/77/120. This means _plane_ must be given (even if optional) if `/pen` is added.  I suggest we honour this via backwards compatibility but change the option to do **-T**[_plane_][**+p**_pen_] like in some many other places in GMT.

This affects **meca** and **coupe**, and **sac** seems OK in this regard.

OK, @seisman?

PaulWessel:
Clean up **meca** and **coupe** syntax and their  parsing of **-T**[_plane_][**+p**_pen_] as well as quiet backwards parsing of the deprecated **-T**_plane_[/_pen_]. Docs look good and the tests pass but not sure if we have many tests are involving **-T**.

Closes #8057.

### DIFF
--- a/doc/rst/source/supplements/seis/coupe.rst
+++ b/doc/rst/source/supplements/seis/coupe.rst
@@ -257,7 +257,7 @@ Optional Arguments
 
     *2*: only the second nodal plane is plotted.
 
-    Append **++p**\ *pen* to set alternative pen attributes for this feature 
+    Append **+p**\ *pen* to set alternative pen attributes for this feature 
     [Default pen is as set by |-W|].
 
     For double couple mechanisms, the |-T| option renders the beach ball transparent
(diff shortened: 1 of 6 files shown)


seisman (approved):

yvonnefroehlich:
Seems like we also have to update the PyGMT [documentation for `Figure.meca`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.meca.html) after GMT 6.5 has been released.

PR #8065, opened 2023-11-23
Add longoptions support for gmtspatial.

### LINKED ISSUE #8064: Add longoptions support for gmtspatial.

Add longoptions support for gmtspatial.

rbdavis:

Python differences:

  No python module.

Julia differences:

  Used in_polygons for -N instead of in_polyg (awkward abbreviation
  and inconsistent with -F force_polygons).

  Used spatial for -S instead of polygons. Almost all options are
  related to polygons, so using 'polygons' alone is not helpful in
  distinguishing between them.

Fixes #8064

**Reminders**


PaulWessel (approved):
See comments

joa-quim:
Well, I agree with your -S `polygons` comment that's all about polygons but that applies exactly to `spatial`. So not sure it's better, and the other name already exists.

src/longopt/gmtspatial_inc.h
```diff
@@ -25,53 +25,44 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
-	{ 0, 'S', "",
-	          "",                  "",
+	{ 0, 'Q', "centroid|area|length",
 	          "",                  "",
+	          "c,h,l,p,s",         "range|limits,header,lines,close,sort",
 		  GMT_TP_STANDARD },
-	{ 0, 'T', "",
-	          "",                  "",
+	{ 0, 'S', "spatial",
+	          "b,h,i,j,s,u",       "buffer,hole,intersection,join,split,union",
```
joa-quim, line 59: The directives (some) are also different in Julia. GMT.jl has `holes` instead of `hole` and `dateline` instead of `split`. I think `dateline` is better because it clearly indicates where the split occurs.

PaulWessel: Tend to agree with this. Could do alias holes here but I like dateline.

PaulWessel: We can add alias holes but not that all the others are singular nouns so _holes_ stand out.

PaulWessel: Best if GMT.jl now adds the hole alias to match CLI.  I did the dateline change and added alias holes.

joa-quim:
Right, but what it detects are `holes`, not just one `hole`: The man construction ``and hole polygons`` is a bit dubious (if I'm interpreting it correctly).

PaulWessel:
I updated the **-S** documentation.  Le the know if this is clearer:

doc

PaulWessel:
> Yes, much better. But now you raised me another doubt. What is meant by _reverses them_? You mean make them of opposite _wise_ of the parent polygon (clockwise or counter-clockwise)?

Yes, any holes will get the opposite handedness as the parent (coco vs counter-clock).  Il'' fix the docs for h to say

- **h** identifies polygons that are holes of other polygons (and flags and reverses their handedness to be the opposite of the outer polygon).
- 
Any other things before I add this?

joa-quim:
No, sounds good.

PaulWessel:
A bit dumb to list **-Si**, **-Sj**, **-Su** when not implemented (after many years).  Leaves two options:

1. Try to implement this (but cannot do so before 6.5 as it will take a while).
2. Put those things under an #if 0 ...#endif so user dont read those and get confused.

Those directives were clearly aspirational but here we are...  I think there is a C++ library out there but perhaps there is a C library somewhere we could use?

Issue #8036, opened 2023-11-13
Clipping fails for grdview


PaulWessel:
Reduced the case reported on the [forum](https://forum.generic-mapping-tools.org/t/strange-data-cuting-with-grdview/4390) reduced to this simple example:
```
gmt begin 3dbug png
	gmt grdview  @earth_relief_10m -JB-90/21/19/23/10c -R-92/-87/19/23 -p-30/90 -Cglobe -B -BWSen -Qi50
	gmt grdimage @earth_relief_10m -p -Cglobe -B -BWSen -X12c
gmt end show
```
Same plot but with two different modules.  The left plot uses **grdview** which has a bug while **grdimage** does not.

![3dbug](https://github.com/GenericMappingTools/gmt/assets/26473567/dcb9cf47-317c-4424-8c06-2771cefb9d05)

Both **-Qi** and **-Qc** give the same problem, but doing tiles (**-Qs**) is unaffected. Suspicion is that the computed mask polygon in **grdview** is not properly 3-D projected, while **grdimage** does not have such a thing.

joa-quim:
I think this was fixed, right?

PaulWessel:
No, and on the road.

PaulWessel:
Strange things going on.  I ran the **grdview** twice: One with **-Qc** (image on left) and another with **-Qs** (surface polygons on right). This yields correct for **-Qs** and a bad clip(?) for **-Qc**:

![3dbug](https://github.com/GenericMappingTools/gmt/assets/26473567/c641c60c-1dba-4b69-aa37-9b5e080bf018)

However, if I edit the two PostScript files and remove all the stuff that plots the color surface or color image, we are left with basic setup, clip paths, and basemap. The diff of those edited files show no differences.  So same clip path, which works fine for surfaces but not for tiling or images.  Conclusion must be that the images and the tiling are not doing the right thing, even though they happen in separate branches in **grdview** code. Using polygon tiles (**-T**) shows the same bad clipping in **grdview** (but not in **grdimage** as the top plot shows).

Clearly, the pixels that do show (on the left image) are in the right place (comparing to surface on right) so it would seem the loop over pixels or tiles is skipping those locations since the clip is the same?  Adding a vertical scale (**-JZ**3c) correctly strengths this in the z-dimensions but the clipping is the same. Any thoughts on how to figure this out is appreciated.

PaulWessel:
[...]

PR #8072, opened 2023-11-24
Fix lack of xyz->xy conversion for graticules in grdview -T

PaulWessel:
See #8036 for background. We use _gmt_graticule_path_ to uild polygons for each grid node but in **grdview** we need to project them to 3-D coordinates (x,y,z to x', y')

Now, 

`gmt grdview  @earth_relief_10m -JB-90/21/19/23/10c -R-92/-87/19/23 -p-30/40 -Cglobe -B -BWSen -T -png map`

yields 

t

`gmt grdview  @earth_relief_10m -JB-90/21/19/23/10c -R-92/-87/19/23 -p-30/40 -Cglobe -B -BWSen -Qc -png map`

still fails (below) but that is a separate branch in **grdview** that needs another fix by another PR.

map

### DIFF
--- a/src/gmt_plot.c
+++ b/src/gmt_plot.c
@@ -10829,7 +10829,7 @@ struct GMT_POSTSCRIPT * gmt_get_postscript (struct GMT_CTRL *GMT) {
 	return (P);
 }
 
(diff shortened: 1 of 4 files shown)


joa-quim (approved):
Congrats

PaulWessel:
I went back to **grdview** 6.0 code and whatever that was it had the same bug, so nothing recent other than a new function that had the same problem.

PR #8074, opened 2023-11-24
Use named constants rater than 0-2

PaulWessel:
We have an _angle_type_ that indicates what angle we should use along a contour for label placement or decorated symbol placement. Hard to remember what 0-2 means so this PR just introduces named enums.  No change to any code or behaviour (yet).


Esteban82 (approved):
Looks fine. Blind approval.

PR #8075, opened 2023-11-26
Make batch -T doc use the same format as for movie -T

PaulWessel:
When there are many modifiers it is much cleaner to have the man page sort these with separate bulleted sections. Using the same layout for **movie** yields this updated **-T** for **batch**:

batch

### DIFF
--- a/doc/rst/source/batch.rst
+++ b/doc/rst/source/batch.rst
@@ -62,23 +62,26 @@ Required Arguments
 .. _-T:
 
 **-T**\ *njobs*\|\ *min*/*max*/*inc*\ [**+n**]\|\ *timefile*\ [**+p**\ *width*]\ [**+s**\ *first*]\ [**+w**\ [*str*]\|\ **W**]
-    Either specify how many jobs to make, create a one-column data set width values from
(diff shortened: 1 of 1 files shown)


seisman:
This doesn't look right:
image

PaulWessel:
> This doesn't look right: image

Agreed, but it basically copied the exact same stuff from movie.rst, which looks good. Not sure how that can happen?

yvonnefroehlich:
> > This doesn't look right: image
> 
> Agreed, but it basically copied the exact same stuff from movie.rst, which looks good. Not sure how that can happen?

Maybe `.. include:: explain_array.rst_` has to added now:
https://github.com/GenericMappingTools/gmt/blob/4fddca00719d998dc09dfd88662ffffaec659c23/doc/rst/source/batch.rst?plain=1#L180-L186
like in movie.rst: 
https://github.com/GenericMappingTools/gmt/blob/4fddca00719d998dc09dfd88662ffffaec659c23/doc/rst/source/movie.rst?plain=1#L349-L357

PaulWessel:
Thanks @yvonnefroehlich - now updated.

seisman (approved):

PR #8083, opened 2023-11-26
Improve the layout of gmtinfo's -I option in usage and docs

PaulWessel:
Again, spread meaning of directives and modifiers over several lines.


joa-quim (approved):
You can aprove these yourself too.

PR #8084, opened 2023-11-26
Improve man page for gmt logo

PaulWessel:
Also avoid repeat of "needs both **-R -J**" in -D docs.  However, the stuff beneath **-D** is tabbed one to many to the right.  I cannot see why as it looks exactly as **-F** which does the right indentation.  Anyone? See difference in alignment for **-D** and **-F** below:

logo


joa-quim (approved):
Mysterious alignment shifts are sometimes caused by tabs disguised as spaces.

PR #8094, opened 2023-11-28
Use math for special symbols

PaulWessel:
Basically use Latex codes for >= and <=.


doc/rst/source/grdclip.rst
```diff
@@ -27,8 +27,8 @@ Description
 -----------
 
 **grdclip** will set values < *low* to *below* and/or values > *high* to
```
seisman, line 29: Why not use `≤` and `≥`?

seisman: I meant to use ≤ instead of latex codes

PaulWessel:
I guess that is option-< etc? On the road now so please fix if you have
time.

PR #8099, opened 2023-11-28
Enable updated Accelerate headers for Darin kernel >= 22.4

PaulWessel:
Getting tired of the tantrum Xcode has added if we call the deprecated `dgemm_` without including the new headers via compiler flag **-DACCELERATE_NEW_LAPACK**. This PR has ben following changes:

1. `cmake/ConfigUserAdvancedTemplate.cmake`: Adds  **-DACCELERATE_NEW_LAPACK** to the compiler flags if we are on Darwin >= 22.4
2. `CMakeLists.txt`: Determine if we are on version 22.4 and if so set Cmake variable **NEW_ACCELERATE_LAPACK** to 1 (else 0).
3.` gmt_vector.c`: Use `#ifdef ACCELERATE_NEW_LAPACK` to cast our parameters using Apple's new `dgemm_` definition.

This PR now compiles quietly for me on macOS Sonoma. Ideally:

- Someone who knows Cmake better than me (@seisman, @remkos ?) should see if the new code in the advanced template can be simplified.
- Give it a try if you have an older kernel since I am unable to test that.
- Also build on non-macOS to ensure nothing funny happened.


remkos (commented):
As you have notices, I needed to make some changes:
- Renaming NEW_ACCELERATE_LAPACK to ACCELERATE_NEW_LAPACK did not have the result I expected, though I should have realised that. Nonetheless, consistency is practical.
- Putting stuff in ConfigUserAdvancedTemplate.cmake did not make much sense. I would not have looked into the template myself. It should be part of the CMakeList hierarchy.
- The whole stuff needed to be moved to src/CMakefile.txt to have effect, because only there it is determined if LAPACK is even there.

remkos:
It works now for me.
@PaulWessel and @seisman please check if this works for you.
It should probably work "out of the box" with a simple `make`. Otherwise redo a `cmake` in your build directory, then `make`.

PaulWessel:
I guess I copied the new stuff from that template to my own advanced cmakefile and as far as I can tell it found everything and compiled as expected.  So I think it works **IF** you update that file.

I agree that users on recent macs who already have GMT via git will not notice the change since they may have an advanced file that is not getting updated.  I can look at this tomorrow pm unless you do something first.

remkos:
It works now for me without having touched my ConfigUser.cmake, which is what I wanted to achieve, and as you pointed out is what users would likely expect. We do not need to bother them with switching logic in their user configuration.

I have nothig to change. Just wait for @seisman and you to see if it works for you.

PaulWessel:
Great, works fine for me - yours is the right solution.  All tests still pass that passed before.

seisman (approved):

PR #8109, opened 2023-11-29
Fix bug in greenspline for -E -C

PaulWessel:
Wherever we need to evalute the solution we compute the distance been two points and if nonzero we evaluate the Green's function.  But, in one place we placed the calculation after the zero-check and of course r was always zero and the calculation of the surface did not happen.

Hopefully not more issues but this one is obvious so we fix it in this simple PR.

### DIFF
--- a/src/greenspline.c
+++ b/src/greenspline.c
@@ -2704,8 +2704,8 @@ EXTERN_MSC int GMT_greenspline (void *V_API, int mode, void *args) {
 						double dev, rms = 0.0, chi2_sum = 0.0;
 						for (j = 0; j < nm; j++) {	/* For each data constraint */
 							for (p = 0, wp = 0.0; p < nm; p++) {	/* Add contribution for each data constraint */
+								r = greenspline_get_radius (GMT, X[j], X[p], 2U);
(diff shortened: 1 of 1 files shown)


Esteban82 (approved):
Thanks. It works!

PaulWessel:
Misfit file now looks like this so I think it is now working (again):

```
# Command : gmt greenspline geo.txt -Sc -C -Ggeo.grd -Z2 -Emisfit.txt -R158:00W/156:58W/18:00N/19:40N -I1m
# eigenno	eigenval	var_percent	rms
0	59177476.7	65.8905533939	752.198173843
1	37281370.3447	92.0418967819	752.120969527
2	19324372.0562	99.0681055505	752.118091663
3	6863523.97604	99.9544541832	752.118066259
4	1200156.86214	99.9815552443	751.647630375
5	641289.879832	99.9892930717	749.851543745
6	527868.265693	99.9945358486	717.129582791
7	390481.243194	99.997404717	707.146351065
...

```

Esteban82:
Really? I have different values. I didn't compare it with the original animation.

https://github.com/GenericMappingTools/gmt/assets/17908951/765b098c-c8a5-4d65-9e9c-0f2f3e3df0da

PR #8118, opened 2023-12-02
Multi-directive support for gmtselect.

rbdavis:

Add multi-directive support for **gmtselect**. The bulk of the changes here are the additions of the GMT_TP_* bitmask fields to the end of each table entry. Hopefully this is the last (maybe?) pre-existing longoptions file requiring this large-scale refit.

Please note that the only changed longoption strings in this PR are:

(1) New multi-directives **"circle,polygons,zero,line,rectangle,gridmask,zrange"** added to -I option. Please do not review any of the other unchanged and already-approved longoption strings in this PR.

(2) The -Z **limits** alias was removed. As noted before **--limits** is a longoption alias already in use for -R within the gmt_common_longoptions.h table. Please do not use **limits** (or anything else used within gmt_common_longoptions.h) as a **--longoption** trigger string.

Fixes #

**Reminders**


src/longopt/gmtselect_inc.h
```diff
@@ -23,27 +23,36 @@
...
-	{ 0, 'F', "polygon",           "", "", "", "" },
-	{ 0, 'G', "gridmask",          "", "", "", "" },
-	{ 0, 'I', "invert|reverse",    "", "", "", "" },
+	          "f",                 "lower",
+                  GMT_TP_STANDARD },
+	{ 0, 'E', "boundary",          "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'F', "polygon",           "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'G', "gridmask",          "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'I', "invert|reverse",
+	          "c,f,g,l,r,s,z",     "circle,polygons,zero,line,rectangle,gridmask,zrange",
```
joa-quim, line 44: I prefer a simple `mask` instead of `gridmask`

rbdavis: I squashed and merged whatever changes were made to that PR when it was approved.

PaulWessel: OK so there are two "mask" concepts here:
1. -Ggridmaskfile>  This says it is an input grid that has NaNs: "Pass all locations that are inside the valid data area of the grid gridmask. Nodes that are outside are either NaN or zero.".  It makes sense that NaN means outside, but 0? What if you have a data set that has a range from -4300 to +1250 and regions of NaN.  A z = 0 would not mean outside then? I think we should fix this and only accept NaN and update the statement.
2. -Nvalues.  Here we mean GSHHG polygons and if points that fall inside or outside etc should be skipped or kept.  Not really mask, I think?  Or maybe poly_mask vs grid_mask?

rbdavis: As far as I can tell from examining the old PR #8004 (and my github skills are not the grewatest), it looks to me like PR #8004 was approved with -G using --gridmask.

rbdavis: And FYI, gmtselect's -N option is using --mask.

PaulWessel: I thought the -I was settled.  Talking about -G and -N, no?

joa-quim: > that are inside the valid data area of the grid gridmask. 

In fact when I quickly re-read this I did not understood it. As I said in my previous/simultaneous comment, a mask is always a matrix of 0's & 1's, even when we pass a polygon.

joa-quim: > And FYI, gmtselect's -N option is using --mask.

So that settles it. -G has to remain `gridmask`. Regarding **-Is** since it refers **-N** it should be `mask` no?

joa-quim: > I thought the -I was settled. Talking about -G and -N, no?

No. This PR is about **-I**

rbdavis: My vote in general is to not revisit previously approved longoption strings, because after you guys approve them I then write up the translation tests accordingly. If there's a problem with new strings just submitted for review (like the new -I gridmask directive) that's a different issue, of course.

PaulWessel: Generally fine.  My thinking is that this first go-around will leave a few things we will realise e need to revisit once through. But it is also hard to remember 150 longopt things so when we detect something that probably should change, cannot just wait 2 months and remember that.

As for here, -I is invert|reverse and we are done, no?  Nothing to do with masking.

rbdavis: The only change to -I proposed here is the addition of the multi-directive strings "circle,polygons,zero,line,rectangle,gridmask,zrange". The "invert|reverse" trigger strings are not being changed.

PaulWessel: OK, I misunderstood.  But s -> grid mask does not make sense since there is no grid involved, just polygons that container or do not contain the point.  If anything g should be grid mask since it refers to the grid given by -G. And by the above, I am not sure about checking for zero. Think that should be NaN but that is a module code change I can do later. The s, pointing to -N, is strange.  I must have good up.  All the other lower-case letter match directly to the upper-case option they invert. Why didn't;t I use n for -N here and we could have n -> mystery word starting with n?

PaulWessel: I wonder if I missed looking at those directives too. The s should be n to match teh lower/upper combo for all the others. The g should be grid mask, the r should be region, and the n should be something that ideally starts with n, but might be hard.  Since how you give the arguments to -N there is no standard form that is inverted.  You could give -Nk/s and that is the inverse of the default -Ns/k.  So inside/outside is not clear at all.  Maybe notout ?

PaulWessel (approved):
Sorry so slow.

PR #8124, opened 2023-12-04
Only reset the column type at level zero.

### LINKED ISSUE #8115: grdimage -I breaks frame annotations.
This one was really waiting for us to release 6.5 to then show up.

With **-I** it doesn't plot the annotations
```
grdimage @HI_topo_02.nc -R160/20/220/30+r -JOc190/25.5/292/69/12 -Ba -I -png lixo
```
lixo

joa-quim:
This doesn't affect the CLI because it will end anyway but makes a difference for externals that keep the API alive between module calls.

Fixes #8115


PaulWessel (commented):
See my points on #8115 first.

PaulWessel:
Now merged so I will add my wrapper checks for the future.

joa-quim:
Ok, but remember that AFAIK all conditions applied to Julia applies to MEX too

PaulWessel:
Yes, I will just use "external" for now.

PR #8137, opened 2023-12-05
Remove unused variables

remkos:

As usual: some code cleanup to remove unused variables that cause warnings during compilation.


PaulWessel (approved):
Thanks for the janitorial effort!

PR #8138, opened 2023-12-05
Upgrade grd2cpt like makecpt and improve docs

PaulWessel:
This PR does several things:

1. It adds **-F** directive **g** for grayscale. While a couple of GMT modules (e.g., **grdimage**, **grdview**) has a **-M** option (for converting colors to black and white television via the YIQ transformation), we cannot actually make sure CPT output.  With **-Fg** we now get a single grayscale column wether it was originally gray or had to be YIQ converted..
2. While **makecpt** had the **+k**_keys_ modifier, **grd2cpt** did not.  No good reason other than obviously a keyed categorical CPT cannot be used with a grid since the look-up values are always numbers.  But users may still want to make a categorical CPT with keys via **gd2cpt** that can then be used for non-grid plotting. Thus I duplicated the **+k** stuff from **makecpt** to **grd2cpt**.
3. Because they now have identical synopsis and usage I made helper functions that are called by both instead of maintaining duplicate usage messages.  The **-F**_arg_ string is now also a constant in `gmt_constants.h`
4. Same deal with the docs: Introduced an include file (`explain_cpt_output.rst_`) to ensure we say the same thing in the two modules.  The updated **-F** docs now look like this:

F

### DIFF
--- a/doc/rst/source/explain_cpt_output.rst_
+++ b/doc/rst/source/explain_cpt_output.rst_
@@ -10,7 +10,7 @@
     - **r**: Output *r*/*g*/*b* codes only (even if a color name apply).
     - **x**: Output *#rrggbb* hex codes.
 
-    Optionally or alternatively, append this modifier:
+    Optionally or alternatively, append these modifiers:
 
     - **+c**: Write discrete palettes in categorical format. If *label* is appended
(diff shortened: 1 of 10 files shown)


remkos (approved):
Compiles and works for me.

PR #8161, opened 2023-12-08
Int16 tif files are more likely a grid than an image

joa-quim:
When deciding if a .tif is a grid or image the Int16 (short int) case was not considered, which resulted in the file being tagged as image. This gave errors like

grdcut wc2.1_10m_prec_01.tif=gd:GTiff ...
...
grdcut [ERROR]: Unsupported image format. Supported formats are: BMP,GIF,JPG,PNG & TIF
grdcut [ERROR]: Alternatively, append :<driver> for a valid GDAL driver

With this PR the grdcut command succeeded.


PaulWessel (approved):
Might it be useful to add "&& I->header->n_bands ==1)" to your GMT_SHORT test?

joa-quim:
OK, added.

Issue #8158, opened 2023-12-07
Greenspline. Add tab in Misfit evaluation


Esteban82:
I want to plot the misfit evaluation of a grid made with greenspline in a graph.

`gmt greenspline Perfiles?.txt -i0,1,4 -R$R -I$inc -Sc -V -Z1 -Ggsplines.nc -Emisfit.txt 2> misfit_stderr.txt`

So, first I use `-Emisfit.txt 2> misfit_stderr.txt`  to save the info of the whole grid into in a file.

Then I use `grep`:

```
grep "Misfit evaluation" misfit_stderr.txt
greenspline [INFORMATION]: Misfit evaluation: N = 217	Mean = -7.35372e-15	Std.dev = 3.84514e-14	RMS = 3.90611e-14
```
With this command I can plot the Mean info (i.e. Mean = -7.35372e-15) in the graph. So far so good.

	grep "Misfit evaluation" misfit_stderr.txt | cut -f 2 | gmt text -F+cTL+f9p -Dj0.1c/18p

### Issue

My problem is when I want to plot "N = 217" in the graph. With `-f 1` doesn't work

```
grep "Misfit evaluation" misfit_stderr.txt | cut -f 1
greenspline [INFORMATION]: Misfit evaluation: N = 217
```
Also try: 

```
grep "Misfit evaluation" misfit_stderr.txt | cut -d" " -f5-7
N = 217 Mean

```
### Solution? 

So, I think that this could be solve if I add a `\t ` before "N" in the lines 2533 and 2535 of https://github.com/GenericMappingTools/gmt/blob/master/src/greenspline.c

			`GMT_Report (API, GMT_MSG_INFORMATION, "Misfit evaluation: N = %u\tMean = %g\tStd.dev = %g\tRMS = %g\n", nm, mean, std, rms);`
			
			
Is my solution ok? Should I make a PR?

PaulWessel:
Sure, you can add that tab.

joa-quim:
We normally separate second and on with tabs in an attempt to align columns vertically. With spaces that's very difficult, but ofc this makes no difference for a single plot line.

Esteban82:
It is not very clear if I should put space or tabs. I would prefer tabs.

joa-quim:
Fixed by a new option

PR #8162, opened 2023-12-08
Add new greenspline option -K for statistics output to file or stdout

PaulWessel:
See #8158 for some background.  Rather than messing with the verbose output, this PR instead adds **-K** option that will wrote a data table with header and 7-8 values which makes it easier to access. Updated docs as well. Suggest @Esteban82 gives this a try and report back if there are any issues or request. You will see I basically write both the misfit and variance statistics to the file (variance first, then misfit) and then if **-W** is active we append one more output (chi^2) at the end.  This ensures that the first 7 columns are in a fixed order and the 8th optional column is only there if **-W** is set. @joa-quim, done merge this into master until @Esteban82 has given it a try later today.

### DIFF
--- a/doc/rst/source/greenspline.rst
+++ b/doc/rst/source/greenspline.rst
@@ -19,6 +19,7 @@ Synopsis
 [ |SYN_OPT-D3| ]
 [ |-E|\ [*misfitfile*] ]
 [ |-I|\ *xinc*\ [/*yinc*\ [/*zinc*]] ]
+[ |-K|\ [*file*] ]
 [ |-L|\ [**t**][**r**] ]
 [ |-N|\ *nodefile* ]
 [ |-Q|\ [*az*\|\ *x/y/z*] ]
@@ -198,6 +199,14 @@ Optional Arguments
 
 **-I**\ *xinc*\ [/*yinc*\ [/*zinc*]]
     Specify equidistant sampling intervals, on for each dimension, separated by slashes.
+
+.. _-K:
+
+**-K**\ [*file*]
+    Write misfit and variance evaluation statistics to *file* (or *stdout* if no file is given).
+    Output order is *Data Model Explained(%) N Mean Std.dev RMS*; if |-W| is used we
+    add *chi^2* as a final 8th column.
+    If |-K| is not used then |-V|\ **i** will report these via two verbose messages.
 
 .. _-L:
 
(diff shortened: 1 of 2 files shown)


joa-quim:
Isn't -K a global option?

joa-quim:
And where is that `#define GMT_N_API_ENUMS 263` coming from that makes the Validator check to fail?

PaulWessel:
> Isn't -K a global option?

No, see **movie -K** for instance.  But, I agree there is no point using -K if there are other letters available, e.g., **-F**, **-M** .  I will change it to -F instead.

PaulWessel:
If we add an enum we must run src/gmt_make_enum_dicts.sh to update that include file for the API. Idid I forget?

Esteban82:
It looks fine. I test it with this:

`gmt greenspline -R0/7.2/-0.2/7 -I0.05 -GMM1993.grd @Table_5_11.txt -Sr0.99 -Z1  -Vi -Kstatistics.txt -Emisfit.txt
`


Statistics.txt:
```
# Misfit & Variance Evaluation: Data	Model	Explained(%)	N	Mean	Std.dev	RMS
3.5767e+07	3.5767e+07	100	52	-5.24708e-14	7.94742e-14	9.45932e-14
```

### Small Issue

The creation of that file depends on `-E`. So, if I use the command below, the file is not generated. 

`gmt greenspline -R0/7.2/-0.2/7 -I0.05 -GMM1993.grd @Table_5_11.txt -Sr0.99 -Z1  -Vi -Kstatistics2.txt `

So, maybe should be clarify in the docs? Or maybe we should a modifier to `-E` instead of a new letter?

PaulWessel:
Sorry, forget.  Just pushed it to master now. It is 263.

PaulWessel:
@Esteban82, please update build and check again since there were several changes.

PaulWessel:
Fixed.

Esteban82:
Ok. BTW I will be away from my PC for the following days.

PaulWessel:
OK, the new feature is now set via **-E**[_misfitfile_][**+r**_reportfile_]. The benefits are

No need to messing with complicated **-F** form the olden days.
No need to add a new option, just a modifier
This also means we would not have to set **-E** in addition to the initial **-K** or **-M** to do this since it is now a modifier of **-E** so **-E** is obviously set.
green spline actually has an unimplemented **-M** for passing in previous eigenvalues.  May happen at some point and then it is nice that **-M** is not taken already.

Give @Esteban82 is off, perhaps @joa-quim can have a look. Should be harmless since only about a report to a file.

Fixed synopsis message and a few issue related to freeing files that were forgotten in gpsgridder and green spline.  gpsgridder now has the same -E modifier etc.

Issue #7591, opened 2023-07-02
Rotating the symbols of a decorated line does not work ("+aangle") 


yvonnefroehlich:
**Description of the problem**

Based on the documentation at https://docs.generic-mapping-tools.org/dev/plot.html#s it should be possible to rotate the symbols of a decorated line by appending **+a**_angle_:

> +aangle
> Force symbols at a fixed angle, +an for line-normal, or +ap for line-parallel [Default].

However, so far this does not work for me. For quoted lines, it works to rotate the label by appending **+a**_angle_.
Related PyGMT comment https://github.com/GenericMappingTools/pygmt/pull/2564#issuecomment-1614331752.

**Full script that generated the error**

_Input data_: [test_data_deco_line.txt](https://github.com/GenericMappingTools/gmt/files/11928926/test_data_deco_line.txt)

```
gmt plot test_data_deco_line.txt -R0/10/-1/1 -JX10c/3c -B -S~d2c:+ss1c+a20 -W1p,black -png deco_lines_rot_gmt
```

**Full error message**

No error message occurs.

**Actual outcome**

![deco_lines_rot_gmt](https://github.com/GenericMappingTools/gmt/assets/94163266/b7c63ff1-2c15-47a5-bbc3-3d2e6b0c103b)

**Expected outcome**

The squares along the line should be rotated by 20 degrees (counter-clockwise from the horizontal); similar to quoted lines:
```
gmt plot  test_data_deco_line.txt  -R0/10/-1/1 -JX10c/3c -B -Sqd1c:+ltext+a20 -W1p,black -png quot_lines_rot_gmt
```
![quot_lines_rot_gmt](https://github.com/GenericMappingTools/gmt/assets/94163266/86717a41-13c2-4da0-98b9-79ef36dace0a)

**System information**

* Operating system: Windows-10-10.0.19045-SP0
* GMT version (`gmt --version`): 6.4.0 as well as 6.5.0_95c3431_2023.06.26

PaulWessel:
Mystery so far.  At the end of _gmtsupport_add_decoration_ we have

```
	/* Build record with Cartesian (hence col = GMT_Z) "x y angle symbol" since we are using -Jx1i */
	S->data[GMT_X][S->n_rows] = L->x;
	S->data[GMT_Y][S->n_rows] = L->y;
	S->data[GMT_Z][S->n_rows] = gmt_M_to_inch (GMT, G->size);
[...]

PR #8182, opened 2023-12-11
Fix bug and allow +apangle for decorated lines

PaulWessel:
This fixes the bug in #7591. While at it, we allow **+ap**_angle_ to let the symbol angle deviate by _angle_ from the line's angle. And **+a**_angle_ still sets a fixed angle regardless of angle contortions.

Original post with **+a**20:

deco_lines_rot_gmt2

New **+ap**20, same line:

deco_lines_rot_gmt


yvonnefroehlich:
Thanks @PaulWessel for fixing this bug. Great that this will be included in GMT 6.5.0 🙂.

PR #8188, opened 2023-12-12
Long-lasting bug with seniority finally killed in gmt_crossover function

PaulWessel:
The test `spline_5.sh` has failed (at least on macOS) for the longest time.  Finally decided to debug the sucker and found that the missing contour annotations had to do with round-off.  If fact there were two missing but only one was seen as a bug.  Problem was that if the crossing was exactly on a data noted then our test to see if it was outside the segment failed because we checked along-track "time":

`tx_a < ta_stop && tx_b < tb_stop`

Clearly, if _tx_a_ or _tx_b_ equal _ta_stop_ or _tb_stop_ then we still fail (cutting at the mode). Solution is to allow one of the ends to be equal:

`if (tx_a < ta_stop && tx_b <= tb_stop)`

Now, the contour map places all the contours crossing the imaginary line set by **-Gl**:

gspline_5

The topmost 40 label in the bottom plot was missing before, but so was the 30 contour.  Now all is well. A bug with seniority bites the dust!


WalterHFSmith:
Nice to see a senior bug bite the dust. Well done! -W

PR #8198, opened 2023-12-16
Enable longoptions support for grdcut.

rbdavis:

Python differences:

  No differences.

Julia differences:

  No differences.

Fixes #

**Reminders**


src/longopt/grdcut_inc.h
```diff
@@ -25,29 +25,27 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
-		  GMT_TP_STANDARD },
-	{ 0, 'F', "",
-	          "",                  "",
+	{ 0, 'D', "dryrun",
 	          "",                  "",
+	          "t",                 "trailer",
 		  GMT_TP_STANDARD },
-	{ 0, 'G', "",
-	          "",                  "",
+	{ 0, 'E', "extract",
```
joa-quim, line 32: I think `rowslice` & `colslice` is better here.

PaulWessel: Well, **-E** can be **--extract** and then the two directives can be columns or row, e.g.

```
--extract=column:54:30W
--extract=row:30N
```

No? We are not replacing -E with two new long-options.

joa-quim:
`zrange` is good.

PR #8199, opened 2023-12-16
Enable longoptions support for grdfft.

rbdavis:

Python differences:

  No Python module.

Julia differences:

  Changed -E radial_power to power_spectrum. Spectrum can be x or y
    as well as radial.

Fixes #

**Reminders**


joa-quim:
This PR has covers two cases: grdfft and grdedit.

And regarding grdfft, it happening again. When you replace a GMT.jl long option that is unique (the -E case) that breaks names intersection with GMT.jl. Besides that, I prefer `radial_power` (rare to find in other softs) to `power_pectrum`.

src/longopt/grdfft_inc.h
```diff
@@ -25,48 +25,30 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
 		  short_directives,    long_directives,
 		  short_modifiers,     long_modifiers,
 		  transproc_mask */
-	{ 0, 'A', "",
-	          "",                  "",
-	          "",                  "",
+	{ 0, 'A', "azimuth|azim",      "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'C', "zcontinue|upward",  "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'D', "differentiate|dfdz", "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'E', "power_spectrum",
```
joa-quim, line 31: Same comment as in the other PR. This breaks with GMT.jl

joa-quim (approved):

src/longopt/grdfft_inc.h
rbdavis: >This PR has covers two cases: grdfft and grdedit.

My apologies to all, but something strange happened this morning while I was uploading the various new longopts-* branches. There was a problem with the longopts-grdedit upload, and because of this all of the subsequent uploads and PRs I did were impacted and you will be seeing a grdeit commit notice in each one. Too long to explain here in any coherent way (and in fact I do not understand exactly what's going on), but in all of these other PRs you should see only one changed file -- in the case of this PR, that's grdfft_inc.h. You should ignore the commit notices at the top of these PRs which refer to grdedit, as that was just me putting grdedit_inc.h back to its original unedited form.

PR #8200, opened 2023-12-16
Enable longoptions support for grdfill.

rbdavis:

Python differences:

  No differences.

Julia differences:

  Changed -N nodata (conflicts with common long options) to no_data.

Fixes #

**Reminders**


joa-quim:
> Changed -N nodata (conflicts with common long options) to no_data.

Is this for programmatic reasons? And since we have not yet decided the long names of the lowercase global options, please remind me which is `nodata`.

PaulWessel:
The **-d** common option is **--nodata**.
Here, **-N** could be **--holevalue**|**hole** as it specifically talks about what z-value should define a hole, e.g.,

`--hole=-99`

PaulWessel (approved):
See comments about -N

PR #8205, opened 2023-12-16
Enable longoptions support for grdlandmask.

rbdavis:

Python differences:

  No differences.

Julia differences:

  No differences.

Fixes #

**Reminders**


PaulWessel (commented):
Looks like -A is used in coast too but has different  words?

rbdavis:
It's true that coast uses only 'area'. Generally, when I'm working on one of these if I recall that another module has similar operability I will go back and look at it, but in this case I did not recall (and very often don't in many cases) and hence never looked at pscoast until just now. Here, I started off with area|area_thresh because those were the strings used respectively in Python and Julia for grdlandmask. I also added min_area because that was the token used in the manual page with this option.

joa-quim (approved):

Issue #8126, opened 2023-12-04
Let movie (and batch) use the variables of -Iincludefile


Esteban82:
When using `movie` (or `batch`) sometimes it is useful to use one of the variables (define in  `movie -Iincludefile`) to set name of the output file, frame rate, etc. 

Now, for this cases, the user have to do somehting like animation 06 where the variables are defined in and out of the `includefile`.

```
rate=24			# Frames per seconds
cat << EOF > init.sh
rate=$rate
EOF

gmt movie main.sh -Iinit.sh -D$rate ...
```

### Proposal
I think that a better solution is to use the `source init.sh` command (see below). I wonder if would be a good idea to include this command under the hood (or something similar) so it would be easier for the not expert users.

```
cat << EOF > init.sh
rate=24			# Frames per seconds
EOF

source init.sh       # Add this command under the hood
gmt movie main.sh -Iinit.sh -D$rate ...
```

Esteban82:
Close because cannot be done.

PR #8212, opened 2023-12-18
Homogenize animations

Esteban82:
I revised all the animations in order to homogenize them. 

I did this changes:
1. Add numbered subtitles for each step.
2. Rename the _includefile_ to `inc.sh`
3. Use `source` for _anim06_ (see #8126).
4. In _anim09_ I reorder the firts two steps.
5. Use `cat << 'EOF'` instead of `EOF`.
6. For _anim14_ I fix an error. 

Now, all the animation should look more similar. The only exception is anim14 which I found I bit harder to remake with `'EOF'` and with an _includefile_ for all the variables.


Esteban82:
> Did any of the movies change, so we should update things on YouTube?

No, no changes.

Please feel free to merge.

PR #8220, opened 2023-12-21
Enable longoptions support for grdbarb.

rbdavis:

Python differences:

  No Python module.

Julia differences:

  No Julia module.

Fixes #

**Reminders**


PaulWessel:
Maybe @masahigaki would like to opine if these long-options align with his thinking since he is the author of this module (sam when we get to barb)?

PaulWessel (approved):
I feel -A goes better with --radial and --cartesian and I am going to edit the use message to use those terms.  Same will be true for grdvector (whose -A usage I will also check).

PaulWessel:
Opps, changed my mind since we use words like polar and cartesian so lets use polar instead of radial since we mean polar versus cartesian coordinates.

joa-quim (approved):

masahigaki:
Sorry for my delayed response.
Basically I agree with the proposed long options for grdbarb.
I am not so familiar with long options of GMT, but if I may comment, --\*_azimuth seems better than --\*_azimuth"s" for consistency with long options of other subcommands.

PaulWessel:
Thanks for your feedback, @masahigaki. Looking more carefully, since **-A** is the same as in **grdvector** we should have the option for these two modules.  I would pick **--polar** for **-A** since we recently updated the documentation to distinguish between Cartesian [Default] and polar coordinates.  That means we don't overabuse the "azimuth" word.  Now, **-T** can be **--adjust_azimuth** and no plural "s" is indeed needed.

FYI, once we complete all 150 or so modules there will be another review to make sure we stay consistent.  The long-option feature won't be released until GMT 7 I think.

PR #8221, opened 2023-12-21
Enable longoptions support for grdedit.

rbdavis:

Python differences:

  No Python module.

Julia differences:

  Changed -C adjust_inc (Julia doc error?) to cmdhist|command_history.

  Changed -D header (conflict with common longoptions) to
    netcdf|netCDF|ncheader.

  Changed -E header (conflict with common longoptions) to transform.

  Changed -S wrap (conflict with common longoptions) to lonshift_region.

  Changed -T toggle_reg to match existing grdfilter longoptions
    toggle_registration|toggle.

Fixes #

**Reminders**


src/longopt/grdedit_inc.h
```diff
@@ -25,42 +25,28 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
-	          "",                  "",
+	{ 0, 'D', "netcdf|netCDF|ncheader",
 	          "",                  "",
+	          "c,d,n,o,r,s,t,v,x,y,z",
+		                       "cpt,dname,invalid,offset,remark,scale,title,varname,xname,yname,zname",
 		  GMT_TP_STANDARD },
-	{ 0, 'G', "",
-	          "",                  "",
+	{ 0, 'E', "transform",
+	          "a,e,h,l,r,t,v",     "hvflip,xyswap,hflip,rot90counterclock,rot90clock,transpose,vflip",
```
remkos, line 39: the rot90... becomes very long.
Can we do (at least as alternative) "rotleft" and "rotright". It should be obvious that it is by 90º.

remkos (changes requested):

src/longopt/grdedit_inc.h
PaulWessel: ******Maybe** rot90ccw and rot90cw are more specific than just a rotation of unspecific value.  I agree that **rot90counterclock** is too long and even not long enough since it begs for the "wise" at the end.

PaulWessel (approved):
Please revise the rotations as suggested in the comment.

src/longopt/grdedit_inc.h
remkos: Well, it would be hard to rotate a 2D grid by anything else than increments of 90º. But, yes, I would also prefer "rot90ccw" and "rot90cw"

PR #8222, opened 2023-12-21
Enable longoptions support for grdmask.

rbdavis:

Python differences:

  No Python module.

Julia differences:

  No differences.

Fixes #

**Reminders**


PaulWessel (approved):
For -A there are two more directives:

t theta|azimuth
r radial

Please add before merge.

src/longopt/grdmask_inc.h
```diff
@@ -25,30 +25,21 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
-	          "",                  "",
+	{ 0, 'C', "clobber",
+	          "f,l,o,u",           "first,low,last,high",
 	          "",                  "",
 		  GMT_TP_STANDARD },
-	{ 0, 'S', "",
-	          "",                  "",
+	GMT_G_OUTGRID_KW,
+	GMT_I_INCREMENT_KW,
+	{ 0, 'N', "outedgein",
```
joa-quim, line 38: In GMT.jl this is `out_edge_in`

PaulWessel: OK, s that should definitively be an alias.  We can handle **outedgein**|**out_edge_in** I think.

rbdavis:
Joaquim, there seems to be some confusion with regard to Julia docs. When I do '? grdmask' in Julia this is what I see (macOS Monterey screenshot):

julia-grdmask

As you can see, there are no underscores anywhere in **outedgein** -- however, there is a strange underline beneath the **edge** substring of **outedgein**. ??? But then in the option immediately following **outedgein** you can clearly see the underscore in **search_radius**, so I don't  think there is any systemic problem with displaying underscores here.

I really am trying to be conscientious about looking at existing Python and Julia strings as part of this process, but I can only go by what I see. This is not the first time this problem has occurred -- a few modules back there was some **errorbarspace**  (or similar, I don't recall exactly) Julia string where I saw no underscores in the doc but you said that there should have been.

joa-quim:
Hi Roger,

This seems to be an unfortunate shit on the docstrings rendering. No idea why sometimes it takes the underscore to mean `begging italicizing` and the word shows up with the underline. If you do `@? grdmask` it opens a web page where things are are rendered correctly. Sorry about this, but don't know what to do about it.

image

rbdavis:
OK, understood, thanks. What I really would prefer is one single method to access the docs, not multiple methods which produce conflicting results, so from now on I will use only '@? module'.

joa-quim:
Yes, I understand that but ...... not al modules got a web page document (that's a LOT of work for each one). For the ones that were not lucky yet, it still has to be `? module`.

PR #8226, opened 2023-12-21
Enable longoptions support for grdproject.

rbdavis:

Python differences:

  Changed -F scaling to one2one|one2one_scale.

Julia differences:

  No differences.

Fixes #

**Reminders**


src/longopt/grdproject_inc.h
```diff
@@ -25,32 +25,20 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
 		  GMT_TP_STANDARD },
-	{ 0, 'E', "",
-	          "",                  "",
-	          "",                  "",
-		  GMT_TP_STANDARD },
-	{ 0, 'F', "",
-	          "",                  "",
+	{ 0, 'E', "dpi",               "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'F', "one2one|one2one_scale",
+	          "c,i,p,e,f,k,M,n,u", "cm,inch,point,meter,foot,km,smile,nmile,ussft",
```
remkos, line 35: This may be very late in the game to ask, but I think we should be more flexible here, since it would be easy to forget the exact syntax. Can we do something like:
"cm,inch|in,point|pt,meter|m,foot|ft,km,smile|mi,nmile|nm,ussft"

I'm not sure that a single letter "m" would work. To be checked.
Also this should then be used consistently in other modules.

src/longopt/grdproject_inc.h
```diff
@@ -25,32 +25,20 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
-	{ 0, 'I', "",
-	          "",                  "",
-	          "",                  "",
-		  GMT_TP_STANDARD },
-	{ 0, 'M', "",
-	          "",                  "",
+	GMT_G_OUTGRID_KW,
+	{ 0, 'I', "inverse",           "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'M', "unit|projected_unit",
+	          "c,i,p",             "cm,inch,point",
```
remkos, line 41: Similar to the above: could we do "cm,inch|in,point|pt"

remkos (changes requested):

src/longopt/grdproject_inc.h
PaulWessel: I am not sure if we want distance units to be spelled out.  That would mean hundred of places where we should allow km for k, point for p, etc, Think of map projections, e.g. Jm72point.  I think it is cleaner that people need to know what the unit abbreviations are instead of having to deal with smile and foot. I think the most commonly used units will be c, i, p, m, k.  Doubt anyone uses USGS survey foot. Mariners may use nautical miles.
However, I am happy to hear what more people think of this.  Seems like unnecessary and much work to mel  @joa-quim?, @seisman ?, @leouieda , @Esteban82 ?

joa-quim:
Agree with no long names here. `250smile` may look modern but is wrong (should be plural)

remkos:
I was not intending to create such a big thing. But it is true that it would be confusing, since we stick to `c,i,p` etc. for units (as in `-W1p,blue`) but then use the units written out here in `grdproject`

So I agree with the consensus among @PaulWessel and @joa-quim to use  "c,i,p,m,k,n,M,f,u " and "c,i,p" resp.

PR #8228, opened 2023-12-21
Enable longoptions support for grdselect.

rbdavis:

Python differences:

  No Python module.

Julia differences:

  No Julia module.

Fixes #

**Reminders**


src/longopt/grdselect_inc.h
```diff
@@ -25,54 +25,33 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
-	{ 0, 'Z', "",
+	{ 0, 'G', "force_remote",      "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'I', "invert|reverse",
+	          "C,D,F,L,N,R,W,Z,r", "points,increment,polygons,lines,nans,region,range,zrange,registration",
 	          "",                  "",
+		  GMT_TP_MULTIDIR },
+	{ 0, 'L', "linefile",          "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'M', "margins",           "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'N', "nans",
+	          "l,h",               "lower,higher",
```
remkos, line 50: I would suggest using here "lower|lower_than,higher|higher_than"

remkos (changes requested):

src/longopt/grdselect_inc.h
PaulWessel: I am find with those aliases, make it so @rbdavis

PaulWessel (approved):
See remko's suggestion and my endorsement

PR #8231, opened 2023-12-21
Fix bad check for when using -T in mapproject

PaulWessel:
See this forum [post](https://forum.generic-mapping-tools.org/t/mapproject-t-option/4534) for details. The parser checked that all the options that cannot work with **-T** were not set ( **-A**, **-E**, **-G**, **-L**, **-W** or **-Z**) but it also erroneously added **-T** in that list so of course we got the dumb message.  Now fixed and works.


PaulWessel:
This is the result:

```
echo "-4725626.3127 -6460896.1797" | gmt mapproject -Jw0/1:1 -R-180/180/-90/90 -C -Fe -I -TGRS-80:0,0,0/- --FORMAT_FLOAT_OUT=%.8f
-67.72575952	-55.93418369
```

joa-quim (approved):

PR #8233, opened 2023-12-22
Forgot to handle Euro dates dd.mm.yyyy in the first-record exam

PaulWessel:
See Forum [post](https://forum.generic-mapping-tools.org/t/regression-in-filter1d/4535) for background. We had a major upgrade for examining what a column might contain a month or two ago but while we did check for multiple periods as possible European-style non-ISO8601 dates, we called another function first that thought it was a string...Now fixed. 

```
cat <<eof | awk '{print $1"T",$2}' | gmt filter1d  -Fm14 --TIME_UNIT=d -V -E --FORMAT_DATE_IN=dd.mm.yyyy
29.01.2015      75.1
30.01.2015      76
31.01.2015      75
01.02.2015      75.7
eof
filter1d [INFORMATION]: Processing input table data
filter1d [INFORMATION]: Reading Data Table from Stream 1e021c3d8
filter1d [INFORMATION]: Writing Data Table to Standard Output stream
filter1d [INFORMATION]: Filter the data columns
filter1d [INFORMATION]: Read 4 records from table 0, segment 0
filter1d [INFORMATION]: F width: 14 Resolution: 1 Start: 16464 Stop: 16467
2015-01-29T00:00:00	75.4
2015-01-30T00:00:00	75.4
2015-01-31T00:00:00	75.4
2015-02-01T00:00:00	75.4
```

It is always possible to pick odd formats for scientific times but always good to stick with the ISO.

### DIFF
--- a/src/gmt_internals.h
+++ b/src/gmt_internals.h
@@ -57,7 +57,7 @@ EXTERN_MSC unsigned int gmtlib_is_coordinate (struct GMT_CTRL *GMT, unsigned int
 EXTERN_MSC unsigned int gmtlib_is_time (struct GMT_CTRL *GMT, char *text);
 EXTERN_MSC unsigned int gmtlib_is_string (struct GMT_CTRL *GMT, char *string);
 EXTERN_MSC unsigned int gmtlib_determine_datatype (struct GMT_CTRL *GMT, char *text);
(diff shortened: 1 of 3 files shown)


joa-quim (approved):

anbj:
Great, now works again.

anbj (approved):

PR #8246, opened 2023-12-30
Allow colorbar -LI as well as -Li

PaulWessel:
If a CPT goes from 0 to 100 in steps of 10 then **-Li** normally labels the first color 0-10 and the last 90-100. With upper case **I** those two are labelled < 10 and > 90 instead, thus includes values outside the CPT range.  In addition, the PR attempts to adjust the default panel size to the longer AAA-BBB annotations.  See this form [post](https://forum.generic-mapping-tools.org/t/colorbar-box-and-annotation/4548) for the issue.  Here is what **-Li** and **-LI** look like.  Unlike the poster in the forum, I don't think we want <= since the previous is 10-20 so <= 10 seems wrong to me.

bars

Do you think this is a useful directive, @joa-quim ?


joa-quim (approved):

joa-quim:
Yes, looks good, and yes, dshould be `<` and nor `<=`

PR #8255, opened 2024-01-02
Do not try to project images that don't have set 'need_to_project'.

### LINKED ISSUE #8254: imshow in Julia is now broken because of the pad
Plot is linear. Should not need any PAD.

```
julia> gmtwrite("lixo1.tif", rand(UInt8,32,32,3), driver=:GTiff)

julia> I = gmtread("lixo1.tif", img=true, band=[0 1 2]);

julia> imshow(I)
GMT [WARNING]: gmtapi_change_imagelayout: reordering function for case BRBa -> BRP not yet written. Doing nothing.
grdimage [ERROR]: gmt_img_project: Input image does not have sufficient (2) padding
ERROR: Something went wrong when calling the module. GMT error number = 79
```

joa-quim:
Fixes #8254

Don't know if it's the best fix but works.


PaulWessel (approved):
All tests still passes so OK

PR #8259, opened 2024-01-05
When given multiple tables, report line number of bad records in the offending file

PaulWessel:
See this [post](https://forum.generic-mapping-tools.org/t/gmt-info-reporting-of-mismatch-of-actual-and-expected-fields/4567) on the Forum.  The problem was that the Warning statement used the total internal record counter (the only option if you cat may files into a table reader) but if you give individual files on the command line we have a secondary counter that gets reset to 0 for each new file. Pretty sure we just forgot.  Here is what it looks like now:

```
cat<<eof > file1
2 4
6 7
2 8
eof
cat<<eof > file2
2 4
trash
6 7
2 8
eof
gmt info file1 file2
gmtinfo [WARNING]: Mismatch between actual (1) and expected (2) fields near line 2 in file file2

```

Of course, the` cat file1 file2 | gmt info` example cannot change since it just see a combined stdin stream. The **gmt** **info** module reads record by record, while many others (e.g., **gmt** **convert**) reads file by file.  That also works now, e.g. 

```
gmt convert file1 file2
gmtconvert [WARNING]: Mismatch between actual (1) and expected (2) fields near line 2 in file file2
```


anbj (approved):
Perfect. Reports as expected. Thanks Paul!

PR #8260, opened 2024-01-06
Add -s to build-release to allow signing with default no signing

PaulWessel:
We now need to pass option **-s** to admin/build-relesae if Paul or Max is to sign the bundle.  Default is no signing.

Max, please check UH account name in `ConfigReleaseBuildSigning.cmake`.


PaulWessel:
Tested and works so I will merge it in.

PR #7805, opened 2023-09-10
Let pscoast -E print coordinates in [-180 180] or FORMAT_GEO_OUT.

joa-quim:
Fix #7540

PR #8266, opened 2024-01-07
The -R string from three coutries crossing 180 was bad

PaulWessel:
The USA, Russia, and Fiji are the only country polygons that cross lon = 180.  That gave us odd **-R**_w/e/s/n_ strings such as 172.436/-66.9489/18.9099/71.3907 for the US. With west > east we run into trouble when building the list of remote file tiles (e.g., the @earth_relief ******_p.jp2 that needs to be downloaded).  The result was a blank map. I verified this was the same with GMT 6.4 so it is not a new bug; it has probably been there since Day 1 of remote data set implementation but nobody used **-RUS**, **-RRU** or **-RFJ** before. This PR simply does the sanity check that if if `east < west` we subtract 360 from west, since it is _assumed_ all global files are on the **-Rd** layout.  With this fix the plots for the three countries now works, as shown below (these were all blank frames before):

`gmt grdimage @earth_relief_05m -B -RUS -png USA`
USA

`gmt grdimage @earth_relief_05m -B -RRU -png RU`
RU

`gmt grdimage @earth_relief_05m -B -RFJ -png FJ`
FJ

Since this is a long-standing bug we can choose what to do:

1. It is Sunday, so we could in a hurry build new GMT6.5 installers and the problem has been dealt with
2. Shit, horse is out of the barn and we simply should do 6.5.1 in a few days.
3. If we have lived with this since 2016 then perhaps we can live with it until 6.6?

Please opine.  Note @joa-quim this particular PR has nothing to do with code freeze concern since it was there so long.

### DIFF
--- a/src/gmt_dcw.c
+++ b/src/gmt_dcw.c
@@ -900,6 +900,9 @@ struct GMT_DATASET * gmt_DCW_operation (struct GMT_CTRL *GMT, struct GMT_DCW_SEL
 			wesn[XLO] = 0.0;
 			wesn[XHI] = 360.0;
 		}
+		if (wesn[XHI] < wesn[XLO]) {	/* Cannot tolerate that west larger than east */
+			wesn[XLO] -= 360.0;
+		}
 		GMT_Report (GMT->parent, GMT_MSG_INFORMATION, "Region implied by DCW polygons is %g/%g/%g/%g\n", wesn[XLO], wesn[XHI], wesn[YLO], wesn[YHI]);
 	}
 	gmt_M_free (GMT, order);
--- a/src/gmt_remote.c
+++ b/src/gmt_remote.c
@@ -1513,7 +1513,7 @@ int gmt_download_file (struct GMT_CTRL *GMT, const char *name, char *url, char *
 		if (be_fussy || !(curl_err == CURLE_REMOTE_FILE_NOT_FOUND || curl_err == CURLE_HTTP_RETURNED_ERROR)) {	/* Unexpected failure - want to bitch about it */
 			GMT_Report (API, GMT_MSG_ERROR, "Libcurl Error: %s\n", curl_easy_strerror (curl_err));
 			if (curl_err == CURLE_HTTP_RETURNED_ERROR)
-				GMT_Report (API, GMT_MSG_ERROR, "Probably means %s does not exist on the remote server\n", name);
+				GMT_Report (API, GMT_MSG_ERROR, "Probably means %s does not exist on the remote server [%s]\n", name, GMT->session.DATASERVER == NULL ? "not set" : GMT->session.DATASERVER);
 			error = curl_err;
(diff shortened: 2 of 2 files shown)


Esteban82:
I would go with 2 or 3. I think 1 is a bad idea.

seisman:
> I verified this was the same with GMT 6.4 so it is not a new bug;

It works well for me with GMT 6.4.0 on macOS:
```
$ gmt --version
6.4.0
$ gmt grdimage @earth_relief_05m -B -RUS -png USA
grdblend [NOTICE]: Remote data courtesy of GMT data server china [http://china.generic-mapping-tools.org]

grdblend [NOTICE]: SRTM15 Earth Relief v2.5.5 at 05x05 arc minutes reduced by Gaussian Cartesian filtering (9.3 km fullwidth) [Tozer et al., 2019].
grdblend [NOTICE]:   -> Download 180x180 degree grid tile (earth_relief_05m_p): S90E000
grdblend [NOTICE]:   -> Download 180x180 degree grid tile (earth_relief_05m_p): S90W180
```

PaulWessel:
OK, maybe you had the necessary tiles.  I started with scratch and it was unable to get them just for those three examples.

joa-quim:
> Please opine. Note @joa-quim this particular PR has nothing to do with code freeze concern since it was there so long.


We have tons of open bugs reports. I have built the Win installers 3 of 4 times (installers build is not an instantaneous step). Let's stop now.

PaulWessel:
I am fine with that, too.

seisman:
Just in case you're still interested, the bug was introduced in https://github.com/GenericMappingTools/gmt/pull/7805.

yvonnefroehlich:
> I verified this was the same with GMT 6.4 so it is not a new bug; 

Hm. Not sure. For me, it works for both Windows with GMT 6.4 and Linux with GMT 6.4.0_965606f_2022.03.14. But it fails with GMT 6.5 on Windows.

PaulWessel:
> Just in case you're still interested, the bug was introduced in #7805.

I see.  So that section sort of assumed that west < east but for those three countries gmt_DCW_operation returns 172.436/-66.9489/18.9099/71.3907 for the US and that get caught in those tests.  So perhaps 6.4 was OK for the images if you already had the tiles.  I did not and I verified there was a problem selecting the right tiles and download then with the above bug.  Obviously work around with coordinates.

FYI, just getting the -R string (which I think @joa-quim was after in #7805):

```
GMT 6.4: gmt pscoast -EUS gives -R172.436111/293.051109/18.909859/71.390686
master before fix (i.e. 6.5 installers): -R172.436111/-66.948891/18.909859/71.390686
branch with fix: -R-187.563889/-66.948891/18.909859/71.390686
```

So first and last are the same ±360 and because the data sets are in -Rd I picked the negatives.

PR #8283, opened 2024-01-11
Add longoptions support for psbarb.

rbdavis:

Python differences:

  No Python module.

Julia differences:

  No Julia module.

Fixes #

**Reminders**


PaulWessel (approved):

src/longopt/psbarb_inc.h
```diff
@@ -25,42 +25,19 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
...
-		  GMT_TP_STANDARD },
-	{ 0, 'W', "",
-	          "",                  "",
-	          "",                  "",
-		  GMT_TP_STANDARD },
-	{ 0, 'Z', "",
-	          "",                  "",
+	{ 0, 'Q', "barbs",
 	          "",                  "",
+	          "a,g,p,j,s,w,z",     "angle,fill,pen,justify,longspeed,width,uvdata",
```
joa-quim, line 38: `longspeed`? Shouldn't be `windspeed`?

rbdavis: That's the speed value which gets drawn as a long barb. _longbarbwindspeed_ seemed a bit long to me.

joa-quim: Sorry, I must be missing something, both  `longspeed` and `windspeed` have exactly the same length and **+s** is meant to set the `the wind speed which corresponds to a long barb`

PR #8285, opened 2024-01-12
Update module names in docs

remkos:

A bunch of modules still appear with their old names in the indices in the documentation.
For example, in `modules.rst` we still have `gmtwhich` instead of `which`, while in `gmtwhich.rst` the synopsis does start with **gmt which**, though it has **gmtwhich** in the header.

One way to deal with this is similar to `psxy` and `plot`: we create near duplicate files `gmtwhich.rst` and `which.rst`, where the former will still appear in the list of classic modules, whereas the later will appear in the modern (default) list. 

I was concerned though that people will forget to update the proper documentation files, i.e. `gmtwhich.rst` **and** `which.rst` with any change. Given the number of modules involved (see below) that would be serious work.

Thus I choose to simply change the top section header of `gmtwhich.rst` (etc.) and write there `which` instead of `gmtwhich`, and of course change the alphabetical order in `modules.rst`.  This writes `which` in the index, without any further changes.

To avoid that `which` would show as well in the classical module list, I used, e.g.:
```
:doc:`gmtwhich <gmtwhich>`
```
which would then show the word `gmtwhich` again.

I identified the following modules that would need this treatment:

- [x] binstats <- gmtbinstats
- [x] connect <- gmtconnect
- [x] convert <- gmtconvert
- [x] defaults <- gmtdefaults
- [x] get <- gmtget
- [x] info <- gmtinfo
- [x] logo <- gmtlogo
- [x] math <- gmtmath
- [x] regress <- gmtregress
- [x] select <- gmtselect
- [x] set <- gmtset
- [x] simplify <- gmtsimplify
- [x] spatial <- gmtspatial
- [x] split <- gmtsplit
- [x] vector <- gmtvector
- [x] which <- gmtwhich

And in the supplemental modules

- [x] flexure <- gmtflexure
- [x] gravmag3d <- gmtgravmag3d
- [x] pmodeler <- gmtpmodeler

### DIFF
--- a/doc/rst/source/gmtbinstats.rst
+++ b/doc/rst/source/gmtbinstats.rst
@@ -1,9 +1,9 @@
 .. index:: ! gmtbinstats
 .. include:: module_core_purpose.rst_
 
-***********
-gmtbinstats
-***********
+********
+binstats
+********
 
 |gmtbinstats_purpose|
 
--- a/doc/rst/source/gmtconnect.rst
+++ b/doc/rst/source/gmtconnect.rst
@@ -1,9 +1,9 @@
 .. index:: ! gmtconnect
 .. include:: module_core_purpose.rst_
 
-**********
-gmtconnect
-**********
+*******
+connect
+*******
 
 |gmtconnect_purpose|
 
--- a/doc/rst/source/gmtconvert.rst
+++ b/doc/rst/source/gmtconvert.rst
@@ -1,9 +1,9 @@
 .. index:: ! gmtconvert
 .. include:: module_core_purpose.rst_
 
-**********
-gmtconvert
-**********
+*******
+convert
+*******
 
 |gmtconvert_purpose|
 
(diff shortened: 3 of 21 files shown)


seisman:
> we create duplicate files `gmtwhich.rst` and `which.rst`, where the former will still appear in the list of classic modules, whereas the later will appear in the modern (default) list.

`which` is also valid in classic mode, so perhaps no need to duplicate.

remkos:
> `which` is also valid in classic mode, so perhaps no need to duplicate.

Good point. That makes my further discussion on how to avoid duplication more appropriate. Then I'll proceed that way.
The only duplication (that's already there) will be with `plot.rst` and `psxy.rst`, as well as `plot3d.rst` and `psxyz.rst`.

I'll proceed that way, undoing my duplications suggested in the commit.

PaulWessel (approved):
Looks good

PR #8159, opened 2023-12-08
Add changelog for 6.5

maxrjones:
**Description of proposed changes**

Starting the 6.5 changelog

Fixes #

**Reminders**

- [ ] Make sure that your code follows our style. Use the other functions/files as a basis.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Describe changes to function behavior and arguments in a comment below the function declaration.
- [ ] If adding new functionality, add a detailed description to the documentation and/or an example.

PR #8287, opened 2024-01-12
colorbar: Add modifier +n|N to -D option for placing colorbar NaN box on the right and to skip label

PaulWessel:
The new modifier for placing a NaN-color box and NAN label at the end of the color bar is now extended to

**+n**|**N**[**-**|_label_]

where the new **+N** means place it on the right side of the bar (top for vertical bars) and if the label is **-** we do not place any _label_.  Default is still NaN unless you append an alternative label.

Added two new test scripts `psscale/colorbar-nans-h.sh` and `psscale/colorbar-nans-v.sh` to ensure it works for both horizontal and vertical bars.


seisman:
Add the `add-changelog` label and change the PR title to a more descriptive one like "colorbar: Add modifier +n|N to -D option for placing colorbar NaN box on the right and to skip label"? If so, life will be much easier when we work on the GMT 6.6.0 changelog (xref #8159).

joa-quim (approved):

PR #8297, opened 2024-01-16
Require movie canvas dimensions in integer pixels when using modifiers +c|i

PaulWessel:
If plot dimensions are given in pixels instead of cm, inches etc then use **+c**|**i** to indicate that units the _dpu_ is in since actual plot dimensions is then the pixel dimensions divided by the _dpu_ which gives either cm or inches.  This feature was implemented to avoid having to write painful things like

`-C16.5517241379cx6.03448275862cx116`

Much cleaner to allow

`-C1920x700x116+c`

This is helpful when it is simpler to specify the dimensions in integers but we will need to relate those via the dpc to physical dimensions (here 16.5517241379c).

### DIFF
--- a/doc/rst/source/movie.rst
+++ b/doc/rst/source/movie.rst
@@ -13,7 +13,7 @@ Synopsis
 .. include:: common_SYN_OPTs.rst_
 
 **gmt movie** *mainscript*
-|-C|\ *canvas*
+|-C|\ *canvassize*
 |-N|\ *prefix*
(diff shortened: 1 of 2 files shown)


Esteban82:
Do you any script to test it?

PaulWessel:
Not a new one but my movie in the paper now uses this and it works.

Esteban82 (approved):
Blind approve

PR #8302, opened 2024-01-17
Remove deprecated doc and usage language from grdview

PaulWessel:
There is no need to explain about the deprecated way of specifying 3 grids to drap (red, green blue) since we stopped doing that for **grdimage** a long time ago.  **grdview** will still check if given 3 but we don't have to talk about it.  The usage and docs now deal with **-G**_drapegrid_|_drapeimage_ only. We have a script that exercises the deprecated forms `-Gr.nc,g.nc,b.nc` and `-Gr.nc -Gg.nc -Gb.nc` and that works fine under the hood without any advertisement.

I left the special option **-z** for passing a single image from GMT.jl. Suggest @joa-quim play with removing that.  Since the KEY need **G** or **I** perhaps some work needs to be done in `GMT_Encode_Options`?

Improved the doc layout for **-G** and **-Q** as well as their synopsis.

Here is -G:

G

Here is -Q:

Q

All tests pass so I think this is mostly a documentation PR.

### DIFF
--- a/doc/rst/source/grdview.rst
+++ b/doc/rst/source/grdview.rst
@@ -15,7 +15,7 @@ Synopsis
 **gmt grdview** *reliefgrid* |-J|\ *parameters*
 [ |SYN_OPT-B| ]
 [ |-C|\ [*section*/]\ *master*\|\ *cpt*\|\ *color*\ :math:`_1`,\ *color*\ :math:`_2`\ [,\ *color*\ :math:`_3`\ ,...]\ [**+h**\ [*hinge*]][**+i**\ *dz*][**+u**\|\ **U**\ *unit*][**+s**\ *fname*] ]
-[ |-G|\ *drapegrid* \| |-G|\ *grd_r* |-G|\ *grd_g* |-G|\ *grd_b* ]
+[ |-G|\ *drapegrid*\|*drapeimage* ]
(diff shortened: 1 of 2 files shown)


PaulWessel:
Yes, saw that so leave for now.  Solution is to learn in `GMT_Encode_Option` that the virtual single file is an image, and then change type from `?` to I or G.
But **grdimage** also accept image or grid and it just does `<G{` and never sees an I even though we detect that in the code, no?

joa-quim:
Yep, that magic is apparently working. I have this Julia test
```
I = mat2img(rand(UInt8,89,120,3), proj4="+proj=longlat +datum=WGS84 +no_defs");
grdview(rand(90,120), G=I, J=:X12, JZ=5, Q=:i, view="145/30", show=1)
```
and a bunch of colored random noise draped on a random grid does show up and looks good. So that `-z` is probably not needed after all.

PaulWessel:
Try to remove the KEY and the case 'z': in this PR and see if all is well.

joa-quim:
Removing the `z` and its KEY as well as replacing `zI(` by `GI(`, all made no difference. The colored trash test still works fine.

PaulWessel:
OK, did you commit this to the PR so we can approve merge etc?

joa-quim:
Did commit now. Also replaced the KEY though it's probably not being used.

joa-quim (approved):

PR #8307, opened 2024-01-18
Add longoptions support for psbasemap.

rbdavis:

Python differences:

  No differences.

Julia differences:

  No differences.

Fixes #

**Reminders**


src/longopt/psbasemap_inc.h
```diff
@@ -25,26 +25,20 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
 		  short_directives,    long_directives,
 		  short_modifiers,     long_modifiers,
 		  transproc_mask */
-	{ 0, 'A', "",
-	          "",                  "",
-	          "",                  "",
+	{ 0, 'A', "polygon",           "", "", "", "", GMT_TP_STANDARD },
+	{ 0, 'D', "inset|inset_box",
+	          "g,j,J,n,x",         "mapcoords,inside,outside,boxcoords,plotcoords",
+	          "r,s,t,u,w,j,o",     "corners,outfile,translate,units,width,janchor,offset",
```
joa-quim, line 31: I perfectly understand the difficulty with this one `janchor` (the difference of the two _anchors_ in the option are difficult to grasp), but carrying the letter modifier (**j**) into the long option name is not much helpfull either.

PaulWessel: This is about upper and lower case j|J, right?  How about inanchor and outanchor?

PaulWessel (approved):

joa-quim:
`janchor` is not good. If we have in other modules it should be replaced there too.

PaulWessel:
Can't it just be justify|justification instead?

joa-quim:
No because we have also **+j** modifiers that take that word, but `inanchor | outanchor` are a good ones.
The point of `janchor` is that it's carrying the one letter code into the long option name, which defeats the all purpose of long options.

Issue #8300, opened 2024-01-17
Do not raise "Unrecognized option -x" error if GMT is not compiled with OpenMP/GThreads support?


seisman:
The `-x` common option is only available if GMT is compiled with OpenMP/GThreads support. If not, using `-x` gives the error: `Unrecognized option -x`. Users may be surprised why a script works on machine A but gives such an error on machine B. After reading the documentation or the man page, they may realize that GMT on machine B is not compiled with parallel support. Then users have to modify their working script unless they're unable to install another parallel-enabled GMT. 

So, instead of giving the "Unrecognized option -x" error, we can make it always available. 

- If GMT is compiled with parallel support, then everything works as before
- If GMT is NOT compiled with parallel support, then raise a **WARNING** and fallback to one core

In this way,

1. the warning message can clearly explain that the GMT is not parallel-enabled
2. users don't have to modify their working script if they don't care about parallel
3. the synopsis message can also be simplified (i.e., `GMT_x_OPT`, `GMT_ADD_x_OPT`)

PaulWessel:
This is of course a recent change dealing with docs for -x and difference between OpenMP, Gthreads, and movie/batch.  I would not care if we always parsed -x and if not compiled to support it we.  Interested in fixing this by adding your warning instead?

PR #8319, opened 2024-01-20
Reduce lack of -x to just an information messag

PaulWessel:
See #8300.  Just say if no parallel support then only once care will be used. However @seisman, my effort at printing that out fails (but the error is gone).  Can you check why

`#if !defined(GMT_MP_ENABLED)`

seems to fail for me? E.g., lines 9169 and 18779 in gmt_init.c

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -9164,9 +9164,12 @@ int gmt_default_error (struct GMT_CTRL *GMT, char option) {
 		case 's': error += GMT->common.s.active == false; break;
 		case 't': error += GMT->common.t.active == false; break;
 		case 'w': error += GMT->common.w.active == false; break;
-#ifdef GMT_MP_ENABLED
-		case 'x': error += GMT->common.x.active == false; break;
+		case 'x': error += GMT->common.x.active == false;
+		error--;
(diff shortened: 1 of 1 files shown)


src/gmt_init.c
```diff
@@ -18770,10 +18773,11 @@ int gmt_parse_common_options (struct GMT_CTRL *GMT, char *list, char option, cha
 			error += gmt_M_more_than_once (GMT, GMT->common.w.active) || gmtinit_parse_w_option (GMT, item);
 			break;
 
-#ifdef GMT_MP_ENABLED
 		case 'x':
 			error += (gmt_M_more_than_once (GMT, GMT->common.x.active) || gmtinit_parse_x_option (GMT, item));
 			GMT->common.x.active = true;
+#!define(GMT_MP_ENABLED)
```
seisman, line 18779: Typo here?
```suggestion
#if !define(GMT_MP_ENABLED)
```

seisman:
I'm getting this error when OpenMP is disabled:
```
/usr/bin/ld: src/libgmt.so.6.6.0: undefined reference to `gmtinit_parse_x_option'
```
The reason is `gmtinit_parse_x_option` is only defined if OpenMP is enabled.

Issue #8326, opened 2024-01-24
Modifier +e detected but not a valid modifier! - ignored in gmt movie -A+e?


Esteban82:
I run the animation [WED-A_Fig_8.sh](https://github.com/GenericMappingTools/gmt-2024-animation/blob/main/WED-A_Fig_8.sh) with this command and I got this messagge:

```
gmt movie main.sh -AIndianaJones_RaidersMarch.mp3+e -Iin.sh -Sbpre.sh -N${FIG} -Tdistance_vs_frame.txt -Mm,png -Etitle.sh+d6s+fo1s -Cfhd -Fmp4 -Vi -D1 -K+p -Zs
movie [WARNING]: Modifier +e detected but not a valid modifier! - ignored

```

BTW, I also got these messagges:

```
movie [INFORMATION]: Frame 28 of 35 completed [ 80.0 %]
text [WARNING]: Record 1 had bad fill transparency (NaN) - set to 0.0
text [WARNING]: Record 1 had bad stroke transparency (NaN) - set to 0.0
movie [INFORMATION]: Frame 30 of 35 completed [ 82.9 %]
```

And this other messagge related to `ffmpeg`.

```
movie [NOTICE]: Running: ffmpeg -loglevel warning -f image2 -framerate 1 -y -i "/home/federico/Github/Esteban82/gmt-2024-animation/WED-A_Fig_8/WED-A_Fig_8_%02d.png" -i IndianaJones_RaidersMarch.mp3 -vcodec libx264  -pix_fmt yuv420p WED-A_Fig_8.mp4
[image2 @ 0x55f5832329c0] Thread message queue blocking; consider raising the thread_queue_size option (current value: 8)
```

I found [here](https://video.stackexchange.com/questions/25643/ffmpeg-thread-queue-size-warning-what-does-this-mean) that this could be solve if I add  `-thread_queue_size 4096` after the frame rate. I don't understand very well what this means. Let me know if you want me to make a PR with this "solution". Or maybe this would be solved when `-A+e` is fixed.

So the command corrected is: 

`ffmpeg -loglevel warning -f image2 -framerate 1 -thread_queue_size 4096  -y -i "/home/federico/Github/Esteban82/gmt-2024-animation/WED-A_Fig_8/WED-A_Fig_8_%02d.png" -i IndianaJones_RaidersMarch.mp3 -vcodec libx264  -pix_fmt yuv420p WED-A_Fig_8.mp4`

[...]

PR #8327, opened 2024-01-24
Fix bad movie -A parsing

PaulWessel:
Would complain about **+e**.  Turn things around and checked for the deprecated **+l +s** first.

Sorry @Esteban82.  See if this works properly, it clearly had a problem as you pointed out.  Now I check for the deprecated **+s +l** first and if found I get out.  If not I check for **+e** and the audio file name.


Esteban82 (approved):

Esteban82:
Closes #8326

PR #8388, opened 2024-03-02
Patch a case where nodes with very close to zero values created an infinite alloc loop.

### LINKED ISSUE #6738: grdcontour having trouble with tiny mignitudes
See the original [PyGMT](https://github.com/GenericMappingTools/pygmt/issues/1927) issue for background.

I have isolated a small 5x4 grid cut from the offending grid above that fails to let us contour the zero-contour.  This hangs:

`gmt grdcontour t.grd -V  -C10 -JX15c > t.ps`

The values are very small on either side of zero.  Here is a graphical view of the values:

nodes

If I scale t.grd by 1000 and try to contour that it works.  Since **grdcontour** works by never allowing any node to be _exactly_ zero (we add a tiny amount), the issue of _tiny_ in relation to these small values is the troubles, so I know where to look.  Posting t.grd here for reference.

[t.grd.zip](https://github.com/GenericMappingTools/gmt/files/8758738/t.grd.zip)

### LINKED ISSUE #8387: grdcontour hanging
**grdcontour gets hung when fed a grid file that has values very close to zero, but not exactly zero.**

**Please see the post on the GMT forum page. Data and two-line script were posted there.**

**System information**

* Operating system: MacOS 13.6.4
* GMT version (`gmt --version`): 6.6.0_1443c0c_2024.02.26

joa-quim:
The situation arose when the condition

if ((dz = z[this_side] - z[this_side+1]) == 0.0f) continue;

was never met by a tiny bit. Comparing to 0.0 is never a good idea. So I made it

if (fabs((dz = z[this_side] - z[this_side+1])) < 1e-10) continue;

I get 6 grdcontour test failures but they don't seem related to this change (more due to the endless headache of slight different grids).

Fixes #8387

Fixes #6738


PaulWessel:
Talking to Garrett to tom orrow, anything I should add?

Issue #8391, opened 2024-03-04
Release of GMT 6.5.1 needed?


Esteban82:
I am working with Paul and Guillaume on an article about movies with GMT (**The Generic Mapping Tools and Animations for the Masses**). 

I think we will need a new GMT release (6.5.1) because I some issues (with 6.5) that I will try to fix myself (like #8209). And also becauses include some features that have already been done (like #8297).

What do you think?

I don't know is somelse is working on a paper that will need some new features or fixes. I know that @joa-quim is working on one.

seisman:
Maybe even 6.6.0, although there are not many new features added. 

I expect that this release may take much longer than any previous releases, especially that Paul was the one that built the macOS bundle.

Esteban82:
> Maybe even 6.6.0, although there are not many new features added.

Yes, probably. I update the scritp to GMT 6.5 for the paper about animations. So, I won't need this release soon.

I will close this issue.

PR #8396, opened 2024-03-05
Add -thread_queue_size 4096

### LINKED ISSUE #8394: Movie. ffmpeg "Thread message queue blocking"
## Intro

I create this command to create a **movie** (**mp4**) with a frame rate of 60 and audio a file (I think that those are the key elements that causes the issue):

```
gmt movie main.sh -Tdistance_vs_frame.txt -Iin.sh -Sbpre.sh -Etitle.sh+d6s+fo1s -N${FIG} -Mm,png \
	-AIndianaJones_RaidersMarch.mp3+e -Cfhd -Fmp4 -Vi -D60 -K+p # -Zs
```
and I got the following message on the terminal:

`[image2 @ 0x55e57165d400] Thread message queue blocking; consider raising the thread_queue_size option (current value: 8)`

I similar issue appear to create a **WEBM**. In this case the files are different (with `diff`) but I can't see it.

## Solution

In both cases, I only have to add  `-thread_queue_size 4096` in the `ffmpeg` comand to fix it. I can do it.

## Doubt

My doubt is that I don't understand well what is going on, so I don't know if my solution may cause major problems later on or if there is no better solution. 

I found an explanation [here](https://video.stackexchange.com/questions/25643/ffmpeg-thread-queue-size-warning-what-does-this-mean) that it says "that your computer cannot keep up with encoding in real-time." I try the solution 

This is what the [ffmpeg docs](https://ffmpeg.org/ffmpeg.html) say:

> -thread_queue_size size (input/output)
> For input, this option sets the maximum number of queued packets when reading from the file or device. With low latency / high rate live streams, packets may be discarded if they are not read in a timely manner; setting this value can force ffmpeg to use a separate input thread and read packets as soon as they arrive. By default ffmpeg only does this if multiple inputs are specified.
> 
> For output, this option specified the maximum number of packets that may be queued to each muxing thread.
> 

So, if someone has something to say, please let me know. I will start making the PR.

Esteban82:
Fix #8394

### DIFF
--- a/src/movie.c
+++ b/src/movie.c
@@ -2721,7 +2721,7 @@ EXTERN_MSC int GMT_movie (void *V_API, int mode, void *args) {
 		else
 			sprintf (extra, "quiet");
 		sprintf (png_file, "%%0%dd", precision);
-		sprintf (cmd, "ffmpeg -loglevel %s -f image2 -framerate %g -y -i \"%s%c%s_%s.%s\"%s -vcodec libx264 %s -pix_fmt yuv420p %s.mp4",
+		sprintf (cmd, "ffmpeg -thread_queue_size 4096 -loglevel %s -f image2 -framerate %g -y -i \"%s%c%s_%s.%s\"%s -vcodec libx264 %s -pix_fmt yuv420p %s.mp4",
 			extra, Ctrl->D.framerate, tmpwpath, dir_sep, Ctrl->N.prefix, png_file, MOVIE_RASTER_EXTENSION, audio_option,
 			(Ctrl->F.options[MOVIE_MP4]) ? Ctrl->F.options[MOVIE_MP4] : "", Ctrl->N.prefix);
 		gmt_sleep (MOVIE_PAUSE_A_SEC);	/* Wait 1 second to ensure all files are synced before building the movie */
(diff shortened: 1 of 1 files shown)


Esteban82:
I test it on my script and it works.

Esteban82:
It is also related with #8391.

joa-quim:
Can't you achieve the same effect with **-F+o**? From man and a quick code reading I would say one can.
Adding fixed numbers in code (`-thread_queue_size 4096`) is normally not a good idea.

Esteban82:
> Can't you achieve the same effect with **-F+o**? From man and a quick code reading I would say one can. Adding fixed numbers in code (`-thread_queue_size 4096`) is normally not a good idea.

I didn't rememeber that. 

But, it didn't work. The mp4 was not even created. The problem is that the code (`-thread_queue_size 4096`) was added at the end of the `ffmpeg` command and it must be added at the start. `ffmpeg` is very sensity to the order of arguments, 

```
movie [NOTICE]: Running: ffmpeg -loglevel warning -f image2 -framerate 60 -y -i "/home/thor/Github/Esteban82/gmt-2024-animation/WED-A_Fig_8/WED-A_Fig_8_%04d.png" -i IndianaJones_RaidersMarch.mp3 -af atempo=1.94612 -vcodec libx264 -thread_queue_size 4096 -pix_fmt yuv420p WED-A_Fig_8.mp4
Option thread_queue_size (set the maximum number of queued packets from the demuxer) cannot be applied to output url WED-A_Fig_8.mp4 -- you are trying to apply an input option to an output file or vice versa. Move this option before the file it belongs to.
Error parsing options for output file WED-A_Fig_8.mp4.
Error opening output files: Invalid argument
movie [ERROR]: Running FFmpeg conversion to MP4 returned error 256 - exiting.
```

Esteban82:
> Adding fixed numbers in code (`-thread_queue_size 4096`) is normally not a good idea.

I also don't like adding fixed numbers. That's why I would like that tell ffmpeg to take all the time it needs to process the data.

joa-quim:
Well, we could change the order of the option insertion, but it looks that that option is [tricky to use](https://stackoverflow.com/questions/61723571/correct-usage-of-thread-queue-size-in-ffmpeg) what pleases one case can error in another. Not sure what to propose.

Esteban82:
> Well, we could change the order of the option insertion,

I don't thik so. `-F+o` is currently used (I think) to pass new arguments for the output file. 

We would need a similar option but for the input file (something like -F+i). And I don't know if we need one map for the video and another for the audio. `ffmpeg` is tricky.

joa-quim:
Can you try the `movie-Fi` branch (#8397)? Is not documented yet but what it does is to implement a **-F+i** option that works like **-F+o**. See if it solves the current issue.

Esteban82:
Ok, I will test it.

joa-quim:
Better addressed in #8397

PR #8425, opened 2024-04-05
Fixes in the algorithm that detects what type of transparency an image has.

joa-quim:
The `grdimage_transparencies()` function detects if an image has transparency as a separate layer (I->alpha) or as a 4th layer as in an RGBA image, but this function was failing in more than one case. To start with it was always failing for pixel interleaved RGBA images, and for other it was also failing depending on image `n_rows x n_cols` and it was not scanning the entire alpha layer. When it failed it fell into the variable transparency case that is quite heavy in PostScript size.

I run the tests and apparently (and I say apparently because I can't make the remote and DVC files to be in sync) only the  `transp_mix` test changed. But that test was already wrong because subplot (2,1) is exactly equal to (3,2) when the meant to be different. With this fix that sub-panel plot is still wrong as the transparency in figure is ignored. After many debugs I quit trying to find why it fails and also what was the original idea of that test. RGB images with a separate alpha layer also fail to set the transparency. Again, after long suffering trying to understand why, when all seem to go right, I gad to give up too. But this affects more the Julia side where  that case is more common and I'm forced to save image in file and read it again (this forces a conversion to RGBA). 

In summary, not all problems were solved but some important ones were.

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -983,10 +983,10 @@ GMT_LOCAL void grdimage_img_set_transparency (struct GMT_CTRL *GMT, struct GRDIM
 	 */
 	double o, t;		/* o - opacity, t = transparency */
 	gmt_M_unused (GMT);
-	o = pix4 / 255.0;	t = 1 - o;
 	if (Conf->Transp->n_transp == 2)
 		gmt_M_rgb_only_copy (rgb, Conf->tr_rgb);
 	else {	/* Blend */
+		o = pix4 / 255.0;	t = 1 - o;
 		if (Conf->invert) t = o, o = 1 - t;
 		rgb[0] = o * rgb[0] + t * Conf->tr_rgb[0];
 		rgb[1] = o * rgb[1] + t * Conf->tr_rgb[1];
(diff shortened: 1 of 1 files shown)


seisman:
> because I can't make the remote and DVC files to be in sync)

can you explain more about this? are there any errors?

joa-quim:
No errors, just > 50 failures mostly due to the use of different grids/images used on creating the original PS. When I do `dvc pull` it says is up to date and even added `export GMT_DATA_SERVER="static"` to the `gmtest.in` but nothing of that works.

seisman (approved):
Blindly approve.

seisman:
> I run the tests and apparently (and I say apparently because I can't make the remote and DVC files to be in sync) only the `transp_mix` test changed. But that test was already wrong because subplot (2,1) is exactly equal to (3,2) when the meant to be different. With this fix that sub-panel plot is still wrong as the transparency in figure is ignored.

Should we mark the `transp_mix` test as `GMT_KNOWN_FAILURE`?

joa-quim:
Yes. One day I may get enough patience again to try to find why the constant transparency is failing.

seisman:
Done in 0a10c48eb0 and pushed directly to master.

PR #8454, opened 2024-04-23
Configure dependabot to automatically update requirements.txt for docs

seisman:
We will be notified if any dependencies for building the GMT docs have updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,15 @@ updates:
     # Specify labels for pull requests
     labels:
       - "maintenance"
+
+  - package-ecosystem: "pip"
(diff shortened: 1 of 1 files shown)


joa-quim:
Can we lower down the frequency of this warnings/updates? Monthly perhaps? I'm starting to get too many of those software updates warnings that we probably do not even need at all but, I know, we also not lag too much behind.

seisman:
> Can we lower down the frequency of this warnings/updates? Monthly perhaps? I'm starting to get too many of those software updates warnings that we probably do not even need at all but, I know, we also not lag too much behind.

Done in https://github.com/GenericMappingTools/gmt/pull/8458.

PR #8466, opened 2024-04-30
CI: Update the SCM-check workflow

seisman:

Fixes #

**Reminders**


joa-quim:
Nothing against updating the workflow but honestly I don't see the interest on these frequent updates of the SCM that cause test breakings.

joa-quim (approved):

seisman:
I agree. I think we should be reminded that a new SCM is available, then we can decide if we want to update the CPTs or not.

PR #8477, opened 2024-05-06
Fix typos when parsing -h option

joa-quim:
Note that inside the `parse_h_option` function it was setting the variables for opt_g


seisman (approved):
Looks good. I've updated the PR title to make it more clear what this PR does.

PR #8482, opened 2024-05-12
CI: Bump to Ghostscript 10.03.0

seisman:
Bump the Ghostscript version to 10.03.0 in the CI jobs. No new failing tests, so this PR should be good to merge.


joa-quim (approved):
OK, but given the terrible past history of Ghostscript breaking versions I think we should restrain for doing updates unless they prove to come with some benefit.

Issue #6649, opened 2022-05-01
Don't use GMT_THEME=cookbook in cookbook scripts


seisman:
In some examples in the cookbook, we use the `GMT_THEME=cookbook` setting. For example:

<img width="915" alt="image" src="https://user-images.githubusercontent.com/3974108/166139304-c27f2e2f-c92a-448d-a130-6faa6a90ed10.png">

Why do we use it? Users may be confused that why `GMT_THEME=cookbook` is always presented in these examples.

joa-quim:
Agree.

meghanrjones:
It's included so that all the output images don't have different font and pen sizes due to the modern mode conf default of auto, which would look bad.

seisman:
> It's included so that all the output images don't have different font and pen sizes due to the modern mode conf default of auto, which would look bad.

I'm not fully convinced by this reason. If the images look bad due to the different font and pen sizes, we should update the scripts so that all basemaps have similar dimensions (thus similar font and pen sizes).

meghanrjones:
> > It's included so that all the output images don't have different font and pen sizes due to the modern mode conf default of auto, which would look bad.
> 
> I'm not fully convinced by this reason. If the images look bad due to the different font and pen sizes, we should update the scripts so that all basemaps have similar dimensions (thus similar font and pen sizes).

I recommend submitting a PR for one section of the cookbook with your solution rather than using the cookbook theme so that we can review the appearance.

joa-quim:
Now close this?

seisman:
there are still a lot to update

seisman:
Closed by #8509.

PR #8488, opened 2024-05-14
doc: Remove the cookbook GMT_THEME setting from more scripts

seisman:
This PR removes the `GMT_THEME` setting from some scripts used in the documentation. 

Only ~10 scripts are updated in this PR, so that it's easier to review the PR.

Address #6649.


Esteban82 (approved):
Looks fine.

PR #8506, opened 2024-06-01
Using gmt in NixOS

miniBill:
Adding NixOS instructions

### DIFF
--- a/doc/rst/source/install.rst
+++ b/doc/rst/source/install.rst
@@ -278,6 +278,15 @@ Install GMT via:
 
    sudo emerge --verbose --ask sci-geosciences/gmt
 
+NixOS
+~~~~~
+
+Install GMT by adding these packages to your system (or home-manager) configuration:
+
+::
+
+   gmt gshhg-gmt dcw-gmt
+
 Cross Platform Install Instructions
 -----------------------------------
 


doc/rst/source/install.rst
```diff
@@ -278,6 +278,15 @@ Install GMT via:
 
    sudo emerge --verbose --ask sci-geosciences/gmt
 
+NixOS
+~~~~~
+
+Use GMT in your shell via:
```
joa-quim, line 284: I guess you mean "Install GMT ..."?

miniBill: No, I intentionally wrote it that way.

`nix shell nixpkgs#packagename` will start a shell where `packagename` is available without the need to permanently install it on your system (when you close the shell, the package is not used/active anymore and is going to be eventually garbage collected).

To install it you would either add it to your system configuration or your home-manager configuration. Instructions on how to do that are probably unneeded for most NixOS users, the important part is the list of packages.

miniBill: You can also install the packages via `nix-env -iA`, but that's nondeclarative and honestly better not used at all in favor of a fully declarative approach.

joa-quim: OK, that sounds wasteful but I know nothing about NixOS (other than the name). But this is a section about how to install GMT. Wouldn't it be appropriate to add that info on NixOS installation as well?

miniBill: The installation instructions would be "add `gmt`, `gshhg-gmt` and `dcw-gmt` to your list of packages". I can add that, I guess.

joa-quim: Yes, please. And explain also that difference between using and installing. NixOS users may know that but others likely do not.

miniBill: Done

joa-quim: I mean, You can leave both informations. `Using` and `Installing` if you find it appropriate.

miniBill: Nah, this is fine

joa-quim (approved):

PR #7885, opened 2023-10-08
doc: Move install instructions from the repository root to the docs

seisman:
**Description of proposed changes**

Address https://github.com/GenericMappingTools/gmt/issues/6687#issuecomment-1746951809.

PR #8511, opened 2024-06-03
install.rst: remove broken links and improve formatting

anbj:
Minor improvements.

Why are building instructions in the wiki and not in the docs?

### DIFF
--- a/doc/rst/source/install.rst
+++ b/doc/rst/source/install.rst
@@ -234,7 +234,7 @@ Ubuntu/Debian
 
 **NOTE:** The Ubuntu/Debian official repositories may provide old GMT
 versions. If you want the latest GMT 6.x release, your best bet then is
-to `build the latest release from source <BUILDING.md>`__.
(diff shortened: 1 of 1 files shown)


seisman:
Building instructions are in https://github.com/GenericMappingTools/gmt/blob/master/BUILDING.md, not in the wiki.

anbj:
Ah, and that document isn't referenced in the technical reference, right?

Does that mean that all references to the wiki (build related) should instead link to BUILDING.md?

seisman:
> Ah, and that document isn't referenced in the technical reference, right?

Yes
 
> Does that mean that all references to the wiki (build related) should instead link to BUILDING.md?

Yes.

Originally, the installation instructions were in the `INSTALL.md` file in the root directory of the repository and were not included in the documentation, so the linking to `BUILDING.md` works. However, a few months ago (#7885), we decided to move the installation instructions to the docs but forgot to update the links to `BUILDING.md`.

PR #8512, opened 2024-06-03
Export 4 gmt_conf_* (they were local).

joa-quim:
Let the gmtlib_setparameter() function also change between CLASSIC & MODERN modes.


src/gmt_keywords.txt
```diff
@@ -176,3 +176,6 @@ TIME_UNIT			# Unit of time measurement
 TIME_SYSTEM			# Shortcut to setting TIME_EPOCH/UNIT [does not go in gmt.conf]
 TIME_WEEK_START			# Day a new week starts on
 TIME_Y2K_OFFSET_YEAR		# For 2-digit years to separate 1900 and 2000 centuries
+#-------------------------------------------------------
+SET_RUN_MODE		# To change between CLASSIC and MODERN mode. Used only by externals.
```
seisman, line 180: What's the use case of the new keyword? Is `GMT_RUN_MODE` a better name?

joa-quim: OK, renamed to GMT_RUN_MODE.

The use case, as mentioned in the comment, is to allow externals change to MODERN (and back if wished) mode. I need these type of manipulations in GMT.jl to use the autoscaling of the modern them while in a session whose API started as CLASSIC. GMT.jl uses almost only classic mode but some times it needs features only available in modern. For example, I have a working prototype of using the _inset_ command from within a classic session. This keyword makes things simpler.

src/gmt_init.c
```diff
@@ -12641,6 +12649,9 @@ char *gmtlib_getparameter (struct GMT_CTRL *GMT, const char *keyword) {
 	case_val = gmt_hash_lookup (GMT, keyword, keys_hashnode, GMT_N_KEYS, GMT_N_KEYS);
 
 	switch (case_val) {
+		case GMTCASE_SET_RUN_MODE:		/* Gat the current running mode CLASSIC or MODERN */
```
seisman, line 12652:
```suggestion
		case GMTCASE_GMT_RUN_MODE:		/* Gat the current running mode CLASSIC or MODERN */
```

src/gmt_init.c
```diff
@@ -10930,6 +10930,14 @@ unsigned int gmtlib_setparameter (struct GMT_CTRL *GMT, const char *keyword, cha
 	case_val = gmt_hash_lookup (GMT, keyword, keys_hashnode, GMT_N_KEYS, GMT_N_KEYS);
 
 	switch (case_val) {
+		case GMTCASE_SET_RUN_MODE:		/* Allow changing to/from CLASSIC/MODERN mode */
```
seisman, line 10933:
```suggestion
		case GMTCASE_GMT_RUN_MODE:		/* Allow changing to/from CLASSIC/MODERN mode */
```

joa-quim:
Ofc, thanks.

seisman (approved):

PR #8522, opened 2024-06-17
Fix the color depth returned when reading the gray.cpt master CPT.

joa-quim:
The master CPT gray.cpt has only one slice that fits both the `is_gray` and `is_bw` tests and hence returns with both as true, which is obviously wrong and cause that when passed to grdimage together with a grid from Julia the image_8 container was not initialized and a crash followed (see https://github.com/GenericMappingTools/GMT.jl/issues/1472).


remkos (changes requested):
Apart from changing "aout" to "out" in the comment above the code line, I think this is the right, and generic, approach.

joa-quim:
Corrected the typo.

Issue #8525, opened 2024-06-20
grdinfo -L0 doesn't work for external wrappers


seisman:
Here is the C source codes that is modified from the `testapi_gmtgrid.c` function:
```
#include "gmt.h"

int main () {
	unsigned int mode = GMT_SESSION_EXTERNAL;
	struct GMT_GRID *G = NULL;
	struct GMTAPI_CTRL *API = NULL;
	uint64_t dim[2] = {5, 4};
	char args[1000] = {""};
	char input[GMT_VF_LEN] = {""};
	gmt_grdfloat *data = NULL;

	API = GMT_Create_Session ("testapi_gmtgrid", 2U, mode, NULL);

	/* Create a container for the grid */
	G = GMT_Create_Data (API, GMT_IS_GRID, GMT_IS_SURFACE, GMT_CONTAINER_ONLY, dim, NULL, NULL, 0, 2, NULL);

	/* Allocate memory of the data array in the external program (C or PyGMT) */
	data = (gmt_grdfloat *)malloc(sizeof(gmt_grdfloat) * 9 * 8);
	for (int i = 0; i < 9 * 8; i++) data[i] = (gmt_grdfloat)i;

	/* Assign the user data to the GMT_GRID structure */
	G->data = data;

	/* Using the grid */
	GMT_Open_VirtualFile (API, GMT_IS_GRID, GMT_IS_SURFACE, GMT_IN, G, input);
	sprintf (args, "%s -L0 -C", input);
	GMT_Call_Module (API, "grdinfo", GMT_MODULE_CMD, args);
	GMT_Close_VirtualFile (API, input);

	/* Free the data array allocated by the external program */
	free (data);

	if (GMT_Destroy_Session (API)) return EXIT_FAILURE;
	exit (0);
}
```
Compiling and running the C program returns:
```
0	0	0	0	47	47	0	0	1	1	0	0
```
Without `-L0`, it returns:
```
0	4	0	3	NaN	NaN	1	1	5	4	0	0
```

joa-quim:
Hmm, it works for me in Julia
```
julia> G = mat2grid(reshape(collect(0.0f0:8*9-1), 8,9))
A GMTgrid object with 1 layers of type Float32
Gridline node registration used
x_min: 1.0      x_max :9.0      x_inc :1.0      n_columns :9
y_min: 1.0      y_max :8.0      y_inc :1.0      n_rows :8
z_min: 0.0      z_max :71.0
[...]

PR #8528, opened 2024-06-24
grdinfo: Fix pad for external wrappers and update grid header for -L option

seisman:
Closes https://github.com/GenericMappingTools/gmt/issues/8525.

### DIFF
--- a/src/grdinfo.c
+++ b/src/grdinfo.c
@@ -174,7 +174,7 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Usage (API, 3, "a: All of the above.");
 	GMT_Usage (API, -2, "Note: If grid is geographic then we report area-weighted statistics.");
 	GMT_Usage (API, 1, "\n-M[c|f]");
-	GMT_Usage (API, -2, "\nSearch for the global data min and max locations (x0,y0{,z0}) and (x1,y1{,z1}) [Default]." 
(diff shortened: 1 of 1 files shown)


src/grdinfo.c
```diff
@@ -689,8 +689,6 @@ EXTERN_MSC int GMT_grdinfo (void *V_API, int mode, void *args) {
 
 	Out = gmt_new_record (GMT, (n_cols) ? out : NULL, (cmode) == GMT_COL_FIX ? record : NULL);	/* the two items */
 
-	gmt_set_pad (API->GMT, 0);	/* Not read pads to simplify operations */
```
seisman, line 692: @joa-quim Do you see any side effects removing this line?

joa-quim:
_simplify operations_ does not seems a reason good enough. Specially when it has buggish side effects, so I would say we should remove it and see if there are any noticeable side effects. 

The pad is big annoyance for externals transfering grids in memory because it implies making a full copy of the array and for many operations, grdinfo for example, the pad is not necessary at all. In Julia I try to distinguish uses where pad is needed (projections, gradients, etc...) from others where it doesn't and in those case I set the pad to zero.

joa-quim (approved):

seisman:
No new failures and the original bug is fixed. Merging.

joa-quim:
Well, a little bit by reasoning + trial-error. Making a projected image needs pad but a linear one (-JX) does not. Shading needs pad because of the gradient. Also, when a module needs the pad and grid/image do not have it, an error is issued.

seisman:
Good to know. Will see how it works in PyGMT.

PR #8538, opened 2024-07-03
Fix the issue with x_0 & y_0 in proj4 string used in pscoast

joa-quim:
The workaround found is to catch these specific cases in `pscoast`, strip the `+x_0=... +y_0=...`, recreate the GDAL CT (coordinate transforms), do the plotting and reset original CTs at the end that may be needed for example in mapproject, though this later may only make sense for externals.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -18256,7 +18256,7 @@ unsigned int gmt_parse_inc_option (struct GMT_CTRL *GMT, char option, char *item
 	return GMT_NOERROR;
 }
 
-GMT_LOCAL int gmtinit_parse_proj4 (struct GMT_CTRL *GMT, char *item, char *dest) {
+GMT_LOCAL int gmtinit_parse_proj4(struct GMT_CTRL *GMT, char *item, char *dest) {
 	/* Deal with proj.4 or EPSGs passed in -J option */
 	char  *item_t1 = NULL, *item_t2 = NULL, epsg2proj[GMT_LEN256] = {""}, wktext[32] = {""}, *pch;
 	bool   do_free = false;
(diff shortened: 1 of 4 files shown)


seisman:
Did you forget to add the changed `test/baseline/pscoast.dvc` file or forget to run `dvc push`?

joa-quim:
Neither, I think.

I'm just fed up with DVC. It always fck for me. I followed the exactly the recipe on
https://docs.generic-mapping-tools.org/dev/devdocs/contributing.html
up to 
``git add test/baseline/<module>/<newplot.ps>.dvc``
that it refused saying some shit of not existing, and I replaced it with
```
git add test/baseline/pscoast/pscoast.dvc
``` 
that had local changes needed to be committed. Then I did a `git commit` like I always do for normal src changes (with Turtoisegit) and checked that only that commit was submitted.

seisman:
I think the guides are incorrect. The correct steps should be:

1. Put the correct PS file at `test/baseline/pscoast/proj4_nox0y0.ps`
2. Run `dvc add test/baseline/pscoast/proj4_nox0y0.ps`
3. Run `git add test/baseline/pscoast.dvc`
4. Run `git commit`

seisman:
If you can upload the PS file to GitHub, then I can try to help add it to dvc.

joa-quim:
Thanks.
It's a tinny file. But this situation with DVC is unsustainable. All tests now fail with an incomprehensible message.

[proj4_nox0y0.zip](https://github.com/user-attachments/files/16091651/proj4_nox0y0.zip)

seisman:
Done in the master branch

PR #8541, opened 2024-07-08
Fix crash when asking grdcut to output file with no extension.

### LINKED ISSUE #8537: grdcut: segmentation fault if output extension is omitted
See forum for details:  https://forum.generic-mapping-tools.org/t/error-recognizing-tif-files-from-remote-dataset/5035/7

I offer to fix it (with Joaquim guidance).

The source code should be fixed. I think I will have to edit this file: 
https://github.com/GenericMappingTools/gmt/blob/master/src/grdcut.c

joa-quim:
Fix #8537

### DIFF
--- a/src/gmt_gdalwrite.c
+++ b/src/gmt_gdalwrite.c
@@ -104,7 +104,7 @@ int gmt_export_image(struct GMT_CTRL *GMT, char *fname, struct GMT_IMAGE *I) {
 
 	ext = gmt_get_ext(fname);
 	if (ext == NULL) {
-		GMT_Report(GMT->parent, GMT_MSG_ERROR, "Cannot NOT provide a file extension for the saving file (-G option). We need it to determine the output format.\n");
+		GMT_Report(GMT->parent, GMT_MSG_ERROR, "MUST provide a file extension for the saving file (-G option). We need it to determine the output format.\n");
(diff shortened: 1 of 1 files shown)


src/gmt_gdalwrite.c
```diff
@@ -102,7 +102,11 @@ int gmt_export_image (struct GMT_CTRL *GMT, char *fname, struct GMT_IMAGE *I) {
 		to_GDALW->ULy = I->header->n_rows;
 	}
 
-	ext = gmt_get_ext (fname);
+	ext = gmt_get_ext(fname);
+	if (ext == NULL) {
+		GMT_Report(GMT->parent, GMT_MSG_ERROR, "Cannot NOT provide a file extension for the saving file (-G option). We need it to determine the output format.\n");
```
seisman, line 107:
```suggestion
		GMT_Report(GMT->parent, GMT_MSG_ERROR, "Can NOT provide a file extension for the saving file (-G option). We need it to determine the output format.\n");
```

seisman (approved):

src/gmt_gdalwrite.c
joa-quim: OK, rephrased it to "MUST provide ..."

PR #8546, opened 2024-07-17
Let pscoast -E accept reading other DCW type files made by users.

joa-quim:
In GMT.jl I added a function that reads an OGR file containing country borders and attributes describing each country polygon with 2 or 3 chars ISO codes and create a file with the same structure as the dcw-gmt.nc file.

That file can then be read by pscoast via -E like the official DCW file. In order to do it the +f<file> modifier was added.

Besides that, this PR adds also the _continent_ **WD**  to mean the entire world.

I have prepared one of such DCW files with data from https://public.opendatasoft.com/explore/dataset/world-administrative-boundaries/export/ and made it available at the gmtsetver /export/gmtserver/gmt/data/geography/dcw/ODS.nc.
It is only 1.3 MB big and with it we can do:

```
gmt coast -E=WD+fc:\v\ODS -JQ15 -Ba -png world_ODS
```
world_ODS

I have also a 60 MB file with the OSM countries if anyone is interested.

### DIFF
--- a/doc/rst/source/coast.rst
+++ b/doc/rst/source/coast.rst
@@ -111,7 +111,7 @@ Optional Arguments
 
 .. _-E:
 
-**-E**\ *code1,code2,...*\ [**+l**\|\ **L**\|\ **n**][**+c**\|\ **C**][**+e**][**+g**\ *fill*][**+p**\ *pen*][**+r**][**+R**][**+z**]
+**-E**\ *code1,code2,...*\ [**+f**\ *file*][**+l**\|\ **L**\|\ **n**][**+c**\|\ **C**][**+e**][**+g**\ *fill*][**+p**\ *pen*][**+r**][**+R**][**+z**]
     Select painting, clipping or dumping country polygons from the Digital Chart of the World (DCW).
     This is another dataset independent of GSHHG and hence the |-A| and |-D| options do not apply. The following codes
     are supported:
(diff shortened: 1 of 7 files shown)


seisman:
It's interesting.

> I have also a 60 MB file with the OSM countries if anyone is interested.

Does it mean it's possible to replace the current DCW data with OSM data?

joa-quim:
Well, yes but probably a lot of work. For now I left out higher administrative levels (before Federico and Andreas ask) because I wanted to make this work first and it was already quite some work. Also some of the DCW polygons are very GMT peculiar. The Russian one goes to ~30 to 190 and when we request data in [-180 180] GMT trims the > 180 and puts it in West. This if fine for mapping, when it works but we have bugs that make this fail often with low resolution data, but definitely not good when one want to do point-in-polygon (`gmtselect`) that is a Cartesian operation. Also the OSM data has very high resolution and not even it has borders that fit perfectly between two neighbor countries.

One intention was if I could expand the julia code to take GADM and(or OSM countries individually and build _DCW's_ with higher administrative levels. We wouldn't be distributing those files (we can't for GADM) but users could build them easily by themselves.

Esteban82 (approved):
It works fine. I got the same map. 

My only doubt is why you especify the file name without extension. Does it searchs for a .nc file?

seisman:
Instead of adding a `+f` modifier, can we add a new GMT configuration (we already have the `DIR_DCW` configuration key) that points to the DCW data by default but can be changed to use user's own nc file like what you're providing?

joa-quim:
> My only doubt is why you especify the file name without extension. Does it search for a .nc file?

Yes, it searches for a .nc file and name given may have, or not, that extension.

Regarding the `+f` modifier. I thought in a new config variable too but gave up because it wouldn't be different of `DIR_DCW` and in fact we can use that variable to indicate the location of the new file but things are a bit more complicated this way and that's why I didn't documented it. The issue is that the files `dcw-countries.txt` (+states +collections) would have to copied too and the file names kept the same (including the nc file). A problem is that because in any module we can give a **-RXX**, where **XX** is a country code, the GMT machinery searches for those `.txt` files even before the parsing of a modules option. This means that when we use **-EIT** it already scanned `dcw-countries.txt` to find the country code is valid and finding the region. This is a difficulty if we ever will want to have files with administrative zones other than those included in  `dcw-states.txt`. A new logic will need to be created and implemented.

All of that to say that the `+f` modifier was the simplest way of implementing this change.

joa-quim:
BTW, I updated the ODS.nc file in the gmtserver with one patched to have also the Antarctica polygons that were missing in the opendatasoft file.

Esteban82:
Great. Is something else to check?

joa-quim:
If there are no other questions I'll merge this today.

PR #8547, opened 2024-07-18
gmt_common_sighandler.c: fix ucontext for Apple

barracuda156:

Fix ucontext macros for macOS.

Current code falls back to x86 version on PowerPC, which is obviously wrong.

**Reminders**

Perhaps this is a trivial bug fix, there is no need to document it specifically.

P. S. I am not sure FreeBSD macro for PowerPC is correct. AFAIK, `__ppc__` is Apple-specific.
@pkubaj Could you please take a look? (I am not changing anything for FreeBSD in this PR, I mean to verify what the source has now.)


pkubaj:
LGTM.

barracuda156:
CI failures are in tests, which is obviously unrelated: an error in macros here would lead to a build-time error on macOS.

seisman (approved):

PR #8548, opened 2024-07-20
Update in the docs to netcdf 1.7

Esteban82:
I create a grid and I saw with grdinfo that the files formats are CF-1.7 (not CF-1.5).

I made this PR to update this.

Approve ONLY if it is correct

### DIFF
--- a/doc/rst/source/reference/features.rst
+++ b/doc/rst/source/reference/features.rst
@@ -2336,15 +2336,15 @@ approach should contact the GMT team for guidance.
 +==========+===============================================================+
 |          | *GMT 4 netCDF standard formats*                               |
 +----------+---------------------------------------------------------------+
-| nb       | GMT netCDF format (8-bit integer, COARDS, CF-1.5)             |
+| nb       | GMT netCDF format (8-bit integer, COARDS, CF-1.7)             |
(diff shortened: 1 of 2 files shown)


seisman (approved):
Looks correct to me.

joa-quim:
But why are we doing this? Nothing changed in the netCDF usage. What is the difference between CF 1.5 and CF 1.7 that affects anything we use in GMT?

seisman:
Just make the documentation consistent with the GMT output.

joa-quim:
OK. Still think useless CI Joules burning, but OK.

Esteban82:
> But why are we doing this? Nothing changed in the netCDF usage. What is the difference between CF 1.5 and CF 1.7 that affects anything we use in GMT?

I agree, nothing major. But I'm going to teach a class on GMT grid formats and I want the documentation to say the same thing as the terminal information. If I was sure, I would have modified it directly on master.

joa-quim:
Than the most correct thing to do is to remove that CF info from the grdinfo print. The thing is that the ndetCDF grids are **not** a GMT format and can, and actually are, be produced by many different programs. They are all reported to conform with CF1.7, but we have no idea if that is even true.

PR #8553, opened 2024-07-25
Fix the grdimage and grdview -T online syntax

seisman:

Current description of the `-T` syntax is inaccurate and misleading:
```
  -T[<width>[c|i|p]],[<color>],[<style>[c|i|p]]
     Image the data without interpolation by painting polygonal tiles in the form [+o[<pen>]][+s].
     <pen> is a comma-separated list of three optional items in the order: <width>[c|i|p], <color>, and
       <style>[c|i|p].
       <width> >= 0.0 sets pen width (Default unit is point); alternatively, give a pen name: Choose from faint,
          default, [thin|thick|fat][er|est] or wide.
       <color> = <gray> or <red>/<green>/<blue>, each in the range 0-255; #rrggbb, each in the range 00-FF using
          hexadecimal numbers; <cyan>/<magenta>/<yellow>/<black> each in 0-100% range; <hue>-<saturation>-
          <value> in ranges 0-360, 0-1, 0-1, respectively; a valid color name.
       <style> = pattern of dashes (-) and dots (.), scaled by <width>; "dashed", "dotted", "dashdot",
          "dotdash", or "solid"; <pattern>[:<offset>]; <pattern> holds lengths [Default unit is points] of any
          number of lines and gaps separated by underscores. The optional <offset> phase-shifts elements from
          start of the line [0].
     For PDF stroke transparency, append @<transparency> in the range 0-100% [0 = opaque].
       +s Skip tiles for nodes with z = NaN [Default paints all tiles].
       +o to draw tile outline; optionally append <pen> [Default uses no outline].
     Note: Cannot be used with -Jz|Z as it produces a flat image.
```

This PR changes it to:
```
  -T[+o[<pen>]][+s]
     Image the data without interpolation by painting polygonal tiles.
       +s Skip tiles for nodes with z = NaN [Default paints all tiles].
       +o to draw tile outline; optionally append <pen> [Default uses no outline].
     Note: Cannot be used with -Jz|Z as it produces a flat image.
```

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -240,7 +240,8 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Usage (API, 3, "+i Invert presumed opacities to transparencies (i.e., 1 - opacity).");
 	GMT_Usage (API, 3, "+z Specify grid <value> to set transparent pixel color via CPT lookup.");
 	GMT_Option (API, "R");
-	gmt_pen_syntax (API->GMT, 'T', NULL, "Image the data without interpolation by painting polygonal tiles in the form [+o[<pen>]][+s].", NULL, 0);
+	GMT_Usage (API, 1, "\n-T[+o[<pen>]][+s]");
+	GMT_Usage (API, -2, "Image the data without interpolation by painting polygonal tiles.");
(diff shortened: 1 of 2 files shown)


joa-quim (approved):
Agree. Too much detail for the online help.

PR #8523, opened 2024-06-17
WIP: Use all cores in modules built with Glib (grdfilter, gmt|grdgravmag3d)

joa-quim:

PR #8577, opened 2024-09-05
CMake: Need to install gmt_glib.h since it's included in gmt_dev.h

seisman:
See https://forum.generic-mapping-tools.org/t/missing-header-file/5261 for the original issue report.

The bug was introduced at https://github.com/GenericMappingTools/gmt/pull/8523.

In previous versions, `gmt_glib.h` is not included in `gmt_dev.h`, but after that PR, `gmt_glib.h` is included, so we need to install this header file.


joa-quim (approved):

thwbecker:
Thanks, guys!

PR #8579, opened 2024-09-08
The psl_patch() function is intended for small polygons but it doesn't fill them.

joa-quim:
This PR fixes it.

This is what we get. Note that the -G+z is ignored but the segment headers have -Z<val> info  
```
makecpt -T-500/0/20 > cpt.cpt
gmt plot3d D.dat -p217.5/30 -JX15c/10c -Baf -Bza -R253/278/7.6/21.6/-400/-0 -JZ3c -Ccpt.cpt -W+z -G+z -png tri_1.png
```

tri_1 png

With this PR we get the triangles filled, though we still have to set -L even though the triangles are closed.
```
gmt plot3d D.dat -p217.5/30 -JX15c/10c -Baf -Bza -R253/278/7.6/21.6/-400/-0 -JZ3c -Ccpt.cpt -G+z -L -png tri_2.png
```
tri_2 png

[D.zip](https://github.com/user-attachments/files/16923926/D.zip)


seisman (approved):
Blindly approve.

PR #8591, opened 2024-10-02
symbol was compared with GMT_SYMBOL_CUBE twice instead of GMT_SYMBOL_COLUMN

joa-quim:
Also moved DEBUG print to outside a for loop


seisman (approved):

seisman:
it worked yesterday. there must be.some changes outside GMT. Will see what happened tomorrow.

PR #8592, opened 2024-10-04
Add a +n option to -S to prevent saving node numbers in headers.

joa-quim:
The reason for this is that information is normally useless but takes up needless space when the dataset is exported to the wrappers.

Also initialized some variables to please link checker.

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -7272,7 +7272,8 @@ GMT_LOCAL void gmtinit_conf_modern_override (struct GMT_CTRL *GMT) {
 	GMT->current.setting.map_grid_pen[GMT_SECONDARY].width = GMT->session.d_NaN;	/* 0.5p (thinner) */
 
 	if (error)
-		GMT_Report (GMT->parent, GMT_MSG_ERROR, "Unrecognized value during gmtdefaults modern initialization.\n");}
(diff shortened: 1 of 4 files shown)


src/gmt_init.c
```diff
@@ -20541,16 +20542,18 @@ void gmt_auto_offsets_for_colorbar (struct GMT_CTRL *GMT, double offset[], int j
...
 	gmt_M_memcpy (sides, GMT->current.map.frame.side, 5U, unsigned int);
 	was = GMT->current.map.frame.draw;
 	gmtinit_conf_modern_override (GMT);	/* Reset */
 	(void)gmt_getdefaults (GMT, NULL);
-	for (opt = options; opt; opt = opt->next) {
-		if (opt->option != '-') continue;   /* Not a parameter setting */
-		if ((c = strchr (opt->arg, '=')) == NULL) continue;
-		c[0] = '\0';  /* Remove = */
-		n_errors += gmtlib_setparameter (GMT, opt->arg, &c[1], false);
+	if (!GMT->parent->external || options->option) {	/* So that externals can send a NULL ptr for options. 'Internal' is not affected */
```
seisman, line 20550: @joa-quim It seems changes in these lines cause a crash for PyGMT, likely because the `options` pointer is `NULL` in our case. Perhaps we should change this line to:
```
if (!GMT->parent->external || options) {
```

PR #8593, opened 2024-10-06
Check if options is NULL rather than options->option

seisman:

Address https://github.com/GenericMappingTools/gmt/pull/8592#discussion_r1788831719.
Patches #8592.


joa-quim (approved):
OK, but this was not changed in #8593. It’s here for quiet some time (if I remember well).

seisman:
It was changed in #8592. See below for the diff:
https://github.com/GenericMappingTools/gmt/pull/8592/files#diff-fbd4b88ec2e0eb0c696a1fef67f37f1ac34e561b2cde8f5adeefbf824cba042f

joa-quim:
Ah, I remember a bit better now (I’m on mobile, cannot use tools). I had a previous older change in my local file. The +n in triangulate had nothing to do with it. But still strange that I hadn’t made a separate PR got it. Sorry.

PR #8598, opened 2024-10-10
anim07.sh. Data registration (_g vs _p).

Esteban82:
grdgradient and grdimage uses different data registration (see  https://docs.generic-mapping-tools.org/dev/datasets/remote-data.html#data-registration).

This causes an error in this animation. 
I append  _p for grdgradient to solve it.

Approve if you think that is the best solution.

It would possible to change in the version 6.5?


Esteban82:
The same issue happens with animation 08 and 11. I made a commit.

joa-quim (approved):

Issue #8607, opened 2024-10-31
Improve gmt events docs for events without duration


Esteban82:
The video and the graphs explain the case when the event has a duration. But this is optional. The docs should be improve to describe better when the event has no duration. 

I could do it but after the AGU.

Esteban82:
For example I made this animation with this code

```
gmt events violet.t -T${MOVIE_COL0} -Sc2c -W1p -Gblueviolet
gmt events blue.t   -T${MOVIE_COL0} -Sc2c -W1p -Gblue   -Ms1.5+c0.25 -Mi1+c-0.5 -Mt+c0 -Es+r0.25
gmt events green.t  -T${MOVIE_COL0} -Sc2c -W1p -Ggreen  -Ms1.5+c0.25 -Mi1+c-0.5 -Mt+c0 -Es+r0.25+d0.25
gmt events yellow.t -T${MOVIE_COL0} -Sc2c -W1p -Gyellow -Ms1.5+c0.25 -Mi1+c-0.5 -Mt+c0 -Es+r0.25+p0.25
gmt events orange.t -T${MOVIE_COL0} -Sc2c -W1p -Gorange -Ms1.5+c0.25 -Mi1+c-0.5 -Mt+c0 -Es+r0.25+p0.25+d0.25
gmt events red.t    -T${MOVIE_COL0} -Sc2c -W1p -Gred    -Ms1.5+c0.25 -Mi1+c-0.5 -Mt+c0 -Es+r0.25+p0.25+d0.25+f0.25
```


https://github.com/user-attachments/assets/c8757dc9-bf8b-4c91-9d5f-e7b382613875

Full script:

```
#!/usr/bin/env bash
#
# Video 1 in this paper: WED-A_Vid_1.sh
# https://github.com/GenericMappingTools/gmt-2024-animation
#
# Wessel, Esteban & Delaviel-Anger, 2024,
#	The Generic Mapping Tools and Animations for the Masses,
#	Geochem. Geophys. Geosyst.
#
# Purpose: Show short movie of the life of two simple events.
# The movie took 53 seconds to render on an 8-core Intel® Core™ i7-3700 CPU @ 3.60GHz.
#--------------------------------------------------------------------------------
NAME=Tutorial_1
NAME=$(basename $0 .sh)
#
# Create a short MP4 showing the life of two simple events.
# Both events starts at 0 and ends at 1, but we add rise
# and fade and coda to one of them (red) and let the other
[...]

PR #8611, opened 2024-11-02
Fix events for infinite symbols

Esteban82:
## Intro

The [description](https://docs.generic-mapping-tools.org/dev/events.html#description) about the coda phase in the events documentation mentions:

> " You may also wish for symbols to disappear completely after they reach their end time (we call this period the coda), or perhaps remain visible but faded, shrunk, or darkened."

Also, at the end of the [-M](https://docs.generic-mapping-tools.org/dev/events.html#m) section, it says:

> Optionally, for finite-duration events that should remain visible for all times after their event end time has been reached you must append +c (for coda) to set the corresponding terminal value during the coda.  If +c is not given then the defaults are 0 (intensity), 0 (size), 100 (transparency) and 0 (value), meaning the symbols are not plotted unless you change these attributes with one or more +c modifiers (one per directive). 

This implies that symbols without a defined duration (infinite) shouldn’t reach the coda phase. However, currently, this is what is happening (as seen in animation #8607 and [animation 08](https://docs.generic-mapping-tools.org/dev/animations/anim08.html)).

This also means that I have to define the symbol properties via the coda phase modifiers ([-M+c](https://docs.generic-mapping-tools.org/dev/events.html#m)). In fact, I must include these modifiers to make the symbols visible (e.g., like -Mt+c0), because the default behavior is for them to become invisible.

## Proposed Fix

I made a fix to address this issue. The problem was (I believe) that only events with finite duration could reach the normal phase. So, in this PR, I remove that condition. I tested it on my PC (with the same script from the previous link) and got this animation, which looks fine to me:

https://github.com/user-attachments/assets/21fb3af1-f8ec-48df-bd1a-580df9bdd310

I think this solution is better because now the symbols at the end retain the properties defined by -Sc and -G (without the -M+c modification). This means the code could be simplified:

```
gmt events violet.t -T${MOVIE_COL0} -Sc2c -W1p -Gblueviolet
gmt events blue.t   -T${MOVIE_COL0} -Sc2c -W1p -Gblue   -Ms1.5 -Mi1 -Mt -Es+r0.25
gmt events green.t  -T${MOVIE_COL0} -Sc2c -W1p -Ggreen  -Ms1.5 -Mi1 -Mt -Es+r0.25+d0.25
gmt events yellow.t -T${MOVIE_COL0} -Sc2c -W1p -Gyellow -Ms1.5 -Mi1 -Mt -Es+r0.25+p0.25
gmt events orange.t -T${MOVIE_COL0} -Sc2c -W1p -Gorange -Ms1.5 -Mi1 -Mt -Es+r0.25+p0.25+d0.25
gmt events red.t    -T${MOVIE_COL0} -Sc2c -W1p -Gred    -Ms1.5 -Mi1 -Mt -Es+r0.25+p0.25+d0.25+f0.25
```

I also try the original animation made by Paul ([WED-A_Vid_1.sh](https://github.com/GenericMappingTools/gmt-2024-animation/blob/main/WED-A_Vid_1.sh)) and I got the same animation.


joa-quim (approved):
I can't say that I fully understand the issue but I've not worked with animations for quite some time. But you do, so I trust that this is an improvement.

Esteban82:
I am quite confident that it is an imporvement necessary. I am confident (although a little less so) that the modification I made to the code is the correct one. But I guess if it is wrong, we will figure it out. My tests were fine.

Esteban82:
This PR affects the output of this test (`test/psevents/events.sh (Failed)`). 
I have to see how to update that.

Esteban82:
I added # GMT_KNOWN_FAILURE in the test script that has to be updated. I will update the reference image (in DVC) in another PR. I have doubts on how to do it.

Esteban82:
What are those error with the tests? May I merge it anyway?  I would think so but I am not very sure.

seisman:
The tests are unrelated to changes in this PR. So please merge.

PR #8612, opened 2024-11-07
Fix the variable transparency case that was broken by the #8255 commit.

joa-quim:
The fix in the #8255 was actually very incomplete and left the working array still pointing to a zero filled array and hence this type of transparency was totally broken. This PR fixes that. But this PR also addressees other situations.

- For the variable transparency case, and given that it implies cresting a **much** bigger _ps_ that considerably slows down the image rasterization, a new modifier **+t** is added to **-Q** to select true transparency. Otherwise, and by default, a blended transparency, that is an opaque combination of colors and alpha's, is computed.
- The single transparency value case was not working at all. This PR now does also the blended composition by default.
- The **-Q** docs are modified but I don't think they are telling the full truth. Debugging the code I don't see where a 2-values transparency layer is creating a true transparent figure. I always see the code path falling into the `grdimage_img_set_transparency` function that either copies a color or blends it. I don't know enough postscript to follow this matter any further.

If someone wants to play with a variable transparency file, here is one.
stefan_rgba

EDIT: The SC bellow sows a cut of the original of the `transp_mix` test. The top left image (that in fact is the second row, first col of the full image) shows the same grid pattern as the lower left. This grid pattern in PS is what we see when using the true variable transparency that works by drawing a rectangle for each pixel (that is why those PS are huge). So, at a given point the single value transparency (obtained for example with `grdmix -A`) must have been working, but as I said above, I have no idea how. I think we should replace that orig PS by the one produced now by the `transp_mix.sh` script and remove the  `GMT_KNOWN_FAILURE`. @seisman Coul you do it after this RR is merged? I am not able to update original PS's.

image

### DIFF
--- a/doc/rst/source/grdimage.rst
+++ b/doc/rst/source/grdimage.rst
@@ -23,7 +23,7 @@ Synopsis
 [ |-I|\ [*file*\|\ *intens*\|\ **+a**\ *azimuth*][**+d**][**+m**\ *ambient*][**+n**\ *args*] ]
 [ |-M| ]
 [ |-N| ]
-[ |-Q|\ [*color*][**+i**][**+z**\ *value*] ]
+[ |-Q|\ [*color*][**+i**][**+t**][**+z**\ *value*] ]
 [ |SYN_OPT-Rz| ]
 [ |-T|\ [**+o**\ [*pen*]][**+s**] ]
 [ |SYN_OPT-U| ]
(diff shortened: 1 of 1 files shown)


joa-quim:
Looks something is holding running the CI tests. I run them locally and no new failures were found.

joa-quim:
Strange. This is what I get with this branch after removing the `GMT_KNOWN_FAILURE` from `transp_mix.sh` and running `ctest -R transp_mix`

transp_mix.png

transp_mix


transp_mix_diff.png

transp_mix_diff

Esteban82:
Ok, I will try again at my home.

Esteban82:
> Ok, I will try again at my home.

I got this image. I don't know what is happening.

transp_mix

seisman:
Sorry I don't get it. There are only documentation changes in this PR, no?

joa-quim:
Ghrrr, I made the commit from thr rst directory and the C files were left in my computer only. Now in #8613

joa-quim:
Ghrrr, I made the commit from the rst directory and the C files were left in my computer only. Now in #8613

PR #8616, opened 2024-11-14
Let the -U+t (i.e. without any argument to the +t modifier) plot no stamp at all.

joa-quim:
Now this plots only the GMT logo and nothing else. Before this PR it still plotted the time stamp.
```
gmt plot -R0/3/0/0.1 -Jx1i -U+t -T  -view GMT_-U
```

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -2527,43 +2527,46 @@ int gmt_parse_model (struct GMT_CTRL *GMT, char option, char *in_arg, unsigned i
 #define is_text(item,k)  (item[k] == 't' && item[k+1])	/* Modifier t followed by random text */
 
 /*! Parse the -U option.  Full syntax: -U[<label>][+c][+j<just>][+o<dx>[/<dy>]][+t<string>]  Old syntax was -U[[<just>]/<dx>/<dy>/][c|<label>] */
-GMT_LOCAL int gmtinit_parse_U_option (struct GMT_CTRL *GMT, char *item) {
+GMT_LOCAL int gmtinit_parse_U_option(struct GMT_CTRL *GMT, char *item) {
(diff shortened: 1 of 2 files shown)


Esteban82:
I got this with master

GMT_-U_previo


and this with this branch

GMT_-U

I use this command with `png` (instead of `view`). Is view a thing? I got a weird message. 

`gmt plot -R0/3/0/0.1 -Jx1i -U+t -T  -png GMT_-U`

Esteban82 (approved):

joa-quim:
I confess I just learnt about the -view in https://docs.generic-mapping-tools.org/latest/reference/options.html#timestamps-on-plots-the-u-option

But yes, the plots you posted correctly show the situation. **+t** is supposed to replace the time stamp but only worked if it has an argument. A trick was to set `+t" "` but that is ugly and hard to discover.

rbdavis:
Hi Roger,

I'm not cc'ing anyone. You are probably receiving these messages because you are (to be confirmed) in the 'core' group that I indeed select for reviewing the PR's I submit.

Cheers

Joaquim
 

-----Original Message-----

seisman:
Yes, Roger is in the core group. I can remove him from the "core" group or maybe we can have a "gmt-core" team for people who'd like to review the GMT C project.

joa-quim:
I just removed Roger (and Paul) from the 'core' group. And yes, we could have a team for reviews, but how different would it be from 'core'?

seisman:
Actually, we already have a "gmt-maintainers" team (https://github.com/orgs/GenericMappingTools/teams). 

image

joa-quim:
Where, curiously, I am not in. Sure I can add me, but I still think the 'core' is the right group for asking/informing about new PR's.

PR #8619, opened 2024-11-16
Move the annoying warning about the sloppy w/e/s/n one level up (GMT_MSG_INFORMATION)

joa-quim:
I propose to move the warning about the sloppy w/e/s/n one level up (require -Ii).

This warning shows up like 
```
gmtread [WARNING]: (w - x_min) must equal (NX + eps) * x_inc), where NX is an integer and |eps| <= 0.0001.
gmtread [WARNING]: w reset from 5.96701 to 5.93333333333
...
```
and I bet a large majority of users don't understand what we are talking about. On the other side, at least to me, that occurs with codes generated inside a wrapper where user has really no control on the exact w/e/s/n being asked, which will be modified anyway. So, for a bit sanity (perhaps mostly mine) let's move it up one level of verbosity.


Esteban82 (approved):
Sounds fine to me. Blind approval. Let me know if you want me to test it.

joa-quim:
No need to test (I did it).

PR #8625, opened 2024-11-19
Let -J & -F have a -J without scale/size (not used anyway).

### LINKED ISSUE #8624: mapproject -F should not need a -J.../scale
A by product of #8623 

```
C:\v>echo -9 40 | mapproject -Ju29 -C -F
mapproject [WARNING]: No UTM zone given; zone 60T selected
-7036643.48288  71786047.2525

C:\v>echo -9 40 | mapproject -Ju60T -C -F
mapproject [WARNING]: Length <unit> T not supported - revert to default unit [cm]
mapproject [WARNING]: No UTM zone given; zone 60T selected
-7036643.48288  71786047.2525
```
The problem is that it wants also a scale. Like this it works.
```
C:\v>echo -9 40 | mapproject -Ju60T/1:1 -C -F
-7036643.48288  71786047.2525
```

joa-quim:
Fixes #8624


joa-quim:
Right. Before the fisrt two commands would complain and give wrong results.

PR #8627, opened 2024-11-20
Apprently new VS headers (or other Win .h) drop "#define NAN".

joa-quim:
This PR defines it when it«s not found (Windows only)

### DIFF
--- a/src/gmt_common_math.h
+++ b/src/gmt_common_math.h
@@ -65,17 +65,6 @@ extern "C" {
 #	define int64_abs(x) ((int64_t)(((x) >= 0) ? (x) : -(x)))
 #endif
 
-/* Apparently NAN may not be defined anymore in new Visual Studio headers.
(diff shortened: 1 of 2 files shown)


seisman:
NAN is handled in https://github.com/GenericMappingTools/gmt/blob/fa3eda07161f236a039eb68f5791edcb06850316/src/gmt_notposix.h#L335

Perhaps `_Nan._Double` is not a constant?

joa-quim:
I don't know where `_Nan._Double` is coming from, but since my previous fix seems to solve it (the Win CI built fine), I now moved that NAN definition to `gmt_notposix.h`

seisman:
`_Nan` was originally defined in `ymath.h` but was moved to another file in https://github.com/microsoft/STL/pull/2828.

seisman (approved):

joa-quim:
Deeper than `ymath.h`. There I looked, but it includes others, that include others ...

PR #8628, opened 2024-11-21
Improve the patch for NaN with MSVC

seisman:

Not sure what's happening. Changes in PR #8627 passed in that PR, but failed after merging into the master branch (see https://github.com/GenericMappingTools/gmt/actions/runs/11934957233/job/33265224118).

ChatGPT told me:

> In C, a static const variable at global scope must be initialized with a constant expression that can be evaluated at compile time. However:
> 
> (1e+300 * 1e+300) * 0.0F involves floating-point arithmetic, which is not considered a compile-time constant in most C compilers.
>
> HUGE_VAL - HUGE_VAL is also not a compile-time constant for the same reason.

So, this PR improves the definition NAN without any floating-point arithmetic.

With the new patch, the conda-forge building in https://github.com/conda-forge/gmt-feedstock/pull/302 passes.

### DIFF
--- a/src/gmt_notposix.h
+++ b/src/gmt_notposix.h
@@ -331,9 +331,16 @@
 
 #ifndef NAN
 #	ifdef _MSC_VER
-#		define NAN (-(float)(((float)(1e+300 * 1e+300)) * 0.0F));
+# 		ifndef INFINITY
+#			ifndef _HUGE_ENUF
+#				define _HUGE_ENUF 1e+300 // _HUGE_ENUF*_HUGE_ENUF must overflow
+# 			endif
+#			define INFINITY ((float)(_HUGE_ENUF * _HUGE_ENUF))
+# 		endif
+# 		define NAN (-(float)(INFINITY * 0.0F))
 #	else /* _MSC_VER */
-#		define NAN (HUGE_VAL-HUGE_VAL);
+		static const double _NAN = (HUGE_VAL-HUGE_VAL);
+#		define NAN _NAN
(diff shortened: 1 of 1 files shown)


seisman:
> Not sure what's happening. Changes in PR #8627 passed in that PR, but failed after merging into the master branch (see https://github.com/GenericMappingTools/gmt/actions/runs/11934957233/job/33265224118).

it turns our in PR #8627, the CI was still using the old MSVC 19.41.34123.0, while in the master branch and this PR, the CI has upgraded to MSVC 19.42.34433.0.

joa-quim:
OK, let's try this that is a distillation of the NAN definition in a header called `corecrt_math.h`

#ifndef NAN
#	ifndef INFINITY
#		ifndef _HUGE_ENUF
#			define _HUGE_ENUF 1e+300 // _HUGE_ENUF*_HUGE_ENUF must overflow
#		endif
#		define INFINITY ((float)(_HUGE_ENUF * _HUGE_ENUF))
#	endif
#	define NAN (-(float)(INFINITY * 0.0F))
#endif

seisman:
> OK, let's try this that is a distillation of the NAN definition in a header called `corecrt_math.h`

can we just include this header file?

joa-quim:
Sounds complicated. That header is part of `Windows Kits`, which I'm not even sure it belongs to MSVC (probably not).

joa-quim:
Sounds complicated. That header is part of `Windows Kits`, which I'm not even sure it belongs to MSVC (probably not).

PR #8629, opened 2024-11-22
gmtlogo: Create the gmt_letters array dynamically to avoid MSVC compiler error

seisman:
As reported in https://github.com/conda-forge/gmt-feedstock/pull/302#issuecomment-2487473474, building GMT source code with MSVC started to fail with following errors:
```
%SRC_DIR%\src\gmtlogo.c(38): error C2099: initializer is not a constant
%SRC_DIR%\src\gmtlogo.c(131): error C2099: initializer is not a constant
%SRC_DIR%\src\gmtlogo.c(145): error C2099: initializer is not a constant
```
I believe it's caused by behavior changes of new MSVC versions. In PR #8627 and #8628, we tried to improve the handling of `NAN` on Windows to avoid the error, but it seems we don't have a good solution there.

Instead, we can refactor the `gmt_letters` array in `gmtlogo.c` to avoid initializing `gmt_letters` with `NAN`. This is done by splitting the `gmt_letters` array into three separate arrays (for the letter G, M and T, respectively), and then creating the `gmt_letters` array by joining the three arrays with `NAN` records. This PR also reverts the changes in PR #8627

Now the build passes.

### DIFF
--- a/src/gmt_notposix.h
+++ b/src/gmt_notposix.h
@@ -331,8 +331,8 @@
 
 #ifndef NAN
 #	ifdef _MSC_VER
-		static const double _NAN = (-(float)(((float)(1e+300 * 1e+300)) * 0.0F));
-#		define NAN _NAN
+#		include <ymath.h>
+#		define NAN _Nan._Double
 #	else /* _MSC_VER */
 		static const double _NAN = (HUGE_VAL-HUGE_VAL);
 #		define NAN _NAN
--- a/src/gmtlogo.c
+++ b/src/gmtlogo.c
@@ -451,12 +451,12 @@ EXTERN_MSC int GMT_gmtlogo (void *V_API, int mode, void *args) {
 	gmt_letters[idx][1] = NAN;
 	idx += 1;
 	memcpy(gmt_letters + idx, gmt_letters_g, sizeof(gmt_letters_g));
-	idx += sizeof(gmt_letters_g);
+	idx += GMT_N_LETTERS_G;
 	gmt_letters[idx][0] = NAN;
 	gmt_letters[idx][1] = NAN;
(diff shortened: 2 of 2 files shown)


seisman:
The failures on macOS is unrelated and can be ignored.

joa-quim (approved):

joa-quim:
We now have some warnings that I don't fully understand. _If it works_ a simpler solution would have been to do

```
const double d_nan = NAN;
```
and use `d_nan` instead of `NAN`

seisman:
I don't see any related warnings in the CI jobs, on both Linux and Windows. What are exactly the warnings?

joa-quim:
Hmm, I don't see them anymore. It was something about possible overflows.
I also see these for sometime (but they are new) that I don't understand.
```
D:\a\gmt\gmt\src\gmt_init.c(3754): warning C4129: 'g': unrecognized character escape sequence
D:\a\gmt\gmt\src\gmt_init.c(3755): warning C4129: 'g': unrecognized character escape sequence
```

seisman:
> Hmm, I don't see them anymore. It was something about possible overflows. 

Yes, it's caused by typos and has been fixed in https://github.com/GenericMappingTools/gmt/pull/8629/commits/cd21d88e589692e8ef80ea87591e3929df8d2393.

> I also see these for sometime (but they are new) that I don't understand.
> 
> ```
> D:\a\gmt\gmt\src\gmt_init.c(3754): warning C4129: 'g': unrecognized character escape sequence
> D:\a\gmt\gmt\src\gmt_init.c(3755): warning C4129: 'g': unrecognized character escape sequence
> ```

I also see these warnings, but they're unrelated to changes in this PR.

PR #8251, opened 2024-01-01
Final version of updates for 6.5: grdimage, colorbar, grdmix

PaulWessel:
This PR takes over the function of the previous PR (#8241) which now has been absorbed and modified into this cleaner PR.  It also makes further adjustments to the placement and pale dimensions related to the enhanced **-Li**|**I**[_gap_] option in **colorbar**.  I also have added a test for this whose plot looks like this:

interval-panel

You can see the older PR for various new test plots that documents the working of grdmix and grdimage.  I get zero failures with this branch, running with or without OpenMP,  Plesae make sure you change to theimg-layout branch, select static as active server (`export GMT_DATA_SERVER=static`) and run all tests.  If you find any issues under Windows or Linux please let me know.

The **colorbar** docs now states this:

L

I believe this is the last item to be merged into master before we release 6.5.

PR #8241, opened 2023-12-28
Final grdimage upgrade related to transparency

PaulWessel:
This branch replaces the deprecated branch #8191 which added some complications.  This branch still passes 100% of the tests.  I will now work on adding tests that highlight new capabilities (like variable transparency via square polygons) while making sure all tests still pass.

PR #8643, opened 2024-12-15
Remove the incompatibility (regression) between the -I and -Q options.

joa-quim:
See forum post
 https://forum.generic-mapping-tools.org/t/grdimage-transparency-of-nan-values/5641/11

But testing this further would be good. There must have been some reason to exclude the -I -Q combination.


seisman:
These codes were initially added in https://github.com/GenericMappingTools/gmt/pull/8241 and #8251

joa-quim:
Yes, I guessed so but unless we find a good reason to keep it (the -Q -I conflict) we should remove it. And even if we find, we may restrict that eventual case under an appropriate code condition.

seisman (approved):
No new failures. So good to merge.

Issue #8655, opened 2025-01-02
Tif won't convert to grd without complaining about xres/yres


cbworden-usgs:
I download a file from ScienceBase in .tif format, and grdinfo is happy with it, but as soon as I do anything with the file that turns it into a .grd (or .nc), grdinfo (and other GMT processes) throw an error: "grdinfo [ERROR]: (x_max-x_min) must equal (NX + eps) * x_inc), where NX is an integer and |eps| <= 0.0001. [etc.]". The file comes from this page on SB: https://www.sciencebase.gov/catalog/item/5a5fa029e4b06e28e9bfc43a

To get the .tif file do:
```
curl https://www.sciencebase.gov/catalog/file/get/5a5fa029e4b06e28e9bfc43a\?f=__disk__2b%2F4b%2Fac%2F2b4bacfcbe272e38b35637f7b9633401e1c2f191 > ca_map.tif
```
You may need to remove the "\\" before the "?" depending on your shell.

Now:
```
gmt grdinfo ca_map.tif
```
is happy, and the numbers look good. But if we do:
```
gmt grdconvert ca_map.tif -Gca_map.grd
```
Now:
```
gmt grdinfo ca_map.grd
```
Gives:
```
grdinfo [ERROR]: (x_max-x_min) must equal (NX + eps) * x_inc), where NX is an integer and |eps| <= 0.0001.
grdinfo [ERROR]: (y_max-y_min) must equal (NY + eps) * y_inc), where NY is an integer and |eps| <= 0.0001.
grdinfo (gmtlib_read_grd_info): Use grdedit -A on your grid file to make region and increments compatible [ca_map.nc]
[...]
```
Despite the numbers still looking good. Following the instructions in the error message:
```
gmt grdedit -A ca_map.grd -Gca_map_fixed.grd
```
Results in the same errors when running
```
gmt grdinfo ca_map_fixed.grd
```

```
curl https://www.sciencebase.gov/catalog/file/get/5a5fa029e4b06e28e9bfc43a\?f=__disk__2b%2F4b%2Fac%2F2b4bacfcbe272e38b35637f7b9633401e1c2f191 > ca_map.tif

gmt grdconvert ca_map.tif -Gca_map.grd

gmt grdedit -A ca_map.grd -Gca_map_fixed.grd

gmt grdinfo ca_map_fixed.grd
```

```
[...]

PR #8660, opened 2025-01-03
Do not apply the gmtgrdio_doctor_geo_increments() small modification if it breaks.

joa-quim:
See issue #8655. The issue is that `gmtgrdio_doctor_geo_increments()` may change the increments in such a way that a grid with a previous good range as a multiple of inc becomes bad (because only the incs were changed). This PR checks that the _doctor_ works does not break the condition. If it does, do not apply it.


seisman:
I think you should ping @GenericMappingTools/gmt-maintainers instead of the core team instead.

Issue #8606, opened 2024-10-28
Can't plot data points out of the [-180, 180, -90, 90] range on geographic maps for external wrappers


seisman:
See https://forum.generic-mapping-tools.org/t/plotting-points-out-of-90-degree-range/5479 for the initial issue report. 

To reproduce the issue:
```
import pygmt
fig = pygmt.Figure()
fig.basemap(region="d", projection="Q10i", frame="afg")
fig.coast(land="#666666", water="skyblue")
fig.plot(x=60, y=110, style="c0.3c", fill="red", pen="black", no_clip=True)
fig.show()
```

The error `plot [WARNING]: Latitude (110) at line # 1 exceeds -|+ 90! - set to NaN` comes from the `gmtlib_process_binary_input` function. https://github.com/GenericMappingTools/gmt/blob/65e0917c21ca5b159f32a87358468d3fe987cb26/src/gmt_io.c#L4839

The equivalent CLI version works as expected. The reason is that, for GMT CLI, `gmtio_ascii_input` is called to read the data points, but for external wrappers, data are passed in binary buffers, so the `gmtio_bin_input` is called instead. 

I guess the same issue exists for GMT.jl, so it should be declared as a GMT bug.

joa-quim:
Yep. This works
```
C:\v>echo 60 110 | gmt plot -Rd -JQ10i -Baf -BWSen -N -Gred -Sc7p -png lixo
```
but this, not
```
Julia> plot([60 110], region="d", projection="Q10i", marker="c0.3c", fill="red", no_clip=true, Vd=1)
        psxy  -Rd -JQ10i -Baf -BWSen -N -Gred -Sc7p -P -K > c:\TMP/GMTjl_j.ps
psxy [WARNING]: Latitude (110) at line # 1 exceeds -|+ 90! - set to NaN
```

joa-quim:
This not an externals issue only. If we use binary input, it does not complain but no plot symbol either
```
echo 60 110 > lixo.dat
gmtconvert lixo.dat -bo > lixo.bin
gmt plot lixo.bin -Rd -JQ10 -Baf -BWSen -N -Gred -Sc7p -png lixo   # No complain and no red point plotted
```

joa-quim:
And is restricted to geogs. Cartesians work
```
gmt psxy lixo.bin -bi -R-180/180/-90/90 -JX10 -Baf -BWSen -N -Gred -Sc7p -png lixo
```

[...]

PR #8663, opened 2025-01-03
Let psxy -N be usable with binary input.

joa-quim:
Fix issue #8606

But solution isn't perfect. For longitudes the 360-wrapping code does not let to plot outside and there is nothing we can do about it without risking to break the general longitudes wrapping.

```
echo 190 110 | gmtconvert -bo | gmt plot -bi -Rd -JQ10i -Baf -BWSen -N -Gred -Sc12p -png lixo
```
lixo

### DIFF
--- a/src/psxy.c
+++ b/src/psxy.c
@@ -1403,6 +1403,8 @@ EXTERN_MSC int GMT_psxy (void *V_API, int mode, void *args) {
 		Return (GMT_RUNTIME_ERROR);
 	GMT_Report (API, GMT_MSG_DEBUG, "Operation will require %d input columns [n_cols_start = %d]\n", n_needed, n_cols_start);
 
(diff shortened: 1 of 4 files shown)


src/psxy.c
```diff
@@ -1403,6 +1403,8 @@ EXTERN_MSC int GMT_psxy (void *V_API, int mode, void *args) {
 		Return (GMT_RUNTIME_ERROR);
 	GMT_Report (API, GMT_MSG_DEBUG, "Operation will require %d input columns [n_cols_start = %d]\n", n_needed, n_cols_start);
 
+	if (Ctrl->N.active && GMT->common.b.active[GMT_IN]) GMT->common.b.active[2] = true; /* To signal gmtlib_process_binary_input that lats may be of |90| */
```
seisman, line 1406: Do we make the same change to plot3d?

remkos (approved):
The solution looks ok by my. As @seisman said: please also implement in `psxyz.c`

joa-quim:
Yeap, thought on that too but forgot meanwhile. Done now.

PR #8665, opened 2025-01-04
Slightly increase the RMS value for test/grdvector/shrink.sh

seisman:
The test `test/grdvector/shrink.sh` has been failing for a long time.

Here is the diff image. As you can see, the only difference is in the arrow of the 3rd row, east hemisphere (near (160, -10)). This difference is most likely caused by numerical round-off errors, so I think the test can be safely ignored. This PR increases the RMS value to make the test pass.

shrink_diff

### DIFF
--- a/test/grdvector/shrink.sh
+++ b/test/grdvector/shrink.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 # Test shrinking of line and heads with/without +n for grdvector
+# GRAPHICSMAGICK_RMS = 0.0042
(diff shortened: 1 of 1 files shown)


remkos (approved):

joa-quim:
> This difference is most likely caused by numerical round-off errors, ...

I don't believer this round-off theory (round-offs occur at 1e-16 bits) but don't have strong enough motivation to dig-up this behavior. So, good to me too.

joa-quim (approved):

Issue #8645, opened 2024-12-25
plot/plot3d: Simplify the documentation for the -A option


seisman:
https://docs.generic-mapping-tools.org/latest/plot.html#a

The `-A` option has the following syntax:
```
-A[m|p|x|y|r|t]
```
in which `m|p` works for geographic map, `x|y` works for Cartensian plot, and `r|t` works for polar projections.

But in the source code, `m`/`y`/`r` works the same way, and `p`/`x`/`t` works the same way.
https://github.com/GenericMappingTools/gmt/blob/b34b125857afedea82fb5b4c2589298df267c774/src/psxy.c#L838-L848

So, instead of showing the `-A[m|p|x|y|r|t]`, we can simplify it to:
```
-A[x|y]
```
then the documentation can be something like:

> **x**: Draw the line along *x* first then along *y*.
> **y**: Draw the line along *y* first then along *x*.
> 
> Here, *x* and *y* have different meanings in different coordinate systems.
> - For Cartesian data, *x* and *y* means X- and Y-axis, respectively
> - For polar projection, *x* means *theta* and *y* means *r*
> - For geographic projections, *x* means *parallel* and *y* means *meridian*

I'd like to work on it if approved.

joa-quim:
Yes, it will look better that way.

seisman:
I'm updating the documentation of this option on the PyGMT side first, then @yvonnefroehlich found an issue at https://github.com/GenericMappingTools/pygmt/pull/3720#discussion_r1899816545, which I think is a GMT bug. Here is the CLI version just in case you want to try it:

```
gmt begin map

gmt basemap -JH10c -R0/360/-90/90 -Bafg -B+t"-Am"
echo 90 -50 | gmt plot -Sc0.2c -Gred 
gmt plot -W1p -Am << EOF
90  -50
150 50
EOF

gmt basemap -JH10c -R0/360/-90/90 -Bafg -B+t"-Ap" -Xw+1c
echo 90 -50 | gmt plot -Sc0.2c -Gred 
gmt plot -W1p -Ap << EOF
90  -50
150 50
EOF
gmt end show
```

Here are the related codes:
https://github.com/GenericMappingTools/gmt/blob/b877d7288d23766014d5fb82e876fe5039dea9b5/src/gmt_vector.c#L1607-L1653

[...]

PR #8674, opened 2025-01-07
plot/plot3d/clip: Simplify the -A syntax to -Ax|y

seisman:
See https://github.com/GenericMappingTools/gmt/issues/8645 for context.

Closes https://github.com/GenericMappingTools/gmt/issues/8645.


remkos (approved):
I felt I should review this since I implemented -Am and -Ap in GMT4 in Sep 2007!
Since then it was extended with -A[x|y|r|t], without it actually requiring a change of code except for needing to parse the options and add to the documentation.

So it is OK to me to bring it back to -A[x|y] as long as the old options are still supported for backward compatibility.

How does this impact the long options?

seisman:
> I felt I should review this since I implemented -Am and -Ap in GMT4 in Sep 2007!

Good to know the history dating back to 2007!

When updating the documentation, we also found a bug that the meaning of `-Am` and `-Ap` are reversed. Please see https://github.com/GenericMappingTools/gmt/issues/8645#issuecomment-2566198810 for details. We already fixed it in #8648, but it would be great if you can verify it again.

> So it is OK to me to bring it back to -A[x|y] as long as the old options are still supported for backward compatibility.

Yes, it's backward compatible.

> How does this impact the long options?

That's a good question. Currently, long names exist for m/p/x/y/r/t (https://github.com/GenericMappingTools/gmt/blob/605da462cbc15e3db40cc86e1e3b863b61444ca8/src/longopt/psxy_inc.h#L29). I guess we should also remove the long names for m/p/r/t, OK?

joa-quim:
Also -E longoptions got it all wrong GMT.jl has `error_bars`, PyGMT `error_bar` but GMT picked `errorbars`. Add all three aliases in master.

seisman:
Not yet. There are other modules (e.g., `clip`) that takes the same `-A` option.

seisman:
I believe this PR is now ready for final review.

joa-quim (approved):
Looks good.

PR #8676, opened 2025-01-15
Update grdvs30.rst

CovMat:
cratons.xy does not exist in GMT server now

### DIFF
--- a/doc/rst/source/supplements/seis/grdvs30.rst
+++ b/doc/rst/source/supplements/seis/grdvs30.rst
@@ -45,7 +45,8 @@ Required Arguments
     Argument *val* can be one of these three:
        - A value *val* between 0 and 1, where 0 means a stable Craton and 1 an Active region.
        - The name of a multi-segment file with the *cratons* polygons. In this case the polygons will be
-         feed to grdmask to compute a cratons/active tectonic mask. 
+         feed to grdmask to compute a cratons/active tectonic mask. Use **-C**\ *@cratons.xy* to download
+         a cratons file from the GMT server.
        - The name of a grid with the cratons/active tectonic regions. In this case the **+g** suffix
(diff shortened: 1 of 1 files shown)


CovMat:
and usage and code of grdvs30.c should be rewrited:

https://github.com/GenericMappingTools/gmt/blob/master/src/seis/grdvs30.c#L183-L197

now the code cannot distinguish value and filename. If you give -C1, it will see 1 as a filename

joa-quim:
Yes, good intentions (saying `-C@cratons.xy` is possible), but I rather keep it in docs and this serving a reminder that we must add that file to the server. So, can you please edit your PR to keep it?

Regarding the 0 _vs_ 1 for `cratons` vs `active` the source is contraditory. The original code starts to say in comments:

https://github.com/usgs/earthquake-global_vs30/blob/master/src/grad2vs30.c#L20-L21

but later the code say the oposite:
https://github.com/usgs/earthquake-global_vs30/blob/master/src/grad2vs30.c#L189-199

So I decided to trust more the information in the code and thus the first part of this PR is correct.

CovMat:
> Yes, good intentions (saying `-C@cratons.xy` is possible), but I rather keep it in docs and this serving a reminder that we must add that file to the server. So, can you please edit your PR to keep it?
> 
> Regarding the 0 _vs_ 1 for `cratons` vs `active` the source is contraditory. The original code starts to say in comments:
> 
> https://github.com/usgs/earthquake-global_vs30/blob/master/src/grad2vs30.c#L20-L21
> 
> but later the code says the oposite: https://github.com/usgs/earthquake-global_vs30/blob/master/src/grad2vs30.c#L189-199
> 
> So I decided to trust more the information in the code and thus the first part of this PR is correct.

@joa-quim  thanks. I think  It can be merged

joa-quim (approved):

PR #8694, opened 2025-02-18
Update grdimage required arguments

Esteban82:
The `grdimage` module does not require the `-J` argument. 

Modify the file so that `-J` appears among the optional arguments when one types `gmt grdimage` in the terminal.
I also remove `J` in `#define THIS_MODULE_NEEDS`. 

On my pc it works fine. But I'm not sure I did it right. 

Then I will modify the rst file.

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -33,7 +33,7 @@
 #define THIS_MODULE_LIB			"core"
 #define THIS_MODULE_PURPOSE		"Project and plot grids or images"
 #define THIS_MODULE_KEYS		"<G{+,CC(,IG(,>X},>IA,<ID"
-#define THIS_MODULE_NEEDS		"Jg"
+#define THIS_MODULE_NEEDS		"g"
 #define THIS_MODULE_OPTIONS		"->BJKOPRUVXYfnpt" GMT_OPT("Sc") GMT_ADD_x_OPT
 
 /* These are images that GDAL knows how to read for us. */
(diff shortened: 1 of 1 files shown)


remkos (approved):
Looks good to me.

src/grdimage.c
```diff
@@ -33,7 +33,7 @@
 #define THIS_MODULE_LIB			"core"
 #define THIS_MODULE_PURPOSE		"Project and plot grids or images"
 #define THIS_MODULE_KEYS		"<G{+,CC(,IG(,>X},>IA,<ID"
-#define THIS_MODULE_NEEDS		"Jg"
+#define THIS_MODULE_NEEDS		"g"
```
seisman, line 36: I don't think you can remove `J` from `THIS_MODULE_NEEDS`. There are some explanations at https://github.com/GenericMappingTools/pygmt/issues/1493#issuecomment-921267870.

Esteban82: Ok, thanks @seisman. I will restore J.

joa-quim:
This is not right. At #L223 it's inserting the -J in the middle of -I description.

Esteban82:
> This is not right. At #L223 it's inserting the -J in the middle of -I description.

Ok. I see. Should it be in Line 231 (just before `GMT_Option (API, "K");`)?

joa-quim:
Yep, alphabetic order.

PR #8699, opened 2025-03-10
Add a new option to grdfft and gravfft to let change the MGAL_AT_45 value

joa-quim:
See this Forum post.
https://forum.generic-mapping-tools.org/t/gravitational-admittance-estimate-for-mars/5776/4

### DIFF
--- a/doc/rst/source/grdfft.rst
+++ b/doc/rst/source/grdfft.rst
@@ -20,6 +20,7 @@ Synopsis
 [ |-E|\ [**r**\|\ **x**\|\ **y**][**+n**][**+w**\ [**k**]] ]
 [ |-F|\ [**r**\|\ **x**\|\ **y**]\ *params* ]
 [ |-I|\ [*scale*\|\ **g**] ]
+[ |-M|\ *mgal_at_45* ]
 [ |-N|\ *params* ]
 [ |-Q|\ ]
 [ |-S|\ *scale*\|\ **d** ]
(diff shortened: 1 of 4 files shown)


Esteban82:
How could I test this? I try some examples from gravfft but I think I did something wrong.

joa-quim:
Honestly, I'm not sure how to test it with new numbers, but what I tested is that it doesn't break. We have quite a few tests for `gravfft` and all passed with this PR.

Regarding `grdfft` we actually do not have any tests this parameter (options -D and -I), but this one gives exactly the same thing before and after this PR
```
grdfft @geoid_1m.nc -Dg -V -Ggrav.nc
``` 

so I would say we re not breaking anything.

Esteban82:
I think that there is something missing.

```
gmt grdfft @geoid_1m.nc -Dg -V -Ggrav1.nc -M980619.9203
grdfft [ERROR]: Unrecognized option -M
```

And I think that I build the branch well because I can see the description of `-M`

```
gmt grdfft
...
  -M<mgal_at_45>
     Value for gravity in mGal at 45 degrees latitude. Default is 980619.9203 mGal (Moritz's 1980 IGF value).
...
```

joa-quim:
Yes, just submitted a commit to fix that. The thing is that `grdfft` is more convoluted in usage of this parameter and I had to parse the -M separately, but missed to make it be ignored on second parsing passage. Now all of these three give the same
```
grdfft @geoid_1m.nc -Dg -Ggrav.nc
build\src\grdfft @geoid_1m.nc -Dg -Ggrav.nc
build\src\grdfft @geoid_1m.nc -Dg -Ggrav.nc -M980619.9203
```

Esteban82:
Yes, now it works for me too.
I try these:

```
gmt grdfft @geoid_1m.nc -Dg -V -Ggrav0.nc
gmt grdfft @geoid_1m.nc -Dg -V -Ggrav1.nc -M980619.9203
gmt grdfft @geoid_1m.nc -Dg -V -Ggrav2.nc -M372635.5697
```

For the first two I got the same grid. For the last one a similar one but with different values.
1_map
2_map

I don't understand much about this subject, but it seems fine to me. I will approve it.

PR #8737, opened 2025-05-29
Fix option -B+n in psscale

joa-quim:
This addresses one of the points raised in the Forum issue https://forum.generic-mapping-tools.org/t/colour-scheme-without-outline/5967

Now we can do `psscale -Dx6.5/1+w340p/36p+h -B+n -Ccyclic > cyclic.ps` and have no frames at all.

I would like also that `-B0` did that work (a nicer syntax, IMO) and in fact, this PR also implements it but it requires that GMT is built with `-DB0_IS_NO_FRAME`. The reason for wrapping it in a `#ifdef` is that changing `-B0` to do what it originally should have been since the beginning (that is, to DO NOT PLOT ANY FRAME, ANNOTS AT ALL) is a breaking change. I would like, however, to discuss if we should do it nevertheless.


Esteban82 (approved):
Works for me.

Esteban82:
I didn't see the comment on the discussion. 
Would this cause would break backwards compatibility? I believe I have used -B+n for maps and had no frame at all.

joa-quim:
Ok, I'm going to merge this and leave discussion to #8740

PR #8745, opened 2025-06-08
WIP: Let us have the Spilhaus projection accessible from pscoast.c

joa-quim:
For this, it was necessary to add the -g option to pscoast and make a gap detection in `gmt_geo_to_xy_line()`.

We can not yet do the interrupted graticule nor plot a frame.


joa-quim:
Guys, we have a working version ready for testing/approval. Not perfect, as painting land/Ocean still screws (but Spilhaus is damn complex to deal with) but this works fine. Don't know also why the annotations fall where they fall but not so bad either.

```
gmt coast -Rd -J+proj=spilhaus+width=10 -W0.1 -Ba30g15 -png lixo
```


lixo

PR #8748, opened 2025-06-13
Update makecpt.rst -H

Esteban82:
I understand that when the `-H` option is used, the CPT is only written to standard output, not to a file. So the documentation should say "instead" rather than "as well".

Here is the relevant line in the source code for reference: https://github.com/GenericMappingTools/gmt/blob/b244cfd57683cf096c03b193ed83c51a48bbf5a8/src/makecpt.c#L668

Please approve this change only if my understanding is correct.

### DIFF
--- a/doc/rst/source/makecpt.rst
+++ b/doc/rst/source/makecpt.rst
@@ -123,7 +123,7 @@ Optional Arguments
 .. _-H:
 
 **-H**\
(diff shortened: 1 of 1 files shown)


joa-quim:
> only written to standard output, not to a file

I'm a bit confused with this. Writing to stdout on modern mode is the same as writing to a file (only that file name is is determined by modern mode machinery).

Issue #8740, opened 2025-05-30
What to do with the -B0 option


joa-quim:
This is a quick start on a discussion on -B0 initiated at https://github.com/GenericMappingTools/gmt/pull/8737 and to not let it die buried in an closed issue.

Another option for -B0 could be "Do not plot frame axes" but let it plot annotations, ticks, grids, etc. I think we do not have that possibility except if we play with `@100` transparencies. And even then, `psscale` uncovered a bug in it.

remkos:
I thought `-B0` used to mean "no frame and no annotations at all" in all other applications, like `psbasemap`. It would overrule the default `WESN`, plotting at least a frame. But I do not see it in the GMT4 manual page (https://docs.generic-mapping-tools.org/4/gmt/html/man/psbasemap.html), so it was likely an undocumented feature.

Nonetheless, it was actually an undocumented feature that I tried to employ myself with `psscale` recently.

But still, the syntax *does* show up in this current GMT6 example using `(ps)coast`: https://docs.generic-mapping-tools.org/latest/gallery/ex44.html#example-44

So I would like to call for a consistent use of it, i.e.,
* Support `-B+n` as documented, and use that in the examples, replacing `-B0`
* For (undocumented) backward compatibility, allow `-B0` to be identical to `-B+n`, also in `psscale`.

anbj:
If you do a **grep** in the doc directory, it gives 28 matches. Not all matches related to this discussion probably, but it shows that it's being used quite a bit.

```
$ grep -r '\-B0\s' | wc -l
28
```

remkos:
Thanks, @anbj, for checking that more thoroughly than I did.
Shall we then conclude to proceed as I suggested above? That is, to replace all the `-B0` with `-B+n` in the scripts and also support `-B0` in `psscale` for the old dudes that have `-B0` engraved in their brains.

joa-quim:
We cannot replace `-B0` by `-B+n`. First they don't do the same thing (though we make them do). `-B0` always draws frame whilst `-B+n` does not. The equivalent of `-B0` is `-Blbrt` but even that we cannot replace, because

```
gmt basemap -R0/1.5/0/1.7 -Jx3 -B0 -B+glightyellow -png lixo1
```

![Image](https://github.com/user-attachments/assets/32dabd86-a103-4741-9419-067019f77bb4)

but if we replace `-B0` by `-Blrbt` (or even `-B+n`)

```
[...]

PR #8749, opened 2025-06-16
Let -B00[...] and -B000<...> do plot a zero width frame or no frame at all (but annots, tics, grids, yes)

joa-quim:
Addresses discussion in #8740


seisman:
Need to document it, otherwise no one will know it.

joa-quim:
I had, but for some reason the commit this not show up.

PR #8776, opened 2025-08-13
Set the -J optional

Esteban82:
This PR updates the documentation to indicate that the -J option is optional.


Esteban82:
Is there someone that would like to revise this?

PR #8793, opened 2025-09-12
gmtlogo/psbasemap: Ensure the long options for -D are consistent with other modules

seisman:
Some GMT modules like `image`/`logo`/`legend`/`colorbar` can add embellishments. These modules have similar syntax for placing these embellishments. For example, `logo`'s `-D` option has the following syntax:
```
-D[g|j|J|n|x]refpoint[+hheight|+wwidth][+jjustify][+odx[/dy]]
```
in which `[g|j|J|n|x]refpoint[+jjustify][+odx[/dy]]` are common to all embellishments and determine the location of embellishments on maps. 

The long option names are `g|j|J|n|x` for `mapcoords,inside,outside,boxcoords,plotcoords`, `janchor` for `+j`, and `anchoroffset` for `+o`. 

This PR fixes the inconsistences in a few modules.

### DIFF
--- a/src/longopt/gmtlogo_inc.h
+++ b/src/longopt/gmtlogo_inc.h
@@ -27,7 +27,7 @@ static struct GMT_KEYWORD_DICTIONARY module_kw[] = {
 		  transproc_mask */
 	{ 0, 'D', "position",
 	          "g,j,J,n,x",         "mapcoords,inside,outside,boxcoords,plotcoords",
-	          "w,h,j,o",           "width,height,janchor,offset",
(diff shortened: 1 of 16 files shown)


seisman:
I've renamed `anchoroffset` to `offset`, but for **+j**, I think `anchor` makes more sense than `justify`, since it sets the justification anchor point.

joa-quim (approved):
OK, I'm not fully convinced about **+j** (after all the synopsis syntax says **+j**justify and not **+j**anchor) but won't insist on this.

PR #8798, opened 2025-09-21
Check if file passed via -G+l has at least one numeric column as it must

joa-quim:
Update docs to make this more clear.


joa-quim:
No idea why the tests are failing before even starting. Complaining about the cache?

PR #8799, opened 2025-09-22
CI: Fix the dvc issue with an access token

seisman:
The same thing happened in the PyGMT repo (https://github.com/GenericMappingTools/pygmt/pull/4113, https://github.com/GenericMappingTools/pygmt/pull/4078).


joa-quim (approved):
Good work.

Issue #8775, opened 2025-08-08
Issues with transparency with psevents


Esteban82:
I created an earthquake animation (similar to Animation 8) and wanted the events to appear with transparency. I noticed a few issues: currently, transparency is only supported during the coda phase, which only exists for events with finite duration.

What confuses me is that transparency behaves differently from other properties, such as size or color (see figure below). I’m wondering if there’s a deeper reason why transparency was implemented this way and I’m just not seeing it.

<img width="1847" height="648" alt="Image" src="https://github.com/user-attachments/assets/25324643-1608-4e67-bce4-9f2e34eda1fa" />

I think I could make the modifications in the source file, but I might need some help.

### Issues observed

1. The -t option is not available for events.
2. Transparency is hard-wired to 0 in the plateau, decay, and normal phases. See lines [839](https://github.com/GenericMappingTools/gmt/blob/9e4631936d831001a6a5459541e6966940c992b4/src/psevents.c#L839), [846](https://github.com/GenericMappingTools/gmt/blob/9e4631936d831001a6a5459541e6966940c992b4/src/psevents.c#L846), and [851](https://github.com/GenericMappingTools/gmt/blob/9e4631936d831001a6a5459541e6966940c992b4/src/psevents.c#L851).
3. Two transparency values can be set with -`Mtval1[+cval2]`. It seems that the first value (val1) is only used in the previous phase (with a transition in the rise phase). For symbols, this makes little sense since their size is set to 0 at that point—perhaps it’s more relevant for text?
4. Transparency works fine in the fade and coda phases.

### Help needed

I believe I can implement the required changes, but right now I would appreciate:

1. Any comments on how transparency is supposed to work—how Paul originally intended it.
2. Confirmation that I can safely use the undocumented `psevents -/ `options to get the event functions at t = 0. See [line 629](https://github.com/GenericMappingTools/gmt/blob/9e4631936d831001a6a5459541e6966940c992b4/src/psevents.c#L629).

remkos:
Just for info. What you need here is indeed the *-t* option. 
The way it is implemented (I did quite a bit of the work at the time) is that it adds command to the Postscript output that is interpreted by Ghostscript in the conversion to PDF (or raster formats) to have the next layer be transparent. (This does actually not work for Postscript (printers) but works great with Ghostscript, as you can see.)

PR #8803, opened 2025-10-01
Add the -t option for psevents

Esteban82:
In this option I add the `-t` option for psevents module. This PR refers to the first issue in #8775.

I try it with this commands:
```
gmt makecpt -Cred,green,blue -T0,70,300,10000 > q.cpt
gmt events -Rg -JG200/5/6i -Baf @quakes_2018.txt -SE- -Cq.cpt  --TIME_UNIT=d -T2018-06-01T  -t50 -png prueba
```
I got this figure
prueba

If it is ok, later I will modify the docs.


Esteban82:
Thanks Remko. I've just add the explanation in the docs.

PR #8804, opened 2025-10-03
fix a bug in coupe ( -A+wwidth is useless when using -N  )

CovMat:
https://github.com/GenericMappingTools/gmt/blob/58850df4f841da24240147432e28b5322cfb1ccc/src/seis/pscoupe.c#L1331-L1333

if `-N`  is used, `!Ctrl->N.active` is always 0, so the whole condition is always 0 too. 

However, if user wants to use **-A+w***width* to choose data points only in the range of the 

> width in km of the cross-section on each side of a vertical plane or above and under an oblique plane 
> https://docs.generic-mapping-tools.org/6.6/supplements/seis/coupe.html#a

, although the data points outside the range let `pscoupe_dans_coupe` returns 0 and `!pscoupe_dans_coupe` equal to 1:

https://github.com/GenericMappingTools/gmt/blob/58850df4f841da24240147432e28b5322cfb1ccc/src/seis/pscoupe.c#L353-L371

these data points will not be skipped because `!Ctrl->N.active` is always 0. 

Actually, in docs: https://docs.generic-mapping-tools.org/6.6/supplements/seis/coupe.html#n , **-N** means 

> Does not skip symbols that fall outside map border [Default plots points inside border only].

it is well done in lines:
https://github.com/GenericMappingTools/gmt/blob/58850df4f841da24240147432e28b5322cfb1ccc/src/seis/pscoupe.c#L1259
https://github.com/GenericMappingTools/gmt/blob/58850df4f841da24240147432e28b5322cfb1ccc/src/seis/pscoupe.c#L1265
https://github.com/GenericMappingTools/gmt/blob/58850df4f841da24240147432e28b5322cfb1ccc/src/seis/pscoupe.c#L1338-L1341
https://github.com/GenericMappingTools/gmt/blob/58850df4f841da24240147432e28b5322cfb1ccc/src/seis/pscoupe.c#L1684

It makes no sense to use `!Ctrl->N.active` in 
```
if (!pscoupe_dans_coupe (in[GMT_X], in[GMT_Y], depth, Ctrl->A.xlonref, Ctrl->A.ylatref, Ctrl->A.fuseau, Ctrl->A.PREF.str,
     Ctrl->A.PREF.dip, Ctrl->A.p_length, Ctrl->A.p_width, &distance, &n_dep) && !Ctrl->N.active)
     continue;
```

### DIFF
--- a/src/seis/pscoupe.c
+++ b/src/seis/pscoupe.c
@@ -1329,7 +1329,7 @@ EXTERN_MSC int GMT_pscoupe (void *V_API, int mode, void *args) {
 				depth = in[GMT_Z];
 
 				if (!pscoupe_dans_coupe (in[GMT_X], in[GMT_Y], depth, Ctrl->A.xlonref, Ctrl->A.ylatref, Ctrl->A.fuseau, Ctrl->A.PREF.str,
-					Ctrl->A.PREF.dip, Ctrl->A.p_length, Ctrl->A.p_width, &distance, &n_dep) && !Ctrl->N.active)
+					Ctrl->A.PREF.dip, Ctrl->A.p_length, Ctrl->A.p_width, &distance, &n_dep) )
 					continue;
 
 				xy[GMT_X] = distance;


CovMat:
If we choose to keep `&& !Ctrl->N.active` in 
https://github.com/GenericMappingTools/gmt/blob/58850df4f841da24240147432e28b5322cfb1ccc/src/seis/pscoupe.c#L1331-L1333

The description of `-N` in https://docs.generic-mapping-tools.org/6.6/supplements/seis/coupe.html#n should be modified as:
> Does not skip symbols that fall outside map border and **outside the cross-section** [Default plots points inside border and **inside the cross-section**].

CovMat:
Here is a example shows that how `-N` influences `-A+wwidth`

data: [nichong.txt](https://github.com/user-attachments/files/22672824/nichong.txt)

```
gmt begin figure png
gmt basemap -BWS -Byaf+l"Focal depth (km)" -Bxaf+l"Distance (km)" -JX15c/-5c -R77.9/79.5/40.70/41.75
gmt coupe nichong.txt -Sa0.9c -Aa78.43/41.26/78.70/41.05+d90+r+w1 -W0.4p -GLIGHTSEAGREEN -Q -N
gmt end
```

With `-N`, all points are plotted, `+w1` is ignored:
figure_2

If remove `-N`, we get:
figure_1

only points within width=1km are chosen

Esteban82:
Hi @CovMat. I'm not quite sure what behavior you consider to be a bug. Could you tell me specifically what problem you want to solve?

CovMat:
> Hi @CovMat. I'm not quite sure what behavior you consider to be a bug. Could you tell me specifically what problem you want to solve?

@Esteban82, sorry I updated the example in above comment to explain this bug. Thank you for reading it

joa-quim (approved):
It took me  a little while to fully understand this issue. We certainly do not want to change the **-N**  meaning/description, which is transversal to several GMT modules and means simply - _clip by map borders_.

PR #8844, opened 2025-12-21
Add the so wished option to plot spheres in psxyz.

joa-quim:
They aren't really spheres but circles in 3D plots.

This is mostly a Claude.ai creation but a one that needed a lot of baby-sitting development.

Claude picked -SQ for this option and I honestly couldn't think of a better alternative, though -Sq is a decorated line and -SQ has nothing to do with it. Ideas for a better syntax are welcome. The following is the summary prepared by Claude

```

● Summary of -SQ Sphere Symbol Option

  Basic Syntax:
  -SQ<size>[c|i|p][+a<azimuth>][+e<elevation>][+f][+n]

  Description:
  The -SQ option plots a 3-D sphere symbol with diameter <size>. The sphere is rendered with a radial gradient shading from white at the light source position to the fill color (specified by -G) at the opposite side, creating a realistic 3-D appearance.

  Size Units:
  - Default unit: cm (centimeters)
  - Supported units: c (cm), i (inches), p (points)

  Modifiers:
  - +a - Set light source azimuth angle [default: 0°, from the right]
  - +e - Set light source elevation angle [default: 90°, perpendicular to viewing plane]
  - +f - Use flat/constant fill color (disables gradient shading)
  - +n - Draw outline only with no fill

  Fill and Outline:
  - Fill color: Specified by -G [default: black if not specified]
  - Outline: Controlled by -W (pen color, width, style)
  - Without -G and without +n, the sphere defaults to black fill
```

The **+a** and **+e** are not working correctly and I'm not even sure that we should have them. Opinions?

A low imagination example (used in the learning work) is:

```
echo 3 3 3 | gmt plot3d -R0/6/0/6/0/6 -JX10c -JZ4c -p135/45 -SQ2c -W0.2p,red -Ba -Bz -png lixo
```

lixo

### DIFF
--- a/src/psxyz.c
+++ b/src/psxyz.c
@@ -1905,8 +1905,16 @@ EXTERN_MSC int GMT_psxyz (void *V_API, int mode, void *args) {
 						break;
 					case PSL_SPHERE:	/* Case created by Claude.ai */
 						gmt_plane_perspective(GMT, GMT_Z, data[i].z);
-						if (S.SP_light_set)	/* Output custom light position for sphere */
-							PSL_command(PSL, "/SP_lx %.12g def /SP_ly %.12g def\n", S.SP_lx, S.SP_ly);
+						if (S.SP_light_set) {	/* Calculate and output custom light position for sphere */
+							/* Simple model: azimuth controls horizontal, elevation controls vertical */
+							/* Relative to viewing direction */
+							double dazim = GMT->current.proj.z_project.view_azimuth - S.SP_light_az;
+							if (dazim > 90) dazim = 90.0;		/* Do let illum from the hidden hemisphere */
+							if (dazim < -90) dazim = -90.0;
+							double SP_lx_proj = sind(dazim);
+							double SP_ly_proj = sind(S.SP_light_el - GMT->current.proj.z_project.view_elevation);
+							PSL_command(PSL, "/SP_lx %.12g def /SP_ly %.12g def\n", SP_lx_proj, SP_ly_proj);
+						}
 						if (S.SP_flat)	/* Output flat/constant color flag for sphere */
 							PSL_command(PSL, "/SP_flat true def\n");
 						if (S.SP_no_fill)	/* Output no-fill flag for sphere */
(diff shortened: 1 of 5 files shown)


joa-quim:
Don't do it. There are more things to change. And `psxy` ill not receive this option

seisman:
I've added the baseline image to DVC, but currently it has a vertical axis on the left, should it be a 3D basemap instead?

sphere

src/psxyz.c
```diff
@@ -276,6 +276,13 @@ static int usage (struct GMTAPI_CTRL *API, int level) {
 	GMT_Usage (API, 2, "\n%s 3-D Cube: Give <size> as the length of all sides; append q if <size> "
 		"is a quantity in x-units.", GMT_LINE_BULLET);
 
+	GMT_Usage (API, 2, "\n%s 3-D Sphere: Give <size> as sphere diameter [Default unit is cm]. "
+		"Sphere is rendered with radial gradient shading from white at light source to fill color. Modifiers:", GMT_LINE_BULLET);
+	GMT_Usage (API, 3, "+a Set light source azimuth [0, from the right].");
+	GMT_Usage (API, 3, "+e Set light source elevation [90, perpendicular to viewing plane].");
```
seisman: Does `+a`/`+e` really work? I tried to plot a sphere with `-Sc1c+a90+e10` and expect to see the white light near the north edge, but it gives me:

Image

seisman:
> > Does +a/+e really work?
> 
> No. I mentioned that in my first post. And asked opinions if it really worth to exist as option. I still think yes but meanwhile other things came up and I left them in the state they are.

I think they're useful and should be implemented.

joa-quim:
>  I tried to plot a sphere with -SP1c+a90+e10 and expect to see the white light near the north edge,

Is it? I'm not even so sure on what to expect. I would say that the expectation would be to have the bright point at 45º `lat` when looking, for example from -p180/45, and that because we also have to take into account the elevation of the perspective view. This means that illumination should only be allowed to come from the illuminated hemisphere (plus some tolerance for the same reason that we still have day light even after sunset).

Esteban82:
> ```
> echo 3 3 3 | build\src\gmt psxyz -R0/6/0/6/0/6 -JX10c -JZ4c -p180/45 -SP4c+e90+a180 -Gblue -W0.2p,red -Ba -Bza -png lixo
> ```

I got the same result.

Esteban82:
I have a question about this PR. I don't know if it's real or if it has already been considered. 

This PR modifies the `PSL_prologue.ps` file. I understand that this means that from now on, all PS files created by GMT will have this modification. Could this cause problems with previous PS files in DVC used for testing?

joa-quim:
If the rendered images do not change there should be no changes, but only running the tests can answer that and I'm unable to run the tests and the CI tests don't even start too.

Esteban82:
Thanks joa-quim!

PR #8846, opened 2025-12-26
Add color gradients to polygons in psxy 

joa-quim:
Again with Claude assistance I was able to create another of the old wishes for coloring polygons. With the new **-G+g** option we can now created color gradients in polygons. So far only in `psxy` (it was a hell of a work for this one alone). Added a `gradients.sh` test but I'm unable to run it since something completely broke by ability to run bash shells on Windows.

An example
```
# Put this in the poly.txt file
1 1 red
5 1 green
5 5 yellow
2.5 5.5 orange
1 5 blue

gmt plot poly.txt -R0/6/0/6 -JX10c -Ba -G+g -W0.5p,black -png lixo
``` 
lixo

### DIFF
--- a/doc/rst/source/plot.rst
+++ b/doc/rst/source/plot.rst
@@ -212,7 +212,7 @@ Optional Arguments
 
         1. **RGB format**: *x y r g b* where RGB values are in the range 0-255.
         2. **Color name format**: *x y colorname* where *colorname* can be a named color (e.g., "red", "blue"),
-           hex color (e.g., "#FF0000"), or slash-separated RGB (e.g., "255/0/0").
+           hex color (e.g., "#FF0000"), slash-separated RGB (e.g., "255/0/0"), H-S-V or C/M/Y/K formats.
         3. **CPT format**: *x y z* where *z* values are mapped to colors via the CPT specified with |-C|.
 
         Polygons with any number of vertices are supported and automatically triangulated using a fan triangulation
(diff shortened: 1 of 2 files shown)


joa-quim:
It should accept all forms of specifying colors that we already have.

The test. The problem is that the images are too small. (5 cm each)

Esteban82:
> It should accept all forms of specifying colors that we already have.

Yes, that's true. But the rst files only mention using RGB, colornames, and CPT. I tried the H-S-V and C/M/Y/K formats and they worked. 


However, for the grayscale format, an error occurs. It does not interpret them as grayscale but as values to use with a CPT.

```
cat << EOF > triangle.txt
0 0 0
3 0 100
1.5 2.6 200
EOF
   
gmt plot triangle.txt -R-1/4/-1/3 -JX10c -G+g -W0.5p,black -Baf -png gradient_gray

plot [ERROR]: Gradient option -G+g detected CPT format (x y z) but no color palette provided. Use -C<cpt> to specify a color palette.
psconvert [ERROR]: Unable to decode BoundingBox file /home/esteban82/.gmt/sessions/gmt_session.77159/psconvert_77209c.bb (maybe no non-white features were plotted?)
```

Esteban82 (approved):

joa-quim:
Yes, grays with a single value are not possible for the reason you found. This was a particular pain point when implementing the schema that guesses the file format and thus allows to do all with a single option (the **+g**) and not having to provide modifiers for each format.

> But the rst files only mention using RGB, colornames

It mentions

2. **Color name format**: *x y colorname* where *colorname* can be a named color (e.g., "red", "blue"),
           hex color (e.g., "#FF0000"), or slash-separated RGB (e.g., "255/0/0").

PR #8855, opened 2026-01-05
Fix a regression that was causing failure of categorical test.

joa-quim:
Also document -Qt option that was missing since its creation.


joa-quim:
That's a different issue (and it lacks -R to use the -JZ)

Esteban82:
Ok. I will make another PR.

PR #8861, opened 2026-01-07
Add a -T option that can be used with -Fxm (like in gmtregress).

joa-quim:
See this forum [post](https://forum.generic-mapping-tools.org/t/trend1d-model-output-question-feature-request/6211) for this feature request

Credits should go mostly to Claude.


Esteban82 (approved):
I tested the script and it worked fine.

PR #8871, opened 2026-01-22
Fix orders of rho_mantle and rho_crust in gmt|grdflexure.

joa-quim:
They where reversed.


Esteban82 (commented):
Looks fine. Sorry by the delay.

PR #8891, opened 2026-02-05
Update cpt-city URL

Esteban82:
See this issue: https://forum.generic-mapping-tools.org/t/cpt-city-now-in-rails/6347

This PR updates the URL.

I updated the link in a file. I don't know if there are others links that should be updated.


Esteban82:
Thanks @weiji14. I will update those links too.

PR #8901, opened 2026-02-11
Detecting pointer behavior resolves data copying issues.

### LINKED ISSUE #8899: GMT_Open_VirtualFile return all 0.0
When grdpaste reads from virtual files created via GMT_Open_VirtualFile, the data is not copied to the destination buffer.     
  Instead, GMT_Read_Data overwrites the data pointer set by grdpaste.

Chuan1937:
In `gmtapi_import_grid` (within the `GMT_IS_REFERENCE|GMT_VIA_MATRIX` branch), check whether the data pointer has been redirected. If redirected, copy data row-by-row to the target location; otherwise, retain the original behavior.

Fixes #8899


joa-quim (approved):
All clear from GMT.jl too.

PR #8904, opened 2026-02-17
Correctly setting the WrbNZ configuration

Chuan1937:
Added a new memory image detection macro that directly accesses loaded image objects when acquiring regions, bypassing GMT_Read_Data.

Local Julia test successful. Continue #8881


joa-quim (approved):
Congratulations. I tested and it looks fine.

PR #8909, opened 2026-02-20
Add new modifier, +i, to -A option.

joa-quim:
This modifier makes the program only report the high-resolution BoundingBox and end the program. Mostly handy for externals that can thus inquire image size without spending time to rasterize it.


joa-quim:
I tested it, and there is not much to test `-A+i` prints the tight BB and exits.

PR #8916, opened 2026-03-09
pstext -D[j|J] is broken

### LINKED ISSUE #8712: pstext -D[j|J] is broken
Here -D is ignored

```
echo 2 2 Hello World | gmt text -JX15c/0 -Baf -BWSen -R1.89/2.1/1.89/2.1 -Dj5c -png lixo
```

but this works (no 'j')

```
echo 2 2 Hello World | gmt text -JX15c/0 -Baf -BWSen -R1.89/2.1/1.89/2.1 -D5c -png lixo
```

Chuan1937:
For MC (Middle Center) alignment, the `just` value is 6. When calculating the offset in `gmt_smart_justify`:                                                 

```
xx = (2 - (just % 4)) * dx = 0
yy = (1 - (just / 4)) * dy = 0
```
So when text is centered horizontally or vertically, the offset in the corresponding direction becomes 0, rendering the `-Dj` option ineffective.

Solution: When text is centered in a given direction, use the default offset direction (+1) instead of 0:

Fixes #8712

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -14917,8 +14917,8 @@ void gmt_smart_justify (struct GMT_CTRL *GMT, int just, double angle, double dx,
 	gmt_M_unused(GMT);
 	f = (mode == 2) ? 1.0 / M_SQRT2 : 1.0;
 	sincosdegree (angle, &s, &c);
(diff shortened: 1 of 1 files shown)


Chuan1937:
ok, I'll try to fix it again.

PR #8919, opened 2026-03-10
Add a report of p-value to the regress's -Fp option.

joa-quim:
The code was ported from an old version of trend1d that I have in Mirone. Claude digested it and came out with this (quite short) implementation. A test shows it agrees with Matlab's corrcoef function.


Esteban82 (approved):
I test it and works fine.

PR #8965, opened 2026-04-05
text: Add a new modifier to -D (-D+f<file>)

joa-quim:
- Add new modifier **+f**<file> to the **-D** option of `text`/`pstext`
  - The file contains one (*dx*, *dy*) pair per text record, allowing each label to have its own displacement instead of a single fixed offset
  - File reading uses `GMT_Read_Data` so the feature works transparently with GMT.jl in-memory datasets (virtual files)
  - Values in the file are in the current length unit (`PROJ_LENGTH_UNIT`)
  - Added shell test `test/pstext/var_offset.sh`

  ## Example

```
  # offsets.txt has one dx,dy per label
  gmt text -R... -J... -F+f12p+jCM -D0+foffsets.txt << EOF
  -3  -3  SW corner
   3   3  NE corner
  EOF
```

Mostly written by Claude.

### DIFF
--- a/doc/rst/source/text.rst
+++ b/doc/rst/source/text.rst
@@ -159,22 +159,23 @@ Optional Arguments
 
     - **o** - Straight rectangle [Default].
     - **O** - Set a rounded rectangle.
-    - **c** - (Paragraph mode (|-M| only) selects a concave rectangle.
-    - **C** - (Paragraph mode (|-M| only) selects a convex rectangle.
+    - **c** - Paragraph mode (|-M| only) selects a concave rectangle.
+    - **C** - Paragraph mode (|-M| only) selects a convex rectangle.
 
 .. _-D:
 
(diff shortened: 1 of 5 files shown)


remkos (approved):
Looks good to me. I see that en passant you fixed a few other things.

joa-quim:
OK I'll merge it, but I will need that some change the mode of the var_offset.sh script.

Esteban82:
Just for the record, I run the test and looks good to me.

seisman:
PR titles would be more informative if they adopted a format such as 

> module: description of the change.

seisman:
Sorry for commenting on a closed PR, but why not put *dx*, *dy* directly in the same input file, i.e., the input file has the following columns:

```
x y dx dy text
```

joa-quim:
Hmm, I'll need to re-dive on this to remember why things were done the way they were.

PR #8966, opened 2026-04-08
psmeca: let it recognize master CPTs in -C

joa-quim:
Address the issue raised in https://forum.generic-mapping-tools.org/t/changing-beach-ball-color-with-focal-depth-using-meca/6392/8, where a -Cjet was raising an error related to a pen setting.


Esteban82 (approved):
I tested and it works fine.

PR #8967, opened 2026-04-10
psxy: Add broken arrows

joa-quim:
From (new) docs

> When used together with -Sv (Cartesian vectors), the x and y directives produce broken (right-angle elbow) arrows: instead of a straight line from tail to tip, the vector is drawn along two perpendicular legs meeting at a corner. x draws the horizontal leg first, then the vertical leg; y draws the vertical leg first, then the horizontal leg. The arrowhead is placed at the tip.

Co-authored (mostly) by Claude.

Example
```
echo 0 0 45 3 | build\src\gmt psxy -R-5/5/-5/5 -JX10c -Sv0.3c+e -Ay -W1p -Gred -Ba -png lixo
```

Note, the test/psxy/elbowvec.sh file needs to be set to execution mode (I can't do it)

lixo


Esteban82 (approved):
It works.

PR #8969, opened 2026-04-11
Fix the secondary annotations plotted twice reported in #8955

joa-quim:
Assisted-by: Claude: opus-4.6

Fix #8955


Esteban82 (approved):
It works fine. I got the correct figure.

joa-quim:
Yep, but it breaks the azim_radial.sh test

PR #8977, opened 2026-04-14
grdimage: fix the -C...+Uk issue with cpt scale

joa-quim:
Assisted-by: Claude Opus 4.6 (Gemini failed miserably in the same job)

Fix #8950


Esteban82 (approved):
It works.

PR #8978, opened 2026-04-15
psconvert: fix spurious -N warning

### LINKED ISSUE #8868: psconvert warning when using gmt movie: "-N+g or -N+p options do not support the -!"
### Description
When running a very simple `gmt movie` script, I consistently get the following warning printed to the terminal:

`psconvert [WARNING]: The -N+g or -N+p options do not support the -!`

I noticed that this warning is linked to the use of the `-Gred` option in `gmt movie`. If `-Gred` is removed, the warning no longer appears.

### Minimal reproducible example:

```
cat << 'EOF' > main.sh
gmt begin
	echo ${MOVIE_FRAME} | gmt text -R0/1/0/1 -JX10c -F+f200p+cCM -B0 -X0 -Y0
gmt end
EOF

gmt movie main.sh -Nprueba_1 -T25 -C10cx10cx30 -Gred -D4 -Zs -Fmp4
```

The output (MP4/GIF) looks correct and nothing unusual is visible in the result.
I do not recall seeing this warning in earlier versions of GMT (I believe it did not appear in GMT 6.6).
What does the -! option refer to in this context?

joa-quim:
Check that the warning is printed only if -N+g|p was used. The -! option is the one that controls if the ps file is modified _insitu_ or duplicated.

Fix #8868

### DIFF
--- a/src/psconvert.c
+++ b/src/psconvert.c
@@ -1764,8 +1764,9 @@ EXTERN_MSC int GMT_psconvert(void *V_API, int mode, void *args) {
 		GMT_Report (API, GMT_MSG_WARNING, "The -TE option does not support the -!\n");
 	}
 	if (Ctrl->N.BB_paint || Ctrl->N.outline) {
+		if (Ctrl->O.active)
+			GMT_Report(API, GMT_MSG_WARNING, "The -N+g or -N+p options do not support the -!\n");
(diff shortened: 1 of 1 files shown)


Esteban82 (approved):
It works.

PR #8980, opened 2026-04-15
basemap: Fix -A in modern mode.

### LINKED ISSUE #8962: `gmt basemap -R... -J... -A` fails as a modern-mode one-liner
`gmt basemap -R... -J... -A` works in a multi-call modern mode session, works as a classic mode call but fails as a modern mode one-liner with weird error messages, see below. Tried with gmt version 6.6.0.

from the [docs](https://docs.generic-mapping-tools.org/6.6/basemap.html#a):
```
-A[file]

    No plotting is performed. Instead, we determine the geographical coordinates of 
the polygon outline for the (possibly oblique) rectangular map domain. The plot domain 
must be given via [-R](https://docs.generic-mapping-tools.org/dev/basemap.html#r) and
 [-J](https://docs.generic-mapping-tools.org/dev/basemap.html#j), with no other options 
allowed. The sampling interval is controlled via [MAP_LINE_STEP](https://docs.generic-mapping-tools.org/dev/gmt.conf.html#term-MAP_LINE_STEP) 
parameter. The coordinates are written to file or to standard output if no file is specified.
```

`gmt basemap -JS0/-90/20c -R3000+uk -Aout.txt` fails with the following error message with and without `out.txt` requested after `-A`:
```
gmt [ERROR]: Shared GMT module not found: basemap

ERROR: No module named basemap was found.  This could mean one of four cases:
```

`gmt basemap -JS0/-90/20c -R3000+uk -Aout.txt -png quick` also fails:
```
psbasemap [ERROR]: Option -p (ng): Syntax is -p[x|y|z]<azim>[/<elev>[/<zlevel>]][+w<lon0>/<lat0>[/<z0>]][+v<x0>/<y0>]
psbasemap [ERROR]: Option -p parsing failure. Correct syntax:

  -p[x|y|z]<azim>[/<elev>[/<zlevel>]][+w<lon0>/<lat0>[/<z0>]][+v<x0>/<y0>]
     Select a 3-D pseudo perspective view. Append the <azimuth>/<elevation> of
     the viewpoint [180/90], and when used with -Jz|Z, optionally add /<zlevel>
     for basemap level [bottom of z-axis]. Prepend x or y to plot against the
     "wall" x = level or y = level [z]. For a plain rotation about the z-
     axis, give rotation angle only and optionally use modifiers +w or +v to
     select location of axis:
       +w Specify a fixed coordinate point [region center].
       +v Set a fixed projected point [panel center].
psbasemap [ERROR]: Offending option -png
```

`gmt basemap -JS0/-90/20c -R3000+uk -Aout.txt -png quick -B` terminates w/o errors but produces neither the `-A` output nor the plot.

A classic mode call `gmt psbasemap -JS0/-90/20c -R3000+uk -A` produces the desired output.

joa-quim:
I didn't intend to consume resources (both organic and silicium) with this issue that, as discussed in not really a bug) but since I asked Claude by mistake to address it, and it did it ... 

Fix #8962

Assisted: Claude code, Sonnet 4.6

### DIFF
--- a/src/psbasemap.c
+++ b/src/psbasemap.c
@@ -349,8 +349,16 @@ EXTERN_MSC int GMT_basemap (void *V_API, int mode, void *args) {
 	/* This is the GMT6 modern mode name */
 	struct GMTAPI_CTRL *API = gmt_get_api_ptr (V_API);	/* Cast from void to GMTAPI_CTRL pointer */
 	if (API->GMT->current.setting.run_mode == GMT_CLASSIC && !API->usage) {
-		GMT_Report (API, GMT_MSG_ERROR, "Shared GMT module not found: basemap\n");
-		return (GMT_NOT_A_VALID_MODULE);
+		/* Allow -A (dump map boundary coordinates) in classic mode since it writes a dataset, not PostScript */
+		struct GMT_OPTION *options = GMT_Create_Options (API, mode, args);
+		bool dump_only = false;
+		if (API->error) return (API->error);
+		dump_only = (GMT_Find_Option (API, 'A', options) != NULL);
+		gmt_M_free_options (mode);
+		if (!dump_only) {
+			GMT_Report (API, GMT_MSG_ERROR, "Shared GMT module not found: basemap\n");
+			return (GMT_NOT_A_VALID_MODULE);
+		}
 	}
 	return GMT_psbasemap (V_API, mode, args);
 }


mkononets:
a minor thing but the comment on line 352 refers to the classic mode, while it's the modern mode.
```
		/* Allow -A (dump map boundary coordinates) in classic mode since it writes a dataset, not PostScript */
```

joa-quim:
Well, it's inside a condition
```
if (API->GMT->current.setting.run_mode == GMT_CLASSIC ...
```
**-A** is now both classic and modern.

PR #8992, opened 2026-04-20
Let PS format be also _cropable_ from modern mode scripts.

### LINKED ISSUE #8979: Difference between begin and psconvert for pngs

There is a difference between some of the option flags in `gmt begin` and `gmt psconvert`.

```
#!/bin/bash
gmt begin test ps,png E720 -C
gmt basemap -R-2/2/-2/2 -JX5i/5i -BNSEW -Bpxa1 -Bpya1 -Xc -Y3.25i 
gmt end show
gmt psconvert test.ps -A -I+m0.1i -Tg -E720
rm -f gmt.conf gmt.history
```

The above will generate a png that has a page border of 0.1i, while putting the `psconvert` flags in the options in begin will result in a full size page png.

```
#!/bin/bash
gmt begin test ps,png A,E720,-I+m0.1i -C
gmt basemap -R-2/2/-2/2 -JX5i/5i -BNSEW -Bpxa1 -Bpya1 -Xc -Y3.25i 
gmt end show
rm -f gmt.conf gmt.history
```

**Expected outcome**

I would have expected `gmt begin test ps,png A,E720,-I+m0.1i -C` to produce a png that has a proper bounding box size in accordance with the `-A` and `-I` flags, but it only works when `psconvert` is called later. In either case, the bounding box resize only applies to the png and not the ps.

**System information**

* Operating system: Linux Ubuntu 22.04
* GMT version: 6.6.0

joa-quim:
See issue #8979 for reference. AFAIK, cropping ps files from a modern mode script never worked. This PR makes it possible (but not an EPS file) but had to change a default value in `gmt_modern.h`. We changed 

#define GMT_SESSION_CONVERT	"A"
to
#define GMT_SESSION_CONVERT	""

Running the tests doesn't reveal any break, but this is nevertheless a somewhat risky change.

Fix #8979

### DIFF
--- a/src/gmt_modern.h
+++ b/src/gmt_modern.h
@@ -36,7 +36,7 @@
 
 #define GMT_SESSION_NAME	"gmtsession"	/* Default prefix for single-figure modern sessions */
 #define GMT_SESSION_FORMAT	0		/* Default entry into gmt_session_format|code arrays -> PDF */
-#define GMT_SESSION_CONVERT	"A"		/* Default psconvert options in gmt figure */
+#define GMT_SESSION_CONVERT	""		/* Default psconvert options in gmt figure */
 
 /* Declarations of modern mode module names.  These functions will
  * call the classically-named modules (e.g., plot will call psxy) but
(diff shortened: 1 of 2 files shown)


Esteban82 (approved):
I tested it with the following script:

```
#!/bin/bash
gmt begin test ps,png A,E720,I+m0.1i -C
gmt basemap -R-2/2/-2/2 -JX5i/5i -BNSEW -Bpxa1 -Bpya1 -Xc -Y3.25i 
gmt end show
rm -f gmt.conf gmt.history
```

and I obtained both a cropped PS and PNG.

Issue #8987, opened 2026-04-17
grdmath [ERROR]: Cannot use -I together with -R<gridfile>


anbj:
Very brief; I'll give more detail if necessary.

**grdmath** complains about giving ``-I`` when I never did.
``grid.tif`` just contains a bounding box in UTM coordinates.

```
$ gmt grdmath -Rgrid.tif 0 = trash.nc

$ gmt grdmath -Rgrid.tif 1 trash.nc ADD = trash.nc
grdmath [ERROR]: Cannot use -I together with -R<gridfile>
```

joa-quim:
I don't think this is a bug. When when use **-R**<file>, we are setting both the region **and** the increment.

anbj:
Ok, but it gives the impression that I, the user, gave an ``-I``, so a bit confusing. Closing then.

joa-quim:
Yes, confusing and don't see why setting that constrain. Removed it in #8994

anbj:
**Edit**: Forget this! I forgot to build from the PR.

Hmm, still getting (and now better bug report)

```
# create dummy grid
$ gmt grdmath -I1 -R0/10/0/10 0 = grid.nc

# make new grid based on grid.nc
$ gmt grdmath -Rgrid.nc 0 = trash.nc

# test
$ gmt grdmath -Rgrid.nc 1 trash.nc ADD = trash.nc
grdmath [ERROR]: Cannot use -I together with -R<gridfile>
```

joa-quim:
> grdmath [ERROR]: Cannot use -I together with -R<gridfile>

Can't be. I removed those lines from code.

anbj:
Yes - I forgot to build from the PR, sorry.

PR #8994, opened 2026-04-21
grdmath: do not error because on an unused -I when using -R<gridfile>

joa-quim:
See issue #8987

Don't know why the removed test condition was set, but removing seems harmless and let -Rgrid be used. Tests do not show any regression.

### DIFF
--- a/src/grdmath.c
+++ b/src/grdmath.c
@@ -6822,11 +6822,7 @@ EXTERN_MSC int GMT_grdmath (void *V_API, int mode, void *args) {
 	subset = GMT->common.R.active[RSET];
 
 	if (G_in) {	/* We read a gridfile header above, now update columns */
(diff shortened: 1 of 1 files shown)


Esteban82 (commented):
It partially works. 

Now I got this error instead:

`grdmath [ERROR]: Cannot use -r when grid files are specified`.

I used this script:

```
gmt grdmath -Rd -I10 X Y MUL = grid.tif=gd:GTiff
gmt grdmath -Rgrid.tif 0 = trash.nc
gmt grdmath -Rgrid.tif 1 trash.nc ADD = trash2.nc
```

joa-quim:
Yep, but that is a separate issue. It now fails on the next test (after the one I removed). But maybe this one is right.

joa-quim:
Just to be clear. I will not touch in the `-r` case.

Esteban82 (approved):
Ok, so I understand that I could approve this.

PR #9013, opened 2026-05-14
grdview: fix clipping failure.

### LINKED ISSUE #8036: Clipping fails for grdview

joa-quim:
Fix #8036

Assisted-by: Claude Opus 4.7


Esteban82 (approved):
It works.

PR #9018, opened 2026-05-15
Add bitflag to allow resetting GMT->init.runtime_bindir

joa-quim:
The issue is that when called from externals `GMT->init.runtime_bindir` is set to the binary of the external caller. For example from Julia `GMT->init.runtime_bindir` is set to `.../Julia/bin` and this is not good for example to find the Ghostscript location on Windows.

This PR adds a new `GMT_SESSION_BIN2LIBDIR`  bit flag that can be used at the time of `GMT_Create_Session()` call and will set `GMT->init.runtime_bindir` equal to `GMT->init.runtime_libdir`. This latter does not seem to be affected by the guessing machinery in `gmt_begin` and holds the right address on Windows. Unix systems may still be affected by the issue this PR tries to address. In all cases, this should be a harmless change when the `GMT_SESSION_BIN2LIBDIR` bit flag is not used.

### DIFF
--- a/src/gmt_api.c
+++ b/src/gmt_api.c
@@ -8471,6 +8471,11 @@ void * GMT_Create_Session (const char *session, unsigned int pad, unsigned int m
 	}
 	GMT_Report (API, GMT_MSG_DEBUG, "GMT_Create_Session initialized GMT structure\n");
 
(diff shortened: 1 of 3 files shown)


Esteban82 (approved):
Approved without testing.

joa-quim:
@seisman this may have influence in PyGMT as well. Please test it.

seisman:
Currently, PyGMT doesn't rely on the bin directory, so it's not affected.

PR #9014, opened 2026-05-14
psimage: fix -B+t>title> issue

joa-quim:
Fix #5635

Assisted-by: Claude Opus 4.7

PR #9022, opened 2026-05-16
Fix a bad fix in #9014 (the psimage -B+t<title> issue)

joa-quim:
A part of the previous fix in #9014 had broken a couple of other tests (oldvec, etc). This now revert that breaking and oddly still fixes the psimage text rotation issue.


joa-quim:
Merging as the tests are working well again.

PR #9029, opened 2026-05-21
Add new documentation on how to create custom supplements libraries.

joa-quim:
New devdocs page (doc/rst/source/devdocs/custom_supplements.rst) explaining the in-tree vs out-of-tree supplement build mechanism: the five-symbol GMT_MODULEINFO contract, how src/CMakeLists.txt auto-generates the glue and module info header for in-tree supps (windbarbs as reference), and what an out-of-tree project must vendor to reproduce the same wiring.

This work was applied into (re)building the MB-system modules lib such that they show up when doing 'gmt --show-modules' and some more worked test examples let be called from Julia and exchange memory data (like the whole GMT).

We have old documentation on this in other repositories but I frankly find it impossible to understand.

### DIFF
--- a/doc/rst/source/devdocs/custom_supplements.rst
+++ b/doc/rst/source/devdocs/custom_supplements.rst
@@ -1,0 +1,313 @@
+:orphan:
+
+Custom GMT Supplements --- In-Tree vs Out-of-Tree
+=================================================
+
+.. note::
+
+   The investigation underlying this document and the document itself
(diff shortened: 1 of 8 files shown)


Esteban82 (approved):
Looks fine. But I am not sure about its content.

joa-quim:
It is the basis of this MB PR https://github.com/GenericMappingTools/gmt/pull/9029

PR #9032, opened 2026-05-24
Add a standalone program to let update GMT between release cycles (tested on Win only).

joa-quim:
For a long time that I think that updating GMT, at least on Windows, should be trivial and not requesting either a local build or wait for the next release. So, after several sessions with Claude we came out with this solution that relies on a standalone program that does not link with the GMT libs and hence can replace them with new ones. And where are the new ones? In 
https://github.com/GenericMappingTools/gmt/releases/tag/6.6.0

It has a manifest file (manifest_win_x64.txt) with info about new binaries hashes and the new binaries itself.

I only tested on Windows and have doubts if it can work on *nix because of the messy situations regarding where the libs are installed (and symlinked) across different distros and MacOS own things. But I would like that people tried it and see if improvements are necessary.

For Windows I have added the `gmt_update.exe` that will be generally available only at GMT 6.7, which when placed at `...\gmt6\bin` will update GMT only by running `gmt_update`

I'm not providing more details because they are well explained in the `gmt_update.rst`


doc/rst/source/gmt_update.rst
```diff
@@ -0,0 +1,310 @@
...
+combination of operating system and processor architecture.  The compile-time
+channel of an individual ``gmt_update`` binary is fixed and printed at the top
+of the ``--help`` output.  The six supported channels are:
+
+==========  =====================================
+Channel      Description
+==========  =====================================
+win_x64      Windows on Intel/AMD 64-bit
+win_arm64    Windows on ARM 64-bit
+linux_x64    Linux on Intel/AMD 64-bit
```
seisman, line 61: I guess the Linux/macOS channel won't work because the github release page doesn't provide any binaries for Linux/macOS.

doc/rst/source/gmt_update.rst
```diff
@@ -0,0 +1,310 @@
...
+Upload that file together with the referenced binaries to the same release
+(or the same directory on a private mirror) and the consumer side is done.
+
+**7. Generate a manifest from a custom file list and a hand-set version:** ::
+
+    gmt_update --make-manifest -o manifest_win_x64.txt --files my_files.txt --version 6.7.0_rc1
+
+with ``my_files.txt`` containing, e.g. ::
+
+    # Core binaries
```
seisman, line 293: If I understand correctly, the updater on Windows can only update a few DLL/EXE files. The risk is that any changes made to the `GMT/share` or `GMT/doc` directories would not be updated locally. As a result, users could end up with GMT 6.7.0 binaries but `share` or `doc` directories from GMT 6.6.0.

I feel that such inconsistence could create even more confusion when users encounter issues.

doc/rst/source/gmt_update.rst
joa-quim: Sure, it won't right now. It's up to us to produce those binaries and out them there and I only did it for Windows. As I said, I anticipate troubles on *unixes and we may very well have to assume that this wont work for them, but we need to test it first to see if eventual troubles are solvable.

doc/rst/source/gmt_update.rst
joa-quim: Nope. The mechanism is set to accept serving more files (the ones we want). Just did not test it if the updater can do it automatically as well or if it has to be done manually.

seisman: > The mechanism is set to accept serving more files (the ones we want).

It means when making a release, we have to know exactly which files are changed and need updates. For example, if we update the SCM CPT files in GMT 6.7.0, then we have to upload all these CPT files to the GitHub release page. It means more maintenance burden for us.

joa-quim: Again nope. Sorry for not being clear on this and the docs were not either. The updater checks in which major.minor version we are and only updates for the on-the-works next version. So _releases/latest_ has the manifest and updated files that can be used to update for the what's in the current development version. 

Imagine that we already had it implemented in 6.6 (and it is to simulate it that I added the `gmt_update.exe` to the releases). Users can update to keep up with 6.7 development. When 6.7 comes out, _releases/latest_ translates to _releases/6.7.0_ and a user with 6.7 can only update from the manifest in there, which means the 6.8dev cycle.

seisman: It makes more sense. Then can we create a dedicated release (e.g., `nightly`) for these files? I feel it's dangerous to frequently add/update files to the 6.6.0 release page.

joa-quim:
I am coming back to this because I want to it merged before 6.7.0 release. I removed the reference to its docs so it will stay as an undocumented feature till things settle to better. But I would like to have this available in official binaries so that I can explore better how to update the GMT on Windows used by the GMT.jl package. As is, it should be harmless and not interfere in any way with the normal GMT functioning.

seisman:
> Then can we create a dedicated release (e.g., `nightly`) for these files? I feel it's dangerous to frequently add/update files to the 6.6.0 release page.

See my comment at https://github.com/GenericMappingTools/gmt/pull/9032#discussion_r3331404685

joa-quim:
Yes, I saw it (at the time). I just have no time to dedicate to this at the moment before the 6.7. I removed the reference to docs so this remains as an undocumented feature so that still we'll have freedom for improvements but at same time will let me experiment with ways of updating the GMT Win version to use in GMT.jl (I would like to replace the GMT that is currently used by GMT.jl on Windows but need to have a way to do easy updates).

seisman:
OK. Could you please add a note to the source code and docs to emphasize that the feature is highly experimental? Just in case someone finds this PR or the binary and uses it in production.

joa-quim:
OK, added that on the .rst and the `gmt_update.c` (as a comment). Anyway, this can never work if we do not add the necessary files to the `releases`.

seisman (approved):

PR #9033, opened 2026-05-24
psconvert: Do not raise an error when gs is not found in registry or GMT/bin

seisman:
`Ghostscript not found in registry nor in GMT/bin. Fallback to PATH.` should not be an error.

### DIFF
--- a/src/psconvert.c
+++ b/src/psconvert.c
@@ -1750,7 +1750,7 @@ EXTERN_MSC int GMT_psconvert(void *V_API, int mode, void *args) {
 	if ((error = parse (GMT, Ctrl, options)) != 0) Return (error);
 
 	/*---------------------------- This is the psconvert main code ----------------------------*/
-	
+
 	gmt_M_tic(GMT);
 
 	if (!Ctrl->L.active && (GMT->current.setting.run_mode == GMT_CLASSIC) && (Ctrl->In.n_files == 0)) {	/* No files given, bail */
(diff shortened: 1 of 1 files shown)


joa-quim (commented):
But it will then very likely fail silently (chances that it is found in path and not in registry not gmt/bin are slim.

seisman:
> But it will then very likely fail silently (chances that it is found in path and not in registry not gmt/bin are slim.

`gs` is commonly found in PATH on Linux and macOS. Even on Windows, Chinese users have to install ghostscript separately to make GMT support Chinese characters.

joa-quim:
That code is under a `#ifdef _WIN32` so Linux does not enter here.
I remember you mentioned this once (fonts needed for Chinese users) and that was also one of the reasons I change the order by which the `gs` is searched on Windows. Now it searches first in the registry because that is how on Windows `gs` is found. Not from `path`, it is from registry. Next it looks for it (but with the fonts short-come) on `gmt\bin`. Only if this also fails is that message printed. But it would not if chaged to `_DEBUG`

seisman:
> That code is under a `#ifdef _WIN32` so Linux does not enter here.

You're right that these errors don't exist on Linux/macOS but exist on Windows when GMT is installed via conda (e.g., in Windows CI). In https://github.com/GenericMappingTools/pygmt/actions/runs/26330974098/job/77516909838?pr=4653, we see lots of errors.

joa-quim (approved):
Ah OK, so conda decided to invent and installs `gs` on it's own way.

PR #9035, opened 2026-05-25
pstext: clip lines by text box

### LINKED ISSUE #3927: Line in text -D..+v is not clipped
When drawing a line from the point to the offset text with **-Dj**_offset_**+v**_pen_, unless you fill in the text box the line shows up:

`echo "1.91739628589 1.59140756803 T" | gmt text -R-5/5/-5/5 -Jx1 -D-5c+v0.5p,black -F+f32,Helvetica,black+a0+jCB -W0.25p,red -png`

map
 map

It would be better to set up a clip path if no fill is used so that the line stops at the (imaginary or drawn) textbox.

[Moved from old GMT wiki issues from 2013].

joa-quim:
Fix #3927

Assisted-by: Claude Opus 4.7

### DIFF
--- a/src/postscriptlight.c
+++ b/src/postscriptlight.c
@@ -4045,6 +4045,64 @@ int PSL_plotsegment (struct PSL_CTRL *PSL, double x0, double y0, double x1, doub
 	return (PSL_NO_ERROR);
 }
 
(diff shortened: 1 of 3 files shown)


Esteban82 (approved):
It works.

PR #9036, opened 2026-05-25
grdimage: grdimage -A -R does not extend image

### LINKED ISSUE #3106: grdimage -A -R crops but does not extend image
If **-R** is used with **-A** and it crops the image, we get the cropped area back.  However, if we use a -R setting that goes beyond the grid, we do not get extra NaN space back, just the same image.  It seems to me that if we run

```
gmt grdmath -R0/20/0/20 -I0.1 X Y MUL = t.nc
gmt grdimage t.nc -JX5i -A1.png
gmt grdimage t.nc -JX5i -A2.png -R0/20/0/30
```

then image 1.png should be 201x201 while 2.png should be 301x201, but it is still 201x201.

joa-quim:
Fixes #3106

Assisted-by: Claude Opus 4.7

With this PR
```
gmt grdmath -R-20/20/10/20 -I0.1 X Y MUL = t.nc
gmt grdimage t.nc -R-25/5/14.2573318521/23.9576963964+r -JL0/15/10/20/15c -Alixo.png
```
produces

lixo


Esteban82 (approved):
It works.

PR #9037, opened 2026-05-25
pscoast: fix issue with crude coastlines and very large oblique maps

### LINKED ISSUE #3923: Very large oblique map and crude coastline fails to fill properly
When the crudest coast line is used (decided by default in modern mode for this plot), the fill is incorrect:L

`gmt coast -R38.25/-15.27/303.04/52.30+r -JE174.79/-37.01/21c -Wdarkgrey -Glightgrey -B -png bad`

bad

If we step up to the low resolution coastline then it works fine:

`gmt coast -R38.25/-15.27/303.04/52.30+r -JE174.79/-37.01/21c -Wdarkgrey -Glightgrey -B -Dl -png good`

good

[Moved from old GMT wiki issue from 2018]

joa-quim:
Fix #3923

Assisted-by: Claude Opus 4.7

### DIFF
--- a/src/gmt_shore.c
+++ b/src/gmt_shore.c
@@ -523,6 +523,10 @@ int gmt_set_resolution (struct GMT_CTRL *GMT, char *res, char opt) {
 					base = 1;	/* high */
 				else
 					base = 0;	/* full */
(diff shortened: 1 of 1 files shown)


Esteban82:
It works.

PR #9045, opened 2026-05-28
-Xc -Yc and -pz fail to center plot

### LINKED ISSUE #3925: OPtions -Xc -Yc and -pz fail to center plot
This is shifted away from paper center, and affects classic mode plus PS output from modern mode, like here:

`gmt  coast -R-75/-60/-50/-40 -JM-62.5/-45/12c -Df -A0/0/1 -S198/236/255 -pz135/30 -Xc -Yc -Bfa -ps map`

joa-quim:
Fix #3925

Assisted-by: Claude Opus 4.7


Esteban82:
I  got this figure. Is it ok? 

map

joa-quim:
Yes, it is centered as it should.

Issue #6741, opened 2022-05-25
Some issues with 3D plots


uleysky:
First, the command
```bash
gmt psbasemap -R0/10/0/20/0/30 -JX5c/10c -JZ15c -Bswnez1 -pz135/40/0
```
 gives an error
```
psbasemap [ERROR]: Bad interval in -B option (x-component, a-info): wnez1 gave interval = 0
psbasemap [ERROR]: Bad interval in -B option (y-component, a-info): wnez1 gave interval = 0
psbasemap [ERROR]: Option -B parsing failure. Correct syntax:

  -B Specify both (1) basemap frame settings and (2) axes parameters.
     Frame settings are modified via an optional single invocation of -B[<axes>][+b][+g<fill>][+i[<val>]][+n][+o<lon>/<lat>][+s<subtitle>][+t<title>][+w[<pen>]][+x<fill>][+y<fill>][+z<fill>]
     Axes parameters are specified via one or more invocations of -B[p|s][x|y|z]<intervals>[+a<angle>|n|p][+e[l|u]][+f][+l|L<label>][+p<prefix>][+s|S<secondary_label>][+u<unit>
     <intervals> is composed of concatenated [<type>]<stride>[l|p] sub-strings. See basemap documentation for more details and examples of all settings.
psbasemap [ERROR]: Offending option -Bswnez1
```
If I replace "sw" with "ws" or "s" with "S" in the last -B option, the command works fine. The bug is small but curious.

Second, when shifting the xy plane by zlevel parameter of the -p option, the "+b" flag does not draw the bottom face of the cube
```bash
gmt psbasemap -R0/10/0/20/0/30 -JX5c/10c -JZ15c -Bwsenz1+b -pz135/40/10 -P >test1.ps
```
![test1](https://user-images.githubusercontent.com/5275343/170194280-83aaf5d2-2d74-404a-abde-72a08c2f42b2.png)
Also, the manual says
```
+b (for 3-D plots) to draw the foreground lines of the 3-D cube
```
but some of the drawn lines are not foreground.

Third, the command
```bash
gmt psbasemap -R0/10/0/20/0/30 -JX5c/10c -JZ15c -Bxa+lx -Bya+ly -Bza+lz -BwsENZ1+b -px135/40/0 -P >test2.ps
```
produces a result that I find it difficult to interpret. It looks like the z-axis is pointing wrong.
[...]

Issue #529, opened 2019-03-18
Bug in plotting XZ and YZ planes in perspective view


seisman:
**Full script that generated the error**

```
gmt begin perspective pdf
gmt basemap -R0/10/0/10/0/20 -JX3c -JZ3c -Baf -Bzaf -BwsENZ1+b -pz135/40/5
gmt basemap -R0/10/0/10/0/20 -JX3c -JZ3c -Baf -Bzaf -BwsENZ1+b -px135/40/5 -X6c
gmt basemap -R0/10/0/10/0/20 -JX3c -JZ3c -Baf -Bzaf -BwsENZ1+b -py135/40/5 -X6c
gmt end
```
![perspective](https://user-images.githubusercontent.com/3974108/54505205-46a2fd00-490d-11e9-9276-6a4fb15af01d.png)

**Description of the problem**

If my understanding of -p option is correct, `-px` would plot a YZ plane at x=5 (the center of X axis). However, it seems the YZ plane is plotted at x=2.5. The same happens to -py.

Is this a bug or do I misunderstand the `-p` option.

**System information**

* Operating system: macOS
* Version of GMT: 6.0.0_b03d6d7_2019.03.06

PaulWessel:
Certainly looks like a bug or two. We agree that the **-pz** option seems to work but not the other two.  I am also not sure about the lack of the "5" annotation on the z-axis, especially for the last two plots.

PaulWessel:
Perhaps @remkos can have a look at this as he implemented this version of **-p** I think.  I am not sure what our intent is with the "z" value in -p when **-px** or **-py** are selected.  We clearly also have code that seeks to skip the plotting of the z-annotation at the level of the horizontal plane, and that code predates **-px** and **-py** as well.

remkos:
I will put this one on me. The whole plotting with `-px` and `-py` was never implemented robustly, and some things simply cannot work given the way the projection is implemented in PostScript.
However, this is clearly an error and it ought to be corrected.

joa-quim:
#7518 Fixed the bad plane location for the second and third cases. However the base of the cube is no longer plotted, though it wasn't anymore before #7518 fix so an unrelated issue.

seisman:
After PR #7518 I'm getting this:
![perspective](https://github.com/GenericMappingTools/gmt/assets/3974108/d67fd3c2-76c0-45c3-aec6-09cb1f9ef1fd)

joa-quim:
[...]

PR #9047, opened 2026-05-29
Tackle the -px... -py... walls plot option

joa-quim:
This one was tough, even with Claude's help, but I think most (though not all) of the issues (many) raised in #529 and #6741 are now solved. Paul's example with subplot still fails to annotate the Z axis. Another missing feature is that if we plot a xz or yz plane we don't have a mean to plot annotated X & Y axis like we do for Z.

Assisted-by: Claude Opus 4.7

@remkos and @uleysky please give this a try.


seisman:
please add a test for it.

joa-quim:
Done but I can't set it the execution mode.

seisman:
> Done but I can't set it the execution mode.

I've done it for you.

PR #9055, opened 2026-06-01
Fix the baseline image for doc/scripts/images/GMT_-B_geo_2.sh and two grdimage tests

seisman:


seisman:
Actually, the new grdimage baseline image may be incorrect. The ticks at 90N are missing.

QQ20260601-105302

joa-quim:
Yes, that was a side effect of #9051

joa-quim:
The fix in #9056 no longer has this side effect.

joa-quim (approved):

PR #9064, opened 2026-06-14
plot: Fix the description of the -M option

seisman:
In `plot -M`, when `+p` modifier is not set, it defaults to not drawing the curve, but the docs say
> Specify a separate pen for drawing curve :math:`y_1(x)` [Default is same pen as :math:`y_0(x)`]. See |-W| for specifying the pen for curve :math:`y_0(x)`.

The behavior can be confirmed by the following script:
```
#!/usr/bin/env bash
# Demonstrate filling of area between two sinusoidal curves with a section of NaNs
gmt begin GMT_fill_curves
	gmt plot -Mc+gblue+r+l"Short @~l@~ exceeds long" -Bxa90g90+u@. -Byafg -BWStr -JX15c/3c -R0/720/-1.25/2.5 both_NaN.txt -Gred -l"Long @~l@~ exceeds short" -W1p,red
gmt end show
```

### DIFF
--- a/doc/rst/source/plot.rst
+++ b/doc/rst/source/plot.rst
@@ -252,7 +252,7 @@ Optional Arguments
 
 .. _-M:
 
-**-M**\ [**c**\|\ **s**][**+l**\ *seclabel*][**+g**\ *fill*][**p**\ *pen*][**+r**\ *pen*][**+y**\ [*level*]]
+**-M**\ [**c**\|\ **s**][**+l**\ *seclabel*][**+g**\ *fill*][**+p**\ *pen*][**+r**\ *pen*][**+y**\ [*level*]]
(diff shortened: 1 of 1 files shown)


doc/rst/source/plot.rst
```diff
@@ -252,7 +252,7 @@ Optional Arguments
 
 .. _-M:
 
-**-M**\ [**c**\|\ **s**][**+l**\ *seclabel*][**+g**\ *fill*][**p**\ *pen*][**+r**\ *pen*][**+y**\ [*level*]]
+**-M**\ [**c**\|\ **s**][**+l**\ *seclabel*][**+g**\ *fill*][**+p**\ *pen*][**+r**\ *pen*][**+y**\ [*level*]]
```
yvonnefroehlich, line 255: Nice catch!

yvonnefroehlich (approved):

Issue #8999, opened 2026-04-23
Allow fancy scale bar (-L+f) in Cartesian projections


Esteban82:
The fancy map scale option (-L+f) is currently restricted to geographic projections, producing this messagge:

`
basemap [ERROR]: Option -L: No fancy map scale modifier allowed for Cartesian projections
`

I would like to use the fancy map scale for maps in UTM coordinates (like the Example 28).

Could support for `-L+f` be extended to Cartesian projections?

A minimal example:

```
gmt begin ex28 png
	gmt basemap -R@Kilauea.utm.nc -Jx1:160000 -Baf 
	gmt basemap -LjRB+w5000+o0.5c		# This works
	gmt basemap -LjLB+w5000+o0.5c+f		# This doesn't work.
gmt end
```

PR #9069, opened 2026-06-24
Allow -L+f map scales for Cartesian projections

Esteban82:
This PR addresses issue #8999.

Made with help of Claude Sonnet 4.6 free version. 

Tested with:
`gmt basemap -R0/1000/0/1000 -JX10/2c -Baf  -LjLB+w200m+o0.5c+f -png test`

test


joa-quim (approved):
Looks good. Just wonder why it was forbidden on purpose.

PR #9072, opened 2026-06-25
Add support for horizontal and vertical lines from a single value in gmt plot

Esteban82:
This PR adds support for plotting horizontal and vertical lines from a single coordinate value in gmt plot.

Two new symbols are introduced:

-SY plots a horizontal line spanning the full plot width.
-SX plots a vertical line spanning the full plot height.

Example script: 

```
echo 1 > line.txt
echo 3 >> line.txt

gmt begin linea png
    gmt basemap -R0/4/0/4 -Baf -Jx1c
    gmt plot line.txt -SX -Wblue 
    gmt plot line.txt -SY -Wred
    echo 2 | gmt plot -SY -Wgreen,-
gmt end 
```
linea

At the moment, the implementation only supports Cartesian coordinates.

The long-term goal is to extend the functionality to geographic and polar coordinate systems, similar to the implementation [available in PyGMT](https://www.pygmt.org/dev/gallery/lines/hlines_vlines.html) to plot parallels and meridians, or arcs and straight lines along radius at a specified azimuth. It may also be possible to add analogous functionality to plot3d in the future, including a -SZ option for drawing reference lines along the Z axis.

### Question:

Are the proposed symbol codes -SX and -SY acceptable, or would you prefer different letters?


seisman:
I'm a little against such a new feature, mainly because preparing a plain text file with two points is easy in GMT CLI (or Bash).

Esteban82:
I find it a little inconvenient that I need to know the minimum and maximum Y values just to plot a vertical line. For example, in animation 13, we currently use:

```
printf "%s -2.6\n%s 2.8\n" ${MOVIE_COL0} ${MOVIE_COL0} | gmt plot -W0.5p
```

This could be simplified to something like:

```
echo ${MOVIE_COL0} | gmt plot -W0.5p -SY
```

This way, only the X coordinate needs to be specified, and there is no need to know the Y range or use `\n` to provide two points.

joa-quim:
My hesitation on this is more that later, partial vertical (from y1 to y2) or horizontal lines are requested and the mechanism will not allow it.

PR #9074, opened 2026-06-26
Add default reference point for map rose ( -Td and -Tm)

Esteban82:
This PR makes the <refpoint> argument optional for the map rose options (`-Td` and `-Tm`).

When omitted, GMT now defaults to **jTR** (top-right corner inside the map frame) instead of raising a parsing error.

This behavior can be tested with:

`gmt basemap -R-50/-40/-20/-18 -JM10c -Bf -Td+f -png test
`
test

Feedback on the default position is welcome.

### DIFF
--- a/src/gmt_support.c
+++ b/src/gmt_support.c
@@ -4973,8 +4973,17 @@ GMT_LOCAL int gmtsupport_getrose_old (struct GMT_CTRL *GMT, char option, char *t
 		snprintf (string, GMT_LEN256, "x%s/%s", txt_a, txt_b);
 	else	/* Set up ancher in geographical coordinates */
 		snprintf (string, GMT_LEN256, "g%s/%s", txt_a, txt_b);
-	if ((ms->refpoint = gmt_get_refpoint (GMT, string, option)) == NULL) {
-		GMT_Report (GMT->parent, GMT_MSG_ERROR, "Option -%c:  Map rose reference point was not accepted\n", option);
+	char default_ref[GMT_LEN256];
+	/* If no reference point is provided (it starts with '+' or is empty), use jTR by default */
+	if (text[1] == '+' || text[1] == '\0') {
+		snprintf(default_ref, GMT_LEN256, "jTR%s", &text[1]);
+		ms->refpoint = gmt_get_refpoint(GMT, default_ref, option);
+	} else {
+		ms->refpoint = gmt_get_refpoint(GMT, &text[1], option);
+	}
+
+	if (ms->refpoint == NULL) {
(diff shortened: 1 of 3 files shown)


src/gmt_support.c
```diff
@@ -14280,19 +14289,37 @@ int gmt_getrose (struct GMT_CTRL *GMT, char option, char *text, struct GMT_MAP_R
...
 	ms->panel = save_panel;	/* In case it is not NULL */
-	if (!strstr (text, "+w") && (strchr ("gjn", text[1]) == NULL || strchr ("fm", text[1]))) return gmtsupport_getrose_old (GMT, option, text, ms);	/* Most likely old-style args */
+
+	/* Inject default reference point (jTR) if omitted by user --- */
+	char refpoint_str[GMT_LEN256];
+	char *parse_text = text;
+	/* If the user omitted the reference point (e.g., -Td or -Tm), inject a default one (jTR = Top-Right inside) */
+	if (text[0] == 'd' || text[0] == 'm') {
+		if (text[1] == '+' || text[1] == '\0') {
+			snprintf(refpoint_str, GMT_LEN256, "%cjTR%s", text[0], &text[1]);
```
joa-quim: Normally we use length-1 to account for the end of string '\0'. That is `GMT_LEN256-1`

joa-quim:
```suggestion
			snprintf(refpoint_str, GMT_LEN256-1, "%cjTR%s", text[0], &text[1]);
```

joa-quim:
Top-Right seems good to me.

Esteban82:
> Top-Right seems good to me.

Ok. But I think that in pygmt it is at the bottom left corner by default ([link](https://www.pygmt.org/dev/gallery/embellishments/directional_roses.html)). Is this correct @seisman? Should we use the same default position? Or is it ok to have different positions for GMT and PyGMT?

joa-quim:
Asked Google and it came out with
```
The traditional symbol for directions on a map (the compass rose) is typically placed in the upper right corner. Placing it here keeps it out of the way of the main map details while remaining easily visible. If an upper corner is unavailable, it is placed in the lower right or lower left.
```
I think UR is the best default location.

Esteban82:
Fine by me.

joa-quim (approved):

seisman:
> > Top-Right seems good to me.
> 
> Ok. But I think that in pygmt it is at the bottom left corner by default ([link](https://www.pygmt.org/dev/gallery/embellishments/directional_roses.html)). Is this correct @seisman? Should we use the same default position? Or is it ok to have different positions for GMT and PyGMT?

Thanks for letting me know the changes. `Figure.directional_rose` and `Figure.magnetic_rose` are newly added methods and are not released yet, so we will update them to match the GMT default behavior.

PR #9076, opened 2026-06-29
Allow map inset dimensions as percentages

Esteban82:
This PR adds support for specifying map inset dimensions as percentages of the figure size. For example, using -DjTR+w24% sets the inset width to 24% of the map width. When a single percentage is given, the inset is drawn as a square. Both dimensions can also be specified independently, as in -DjTR+w24%/30%, or mixed with absolute units, like -DjTR+w4c/30%. The change is fully backward compatible with existing scripts.

Example:
```
gmt begin ex44 png
	gmt coast -R110E/170E/44S/9S -JM15c -Gbrown -Baf
    gmt inset begin -DjTR+w24% -F+gwhite+p1p+c0.1c+s -Vi
		gmt coast -Rg -JG120/30S/? -Gbrown -Bg
	gmt inset end
gmt end

```
ex44

As a future enhancement, it would be possible to add modifiers such as `+dw`, `+dh`, `+du`, and `+dl` (similar to those used in `-J`) to let the user choose which map dimension a single percentage refers to (width, height, maximum, or minimum).

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -8903,7 +8903,8 @@ void gmt_mapinset_syntax (struct GMT_CTRL *GMT, char option, char *string) {
 	GMT_Usage (API, 3, "%s Give <xmin>/<xmax>/<ymin>/<ymax>[+u<unit>] of bounding rectangle in projected coordinates [meters].", GMT_LINE_BULLET);
 	GMT_Usage (API, 3, "%s Set reference point and dimensions of the inset:", GMT_LINE_BULLET);
 	gmt_refpoint_syntax (GMT, "D", NULL, GMT_ANCHOR_INSET, 1);
-	GMT_Usage (API, 3, "Append +w<width>[<u>]/<height>[<u>] of bounding rectangle (<u> is a unit from %s).", GMT_DIM_UNITS_DISPLAY);
+	GMT_Usage (API, 3, "Append +w<width>[<u>][%%][/<height>[<u>][%%]] of bounding rectangle (<u> is a unit from %s). "
+		"Append %% instead of a unit to set that dimension as a percentage of the map width or height, respectively.", GMT_DIM_UNITS_DISPLAY);
 	gmt_refpoint_syntax (GMT, "D", NULL, GMT_ANCHOR_INSET, 2);
 	if (GMT->current.setting.run_mode == GMT_CLASSIC) {
 		GMT_Usage (API, 2, "Append +s<file> to save inset lower left corner and dimensions to <file>.");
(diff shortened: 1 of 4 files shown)


joa-quim:
Federico, when using AI, please add info on what model was used.

Esteban82:
I used Sonnet 4.6 and Qwen3.7-Plus.

joa-quim (commented):
My first reaction is that this over complicates what seems an interesting option. A simple % for the width seams enough for me. Too many options add to the already complicated cases we have in GMT.

Esteban82:
> My first reaction is that this over complicates what seems an interesting option. A simple % for the width seams enough for me. Too many options add to the already complicated cases we have in GMT.

I agree completely. This PR implements exactly that: a simple % for the width (and optionally height). When only one percentage is given, it produces a square inset, which is the most intuitive behavior.
I don't plan to implement the +dw, +dh, +du, +dl modifiers in the near future. The current implementation keeps things simple and covers the most common use case.

Esteban82:
> I used Sonnet 4.6 and Qwen3.7-Plus.

For further details. First, I used Sonnet to implement the use of %. Then I used Qwen to produce a square when a single value is provided (the first implementation applied a % to both dimensions).

joa-quim:
But specifying %w/%h would deform the inset. That's why I said a single value is what it needs.

Esteban82:
Ok. I see. Do you want me to change the implementation to accept only a single percentage value, or should I leave it as it is?

joa-quim:
Well, didn't test it but I'm assuming that by setting only %w the inset's aspect ratio is preserved. Giving tho percentages wouldn't, so I guess yes.

Esteban82:
> Well, didn't test it but I'm assuming that by setting only %w the inset's aspect ratio is preserved. Giving tho percentages wouldn't, so I guess yes.

If a single value is used (such as 20%), a square is drawn that is 20% of the figure's width.

joa-quim:
Squares are not good test cases, you must test test it with a rectangular inset region and check that a single %w keep the inset's aspect ratio.

seisman:
> setting only %w the inset's aspect ratio is preserved.

I don't think it is useful in the real world. The inset module is primarily used to display a large-region map which usually can have a different aspect ratio than the main map. For instance, let's take an 8cm X 4cm map of a small area for the main map and an inset map showing the globe.

joa-quim:
I didn't mean it to keep the aspect ratio of the large region. I mean, the aspect ratio of the inset itself.

Esteban82:
I just realized that inset begin can receive -R and -J, e.g.

`gmt inset begin ... -R15W/25E/30N/48N -JM5c`

In that case GMT already knows the inset map geometry at inset begin time.

Is that what you meant by preserving the aspect ratio of the inset itself?

BTW, I'll update the documentation to better describe this option.

joa-quim:
I think yes. For what I understand the percentage option it would replace the `5c` of your example. But maybe this is too much complication for a little gain. User will always have to set the image dimension, so doing the same for the inset doesn't seem a big extra. And what would happen if scale is given as 1:xxxxxx, does this % option still work?

PR #9077, opened 2026-06-29
Add default reference point for map inset (-D)

Esteban82:
Assisted-by: **Claude Sonnet 4.6**

This PR makes the <refpoint> argument optional for the map inset option (`-D`).

When omitted, GMT now defaults to `jTL` (top-left corner inside the map frame) instead of raising a parsing error.

Test with:

```
gmt begin ex44 png
	gmt coast -R110E/170E/44S/9S -JM15c -Gbrown -Baf
    gmt inset begin -D+w4c -F+gwhite+p1p+c0.1c+s -Vi
		gmt coast -Rg -JG120/30S/? -Gbrown -Bg
	gmt inset end
gmt end
```
ex44

Please let me know if you’d prefer a different default setting. In PyGMT the Bottom Left corner is used.

### DIFF
--- a/doc/rst/source/inset.rst
+++ b/doc/rst/source/inset.rst
@@ -53,8 +53,8 @@ Required Arguments (begin mode)
 
     .. include:: explain_refpoint.rst_
 
-    If *refpoint* is omitted, the map inset is placed by default in the upper-left corner of the map
-    (using **jTL**).
+    If *refpoint* is omitted, the map inset is placed by default in the upper-right corner of the map
+    (using **jTR**).
 
     Alternatively, Give *west/east/south/north* of geographic rectangle bounded by parallels
     and meridians; append **+r** if the coordinates instead are the lower left and
--- a/doc/rst/source/inset.rst
+++ b/doc/rst/source/inset.rst
@@ -48,10 +48,13 @@ Required Arguments (begin mode)
 
 .. _-D:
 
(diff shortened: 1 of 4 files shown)


joa-quim:
Maybe it's only my taste, but I would put it at TR (that's what you did in the other PR) or BR

Esteban82:
Yes, I would put it there too, but it would be the same corner used by the directional arrow (in #9074). That's why I chose "top left" so that the default positions wouldn't overlap.

joa-quim:
Well, in that (very very rare) case the user has to decide where he/she wants to pout each of them.

Esteban82:
I don't think it's that rare to have both an inset and a north arrow on the same map. 

But if you insist with TR, I can change it.

Esteban82:
Changed to TR.

joa-quim (approved):

yvonnefroehlich:
For PyGMT we have currently the Bottom Left corner of the plot as default

Esteban82:
I have no problem to change it. @joa-quim ?

joa-quim:
GMT.jl has [insets](https://www.generic-mapping-tools.org/GMTjl_doc/tutorials/insets/insets.html) too, but more elaborated as they can search _best locations_, but for full filled base image the default is TR. And the example in the GMT inset page, also set TR as the location. I don't think BL is a common default location for an inset.

PR #9079, opened 2026-07-01
Add default position for map scales when refpoint is omitted

Esteban82:
**Assisted by Claude Sonnet 4.6** 

This PR adds a default placement for map scales when refpoint is omitted. The scale is automatically placed in the lower-right corner of the map (**+jBR+o0.5c**), slightly offset from the frame.

Testing
`gmt basemap -R0/10/0/1 -JM10c -Baf -L+f+w150k -png test`

test

Let me know if you prefer another position.


joa-quim:
Hmm, sorry disagreeing again but this is a case where the default must clearly be LL.

Esteban82:
Ok. I will change to LL. Thanks for the feedback.

Esteban82:
I just changed to jBL.

Esteban82:
Now I got this figure.
test

yvonnefroehlich:
In PyGMT we have for the high-level method [ `Figure.scalebar`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.scalebar.html#pygmt.Figure.scalebar):

> If not specified, defaults to the Bottom Left corner of the plot with a 0.2-cm and 0.4-cm offset in the x- and y-directions, respectively.

as default.

Esteban82:
Okay, I'll update the offset with those values. I suspect they were intended to be higher than the one I used. Thanks Yvonne for letting me know.

Esteban82:
Offset changed.

`gmt basemap -R0/10/0/2 -JM10c -Baf  -L+f+w150k -png test`

test

joa-quim (approved):

remkos (approved):

PR #9086, opened 2026-07-03
Fix double filtering in psevents when using -q option

### LINKED ISSUE #8460: Bug. Data get filter twice with psevents -q0:inc:
With the improve version of animation 1 (#8459) I use the [-q option](https://docs.generic-mapping-tools.org/dev/gmt.html#qi-full) to plot every 10 values (so I can avoid creating the point_sin.txt file). 

I add `-qi0:10: ` to `events` and I the values of 90, 190, 290 were plot (so every 100th value starting in 0). If I use plot instead, then I got the image ok. 

	gmt plot     sin_curve.txt -Sc0.3 -Gyellow -qi0:10:${MOVIE_FRAME}
	gmt events sin_curve.txt -i0,1,0 -T${MOVIE_FRAME} -Sc0.2c -Gdarkred -qi0:10: -Vi

In the terminal, I got this message:
  `events [INFORMATION]: Events read: 37 Event symbols plotted: 35 Event labels plotted: 0`

So, it thinks that 35 symbols were plotted (as `gmt plot` did) but there are only 3. 

I suspect the `events` uses `plot` under the hood. And the data is filter by both modules. I tried to understand if this really happens (and where) but I couldn't. If someone could give a hint, I could try to solve it. 

Bug_anim01

## Full script:

```
# 1. Create files needed in the loop
cat << 'EOF' > pre.sh
gmt math -T0/360/1 T SIND = sin_curve.txt
gmt begin
	gmt basemap -R0/360/-1.2/1.6 -JX22c/11.5c -X1c -Y1c \
	-BWSne+glightskyblue -Bxagf -Bya0.5f0.1g1 --FONT_ANNOT_PRIMARY=9p
gmt end
EOF
# 2. Set up the main frame script
cat << 'EOF' > main.sh
gmt begin
	gmt basemap -R0/360/-1.2/1.6 -JX22c/11.5c -X1c -Y1c -B+n
	i=10
	gmt plot   sin_curve.txt -Sc0.3 -Gyellow -qi0:$i:${MOVIE_FRAME}
	gmt events sin_curve.txt -i0,1,0 -T${MOVIE_FRAME} -Sc0.2c -Gdarkred -qi0:$i: -Vi
	
gmt end
EOF
# 3. Run the movie
gmt movie main.sh -Sbpre.sh -Chd -Tsin_curve.txt -D50 -NBug_anim01 \
	-Lf+t"a = %3.3d"+f14p,Helvetica-Bold+jTL+o1.25/1.15 -M345,png
```

Esteban82:
**Assisted by Qwen 3.7 Plus**

### Bug:
psevents applies the -q filter, and writes the selected data to temporary files. Afterwards, it calls internal modules to plot the symbols/labels from these temporary files. Because the -q option remained active, the internal modules applied the exact same filter a second time on the already-filtered temporary files.

### Solution:
Implemented a local patch to temporarily suspend the -q option before calling the internal modules, and restore it immediately after.

### Alternative solution:
I also considered modifying the global `gmt_disable_bghio_opts()` and `gmt_reenable_bghio_opts()` functions in `gmt_io.c` to include `-q`. However, this will affect dozens of other modules, and I'm not sure if the -q option should also be disabled for them. But maybe it should. 

Test with:

I use -`qi0:30: `to plot a data point every 30. 

```
gmt math -T0/360/1 T SIND = sin_curve.txt
cat << 'EOF' > main.sh
gmt begin
	gmt basemap -R0/360/-1.2/1.6 -JX22c/11.5c -X1c -Y1c -Bxa30g30f -By
	gmt events sin_curve.txt -i0,1,0 -T${MOVIE_FRAME} -Sc0.2c -Gblack -qi0:30:
gmt end
EOF
gmt movie main.sh -Chd -Tsin_curve.txt -Ntest -Ml,png -Zs

```
test

Closes #8460.

### DIFF
--- a/src/gmt_common.h
+++ b/src/gmt_common.h
@@ -251,7 +251,7 @@ struct GMT_COMMON {
 		char string[2][GMT_LEN64];
 		unsigned int col;	/* When +c<col> sets a specific data column */
 		unsigned int mode;	/* 1 for in row-range check, 2 for in time check, 3 for out-row check, 4 for out-time check */
-		uint64_t *rec;		/* POinter to the relevant record counter (dataset, table, segment) */
+		uint64_t *rec;		/* Pointer to the relevant record counter (dataset, table, segment) */
 	} q;
 	struct s {	/* -s[r] */
 		bool active;
--- a/src/psevents.c
+++ b/src/psevents.c
@@ -913,8 +913,8 @@ EXTERN_MSC int GMT_psevents (void *V_API, int mode, void *args) {
 	bool do_coda, finite_duration, has_text, out_segment = false, have_previous_point = false;
 
 	/* Save state for -q option to prevent double filtering in internal calls. Fixes #8460. */
-	bool q_active_in_save, q_active_out_save;
-	unsigned int q_mode_save; 
+	bool q_active_in_save, q_active_out_save = false;
+	unsigned int q_mode_save = 0; 
 
 	int error;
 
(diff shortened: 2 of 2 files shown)


joa-quim (approved):
I'll be very extra cautious in using other models than Anthropic's (even Sonnet, though Sonnet 5 is a good surprise but previous one was moderately bad). GLM, for example, is ver acclaimed but my experience with it was bad. Have no experience with OpenAI top models though.
This fix seams simple so it should not harm.

Esteban82:
I agree. I will ask Sonnet to review the patch before merging, just to get an additional set of eyes on it.

Esteban82:
Based on Claude's suggestion:

-  I initialized the values to false or 0.
-  I added the {} when the settings are recovered so the `if` applies to all three statements instead of only the first one.

PR #9094, opened 2026-07-08
Fix the access to layers in a nc cube file that has more that one variable.

joa-quim:
Before, this cube file with two variables: 'z' and 'bathymetry' always returned the 'bathymetry' layer.

```grdinterpolate tsu.nc?z -T570 -Glay570.grd```

Now it works, (note the 'tsu.nc' file was created with the new nswing supplement).

Assisted-by: Sonnet 5.0


joa-quim:
Shit. I added the grdinterpolate fix in this branch.

PR #9095, opened 2026-07-08
Fix extracting layers from a nc cube containing more that one variable.

joa-quim:
Before, this cube file with two variables: 'z' and 'bathymetry' alway…s returned the 'bathymetry' layer.

grdinterpolate tsu.nc?z -T570 -Glay570.grd

Now it works, (note the 'tsu.nc' file was created with the new nswing supplement).

Assisted-by: Sonnet 5.0


Esteban82:
Let me know if you want me to test it.

joa-quim:
I've merged this already. I did test it with the situation where the bug showed up (the nc file created by nswing).

Issue #9073, opened 2026-06-26
Segmentation fault in coast using -JE and -R+r modifier


biggle1856:
**Description of the problem**
Trying to draw an Azimuthal equidistant projection based on Singapore in order to show the full extent of SIA's route network. 
Segmentation fault in coast (and only in coast) when I use the -R+r modifier

**Full script that generated the error**

```
gmt coast -Vd -R6.97229795181/-37.0769712136/278.20744191/7.44518494882r -JE103.989333/1.359211/17.7c -Glightgrey -pdf t
```

**Full error message**

```
coast [INFORMATION]: Selected ice front line as Antarctica coastline
coast [INFORMATION]: GSHHG version 2.3.7
coast [INFORMATION]: Derived from World Vector Shoreline, CIA WDB-II, and Atlas of the Cryosphere
coast [INFORMATION]: Processed by Paul Wessel and Walter H. F. Smith, 1994-2017
coast [DEBUG]: Running in PS mode modern
coast [DEBUG]: Use PS filename /home/jch/.gmt/sessions/gmt_session.88313/gmt_0.ps-
coast [DEBUG]: Create hidden PS file /home/jch/.gmt/sessions/gmt_session.88313/gmt_0.ps-
coast [DEBUG]: No figure file /home/jch/.gmt/sessions/gmt_session.88313/gmt.figures - nothing to do
coast [DEBUG]: Got session name as t and default graphics formats as pdf
coast [DEBUG]: Basemap order: Frame = above  Grid = above  Tick/Annot = below
pscoast (gmtlib_lonpath): Could not reallocate memory [68719476736.00 Gb, 9223372036854775809 items of 8 bytes]
pscoast (gmtlib_lonpath): Could not reallocate memory [68719476736.00 Gb, 9223372036854775809 items of 8 bytes]
ERROR: Caught signal number 11 (Segmentation fault) at
/usr/local/lib64/libgmt.so.6(gmtlib_lonpath+0x1ea) [0x7f88d5da3eaa]
[0x0]
Stack backtrace:
/usr/local/lib64/libgmt.so.6(gmt_sig_handler_unix+0xbd) [0x7f88d5d3ba6d]
/lib64/libc.so.6(+0x1a0b0) [0x7f88d2c210b0]
/usr/local/lib64/libgmt.so.6(gmtlib_lonpath+0x1ea) [0x7f88d5da3eaa]
[...]

PR #9104, opened 2026-07-15
Patch the crash reported in #9073

joa-quim:
The crash occurs because lat1 = NaN, but only on unix. On Windows it works fine. The true fix must be to find WHY lat1 = NaN in the first place, bug debugging on Linuxd is behind my suffering threshold.

And, on Windows the command:

gmt coast -R6.97229795181/-37.0769712136/278.20744191/7.44518494882r -JE103.989333/1.359211/17.7c -Glightgrey -Dc -png lixo

shows a vertical stray edge gray stripe, but on Linux the example now works fine.


Esteban82:
I got this figure. There is an issue on the edge but it didn't crush.

lixo

Esteban82 (approved):

joa-quim:
Gremlins are attacking us from all fronts. That is what I get on Windows, but not what I get on WSL.

joa-quim:
This patch fix, also fixed the Linux test failure. Now there all tests pass again.

PR #9122, opened 2026-08-06
Allow setting angles of primary and secondary annotations separately

### LINKED ISSUE #7036: Allow setting angles of primary and secondary annotations separately

```
gmt basemap -R0/10/0/10 -JX10c -Bpxa1+a30 -Bsxa5+a60 -Byaf -png map
```

map

I expect that the primary and secondary annotations should have different angles.

Esteban82:
**Done with Claude Sonnet 5**

Tested with (secondary axis with a different angle:

`    gmt basemap -R0/10/0/10 -JX10c -BS -Bpxa1+a30 -Bsxa5+a-60 -png test`

Fixed: 
7036_fixed

Without fixed (all the numbers have the same angle):

7036_without_fix

Fixes #7036. Would you like to test it @seisman?

Should I add a test?

### DIFF
--- a/src/gmt_init.c
+++ b/src/gmt_init.c
@@ -4933,7 +4933,7 @@ GMT_LOCAL int gmtinit_parse5_B_option (struct GMT_CTRL *GMT, char *in) {
 
 	char string[GMT_BUFSIZ] = {""}, orig_string[GMT_BUFSIZ] = {""}, text[GMT_BUFSIZ] = {""}, *mod = NULL, *the_axes = "xyz";
 	struct GMT_PLOT_AXIS *A = NULL;
-	unsigned int no;
+	unsigned int no, pr;
 	int k, error = 0;
 	bool side[3] = {false, false, false}, implicit = false;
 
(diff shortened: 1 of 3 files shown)


joa-quim:
The _fixed_ case is not right. **+a** is the angle with respect to horizontal. The _non-fixed_ angles are the right ones.

Esteban82:
The docs actually back up the fixed behavior here:

> The terms "primary" and "secondary" do not reflect any hierarchical order of units: the "primary" annotation interval is usually smaller (e.g., days) while the "secondary" annotation interval typically is larger (e.g., months).
> (`doc/rst/source/explain_-B.rst_`, describing **-B**\[**p**\|**s**\])

Following that convention (primary = smaller interval, secondary = larger interval), this makes it very clear:

```
gmt basemap -R0/15/0/15 -JX5c -BS -Bpa1+a5 -Bsa5+a85 -png test_7036
```
test_7036

Primary (`-Bpa1+a5`, interval 1) → minimal rotation (5°, essentially horizontal), closest to the axis.
Secondary (`-Bsa5+a85`, interval 5) → maximal rotation (85°, near vertical), farther from the axis.

Each one clearly uses its own `+a` value, measured from horizontal, exactly as requested — no mixing between primary and secondary. Let me know if there's a different combination where you're still seeing the issue.

joa-quim:
OK, you are right. The bug is that it applies the angles of the secondary axes to the primary as well.

joa-quim (approved):

PR #9133, opened 2026-08-21
Remove executable permission check from code-validator

### LINKED ISSUE #9048: Remove unnecessary executable permissions from Bash example and test scripts
Currently, many Bash scripts in the repository are marked as executable. I wonder if we really need executable permissions for all of them.

In Linux, the executable permission is mainly useful for scripts that are intended to be run directly from the command line, e.g., `./build.sh`. However, most of the Bash scripts in this repository are examples or test scripts. They are typically executed through an interpreter, e.g., `bash ex01.sh`. In such cases, the executable permission is not required.

Removing the executable permission can also solve the @joa-quim's issue on Windows (https://github.com/GenericMappingTools/gmt/pull/9047#issuecomment-4569588832).

Therefore, I suggest removing executable permissions from Bash scripts that are not intended to be executed directly by users, and keeping executable permissions only for a few scripts in the GMT/bin directory.

Esteban82:
**Assisted-by: Claude Sonnet 5 (High effort)**

Removed the executable permission from example and test `.sh` scripts under `doc/` and `test/`.
Updated the `code-validator.yml` CI checker to reflect this new policy.

Closes [#9048](https://github.com/GenericMappingTools/gmt/issues/9048)


seisman:
I don't think these scripts need executable permissions. They can be run via `bash xxx.sh`.
```
$ find . -name "*.sh" ! -path "./doc/*" ! -path "./test/*" \
            ! -path "./share/tools/gmt_functions.sh"
./ci/config-gmt-windows.sh
./ci/config-gmt-unix.sh
./ci/download-coastlines.sh
./ci/install-dependencies-linux.sh
./ci/install-dependencies-macos.sh
./ci/simple-gmt-tests.sh
./ci/deploy-gh-pages.sh
./ci/build-gmt.sh
./ci/install-dependencies-windows.sh
./admin/src_replace_word.sh
./admin/src_convention_check.sh
./admin/build-scientific-colors-cpt.sh
./admin/build-macos-external-list.sh
./admin/copyright_year.sh
./admin/list_known_failures.sh
./admin/remote_files_used.sh
./admin/run_cppcheck.sh
./admin/git_release_history.sh
./admin/image_optimize.sh
./admin/gs_check.sh
./admin/build-macos-cmake-cache.sh
./admin/gmt_make_pattern_include.sh
./admin/build-cmocean-cpt.sh
./admin/examine-failure-log.sh
./admin/build-release.sh
./admin/place-release.sh
./share/tools/gmt_links.sh
./share/tools/gmt_prepmex.sh
./share/tools/gmt_uninstall.sh
./share/tools/christmas-cube.sh
./src/gsfml/fz_funcs.sh
./src/gmt_make_PSL_strings.sh
./src/gmt_make_enum_dicts.sh
./src/gmt_make_module_purpose.sh
./src/spotter/spotter.sh
./src/shelltest.sh
./src/windbarbs/test/grdbarb_example.sh
./src/windbarbs/test/psbarb_example.sh
./src/mgd77/make_new_dst.sh
./src/mgd77/reformat_kyoto_dst.sh
./src/mgd77/mgd77netcdfhelper.sh
./src/mgd77/mgd77_codes.sh
./src/x2sys/test_x2sys.sh
./src/gmt_shell_functions.sh
```

seisman:
`gmt_functions.sh` is expected to be used like `source gmt_functions.sh`, so it also doesn't need executable permission.

Esteban82:
Fixed now. No more exceptions, no *.sh should be executable.

seisman:
The Tests workflow fails and needs to be fixed. it can be done in a separate PR because this PR already contains too many changed files.

Esteban82:
> The Tests workflow fails and needs to be fixed. it can be done in a separate PR because this PR already contains too many changed files.

They are failing because of this PR, right?

seisman:
I think we need to update https://github.com/GenericMappingTools/gmt/blob/63127378c54970ca1d21d5223c9750b2c596891d/test/gmtest.in#L309

to 
```
bash "${local_script}"
```

Esteban82:
It sounds fine. I will do make another PR for this.

joa-quim:
I don’t understand why is needed this to change all these files. If the solution is to call them prefixed with bash it doesn’t matter if they have the executable mode or not.

seisman:
As I understand it, scripts created under Windows do not have executable permissions by default. Consequently, we may encounter inconsistent executable permissions across our scripts in the future.

Esteban82:
I think Joaquim meant that it isn't necessary to remove the execution permission from all existing .sh files. It makes sense to be consistent, but it isn't required.

seisman:
Yes, not required, but better to be consistent.

joa-quim:
Sorry but I fail to see where consistency plays here. For me avoiding changing > 150 files needlessly (and a change that Windows will not see) seems more important.

seisman:
I'm OK with keeping the executable permissions.

.github/workflows/code-validator.yml
```diff
@@ -65,9 +65,12 @@ jobs:
 
       - name: Check execute permission of bash scripts
```
seisman: I think we should remove the check because we no longer care about execute permissions.

Esteban82:
Now it can be review.

.github/workflows/code-validator.yml
```diff
@@ -63,11 +63,14 @@ jobs:
           gmt_version_year=$(grep GMT_VERSION_YEAR cmake/ConfigDefault.cmake | awk -F'"' '{print $2}')
           if [ "$current_year" != "$gmt_version_year" ]; then exit 1; fi
 
-      - name: Check execute permission of bash scripts
+      - name: Check read permission of bash scripts
```
seisman: Perhaps checking read permission is not necessary, because files without read permission are very unusual.

Esteban82: ok, so should I delete all this step?

seisman: I feel so.

Esteban82: Ok

seisman (approved):
remember to update the pr title

PR #9137, opened 2026-08-22
CI: Replace setup-miniconda with setup-micromamba on macOS

seisman:
Replace `setup-miniconda` with `setup-micromamba`, which is much faster, and also fix the failures in the macOS CI.


Esteban82:
I understand that it can be approved, given that it built successfully on macOS.

Esteban82 (approved):

PR #9140, opened 2026-08-23
CI: Deduplicate links in the Check Links report

Esteban82:
The same broken link often gets listed once per file that references
it, so a single dead link can flood the weekly "Link Checker Report"
issue with repeated entries.

Added a post-processing step that keeps only the first occurrence of
each link in the report and drops the per-file subsections that end
up empty after deduplication.

Assisted-by: Claude Sonnet 5 (High effort)

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -58,6 +58,54 @@ jobs:
           **/*.rst_
           **/*.md
 
+    - name: Remove duplicate links from the report
+      if: steps.lychee.outputs.exit_code != 0
+      shell: python3 {0}
+      run: |
+        import re
+        from pathlib import Path
+
+        report = Path("/tmp/lychee-out.md")
+        lines = report.read_text().splitlines()
+
+        # Each bullet line reports one link, e.g. "* [404] <https://example.com> (at ...)".
+        # The same link is often referenced from many files, so lychee lists it once
+        # per file. Keep only the first occurrence of each link.
+        bullet_re = re.compile(r"^\* (?:\[[^\]]+\] )?<?([^\s>]+)>?")
+
+        seen = set()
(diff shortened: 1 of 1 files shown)


seisman:
It's easy to write a ~100-line script with Claude to process the output, but I feel it's too complex for us to maintain. 

Looking at the latest report in https://github.com/GenericMappingTools/gmt/issues/9139, maybe we should ignore timeout errors and 403 errors? I think we care about 404 errors the most.

Esteban82:
I could simplify this to a short bash pipeline instead:

```
- name: Deduplicate links in the report
  if: steps.lychee.outputs.exit_code != 0
  run: |
    {
      awk '/^## /{exit} {print}' /tmp/lychee-out.md
      echo
      echo "## Broken/timed-out links"
      echo
      grep '^\* \[' /tmp/lychee-out.md | sed -E 's/ \(at [^)]*\)//' | sort -u
    } > /tmp/lychee-dedup.md
    mv /tmp/lychee-dedup.md /tmp/lychee-out.md

```    
The trade-off is that the report becomes a flat, deduplicated list instead of being grouped per source file, so it wouldn't say which file(s) each broken link came from anymore.

Would that be better for you?

seisman:
> The trade-off is that the report becomes a flat, deduplicated list instead of being grouped per source file, so it wouldn't say which file(s) each broken link came from anymore.

The Python script has the same drawback because it reports only the first occurrence of a broken link.

I usually use `grep`/`ack` to find the files with broken links, so I feel it's not a big issue. In other words, I feel the bash script version is better than the Python version.

seisman (approved):

Esteban82:
Done — switched to the bash pipeline (grep/sed/sort -u) in the latest commit.

PR #9148, opened 2026-08-25
Let gmtread (externals only) also respect the -R option.

joa-quim:
It has to resort to the trick to call gmtselect -R.


Esteban82 (approved):
Blindly approve.

PR #9150, opened 2026-08-26
CI: Add workflow_dispatch trigger to workflows missing it

seisman:
Adds `workflow_dispatch:` to workflows so that we can run the workflows manually in the Actions tab.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Esteban82 (approved):
Looks fine.

PR #9151, opened 2026-08-26
filter1d: Add -N+e<cols> modifier to exclude columns from filtering

Esteban82:
`filter1d` filters every column except the independent one (`-N<t_col>`). This adds a `+e<cols>` modifier to `-N` so you can also exclude other columns from being filtered.

Motivating example, added to filter1d.rst: given a profile with columns X, Y, Distance, Elevation, filtering normally applies to every column except the independent one, which would also smooth X and Y. To filter only Elevation as a function of Distance while leaving X,Y untouched:

`gmt filter1d profile.txt -Fb100 -N2+e0,1 -E > profile_filt.txt`

**Done with Claude Opus 5**.


Esteban82:
Could someone please have a look at this PR? I’d like to use it. Thanks!

joa-quim (approved):

joa-quim (approved):

joa-quim:
I can't keep with the pace at which e-mails are arriving.

PR #9152, opened 2026-08-26
Fit 3-D subplot panels, and their heading, inside the figure (#4450)

### LINKED ISSUE #4450: 3D subplot title positioning doesn't account for perspective angle

The new [`subplot`](https://docs.generic-mapping-tools.org/latest/subplot.html) module is great, but the positioning of the figure title on 3D perspective plots seems to be using the 2D area and can overlap with the plotted content.

```
gmt begin map png
  gmt subplot begin 1x2 -Ff20c/10c+pblack+wblue -T"3D Subplots"
    gmt subplot set 0,0
    gmt basemap -BneSWZ+b -R0/10/20/30/40/50 -Jz1 -p157.7/45

    gmt subplot set 0,1
    gmt basemap -BSEnwZ2+b -R0/10/20/30/40/50 -Jz1 -p157.7/45
  gmt subplot end
gmt end show
```

**Actual outcome**

map

**Expected outcome**

Title position should be placed higher, i.e. above the 3D plots.

**System information**

* Operating system: Linux
* GMT version (`gmt --version`): 6.2.0_92c5676_2020.11.10

joa-quim:
A perspective panel is drawn as a projected box that is larger than the flat map rectangle subplot lays its panels out from, and nothing ever fitted that box to the panel.  In the reported example both cubes burst out of the -Ff20c/10c figure area and the heading was laid out against the flat rectangle, so it ended up among them.

gmt_map_setup could not have fitted the box earlier: it is only known once gmtmap_init_three_D has run, at the very end of the setup.  So the fit now happens there, for a panel in 3-D: the footprint is measured from the eight corners of the x/y/z region, padded for the ticks and annotations that the frame draws outside it (a quarter of that above the box, where nothing is annotated), and if it does not fit the panel the projection is scaled down and the box centered in the panel via z_project.x_off/y_off.  The scaling includes proj.zmax/zmin, the z-axis length in plot units that the PostScript perspective matrix works from; without those the numbers changed but the picture did not.

The heading is now placed at subplot end rather than at subplot begin, since only then is it known how far a 3-D panel reaches: begin records the position it would have used in gmt.subplotheading.<fig> together with the canvas corner (as PSL_SUBPLOT_ox/oy in the PostScript, so no assumption about -X -Y is needed), gmt_plotinit records the highest projected y of any 3-D panel in gmt.subplottop.<fig>, and end returns to that corner, raises the heading by any excess and puts the origin back for whatever follows.  The panel settings are cleared around that call, or the heading is scaled into the last panel, and restored afterwards since the -R history is built from them.

test/subplot/subplot_3d_fit.sh plots the example from the issue and requires the drawn height to stay near the 10 cm of the figure; the unfitted cubes reach some 19 cm.  It measures the plot instead of comparing it, so it needs no baseline PostScript.

Fix #4450

Assisted-by: Claude Opus 5.0 But it resisted like hell to fix this.

The issue's example now produces.
issue4450


Esteban82 (approved):
Works fine for me.

joa-quim:
These tests make no sense. Now all tests pass on Linux, a stray one on MacOS a few failures on Windows, non of them fail for me locally.

PR #9161, opened 2026-08-28
image: Fix the description of the -G option

seisman:
In `image`, the `-G` option can be repeated, e.g., `-Gblue+b -Gred+f`, while the `-G<color>+t` can not be repeated. So the original docstring is misleading.


doc/rst/source/image.rst
```diff
@@ -111,7 +111,9 @@ Optional Arguments
     - **+f** - For 1-bit images you can specify an alternate *color* for the foreground pixels.
       Give no color to make those pixels transparent.
     - **+t** - For color images you can select a single *color* that should be made transparent
-      instead. This option may be repeated with different settings.
+      instead.
+
+    This option may be repeated with different settings.
```
joa-quim: I see only a line break into a new paragraph. Is that the intention?

Esteban82 (approved):

PR #9131, opened 2026-08-19
Fix polar projection (-JP+t)

Esteban82:
**Done with Claude Sonnet 5**

The axis is now the correct length and the labels are displayed for `+t90`.

Tested with: 
```
gmt begin test_7059 png
    gmt basemap  -R-0.0001/0.0001/0/1 -JP12c     -Bya
    gmt basemap  -R-0.0001/0.0001/0/1 -JP12c+t45 -Bya
    gmt basemap  -R-0.0001/0.0001/0/1 -JP12c+t90 -Bya
gmt end
```

test_7059

Fixes #7059

PR #9170, opened 2026-08-30
Fix thin polar sectors that end up parallel to an axis (#7059)

### LINKED ISSUE #7059: Multiple -JP plots start at end of previous plot (well, sort off)
Background, I want to create a [radar plot](https://en.wikipedia.org/wiki/Radar_chart)

This is already odd. Origin of second axis is not 0,0 and its length is not 12c but rather 1/cos(45)

```
gmt psbasemap  -R-0.0001/0.0001/0/1 -JP12c -Bya -P -K > lixo.ps
gmt psbasemap  -R-0.0001/0.0001/0/1 -JP12c+t45 -Bya -O >> lixo.ps
```
lixo

and trying ``+t90`` hangs it or maybe its creating a such big PS that I aborted after a while.

joa-quim:
A single polar axis is a sector so thin that its bounding box has a width but no height.  Rotating it by 90 degrees swaps the two and broke both the scale and the annotations:

* gmtmap_setinfo matched -JP<size> to the map width, which for a sector standing on end is nothing but round-off, so -JP12c+t90 produced a plot 12 x 3437746 cm.  It now checks the dimension it was asked to match and falls back to the other one when the first is under 0.01 % of it, with a warning (and a pointer to -Jp<scale> for polar plots).  A fully degenerate map no longer divides by zero.
* n_lon_nodes/n_lat_nodes are derived from the map dimensions and rounded down to zero for a map with no width, so the boundary-walking loops in gmtlib_map_latcross/gmtlib_map_loncross never ran and no ticks or annotations were placed.  The three places that derive them now floor them at 2.

Adds test/psbasemap/polar_thin_sector.sh and documents the meaning of the -JP size for sectors, plus the -Jp recipe for radar plots.

Fix #7059

Supersedes #9131 

Assisted-by: GPT Terra + Claude Opus 5.0

### DIFF
--- a/doc/rst/source/reference/coordinate-transformations.rst
+++ b/doc/rst/source/reference/coordinate-transformations.rst
@@ -312,6 +312,24 @@ in the plane it is a **two-dimensional** projection. The transformation comes in
 #. The radius *r* can either be radius or inverted to mean depth from the surface,
    planetary radii, or even elevations in degrees.
 
+**NOTE**: With **-JP** the given dimension is the dimension of the *plot*, not the radius: a full circle is
+2\ *r* across, so **-JP**\ 12c draws a circle 12 cm wide, while the quarter sector **-R**\ 0/90/0/1 **-JP**\ 12c
+is also 12 cm wide but has a 12 cm radius. Since the bounding box of a sector changes when the sector is rotated
+with **+t**, the same sector given different **+t** values will come out at different radial scales, and a thin
+sector that ends up parallel to the *y*-axis has no width for the given dimension to set (GMT then sizes the height
+instead and warns). Furthermore, each plot is placed with the lower left corner of its bounding box at the current
+plot origin, so overlays that differ in **+t** line up by their bounding boxes and not by their *r* = 0 point. Use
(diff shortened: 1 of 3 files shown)


Esteban82:
I tested it and the test works fine. 

But I run the command of #7059 and I got this figure

lixo

I suspect that the origin of the second axis should be at 0.0.

joa-quim:
Damn it (thanks for testing).

Esteban82:
I tested further. With -45 works fine:

```
gmt begin radar png
	gmt basemap -R-0.0001/0.0001/0/1 -Jp12c -Bya
	gmt basemap -Jp12c+t-45 -Bya
	gmt basemap -Jp12c+t-90 -Bya
gmt end
```

radar




----
But fails with `+t45`.
radar


```
gmt begin radar png
	gmt basemap -R-0.0001/0.0001/0/1 -Jp12c -Bya
	gmt basemap -Jp12c+t45 -Bya
	gmt basemap -Jp12c+t-90 -Bya
gmt end
```

joa-quim:
It smells like the _thing_ has tailored the code to comply with the example given and not with a general solution.

PR #9172, opened 2026-09-01
Add RGBnoQUANT option to PS_COLOR_MODEL

### LINKED ISSUE #9171: Compatibility issue with low-bit-depth indexed RGB images in Adobe Illustrator
## Description

I encountered a color compatibility issue with PDFs generated by GMT.

The PDF renders correctly in Adobe Acrobat, and it also looks correct when placed as a linked PDF in Adobe Illustrator. However, when the same PDF is opened directly in Illustrator, or when the placed PDF is embedded, the raster colors are interpreted incorrectly.

In my case, blue colors become red and differences between neighboring colors are largely lost.

I have attached screenshots showing the difference between the correct rendering and the rendering after opening/embedding in Illustrator.

## Environment

- GMT: 6.6.0
- Ghostscript: 9.54.0
- Adobe Illustrator: 2025

## Observed behavior

The same GMT-generated PDF behaves differently depending on how it is handled:

- Adobe Acrobat: colors are correct
- Adobe Illustrator, Place as linked PDF: colors are correct
- Adobe Illustrator, Open PDF directly: colors are incorrect
- Adobe Illustrator, Place PDF and then Embed: colors become incorrect

This suggests that the PDF itself can be rendered correctly, but something goes wrong when Illustrator converts the PDF contents into editable internal objects.

## Possible relation to indexed-color images

After looking at the GMT source code, I noticed that `PSL_plotcolorimage()` in `src/postscriptlight.c` attempts to convert 24-bit RGB raster images with a small number of unique colors into indexed-color images.

The relevant path appears to be:

```text
PSL_plotcolorimage()
    -> psl_makecolormap()
    -> psl_bitreduce()
    -> Indexed /DeviceRGB image

RyanHaotianPang:

Adds `PS_COLOR_MODEL=RGBnoQUANT`, an opt-in compatibility setting for
third-party PostScript/PDF editors that incorrectly interpret low-bit-depth
indexed RGB raster images. It addresses #9171, where Adobe Illustrator may
change raster colors when opening or embedding a GMT-generated PDF.

`RGBnoQUANT` keeps the normal RGB color mode for all PostScript colors. Its
only difference is that 24-bit RGB raster images bypass the existing
indexed-color optimization and are emitted through the existing direct
`/DeviceRGB` output path. Default `PS_COLOR_MODEL=RGB` output, CPT
interpolation, and `grdimage` color calculations are unchanged.

The setting is parsed case-insensitively and `gmt get PS_COLOR_MODEL` reports
the canonical value `rgbnoquant`.

Fixes #9171

## Tests

- Added `test/postscriptlight/rgbnoquant.sh` with a four-color RGB PPM image.
  It verifies that normal RGB still emits `/Indexed /DeviceRGB` with reduced
  bits, while mixed-case `RGBnoQUANT` round-trips through `gmt get`, emits
  direct `/DeviceRGB`, and contains no indexed color space. It also checks
  that switching back to RGB restores indexed output.
- `ctest --test-dir build --output-on-failure -R
  'test/postscriptlight/rgbnoquant\\.sh'` (1/1 passed).
- Ghostscript renders of the normal-RGB and RGBnoQUANT regression outputs were
  pixel-identical.
- Manually verified the modified PDF resolves the Illustrator color issue.

41980b63fd58ea5ba58434400f064c2b

**Reminders**

- [x] If adding new functionality, add a detailed description to the documentation and/or an example.

### DIFF
--- a/doc/rst/source/gmt.conf.rst
+++ b/doc/rst/source/gmt.conf.rst
@@ -1091,8 +1091,13 @@ PostScript Parameters
         with standard fonts.
 
     **PS_COLOR_MODEL**
-        Determines whether PostScript output should use **RGB**, **HSV**, **CMYK**,
-        or **GRAY** when specifying color [default is **rgb**]. Note if **HSV**
+        Determines whether PostScript output should use **RGB**, **RGBnoQUANT**,
+        **HSV**, **CMYK**, or **GRAY** when specifying color [default is **rgb**].
+        **RGBnoQUANT** uses
+        normal RGB colors but forces direct RGB raster output instead of indexed
+        images. This may increase file size and is intended for compatibility with
+        third-party PostScript/PDF editors that have problems with indexed images.
+        Note if **HSV**
         is selected it does not apply to images which in that case uses **RGB**.
         When selecting **GRAY**, all colors will be converted to gray scale using
         YIQ (television) conversion.
--- a/src/gmt_defaults.h
+++ b/src/gmt_defaults.h
@@ -192,6 +192,7 @@ struct GMT_DEFAULTS {
 	double ps_transparency;			/* Later transparency [0] */
 	double ps_penwidth;			/* Current pen width */
 	unsigned int ps_color_mode;		/* Postscript encoding of color [PSL_RGB | PSL_CMYK | PSL_HSV | PSL_GRAY] */
+	bool ps_color_no_quantization;	/* Skip RGB image quantization and indexed-color output [false] */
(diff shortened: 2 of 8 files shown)


joa-quim (approved):
Thanks for this PR, that looks good. I tested it locally and no side effects were detected.

Esteban82:
Hi @RyanHaotianPang, thanks for the contribution! Did you use any AI assistant to help write this PR? If so, which one?

RyanHaotianPang:
Hi, thanks! @Esteban82   I used OpenAI Codex to assist with the implementation and testing.

PR #9175, opened 2026-09-03
Warn when a table module reads from an interactive terminal

Esteban82:
When a module that reads a data table (`blockmean, gmtinfo, psxy, triangulate, surface, mapproject`, etc.) is run with no input file and no piped/redirected data, GMT falls back to reading from standard input. If you run it directly in a terminal — often just because you forgot the input file — the command appears to hang, with no indication of what it's waiting for.

Now a message appears in that case:

```
gmt gmtinfo
gmtinfo [WARNING]: No input file given - reading table data from standard input
(this terminal). Type your data and press Ctrl-D when done. To use a file or a
pipe instead, press Ctrl-C now and rerun the command with that input.
```

It's currently a warning, so the command keeps waiting on stdin as before (you can still type data and finish with Ctrl-D). 

**Question:**
Would a hard error instead be preferred, so it exits immediately instead of continuing to wait?

**Assisted by Claude Opus 5**


joa-quim:
_gmtinfo_ is a particular module. All other mentioned here (_blockmean_, etc) print the online help when only the program names is typed. I think this would introduce unnecessary complication for very little gain to distinguish when options were passed but not the file name

Esteban82:
Fair point, `gmtinfo` was a bad example — it's the only one here with no required options, so it's the only one that reaches this code on a bare invocation.  

Tested with `blockmean`:

`gmt blockmean` → full module usage (as you said)
`gmt blockmean -I1` → "must specify -R option" error
`gmt blockmean -I1 -R0/10/0/10` → the new message. Only fires when the table is the sole thing missing.

seisman:
I'm also against such changes. 

If I understand it correctly, the error message only appears in interactive terminals. But in the real world, it's more common to run GMT commands from scripts, so the changes add little to GMT.

Esteban82:
To be clear, this is just a warning, not an error — nothing else changes.

What I'm targeting isn't someone typing into a terminal on purpose, but a script bug (typo, empty variable, broken pipe) that silently drops a module's input. If the script was launched from an interactive shell, the module then hangs on the inherited terminal with no way to tell "still computing" from "waiting for input". A warning fixes that without breaking anything, which is also why I went with a warning instead of an error.

Still fine closing this if you don't think it's worth it.

joa-quim:
I thinks it's better not to complicate what never has been reported as a problem.

Esteban82:
Ok, I will close it.

PR #9179, opened 2026-09-04
CI: Touch gmt_data_server.txt and gmt_hash_server.txt so they're not outdated

seisman:
As I understand it, GMT tries to refresh remote files if `gmt_data_server.txt` or `gmt_hash_server.txt` are 24-hours old.

The files stored in cache are refreshed weekly, so GMT thinks they're outdated. This PR fixes the issue by changing the modifications times of the two files, which has been working well in PyGMT.


Esteban82 (approved):
Looks fine.

PR #9183, opened 2026-09-07
Centralize info on character escape sequences in separate file

### LINKED ISSUE #7210: Documentation of character escape sequences in text module and Cookbook not ideal

Esteban82:
Continues #7221 by @KristofKoch, rebased onto master with conflicts resolved (caused by the `cookbook/` → `reference/` docs restructuring). Also fixed two stale `/cookbook/...` links left dangling by that same restructuring.

Fixes #7210
Supersedes #7221

**Assisted by Claude Sonnet 5**


joa-quim:
I guess this is only organizational and does not change how it looks in docs, right?

Esteban82:
Not exactly. 

In the `text` module, this:

> @a, @c, @e, @i, @n, @o, @s, @u, @A, @C, @E, @N, @O, and @U give various accented European characters, as indicated in Table [escape](https://docs.generic-mapping-tools.org/dev/reference/features.html#tbl-shorthand). Composite characters (overstrike) may be indicated with the @!<char1><char2> sequence, which will print the two characters on top of each other.

is replaced by the table from section 3.22 (Character escape sequences), embedded directly instead of linked to.

image

joa-quim (approved):

PR #9184, opened 2026-09-07
CI: Retry graphicsmagick choco install on Windows

Esteban82:
This PR wraps the `choco install` call in a retry loop (3 attempts, 15s apart) so a transient SourceForge timeout doesn't fail the whole run.

**Assisted by Claude Sonnet 5**


Esteban82:
Now all checks passed!!

joa-quim (approved):
If it works better ...

PR #9185, opened 2026-09-07
grdimage: Fix uninitialized inc[] read with variable transparency

Esteban82:
`test/grdimage/image_vartrans.sh` has been failing intermittently, on roughly half the runs. I have seen this myself running the same script on different occasions: sometimes panels 3 and 4 come out fine, and sometimes they are blank.

According to Claude the cause is an uninitialized variable. For images with a variable alpha channel, `GMT_grdimage` does `goto tr_image`, which jumps into the `if (need_to_project)` block past the declaration of `double inc[2] = {0.0, 0.0}`, so the initializer never runs. `gmt_project_init()` then branches on that leftover stack value and, when it happens to look positive, projects the image to 1x1 pixels — a blank panel, with no warning. Hence the coin-flip behavior.

This PR adds the fix proposed by **Claude Opus 5**.

### DIFF
--- a/src/grdimage.c
+++ b/src/grdimage.c
@@ -1454,6 +1454,11 @@ EXTERN_MSC int GMT_grdimage(void *V_API, int mode, void *args) {
 	unsigned int grid_registration = GMT_GRID_NODE_REG, try, row, col, mixed = 0, pad_mode = 0;
 	uint64_t node, k, kk, dim[GMT_DIM_SIZE] = {0, 0, 3, 0};
 	int error = 0, ret_val = GMT_NOERROR, ftype = GMT_NOTSET, ftype2 = GMT_NOTSET;
+	/* Dimensions and increments of the projected grid/image.  These must be declared (and initialized) here at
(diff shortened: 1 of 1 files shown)


src/grdimage.c
```diff
@@ -1454,6 +1454,11 @@ EXTERN_MSC int GMT_grdimage(void *V_API, int mode, void *args) {
 	unsigned int grid_registration = GMT_GRID_NODE_REG, try, row, col, mixed = 0, pad_mode = 0;
 	uint64_t node, k, kk, dim[GMT_DIM_SIZE] = {0, 0, 3, 0};
 	int error = 0, ret_val = GMT_NOERROR, ftype = GMT_NOTSET, ftype2 = GMT_NOTSET;
+	/* Dimensions and increments of the projected grid/image.  These must be declared (and initialized) here at
```
seisman: I feel AI adds too many unnecessary comments to the codes

Esteban82: Yes, I thought that too.

Esteban82:
I simplified the comment.

seisman (approved):

PR #9187, opened 2026-09-08
CMake: Normalize installation path on Windows to avoid mixing slashes and backslashes

seisman:
Previously, we had issues building GMT on Windows with errors like:
```
CMake Error at cmake_install.cmake:5 (set):
  Syntax error in cmake code at

    D:/a/gmt/gmt/build/cmake_install.cmake:5

  when parsing string

    D:\a\gmt\gmt/gmt-install-dir

  Invalid character escape '\a'.
```
The real cause is that Windows uses backslashes (`D:\a\gmt\gmt`) while CI (and CMake) uses slashes.

PR https://github.com/GenericMappingTools/gmt/pull/9097 fixed the issue by applying a workaround to convert backslashes to slashes. It works for CI, but users who build GMT locally (via WSL or similar) may still have the same issue.

This PR fixes the root issue by normalizing paths on Windows using CMake's `TO_CMAKE_PATH` directive (xref: https://cmake.org/cmake/help/latest/command/file.html#to-cmake-path). With the fix, the CI workflows can be simplified.

Helped with Claude Opus 5.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -118,15 +118,10 @@ jobs:
         run: |
           mkdir build
           cd build
-          setlocal enabledelayedexpansion
-          set "INSTALLDIR=%INSTALLDIR:\=/%"
-          set "WINVCPKGROOT=%VCPKG_INSTALLATION_ROOT:/=\%"
-          cmake -G Ninja .. -DCMAKE_TOOLCHAIN_FILE=!WINVCPKGROOT!\scripts\buildsystems\vcpkg.cmake -DCMAKE_BUILD_TYPE=Release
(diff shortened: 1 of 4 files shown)


seisman:
@joa-quim Could you please check if it works for you on Windows locally?

joa-quim:
Hmm, I don't see what I can test locally. I don't use those recipes to build GMT. I have a ConfigUser.cmake to set all that I need.

seisman:
This PR adds the following lines to `cmake/modules/ConfigCMake.cmake`, so it should affect your local builds:
```
# Normalize the installation prefix on Windows (avoid mixing slashes and backslashes).
if (WIN32)
	file (TO_CMAKE_PATH "${CMAKE_INSTALL_PREFIX}" CMAKE_INSTALL_PREFIX)
endif (WIN32)
```

joa-quim:
Rebuilt and installed locally and no issues.

Esteban82 (approved):

PR #9191, opened 2026-09-09
CI: Combine config-gmt-unix.sh and config-gmt-windows.sh into one single config script

seisman:
Combine `config-gmt-unix.sh` and `config-gmt-windows.sh` into `config-gmt.sh`, so that we have one fewer file to maintain.


Esteban82 (approved):
Looks fine.

PR #9192, opened 2026-09-09
inset: Fix gmt inset end not removing its own tagged history/settings files

Esteban82:
## Summary

`gmt inset end` is supposed to delete two temporary files (`gmt.history.inset`, `gmt.conf.inset`) it created when the inset began, so that a later inset in the same figure does not inherit state from a previous, already-closed one. 

## Reproduction

```bash
gmt begin inset_leak png
	gmt basemap -JM7c -R-70/-60/-56/-50 -Baf
	gmt inset begin -DjTR+w2.2c+o-0.08c
		gmt set MAP_FRAME_PEN thick,red
		gmt basemap -Rg -JG-67/-54/? -B0
	gmt inset end
	gmt inset begin -DjBL+w3.5c
		gmt basemap -B0
	gmt inset end
gmt end show
```

BUG_Inset_State_Leak

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/inset.c
+++ b/src/inset.c
@@ -420,11 +420,11 @@ EXTERN_MSC int GMT_inset (void *V_API, int mode, void *args) {
 
 		/* Remove the inset history file */
 		gmt_hierarchy_tag (API, GMT_HISTORY_FILE, GMT_OUT, tag);
-		sprintf (ffile, "%s/%s.%s", API->gwf_dir, GMT_HISTORY_FILE, tag);
+		sprintf (ffile, "%s/%s%s", API->gwf_dir, GMT_HISTORY_FILE, tag);
 		gmt_remove_file (GMT, ffile);
 		/* Remove the gmt settings file */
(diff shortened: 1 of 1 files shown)


joa-quim (approved):
Looks good. Were there files with 2 dots in extension?

In future please say which Claude version. Sonnet is quite low trust.

Esteban82:
Reviewed with Claude Opus 5.

And GMT does create those 2 files (gmt.history.inset and gmt.conf.inset), if that's what you mean.

joa-quim:
Well, my comment was based on seeing what C code changes are. From it I deduced that created files were something like `gmt.history..inset` or `gmt..history.inset`

Esteban82:
Other way around: tag already has the leading dot ([gmt_init.c:3480](https://github.com/GenericMappingTools/gmt/blob/master/src/gmt_init.c#L3480)). The buggy format added a second one, so inset end was trying to delete `gmt.history..inset` / `gmt.conf..inset` — paths that never existed.

PR #9193, opened 2026-09-10
doc: Fix macOS test failure in GMT_App_M_1c.sh (locale-dependent sort)

Esteban82:
Claude Opus suggests this edit to fix the macOS-only test failure in `GMT_App_M_1c.sh` (and the same latent bug in 1a/1b/1d).

**Cause**: the colorbar order comes from a locale-dependent `sort` over a mixed-case CPT name list, and the baseline image was made under a different collation than the macOS runner uses — so the colorbars land in the wrong slots and the image comparison fails (RMS 0.0852). Pinning `LC_ALL=C` on the `sort` fixes it.

I don't have a macOS machine, so I could only reproduce the failure on Linux by forcing the same locale mismatch; I'm relying on this PR's CI run to confirm it actually fixes macOS.

**Assisted-by: Claude Opus 5**

### DIFF
--- a/doc/scripts/GMT_App_M_1a.sh
+++ b/doc/scripts/GMT_App_M_1a.sh
@@ -30,7 +30,7 @@ plot () {
 GMT_SHAREDIR=$(gmt --show-sharedir)
 
 # Here we list all cpt, except cyclic, categorical, cmocean, SCM, srtm
-sed -e 's/"//g' "${GMT_SOURCE_DIR}"/src/gmt_cpt_masters.h | egrep -vi "cyclic|categorical|cmocean|SCM|srtm" | awk '{print $1}' | sort > tt.lis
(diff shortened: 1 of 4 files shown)


seisman:
**Off-topic**: Do we really need to add the "AI-assistant" label manually? I feel most of recent PRs (if not all) are assistanted by AIs.

seisman (approved):

Esteban82:
> **Off-topic**: Do we really need to add the "AI-assistant" label manually?

Are you suggesting we automate this? Or stop using them altogether?

seisman:
Just stop using the label, assuming that all PRs are done with some help from AI.

Esteban82:
The label isn't really about crediting the tool — the point is to make AI-assisted PRs/issues quickly filterable. If something turns out to be wrong down the line, we can pull up exactly that filtered set and give it a closer, more skeptical look, rather than treating everything as equally suspect.

joa-quim:
I also think that we should drop that label. If they are all AI assisted there is no point in repeating it all the time. And, in case of need, we have the commit message that should have what model/version was used.

Esteban82:
Okay, let's drop the label.

PR #9194, opened 2026-09-10
CI: Don't leave required checks pending on documentation-only PRs

Esteban82:
Documentation-only PRs leave the required checks (`Linux`, `macOS`, `Windows`, `Linux (without optional dependencies)`, `Code Validator`) stuck at "Expected — Waiting for status to be reported", which blocks the merge: with `paths:` on the trigger the workflow never runs, so no status is ever reported for those names. The filtering now happens in a small `changes` job and the real job is skipped via `if:` instead, which does report a status and satisfies branch protection.

Assisted-by: Claude Sonnet 5 (implementation) and Claude Opus 5 (review)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,19 +8,7 @@ on:
     branches:
       - master
       - 6.[0-9]+
-    paths:
-      - 'ci/**'
-      - 'cmake/**'
-      - 'src/**'
-      - '**/CMakeLists.txt'
-      - '.github/workflows/build.yml'
   pull_request:
-    paths:
-      - 'ci/**'
-      - 'cmake/**'
-      - 'src/**'
-      - 'CMakeLists.txt'
-      - '.github/workflows/build.yml'
   workflow_dispatch:
 
 defaults:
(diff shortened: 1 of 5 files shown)


seisman:
Make similar changes in docker.yml and docs.yml?

Esteban82:
> Make similar changes in docker.yml and docs.yml?

Ok, I will do it tomorrow.

Esteban82:
Added the same pattern to `docker.yml`, plus `pull-requests: read` — it has its own `permissions:` block, which otherwise defaults every unlisted scope to `none` and would have made `paths-filter` fail on every PR.

Left `docs.yml` out: it has no `pull_request` trigger, so it never posts a status on a PR and isn't part of the pending-checks problem this PR fixes. Gating it the same way would only add risk on the `release` event — if the `changes` job errored, the `docs` job would be skipped rather than failed, silently dropping the doc build, the gh-pages deploy, and the release package.

@seisman do you still want the same change in `docs.yml` despite that, or is it fine to leave it as is?

seisman (approved):

PR #9201, opened 2026-09-12
ci: Remove stale `gcov` input from codecov-action step

Esteban82:
`gcov: true` has been ignored since codecov-action v4.0.0 moved to the Codecov CLI, where gcov runs by default — it only produces an "Unexpected input(s)" warning on every OS. Coverage collection is unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Sonnet 5, reviewed with Opus 5)


Esteban82:
The gcov warning is gone (confirmed on all three OS).

seisman (approved):

PR #9092, opened 2026-07-07
Preserve selected data server in movie frames

Esteban82:
**Done with Claude Sonnet 5**

Fix #8209.

When running gmt movie, the selected GMT_DATA_SERVER was not consistently preserved inside the frame sessions. As a result, gmt --show-dataserver could report a different server in the movie script than the one configured in the parent session.

This PR ensures that the active data server setting is propagated correctly to movie frame sessions.

**Fix**: swapped the order at all 6 locations where scripts are generated (preflight, postflight, intro title-frame loop x2, main script x2), so `cp -f gmt.conf .` now runs before `gmt begin`.
### Test

The issue can be reproduced with:

```
gmt set GMT_DATA_SERVER oceania
gmt --show-dataserver
cat << 'EOF' > main.sh
gmt begin
	gmt --show-dataserver
	gmt basemap -R-114/-103/35/40 -JM10c -Baf
gmt end
EOF
gmt movie main.sh -Chd -Nanim02 -T1 -Zs -Ml,png
```
Expected result: the same data server name should be printed twice (once in the terminal and once from inside the movie script). For example:
```
http://oceania.generic-mapping-tools.org
http://oceania.generic-mapping-tools.org
```

PR #9207, opened 2026-09-14
Pin velo vector scale to explicit cm unit in geodesy_08 test

### LINKED ISSUE #9204: Something broke the test/geodesy/geodesy_08.sh test
As it says.

Esteban82:
#9092 fixed gmt movie to correctly propagate the top directory's gmt.conf into each frame before "gmt begin" runs. As a side effect, movie frames now inherit PROJ_LENGTH_UNIT from the test harness's gmt.conf (inch) instead of silently falling back to GMT's SI default (cm). The velo scale in "-Se0.50/0" had no explicit unit, so it now resolves against the wrong default and renders vectors ~2.5x too long, same class of issue already patched for trajectory2/3/5 in #9092 itself.

Fixes #9204

**Done with Claude Sonnet 5.**


joa-quim (commented):
Hmm, I guess I don't understand this. both `-Se0.50c/0` and `-Se0.50/0` should give the same result (default units in GMT is cm). If it doesn't, that sounds like it is hiding a more serious issue.

Esteban82:
This is Claude's response: 

GMT's own default is cm, but the test harness overrides it: `test/gmtest.in` runs `gmt set -Du GMT_GRAPHICS_FORMAT ps` for these tests, which writes `PROJ_LENGTH_UNIT=inch` into `gmt.conf`. Before #9092, movie frames never saw that file (the `cp` happened after `gmt begin`); now they do, so unitless lengths like `-Se0.50/0` resolve against inch instead of cm.

Esteban82:
[-Du](https://docs.generic-mapping-tools.org/dev/gmtset.html#d) stands for US units, so that's why `PROJ_LENGTH_UNIT` was set to inches.

joa-quim (approved):
Ok, Ok.

PR #9215, opened 2026-09-18
histogram: Add +ol|c|r alignment directives to -E

### LINKED ISSUE #9188: histogram: Support aligning bars at the right edge of bins
In `histogram`, by default, bars are aligned at the left edge of the bins. With `-F` option, bars are centered in the bins. 

The behavior can be confirmed below:
```
gmt begin map png
    gmt histogram  @v3206_06.txt -R-7000/1000/0/2000 -T250 -W0.25p -Baf250 -Gblue
    gmt histogram  @v3206_06.txt -R-7000/1000/0/2000 -T250 -W0.25p -Baf250 -Gblue -F -Xw+1c
gmt end show
```
Image

What's missing is aligning the bars at the right edge of the bins. For comparison, matplotlib's `hist` method provides the `algin` parameter which can be "left", "right", and "mid".

It would be great if GMT's histogram can also support it. For backward compatibility, I expect the following syntax:

- `-Fc` or `-F`: centering
- `-Fl`: left alignment (the default without -F)
- `-Fr`: right alignment

Esteban82:
Adds shorthand directives to the **+o** modifier of **-E** so a bar can be placed relative to the bin center without computing the offset by hand: **+oc** (centered, the default), **+ol** (= `+o-<width>/2`) and **+or** (= `+o+<width>/2`), as proposed in the issue. 

Is this how you wanted @Seisman to work?

Tested with:

```bash
gmt begin grouped png
  gmt histogram @v3206_06.txt -T-6000/0/1000 -R-6000/0/0/0 -JX12c/7c -Gred  -W0.5p -E500+ol -Baf1000
  gmt histogram @v3206_06.txt -T-6000/0/1000 -R-6000/0/0/0 -JX12c/7c -Gblue -W0.5p -E500+or
gmt end
```
grouped

Fixes #9188

Implemented with Claude Sonnet 5, reviewed with Claude Opus 5.


src/pshistogram.c
```diff
@@ -649,7 +654,7 @@ static int parse (struct GMT_CTRL *GMT, struct PSHISTOGRAM_CTRL *Ctrl, struct GM
 	unsigned int n_errors = 0, mode = 0, pos = 0;
 	int sval;
 	size_t L;
-	char *c = NULL, *l_arg = NULL, *t_arg = NULL, *w_arg = NULL, p[GMT_BUFSIZ] = {""};
+	char *c = NULL, *l_arg = NULL, *t_arg = NULL, *w_arg = NULL, p[GMT_BUFSIZ] = {""}, align = '\0';
```
joa-quim: I don't think this is good C.
char *c = NULL, *l_arg = NULL, *t_arg = NULL, *w_arg = NULL, p[GMT_BUFSIZ] = {""}, align = NULL;
should be better (I think)

Esteban82: Ok, I will take a look.

joa-quim: Sorry, I meant
*align = NULL

Esteban82: Fixed. `align` is now `char *.`

joa-quim (approved):

PR #9216, opened 2026-09-18
Fix bugs in the Equal Earth projection implementation (bad lon_0's)

joa-quim:
Existing test will now fail because previous baseline was wrong.

Discussed with Claude Opus 5.0


joa-quim:
BTW, this now works

```
gmt coast -Rg -W0.5 -Bag -J+proj=eqearth+width=15 -png lixo
```

Esteban82 (approved):
It works for me. I got this map

Image

PR #9218, opened 2026-09-21
histogram: Fix bug with the -N normal curve for reverse cumulative histograms (-Qr)

Esteban82:
This fails:

`gmt histogram @v3206_06.txt -T-7000/0/250 -Z0 -JX12c/8c -Baf -Ggray -W0.5p -N0+p2p,red -Qr -png lixo`

Output with bug:

lixo_bug

Output with this fix:

lixo_fix

**Done with Claude Sonnet 5 and review by Opus 5.**

This PR also adds a test asserting that the reported area equals the number of input points for both -Q and -Qr; it fails on the previous code for -Qr.


joa-quim (approved):
Looks good

